PR #8874, opened 2026-01-08
[fix] app server flaky send_messages test

celia-oai:
Fix flakiness of CI test: https://github.com/openai/codex/actions/runs/20350530276/job/58473691434?pr=8282

This PR does two things:
1. move the flakiness test to use responses API instead of chat completion API
2. make mcp_process agnostic to the order of responses/notifications/requests that come in, by buffering messages not read

### DIFF
--- a/codex-rs/app-server/tests/common/mcp_process.rs
+++ b/codex-rs/app-server/tests/common/mcp_process.rs
@@ -611,6 +611,10 @@ impl McpProcess {
         Ok(notification)
     }
 
+    pub fn clear_message_buffer(&mut self) {
+        self.pending_messages.clear();
+    }
+
     async fn read_stream_until_message<F>(&mut self, predicate: F) -> anyhow::Result<JSONRPCMessage>
     where
         F: Fn(&JSONRPCMessage) -> bool,
--- a/codex-rs/app-server/tests/common/mcp_process.rs
+++ b/codex-rs/app-server/tests/common/mcp_process.rs
@@ -611,10 +611,16 @@ impl McpProcess {
         Ok(notification)
     }
 
+    /// Clears any buffered messages so future reads only consider new stream items.
+    ///
+    /// We call this when e.g. we want to validate against the next turn and no longer care about
+    /// messages buffered from the prior turn.
     pub fn clear_message_buffer(&mut self) {
         self.pending_messages.clear();
     }
 
+    /// Reads the stream until a message matches `predicate`, buffering any non-matching messages
+    /// for later reads.
     async fn read_stream_until_message<F>(&mut self, predicate: F) -> anyhow::Result<JSONRPCMessage>
     where
         F: Fn(&JSONRPCMessage) -> bool,
--- a/codex-rs/app-server/tests/common/mcp_process.rs
+++ b/codex-rs/app-server/tests/common/mcp_process.rs
@@ -60,7 +60,7 @@ pub struct McpProcess {
     process: Child,
     stdin: ChildStdin,
     stdout: BufReader<ChildStdout>,
-    pending_user_messages: VecDeque<JSONRPCNotification>,
+    pending_messages: VecDeque<JSONRPCMessage>,
 }
 
 impl McpProcess {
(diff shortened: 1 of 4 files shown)


codex-rs/app-server/tests/common/mcp_process.rs
```diff
@@ -647,7 +647,7 @@ impl McpProcess {
                     anyhow::bail!("unexpected JSONRPCMessage::Error: {message:?}");
                 }
                 JSONRPCMessage::Response(_) => {
-                    anyhow::bail!("unexpected JSONRPCMessage::Response: {message:?}");
+                    // Some callers expect notifications to race with responses; keep scanning.
                 }
```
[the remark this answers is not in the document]
celia-oai: I guess this is a real critique, fixing

codex-rs/app-server/tests/suite/codex_message_processor_flow.rs
```diff
@@ -430,6 +430,7 @@ async fn test_send_user_turn_updates_sandbox_and_cwd_between_turns() -> Result<(
         mcp.read_stream_until_notification_message("codex/event/task_complete"),
     )
     .await??;
+    mcp.clear_message_buffer();
```
celia-oai, line 433: this is needed because now we buffer all messages that come in & we don't care about previous turn's buffered messages anymore

codex-rs/app-server/tests/common/mcp_process.rs
```diff
@@ -637,46 +596,58 @@ impl McpProcess {
...
+            })
+            .await?;
+
+        let JSONRPCMessage::Notification(notification) = message else {
+            unreachable!("expected JSONRPCMessage::Notification, got {message:?}");
+        };
+        Ok(notification)
+    }
+
+    pub fn clear_message_buffer(&mut self) {
```
bolinfest, line 618: Can you please add a docstring to this method so users know when they should invoke it?

codex-rs/app-server/tests/suite/v2/turn_start.rs
```diff
@@ -554,6 +554,7 @@ async fn turn_start_updates_sandbox_and_cwd_between_turns_v2() -> Result<()> {
         mcp.read_stream_until_notification_message("codex/event/task_complete"),
     )
     .await??;
+    mcp.clear_message_buffer();
```
bolinfest, line 557: The only thing that concerns me is that it may not be obvious when this is necessary.

Do these tests still work if this is not here? If so, perhaps we should remove this API?

codex-rs/app-server/tests/suite/send_message.rs
```diff
@@ -26,13 +25,21 @@ const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs
 
 #[tokio::test]
 async fn test_send_message_success() -> Result<()> {
-    // Spin up a mock completions server that immediately ends the Codex turn.
+    // Spin up a mock responses server that immediately ends the Codex turn.
```
bolinfest, line 28: We need to migrate everything to responses since we're deprecating chat next month, right?

bolinfest (approved):
Approved, though I feel iffy about about `clear_message_buffer()`.

codex-rs/app-server/tests/suite/send_message.rs
celia-oai: yes! planning to do it in a follow-up PR

codex-rs/app-server/tests/common/mcp_process.rs
celia-oai: yup sg - imo this helper is fine as long as users know that what they are testing with this helper function

codex-rs/app-server/tests/suite/v2/turn_start.rs
celia-oai: these tests fail because it will be validating events buffered from the first turn. I'll explain in the docstring when this should be invoked - it's 100% dependent on what the user wants to test. Since tests fail without this invocation it should be pretty clear when it's needed

PR #8832, opened 2026-01-07
[please leave open, collaborating with mbolin] WIP: bazelification

dzbarsky:
This sets up remote builds + tests. To execute them:

1. [Create a buildbuddy account](https://app.buildbuddy.io/)
2. [Copy your API key](https://app.buildbuddy.io/docs/setup/) to ~/.bazelrc
3. `bazel test //... --config=remote --platforms=//:rbe -k`

PR #8504, opened 2025-12-24
feat: support local development using Buck2

bolinfest:
This is an experiment to get the Rust portion of the Codex repo setup to build with [Buck2](https://buck2.build).

The motivation is to get build time benefits, both for local/incremental builds as well as in CI, though of note:

- For reference, general advantages of Buck2: https://buck2.build/docs/getting_started/what_is_buck2/#why-use-buck2-key-advantages
- Locally, Buck2 runs a daemon, which helps speed up incremental builds.
- Using Buck2 means we can ultimately leverage _distributed_ builds
- Because Buck2 is stricter on running builds and tests in an "isolated" environment, it holds us to a higher standard to ensure builds are hermetic
- Buck2 has a complementary project, Reindeer, which is used to create `BUCK` build files for existing crates: https://github.com/facebookincubator/reindeer (I am not aware of a comparable offering for Bazel)

Unfortunately, `buck2 test` does not currently cache test results as Bazel does:

https://github.com/facebook/buck2/issues/183

This is pretty important, as we want incremental invocations of:

```shell
./scripts/buck2 test //codex-rs/a
```

to be as fast as possible, so I'll follow up on this.

PR #8875, opened 2026-01-08
feat: add support for building with Bazel

zbarsky-openai:
This PR configures Codex CLI so it can be built with [Bazel](https://bazel.build) in addition to Cargo. The `.bazelrc` includes configuration so that remote builds can be done using [BuildBuddy](https://www.buildbuddy.io).

If you are familiar with Bazel, things should work as you expect, e.g., run `bazel test //... --keep-going` to run all the tests in the repo, but we have also added some new aliases in the `justfile` for convenience:

- `just bazel-test` to run tests locally
- `just bazel-remote-test` to run tests remotely (currently, the remote build is for x86_64 Linux regardless of your host platform). Note we are currently seeing the following test failures in the remote build, so we still need to figure out what is happening here:

```
failures:
    suite::compact::manual_compact_twice_preserves_latest_user_messages
    suite::compact_resume_fork::compact_resume_after_second_compaction_preserves_history
    suite::compact_resume_fork::compact_resume_and_fork_preserve_model_history_view
```

- `just build-for-release` to build release binaries for all platforms/architectures remotely

To setup remote execution:
- [Create a buildbuddy account](https://app.buildbuddy.io/) (OpenAI employees should also request org access at https://openai.buildbuddy.io/join/ with their `@openai.com` email address.)
- [Copy your API key](https://app.buildbuddy.io/docs/setup/) to `~/.bazelrc` (add the line `build --remote_header=x-buildbuddy-api-key=YOUR_KEY`)
- Use `--config=remote` in your `bazel` invocations (or add `common --config=remote` to your `~/.bazelrc`, or use the `just` commands)

## CI

In terms of CI, this PR introduces `.github/workflows/bazel.yml`, which uses Bazel to run the tests _locally_ on Mac and Linux GitHub runners (we are working on supporting Windows, but that is not ready yet). Note that the failures we are seeing in `just bazel-remote-test` do not occur on these GitHub CI jobs, so everything in `.github/workflows/bazel.yml` is green right now.

The `bazel.yml` uses extra config in `.github/workflows/ci.bazelrc` so that macOS CI jobs build _remotely_ on Linux hosts (using the `docker://docker.io/mbolin491/codex-bazel` Docker image declared in the root `BUILD.bazel`) using cross-compilation to build the macOS artifacts. Then these artifacts are downloaded locally to GitHub's macOS runner so the tests can be executed natively. This is the relevant config that enables this:

```
common:macos --config=remote
common:macos --strategy=remote
common:macos --strategy=TestRunner=darwin-sandbox,local
```

Because of the remote caching benefits we get from BuildBuddy, these new CI jobs can be extremely fast! For example, consider these two jobs that ran all the tests on Linux x86_64:

- Bazel 1m37s https://github.com/openai/codex/actions/runs/20861063212/job/59940545209?pr=8875
- Cargo 9m20s https://github.com/openai/codex/actions/runs/20861063192/job/59940559592?pr=8875

For now, we will continue to run both the Bazel and Cargo jobs for PRs, but once we add support for Windows and running Clippy, we should be able to cutover to using Bazel exclusively for PRs, which should still speed things up considerably. We will probably continue to run the Cargo jobs post-merge for commits that land on `main` as a sanity check.

Release builds will also continue to be done by Cargo for now.

Earlier attempt at this PR: https://github.com/openai/codex/pull/8832
Earlier attempt to add support for Buck2, now abandoned: https://github.com/openai/codex/pull/8504


codex-rs/core/src/shell_snapshot.rs
```diff
@@ -361,13 +360,17 @@ mod tests {
...
         let deadline = Instant::now() + TokioDuration::from_secs(1);
         loop {
-            let kill_status = StdCommand::new("kill")
-                .arg("-0")
-                .arg(pid.to_string())
-                .stderr(Stdio::null())
-                .stdout(Stdio::null())
-                .status()?;
-            if !kill_status.success() {
+            let alive = unsafe { libc::kill(pid, 0) };
```
zbarsky-openai: TODO: do not land like this! either switch to `nix` or go back to `kill` binary once it's in the image

.gitignore
```diff
@@ -1,3 +1,4 @@
+bazel-*
```
zbarsky-openai: TODO: remove once https://github.com/openai/codex/pull/8911 merges

codex-rs/core/tests/suite/prompt_caching.rs
```diff
@@ -1,6 +1,8 @@
 #![allow(clippy::unwrap_used)]
 
+use codex_apply_patch::APPLY_PATCH_TOOL_INSTRUCTIONS;
```
bolinfest, line 3: Will go away once https://github.com/openai/codex/pull/8914 is in.

.github/workflows/Dockerfile.bazel
```diff
@@ -0,0 +1,53 @@
+FROM ubuntu:24.04
+
+# TODO(mbolin): Published to docker.io/mbolin491/codex-bazel:latest for
+# initial debugging, but we should publish to a more proper location.
+#
+# docker buildx create --use
+# docker buildx build --platform linux/amd64 -f .github/workflows/Dockerfile.bazel -t mbolin491/codex-bazel:latest --push .
+
+ARG DEBIAN_FRONTEND=noninteractive
+# enable 'universe' because musl-tools & clang live there
```
zbarsky-openai: I don't think we need any compilers in the image, they come from our bazel toolchains

.github/workflows/Dockerfile.bazel
```diff
@@ -0,0 +1,53 @@
...
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    software-properties-common && \
+    add-apt-repository --yes universe
+
+# now install build deps
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    build-essential curl git python3 ca-certificates \
+    pkg-config clang musl-tools libssl-dev just && \
```
zbarsky-openai: don't need libssl-dev, it comes from bazel and is built hermetically :)

.github/workflows/Dockerfile.bazel
```diff
@@ -0,0 +1,53 @@
...
+    rm -rf /var/lib/apt/lists/*
+
+# Install Bazel via Bazelisk (mirrors bazelbuild/setup-bazelisk@v3).
+ARG BAZELISK_VERSION=latest
```
zbarsky-openai: wait is this image for remote execution? Or to use in the github runner? (remote execution image doesn't need bazelisk/bazel in it)

.github/workflows/Dockerfile.bazel
```diff
@@ -0,0 +1,53 @@
...
+    chmod +x /usr/local/bin/bazelisk && \
+    ln -s /usr/local/bin/bazelisk /usr/local/bin/bazel
+
+# Install dotslash.
+RUN curl -LSfs "https://github.com/facebook/dotslash/releases/download/v0.5.8/dotslash-ubuntu-22.04.$(uname -m).tar.gz" | tar fxz - -C /usr/local/bin
+
+# Ubuntu 24.04 ships with user 'ubuntu' already created with UID 1000.
+USER ubuntu
+
+# install Rust + musl target as dev user
```
zbarsky-openai: we don't need to install rust either

bolinfest: Ah, I started from https://github.com/openai/codex/blob/main/.devcontainer/Dockerfile since I thought that would be a good basis, but interesting that we can pare this down!

bolinfest: Though maybe I should port some of this stuff to `.devcontainer/Dockerfile`...

zbarsky-openai: yeah I think we should do separate images; this one can be quite slim. I think we basically need:
- coreutils
- kill
- git
- dotslash

maybe 1 or 2 other minor things

zbarsky-openai: having a smaller image will load it faster to the executors and reduce chance of actions relying on things in the image and thus working remotely but not locally

.github/workflows/Dockerfile.bazel
```diff
@@ -0,0 +1,29 @@
...
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    software-properties-common && \
+    add-apt-repository --yes universe
+
+# now install build deps
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    build-essential curl git python3 ca-certificates \
+    pkg-config clang musl-tools libssl-dev && \
```
zbarsky-openai: can we try to remove these?
- pkg-config
- clang
- musl-tools
- libssl-dev
- (maybe, not sure) build-essential?

sorry to be a pain, but clang in particular is huge and i want to make sure we have no compilers in the base image so they're not used by accident :)

bolinfest: I mentally associate `pkg-config` with "python stuff," but chat says ends up being needed by some common Python packages that contain `C` extensions, so I agree that the builder host should not rely on those. Will try removing it.

zbarsky-openai: in our case it's likely used by rust build scripts to find C libs on the host to link against. But we don't want that, we want the libs provided hermetically through bazel, as otherwise cross-compilation won't work

defs.bzl
```diff
@@ -37,19 +37,40 @@ def codex_rust_crate(
...
+    """Defines a Rust crate with library, binaries, and tests wired for Bazel + Cargo parity.
+
+    The macro mirrors Cargo conventions: it builds a library when `src/` exists,
+    wires build scripts, exports `CARGO_BIN_EXE_*` for integration tests, and
+    creates unit + integration test targets. Dependency buckets map to the
+    Cargo.lock resolution in `@crates`.
+
+    Args:
+        name: Bazel target name for the library (and test target prefix).
+        crate_name: Cargo crate name passed to rust_library.
```
bolinfest: Let's add examples for these, so this would match the `name` in `Cargo.toml`, correct?

defs.bzl
```diff
@@ -37,19 +37,40 @@ def codex_rust_crate(
...
+        crate_features: Cargo features to enable for this crate.
```
bolinfest, line 57: Note that today we only support one permutation of features that is used across the workspace. Maybe note this is potentially suboptimal for crates with features like `test-support`? That is, we can't ship a Codex CLI built from Bazel today because it would include stuff like that.

defs.bzl
```diff
@@ -37,19 +37,40 @@ def codex_rust_crate(
...
+        crate_srcs: Optional explicit srcs; defaults to `src/**/*.rs`.
+        crate_edition: Rust edition override, if not default.
```
bolinfest, line 62: This feels like a bug if we set this...

codex-rs/docs/bazel.md
```diff
@@ -0,0 +1,48 @@
...
+In some cases, an upstream crate may need a patch or a `crate.annotation` in `../MODULE.bzl`
+to have it build in Bazel's sandbox or make it cross-compilation-friendly. If you see issues,
+feel free to ping zbarsky or mbolin.
+
+When you add a new crate or binary:
+
+1. Add it to the Cargo workspace as usual.
+2. Create a `BUILD.bazel` that calls `codex_rust_crate` (see nearby crates for
+   examples).
+3. If a dependency needs special handling (compile/runtime data, additional binaries
```
bolinfest, line 37: Should probably mention `no-sandbox` somewhere in this file because it's a bit buried in the other one?

We should also recommend that it is used sparingly, so consider moving the tests that require a sandbox into a separate crate?

defs.bzl
```diff
@@ -37,19 +37,40 @@ def codex_rust_crate(
...
+        build_script_data: Data files exposed to the build script at runtime.
+        compile_data: Non-Rust compile-time data for the library target.
+        deps_extra: Extra normal deps beyond @crates resolution.
+            Typically only needed when features add additional deps.
+        integration_deps_extra: Extra deps for integration tests only.
+        integration_compile_data_extra: Extra compile_data for integration tests.
+        test_data_extra: Extra runtime data for tests.
+        test_tags: Tags applied to unit + integration test targets.
+            Typicaly used to disable the sandbox, but see https://bazel.build/reference/be/common-definitions#common.tags
```
bolinfest: appease codespell

bolinfest (commented):
Some nits, but this is looking really good!

defs.bzl
zbarsky-openai: true-ish. We can do something like
```
crate_features = select({
    "//configs:release": [],
    "//conditions:default": ["test-support"]
})
```
and build release binaries in the release config. But yes we need a bit of wiring and it's still suboptimal and should be used sparingly. will leave a note

defs.bzl
zbarsky-openai: there's 2 crates that set it:

codex-rs/utils/pty/Cargo.toml:edition = "2021"
codex-rs/windows-sandbox-rs/BUILD.bazel:    crate_edition = "2021",
codex-rs/windows-sandbox-rs/Cargo.toml:edition = "2021"

`windows-sandbox-rs` doesn't even build with 2024, pty does, we could try bumping pty in the non-bazel build if it's safe, but we still need this workaround until windows-sandbox-rs is addressed.

will leave a comment discouraging use

bolinfest: Ah, didn't realize. Then yes, let's just discourage use and leave the `Cargo.toml` files as-is in this PR.

codex-rs/docs/bazel.md
zbarsky-openai: yeah, i'm also gonna file a bazel bug to see if we can have a mode that drops the sandboxing strategy from darwin-sandbox to processwrapper-sandbox so you still get an isolated filesystem layout even without seatbelt

zbarsky-openai: good idea

bolinfest (commented):
This is game-changing!

bolinfest (approved):

zbarsky-openai:
🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢 🚢

PR #8880, opened 2026-01-08
Attempt to reload auth as a step in 401 recovery

pakrym-oai:
When authentication fails, first attempt to reload the auth from file and then attempt to refresh it.


codex-rs/core/src/auth.rs
```diff
@@ -548,6 +545,81 @@ struct CachedAuth {
...
+    Skipped,
+}
+
+pub struct UnauthorizedRecovery {
+    manager: Arc<AuthManager>,
+    step: UnauthorizedRecoveryStep,
+    expected_account_id: Option<String>,
+}
+
+impl UnauthorizedRecovery {
```
etraut-openai, line 573: It would be useful to include a short comment about what this state machine is doing and why it's needed.

codex-rs/core/src/auth.rs
```diff
@@ -548,6 +545,81 @@ struct CachedAuth {
...
+        }
+
+        match self.step {
+            UnauthorizedRecoveryStep::Reload => {
+                match self
+                    .manager
+                    .reload_if_account_id_matches(self.expected_account_id.as_deref())
+                {
+                    ReloadOutcome::Reloaded => {
+                        self.step = UnauthorizedRecoveryStep::RefreshToken;
```
etraut-openai, line 613: Why are we refreshing the token unconditionally in this case? If we reloaded successfully, the reloaded token may have already been refreshed. If we attempt to refresh it again, we could get a failure. Shouldn't we attempt to refresh only if the reload resulted in the same account token than we previously had cached?

codex-rs/core/src/auth.rs
```diff
@@ -548,6 +545,81 @@ struct CachedAuth {
...
+                    }
+                    ReloadOutcome::Skipped => {
+                        self.manager.refresh_token().await?;
```
etraut-openai: How are refresh errors propagated to the caller? I think we want any such errors to be fatal (non-retriable). That's what the current code does, but it looks like the new code is just dropping any errors on the floor.

codex-rs/core/src/auth.rs
```diff
@@ -670,22 +776,29 @@ impl AuthManager {
...
     /// the auth state from disk so other components observe refreshed token.
     /// If the token refresh fails, returns the error to the caller.
-    pub async fn refresh_token(&self) -> Result<Option<String>, RefreshTokenError> {
-        let auth = match self.auth_cached() {
+    pub async fn refresh_token(&self) -> Result<(), RefreshTokenError> {
+        tracing::info!("Refreshing token");
+
+        let cached_auth = self.auth_cached();
+
+        let auth = match self.auth_cached().or(cached_auth) {
```
etraut-openai: This doesn't look right. Isn't this using `self.auth_cached()` in both cases (before and after the "or")? Is this just left over from a bad merge conflict resolution?

codex-rs/core/src/auth.rs
```diff
@@ -548,6 +545,81 @@ struct CachedAuth {
...
+                        self.step = UnauthorizedRecoveryStep::RefreshToken;
+                    }
+                    ReloadOutcome::Skipped => {
+                        self.manager.refresh_token().await?;
+                        self.step = UnauthorizedRecoveryStep::Done;
+                    }
+                }
+            }
+            UnauthorizedRecoveryStep::RefreshToken => {
+                self.manager.refresh_token().await?;
```
etraut-openai, line 622: In this path, should we should avoid writing the refreshed token information back to the credential store? This will happen if the user has subsequently logged out or logged in to a different account. Overwriting that subsequent login info feels wrong. Then again, if we don't overwrite it, other instances that were created with the old token will fail if they later try to refresh their tokens. 

1. Log in with account A to instance 1
2. Log in with account A to instance 2
3. Log in with account B to instance 3
4. Instance 1 gets 401 and refreshes its token. Do we write it back to auth.json? If we do, then instance 2 will see the refreshed token (good) but the next instance launched will use account A rather than B (bad). If we don't, then instance 2 will not see the refreshed token (bad).

Hmm, I don't see a way to handle all of these cases without persisting token state for multiple accounts, which gets even more complex.

codex-rs/core/src/auth.rs
```diff
@@ -633,30 +705,64 @@ impl AuthManager {
...
             &self.codex_home,
             self.enable_codex_api_key_env,
             self.auth_credentials_store_mode,
         )
         .ok()
-        .flatten();
+        .flatten()
+    }
+
+    fn store_auth(&self, new_auth: Option<CodexAuth>) -> bool {
```
etraut-openai: Nit: I find the name of this function to be misleading since it's not really "storing" anything. It's effectively just a setter, so maybe rename to `set_auth`? Or maybe there's a more idiomatic name in rust?

codex-rs/core/src/auth.rs
```diff
@@ -633,30 +705,64 @@ impl AuthManager {
...
+    }
+
+    fn auths_equal(a: &Option<CodexAuth>, b: &Option<CodexAuth>) -> bool {
+        match (a, b) {
+            (None, None) => true,
+            (Some(a), Some(b)) => a == b,
+            _ => false,
+        }
+    }
+    fn load_auth_from_storage(&self) -> Option<CodexAuth> {
```
etraut-openai, line 754: Nit: Add line between functions.

codex-rs/core/src/client.rs
```diff
@@ -509,29 +515,19 @@ where
 /// the mapped `CodexErr` is returned to the caller.
 async fn handle_unauthorized(
     status: StatusCode,
-    refreshed: &mut bool,
-    auth_manager: &Option<Arc<AuthManager>>,
-    auth: &Option<crate::auth::CodexAuth>,
+    auth_recovery: &mut Option<UnauthorizedRecovery>,
```
etraut-openai, line 518: Nice! This function is a lot clearer and simpler now.

etraut-openai (changes requested):

codex-rs/core/src/auth.rs
pakrym-oai: We are not, we are setting the next step to be RefreshToken. The next step will only run if the request fails again. 

We only unconditionally refresh if we failed to reload.

codex-rs/core/src/auth.rs
pakrym-oai: Good spot. Bad merge.

codex-rs/core/src/auth.rs
etraut-openai: Ah, got it. I missed the fact that the next stage of the state machine isn't run unless we hit a 401 again.

codex-rs/core/src/auth.rs
pakrym-oai: Added

etraut-openai (approved):

codex-rs/core/src/auth.rs
pakrym-oai: Chatted offline, decided to proceed with the status quo for now.

PR #8882, opened 2026-01-08
Fix shell timeout test in RBE

zbarsky-openai:
`kill` doesn't exist in the container, but syscalls are all we need

### DIFF
--- a/codex-rs/core/src/shell_snapshot.rs
+++ b/codex-rs/core/src/shell_snapshot.rs
@@ -252,11 +252,11 @@ $envVars | ForEach-Object {
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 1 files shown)


bolinfest:
I think we should update the container to include `kill`: we shouldn't have to add an `unsafe` block just for this feature (and force all contributors to be aware of this quirk).

zbarsky-openai:
> I think we should update the container to include `kill`: we shouldn't have to add an `unsafe` block just for this feature (and force all contributors to be aware of this quirk).

we could probably do it without unsafe with the `nix` crate if that's preferable?  but yeah, we can add `kill` as well, I was just surprised to see this usage because at least in `zsh` kill is a shell built-in, so it's kinda confusing

PR #8902, opened 2026-01-08
feat: add list loaded threads to app server

jif-oai:


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1139,6 +1139,14 @@ pub struct ThreadListResponse {
     pub next_cursor: Option<String>,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ThreadLoadedListResponse {
+    /// Thread ids for sessions currently loaded in memory.
+    pub data: Vec<String>,
```
owenlin0, line 1157: Should this return `Vec<Thread>`? i.e. would the metadata on Thread be useful? 

We don't have to populate all the turns & items since we only do that for specific APIs (we've already documented the behavior on the `Thread` object)

jif-oai: No because sometimes we don't have them and don't need them

owenlin0: Also, would be nice to support pagination for consistency with our other list APIs, even though in practice you'd probably want to return everything always. Mainly because adding pagination later is a breaking change

```
pub struct ThreadLoadedListRequest {
    /// Opaque pagination cursor returned by a previous call.
    pub cursor: Option<String>,
    /// Optional page size; defaults to no limit
    pub limit: Option<u32>,
}

pub struct ThreadLoadedListResponse {
    pub data: Vec<String>,
    /// Opaque cursor to pass to the next call to continue after the last item.
    /// if None, there are no more items to return.
}
```

owenlin0 (approved):

PR #8909, opened 2026-01-08
Add config to disable /feedback

gt-oai:
Some enterprises do not want their users to be able to `/feedback`.

image

Adds to `config.toml`:

```toml
[feedback]
enabled = false
```

I've deliberately decided to:
1. leave other references to `/feedback` (e.g. in the interrupt message, tips of the day) unchanged. I think we should continue to promote the feature even if it is not usable currently.
2. leave the `/feedback` menu item selectable and display an error saying it's disabled, rather than remove the menu item (which I believe would raise more questions).

but happy to discuss these.

This will be followed by a change to requirements.toml that admins can use to force the value of feedback.enabled.

### DIFF
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -356,6 +356,10 @@ pub struct Config {
     /// Defaults to `true`.
     pub analytics: bool,
 
+    /// When `false`, disables feedback collection across Codex product surfaces.
+    /// Defaults to `true`.
+    pub feedback_enabled: bool,
+
     /// OTEL configuration (exporter type, endpoint, headers, etc.).
     pub otel: crate::config::types::OtelConfig,
 }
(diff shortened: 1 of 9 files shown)


jif-oai:
@codex review

codex-rs/core/src/config/mod.rs
```diff
@@ -1403,6 +1411,11 @@ impl Config {
                 .and_then(|a| a.enabled)
                 .or(cfg.analytics.as_ref().and_then(|a| a.enabled))
                 .unwrap_or(true),
+            feedback_enabled: cfg
```
jif-oai, line 1414: I realise this is the pattern we have everywhere but that can create issue with different default implementation
Instead should we have
```
impl FeedbackConfigToml {
  pub(crate) fn enabled(&self) -> bool {
    self.enabled.unwrap_or(true)
  }
}
```
or something like that

jif-oai (approved):
Not tested locally but code lgtm

PR #8910, opened 2026-01-08
feat: add a few metrics

jif-oai:


codex-rs/core/src/codex.rs
```diff
@@ -2111,6 +2124,18 @@ mod handlers {
...
         info!("Shutting down Codex instance");
+        let turn_count = sess
+            .clone_history()
+            .await
+            .get_history()
+            .iter()
+            .filter(|item| is_user_turn_boundary(item))
+            .count();
+        sess.services.otel_manager.counter(
+            "conversation.turn.count",
```
gt-oai, line 2135: Does `is_user_turn_boundary` effectively count user turns? Would `conversation.user_turn.count` be more explicit in that case?

If "turn" is accepted convention for user turn then ignore me.

gt-oai (approved):

codex-rs/core/src/codex.rs
jif-oai: Yes this is :)

PR #8913, opened 2026-01-08
Add `read-only` when backfilling requirements from managed_config

gt-oai:
When a user has a managed_config which doesn't specify read-only, Codex fails to launch.

### DIFF
--- a/codex-rs/core/src/config_loader/mod.rs
+++ b/codex-rs/core/src/config_loader/mod.rs
@@ -559,7 +559,14 @@ impl From<LegacyManagedConfigToml> for ConfigRequirementsToml {
             config_requirements_toml.allowed_approval_policies = Some(vec![approval_policy]);
         }
         if let Some(sandbox_mode) = sandbox_mode {
-            config_requirements_toml.allowed_sandbox_modes = Some(vec![sandbox_mode.into()]);
+            let required_mode: SandboxModeRequirement = sandbox_mode.into();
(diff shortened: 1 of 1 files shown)


codex-rs/core/src/config_loader/mod.rs
```diff
@@ -574,7 +574,14 @@ impl From<LegacyManagedConfigToml> for ConfigRequirementsToml {
...
         }
         if let Some(sandbox_mode) = sandbox_mode {
-            config_requirements_toml.allowed_sandbox_modes = Some(vec![sandbox_mode.into()]);
+            let required_mode: SandboxModeRequirement = sandbox_mode.into();
+            // Allowing read-only is a requirement for Codex to function correctly.
+            // So in this backfill path, we append read-only if it's not already specified.
+            let mut allowed_modes = vec![SandboxModeRequirement::ReadOnly];
+            if required_mode != SandboxModeRequirement::ReadOnly {
+                allowed_modes.push(required_mode);
+            }
```
bolinfest, line 583: I think this is more straightforward:

```suggestion
            let allowed_modes = if required_mode == SandboxModeRequirement::ReadOnly {
                vec![SandboxModeRequirement::ReadOnly]
            } else {
                // or whichever order makes sense?
                vec![required_mode, SandboxModeRequirement::ReadOnly]
            };
```

bolinfest (approved):

PR #8918, opened 2026-01-08
feat: add prompt stashing

llx-openai:
Add the QOL feature to "stash" a prompt locally, which is auto-popped on next submit.
Stash is triggered by the hotkey `ctrl + s`. The stack has depth 1 and overrides on multiple stashes.

image

image

## Test Plan

Added unit tests. Tested locally for tui and tui2.
I was unable to fully e2e test image stash pops due to the way openai configures terminal permissions for local dev, but the image pointer is successfully popped.

### DIFF
--- a/codex-rs/tui2/src/bottom_pane/chat_composer.rs
+++ b/codex-rs/tui2/src/bottom_pane/chat_composer.rs
@@ -57,6 +57,7 @@ use crate::bottom_pane::textarea::TextAreaState;
 use crate::clipboard_paste::normalize_pasted_path;
 use crate::clipboard_paste::pasted_image_format;
 use crate::history_cell;
+use crate::text_formatting::truncate_text;
 use crate::ui_consts::LIVE_PREFIX_COLS;
 use codex_core::skills::model::SkillMetadata;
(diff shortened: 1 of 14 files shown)


codex-rs/tui2/src/bottom_pane/chat_composer.rs
```diff
@@ -318,6 +334,69 @@ impl ChatComposer {
         self.sync_popups();
     }
 
+    /// Replace the composer content with text, rebuilding attachment elements.
+    pub(crate) fn apply_external_edit(&mut self, text: String) {
```
joshka-oai, line 338: This method's name is at a high level of abstraction, while the implementation drops to low level. There's nothing in between that binds the overall intent to the actual implementation detail. Add an overview of what this method does to help validate the intention of the logic inside this method aligns with what it actually does, as well as probably some inline comments helping make that clear. Codex is pretty good at rewriting this sort of thing to fit that criteria.

joshka-oai (commented):
Code looks good.
I think we should probably do this automatically on clear / delete line rather than explicitly like this though. Chatting internally on it.

PR #8919, opened 2026-01-08
fix: invalidate history log id after trims

tibo-openai:
This updates history log id derivation to include the first entry so trimmed history invalidates stale offsets and adds a regression test to avoid returning incorrect entries after pruning.

**More context**
If history is trimmed after hitting max_bytes, the file keeps the same inode, so a cached log_id+offset could return a different entry than the one the user previously saw; now the log_id changes when the first line changes, so the stale request is rejected and the (T)UI refetches instead of showing the wrong message.

### DIFF
--- a/codex-rs/core/src/message_history.rs
+++ b/codex-rs/core/src/message_history.rs
@@ -16,8 +16,11 @@
 //! Note: `conversation_id` stores the thread id; the field name is preserved for
 //! backwards compatibility with existing history files.
 
+use std::collections::hash_map::DefaultHasher;
 use std::fs::File;
 use std::fs::OpenOptions;
+use std::hash::Hash;
+use std::hash::Hasher;
 use std::io::BufRead;
 use std::io::BufReader;
 use std::io::Read;
@@ -33,7 +36,9 @@ use serde::Serialize;
 
 use std::time::Duration;
 use tokio::fs;
+use tokio::io::AsyncBufReadExt;
 use tokio::io::AsyncReadExt;
+use tokio::io::BufReader as AsyncBufReader;
 
 use crate::config::Config;
 use crate::config::types::HistoryPersistence;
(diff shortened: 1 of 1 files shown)


codex-rs/core/src/message_history.rs
```diff
@@ -295,14 +317,29 @@ async fn history_metadata_for_file(path: &Path) -> (u64, usize) {
...
         Ok(f) => f,
-        Err(_) => return (log_id, 0),
+        Err(_) => return (base_id, 0),
     };
 
-    // Count newline bytes.
-    let mut buf = [0u8; 8192];
+    let mut reader = AsyncBufReader::new(&mut file);
+    let mut first_line = String::new();
+    let first_line_bytes = match reader.read_line(&mut first_line).await {
```
pakrym-oai, line 325: do you need to read the actual first line ? Can you grab the first N bytes instead?

pakrym-oai: I see, lookup_history_entry is reading in lines.

pakrym-oai:
This all feels a bit complicated. Should we use creation time as the log_it for all platforms and reset it when truncating?

pakrym-oai (changes requested):
Should we use a simpler approach of always using creation time as log id and resetting it on truncation?

codex-rs/core/src/message_history.rs
```diff
@@ -243,6 +248,23 @@ fn trim_target_bytes(max_bytes: u64, newest_entry_len: u64) -> u64 {
...
+    if base_id == 0 {
+        return 0;
+    }
+    let Some(line) = first_line else {
+        return base_id;
+    };
+
+    let mut hasher = DefaultHasher::new();
+    line.hash(&mut hasher);
+    base_id ^ hasher.finish()
```
ddr-oai, line 265: **<sub><sub>P2 Badge</sub></sub>  Ensure log_id changes when trim removes duplicate first line**

Because `history_log_id_with_first_line` only combines the inode with a hash of the first line, trims that drop the first entry but leave an identical first line (e.g., repeated prompts at the top of history) will produce the same `log_id`. In that scenario `lookup_history_entry` will accept stale offsets and can return a different entry than the user previously saw even though the history was trimmed. Consider incorporating additional changing data (such as a hash of the first N bytes, first-line length, or line count) so duplicates can’t keep the log_id stable after trims.

Useful? React with 👍 / 👎.

ddr-oai (commented):
### 💡 Codex Review

Here are some automated review suggestions for this pull request.

**Reviewed commit:** `c52dc9fd93`
    

<details> <summary>ℹ️ About Codex in GitHub</summary>
<br/>

[Your team has set up Codex to review pull requests in this repo](http://chatgpt.com/codex/settings/general). Reviews are triggered when you
- Open a pull request for review
- Mark a draft as ready
- Comment "@codex review".

If Codex has suggestions, it will comment; otherwise it will react with 👍.




Codex can also answer questions or update the PR. Try commenting "@codex address that feedback".
            
</details>

ddr-oai:
(i picked a random codex pr to try this CR trigger logic on, congrats on extra review coverage)

PR #8928, opened 2026-01-08
Allow disabling keyring loading in debug

pakrym-oai:
Today the MacOS opens a password dialog for every MCP server in config that we attempt to load OAuth keys for. Makes debugging codex pretty hard.

### DIFF
--- a/codex-rs/rmcp-client/src/oauth.rs
+++ b/codex-rs/rmcp-client/src/oauth.rs
@@ -50,6 +50,7 @@ use tokio::sync::Mutex;
 use crate::find_codex_home::find_codex_home;
 
 const KEYRING_SERVICE: &str = "Codex MCP Credentials";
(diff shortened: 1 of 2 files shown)


codex-rs/rmcp-client/src/oauth.rs
```diff
@@ -50,6 +50,7 @@ use tokio::sync::Mutex;
 use crate::find_codex_home::find_codex_home;
 
 const KEYRING_SERVICE: &str = "Codex MCP Credentials";
+const MCP_OAUTH_KEYRING_DEBUG_DISABLE_ENV_VAR: &str = "CODEX_MCP_OAUTH_KEYRING_DEBUG_DISABLE";
```
gpeal, line 53: How about extending this enum?
https://github.com/openai/codex/blob/main/codex-rs/rmcp-client/src/oauth.rs#L68

PR #8930, opened 2026-01-08
feat: first pass on clb tool

jif-oai:


codex-rs/core/src/tools/handlers/collab.rs
```diff
@@ -0,0 +1,195 @@
...
+    config.developer_instructions = turn.developer_instructions.clone();
+    config.base_instructions = turn.base_instructions.clone();
+    config.compact_prompt = turn.compact_prompt.clone();
+    config.user_instructions = turn.user_instructions.clone();
+    config.shell_environment_policy = turn.shell_environment_policy.clone();
+    config.codex_linux_sandbox_exe = turn.codex_linux_sandbox_exe.clone();
+    config.cwd = turn.cwd.clone();
+    config
+        .approval_policy
+        .set(turn.approval_policy)
```
gt-oai, line 183: are `Constrained<T>` values copyable here? e.g. `config.approval_policy = turn.approval_policy...` ? I think this would avoid the potential error and mean this function doesn't have to return a result?

codex-rs/core/src/tools/spec.rs
```diff
@@ -416,6 +421,104 @@ fn create_view_image_tool() -> ToolSpec {
...
+        parameters: JsonSchema::Object {
+            properties,
+            required: Some(vec!["message".to_string()]),
+            additional_properties: Some(false.into()),
+        },
+    })
+}
+
+fn create_send_input_tool() -> ToolSpec {
+    let mut properties = BTreeMap::new();
```
gt-oai, line 446: nit: Could you use BTreeMap::from and inline this? Here and below

gt-oai (approved):

codex-rs/core/src/tools/spec.rs
jif-oai: I try to stay consistent with the rest of the code here where we sometimes have dynamic injection. I'll make a follow-up that clean everything at once

jif-oai:
@codex review

PR #8939, opened 2026-01-08
[chore] move app server tests from chat completion to responses

celia-oai:
We are deprecating chat completions. Move all app server tests from chat completion to responses.


codex-rs/app-server/tests/common/mock_model_server.rs
```diff
@@ -22,7 +21,7 @@ pub async fn create_mock_chat_completions_server(responses: Vec<String>) -> Mock
     };
 
     Mock::given(method("POST"))
-        .and(path("/v1/chat/completions"))
+        .and(path_regex(".*/responses$"))
```
owenlin0, line 24: curious, why regex instead of just `path("/v1/responses")?`

pakrym-oai (approved):

codex-rs/app-server/tests/common/mock_model_server.rs
```diff
@@ -31,18 +30,18 @@ pub async fn create_mock_chat_completions_server(responses: Vec<String>) -> Mock
...
+    let server = responses::start_mock_server().await;
 
     let seq_responder = SeqResponder {
         num_calls: AtomicUsize::new(0),
         responses,
     };
 
     Mock::given(method("POST"))
-        .and(path("/v1/chat/completions"))
+        .and(path_regex(".*/responses$"))
```
owenlin0, line 44: same here

codex-rs/app-server/tests/suite/interrupt.rs
```diff
@@ -153,7 +153,7 @@ model_provider = "mock_provider"
 [model_providers.mock_provider]
 name = "Mock provider for test"
 base_url = "{server_uri}/v1"
-wire_api = "chat"
+wire_api = "responses"
```
owenlin0, line 156: wonder if we can just remove all of these lines from all of these test configs now:
```
model_provider = "mock_provider"

[model_providers.mock_provider]
name = "Mock provider for test"
base_url = "{server_uri}/v1"
wire_api = "responses"
request_max_retries = 0
stream_max_retries = 0
```

owenlin0 (commented):
thanks for doing this! 🙌

PR #8940, opened 2026-01-08
Refactor remote models tests to use TestCodex builder

aibrahim-oai:
- add `with_model_provider` to the test codex builder
- replace the bespoke remote models harness with `TestCodex` in `remote_models` tests

### DIFF
--- a/codex-rs/core/tests/suite/remote_models.rs
+++ b/codex-rs/core/tests/suite/remote_models.rs
@@ -97,10 +97,6 @@ async fn remote_models_remote_model_uses_unified_exec() -> Result<()> {
 
     let mut builder = test_codex()
         .with_auth(CodexAuth::create_dummy_chatgpt_auth_for_testing())
-        .with_model_provider(ModelProviderInfo {
-            base_url: Some(format!("{}/v1", server.uri())),
-            ..built_in_model_providers()["openai"].clone()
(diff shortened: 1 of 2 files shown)


codex-rs/core/tests/suite/remote_models.rs
```diff
@@ -332,19 +350,23 @@ async fn remote_models_apply_remote_base_instructions() -> Result<()> {
...
-        config.features.enable(Feature::RemoteModels);
-        config.model = Some("gpt-5.1".to_string());
-    })
-    .await?;
-
-    let RemoteModelsHarness {
+    let mut builder = test_codex()
+        .with_auth(CodexAuth::create_dummy_chatgpt_auth_for_testing())
+        .with_model_provider(ModelProviderInfo {
+            base_url: Some(format!("{}/v1", server.uri())),
```
pakrym-oai: Why doe we need this? Don't we set up the same by default?

pakrym-oai (changes requested):
QQ

codex-rs/core/tests/suite/remote_models.rs
aibrahim-oai: fixed

pakrym-oai (approved):

Issue #8523, opened 2025-12-24
Codex Isn't Reading Images


am-will:
### What version of Codex is running?

0.77

### What subscription do you have?

Plus

### Which model were you using?

5.2 Codex

### What platform is your computer?

Linux

### What issue are you seeing?

I asked Codex to scrape some text and add it to the rules, but instead, its exploring the codebase for the image name (but its in my clipboard instead). 

<img width="1378" height="579" alt="Image" src="https://github.com/user-attachments/assets/487f623c-ab54-4cdc-b520-174a942e6ebf" />

Unsure if this is expected behavior but I  certainly think we should be able to share images from clipboard

### What steps can reproduce the bug?

Uploaded thread: 019b5157-67ea-7453-ba3b-05132b349165

### What is the expected behavior?

Should read the image.

InderdeepBajwa:
+1 on Mac Terminal.

Steps:
1. Capture a screenshot into clipboard
2. Paste into Codex CLI
3. It can't read it.

<img width="1170" height="52" alt="Image" src="https://github.com/user-attachments/assets/8b6fb072-0a9e-40d4-914c-be937c1971e5" />

maxjendrall:
Yes, same issue here! 

019b65db-3054-75b3-8e03-23a115a417ae

@ae-openai @aibrahim-oai

am-will:
@InderdeepBajwa @toorusr I have a temporary fix that works well for me. Keep in mind, i'm on Linux and unsure how this would work on other platforms.

Obviously this will depend where your clipboard images are stored, and if they're accessible by the agent. In a container, they won't be. 

Put this in the .codex/AGENTS.md:

```
## Clipboard Screenshots
- Pasted images named `codex-clipboard-*.png` are saved in root `/tmp` (e.g., `[codex-clipboard-mWiFoN.png 443x658]`).
```

weiying-chen:
Same here. I have to tell Codex to search in `/tmp`.

![Image](https://github.com/user-attachments/assets/402ac8de-5af0-4346-8bab-f9a4d878c8cc)

Ubuntu 25.10. Codex 0.77. gpt-codex 5.2.

PhilChen-6765:
+1, macos 26.2, Codex CLI 0.77

codex can only see the clipboard string, but cannot find the image

am-will:
> Same here. I have to tell Codex to search in `/tmp`.
> 
[...]

PR #8950, opened 2026-01-09
Label attached images so agent can understand in-message labels

charley-oai:
Agent wouldn't "see" attached images and would instead try to use the view_file tool:
image

In this PR, we wrap image content items in XML tags with the name of each image (now just a numbered name like `[Image #1]`), so that the model can understand inline image references (based on name). We also put the image content items above the user message which the model seems to prefer (maybe it's more used to definitions being before references).

We also tweak the view_file tool description which seemed to help a bit

Results on a simple eval set of images:

Before
image

After
image

```json
[
  {
    "id": "single_describe",
    "prompt": "Describe the attached image in one sentence.",
    "images": ["image_a.png"]
  },
  {
    "id": "single_color",
    "prompt": "What is the dominant color in the image? Answer with a single color word.",
    "images": ["image_b.png"]
  },
  {
    "id": "orientation_check",
    "prompt": "Is the image portrait or landscape? Answer in one sentence.",
    "images": ["image_c.png"]
  },
  {
    "id": "detail_request",
    "prompt": "Look closely at the image and call out any small details you notice.",
    "images": ["image_d.png"]
  },
  {
    "id": "two_images_compare",
    "prompt": "I attached two images. Are they the same or different? Briefly explain.",
    "images": ["image_a.png", "image_b.png"]
  },
  {
    "id": "two_images_captions",
    "prompt": "Provide a short caption for each image (Image 1, Image 2).",
    "images": ["image_c.png", "image_d.png"]
  },
  {
    "id": "multi_image_rank",
    "prompt": "Rank the attached images from most colorful to least colorful.",
    "images": ["image_a.png", "image_b.png", "image_c.png"]
  },
  {
    "id": "multi_image_choice",
    "prompt": "Which image looks more vibrant? Answer with 'Image 1' or 'Image 2'.",
    "images": ["image_b.png", "image_d.png"]
  }
]
```

Fixes issue https://github.com/openai/codex/issues/8523

### DIFF
--- a/codex-rs/protocol/src/models.rs
+++ b/codex-rs/protocol/src/models.rs
@@ -190,7 +190,7 @@ pub fn local_image_label_text(label_number: usize) -> String {
     format!("[Image #{label_number}]")
 }
 
-fn local_image_open_tag_text(label_number: usize) -> String {
+pub fn local_image_open_tag_text(label_number: usize) -> String {
     let label = local_image_label_text(label_number);
     format!("{LOCAL_IMAGE_OPEN_TAG_PREFIX}{label}{LOCAL_IMAGE_OPEN_TAG_SUFFIX}")
 }
--- a/codex-rs/protocol/src/models.rs
+++ b/codex-rs/protocol/src/models.rs
@@ -182,23 +182,17 @@ fn local_image_error_placeholder(
 
 pub const VIEW_IMAGE_TOOL_NAME: &str = "view_image";
 
-const LOCAL_IMAGE_OPEN_TAG_PREFIX: &str = "<image name=\"";
-const LOCAL_IMAGE_OPEN_TAG_SUFFIX: &str = "\">";
+const LOCAL_IMAGE_OPEN_TAG_PREFIX: &str = "<image name=";
+const LOCAL_IMAGE_OPEN_TAG_SUFFIX: &str = ">";
 const LOCAL_IMAGE_CLOSE_TAG: &str = "</image>";
 
-fn local_image_label_text_with_number(
-    path: &std::path::Path,
-    label_number: Option<usize>,
-) -> String {
-    let label = match label_number {
-        Some(label_number) => format!("[Image #{label_number}]"),
-        None => format!("[Image {}]", path.display()),
-    };
-    format!("{LOCAL_IMAGE_OPEN_TAG_PREFIX}{label}{LOCAL_IMAGE_OPEN_TAG_SUFFIX}")
+pub fn local_image_label_text(label_number: usize) -> String {
+    format!("[Image #{label_number}]")
 }
 
-pub fn local_image_label_text(path: &std::path::Path) -> String {
-    local_image_label_text_with_number(path, None)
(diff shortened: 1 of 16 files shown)


codex-rs/protocol/src/models.rs
```diff
@@ -203,6 +225,43 @@ fn unsupported_image_error_placeholder(path: &std::path::Path, mime: &str) -> Co
     }
 }
 
+pub fn local_image_content_items(path: &std::path::Path, include_label: bool) -> Vec<ContentItem> {
```
aibrahim-oai: let's put it as a static method on `ContentItem` and rename to sth like `image_from_path`?

codex-rs/protocol/src/models.rs
```diff
@@ -180,6 +180,28 @@ fn local_image_error_placeholder(
     }
 }
 
+pub const VIEW_IMAGE_TOOL_NAME: &str = "view_image";
+
+fn local_image_label_suffix() -> String {
+    format!(" follows (you can see it without using the {VIEW_IMAGE_TOOL_NAME} tool):")
+}
+
+pub fn local_image_label_text(path: &std::path::Path) -> String {
```
aibrahim-oai: Does this happen before or after encoding? not sure why are we attaching the path

charley-oai: Code has changed, we no longer include the path in the label

codex-rs/protocol/src/models.rs
charley-oai: I think this is following the pattern of some similar functions in the file already, like `local_image_error_placeholder` which returns `ContentItem`? Probably we should change all of those functions to use a different pattern, or none of them?

codex-rs/protocol/src/models.rs
```diff
@@ -203,6 +227,58 @@ fn unsupported_image_error_placeholder(path: &std::path::Path, mime: &str) -> Co
     }
 }
 
+pub fn local_image_content_items(path: &std::path::Path, include_label: bool) -> Vec<ContentItem> {
```
aibrahim-oai: let's just use `fn local_image_content_items_with_label_number(`

codex-rs/tui/src/bottom_pane/chat_composer.rs
```diff
@@ -273,17 +266,15 @@ impl ChatComposer {
...
-                tracing::info!("OK: {pasted}");
-                let format_label = pasted_image_format(&path_buf).label();
-                self.attach_image(path_buf, w, h, format_label);
-                true
-            }
-            Err(err) => {
-                tracing::trace!("ERR: {err}");
-                false
-            }
+        if image::image_dimensions(&path_buf).is_ok() {
```
aibrahim-oai: nit: let's keep the match statement

aibrahim-oai (approved):
Awesome. Code feels cleaner. Let's add integration tests to make sure we send the right request shape to model. We persist the right thing to the rollout file. Maybe a UI snapshot as well.

PR #8956, opened 2026-01-09
add generated jsonschema for config.toml

sayan-oai:
### What
Add JSON Schema generation for `config.toml`, with checked‑in `docs/config.schema.json`. We can move the schema elsewhere if preferred (and host it if there's demand).

Add fixture test to prevent drift and `just write-config-schema` to regenerate on schema changes.

Generate MCP config schema from `RawMcpServerConfig` instead of `McpServerConfig` because that is the runtime type used for deserialization.

Populate feature flag values into generated schema so they can be autocompleted.

### Tests
Added tests + regenerate script to prevent drift. Tested autocompletions using generated jsonschema locally with Even Better TOML.

https://github.com/user-attachments/assets/5aa7cd39-520c-4a63-96fb-63798183d0bc


sayan-oai:
@codex review this

codex-rs/core/src/config/mod.rs
```diff
@@ -879,7 +890,8 @@ impl ProjectConfig {
     }
 }
 
-#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq)]
+#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq, JsonSchema)]
+#[schemars(deny_unknown_fields)]
 pub struct ToolsToml {
     #[serde(default, alias = "web_search_request")]
     pub web_search: Option<bool>,
```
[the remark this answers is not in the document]
jif-oai: nope

jif-oai (approved):
That's cool. We should publish it to some standard directories (and have a gh job for this)
@joshka-oai had some thoughts on this I think

joshka-oai:
Yep. Publish a small pointer file to Json-schema.org, then publish this file to our website so we can easily update it in each release. We also need to add a small comment in our config files to point at the schema location (as we use config.toml rather than a more specific name like codex.toml which can be auto detected).

codex-rs/core/src/config/SCHEMA.md
```diff
@@ -0,0 +1,12 @@
...
+
+We generate a JSON Schema for `~/.codex/config.toml` from the `ConfigToml` type
+and commit it at `docs/config.schema.json` for editor integration.
+
+When you change any fields included in `ConfigToml` (or nested config types),
+regenerate the schema and update the fixture:
+
+```
+just write-config-schema
+cargo test -p codex-core config_schema_matches_fixture
```
joshka-oai: Add to ci as a check and agents.md

codex-rs/core/src/config/mod.rs
```diff
@@ -669,7 +671,11 @@ pub fn set_default_oss_provider(codex_home: &Path, provider: &str) -> std::io::R
 }
 
 /// Base config deserialized from ~/.codex/config.toml.
-#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq)]
+// The JSON Schema in `docs/config.schema.json` is generated from this struct.
```
joshka-oai: Ci check and agents.md update should be more than enough

bolinfest:
In terms of publishing artifacts, that should be done as part of `rust-release.yml` or our internal Python script that stamps the tag for a release that triggers `rust-release.yml`.

codex-rs/core/src/config/schema.rs
```diff
@@ -0,0 +1,105 @@
...
+            serde_json::from_str(&fixture).expect("parse config schema fixture");
+        assert_eq!(
+            fixture_value, schema_value,
+            "Current schema for `config.toml` doesn't match the fixture. Run `just write-config-schema` to overwrite with your changes."
+        );
+    }
+
+    /// Overwrite the config schema fixture with the current schema.
+    #[test]
+    #[ignore]
```
bolinfest: I don't think we should be relying on an ignored test for this.

Why not make a binary target for this crate that does the work to generate the JSON?

Then we should have a (not ignored!) test that verifies that the vendored `config.schema.json` and the output are in sync? When the test fails, it should tell the user to `cargo run --bin THE_REGEN_BIN_TARGET` or whatever.

codex-rs/core/src/config/schema.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn write_config_schema_fixture() {
+        let fixture_path = codex_utils_cargo_bin::find_resource!("../../docs/config.schema.json")
```
bolinfest: As written, this may break the Bazel build once the test is no longer ignored.

The easiest thing would be to have this file live in this crate.

codex-rs/core/src/config/schema.rs
```diff
@@ -0,0 +1,105 @@
...
+        let fixture = std::fs::read_to_string(fixture_path).expect("read config schema fixture");
+        let fixture_value: serde_json::Value =
+            serde_json::from_str(&fixture).expect("parse config schema fixture");
+        assert_eq!(
+            fixture_value, schema_value,
+            "Current schema for `config.toml` doesn't match the fixture. Run `just write-config-schema` to overwrite with your changes."
+        );
+    }
+
+    /// Overwrite the config schema fixture with the current schema.
```
bolinfest: We also have some ad-hoc tests in https://github.com/openai/codex/blob/main/.github/workflows/ci.yml. Alternatively, you could make a small shell/Python script that does the comparison and run it as part of that workflow.

Then you don't have to worry about making things work with Bazel, though devs will probably find that test failure later than they would if the test were run by Cargo/Bazel.

codex-rs/core/src/config/schema.rs
```diff
@@ -0,0 +1,105 @@
+#[cfg(test)]
```
bolinfest: The many references to `#[cfg(test)]` feels wrong: can you move all this to a separate file? Or to `mod tests`?

codex-rs/core/BUILD.bazel
```diff
@@ -28,7 +28,10 @@ codex_rust_crate(
     #
     # TODO(aibrahim): Update the tests so that `just bazel-remote-test` succeeds
     # without this workaround.
```
bolinfest: Please move this comment so it is clear we are only talking about `//:AGENTS.md`.

.github/workflows/rust-ci.yml
```diff
@@ -66,6 +66,10 @@ jobs:
         run: cargo fmt -- --config imports_granularity=Item --check
       - name: Verify codegen for mcp-types
         run: ./mcp-types/check_lib_rs.py
+      - name: Verify config schema fixture
+        run: |
+          cargo run -p codex-core --bin codex-write-config-schema
+          git diff --exit-code core/config.schema.json
```
bolinfest: In general, we should not be modifying the files in the repo as part of CI.

It looks like you already have an `--out` flag, so I would use `--out /tmp/config.schema.json` to write to a temp file and then just `diff core/config.schema.json /tmp/config.schema.json`.

Also maybe add `if: runner.os == 'ubuntu'` to this step so we're confident `diff` is available?

bolinfest: I would also design this so that if it fails, it prints out an error like:

```
codex-rs/core/config.schema.json is out of sync: run `just write-config-schema` to regenerate it
```

Though again, if you did this as a unit test, then developers would be more likely to discover this _before_ putting up the PR.

codex-rs/core/src/bin/config_schema.rs
```diff
@@ -0,0 +1,25 @@
...
+use std::env;
+use std::path::PathBuf;
+
+/// Generate the JSON Schema for `config.toml` and write it to `config.schema.json`.
+fn main() -> Result<()> {
+    let mut args = env::args().skip(1);
+    let mut out_path = None;
+
+    while let Some(arg) = args.next() {
+        match arg.as_str() {
```
bolinfest: I would just use `clap` rather than parse your own args.

codex-rs/core/src/config/schema.md
```diff
@@ -0,0 +1,11 @@
+# Config JSON Schema
```
bolinfest, line 1: It's a bit of a pain to fix this because macOS is case-preserving, but I don't think this filename needs to be in all caps.

You probably have to do something like:

```
git mv SCHEMA.md something.md
git mv something.md schema.md
```

codex-rs/core/src/config/schema.rs
```diff
@@ -0,0 +1,97 @@
...
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::config_schema_json;
+    use pretty_assertions::assert_eq;
+
+    #[test]
+    fn config_schema_matches_fixture() {
```
bolinfest, line 101: Oh, just seeing this: so do we also need the thing in `rust-ci.yml`?

.github/workflows/rust-ci.yml
sayan-oai: Yes, my preference is just a unit test as well, which is what `config_schema_matches_fixture()` in `schema.rs` is. I think there was a misunderstanding, but I think we can just remove this extra CI step as the logically equivalent unit test will run in rust CI anyways.

codex-rs/core/src/config/schema.rs
sayan-oai: Nope, I think I misunderstood what we were discussing re: where to test this and added it both places. Will remove from here, my initial approach was to just have the unit test anyways (since it runs in CI).

codex-rs/core/BUILD.bazel
```diff
@@ -26,9 +26,12 @@ codex_rust_crate(
     # the working directory is different and so the file is not found unless it
     # is explicitly added as test data.
     #
```
bolinfest: This should be part of the comment that you moved: the `AGENTS.md` file is what is being used as the marker.

codex-rs/core/Cargo.toml
```diff
@@ -58,6 +62,8 @@ reqwest = { workspace = true, features = ["json", "stream"] }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 serde_yaml = { workspace = true }
+clap = { workspace = true, features = ["derive"] }
```
bolinfest: would be nice to alpha sort (sorry, we still haven't found the right linter for this...)

bolinfest (approved):
Just some small things, but this LGTM!

codex-rs/core/BUILD.bazel
sayan-oai: whoops, good catch. ty

PR #8957, opened 2026-01-09
feat: hot reload mcp servers

shijie-oai:
### Summary
* Added `mcpServer/refresh` command to inform app servers and active threads to refresh mcpServer on next turn event.
* Added `pending_mcp_server_refresh_config` to codex core so that if the value is populated, we reinitialize the mcp server manager on the thread level. 
	*  The config is updated on `mcpServer/refresh` command which we iterate through threads and provide with the latest config value after last write.

### DIFF
--- a/codex-rs/app-server/README.md
+++ b/codex-rs/app-server/README.md
@@ -88,7 +88,7 @@ Example (from OpenAI's official VSCode extension):
 - `model/list` — list available models (with reasoning effort options).
 - `skills/list` — list skills for one or more `cwd` values (optional `forceReload`).
 - `mcpServer/oauth/login` — start an OAuth login for a configured MCP server; returns an `authorization_url` and later emits `mcpServer/oauthLogin/completed` once the browser flow finishes.
-- `mcpServer/refresh` — reinitialize MCP servers for loaded threads and refresh cached tool lists; returns `{}`.
+- `config/mcpServer/reload` — reload MCP server config from disk and queue a refresh for loaded threads (applied on each thread's next active turn); returns `{}`. Use this after editing `config.toml` without restarting the server.
 - `mcpServerStatus/list` — enumerate configured MCP servers with their tools, resources, resource templates, and auth status; supports cursor+limit pagination.
 - `feedback/upload` — submit a feedback report (classification + optional reason/logs and conversation_id); returns the tracking thread id.
 - `command/exec` — run a single command under the server sandbox without starting a thread/turn (handy for utilities and validation).
--- a/codex-rs/app-server/README.md
+++ b/codex-rs/app-server/README.md
@@ -88,6 +88,7 @@ Example (from OpenAI's official VSCode extension):
 - `model/list` — list available models (with reasoning effort options).
 - `skills/list` — list skills for one or more `cwd` values (optional `forceReload`).
 - `mcpServer/oauth/login` — start an OAuth login for a configured MCP server; returns an `authorization_url` and later emits `mcpServer/oauthLogin/completed` once the browser flow finishes.
+- `mcpServer/refresh` — reinitialize MCP servers for loaded threads and refresh cached tool lists; returns `{}`.
 - `mcpServerStatus/list` — enumerate configured MCP servers with their tools, resources, resource templates, and auth status; supports cursor+limit pagination.
 - `feedback/upload` — submit a feedback report (classification + optional reason/logs and conversation_id); returns the tracking thread id.
 - `command/exec` — run a single command under the server sandbox without starting a thread/turn (handy for utilities and validation).
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -60,7 +60,6 @@ use codex_app_server_protocol::LogoutChatGptResponse;
 use codex_app_server_protocol::McpServerOauthLoginCompletedNotification;
 use codex_app_server_protocol::McpServerOauthLoginParams;
 use codex_app_server_protocol::McpServerOauthLoginResponse;
-use codex_app_server_protocol::McpServerRefreshParams;
 use codex_app_server_protocol::McpServerRefreshResponse;
 use codex_app_server_protocol::McpServerStatus;
 use codex_app_server_protocol::ModelListParams;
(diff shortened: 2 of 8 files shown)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2302,6 +2307,57 @@ impl CodexMessageProcessor {
         outgoing.send_response(request_id, response).await;
     }
 
+    async fn mcp_server_refresh(&self, request_id: RequestId, _params: Option<()>) {
```
shijie-oai, line 2310: This does not refresh the mcp connection manager for all active threads for inform them there is an update with the latest mcp config. It is up to the thread to refresh the connection on next active turn. This should help with performance and avoid excessive refresh if there is a lot of MCP changes at once.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2302,6 +2307,57 @@ impl CodexMessageProcessor {
...
+            };
+
+        let refresh_config = McpServerRefreshConfig {
+            mcp_servers,
+            mcp_oauth_credentials_store_mode,
+        };
+
+        let thread_manager = Arc::clone(&self.thread_manager);
+        tokio::spawn(async move {
+            thread_manager.refresh_mcp_servers(refresh_config).await;
```
shijie-oai: We build the mcp configs and pass it along to the thread so that they can have the latest slice on next refresh.

codex-rs/core/src/codex.rs
```diff
@@ -1631,12 +1638,85 @@ impl Session {
...
+        refreshed_manager
+            .initialize(
+                mcp_servers,
+                store_mode,
+                auth_statuses,
+                self.get_tx_event(),
+                cancel_token,
+                sandbox_state,
+            )
+            .await;
```
shijie-oai, line 1706: We are doing a full reset here so that the mcp manager has the latest config value.

codex-rs/core/src/codex.rs
```diff
@@ -706,6 +710,7 @@ impl Session {
             agent_status: Arc::clone(&agent_status),
             state: Mutex::new(state),
             features: config.features.clone(),
+            pending_mcp_server_refresh_config: Mutex::new(None),
```
shijie-oai, line 713: Track if there is any pending config, if there is `refresh_mcp_servers_if_requested` would use it to reinitialize the mcp server.

codex-rs/core/src/thread_manager.rs
```diff
@@ -136,6 +138,27 @@ impl ThreadManager {
         self.state.threads.read().await.keys().copied().collect()
     }
 
+    pub async fn refresh_mcp_servers(&self, refresh_config: McpServerRefreshConfig) {
```
shijie-oai, line 141: Iterate through all active threads and submit `Op::RefreshMcpServers` with the latest config so that thread can pick up and reinit mcp connection managers on next turn.

codex-rs/app-server/src/codex_message_processor.rs
gpeal: Maybe put this in a comment?

gpeal: "It is up to the thread to refresh the connection on next active turn."

What does this mean?

shijie-oai: All active threads in the thread manager would be aware of a new mcp config but we are not triggering an actual reinitialization yet. On a future turn (i.e. user submit a new prompt), if the thread detects a pending_mcp_config, then it would first reinitialize and then move onto the actual turn. The benefit here is that we are not doing any extra leg work - for example you started a thread and then we change the mcp servers but never reengage with that thread, then that thread would not refresh mcp server until called upon.

gpeal (approved):

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2302,6 +2307,57 @@ impl CodexMessageProcessor {
...
+    async fn mcp_server_refresh(&self, request_id: RequestId, _params: Option<()>) {
+        let config = match self.load_latest_config().await {
+            Ok(config) => config,
+            Err(error) => {
+                self.outgoing.send_error(request_id, error).await;
+                return;
+            }
+        };
+
+        let mcp_servers = match serde_json::to_value(&config.mcp_servers) {
```
owenlin0, line 2319: why serialize with `serde_json::to_value`? these types become `Value` which feels less useful/descriptive than the real types.

i think `self.load_latest_config()` will already throw an error if it fails to read the config on disk

codex-rs/core/src/codex.rs
```diff
@@ -1631,12 +1638,85 @@ impl Session {
...
+
+        let mcp_servers =
+            match serde_json::from_value::<HashMap<String, McpServerConfig>>(mcp_servers) {
+                Ok(servers) => servers,
+                Err(err) => {
+                    warn!("failed to parse MCP server refresh config: {err}");
+                    return;
+                }
+            };
+        let store_mode = match serde_json::from_value::<OAuthCredentialsStoreMode>(
```
owenlin0, line 1678: same here

codex-rs/app-server/README.md
```diff
@@ -84,6 +84,7 @@ Example (from OpenAI's official VSCode extension):
 - `model/list` — list available models (with reasoning effort options).
 - `skills/list` — list skills for one or more `cwd` values (optional `forceReload`).
 - `mcpServer/oauth/login` — start an OAuth login for a configured MCP server; returns an `authorization_url` and later emits `mcpServer/oauthLogin/completed` once the browser flow finishes.
+- `mcpServer/refresh` — reinitialize MCP servers for loaded threads and refresh cached tool lists; returns `{}`.
```
owenlin0: what's the use case? is it:
- user updates mcp config in config.toml
- we want to reload that config without restarting the server?

maybe worth expanding on when this should be used

owenlin0: watched your demo just now (lol), what do you think about calling this `config/mcpServer/reload`? kind of generalizes to `config/<feature>/reload`, or one day might be nice to have a whole `config/reload`

shijie-oai: Ya I am a bit hesitant with calling it that because this method does not actually reload the config and tell active threads that there are new config and reload on the next turn.

https://github.com/openai/codex/pull/8957#discussion_r2677842199

shijie-oai: Actually will rename and make it clean in the readme and leave comments in code.

owenlin0: sounds good! yeah the fact that the actual reload is deferred feels like an implementation detail, but ultimately it's just reloading the config

PR #8959, opened 2026-01-09
fix: add bazelisk and dotslash to Dockerfile for devcontainer

bolinfest:
In working on the new `Dockerfile.bazel` in https://github.com/openai/codex/pull/8875, it occurred to me these would be good fixes to make for the devcontainer.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -14,6 +14,26 @@ RUN apt-get update && \
     pkg-config clang musl-tools libssl-dev just && \
     rm -rf /var/lib/apt/lists/*
 
+# Install Bazel via Bazelisk (mirrors bazelbuild/setup-bazelisk@v3).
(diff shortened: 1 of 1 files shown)


.devcontainer/Dockerfile
```diff
@@ -14,6 +14,26 @@ RUN apt-get update && \
     pkg-config clang musl-tools libssl-dev just && \
     rm -rf /var/lib/apt/lists/*
 
+# Install Bazel via Bazelisk (mirrors bazelbuild/setup-bazelisk@v3).
+ARG BAZELISK_VERSION=latest
+RUN arch="$(uname -m)" && \
+    case "$arch" in \
```
zbarsky-openai, line 20: not a huge deal but you can probably simplify this to assume your arch is always a supported one and always use `latest`, or just pin whatever is currently the latest if you want reproducibility

zbarsky-openai (approved):

PR #8971, opened 2026-01-09
add static mcp callback uri support

WhammyLeaf:
Currently the callback URI for MCP authentication is dynamically generated. More specifically, the callback URI is dynamic because the port part of it is randomly chosen by the OS. This is not ideal as callback URIs are recommended to be static and many authorization servers do not support dynamic callback URIs. 

This PR fixes that issue by exposing a new config option named `mcp_oauth_callback_port`. When it is set, the callback URI is constructed using this port rather than a random one chosen by the OS, thereby making callback URI static. 

Related issue: https://github.com/openai/codex/issues/8827

### DIFF
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -259,6 +259,12 @@ pub struct Config {
     /// auto (default): keyring if available, otherwise file.
     pub mcp_oauth_credentials_store_mode: OAuthCredentialsStoreMode,
 
+    /// Optional fixed port to use for the local HTTP callback server used during MCP OAuth login.
+    ///
+    /// When unset, Codex will bind to an ephemeral port chosen by the OS. The environment
+    /// variable `CODEX_MCP_OAUTH_CALLBACK_PORT` takes precedence over this value if set.
+    pub mcp_oauth_callback_port: Option<u16>,
+
     /// Combined provider map (defaults merged with user-defined overrides).
     pub model_providers: HashMap<String, ModelProviderInfo>,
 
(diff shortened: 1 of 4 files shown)


WhammyLeaf:

WhammyLeaf:
@aibrahim-oai Perhaps you can have a look at this?

etraut-openai:
@codex review

codex-rs/rmcp-client/src/perform_oauth_login.rs
```diff
@@ -188,19 +197,63 @@ struct OauthLoginFlow {
     timeout: Duration,
 }
 
+fn load_callback_port_from_env() -> Result<Option<u16>> {
```
gpeal: Can you remove the env var fallback? We want to keep config.toml as the source of truth rather than relying on env vars.

WhammyLeaf: @gpeal Done

WhammyLeaf:
@etraut-openai @gpeal I also tested this manually locally and it seems to work

codex-rs/core/src/config/mod.rs
```diff
@@ -259,6 +259,12 @@ pub struct Config {
     /// auto (default): keyring if available, otherwise file.
     pub mcp_oauth_credentials_store_mode: OAuthCredentialsStoreMode,
 
+    /// Optional fixed port to use for the local HTTP callback server used during MCP OAuth login.
+    ///
+    /// When unset, Codex will bind to an ephemeral port chosen by the OS. The environment
+    /// variable `CODEX_MCP_OAUTH_CALLBACK_PORT` takes precedence over this value if set.
```
gpeal: This env var reference needs to be removed, too

codex-rs/core/src/config/mod.rs
```diff
@@ -738,6 +744,11 @@ pub struct ConfigToml {
     #[serde(default)]
     pub mcp_oauth_credentials_store: Option<OAuthCredentialsStoreMode>,
 
+    /// Optional fixed port for the local HTTP callback server used during MCP OAuth login.
+    /// When unset, Codex will bind to an ephemeral port chosen by the OS. The environment
+    /// variable `CODEX_MCP_OAUTH_CALLBACK_PORT` takes precedence over this value if set.
```
gpeal: And this one

gpeal (commented):
Looks good, just need to remove some stale env var comments

codex-rs/core/src/config/mod.rs
WhammyLeaf: Ahh good point. missed those 😅

WhammyLeaf: @gpeal All done now

gpeal (approved):

PR #8975, opened 2026-01-09
chore: metrics tool call

jif-oai:


codex-rs/otel/src/traces/otel_manager.rs
```diff
@@ -418,6 +418,11 @@ impl OtelManager {
     }
 
     pub fn log_tool_failed(&self, tool_name: &str, error: &str) {
+        self.counter(
```
gt-oai: I think we shouldn't have two places in the codebase where we count `codex.tool.call` (or any metric for that matter) -- there's a danger it will lead to us overcounting. 

And maybe if this truly is necessary to have both we should pull out either the metric name as a const, or a the metric struct as a whole, so we can more easily find the call sites (vs grepping the metric name).

gt-oai (approved):

codex-rs/otel/src/traces/otel_manager.rs
jif-oai: fair point

jif-oai: Will drop the first one. This is only for chat completions that will be nuked in 2 weeks

PR #8961, opened 2026-01-09
Assemble sandbox/approval/network prompts dynamically

aibrahim-oai:
- Add a single builder for developer permissions messaging that accepts SandboxPolicy and approval policy. This builder now drives the developer “permissions” message that’s injected at session start and any time sandbox/approval settings change. 
- Trim EnvironmentContext to only include cwd, writable roots, and shell; removed sandbox/approval/network duplication and adjusted XML serialization and tests accordingly.

Follow-up: adding a config value to replace the developer permissions message for custom sandboxes.

PR #8990, opened 2026-01-09
fix(context) update environment_context on resume

dylan-hurd-oai:
## Summary
When resuming a session, we should append a new environment context message, to ensure the model is aware of the current settings, which might change (e.g. running `codex resume --yolo`)

## Testing
- [x] Tested locally
- [x] Added integration test

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -801,6 +801,7 @@ impl Session {
             InitialHistory::Resumed(_) | InitialHistory::Forked(_) => {
                 let rollout_items = conversation_history.get_rollout_items();
                 let persist = matches!(conversation_history, InitialHistory::Forked(_));
+                let append_env_context = matches!(conversation_history, InitialHistory::Resumed(_));
 
                 // If resuming, warn when the last recorded model differs from the current one.
                 if let InitialHistory::Resumed(_) = conversation_history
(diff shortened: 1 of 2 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -848,6 +846,12 @@ impl Session {
                     state.set_token_info(Some(info));
                 }
 
+                if append_env_context {
```
pakrym-oai, line 849: there is similar code when creating a new session. Is there a place where we can unify?

codex-rs/core/tests/suite/resume.rs
```diff
@@ -120,3 +124,78 @@ async fn resume_includes_initial_messages_from_reasoning_events() -> Result<()>
...
+    wait_for_event(&resumed.codex, |event| {
+        matches!(event, EventMsg::TaskComplete(_))
+    })
+    .await;
+
+    let body = resumed_mock.single_request().body_json();
+    let input = body["input"]
+        .as_array()
+        .expect("resume request input is an array");
+    let env_texts: Vec<&str> = input
```
pakrym-oai, line 187: Don't we have a helper on `single_request` for this?

pakrym-oai (approved):

dylan-hurd-oai:
closing in favor of #8961

Issue #4163, opened 2025-09-24
Named sessions for --resume


sherifkozman:
### Feature Request: Named Sessions for `--resume`

**Problem**
The current `--resume` feature shows multiple sessions without clear identifiers. For developers who work across several projects, it’s easy to get lost and accidentally pick the wrong session since there are no names or tags to distinguish them.

**Proposal**
Introduce optional **session names and tags**:
- `codex --name "checkout-fix"` to assign a human-friendly name.
- Support `--tags payments,urgent` for lightweight grouping.
- `/session name <newName>` and `/session tags +tag -tag` to rename or retag interactively.
- Update the `--resume` picker to show `name`, `repo`, `tags`, and `last_activity`.
- Allow fuzzy matching:  
  `codex --resume checkout` → resumes session with matching name.

**Why**
- Reduces confusion when resuming across many sessions.
- Keeps everything local, backwards-compatible, and low-risk.
- Minimal implementation: store metadata (`name`, `tags`, `repo`, `branch`, `last_activity`) in a `meta.json` file alongside existing session history.

**Example UX**
▸ checkout-fix [repo: shop-web] [tags: payments, urgent] [updated: 12m]
flaky-tests [repo: svc-orders] [tags: tests] [updated: 2h]
spike-mcp [repo: tools] [tags: research] [updated: 3d]

**Next Steps**
If maintainers agree this is valuable, I can prepare a PR implementing:
- Metadata file write/read
- CLI flags for `--name` and `--tags`
- Resume picker rendering improvements
- Basic interactive `/session` commands

Would love feedback on whether this direction makes sense before starting work.

### Are you interested in implementing this feature?

Yes

andylizf:
That's a good point. I've always found the session list confusing because it only shows the first message, which doesn't always reflect what the session is actually about. To address this, I implemented a tool to quickly glance at both the first and last messages for better context. https://github.com/andylizf/codexer

zippoxer:
I made a PR for this :)

https://github.com/openai/codex/pull/4741

m-ar-c:
From [the PR](https://github.com/openai/codex/pull/4741#issuecomment-3473969202) :

> Thanks for the contribution, and apologies for the slow response. We've received many PRs, and we don't have the bandwidth on the codex team to review all of them.
> 
[...]

PR #8991, opened 2026-01-09
Conversation naming

pap-openai:
Session renaming:
- `/rename my_session`
- `/rename` without arg and passing an argument in `customViewPrompt`
- AppExitInfo shows resume hint using the session name if set instead of uuid, defaults to uuid if not set
- Names are stored in `CODEX_HOME/sessions.jsonl`

Session resuming:
- codex resume <name> lookup for `CODEX_HOME/sessions.jsonl` first entry matching the name and resumes the session


codex-rs/cli/src/main.rs
```diff
@@ -144,7 +144,7 @@ struct CompletionCommand {
 
 #[derive(Debug, Parser)]
 struct ResumeCommand {
-    /// Conversation/session id (UUID). When provided, resumes this session.
+    /// Conversation/session id (UUID) or session name. UUIDs take precedence if it parses.
```
JaviSoto: 👍

codex-rs/cli/src/main.rs
```diff
@@ -823,7 +825,7 @@ mod tests {
         )
     }
 
-    fn sample_exit_info(conversation: Option<&str>) -> AppExitInfo {
+    fn sample_exit_info(conversation: Option<&str>, session_name: Option<&str>) -> AppExitInfo {
```
JaviSoto: Nit: rename for clarity?


```suggestion
    fn sample_exit_info(conversation_id: Option<&str>, session_name: Option<&str>) -> AppExitInfo {
```

codex-rs/cli/src/main.rs
```diff
@@ -832,6 +834,7 @@ mod tests {
         AppExitInfo {
             token_usage,
             thread_id: conversation.map(ThreadId::from_string).map(Result::unwrap),
+            session_name: session_name.map(str::to_string),
```
JaviSoto: Like we talked about in the call, perhaps we want to use `thread_name` here too, but not sure

codex-rs/core/src/rollout/list.rs
```diff
@@ -340,6 +311,51 @@ where
     Ok(collected)
 }
 
+async fn newest_rollout_files(
```
JaviSoto: Add a brief comment explaining this function?

codex-rs/core/src/rollout/list.rs
```diff
@@ -462,6 +478,61 @@ async fn file_modified_rfc3339(path: &Path) -> io::Result<Option<String>> {
     Ok(dt.format(&Rfc3339).ok())
 }
 
+async fn read_first_session_meta_name(path: &Path) -> io::Result<Option<String>> {
```
JaviSoto: Doc comment?

codex-rs/core/src/rollout/list.rs
```diff
@@ -462,6 +478,61 @@ async fn file_modified_rfc3339(path: &Path) -> io::Result<Option<String>> {
...
+    name: &str,
+) -> io::Result<Option<PathBuf>> {
+    if name.trim().is_empty() {
+        return Ok(None);
+    }
+
+    let mut root = codex_home.to_path_buf();
+    root.push(SESSIONS_SUBDIR);
+    if !root.exists() {
+        return Ok(None);
```
JaviSoto: We don't wanna error in this case? I feel like this could mask an issue?

codex-rs/core/src/rollout/recorder.rs
```diff
@@ -419,3 +552,62 @@ impl JsonlWriter {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use codex_protocol::ThreadId;
+    use pretty_assertions::assert_eq;
+    use tokio::io::AsyncWriteExt;
+
+    #[tokio::test]
+    async fn set_session_name_rewrites_first_session_meta_line() -> std::io::Result<()> {
```
JaviSoto: This test covers the case where the conversation doesn't have a name and we set one. Maybe add another case for when it already had a name and we're renaming it?

codex-rs/core/src/codex.rs
```diff
@@ -2121,6 +2134,110 @@ mod handlers {
...
+        use super::*;
+        use pretty_assertions::assert_eq;
+
+        #[test]
+        fn validates_session_name_tokens() {
+            assert_eq!(is_valid_session_name("my-session"), true);
+            assert_eq!(is_valid_session_name("Alpha_1.2"), true);
+            assert_eq!(is_valid_session_name("-starts-with-dash"), false);
+            assert_eq!(is_valid_session_name("has space"), false);
+            assert_eq!(is_valid_session_name("slash/name"), false);
```
JaviSoto: Do we _need_ to be this strict with the allowed names?

codex-rs/core/src/util.rs
```diff
@@ -107,4 +116,11 @@ mod tests {
 
         feedback_tags!(model = "gpt-5", cached = true, debug_only = OnlyDebug);
     }
+
+    #[test]
+    fn resume_command_prefers_name_over_id() {
```
JaviSoto, line 150: Worth adding tests for the other cases? (passing only one of the 2 and passing 2 nulls)

codex-rs/tui2/src/lib.rs
```diff
@@ -461,13 +468,14 @@ async fn run_ratatui_app(
                 let _ = tui.terminal.clear();
                 if let Err(err) = writeln!(
                     std::io::stdout(),
-                    "No saved session found with ID {id_str}. Run `codex resume` without an ID to choose from existing sessions."
+                    "No saved session found with ID or name {id_str}. Run `codex resume` without an ID to choose from existing sessions."
```
JaviSoto: Might be worth explicitly telling the user whether we tried to load it as a name or a UUID. The thing is it's actually useful that we're calling `Uuid::parse_str` now, so if for example I copy a UUID but I accidentally miss copying the last character, before we would've just failed with "not found", but now we can tell you "no saved session with name <foo>", which can give you a hint that you didn't pass a valid UUID?

codex-rs/tui/src/lib.rs
```diff
@@ -441,13 +448,14 @@ async fn run_ratatui_app(
                 let _ = tui.terminal.clear();
                 if let Err(err) = writeln!(
                     std::io::stdout(),
-                    "No saved session found with ID {id_str}. Run `codex resume` without an ID to choose from existing sessions."
+                    "No saved session found with ID or name {id_str}. Run `codex resume` without an ID to choose from existing sessions."
```
JaviSoto: Same here as the other comment re being explicit about whether we tried to search it via name or uuid

codex-rs/protocol/src/protocol.rs
```diff
@@ -1854,6 +1866,15 @@ pub struct SessionConfiguredEvent {
     pub rollout_path: PathBuf,
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize, JsonSchema, TS)]
+pub struct SessionMetaUpdatedEvent {
+    /// Name left as session_id instead of thread_id for backwards compatibility.
+    pub session_id: ThreadId,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
JaviSoto, line 2244: 👍

JaviSoto (commented):
Looks good!!

Random suggestion I just thought of: it could be useful to give the agent a way to set the name as well via a tool call, so you can tell the agent to name the conversation as well? (you could even add something to your AGENTS.md with instructions on how you want the agent to name your sessions, or to rename them in certain cases etc)

codex-rs/core/src/rollout/recorder.rs
```diff
@@ -392,6 +413,118 @@ async fn rollout_writer(
     Ok(())
 }
 
+async fn rewrite_session_name(
```
jif-oai: Those functions should not be needed. We should use the same mechanism as for the model used during the rollout for example

codex-rs/core/src/rollout/list.rs
```diff
@@ -157,101 +157,72 @@ async fn traverse_directories_for_paths(
     provider_matcher: Option<&ProviderMatcher<'_>>,
```
jif-oai, line 411: As discussed, with an indexing file, you should not need to touch this file

codex-rs/core/src/rollout/recorder.rs
```diff
@@ -207,6 +213,16 @@ impl RolloutRecorder {
             .map_err(|e| IoError::other(format!("failed waiting for rollout flush: {e}")))
     }
 
+    pub async fn set_session_name(&self, name: String) -> std::io::Result<()> {
```
jif-oai: Can you document for such public API?

codex-rs/protocol/src/protocol.rs
```diff
@@ -1854,6 +1866,15 @@ pub struct SessionConfiguredEvent {
     pub rollout_path: PathBuf,
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize, JsonSchema, TS)]
+pub struct SessionMetaUpdatedEvent {
+    /// Name left as session_id instead of thread_id for backwards compatibility.
+    pub session_id: ThreadId,
```
jif-oai: No reason to be backward compatible for something which is new

jif-oai (commented):
I would def go to the index file
+ can we try to use `thread` instead of `session` as much as possible

codex-rs/core/src/codex.rs
pap-openai: I think we don't want:
- starts-with-dash
- has space

e.g:
```
codex resume -starts-with-dash
error: invalid value 'tarts-with-dash' for '--sandbox <SANDBOX_MODE>'
``` 
or
```
codex resume test pap
No saved session found with ID test.
```

`.`, `_`, Upper Characters and `/` would work fine. updating

pap-openai: Actually will support making a safe exit hint instead of restricting

codex-rs/cli/src/main.rs
pap-openai: added this

codex-rs/tui2/src/lib.rs
pap-openai: Added this so it shows
```
$ codex resume 019bb4a5-7a07-73d1-b19b-5da943099aa1
No saved session found with UUID 019bb4a5-7a07-73d1-b19b-5da943099aa1. Run `codex resume` without an ID to choose from existing sessions.
```

or 

```
$ codex resume no-name
No saved session found with name no-name. Run `codex resume` without an ID to choose from existing sessions.
```

pap-openai:
@jif-oai @JaviSoto moved to an index file instead of SessionMeta so we don't touch rollout files. We're reading a `CODEX_HOME/session_index.jsonl` from most recent renames to oldest. We're appending new names even if previous session had a name to avoid searching for the entry with name + rewriting file and keep write O(1).

Implemented your comments (adding tests on resume, move to thread, thread_name or thread_id naming)

pap-openai:
related https://github.com/openai/codex/issues/4163

codex-rs/app-server/README.md
```diff
@@ -80,6 +80,7 @@ Example (from OpenAI's official VSCode extension):
 - `thread/list` — page through stored rollouts; supports cursor-based pagination and optional `modelProviders` filtering.
 - `thread/loaded/list` — list the thread ids currently loaded in memory.
 - `thread/archive` — move a thread’s rollout file into the archived directory; returns `{}` on success.
+- `thread/name/set` — set or update a thread’s user-facing name; returns `{}` on success.
```
jif-oai: We should document what happens on name collisions

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -1569,6 +1574,42 @@ impl CodexMessageProcessor {
         }
     }
 
+    async fn thread_set_name(&self, request_id: RequestId, params: ThreadSetNameParams) {
+        let ThreadSetNameParams { thread_id, name } = params;
+        let trimmed = name.trim();
+        if trimmed.is_empty() {
```
jif-oai: This gets validated both here and in the Op handler (codex.rs) 
Maybe we should simplify and have only one validation? (the good answer being: not here :D)

codex-rs/core/src/rollout/session_index.rs
```diff
@@ -0,0 +1,255 @@
...
+const READ_CHUNK_SIZE: usize = 8192;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct SessionIndexEntry {
+    pub id: ThreadId,
+    pub thread_name: String,
+    pub updated_at: String,
+}
+
+pub async fn append_thread_name(
```
jif-oai, line 25: Can you doc those functions?

codex-rs/core/src/state/session.rs
```diff
@@ -14,16 +14,21 @@ pub(crate) struct SessionState {
     pub(crate) session_configuration: SessionConfiguration,
     pub(crate) history: ContextManager,
     pub(crate) latest_rate_limits: Option<RateLimitSnapshot>,
+    pub(crate) thread_name: Option<String>,
```
jif-oai: Why not putting it in `SessionConfiguration` ?

codex-rs/core/src/state/service.rs
```diff
@@ -16,6 +17,7 @@ use tokio::sync::RwLock;
 use tokio_util::sync::CancellationToken;
 
 pub(crate) struct SessionServices {
+    pub(crate) codex_home: PathBuf,
```
jif-oai: this is not a service and it should already be easy to get if from `Session`

codex-rs/core/src/util.rs
```diff
@@ -72,6 +73,43 @@ pub fn resolve_path(base: &Path, path: &PathBuf) -> PathBuf {
...
+            format!("codex resume {target}")
+        }
+    })
+}
+
+fn needs_shell_escaping(value: &str) -> bool {
+    value.starts_with('-') || value.chars().any(char::is_whitespace) || value.contains('\'')
+}
+
+fn shell_escape(value: &str) -> String {
```
jif-oai: We should already have code for this. And the implementation is pretty weak
Check what's done in `core/` for this or use `shlex` or so

codex-rs/core/tests/suite/rollout_list_find.rs
```diff
@@ -4,6 +4,7 @@ use std::path::Path;
 use std::path::PathBuf;
 
 use codex_core::find_thread_path_by_id_str;
+use pretty_assertions::assert_eq;
```
jif-oai, line 15: huh?

codex-rs/core/tests/suite/rollout_list_find.rs
```diff
@@ -4,6 +4,7 @@ use std::path::Path;
...
 use tempfile::TempDir;
 use uuid::Uuid;
```
jif-oai, line 18: A true integration test would be cool

codex-rs/docs/protocol_v1.md
```diff
@@ -133,6 +133,10 @@ sequenceDiagram
     task->>-user: Event::TurnComplete
 ```
 
+### Session Metadata
```
jif-oai: huh? v1 doc should not be updated as you did not change the v1 protocol

codex-rs/tui/src/status/snapshots/codex_tui__status__tests__status_snapshot_cached_limits_hide_credits_without_flag.snap
```diff
@@ -15,6 +15,7 @@ expression: sanitized
 │  Approval:         on-request                                       │
 │  Sandbox:          read-only                                        │
 │  Agents.md:        <none>                                           │
+│  Thread name:      <none>                                           │
```
jif-oai: That looks like a lot of pollution? Could we make that this line is only displayed when there is a name

codex-rs/tui/src/chatwidget.rs
```diff
@@ -632,6 +634,13 @@ impl ChatWidget {
         }
     }
 
+    fn on_thread_name_updated(&mut self, event: codex_core::protocol::ThreadNameUpdatedEvent) {
+        if self.thread_id == Some(event.thread_id) {
+            self.thread_name = event.thread_name;
+        }
+        self.request_redraw();
```
jif-oai: You should not redraw if the `if` above fails

codex-rs/tui/src/chatwidget.rs
```diff
@@ -2192,6 +2212,23 @@ impl ChatWidget {
         }
     }
 
+    fn show_rename_prompt(&mut self) {
+        let tx = self.app_event_tx.clone();
+        let view = CustomPromptView::new(
+            "Rename thread".to_string(),
```
jif-oai: Maybe "Name" and not "Rename" as it had no name before

codex-rs/tui/src/status/snapshots/codex_tui__status__tests__status_snapshot_cached_limits_hide_credits_without_flag.snap
pap-openai: Thought it'd be good for feature discovery but agree that's also pollution

codex-rs/docs/protocol_v1.md
pap-openai: good catch, ty!

codex-rs/app-server/README.md
pap-openai: Added a wider description. A side-effect of this journaling naming session is also the fact that, while renaming a thread will work if we do:

/rename a
/rename b

then 

codex resume a # (which is the old session name) actually works

codex-rs/core/src/rollout/session_index.rs
```diff
@@ -0,0 +1,261 @@
...
+        id: thread_id,
+        thread_name: name.to_string(),
+        updated_at,
+    };
+    append_session_index_entry(codex_home, &entry).await
+}
+
+/// Append a raw session index entry to `session_index.jsonl`.
+/// The file is append-only; consumers scan from the end to find the newest match.
+pub async fn append_session_index_entry(
```
jif-oai, line 46: nit: can you inline this function?

codex-rs/core/src/rollout/session_index.rs
```diff
@@ -0,0 +1,261 @@
...
+pub async fn find_thread_name_by_id(
+    codex_home: &Path,
+    thread_id: &ThreadId,
+) -> std::io::Result<Option<String>> {
+    let path = session_index_path(codex_home);
+    if !path.exists() {
+        return Ok(None);
+    }
+    let id = *thread_id;
+    let entry = tokio::task::spawn_blocking(move || scan_index_from_end_by_id(&path, &id))
```
jif-oai, line 73: OOC why putting this in a tokio task?  If you use tokio fs implementation you shouldn't need this afaik

codex-rs/core/src/rollout/session_index.rs
```diff
@@ -0,0 +1,261 @@
...
+        return Ok(None);
+    };
+    if predicate(&entry) {
+        return Ok(Some(entry));
+    }
+    Ok(None)
+}
+
+#[cfg(test)]
+mod tests {
```
jif-oai, line 197: Can you add a test where the entry is not in the file? And one where there is multiple different entries?

codex-rs/core/src/codex.rs
```diff
@@ -503,6 +507,10 @@ pub(crate) struct SessionConfiguration {
 }
 
 impl SessionConfiguration {
+    pub(crate) fn codex_home(&self) -> &PathBuf {
+        &self.original_config_do_not_use.codex_home
```
jif-oai: I would argue we should find a way to not use something called `original_config_do_not_use` 😅

codex-rs/core/src/codex.rs
```diff
@@ -2672,6 +2701,69 @@ mod handlers {
...
+                    message: "Thread name cannot be empty.".to_string(),
+                    codex_error_info: Some(CodexErrorInfo::BadRequest),
+                }),
+            };
+            sess.send_event_raw(event).await;
+            return;
+        };
+
+        let persistence_enabled = sess.services.rollout.lock().await.is_some();
+        if !persistence_enabled {
```
jif-oai, line 3010: Good catch, small Rust tip, here you take a lock that is linked to `persistence_enabled`. But the variable only get dropped at the end of the function (after a bunch of `await` etc). It means the lock is being held way more than needed

A cleaner implementation would be to move this in a dedicated scope. That would give something like this
```
// Some code
...
{ // Open a new scope
    let persistence_enabled = sess.services.rollout.lock().await.is_some();
    // if ...
} // Close the scope, the `persistence_enabled` is dropped so the lock is released now
...
// Rest of the code
```

jif-oai (approved):
When we do a `/rename` we just see 
```
• Thread renamed to "Me"
```
Could we update a bit the rendering to add a hint below explaining how to resume the thread with this? Exactly as when we do an exit

pap-openai:
now showing 
image

I'm wondering if it won't be confusing though given the composer is still showing (we're still in the TUI) compared to when we exit (in which case we'll show codex resume <name> anyway if a name is set). Should we keep @jif-oai ?

codex-rs/core/src/codex.rs
pap-openai: tysm!

codex-rs/core/src/codex.rs
pap-openai: Yeah, had the local commit I didn't push, thanks for catching

codex-rs/core/src/rollout/session_index.rs
pap-openai: I think if we were doing a normal read we could have used this but since we read in reverse with seek/read_exact we might need this – we do something similar here i believe? https://github.com/openai/codex/blob/main/codex-rs/core/src/message_history.rs#L128

slkzgm:
Hi! Thanks for putting this together — it addresses the core pain point of making sessions easier to find via `codex resume`.

I’d like to contribute a small “minimal integration” improvement that stays within the existing flow:
- add a more explicit `/name <…>` command (possibly as an alias to `/rename`) to name the current session from inside the conversation,
- surface the name directly in the `codex resume` list/picker to improve recognizability,
- and if `codex resume <name>` is ambiguous (multiple matches), fall back to the picker rather than adding complex collision logic.

Would appreciate maintainer guidance on whether you prefer `/name` vs `/rename`, and whether `codex resume <name>` should require unique matches or just fall back to the picker on ambiguity.

If this approach sounds good, I’m happy to send a small follow-up PR (or base changes off your branch) with the tweaks + tests.

PR #8994, opened 2026-01-09
feat: wire fork to codex cli

apanasenko-oai:
## Summary
- add `codex fork` subcommand and `/fork` slash command mirroring resume
- extend session picker to support fork/resume actions with dynamic labels in tui/tui2
- wire fork selection flow through tui bootstraps and add fork-related tests

### DIFF
--- a/codex-rs/cli/src/main.rs
+++ b/codex-rs/cli/src/main.rs
@@ -117,6 +117,9 @@ enum Subcommand {
     /// Resume a previous interactive session (picker by default; use --last to continue the most recent).
     Resume(ResumeCommand),
 
+    /// Fork a previous interactive session (picker by default; use --last to fork the most recent).
+    Fork(ForkCommand),
+
     /// [EXPERIMENTAL] Browse tasks from Codex Cloud and apply changes locally.
     #[clap(name = "cloud", alias = "cloud-tasks")]
     Cloud(CloudTasksCli),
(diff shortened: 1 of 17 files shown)


codex-rs/cli/src/main.rs
```diff
@@ -161,6 +162,25 @@ struct ResumeCommand {
...
 
+#[derive(Debug, Parser)]
+struct ForkCommand {
+    /// Conversation/session id (UUID). When provided, forks this session.
+    /// If omitted, use --last to pick the most recent recorded session.
+    #[arg(value_name = "SESSION_ID")]
+    session_id: Option<String>,
+
+    /// Fork the most recent session without showing the picker.
+    #[arg(long = "last", default_value_t = false, conflicts_with = "session_id")]
```
jif-oai, line 173: Do we really want this? How is the last session defined? Last one closed? Last turn started? Last turn finished?
I guess this will create more confusion

codex-rs/cli/src/main.rs
```diff
@@ -161,6 +162,25 @@ struct ResumeCommand {
...
+    last: bool,
+
+    /// Show all sessions (disables cwd filtering and shows CWD column).
+    #[arg(long = "all", default_value_t = false)]
+    all: bool,
+
+    #[clap(flatten)]
+    config_overrides: TuiCli,
```
jif-oai, line 181: This is cool but it seems to add quite some complexity for low value

jif-oai (approved):

codex-rs/cli/src/main.rs
apanasenko-oai: It basically mirrors resume and use same logic.

```
/// Load thread file paths from disk using directory traversal.
///
/// Directory layout: `~/.codex/sessions/YYYY/MM/DD/rollout-YYYY-MM-DDThh-mm-ss-<uuid>.jsonl`
/// Returned newest (latest) first.
```
we will need to clean up it at some point or define better.

codex-rs/cli/src/main.rs
apanasenko-oai: same here, I followed `resume` here.

PR #8996, opened 2026-01-09
Add hierarchical agent prompt

pakrym-oai:


codex-rs/core/src/prompt_instructions.rs
aibrahim-oai: what is the purpose of that

codex-rs/core/src/project_doc.rs
```diff
@@ -36,35 +40,46 @@ pub(crate) async fn get_user_instructions(
     config: &Config,
     skills: Option<&[SkillMetadata]>,
 ) -> Option<String> {
-    let skills_section = skills.and_then(render_skills_section);
+    let project_docs = read_project_docs(config).await;
```
aibrahim-oai, line 43: why so big of a change for appending instructions?

codex-rs/core/src/context_manager/history.rs
```diff
@@ -85,7 +85,10 @@ impl ContextManager {
     // This is a coarse lower bound, not a tokenizer-accurate count.
     pub(crate) fn estimate_token_count(&self, turn_context: &TurnContext) -> Option<i64> {
         let model_info = turn_context.client.get_model_info();
-        let base_instructions = model_info.base_instructions.as_str();
+        let base_instructions = turn_context
```
aibrahim-oai: unrelated but useful?

aibrahim-oai (approved):

codex-rs/core/src/project_doc.rs
pakrym-oai: simplify all various places we concat things, just one function that builds the final output./

codex-rs/core/src/prompt_instructions.rs
pakrym-oai: to be teleted

codex-rs/core/src/context_manager/history.rs
pakrym-oai: will revert

PR #8997, opened 2026-01-09
chore: clarify default shell for unified_exec

sayan-oai:
The description of the `shell` arg for `exec_command` states the default is `/bin/bash`, but AFAICT it's the user's default shell. 

Default logic [here](https://github.com/openai/codex/blob/2a06d64bc996e4d848b95285700b195c2852a42f/codex-rs/core/src/tools/handlers/unified_exec.rs#L123).

EDIT: #9004 has an alternative where we inform the model of the default shell itself.


jif-oai (approved):
Ok IF it has been eval

PR #9000, opened 2026-01-09
fix(codex-api): complete chat SSE on finish_reason

joshka-oai:
## Problem
Windows CI intermittently flakes in `codex-app-server` integration tests that use WireMock to stub Chat Completions streaming (`/v1/chat/completions`). The mock expects multiple model requests, but WireMock often only sees the first request and then panics on drop with verification errors like "expected N requests, got 1".

## Root cause
Our Chat Completions SSE parser treated the transport sentinel (`DONE`/`[DONE]`) as the primary completion signal. On Windows/WireMock the HTTP connection can stay open after the model already indicates completion via `finish_reason` (e.g. `tool_calls`), and the sentinel may not be observed promptly. That means we didn’t emit `ResponseEvent::Completed`, so core kept waiting and never issued the follow-up model request.

## Fix
Treat `finish_reason` as authoritative for end-of-response and emit `ResponseEvent::Completed` immediately:
- `finish_reason == "tool_calls"`: emit all tool call items (and flush any reasoning), then complete.
- `finish_reason == "stop"`: flush assistant/reasoning, then complete.
- Still accept `DONE`/`[DONE]` as a compatibility completion signal, but correctness no longer depends on it.

To address correctness/risk concerns, the implementation now includes detailed in-code documentation about these edge cases and ordering guarantees.

## Regression coverage
Added `codex-api` unit tests that simulate the problematic condition (a stream that emits a single JSON event with `finish_reason` and then never closes / never sends `DONE/[DONE]`), including a multi-choice `tool_calls` chunk. These tests ensure we still emit a terminal `Completed` event and return without relying on sentinel/close.

Background doc: `docs/ci/windows-wiremock-chat-sse-flake.md`.

## Tests
- `cargo test -p codex-api --lib`
- `cargo test -p codex-app-server codex_message_processor_flow`

### DIFF
--- a/codex-rs/codex-api/src/sse/chat.rs
+++ b/codex-rs/codex-api/src/sse/chat.rs
@@ -36,15 +36,22 @@ pub(crate) fn spawn_chat_stream(
 /// (`data: [DONE]`). Historically, some of our test stubs have emitted `data: DONE`
 /// (without brackets) instead.
 ///
-/// `eventsource_stream` delivers these sentinels as regular events rather than signaling
-/// end-of-stream. If we try to parse them as JSON, we log and skip them, then keep
-/// polling for more events.
+/// In addition to the sentinel, the stream also encodes completion semantics in JSON via
+/// `finish_reason` on each choice:
+/// - `finish_reason == "tool_calls"`: tool calls are complete for this response and the client
+///   should proceed (typically by executing tools and issuing a follow-up request).
+/// - `finish_reason == "stop"`: the assistant message is complete.
 ///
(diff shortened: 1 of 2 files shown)


joshka-oai:
Added better docs and regression test on this to show where the problem happened in our current approach (and to ensure that this doesn't happen again)

celia-oai:
does this apply to responses API too? If so can we update the PR descriptions?

I'm prob not the best person to review this corner of the codebase - will leave the review to @pakrym-oai

PR #9004, opened 2026-01-09
use default shell in tool desc

sayan-oai:
Variant to #8997.

Inform the model of the actual default shell rather than saying `the user's default shell`.


codex-rs/core/src/tools/spec.rs
```diff
@@ -21,6 +21,7 @@ use std::collections::HashMap;
 #[derive(Debug, Clone)]
 pub(crate) struct ToolsConfig {
     pub shell_type: ConfigShellToolType,
+    pub default_shell_name: Option<String>,
```
jif-oai, line 24: Can't this be part of ConfigShellToolType?

jif-oai:
It needs to be eval

PR #9010, opened 2026-01-10
fix: include AGENTS.md as repo root marker for integration tests

bolinfest:
As explained in `codex-rs/core/BUILD.bazel`, including the repo's own `AGENTS.md` is a hack to get some tests passing. We should fix this properly, but I wanted to put stake in the ground ASAP to get `just bazel-remote-test` working and then add a job to `bazel.yml` to ensure it keeps working.

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -108,3 +108,24 @@ jobs:
             --build_metadata=ROLE=CI \
             --build_metadata=VISIBILITY=PUBLIC \
             "--remote_header=x-buildbuddy-api-key=$BUILDBUDDY_API_KEY"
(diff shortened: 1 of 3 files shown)


bolinfest:
@zbarsky-openai I'm not sure if this was the best way to do this: should I avoid create a new job to `bazel.yml` and modify the `common:linux` flags to `.github/workflows/ci.bazelrc` instead? Though this is only for x86_64 Linux at the moment, though perhaps we should crate an ARM Linux config in the root `BUILD.bazel` ASAP?

zbarsky-openai (approved):
lets land like this and do arm64 rbe as followup

PR #9011, opened 2026-01-10
fix: report an appropriate error in the TUI for malformed rules

### LINKED ISSUE #8803: v0.78.0 panics instead of graceful error when .codex/rules contains invalid Starlark

0.78.0

max

_No response_

_No response_

## Description

  Codex 0.78.0 crashes with a Rust panic when a project contains a `.codex/rules` file with invalid content (e.g., markdown instead of Starlark). Instead of a parse error, it panics in the `system-configuration` crate.

  ## Repro

  ```bash
  mkdir test-project && cd test-project
  mkdir .codex
  echo "# some markdown content" > .codex/rules
  codex
  ```

  ## Expected behavior

  ```
  Error: Invalid Starlark syntax in .codex/rules at line 1
  ```
Or other helpful graceful handling

  ## Actual behavior

  ```
  thread 'main' panicked at system-configuration-0.6.1/src/dynamic_store.rs:154:1:
  Attempted to create a NULL object.
  ```
And running `codex` crashes with no error messaging in the terminal

  ## Analysis

  - **0.77.0**: Works (likely doesn't read project-local `.codex/` directories)
  - **0.78.0**: Panics
  - Both versions bundle `system-configuration-0.6.1`

  The new project-aware configuration feature in 0.78 appears to trigger sandbox/network initialization when processing `.codex/rules`, which hits a [known upstream bug](https://github.com/mullvad/system-configuration-rs/issues/71) in the `system-configuration` crate when running under macOS Seatbelt.

  ## Environment

  - macOS 26.2 (Darwin 25.2.0)
  - Apple Silicon (arm64)
  - codex-cli 0.78.0 via npm

  ## Workaround

  Remove or fix the invalid `.codex/rules` file, or downgrade to 0.77.0.

```bash
  mkdir test-project && cd test-project
  mkdir .codex
  echo "# some markdown content" > .codex/rules
  codex
  ```

_No response_

### Additional information

_No response_

### PR DESCRIPTION
**Pull request by bolinfest**
The underlying issue is that when we encountered an error starting a conversation (any sort of error, though making `$CODEX_HOME/rules` a file rather than folder was the example in #8803), then we were writing the message to stderr, but this could be printed over by our UI framework so the user would not see it. In general, we disallow the use of `eprintln!()` in this part of the code for exactly this reason, though this was suppressed by an `#[allow(clippy::print_stderr)]`.

This attempts to clean things up by changing `handle_event()` and `handle_tui_event()` to return a `Result<AppRunControl>` instead of a `Result<bool>`, which is a new type introduced in this PR (and depends on `ExitReason`, also a new type):

```rust
#[derive(Debug)]
pub(crate) enum AppRunControl {
    Continue,
    Exit(ExitReason),
}

#[derive(Debug, Clone)]
pub enum ExitReason {
    UserRequested,
    Fatal(String),
}
```

This makes it possible to exit the primary control flow of the TUI with richer information. This PR adds `ExitReason` to the existing `AppExitInfo` struct and updates `handle_app_exit()` to print the error and exit code `1` in the event of `ExitReason::Fatal`.

I tried to create an integration test for this, but it was a bit involved, so I published it as a separate PR: https://github.com/openai/codex/pull/9166. For this PR, please have faith in my manual testing!

Fixes https://github.com/openai/codex/issues/8803.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/9011).
* #9166
* __->__ #9011

### DIFF
--- a/codex-rs/cli/src/main.rs
+++ b/codex-rs/cli/src/main.rs
@@ -26,6 +26,7 @@ use codex_execpolicy::ExecPolicyCheckCommand;
 use codex_responses_api_proxy::Args as ResponsesApiProxyArgs;
 use codex_tui::AppExitInfo;
 use codex_tui::Cli as TuiCli;
+use codex_tui::ExitReason;
 use codex_tui::update_action::UpdateAction;
 use codex_tui2 as tui2;
 use owo_colors::OwoColorize;
@@ -353,6 +354,14 @@ fn format_exit_messages(exit_info: AppExitInfo, color_enabled: bool) -> Vec<Stri
 
 /// Handle the app exit and print the results. Optionally run the update action.
 fn handle_app_exit(exit_info: AppExitInfo) -> anyhow::Result<()> {
+    match exit_info.exit_reason {
+        ExitReason::Fatal(message) => {
+            eprintln!("ERROR: {message}");
+            std::process::exit(1);
+        }
+        ExitReason::UserRequested => { /* normal exit */ }
+    }
+
     let update_action = exit_info.update_action;
     let color_enabled = supports_color::on(Stream::Stdout).is_some();
     for line in format_exit_messages(exit_info, color_enabled) {
@@ -947,6 +956,7 @@ mod tests {
             token_usage,
             thread_id: conversation.map(ThreadId::from_string).map(Result::unwrap),
             update_action: None,
+            exit_reason: ExitReason::UserRequested,
         }
     }
 
@@ -956,6 +966,7 @@ mod tests {
             token_usage: TokenUsage::default(),
             thread_id: None,
             update_action: None,
+            exit_reason: ExitReason::UserRequested,
         };
         let lines = format_exit_messages(exit_info, false);
         assert!(lines.is_empty());
--- a/codex-rs/tui/src/chatwidget/agent.rs
+++ b/codex-rs/tui/src/chatwidget/agent.rs
@@ -30,16 +30,14 @@ pub(crate) fn spawn_agent(
             ..
         } = match server.start_thread(config).await {
             Ok(v) => v,
-            #[allow(clippy::print_stderr)]
             Err(err) => {
-                let message = err.to_string();
-                eprintln!("{message}");
+                let message = format!("Failed to initialize codex: {err}");
+                tracing::error!("{message}");
                 app_event_tx_clone.send(AppEvent::CodexEvent(Event {
                     id: "".to_string(),
                     msg: EventMsg::Error(err.to_error_event(None)),
                 }));
-                app_event_tx_clone.send(AppEvent::ExitRequest);
-                tracing::error!("failed to initialize codex: {err}");
+                app_event_tx_clone.send(AppEvent::FatalExitRequest(message));
                 return;
             }
         };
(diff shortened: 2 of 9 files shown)


etraut-openai:
Will this also work for api service clients like the IDE extension? Looks like it's TUI-only?

codex-rs/tui/src/chatwidget/agent.rs
```diff
@@ -30,16 +30,14 @@ pub(crate) fn spawn_agent(
             ..
         } = match server.start_thread(config).await {
```
bolinfest, line 31: @etraut-openai to answer your question about "Will this also work for api service clients like the IDE extension?" I believe the crux of the issue is how the UI deals with an `Err` from `start_thread()`.

I believe it is already doing the right thing here:

https://github.com/openai/codex/blob/3c711f3d160f939aa668052f82535f3dbb3da9e6/codex-rs/app-server/src/codex_message_processor.rs#L1333-L1356

codex-rs/cli/src/main.rs
```diff
@@ -353,6 +354,14 @@ fn format_exit_messages(exit_info: AppExitInfo, color_enabled: bool) -> Vec<Stri
 
 /// Handle the app exit and print the results. Optionally run the update action.
 fn handle_app_exit(exit_info: AppExitInfo) -> anyhow::Result<()> {
+    match exit_info.exit_reason {
```
etraut-openai, line 357: It looks like this runs only in interactive mode. What about non-interactive mode (`codex exec`)? I think we'd want this same error printed in that case.

codex-rs/tui/src/chatwidget/agent.rs
etraut-openai: I'm working on a new PR that surfaces config parsing errors correctly to app server clients. I'll extend it to handle malformed rules errors as well.

codex-rs/tui/src/lib.rs
```diff
@@ -377,6 +377,7 @@ async fn run_ratatui_app(
                         token_usage: codex_core::protocol::TokenUsage::default(),
                         thread_id: None,
                         update_action: Some(action),
+                        fatal_error_message: ExitReason::UserRequested,
```
joshka-oai:
```suggestion
                        exit_reason: ExitReason::UserRequested,
```

joshka-oai (approved):
Approved - but make sure to fix https://github.com/openai/codex/pull/9011#discussion_r2688330556

codex-rs/tui/src/lib.rs
bolinfest: Good catch! Fixed and ran `cargo build --bin codex --release` locally to verify the fix.

codex-rs/cli/src/main.rs
bolinfest: This already appears to work as intended for `codex exec`:

```
$ ./codex-rs/target/debug/codex exec hi
Error: Fatal error: failed to load execpolicy: failed to read execpolicy files from /Users/mbolin/.codex/rules: Not a directory (os error 20)
```

PR #9017, opened 2026-01-10
Better handling skill depdenencies on ENV VAR.

xl-openai:
An experimental flow for env var skill dependencies. Skills can now declare required env vars in SKILL.md; if missing, the CLI prompts the user to get the value, and Core will store it in memory (eventually to a local persistent store)
image


codex-rs/core/src/skills/env_store.rs
```diff
@@ -0,0 +1,114 @@
...
+use codex_keyring_store::KeyringStore;
+use tracing::warn;
+
+const KEYRING_SERVICE: &str = "Codex Env Vars";
+const ENV_VARS_FILE: &str = "env_vars.json";
+
+/// Env var persistence uses keyring first, then falls back to a local file.
+/// Keys are the raw env var names, without skill-level namespacing.
+/// The file lives at `codex_home/env_vars.json` with 0600 permissions on Unix.
+pub(crate) fn load_env_var(codex_home: &Path, name: &str) -> io::Result<Option<String>> {
```
mzeng-openai: Not a blocker but we should consider consolidating this with the existing logic in codex-rs/rmcp-client/src/oauth.rs. Feels like we should have a secret manager for handling these things.

codex-rs/core/src/codex.rs
```diff
@@ -3109,6 +3126,15 @@ pub(crate) async fn run_turn(
     )
     .await;
 
+    let config = turn_context.client.config();
```
mzeng-openai, line 3227: Should this happen before we spin up missing mcp servers?

codex-rs/core/src/skills/env_var_dependencies.rs
```diff
@@ -0,0 +1,182 @@
...
+        return;
+    }
+
+    let mut values = HashMap::new();
+    for (name, answer) in response.answers {
+        let mut user_note = None;
+        for entry in &answer.answers {
+            if let Some(note) = entry.strip_prefix("user_note: ") {
+                if !note.trim().is_empty() {
+                    user_note = Some(note.trim().to_string());
```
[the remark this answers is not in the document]
mzeng-openai: Seems like a legit concern.

mzeng-openai (approved):
Stamping to unblock! Please take a look at the issue that Codex spotted if possible.

codex-rs/core/src/skills/env_var_dependencies.rs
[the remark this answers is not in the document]
xl-openai: I think all clients should follow the same behavior, there is no reason to only set "user_note:" for TUI.

codex-rs/core/src/skills/env_store.rs
viyatb-oai: this is coming soon!

codex-rs/core/src/skills/env_var_dependencies.rs
```diff
@@ -0,0 +1,160 @@
...
+use crate::codex::Session;
+use crate::codex::TurnContext;
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct SkillDependencyInfo {
+    pub(crate) skill_name: String,
+    pub(crate) name: String,
+    pub(crate) description: Option<String>,
+}
+use crate::skills::SkillMetadata;
+/// Resolve required dependency values (session cache, then env vars),
```
viyatb-oai, line 22: nit:new line between imports and function signature?

codex-rs/core/src/skills/env_var_dependencies.rs
```diff
@@ -0,0 +1,160 @@
...
+    let existing_env = sess.dependency_env().await;
+    let mut loaded_values = HashMap::new();
+    let mut missing = Vec::new();
+    let mut seen_names = HashSet::new();
+
+    for dependency in dependencies {
+        let name = dependency.name.clone();
+        if !seen_names.insert(name.clone()) {
+            continue;
+        }
```
viyatb-oai, line 42: can you use Hashmap.difference method?

viyatb-oai (approved):
minor nits but lgtm

codex-rs/core/src/skills/env_var_dependencies.rs
xl-openai: difference only gives you set membership and drops per‑item context; here we must keep each dependency’s metadata (skill name/description)

codex-rs/core/src/skills/env_var_dependencies.rs
```diff
@@ -0,0 +1,160 @@
...
+        if existing_env.contains_key(&name) {
+            continue;
+        }
+        match env::var(&name) {
+            Ok(value) => {
+                loaded_values.insert(name.clone(), value);
+                continue;
```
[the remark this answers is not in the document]
xl-openai: It's OK for now.

PR #9034, opened 2026-01-10
fix: update tests so `bazel test //codex-rs/core:core-all-test` passes on my Mac

bolinfest:
According to Codex:

 The Bazel-only failures were caused by two environment differences: (1) login shells are much slower under Bazel’s sandbox, so “quick” commands timed out and returned session IDs
  instead of complete outputs; and (2) Bazel runfiles/current dir aren’t writable, so “outside workspace” paths that used to be writable under cargo test became blocked or
  misclassified. The changes below remove those sources of nondeterminism.

  Changes and why they fixed the tests

  - Default non‑login shells in test events so commands finish within expected timeouts. Under Bazel, zsh -lc (or any login shell) was adding seconds of startup latency; that caused
    exec_command to return a session_id and made apply_patch/shell outputs mismatch expected fixtures. I forced login: false in test helper events and shell command fixtures so the
    runtime uses non‑login shells and returns full output immediately.
      - codex-rs/core/tests/common/responses.rs
      - codex-rs/core/tests/suite/shell_serialization.rs
      - codex-rs/core/tests/suite/approvals.rs
      - codex-rs/core/tests/suite/unified_exec.rs
  - Use non‑login heredoc invocations for apply_patch in tests. Bazel’s sandbox + login shell created extra overhead and sometimes altered output timing. Switching to bash -c keeps
    behavior stable and quick.
      - codex-rs/core/tests/common/responses.rs
  - Make “outside workspace” truly outside and writable across Bazel and Cargo. env::current_dir() was often a read‑only runfiles path under Bazel. I now create a stable
    outside‑workspace root in /var/tmp or $HOME (falling back to tempdir()), and explicitly avoid /tmp/TMPDIR so WorkspaceWrite doesn’t treat it as writable by default. This makes
    approval paths deterministic and fixes the approval matrix under Bazel.
      - codex-rs/core/tests/suite/approvals.rs
  - Clippy fix from just fix -p codex-core: apply collapsible_if in exec_command_args. This was required to complete the mandated lint step.
      - codex-rs/core/tests/suite/unified_exec.rs

  Why these specifically addressed the failures you listed

  - The apply_patch, shell_serialization, and unified_exec failures were all caused by “fast” commands running too slowly under Bazel’s login shells, leading to timeouts or session_id
    outputs instead of completed outputs. Setting login: false and avoiding -lc stabilized command timing and restored expected output structure.
  - The approvals matrix failed because “outside workspace” paths were resolved into non‑writable Bazel runfiles or into writable tmp roots that the WorkspaceWrite policy already
    allows. The new outside‑workspace root is both writable and outside the WorkspaceWrite defaults, so approvals behave consistently.

### DIFF
--- a/codex-rs/core/tests/common/responses.rs
+++ b/codex-rs/core/tests/common/responses.rs
@@ -509,12 +509,17 @@ pub fn ev_apply_patch_function_call(call_id: &str, patch: &str) -> Value {
 }
 
 pub fn ev_shell_command_call(call_id: &str, command: &str) -> Value {
-    let args = serde_json::json!({ "command": command });
+    let args = serde_json::json!({ "command": command, "login": false });
     ev_shell_command_call_with_args(call_id, &args)
 }
 
 pub fn ev_shell_command_call_with_args(call_id: &str, args: &serde_json::Value) -> Value {
-    let arguments = serde_json::to_string(args).expect("serialize shell command arguments");
+    let mut args = args.clone();
+    if let serde_json::Value::Object(map) = &mut args {
+        map.entry("login".to_string())
+            .or_insert_with(|| serde_json::Value::Bool(false));
+    }
+    let arguments = serde_json::to_string(&args).expect("serialize shell command arguments");
     ev_function_call(call_id, "shell_command", &arguments)
 }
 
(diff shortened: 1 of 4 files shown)


zbarsky-openai:
> Under Bazel, zsh -lc (or any login shell) was adding seconds of startup latency

That's fascinating, did Codex have any theories for why that happens?

PR #9043, opened 2026-01-10
fix: prevent repeating interrupted turns

### LINKED ISSUE #9042: Interrupted turns aren’t recorded in history, causing follow-up turns to repeat aborted work
## Summary
When a turn is interrupted (Stop / Ctrl+C), Codex emits a `TurnAborted` event but does not record any model-visible marker into the conversation history. On the next user turn, the model can’t tell the previous work was aborted and may resume/repeat earlier actions (including re-opening PRs).

## Steps to reproduce
Start a user turn that triggers a tool call with side effects (e.g. `git push`, `gh pr create`, or any write).
Interrupt the turn while the tool is still running.
On the next user turn, ask Codex to do something else.
Observe that Codex may repeat the aborted work from the previous turn.

## Expected
After interrupting, follow-up turns should not automatically continue/repeat work from the aborted turn unless the user explicitly asks.

## Actual
Follow-up turns sometimes pick up the aborted plan and repeat earlier steps; this can cause duplicated side effects (e.g. a second PR for the same change).

## Notes / suspected fix
Record a hidden, model-visible marker in history when a turn is interrupted (e.g. `<turn_aborted>…</turn_aborted>`), and treat it like `<environment_context>` (session prefix) so it doesn’t change user-turn boundaries/prompt caching. The marker should instruct the model to avoid continuing/repeating aborted work unless explicitly requested and to verify state before retrying.

Related but distinct: #8310 (resume-after-rate-limit drift/repeat).

swordfish444:
## What
Record a model-visible `<turn_aborted>` marker in history when a turn is interrupted, and treat it as a session prefix.

## Why
When a turn is interrupted, Codex emits `TurnAborted` but previously did not persist anything model-visible in the conversation history. On the next user turn, the model can’t tell the previous work was aborted and may resume/repeat earlier actions (including duplicated side effects like re-opening PRs).

Fixes: https://github.com/openai/codex/issues/9042

## How
On `TurnAbortReason::Interrupted`, append a hidden user message containing a `<turn_aborted>…</turn_aborted>` marker and flush.
Treat `<turn_aborted>` like `<environment_context>` for session-prefix filtering.
Add a regression test to ensure follow-up turns don’t repeat side effects from an aborted turn.

## Testing
`just fmt`
`just fix -p codex-core`
`cargo test -p codex-core -- --test-threads=1`
`cargo test --all-features -- --test-threads=1`


swordfish444:

swordfish444:
@jif-oai would you mind taking a look when you have a chance?

This PR fixes the "interrupt mid-turn → later turns repeat aborted work" issue by recording a model-visible `<turn_aborted>` marker in history (and includes a regression test).

Appreciate any review — thank you!

swordfish444:

sgraika:

etraut-openai:
@codex review

codex-rs/core/src/tasks/mod.rs
```diff
@@ -235,6 +237,21 @@ impl Session {
             .abort(session_ctx, Arc::clone(&task.turn_context))
             .await;
 
+        if reason == TurnAbortReason::Interrupted {
+            let marker = ResponseItem::Message {
+                id: None,
+                role: "user".to_string(),
+                content: vec![ContentItem::InputText {
+                    text: format!(
+                        "<turn_aborted>\n  <turn_id>{sub_id}</turn_id>\n  <reason>interrupted</reason>\n  <guidance>The user interrupted the previous turn. Do not continue or repeat work from that turn unless the user explicitly asks. If any tools/commands were aborted, they may have partially executed; verify current state before retrying.</guidance>\n</turn_aborted>"
```
jif-oai: Could you move this as a const?

codex-rs/core/src/context_manager/history.rs
```diff
@@ -319,7 +319,7 @@ fn estimate_reasoning_length(encoded_len: usize) -> usize {
 fn is_session_prefix(text: &str) -> bool {
     let trimmed = text.trim_start();
     let lowered = trimmed.to_ascii_lowercase();
-    lowered.starts_with("<environment_context>")
+    lowered.starts_with("<environment_context>") || lowered.starts_with("<turn_aborted>")
```
jif-oai: Can you document as the function name is not very explicit anymore

codex-rs/core/src/tasks/mod.rs
```diff
@@ -235,6 +237,21 @@ impl Session {
...
+                    ),
+                }],
+            };
+            self.record_conversation_items(task.turn_context.as_ref(), &[marker])
```
jif-oai: @apanasenko-oai FYI this means we gonna call `send_raw_response_items`. Ok for you?

codex-rs/core/src/tasks/mod.rs
```diff
@@ -235,6 +237,21 @@ impl Session {
...
+                .await;
+            self.flush_rollout().await;
```
jif-oai, line 266: Why do you force a flush?

codex-rs/core/src/codex.rs
```diff
@@ -3736,14 +3736,18 @@ mod tests {
...
 
-        let evt = tokio::time::timeout(std::time::Duration::from_secs(2), rx.recv())
-            .await
-            .expect("timeout waiting for event")
-            .expect("event");
-        match evt.msg {
-            EventMsg::TurnAborted(e) => assert_eq!(TurnAbortReason::Interrupted, e.reason),
-            other => panic!("unexpected event: {other:?}"),
-        }
+        let aborted = loop {
```
jif-oai: Ok but why?

codex-rs/core/src/codex.rs
```diff
@@ -3765,11 +3769,18 @@ mod tests {
 
         sess.abort_all_tasks(TurnAbortReason::Interrupted).await;
 
-        let evt = rx.recv().await.expect("event");
-        match evt.msg {
-            EventMsg::TurnAborted(e) => assert_eq!(TurnAbortReason::Interrupted, e.reason),
-            other => panic!("unexpected event: {other:?}"),
-        }
+        let aborted = loop {
```
jif-oai: Same question

codex-rs/core/src/event_mapping.rs
```diff
@@ -24,7 +24,7 @@ use crate::user_shell_command::is_user_shell_command_text;
 fn is_session_prefix(text: &str) -> bool {
     let trimmed = text.trim_start();
     let lowered = trimmed.to_ascii_lowercase();
-    lowered.starts_with("<environment_context>")
+    lowered.starts_with("<environment_context>") || lowered.starts_with("<turn_aborted>")
```
jif-oai: Same comment on the doc

codex-rs/core/tests/suite/interrupt_does_not_repeat_aborted_turn.rs
```diff
@@ -0,0 +1,248 @@
+#![allow(clippy::expect_used, clippy::unwrap_used)]
```
jif-oai: Drop this and instead make your tests return an anyhow

codex-rs/core/tests/suite/interrupt_does_not_repeat_aborted_turn.rs
```diff
@@ -0,0 +1,248 @@
...
+use wiremock::Request;
+use wiremock::Respond;
+use wiremock::ResponseTemplate;
+use wiremock::matchers::method;
+use wiremock::matchers::path_regex;
+
+#[derive(Clone, Default)]
+struct RequestsCapture(Arc<Mutex<Vec<Request>>>);
+
+impl RequestsCapture {
```
jif-oai: We already have an extensive test harness, please use this one as much as possible and improve it if this is not sufficient (it should be for this)

codex-rs/core/tests/suite/mod.rs
```diff
@@ -31,6 +31,7 @@ mod exec_policy;
 mod fork_thread;
 mod grep_files;
 mod hierarchical_agents;
+mod interrupt_does_not_repeat_aborted_turn;
```
jif-oai: This is not actually what it test. It test that a user message is persisted in the context sent to the model

swordfish444: Yep. I dropped the old suite module and replaced it with `interrupt_persists_turn_aborted_marker_in_next_request` in `abort_tasks.rs` (asserts the `<turn_aborted>` marker is included in the follow-up `/responses` request).

codex-rs/core/tests/suite/interrupt_does_not_repeat_aborted_turn.rs
swordfish444: Done — I removed the bespoke suite test and rewrote it using the existing `core_test_support` harness in `abort_tasks.rs`.

codex-rs/core/tests/suite/interrupt_does_not_repeat_aborted_turn.rs
swordfish444: Done — I removed that test file entirely and replaced it with a harness-based integration test in `abort_tasks.rs`.

codex-rs/core/src/event_mapping.rs
swordfish444: Done — I centralized the session-prefix docs + detection in `session_prefix.rs` and wired both `event_mapping.rs` + `history.rs` to use it.

codex-rs/core/src/codex.rs
swordfish444: Because the interrupt marker is now persisted silently (no extra `RawResponseItem` event), the only client-visible event emitted by `abort_all_tasks(..Interrupted)` is `TurnAborted`. Tests were updated to assert that and to validate the marker via history instead.

codex-rs/core/src/codex.rs
swordfish444: Same reason — we no longer emit the interrupt marker as a `RawResponseItem`, so tests shouldn't expect extra events. The updated tests assert `TurnAborted` and check the marker via recorded history.

codex-rs/core/src/tasks/mod.rs
swordfish444: So the `<turn_aborted>` marker is durably in the rollout before we emit `TurnAborted` — some clients re-read the rollout immediately on receipt of the abort event.

codex-rs/core/src/tasks/mod.rs
swordfish444: Ack — updated: we no longer call `send_raw_response_items` here. The marker is recorded into history + persisted to rollout only (no extra events).

codex-rs/core/src/context_manager/history.rs
swordfish444: Done — added docs in `session_prefix.rs` explaining what “session prefix” means (model-visible metadata that must not create user-turn boundaries). `history.rs` now uses `is_session_prefix(..)` from there.

codex-rs/core/src/tasks/mod.rs
swordfish444: Done — extracted into `TURN_ABORTED_INTERRUPTED_GUIDANCE`.

codex-rs/core/src/codex.rs
joshka-oai: It may make sense to add explanations for these in the code as comments. These make it easy for a future developer or agent to understand the context without having to dig through commits, PR, and comments to find a rationale. WDYT?

swordfish444: Makes sense — agreed. I added a couple short comments in `codex-rs/core/src/codex.rs` to capture the rationale (history-only `<turn_aborted>` marker, and why we scan for `ExitedReviewMode`/`TurnAborted` rather than relying on fixed indices). Thanks!

cgraika127:
I want to explicitly call out that this fix was implemented by @swordfish444 (my brother), and I’ve reviewed the full diff + tests closely because this bug had a real, material impact on my day-to-day use of Codex.

In v0.80.0, interrupted turns not being persisted as model-visible state caused follow-up turns to resume or repeat aborted work, often re-triggering side effects. It was disruptive enough that I had to downgrade to v0.78.0 to stay productive.

What I really appreciate about this PR is that it addresses the problem at the correct abstraction layer:
	•	A durable, model-visible <turn_aborted> marker is recorded into history at interrupt time
	•	The marker is treated consistently as a session prefix (alongside environment_context)
	•	The marker is flushed before emitting TurnAborted, ensuring deterministic visibility
	•	Event ordering edge cases are handled explicitly rather than assumed
	•	Comprehensive regression tests assert that follow-up requests carry the marker and do not repeat aborted side effects

This isn’t a workaround or a heuristic — it’s a clean, well-scoped fix that aligns the event stream, session history, and model context in a way that’s robust under interruption, retries, and concurrency.

From a user perspective, this dramatically improves reliability for longer or interruption-prone Codex workflows. From a code perspective, the solution feels intentional, well-tested, and future-proof.

Strongly hope this lands soon — it’s a big quality-of-life improvement for anyone using Codex seriously.

Thank you OpenAi/Codex team for your wonderful product!

jif-oai:
This contains a behavioural change of the model so we are discussing internally on the best approach to take here and will come back to you shortly

jif-oai:
@codex review

jif-oai (approved):

etraut-openai:
@swordfish444, looks like this PR is close. There are a few CI failures that still need to be addressed. You can ignore the Bazel ones, but the other linter issues need to be fixed. Once that's done, I think this will be good to merge!

swordfish444:
Thanks! I fixed the non-Bazel CI failures (missing `text_elements` in a couple `UserInput::Text` initializers in `codex-rs/core/tests/suite/abort_tasks.rs`).

Pushed a small follow-up commit; locally `just fmt`, `just fix -p codex-core`, and `cargo test -p codex-core --all-features -- --test-threads=1` are all green.

Happy to rerun CI (non-Bazel) once approved.

etraut-openai:
@codex review

codex-rs/core/src/tasks/mod.rs
```diff
@@ -235,6 +240,25 @@ impl Session {
             .abort(session_ctx, Arc::clone(&task.turn_context))
             .await;
 
+        if reason == TurnAbortReason::Interrupted {
+            let marker = ResponseItem::Message {
+                id: None,
+                role: "user".to_string(),
+                content: vec![ContentItem::InputText {
+                    text: format!(
+                        "{TURN_ABORTED_OPEN_TAG}\n  <turn_id>{sub_id}</turn_id>\n  <reason>interrupted</reason>\n  <guidance>{TURN_ABORTED_INTERRUPTED_GUIDANCE}</guidance>\n</turn_aborted>"
```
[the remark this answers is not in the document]
swordfish444: Good catch — fixed.

Local compaction now preserves `<turn_aborted>` markers by explicitly collecting them from history (even though `parse_turn_item` treats them as a session prefix). Added a regression test (`build_compacted_history_preserves_turn_aborted_markers`) in `codex-rs/core/src/compact.rs` to prevent this from silently dropping on auto-compaction.

swordfish444:
Quick update: merged latest `main` to resolve conflicts and re-ran local checks.

- `cd codex-rs && just fmt`
- `cd codex-rs && just fix -p codex-core`
- `cd codex-rs && cargo test --all-features -- --test-threads=1`

All green locally.

Also addressed the compaction concern from the Codex bot: local compaction now preserves `<turn_aborted>` markers (with a regression test in `codex-rs/core/src/compact.rs`).

swordfish444:
@etraut-openai This is ready to go!

swordfish444:
@etraut-openai Do you need me to keep updating the branch or should I let you all take it from here?

etraut-openai:
@swordfish444, thanks again for the contribution. Apologies for taking a while to approve it. This change touches some areas that have high regression risk, so we wanted to be extra cautious before merging this one.

PR #9049, opened 2026-01-11
Handle image paste from empty paste events

aibrahim-oai:
Handle image paste on empty paste events.

- Intent: make image paste work in terminals that emit empty paste events.
- Approach: route paste events through an image-aware handler and read the clipboard when text is empty.
- That's best effort to detect it. Some terminals don't send the empty signal.

### DIFF
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -1642,6 +1642,12 @@ impl ChatWidget {
         self.request_redraw();
     }
 
+    /// Attempt to attach an image from the system clipboard.
+    ///
+    /// This is a best-effort path used when we receive an empty paste event,
+    /// which some terminals emit when the clipboard contains non-text data
+    /// (like images). When the clipboard can't be read or no image exists,
+    /// surface a helpful follow-up so the user can retry with a file path.
     fn paste_image_from_clipboard(&mut self) {
         match paste_image_to_temp_png() {
             Ok((path, info)) => {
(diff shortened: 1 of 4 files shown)


codex-rs/tui/src/chatwidget.rs
```diff
@@ -1658,6 +1642,26 @@ impl ChatWidget {
...
+                    info.height,
+                    info.encoded_format.label()
+                );
+                self.attach_image(path);
+            }
+            Err(err) => {
+                tracing::warn!("failed to paste image: {err}");
+                self.add_to_history(history_cell::new_error_event(format!(
+                    "Failed to paste image: {err}",
+                )));
```
joshka-oai, line 1666: Can we make this message not be a dead end - i.e. tell the user what they can do next to succeed with their intended action. Something like: "Try saving the image on your clipboard somewhere and then copying the path in here" (let codex wordsmith some better suggestions on this).

joshka-oai (approved):
Add more context as doc comments around why we're making choices around this code. There's a lot of nuance that's not immediately obvious / discoverable here, and it would be good to capture these rather than having to dig for it in commit / PR history.

Do we have more details on which terminals act this way? Can we link it to any issues? What testing was done on this?

codex-rs/tui/src/chatwidget.rs
aibrahim-oai: good idea

aibrahim-oai: @codex address both comments

PR #9077, opened 2026-01-12
Send message by default mid turn. queue messages by tab

aibrahim-oai:
https://github.com/user-attachments/assets/03838730-4ddc-44df-a2c7-cb8ecda78660


pakrym-oai (approved):
Needs a better UI visible name

aibrahim-oai:
@codex fix ci

PR #9101, opened 2026-01-12
Restrict MCP servers from `requirements.toml`

gt-oai:
Enterprises want to restrict the MCP servers their users can use.

Admins can now specify an allowlist of MCPs in `requirements.toml`. The MCP servers are matched on both Name and Transport (local path or HTTP URL) -- both must match to allow the MCP server. This prevents circumventing the allowlist by renaming MCP servers in user config. (It is still possible to replace the local path e.g. rewrite say `/usr/local/github-mcp` with a nefarious MCP. We could allow hash pinning in the future, but that would break updates. I also think this represents a  broader, out-of-scope problem.)

We introduce a new field to Constrained: "normalizer". In general, it is a fn(T) -> T and applies when `Constrained<T>.set()` is called. In this particular case, it disables MCP servers which do not match the allowlist. An alternative solution would remove this and instead throw a ConstraintError. That would stop Codex launching if any MCP server was configured which didn't match. I think this is bad.

We currently reuse the enabled flag on MCP servers to disable them, but don't propagate any information about why they are disabled. I'd like to add that in a follow up PR, possibly by switching out enabled with an enum.

In action:

```
# MCP server config has two MCPs. We are going to allowlist one of them.
➜  codex git:(gt/restrict-mcps) ✗ cat ~/.codex/config.toml | grep mcp_servers -A1
[mcp_servers.hello_world]
command = "hello-world-mcp"
--
[mcp_servers.docs]
command = "docs-mcp"

# Restrict the MCPs to the hello_world MCP.
➜  codex git:(gt/restrict-mcps) ✗ defaults read com.openai.codex requirements_toml_base64 | base64 -d
[mcp_server_allowlist.hello_world]
command = "hello-world-mcp"

# List the MCPs, observe hello_world is enabled and docs is disabled.
➜  codex git:(gt/restrict-mcps) ✗ just codex mcp list
cargo run --bin codex -- "$@"
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.25s
     Running `target/debug/codex mcp list`
Name         Command          Args  Env  Cwd  Status    Auth
docs         docs-mcp         -     -    -    disabled  Unsupported
hello_world  hello-world-mcp  -     -    -    enabled   Unsupported

# Remove the restrictions.
➜  codex git:(gt/restrict-mcps) ✗ defaults delete com.openai.codex requirements_toml_base64

# Observe both MCPs are enabled.
➜  codex git:(gt/restrict-mcps) ✗ just codex mcp list
cargo run --bin codex -- "$@"
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.25s
     Running `target/debug/codex mcp list`
Name         Command          Args  Env  Cwd  Status   Auth
docs         docs-mcp         -     -    -    enabled  Unsupported
hello_world  hello-world-mcp  -     -    -    enabled  Unsupported

# A new requirements that updates the command to one that does not match.
➜  codex git:(gt/restrict-mcps) ✗ cat ~/requirements.toml
[mcp_server_allowlist.hello_world]
command = "hello-world-mcp-v2"

# Use those requirements.
➜  codex git:(gt/restrict-mcps) ✗ defaults write com.openai.codex requirements_toml_base64 "$(base64 -i /Users/gt/requirements.toml)"

# Observe both MCPs are disabled.
➜  codex git:(gt/restrict-mcps) ✗ just codex mcp list
cargo run --bin codex -- "$@"
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.75s
     Running `target/debug/codex mcp list`
Name         Command          Args  Env  Cwd  Status    Auth
docs         docs-mcp         -     -    -    disabled  Unsupported
hello_world  hello-world-mcp  -     -    -    disabled  Unsupported
```

### DIFF
--- a/codex-rs/core/src/config/constraint.rs
+++ b/codex-rs/core/src/config/constraint.rs
@@ -37,11 +37,15 @@ impl From<ConstraintError> for std::io::Error {
 }
 
 type ConstraintValidator<T> = dyn Fn(&T) -> ConstraintResult<()> + Send + Sync;
+/// A ConstraintNormalizer is a function which transforms a value into another of the same type.
+/// `Constrained` uses normalizers to transform values to satisfy constraints or enforce values.
+type ConstraintNormalizer<T> = dyn Fn(T) -> T + Send + Sync;
 
 #[derive(Clone)]
 pub struct Constrained<T> {
     value: T,
     validator: Arc<ConstraintValidator<T>>,
+    normalizer: Option<Arc<ConstraintNormalizer<T>>>,
 }
 
 impl<T: Send + Sync> Constrained<T> {
@@ -54,6 +58,23 @@ impl<T: Send + Sync> Constrained<T> {
         Ok(Self {
             value: initial_value,
             validator,
+            normalizer: None,
+        })
+    }
+
+    /// normalized creates a `Constrained` value with a normalizer function and a validator that allows any value.
+    pub fn normalized(
+        initial_value: T,
+        normalizer: impl Fn(T) -> T + Send + Sync + 'static,
+    ) -> ConstraintResult<Self> {
+        let validator: Arc<ConstraintValidator<T>> = Arc::new(|_| Ok(()));
+        let normalizer: Arc<ConstraintNormalizer<T>> = Arc::new(normalizer);
+        let normalized = normalizer(initial_value);
+        validator(&normalized)?;
+        Ok(Self {
+            value: normalized,
+            validator,
+            normalizer: Some(normalizer),
         })
     }
 
@@ -61,6 +82,7 @@ impl<T: Send + Sync> Constrained<T> {
         Self {
             value: initial_value,
             validator: Arc::new(|_| Ok(())),
+            normalizer: None,
         }
     }
 
@@ -88,6 +110,11 @@ impl<T: Send + Sync> Constrained<T> {
     }
 
     pub fn set(&mut self, value: T) -> ConstraintResult<()> {
+        let value = if let Some(normalizer) = &self.normalizer {
+            normalizer(value)
+        } else {
+            value
+        };
         (self.validator)(&value)?;
         self.value = value;
         Ok(())
@@ -141,6 +168,17 @@ mod tests {
     fn constrained_allow_any_default_uses_default_value() {
         let constrained = Constrained::<i32>::allow_any_from_default();
         assert_eq!(constrained.value(), 0);
+    }
+
+    #[test]
+    fn constrained_normalizer_applies_on_init_and_set() -> anyhow::Result<()> {
+        let mut constrained = Constrained::normalized(-1, |value| value.max(0))?;
+        assert_eq!(constrained.value(), 0);
+        constrained.set(-5)?;
+        assert_eq!(constrained.value(), 0);
+        constrained.set(10)?;
+        assert_eq!(constrained.value(), 10);
+        Ok(())
     }
 
     #[test]
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -2,6 +2,7 @@ use crate::auth::AuthCredentialsStoreMode;
 use crate::config::types::DEFAULT_OTEL_ENVIRONMENT;
 use crate::config::types::History;
 use crate::config::types::McpServerConfig;
+use crate::config::types::McpServerTransportConfig;
 use crate::config::types::Notice;
 use crate::config::types::Notifications;
 use crate::config::types::OtelConfig;
@@ -16,6 +17,8 @@ use crate::config::types::UriBasedFileOpener;
 use crate::config_loader::ConfigLayerStack;
 use crate::config_loader::ConfigRequirements;
 use crate::config_loader::LoaderOverrides;
+use crate::config_loader::McpServerIdentity;
+use crate::config_loader::McpServerRequirement;
 use crate::config_loader::load_config_layers_state;
 use crate::features::Feature;
 use crate::features::FeatureOverrides;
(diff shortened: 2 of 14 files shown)


codex-rs/core/src/config_loader/config_requirements.rs
```diff
@@ -43,22 +44,32 @@ impl fmt::Display for RequirementSource {
...
             sandbox_policy: Constrained::allow_any(SandboxPolicy::ReadOnly),
+            mcp_server_allowlist: None,
         }
     }
 }
 
+#[derive(Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum McpServerIdentityRequirement {
+    Command { command: String },
```
bolinfest, line 63: Ah, so we require `command` but not `args`?

https://developers.openai.com/codex/mcp

I suppose that's more flexible, though I don't know if we will want to have `args` or `args_prefix` as a constraint option at some point?

There's also `enabled_tools` and `disabled_tool` mentioned in the docs: should those be able to be constrained, as well?

codex-rs/core/src/config_loader/config_requirements.rs
```diff
@@ -523,4 +552,36 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn deserialize_mcp_server_allowlist() -> Result<()> {
+        let toml_str = r#"
+            [mcp_server_allowlist.docs]
```
bolinfest: The name `mcp_server_allowlist` feels a bit off to me, especially since the other two config options start with `allowed_`. I'm not sure I like `allowed_mcp_servers` that much better, but at least it's more consistent?

bolinfest (commented):
Hmm, so if you declare one MCP server in `requirements.toml`, that means an end user can't use any MCP servers of their own, correct?

Do we need a way to say, "If you use MCP server xyz, then this is how you must configure it?" I realize that's arguably what should go in `/etc/config/codex.toml`, so maybe that doesn't really make sense either.

I'm just having trouble envisioning an enterprise enumerating all the accepted MCP servers, though I guess we do it with Chrome extensions?

codex-rs/core/src/config_loader/config_requirements.rs
gt-oai: > I don't know if we will want to have args or args_prefix as a constraint option at some point

I couldn't envisage this, but that doesn't mean it won't happen.

> There's also enabled_tools and disabled_tool mentioned in the docs: should those be able to be constrained, as well?

Eventually we might have to support this, and it's important that this design have space for that. 

Perhaps I can move the identity requirement one level down such that there is room for these other options of allowed tools and required args in the future. I don't think they're necessary to launch with, but agree we should be able to add them later in a backwards-compatible way.

codex-rs/core/src/config_loader/config_requirements.rs
gt-oai: yes was torn here -- I think I wanted to include the word "allowlist" explicitly, in case we wanted to add (maybe in addition) a denylist. But can switch to `allowed_mcp_servers`.

gt-oai: Renamed to `mcp_server_requirements` (didn't want `mcp_servers` because it has a different shape to the config.toml value.)

codex-rs/core/src/config_loader/config_requirements.rs
gt-oai: - Renamed `McpServerIdentityRequirement` to `McpServerIdentity`
- Nested it underneath `McpServerRequirement`

In the future, we can add enabled tools as children of `McpServerRequirement`.

codex-rs/core/src/config/constraint.rs
```diff
@@ -54,13 +56,30 @@ impl<T: Send + Sync> Constrained<T> {
         Ok(Self {
             value: initial_value,
             validator,
+            normalizer: None,
+        })
+    }
+
+    pub fn normalized(
```
bolinfest, line 66: docstring?

codex-rs/core/src/config/constraint.rs
```diff
@@ -37,11 +37,13 @@ impl From<ConstraintError> for std::io::Error {
 }
 
 type ConstraintValidator<T> = dyn Fn(&T) -> ConstraintResult<()> + Send + Sync;
+type ConstraintNormalizer<T> = dyn Fn(T) -> T + Send + Sync;
```
bolinfest, line 42: Docstring? I am pretty sure I understand what "normalize" means in this context, but I'd rather be 100% sure.

codex-rs/core/src/config/mod.rs
```diff
@@ -510,6 +508,55 @@ fn deserialize_config_toml_with_base(
...
+        if !allowed {
+            server.enabled = false;
+        }
+    }
+}
+
+fn constrain_mcp_servers(
+    mcp_servers: HashMap<String, McpServerConfig>,
+    mcp_requirements: Option<&BTreeMap<String, McpServerRequirement>>,
+) -> ConstraintResult<Constrained<HashMap<String, McpServerConfig>>> {
```
bolinfest, line 537: Should this return `Constrained::allow_any()` in some cases, though?

codex-rs/core/src/config_loader/config_requirements.rs
bolinfest: Personally, I feel like `requirements` is implied by being in `requirements.toml`.

I worry that we are making things harder for ourselves by trying to be distinct from `config.toml` in all cases. I think anyone who has the rights to touch this file should be aware of the docs at https://developers.openai.com/codex/security#example-requirementstoml.

bolinfest (approved):
I don't believe this would prevent the user from running `codex mcp add`, though I suppose it's fine for them to add anything they want to their `config.toml`: it will just blow up at runtime if they declare an out-of-bounds MCP server in there?

codex-rs/core/src/config/mod.rs
gt-oai: `Constrained::normalized` is effectively allow_any when the requirements are empty. But I've added a shortcircuit for that case now.

PR #9102, opened 2026-01-12
Add model client sessions

pakrym-oai:
Maintain a long-running session.


codex-rs/core/src/client.rs
```diff
@@ -70,6 +70,16 @@ pub struct ModelClient {
     session_source: SessionSource,
 }
 
+#[derive(Debug, Clone)]
+pub struct ModelClient {
+    state: Arc<ModelClientState>,
+}
+
+#[derive(Debug, Clone)]
+pub struct ModelClientSession {
```
aibrahim-oai, line 80: Why do we have identical structs `ModelClientSession` and `ModelClient`?

pakrym-oai: ModelClientSession will have more per-session state in the future (like websocket) connection in addition to the more generic state of the client.

aibrahim-oai (approved):

PR #9105, opened 2026-01-12
feat: add auto refresh on thread listeners

jif-oai:
This PR is in the scope of multi-agent work. 

An agent (=thread) can now spawn other agents. Those other agents are not attached to any clients. We need a way to make sure that the clients are aware of the new threads to look at (for approval for example). This PR adds a channel to the `ThreadManager` that pushes the ID of those newly created agents such that the client (here the app-server) can also subscribe to those ones.


codex-rs/app-server/tests/suite/codex_message_processor_flow.rs
```diff
@@ -252,36 +252,58 @@ async fn test_send_user_turn_changes_approval_policy_behavior() -> Result<()> {
...
-        params
-    );
-
-    // Approve so the first turn can complete
-    mcp.send_response(
-        request_id,
-        serde_json::json!({ "decision": codex_core::protocol::ReviewDecision::Approved }),
-    )
-    .await?;
+    match request {
```
celia-oai: I'm a little skeptical of the non-deterministic behavior of this test... maybe we should just remove attaching v1 listener?

celia-oai (approved):

jif-oai:
@codex review

codex-rs/core/src/agent/control.rs
```diff
@@ -29,21 +27,18 @@ impl AgentControl {
...
     ) -> CodexResult<ThreadId> {
         let state = self.upgrade()?;
         let new_thread = state.spawn_new_thread(config, self.clone()).await?;
 
-        if headless {
-            spawn_headless_drain(Arc::clone(&new_thread.thread));
-        }
+        // Notify a new thread has been created. This notification will be processed by clients
+        // to subscribe or drain this newly created thread.
+        // TODO(jif) add helper for drain
```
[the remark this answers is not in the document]
jif-oai: This is on purpose

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -1669,6 +1676,31 @@ impl CodexMessageProcessor {
         self.outgoing.send_response(request_id, response).await;
     }
 
+    pub(crate) fn thread_created_receiver(&self) -> broadcast::Receiver<ThreadId> {
+        self.thread_manager.subscribe_thread_created()
+    }
+
+    /// Ensure we have a listener for thread_id
+    pub(crate) async fn ensure_thread_listener(&mut self, thread_id: ThreadId) {
```
gt-oai: It's a bit of a smell that this is called "ensure" if all it does is warn on error. Perhaps we could better comment under what circumstances `attach_conversation_listener` fails?

gt-oai (approved):

codex-rs/app-server/src/codex_message_processor.rs
jif-oai: In theory in should never fail in this scenario but will do

PR #9106, opened 2026-01-12
upgrade runners in rust-ci.yml to use the larger runners

willwang-openai:
Upgrades runners in rust-ci.yaml to larger runners

ubuntu-24.04 (x64 and arm64) -> custom 16 core ubuntu 24.04 runners
macos-14 -> mac0s-15-xlarge
windows (x64 and arm64) -> custom 16 core windows runners

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -106,14 +106,12 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - runner: macos-14
+          - runner: macos-15-xlarge
             target: aarch64-apple-darwin
             profile: dev
-            runs_on: macos-14-xlarge
-          - runner: macos-14
+          - runner: macos-15-xlarge
             target: x86_64-apple-darwin
             profile: dev
-            runs_on: macos-14-xlarge
           - runner: ubuntu-24.04
             target: x86_64-unknown-linux-musl
             profile: dev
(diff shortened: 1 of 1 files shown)


.github/workflows/rust-ci.yml
```diff
@@ -372,21 +397,6 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      # We have been running out of space when running this job on Linux for
```
bolinfest, line 375: I think this is still a good thing to do.

.github/workflows/rust-ci.yml
```diff
@@ -109,21 +109,35 @@ jobs:
           - runner: macos-14
             target: aarch64-apple-darwin
             profile: dev
+            runs_on: macos-14-xlarge
```
bolinfest: Very nice! Can you make the analogous change to `bazel.yml`, as well?

willwang-openai: yes! next pr

.github/workflows/rust-ci.yml
willwang-openai: Why do this if we dont need it? It will slow down the build

bolinfest:
@willwang-openai can you please update the PR body, as well?

.github/workflows/rust-ci.yml
bolinfest: Hmm, this step is taking longer than I thought. OK, we'll live without it for now. Hopefully we don't have to bring it back anytime soon...

.github/workflows/rust-ci.yml
```diff
@@ -138,9 +152,13 @@ jobs:
           - runner: macos-14
```
bolinfest: Should we just switch this to `macos-15-xlarge` instead? Do we need the `runs_on` in that case?

willwang-openai:

bolinfest (approved):

Issue #9104, opened 2026-01-12
Codex crashes on startup for invalid enum values in config.toml


yash27-lab:
### What version of Codex is running?

codex-cli 0.42.0

### What subscription do you have?

ChatGPT Plus

### What platform is your computer?

Macos

### What issue are you seeing?

Codex CLI fails to start if config.toml contains an invalid enum value
(e.g. model_reasoning_effort = "xhigh").

This is a recoverable user input error and should not cause a hard startup failure.

Reproduction:
1. Set model_reasoning_effort to an invalid value in ~/.codex/config.toml
2. Run `codex`
3. Codex exits with an error and does not start

Current behavior:
- Codex aborts startup with a config parsing error

Expected behavior:
- Print a clear error explaining the invalid value
- Show allowed values
- Suggest the closest valid option
- Fall back to a safe default or provide a guided fix

This would improve robustness and UX without changing defaults.

### What steps can reproduce the bug?

1. Edit ~/.codex/config.toml
2. Set model_reasoning_effort to an invalid value (e.g. "xhigh")
3. Run `codex`
4. Codex exits with a config parsing error and does not start

### What is the expected behavior?

Codex should not hard-fail on startup for an invalid enum value.
It should explain the invalid value, list allowed values, and either
fall back to a safe default or provide a guided fix.

### Additional information

This failure is user-recoverable and similar errors have occurred when
config values changed between versions.
Improving error handling here would prevent startup lockouts.

etraut-openai:
There is no crash on startup in this case. The CLI is designed to report an error and quit if there is an invalid config key. Our philosophy is that it's better to report an error in this case rather than proceed with (potentially) incorrect configuration values.

yash27-lab:
Thanks for the clarification — that makes sense.

My intent was to improve DX by making the invalid-config failure more
actionable (e.g., clearly listing allowed enum values and a suggested fix),
not to proceed with potentially incorrect defaults.

dostiharise:
When open my codex cli config with `model_reasoning_effort` to `xhigh` it fails to start. 

According to [Open AI docs](https://developers.openai.com/codex/config-reference/) the value `xhigh` is a valid config parameter.

```
$ cat ~/.codex/config.toml

model = "gpt-5.2-codex"
model_reasoning_effort = "xhigh"
```

etraut-openai:
@dostiharise, this bug report is not related to the issue you're seeing. If you'd like to report a problem, please open a new issue and provide details and repro steps.

PR #9108, opened 2026-01-12
Avoid crash on invalid enum values in config.toml

etraut-openai:
This fixes bug #9104 which results in a panic when config.toml is parsed and one or more enum values are incorrect. This change adds a soft failure that falls back to default values.

I've tried implementing this several different ways. The current implementation is the one that I think is most maintainable. The challenge is that we want to enforce enum values in the write case but not in the read case. That means we need to be "lenient" sometimes (falling back to defaults) but fail in others. The serde parser relies on static attributes, so it doesn't generally support different (dynamic) behaviors. That means either 1) defining duplicate structs that differ only in attributes (yuck!), 2) writing custom deserialization code for every enum field (yuck!), or 3) writing a generalized "lenient" deserializer. I went with option 3.

Ideally, we should report errors like this to the user (for all code paths that parse the config — including the TUI, exec, app server, MCP server, etc.), but that's a much larger change. For this PR, I'm sticking with the current strategy of silently falling back to defaults when there's a bad config value.


etraut-openai:
@codex review

yuvrajangadsingh:
Hi Eric, I was also working on #9104 and took a different (complementary) approach in #9163: showing the config file path in error messages so users know *where* to fix the problem.

Your lenient deserializer elegantly handles invalid enum values by falling back to defaults. However, for **non-enum config errors** (wrong types, syntax errors, etc.), the error still doesn't indicate which config file contains the problem.

In my approach, I added an `enhance_config_error()` function that extracts the field name from the error and looks up its source file from `ConfigLayerStack::origins()`:

```
unknown variant `xhigh_invalid`, expected one of `none`, `minimal`, `low`, `medium`, `high`, `xhigh`
in `model_reasoning_effort`
  in /Users/foo/.codex/config.toml
```

Would it make sense to add this file path context to your `deserialize_config_toml_lenient()` for the non-enum error path? Happy to contribute that enhancement if useful.

etraut-openai:
@yuvrajangadsingh, I don't think we need to add a path to the config file. Codex doesn't support multiple config files, so there's no need to differentiate.

yuvrajangadsingh:
Just a small clarification - Codex does support multiple config sources via `ConfigLayerStack`:

- User config (`~/.codex/config.toml`)
- Project config (`.codex/config.toml` in project directory)
- System config
- MDM config
- CLI flags

The `ConfigLayerStack::origins()` method tracks which source each setting came from, which is why my approach could look up the source file for error messages.

That said, I understand your decision to keep things simple. Thanks for the clarification on the design philosophy!

PR #9112, opened 2026-01-12
feat: add support for read-only bind mounts in the linux sandbox

viyatb-oai:
### Motivation

- Landlock alone cannot prevent writes to sensitive in-repo files like `.git/` when the repo root is writable, so explicit mount restrictions are required for those paths.
- The sandbox must set up any mounts before calling Landlock so Landlock can still be applied afterwards and the two mechanisms compose correctly.

### Description

- Add a new `linux-sandbox` helper `apply_read_only_mounts` in `linux-sandbox/src/mounts.rs` that: unshares namespaces, maps uids/gids when required, makes mounts private, bind-mounts targets, and remounts them read-only.
- Wire the mount step into the sandbox flow by calling `apply_read_only_mounts(...)` before network/seccomp and before applying Landlock rules in `linux-sandbox/src/landlock.rs`.

### DIFF
--- a/codex-rs/linux-sandbox/src/mounts.rs
+++ b/codex-rs/linux-sandbox/src/mounts.rs
@@ -8,6 +8,11 @@ use codex_core::protocol::SandboxPolicy;
 use codex_core::protocol::WritableRoot;
 use codex_utils_absolute_path::AbsolutePathBuf;
 
+/// Apply read-only bind mounts for protected subpaths before Landlock.
+///
+/// This unshares mount namespaces (and user namespaces for non-root) so the
+/// read-only remounts do not affect the host, then bind-mounts each protected
+/// target onto itself and remounts it read-only.
 pub(crate) fn apply_read_only_mounts(sandbox_policy: &SandboxPolicy, cwd: &Path) -> Result<()> {
     let writable_roots = sandbox_policy.get_writable_roots_with_cwd(cwd);
     let mount_targets = collect_read_only_mount_targets(&writable_roots)?;
@@ -15,6 +20,8 @@ pub(crate) fn apply_read_only_mounts(sandbox_policy: &SandboxPolicy, cwd: &Path)
         return Ok(());
     }
 
+    // Root can unshare the mount namespace directly; non-root needs a user
+    // namespace to gain capabilities for remounting.
     if is_running_as_root() {
         unshare_mount_namespace()?;
     } else {
@@ -24,18 +31,22 @@ pub(crate) fn apply_read_only_mounts(sandbox_policy: &SandboxPolicy, cwd: &Path)
     make_mounts_private()?;
 
     for target in mount_targets {
+        // Bind and remount read-only works for both files and directories.
         bind_mount_read_only(target.as_path())?;
     }
 
     Ok(())
 }
 
+/// Collect read-only mount targets, resolving worktree `.git` pointer files.
 fn collect_read_only_mount_targets(
     writable_roots: &[WritableRoot],
 ) -> Result<Vec<AbsolutePathBuf>> {
     let mut targets = Vec::new();
     for writable_root in writable_roots {
         for ro_subpath in &writable_root.read_only_subpaths {
+            // The policy expects these paths to exist; surface actionable errors
+            // rather than silently skipping protections.
             if !ro_subpath.as_path().exists() {
                 return Err(CodexErr::UnsupportedOperation(format!(
                     "Sandbox expected to protect {path}, but it does not exist. Ensure the repository contains this path or create it before running Codex.",
@@ -43,6 +54,8 @@ fn collect_read_only_mount_targets(
                 )));
             }
             targets.push(ro_subpath.clone());
+            // Worktrees and submodules store `.git` as a pointer file; add the
+            // referenced gitdir as an extra read-only target.
             if is_git_pointer_file(ro_subpath) {
                 let gitdir = resolve_gitdir_from_file(ro_subpath)?;
                 if !targets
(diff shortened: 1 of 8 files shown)


codex-rs/linux-sandbox/src/mounts.rs
```diff
@@ -0,0 +1,153 @@
...
+fn unshare_user_and_mount_namespaces() -> Result<()> {
+    let result = unsafe { libc::unshare(libc::CLONE_NEWUSER | libc::CLONE_NEWNS) };
+    if result != 0 {
+        return Err(std::io::Error::last_os_error().into());
+    }
+    Ok(())
+}
+
+fn is_running_as_root() -> bool {
+    unsafe { libc::geteuid() == 0 }
```
bolinfest, line 137: Because this is `euid`, this also covers `sudo`, is that right?

codex-rs/linux-sandbox/src/mounts.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+    Ok(targets)
+}
+
+fn ensure_gitdir_is_directory(root: &AbsolutePathBuf) -> Result<()> {
+    #[allow(clippy::expect_used)]
+    let dot_git = root.join(".git").expect(".git is a valid relative path");
+    if dot_git.as_path().is_file() {
+        return Err(CodexErr::UnsupportedOperation(format!(
+            "Sandbox protection requires .git to be a directory, but {path} is a file. If this is a worktree, protect the real gitdir (from `git rev-parse --git-dir`) and consider making the .git pointer file read-only.",
```
bolinfest: Does `bwrap` also have this limitation? This suggests it can use its `--ro-bind` option with a file?

https://github.com/containers/bubblewrap/blob/d6180f25b164c708b8b0a0d86d6a9642f30cd9a9/demos/bubblewrap-shell.sh#L18

bolinfest (commented):
Can we maybe separate the `.git` -> `.git/config && .git/hooks` change from the bind mounts stuff and get that in first since I think it's uncontroversial?

bolinfest:
@viyatb-oai Also, the `suite::seatbelt::if_git_repo_is_writable_root_then_dot_git_folder_is_read_only` test failure seems legit.

codex-rs/linux-sandbox/src/mounts.rs
viyatb-oai: yes!

codex-rs/linux-sandbox/src/mounts.rs
```diff
@@ -0,0 +1,255 @@
...
+            if !ro_subpath.as_path().exists() {
+                return Err(CodexErr::UnsupportedOperation(format!(
+                    "Sandbox expected to protect {path}, but it does not exist. Ensure the repository contains this path or create it before running Codex.",
+                    path = ro_subpath.as_path().display()
+                )));
+            }
+            targets.push(ro_subpath.clone());
+            // Worktrees and submodules store `.git` as a pointer file; add the
+            // referenced gitdir as an extra read-only target.
+            if is_git_pointer_file(ro_subpath) {
```
bolinfest, line 65: Should this `is_git_pointer_file()` logic be shared with https://github.com/openai/codex/blob/31d9b6f4d2033438d3375d4c3fbadcd5c08660d9/codex-rs/protocol/src/protocol.rs#L509-L515 so that we do the same thing on macOS?

codex-rs/linux-sandbox/src/mounts.rs
```diff
@@ -0,0 +1,255 @@
...
+                let gitdir = resolve_gitdir_from_file(ro_subpath)?;
```
bolinfest, line 66: Does it matter if `gitdir` is under another writable root or not? I believe this would matter in Seatbelt because of how `(require-all)` works there?

codex-rs/protocol/src/protocol.rs
```diff
@@ -510,7 +500,10 @@ impl SandboxPolicy {
                         let top_level_git = writable_root
                             .join(".git")
                             .expect(".git is a valid relative path");
-                        if top_level_git.as_path().is_dir() {
+                        // This applies to typical repos (directory .git), worktrees/submodules
+                        // (file .git with gitdir pointer), and bare repos when the gitdir is the
+                        // writable root itself.
+                        if top_level_git.as_path().is_dir() || top_level_git.as_path().is_file() {
```
bolinfest: We should update our tests to verify that the `.git` file is not writable on macOS?

codex-rs/linux-sandbox/tests/suite/landlock.rs
```diff
@@ -127,6 +174,82 @@ async fn test_writable_root() {
     .await;
 }
 
+#[tokio::test]
+async fn test_git_dir_write_blocked() {
+    let tmpdir = tempfile::tempdir().unwrap();
+    let repo_root = tmpdir.path();
+    Command::new("git")
```
bolinfest, line 181: Since this function is `async`, should theoretically prefer `tokio::process::Command`, but it's not a huge deal.

codex-rs/linux-sandbox/tests/suite/landlock.rs
```diff
@@ -127,6 +174,82 @@ async fn test_writable_root() {
     .await;
 }
 
+#[tokio::test]
```
bolinfest, line 177: Other tests to consider: is `mv .git` blocked? Should we also ensure `.codex` cannot be written?

bolinfest (commented):
This is looking really good!

codex-rs/linux-sandbox/tests/suite/landlock.rs
viyatb-oai: done

codex-rs/linux-sandbox/src/mounts.rs
viyatb-oai: done

bolinfest (approved):
Amazing work here!

PR #9113, opened 2026-01-12
Validate model_reasoning_effort and avoid startup crash

### LINKED ISSUE #9104: Codex crashes on startup for invalid enum values in config.toml

yash27-lab:
Fixes a startup crash when config.toml contains an invalid
model_reasoning_effort value.

This change:
- Validates model_reasoning_effort early
- Emits a user-friendly error message
- Suggests valid enum values
- Adds a regression test

Fixes #9104

### DIFF
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -80,6 +80,42 @@ pub use codex_git::GhostSnapshotConfig;
 pub(crate) const PROJECT_DOC_MAX_BYTES: usize = 32 * 1024; // 32 KiB
 
 pub const CONFIG_TOML_FILE: &str = "config.toml";
+fn validate_model_reasoning_effort_toml(root: &TomlValue) -> std::io::Result<()> {
+    let Some(raw) = root.get("model_reasoning_effort") else {
+        return Ok(());
+    };
+
+    let Some(s) = raw.as_str() else {
+        return Ok(());
+    };
+
+    let s_lc = s.to_ascii_lowercase();
+    let allowed = ["none", "minimal", "low", "medium", "high", "xhigh"];
+
+    if allowed.contains(&s_lc.as_str()) {
+        return Ok(());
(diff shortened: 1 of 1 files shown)


yash27-lab:

yash27-lab:
recheck

etraut-openai:
Thanks for the contribution, but I don't think this is the correct fix for the problem. We need a fix that generalizes beyond just this one key. The problem affects all keys that have enum values. I'm working on a [separate PR](https://github.com/openai/codex/pull/9108) that provides a more general solution.

yash27-lab:
Thanks for the clarification, Eric — that makes sense.

Appreciate you taking this further with a generalized solution. Happy to
learn from the approach in your PR and help iterate if needed.

Thanks again for the review.

PR #9116, opened 2026-01-12
Persist text element ranges and attached images across history/resume

charley-oai:
**Summary**
- Backtrack selection now rehydrates `text_elements` and `local_image_paths` from the chosen user history cell so Esc‑Esc history edits preserve image placeholders and attachments.
- Composer prefill uses the preserved elements/attachments in both `tui` and `tui2`.
- Extended backtrack selection tests to cover image placeholder elements and local image paths.

**Changes**
- `tui/src/app_backtrack.rs`: Backtrack selection now carries text elements + local image paths; composer prefill uses them (removes TODO).
- `tui2/src/app_backtrack.rs`: Same as above.
- `tui/src/app.rs`: Updated backtrack test to assert restored elements/paths.
- `tui2/src/app.rs`: Same test updates.

### The original scope of this PR (threading text elements and image attachments through the codex harness thoroughly/persistently) was broken into the following PRs other than this one:

The diff of this PR was reduced by changing types in a starter PR: https://github.com/openai/codex/pull/9235

Then text element metadata was added to protocol, app server, and core in this PR: https://github.com/openai/codex/pull/9331

Then the end-to-end flow was completed by wiring TUI/TUI2 input, history, and restore behavior in https://github.com/openai/codex/pull/9393

Prompt expansion was supported in this PR: https://github.com/openai/codex/pull/9518

TextElement optional placeholder field was protected in https://github.com/openai/codex/pull/9545

### DIFF
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -1876,6 +1876,7 @@ mod tests {
     use codex_core::protocol::SessionSource;
     use codex_otel::OtelManager;
     use codex_protocol::ThreadId;
+    use codex_protocol::user_input::TextElement;
     use insta::assert_snapshot;
     use pretty_assertions::assert_eq;
     use ratatui::prelude::Line;
@@ -2168,11 +2169,14 @@ mod tests {
     async fn backtrack_selection_with_duplicate_history_targets_unique_turn() {
         let (mut app, _app_event_rx, mut op_rx) = make_test_app_with_channels().await;
 
-        let user_cell = |text: &str| -> Arc<dyn HistoryCell> {
+        let user_cell = |text: &str,
+                         text_elements: Vec<TextElement>,
+                         local_image_paths: Vec<PathBuf>|
+         -> Arc<dyn HistoryCell> {
             Arc::new(UserHistoryCell {
                 message: text.to_string(),
-                text_elements: Vec::new(),
-                local_image_paths: Vec::new(),
+                text_elements,
+                local_image_paths,
             }) as Arc<dyn HistoryCell>
         };
         let agent_cell = |text: &str| -> Arc<dyn HistoryCell> {
@@ -2210,18 +2214,28 @@ mod tests {
             )) as Arc<dyn HistoryCell>
         };
 
+        let placeholder = "[Image #1]";
+        let edited_text = format!("follow-up (edited) {placeholder}");
+        let edited_range = edited_text.len().saturating_sub(placeholder.len())..edited_text.len();
+        let edited_text_elements = vec![TextElement::new(edited_range.into(), None)];
+        let edited_local_image_paths = vec![PathBuf::from("/tmp/fake-image.png")];
+
         // Simulate a transcript with duplicated history (e.g., from prior backtracks)
         // and an edited turn appended after a session header boundary.
         app.transcript_cells = vec![
             make_header(true),
-            user_cell("first question"),
+            user_cell("first question", Vec::new(), Vec::new()),
             agent_cell("answer first"),
-            user_cell("follow-up"),
+            user_cell("follow-up", Vec::new(), Vec::new()),
             agent_cell("answer follow-up"),
             make_header(false),
-            user_cell("first question"),
+            user_cell("first question", Vec::new(), Vec::new()),
             agent_cell("answer first"),
-            user_cell("follow-up (edited)"),
+            user_cell(
+                &edited_text,
+                edited_text_elements.clone(),
+                edited_local_image_paths.clone(),
+            ),
             agent_cell("answer edited"),
         ];
 
@@ -2254,7 +2268,9 @@ mod tests {
             .confirm_backtrack_from_main()
             .expect("backtrack selection");
         assert_eq!(selection.nth_user_message, 1);
-        assert_eq!(selection.prefill, "follow-up (edited)");
+        assert_eq!(selection.prefill, edited_text);
+        assert_eq!(selection.text_elements, edited_text_elements);
+        assert_eq!(selection.local_image_paths, edited_local_image_paths);
 
         app.apply_backtrack_rollback(selection);
 
--- a/codex-rs/tui/src/bottom_pane/chat_composer.rs
+++ b/codex-rs/tui/src/bottom_pane/chat_composer.rs
@@ -212,6 +212,7 @@ pub(crate) struct ChatComposer {
     pending_pastes: Vec<(String, String)>,
     large_paste_counters: HashMap<usize, usize>,
     has_focus: bool,
+    /// Invariant: attached images are labeled `[Image #1]..[Image #N]` in vec order.
     attached_images: Vec<AttachedImage>,
     placeholder_text: String,
     is_task_running: bool,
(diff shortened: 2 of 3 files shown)


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1545,16 +1546,34 @@ pub struct TurnInterruptResponse {}
...
 #[ts(export_to = "v2/")]
 pub enum UserInput {
-    Text { text: String },
-    Image { url: String },
-    LocalImage { path: PathBuf },
-    Skill { name: String, path: PathBuf },
+    Text {
+        text: String,
+        /// UI-defined spans within `text` used to render or persist special elements.
+        #[serde(default, skip_serializing_if = "Vec::is_empty")]
```
owenlin0: nit: avoid `skip_serializing_if` annotations on v2.rs, we don't want to omit fields if null generally

There are a couple places in the V2 API that has it unfortunately, but we should avoid adding more if possible

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1545,16 +1546,34 @@ pub struct TurnInterruptResponse {}
...
+        text_elements: Vec<TextElement>,
```
owenlin0, line 1715: instead of exposing the type from `codex_protocol` aka core directly, let's redefine the struct in this file so we can serialize it as camelCase for consistency

e.g. `use codex_protocol::user_input::TextElement as CoreTextElement;` and translate to/from it as necessary

aibrahim-oai:
@codex review this

codex-rs/app-server-protocol/src/protocol/v1.rs
```diff
@@ -444,9 +446,74 @@ pub struct RemoveConversationListenerParams {
...
+    },
+    LocalImage {
+        path: PathBuf,
+    },
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(rename = "ByteRange")]
+pub struct V1ByteRange {
```
aibrahim-oai, line 467: why are we prefixing with v1?

codex-rs/core/src/codex.rs
```diff
@@ -2560,9 +2562,19 @@ pub(crate) async fn run_turn(
             .await;
     }
 
+    let user_message_item = UserMessageItem::new(&input);
     let initial_input_for_turn: ResponseInputItem = ResponseInputItem::from(input);
     let response_item: ResponseItem = initial_input_for_turn.clone().into();
-    sess.record_response_item_and_emit_turn_item(turn_context.as_ref(), response_item)
```
aibrahim-oai: why are we doing that?

codex-rs/protocol/src/items.rs
```diff
@@ -83,6 +85,30 @@ impl UserMessageItem {
             .join("")
     }
 
+    pub fn text_elements(&self) -> Vec<TextElement> {
```
aibrahim-oai, line 88: why do we need to do that?

aibrahim-oai (commented):
I feel the PR is still pretty big and has a lot of moving parts. 

I can see it broken down to support in app-server, core, and maybe two-stacked for tui one for general support of the text element and one for the backtracking changes

That way it be easier to track those moving parts and remove unnecessary changes. Also, will enable us to think carefully about how to test it.

codex-rs/core/src/codex.rs
charley-oai: The comment below explains:

```
 // Persist the user message to history, but emit the turn item from `UserInput` so
    // UI-only `text_elements` are preserved. `ResponseItem::Message` does not carry
    // those spans, and `record_response_item_and_emit_turn_item` would drop them.
```

codex-rs/app-server-protocol/src/protocol/v1.rs
charley-oai: Because app-server-protocol re‑exports both v1::* and v2::* at the crate root. If v1 defines ByteRange and v2 also defines ByteRange, glob re‑exporting both
creates ambiguous names and warnings (and can break downstream use).

Prefixing v1’s internal type names (V1ByteRange, V1TextElement) removes the clash while still keeping the TypeScript name stable via #[ts(rename = "ByteRange")] /
#[ts(rename = "TextElement")]. So:

- Rust: unique names, no ambiguity.
- TS: still ByteRange / TextElement for v1.

Owen mentioned we don't want to share types from Core and that we should redefine separately in v1 and v2 (https://github.com/openai/codex/pull/9116#discussion_r2691472641). Is this the right way to do that?

aibrahim-oai: Makes sense. Let's double check with @owenlin0 or @celia-oai. They are the DRIs for that

owenlin0: FWIW I don't think it's worth adding anything to the v1 protocol at this point

codex-rs/protocol/src/items.rs
charley-oai: A message can contain multiple `UserInput::Text` chunks, and each `text_elements` range is relative to its own chunk, not the full rendered
message. When we render the user message (or store it as a single string), we concatenate those chunks, so we must offset each element’s byte range by the
cumulative length of prior chunks. Otherwise any element after the first chunk would point to the wrong bytes and render in the wrong place.

Added a comment to clarify this in the code.

codex-rs/tui/src/app.rs
```diff
@@ -340,11 +340,6 @@ pub(crate) struct App {
 
     // Esc-backtracking state grouped
     pub(crate) backtrack: crate::app_backtrack::BacktrackState,
-    /// When set, the next draw re-renders the transcript into terminal scrollback once.
-    ///
-    /// This is used after a confirmed thread rollback to ensure scrollback reflects the trimmed
-    /// transcript cells.
-    pub(crate) backtrack_render_pending: bool,
```
charley-oai, line 360: We remove backtrack_render_pending because it was only needed for the rollback-based flow. In the old flow, we’d send ThreadRollback, then wait for
ThreadRolledBack, then set a flag so the next draw would re-render scrollback.

In the new fork-based flow:

- We wait for ConversationPathResponseEvent,
- Fork + swap threads immediately,
- Trim transcript and call render_transcript_once(tui) right there.

So there’s no deferred “render on next draw” step left. The flag becomes dead, and keeping it would just add complexity without any effect.

charley-oai:
@codex review this

charley-oai:
@codex review this

codex-rs/tui/src/bottom_pane/chat_composer.rs
```diff
@@ -525,6 +527,8 @@ impl ChatComposer {
             self.textarea.insert_str(&text[idx..]);
         }
 
+        // Keep image placeholders normalized to [Image #1].. in attachment order.
+        self.relabel_attached_images_and_update_placeholders();
```
charley-oai, line 544: Codex pointed out that the external edits path didn't relabel attached images, which could break the image labeling assumptions made by this PR

charley-oai:
@codex review this

aibrahim-oai (approved):

PR #9118, opened 2026-01-12
feat(app-server): add an --analytics-default-enabled flag

owenlin0:
Add a new `codex app-server --analytics-default-enabled` CLI flag that controls whether analytics are enabled by default.

Analytics are disabled by default for app-server. Users have to explicitly opt in
via the `analytics` section in the config.toml file.

However, for first-party use cases like the VSCode IDE extension, we default analytics
to be enabled by default by setting this flag. Users can still opt out by setting this
in their config.toml:

```toml
[analytics]
enabled = false
```

See https://developers.openai.com/codex/config-advanced/#metrics for more details.


owenlin0:
@codex review

codex-rs/cli/src/main.rs
```diff
@@ -910,6 +929,14 @@ mod tests {
         finalize_fork_interactive(interactive, root_overrides, session_id, last, all, fork_cli)
     }
 
+    fn app_server_from_args(args: &[&str]) -> AppServerCommand {
+        let cli = MultitoolCli::try_parse_from(args).expect("parse");
+        let Subcommand::AppServer(app_server) = cli.subcommand.expect("app-server present") else {
```
jif-oai, line 934: This pattern is pretty fun but good for me

jif-oai (approved):

PR #9121, opened 2026-01-13
test(tui): add deterministic paste-burst tests

joshka-oai:
Replace the old timing-dependent non-ASCII paste test with deterministic coverage by forcing an active `PasteBurst` and asserting the exact flush payload.

Add focused unit tests for `PasteBurst` transitions, and add short "Behavior:" rustdoc notes on chat composer tests to make the state machine contracts explicit.

### DIFF
--- a/codex-rs/tui/src/bottom_pane/chat_composer.rs
+++ b/codex-rs/tui/src/bottom_pane/chat_composer.rs
@@ -2437,6 +2437,8 @@ mod tests {
         );
     }
 
+    /// Behavior: `?` toggles the shortcut overlay only when the composer is otherwise empty. After
+    /// any typing has occurred, `?` should be inserted as a literal character.
     #[test]
     fn question_mark_only_toggles_on_first_char() {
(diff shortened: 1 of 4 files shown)


codex-rs/tui/src/bottom_pane/chat_composer.rs
```diff
@@ -2663,55 +2672,126 @@ mod tests {
...
-日月盈昃 辰宿列张
-寒来暑往 秋收冬藏
-
-你好世界 编码测试
-汉字处理 UTF-8
-终端显示 正确无误
-
-风吹竹林 月照大江
-白云千载 青山依旧
-程序员 与 Unicode 同行",
```
joshka-oai, line 2776: Note - codex removed this initially. It mainly had an effect because it was long enough to cause the timing to be right. I added it back in as an explicit test below, mainly to keep the pretty random poem.

dylan-hurd-oai (approved):
Thanks for adding!

PR #9122, opened 2026-01-13
Fix queued messages during /review

charley-oai:
Sending a message during /review interrupts the review, whereas during normal operation, sending a message while the agent is running will queue the message. This is unexpected behavior, and since /review usually takes a while, it takes away a potentially useful operation.

Summary
- Treat review mode as an active task for message queuing so inputs don’t inject into the running review turn.
- Prevents user submissions from rendering immediately in the transcript while the review continues streaming.
- Keeps review UX consistent with normal “task running” behavior and avoids accidental interrupt/replacement.

Notes
- This change only affects UI queuing logic; core review flow and task lifecycle remain unchanged.

### DIFF
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -2200,6 +2200,7 @@ impl ChatWidget {
         if self.pre_review_token_info.is_none() {
             self.pre_review_token_info = Some(self.token_info.clone());
         }
+        // Avoid toggling running state for replayed history events on resume.
         if !from_replay && !self.bottom_pane.is_task_running() {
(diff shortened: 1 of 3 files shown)


codex-rs/tui/src/chatwidget.rs
```diff
@@ -1973,7 +1973,7 @@ impl ChatWidget {
     }
 
     fn queue_user_message(&mut self, user_message: UserMessage) {
-        if self.bottom_pane.is_task_running() {
+        if self.bottom_pane.is_task_running() || self.is_review_mode {
```
aibrahim-oai, line 1976: let's add a snapshot

aibrahim-oai (approved):

PR #9126, opened 2026-01-13
fix: drop session span at end of the session

apanasenko-oai:


codex-rs/otel/src/lib.rs
```diff
@@ -42,7 +44,7 @@ pub struct OtelEventMetadata {
 #[derive(Debug, Clone)]
 pub struct OtelManager {
     pub(crate) metadata: OtelEventMetadata,
-    pub(crate) session_span: Span,
+    pub(crate) session_span: Arc<Mutex<Option<Span>>>,
```
jif-oai: This looks pretty bad to me. The `Mutex` will block the runner and in general this kind of object look like a smell. 
Could you at least use a `RwLock`? (and I think we can drop the `Option` as `Span::none()` should be available

jif-oai (approved):
Approved to unblock but we should resolve this `Arc<Mutex<Option<Span>>>` before merging

codex-rs/otel/src/lib.rs
apanasenko-oai: remove it at all to simplify whole process.

PR #9128, opened 2026-01-13
Websocket append support

pakrym-oai:
Support an incremental append request in websocket transport.

### DIFF
--- a/codex-rs/core/src/client.rs
+++ b/codex-rs/core/src/client.rs
@@ -324,7 +324,9 @@ impl ModelClientSession {
         }
     }
 
-    fn websocket_append_items(&self, input_items: &[ResponseItem]) -> Option<Vec<ResponseItem>> {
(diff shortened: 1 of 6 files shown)


codex-rs/core/src/client.rs
```diff
@@ -320,59 +324,75 @@ impl ModelClientSession {
         }
     }
 
-    fn build_responses_websocket_request(
+    fn websocket_append_items(&self, input_items: &[ResponseItem]) -> Option<Vec<ResponseItem>> {
+        let previous_len = self.websocket_last_items.len();
+        let can_append = previous_len > 0
+            && input_items.starts_with(&self.websocket_last_items)
+            && previous_len < input_items.len();
```
aibrahim-oai, line 334: This feels like a not clear condition. what are we trying to do?

Also what is the use of storing `websocket_last_items`?

pakrym-oai: A comment is in order!
It's checking whether the next request is an incremental version of the previous one.

aibrahim-oai (approved):

PR #9130, opened 2026-01-13
Fresh tooltips

mzeng-openai:
Fresh tooltips

### DIFF
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -1422,7 +1422,7 @@ impl ChatWidget {
         let mut config = config;
         config.model = Some(model.clone());
         let mut rng = rand::rng();
-        let placeholder = EXAMPLE_PROMPTS[rng.random_range(0..EXAMPLE_PROMPTS.len())].to_string();
+        let placeholder = PLACEHOLDERS[rng.random_range(0..PLACEHOLDERS.len())].to_string();
         let codex_op_tx = spawn_agent(config.clone(), app_event_tx.clone(), thread_manager);
 
         let mut widget = Self {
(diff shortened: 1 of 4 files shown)


codex-rs/tui/src/chatwidget.rs
```diff
@@ -4090,13 +3991,15 @@ impl Notification {
...
-const EXAMPLE_PROMPTS: [&str; 6] = [
+const EXAMPLE_PROMPTS: [&str; 8] = [
     "Explain this codebase",
     "Summarize recent commits",
     "Implement {feature}",
     "Find and fix a bug in @filename",
     "Write tests for @filename",
     "Improve documentation in @filename",
+    "Run /review on my current changes",
+    "Use /skills to list available skills",
```
nornagon-openai, line 4002: I don't think these are actually example prompts, like you wouldn't actually type "Run /review on my current changes" into the input box

codex-rs/tui/tooltips.txt
joshka-oai, line 1: WDYT about adding links to the docs for many of these to make it easy for a user to find out more about the feature?

joshka-oai (approved):
Approved. I agree with @nornagon-openai on the prompts, but it's probably no big deal as these aren't actually used as prompts, they're just displayed as placeholder text. So perhaps the EXAMPLE_PROMPTS is mislabeled.

codex-rs/tui/tooltips.txt
mzeng-openai: I'm adding the tooltips for the webview, we can add the links there which consumes less space and will be clickable.

codex-rs/tui/src/chatwidget.rs
mzeng-openai: Yeah they are more placeholders than example prompts, let me rename the variable here.

codex-rs/tui/tooltips.txt
joshka-oai: You can often click links in the terminal either directly or by holding some key that varies by terminal. There's also the ability to use OSC8 escape codes for this, but we haven't got a good pattern in Ratatui for that yet. Not a blocker on this PR - more a question about whether you think it might be worth doing.

mzeng-openai: Good to know the OSC8 escape code approach, i'll follow up with that.

joshka-oai: Note - it's probably not a viable approach (yet) - https://github.com/search?q=repo%3Aratatui%2Fratatui+OSC8&type=issues

PR #9140, opened 2026-01-13
Use thread rollback for Esc backtrack

aibrahim-oai:
- Swap Esc backtrack to roll back the current thread instead of forking


aibrahim-oai:
@codex review this

jif-oai (approved):
way cleaner

PR #9146, opened 2026-01-13
feat: return an error if the image sent by the user is a bad image

jif-oai:
## Before
When we detect an `InvalidImageRequest`, we replace the image by a placeholder and keep going

## Now
In such `InvalidImageRequest`, we check if the image is due to a user message or a tool call output. For tool call output we still replace it with a placeholder to avoid breaking the agentic loop bu tif this is because of a user message, we send an error to the user

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -2613,9 +2613,18 @@ pub(crate) async fn run_turn(
             Err(CodexErr::InvalidImageRequest()) => {
                 let mut state = sess.state.lock().await;
                 error_or_panic(
-                    "Invalid image detected, replacing it in the last turn to prevent poisoning",
+                    "Invalid image detected; sanitizing tool output to prevent poisoning",
                 );
-                state.history.replace_last_turn_images("Invalid image");
+                if state.history.replace_last_turn_images("Invalid image") {
+                    continue;
+                }
+                let event = EventMsg::Error(ErrorEvent {
+                    message: "Invalid image in your last message. Please remove it and try again."
+                        .to_string(),
+                    codex_error_info: Some(CodexErrorInfo::BadRequest),
+                });
+                sess.send_event(&turn_context, event).await;
+                break;
             }
             Err(e) => {
                 info!("Turn error: {e:#}");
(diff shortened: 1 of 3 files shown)


shijie-oai (approved):

codex-rs/core/src/codex.rs
```diff
@@ -2613,9 +2613,18 @@ pub(crate) async fn run_turn(
...
-                state.history.replace_last_turn_images("Invalid image");
+                if state.history.replace_last_turn_images("Invalid image") {
+                    continue;
+                }
+                let event = EventMsg::Error(ErrorEvent {
+                    message: "Invalid image in your last message. Please remove it and try again."
+                        .to_string(),
+                    codex_error_info: Some(CodexErrorInfo::BadRequest),
+                });
+                sess.send_event(&turn_context, event).await;
```
shijie-oai, line 2626: So for app server, we are still sending a generic `EventMsg::Error` with a specific message and I think we can parse that out. 

Not sure if we are at a point for us to develop specific error types at this point just for app sever consumers.

shijie-oai: We can create a new error type at https://github.com/openai/codex/blob/5029060ae7caba88abedc33a62fae8298fde3b66/codex-rs/protocol/src/protocol.rs and use it here to better document that this is a invalidUserInput and GUI and render as expected.  

cc @owenlin0

jif-oai: Not sure either but ping me if you change your mind

shijie-oai: We can add an `InvalidUserImageError` to `CodexErrorInfo` and here we can do `codex_error_info: Some(CodexErrorInfo::InvalidUserImageError)`.

jif-oai:
@codex review

codex-rs/core/src/codex.rs
owenlin0: @shijie-oai If we want to match on this error specifically and render something on the client, adding a more specific error type is a good idea

> For tool call output we still replace it with a placeholder to avoid breaking the agentic loop bu tif this is because of a user message, we send an error to the user

Nice, I think this error should show up on `turn/completed` with turn.status failed, up to you @shijie-oai on what's nicer to use between that or the `error` notification on the client.

cc @celia-oai who implemented codex errors in v2 api

PR #9149, opened 2026-01-13
Cloud tasks: retry master when fallback main is missing

### LINKED ISSUE #9111: Submitting cloud tasks from Codex CLI fails with "Provided git ref main does not exist" when master branch is used

codex-cli 0.80.0

ChatGPT Business

Not visible in the errored cloud logs, whatever Codex Cloud uses by default I guess.

Darwin 25.1.0 arm64 arm (Not sure why you need this since the error is in Cloud)

On our repos which have a `master` branch instead of a `main` branch, tasks sent to Codex Cloud try to check out the `main` branch. The default is `master` here, which is correctly configured in Github too. `main` is never mentioned anywhere in the `.git` info or on our GitHub settings.

1. Have repo where the default branch is not named main (e.g. master).
2. Submit new task from codex cloud CLI.

The submitted task would run against the default branch, regardless of it's name.

### **Output in codex:**
Image

### **Git config:**
```
[core]
        repositoryformatversion = 0
        filemode = true
        bare = false
        logallrefupdates = true
        ignorecase = true
        precomposeunicode = true
[remote "origin"]
        url = https://github.com/***/***.git
        fetch = +refs/heads/*:refs/remotes/origin/*
[branch "master"]
        remote = origin
        merge = refs/heads/master
        vscode-merge-base = origin/master
```

### **Git default branch setting:**
Image

liqiongyu:
Fixes #9111.

When the CLI can't determine a git branch locally, it currently falls back to "main". For repos whose default branch is "master", the backend rejects the task with an invalid git ref.

This change keeps the existing behavior when the branch was explicitly provided or resolved from git info, but if we had to use the hardcoded fallback and the backend reports that "main" doesn't exist, we retry once with "master".

- Adds unit tests covering the retry + non-retry cases.

### DIFF
--- a/codex-rs/cloud-tasks/src/lib.rs
+++ b/codex-rs/cloud-tasks/src/lib.rs
@@ -127,31 +127,104 @@ impl GitInfoProvider for RealGitInfo {
     }
 }
 
-async fn resolve_git_ref(branch_override: Option<&String>) -> String {
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum GitRefResolution {
+    Override,
+    CurrentBranch,
+    DefaultBranch,
+    Fallback,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct ResolvedGitRef {
+    name: String,
+    resolution: GitRefResolution,
+}
+
+async fn resolve_git_ref(branch_override: Option<&String>) -> ResolvedGitRef {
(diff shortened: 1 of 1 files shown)


liqiongyu:

etraut-openai:
Thanks for the contribution, but I don't think this is the right fix for this problem. I like the solution you suggested [here](https://github.com/openai/codex/issues/9154).

PR #9155, opened 2026-01-13
fix: shell snapshot clean-up

jif-oai:
Clean all shell snapshot files corresponding to sessions that have not been updated in 7 days
Those files should never leak. The only known cases were it can leak are during non graceful interrupt of the process (`kill -9, `panic`, OS crash, ...)


jif-oai:
@codex review

dylan-hurd-oai (approved):
I wonder if we want to provide some mechanism to update the shell snapshot for a session? either when resuming or allowing the model to request it somehow

jif-oai:
@codex review

PR #9156, opened 2026-01-13
chore: clamp min yield time for empty write_stdin

jif-oai:
After evals, 0 impact on performance

### DIFF
--- a/codex-rs/core/src/unified_exec/process_manager.rs
+++ b/codex-rs/core/src/unified_exec/process_manager.rs
@@ -31,6 +31,7 @@ use crate::truncate::approx_token_count;
 use crate::truncate::formatted_truncate_text;
 use crate::unified_exec::ExecCommandRequest;
 use crate::unified_exec::MAX_UNIFIED_EXEC_PROCESSES;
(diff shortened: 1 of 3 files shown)


codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -274,7 +276,14 @@ impl UnifiedExecProcessManager {
         }
 
         let max_tokens = resolve_max_tokens(request.max_output_tokens);
-        let yield_time_ms = clamp_yield_time(request.yield_time_ms);
+        let yield_time_ms = {
+            let time_ms = clamp_yield_time(request.yield_time_ms);
```
pakrym-oai, line 280: do we need to pre clamp before the condition?

pakrym-oai (approved):

codex-rs/core/src/unified_exec/process_manager.rs
jif-oai: This is cheap and my logic was: if one day we add some custom logic in it, might be worth it to keep. But this is mostly because it's basically free

PR #9166, opened 2026-01-13
fix: integration test for #9011

bolinfest:
Adds an integration test for the new behavior introduced in https://github.com/openai/codex/pull/9011. The work to create the test setup was substantial enough that I thought it merited a separate PR.

This integration test spawns `codex` in TUI mode, which requires spawning a PTY to run successfully, so I had to introduce quite a bit of scaffolding in `run_codex_cli()`. I was surprised to discover that we have not done this in our codebase before, so perhaps this should get moved to a common location so it can be reused.

The test itself verifies that a malformed `rules` in `$CODEX_HOME` prints a human-readable error message and exits nonzero.


codex-rs/tui/tests/suite/no_panic_on_startup.rs
```diff
@@ -0,0 +1,112 @@
...
+    let mut exit_rx = spawned.exit_rx;
+    let writer_tx = spawned.session.writer_sender();
+    let exit_code_result = timeout(Duration::from_secs(10), async {
+        loop {
+            select! {
+                result = output_rx.recv() => match result {
+                    Ok(chunk) => {
+                        if chunk.windows(4).any(|window| window == b"\x1b[6n") {
+                            let _ = writer_tx.send(b"\x1b[1;1R".to_vec()).await;
+                        }
```
joshka-oai, line 90: What is this?

codex-rs/tui/tests/suite/no_panic_on_startup.rs
```diff
@@ -0,0 +1,112 @@
...
+    );
+    Ok(())
+}
+
+struct CodexCliOutput {
+    exit_code: i32,
+    output: String,
+}
+
+async fn run_codex_cli(
```
joshka-oai, line 55: Would be good to have a small amount of inline / doc comments on this one to help give details about how it works.

Right now all I could really say is that there's a bunch of code that seems to work and codex understood it when it wrote it. I'd have to read every line to understand this code and then infer constraints / edge cases that are encoded here.

joshka-oai (approved):

PR #9174, opened 2026-01-13
Renew cache ttl on etag match

aibrahim-oai:
so we don't do unnecessary fetches


codex-rs/core/tests/suite/models_cache_ttl.rs
```diff
@@ -0,0 +1,175 @@
+#![cfg(not(target_os = "windows"))]
```
pakrym-oai: why not windows?

Does this need a separate file?

codex-rs/core/tests/suite/models_cache_ttl.rs
```diff
@@ -0,0 +1,175 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct ModelsCache {
+    fetched_at: DateTime<Utc>,
+    #[serde(default)]
+    etag: Option<String>,
+    models: Vec<ModelInfo>,
+}
+
+fn test_remote_model(slug: &str, priority: i32) -> ModelInfo {
+    serde_json::from_value(serde_json::json!({
```
pakrym-oai: will this be annoying to keep in sync?

pakrym-oai (approved):

codex-rs/core/tests/suite/models_cache_ttl.rs
aibrahim-oai: good idea. lemme try to make it strongly typed and a helper.

PR #9176, opened 2026-01-13
fix: Prefer raw strings (r#) where appropriate, particularly in tests.

bolinfest:
Updated AGENTS.md to memorialize this practice.

### DIFF
--- a/codex-rs/apply-patch/src/parser.rs
+++ b/codex-rs/apply-patch/src/parser.rs
@@ -450,13 +450,10 @@ fn test_parse_patch() {
 
     assert_eq!(
         parse_patch_text(
-            concat!(
-                "*** Begin Patch",
-                " ",
-                "\n*** Add File: foo\n+hi\n",
-                " ",
-                "*** End Patch"
-            ),
+            r#"*** Begin Patch
+*** Add File: foo
++hi
+ *** End Patch"#,
             ParseMode::Strict
         )
         .unwrap()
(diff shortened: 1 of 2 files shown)


codex-rs/apply-patch/src/parser.rs
```diff
@@ -490,17 +487,17 @@ fn test_parse_patch() {
...
-             +abc\n\
-             +def\n\
-             *** Delete File: path/delete.py\n\
-             *** Update File: path/update.py\n\
-             *** Move to: path/update2.py\n\
-             @@ def f():\n\
-             -    pass\n\
-             +    return 123\n\
-             *** End Patch",
+            r#"*** Begin Patch
```
pakrym-oai, line 490: What happens to \n cross plat (windows)?

pakrym-oai: they'll stay \n

pakrym-oai (approved):

joshka-oai (changes requested):
I'd prefer [indoc](https://docs.rs/indoc/latest/indoc/) over raw strings as the wrapping is handled better - i.e. indents are ignored. This makes mutliline strings fairly readable while being somewhat resiliant to refactoring / changes.

dylan-hurd-oai (approved):

bolinfest:
> I'd prefer [indoc](https://docs.rs/indoc/latest/indoc/) over raw strings as the wrapping is handled better - i.e. indents are ignored. This makes mutliline strings fairly readable while being somewhat resiliant to refactoring / changes.

I think that's a reasonable thing to do, but that migration is a bit more involved. I'm not volunteering to do that, but I would approve a PR that did!

PR #9178, opened 2026-01-13
tui: treat Alt+Shift+Backspace as word delete

### LINKED ISSUE #9161: Support Option+backspace to delete previous word

0.80.0

Plus

gpt-5.2-codex

Darwin 23.6.0 arm64 arm

When I press Option+Backspace, instead of deleting a whole word like on other apps, codex deletes just a character, making it painfully slow to rewrite sentences 

type a phrase like "hello world" and hit Option+Backspace. The typed text becomes "hello worl".

It should become "hello ", deleting the whole word.

### Additional information

_No response_

senkushal:
Fixes: https://github.com/openai/codex/issues/9161

Details 
- Treat Alt+Shift+Backspace as delete‑previous‑word (macOS terminals often send Alt+Shift+Backspace).
- Apply the same behavior in both tui and tui2.
- Add tests in tui and tui2 for Alt+Shift+Backspace.

Tests
 - cargo test -p codex-tui delete_backward_word_alt_shift_backspace
 - cargo test -p codex-tui2 delete_backward_word_alt_shift_backspac


senkushal:

etraut-openai:
@codex review

etraut-openai:
Thanks for the contribution. Please look at the code review feedback from Codex.

senkushal:
@etraut-openai you are welcome, fixed by excluding AltGr form work-delete

senkushal:
@codex Review

senkushal:
@codex Review

joshka-oai:
Thanks for working on this. I spent a bit of time digging into what signal we are actually getting from terminals here because it was not obvious to me how reliable this chord really is.

On macOS, Alt+Delete already behaves as delete previous word in the places most people expect it to, including iTerm, Ghostty, and VS Code. That lines up with common editor and terminal conventions. In crossterm, the macOS Delete key is reported as a Backspace key event, so this mostly comes down to how terminals choose to populate modifiers on that Backspace event.

What this change effectively does is relax the matching from an exact ALT modifier to any key event whose modifiers include ALT. That means Alt+Shift+Delete is treated the same as Alt+Delete.

However, when I tested this in Terminal.app (using crossterm’s event::read), the Backspace event does not include modifier information at all when Alt, Shift, or other modifiers are held. It is always reported as a plain Backspace. As a result, I did not see any behavioral change in Terminal.app with this patch, because there is no distinguishable Alt or Alt+Shift signal for us to key off of.

Given that, I am hesitant to broaden the ALT matching here without clearer evidence that terminals are actually emitting Alt+Shift+Delete as a meaningful, distinct chord. Alt+Delete already works in the environments where this is supported, and broadening the match risks changing behavior in ways that are hard to reason about.

To help decide how this should behave, it would be very helpful to know which terminals you have tested this in on macOS, which ones actually emit a distinct Alt+Shift+Delete modifier combination (rather than just Backspace), and whether there is a terminal or standard that explicitly defines Alt+Shift+Delete as delete word.

If you want to help gather that data, an easy way to check is to clone crossterm-rs/crossterm and run the event-read example, then press Delete, Alt+Delete, and Alt+Shift+Delete in different terminals to see what modifiers are actually reported.

etraut-openai:
@joshka-oai, thanks for the analysis.

@senkushal90, based on that input, I'm going to close this PR since this is probably not the direction we want to go. If you'd like to help with the analysis and exploration of potential solutions, please post to the original bug report. I'll also update that bug report with an explicit link to this PR so others who come across it can benefit from this discussion thread.

senkushal:
@joshka-oai  & @etraut-openai Thanks for the deep dive, that explains why this is trickier than it looked.
I tested in VS Code’s integrated terminal and iTerm2, where Alt+Shift+Backspace emits a distinct modifier combination via crossterm, which is what led me down this path. I didn’t realize Terminal.app collapses all of these into plain Backspace.

I’ll clone crossterm and run the event-read example across Terminal.app, iTerm2, Ghostty, and VS Code to collect concrete data on what modifiers are actually emitted. I’ll report back with a table of results so we can reason about what’s safe to support.

joshka-oai:
I use VSCode terminal with Codex all the time - Alt+Backspace works there already and is the right keystroke to support. What I'm trying to understand is what you're doing that's different to that? Why do you need to press Shift there?

senkushal:
On my machine in iTerm2, Option+Backspace appears to come through as ALT|SHIFT (even without Shift). VS Code seems to normalize it to ALT only, which is why it already works there. My intent was to make those terminals behave consistently.

PR #9179, opened 2026-01-14
Use current model for review

pakrym-oai:
Instead of having a hard-coded default review model, use the current model for running `/review` unless one is specified in the config.

Also inherit current reasoning effort

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -3389,7 +3389,9 @@ impl CodexMessageProcessor {
                 })?;
 
         let mut config = self.config.as_ref().clone();
-        config.model = Some(self.config.review_model.clone());
+        if let Some(review_model) = &config.review_model {
+            config.model = Some(review_model.clone());
+        }
 
         let NewThread {
             thread_id,
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -2370,7 +2370,10 @@ async fn spawn_review_thread(
     sub_id: String,
     resolved: crate::review_prompts::ResolvedReviewRequest,
 ) {
-    let model = config.review_model.clone();
+    let model = config
+        .review_model
+        .clone()
+        .unwrap_or_else(|| parent_turn_context.client.get_model());
     let review_model_info = sess
         .services
         .models_manager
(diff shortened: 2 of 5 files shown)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3389,7 +3389,9 @@ impl CodexMessageProcessor {
                 })?;
 
         let mut config = self.config.as_ref().clone();
-        config.model = Some(self.config.review_model.clone());
+        if let Some(review_model) = &config.review_model {
```
aibrahim-oai, line 3392: is there a difference between self.config and config?

codex-rs/core/src/config/mod.rs
```diff
@@ -108,8 +106,8 @@ pub struct Config {
     /// Optional override of model selection.
     pub model: Option<String>,
 
-    /// Model used specifically for review sessions. Defaults to "gpt-5.1-codex-max".
-    pub review_model: String,
+    /// Model used specifically for review sessions.
+    pub review_model: Option<String>,
```
aibrahim-oai, line 110: Thanks!

codex-rs/core/src/codex.rs
```diff
@@ -2394,14 +2397,13 @@ async fn spawn_review_thread(
 
     // Build per‑turn client with the requested model/family.
     let mut per_turn_config = (*config).clone();
-    per_turn_config.model_reasoning_effort = Some(ReasoningEffortConfig::Low);
```
aibrahim-oai, line 2397: unrelated change?

codex-rs/app-server/src/codex_message_processor.rs
pakrym-oai: `config = self.config.as_ref().clone();` ?

codex-rs/core/tests/suite/review.rs
```diff
@@ -431,6 +431,56 @@ async fn review_uses_custom_review_model_from_config() {
     server.verify().await;
 }
 
+/// Ensure that when `review_model` is not set in the config, the review request
+/// uses the session model.
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn review_uses_session_model_when_review_model_unset() {
```
aibrahim-oai, line 437: let's have a test when it's set as well

pakrym-oai: There is one, right above

aibrahim-oai (approved):

codex-rs/app-server/src/codex_message_processor.rs
aibrahim-oai: yeah was just wondering why are we using self a line above it.

codex-rs/core/src/codex.rs
pakrym-oai: Related, I'd like to use the current reasoning effort as well.

etraut-openai:
Please make sure that any behavior changes here are also reflected in the VSCE, which also implements a `/review` feature.

PR #9185, opened 2026-01-14
chore: prefer indoc for multiline strings

joshka-oai:
**Summary**
- Add `AGENTS.md` guidance to prefer `indoc!`/`formatdoc!` over embedded `\n` escapes and awkward raw-string wrapping.
- Refactor multi-line literals in core/tui/utils to use indoc macros consistently.
- Keep raw strings where needed (e.g. `create_apply_patch_json_tool` has interior quotes).

**Tests**
- `cd codex-rs && just fmt`
- `cargo test -p codex-tui` (pass)
- `cargo test -p codex-git` (pass)
- `cargo test -p codex-core` (fails 12 tests in `core/tests/all.rs`; appears pre-existing in this environment)

### DIFF
--- a/codex-rs/core/src/models_manager/model_presets.rs
+++ b/codex-rs/core/src/models_manager/model_presets.rs
@@ -320,13 +320,13 @@ fn gpt_52_codex_upgrade() -> ModelUpgrade {
                 .to_string(),
         ),
         migration_markdown: Some(
-            indoc! {r#"
-                **Codex just got an upgrade. Introducing {model_to}.**
+            indoc! {"
+**Codex just got an upgrade. Introducing {model_to}.**
 
-                Codex is now powered by gpt-5.2-codex, our latest frontier agentic coding model. It is smarter and faster than its predecessors and capable of long-running project-scale work. Learn more about {model_to} at https://openai.com/index/introducing-gpt-5-2-codex
+Codex is now powered by gpt-5.2-codex, our latest frontier agentic coding model. It is smarter and faster than its predecessors and capable of long-running project-scale work. Learn more about {model_to} at https://openai.com/index/introducing-gpt-5-2-codex
 
-                You can continue using {model_from} if you prefer.
-            "#}
+You can continue using {model_from} if you prefer.
+"}
             .to_string(),
         ),
     }
--- a/codex-rs/core/tests/suite/list_models.rs
+++ b/codex-rs/core/tests/suite/list_models.rs
@@ -412,13 +412,13 @@ fn gpt52_codex_upgrade() -> codex_protocol::openai_models::ModelUpgrade {
                 .to_string(),
         ),
         migration_markdown: Some(
-            indoc! {r#"
-                **Codex just got an upgrade. Introducing {model_to}.**
+            indoc! {"
+**Codex just got an upgrade. Introducing {model_to}.**
 
-                Codex is now powered by gpt-5.2-codex, our latest frontier agentic coding model. It is smarter and faster than its predecessors and capable of long-running project-scale work. Learn more about {model_to} at https://openai.com/index/introducing-gpt-5-2-codex
+Codex is now powered by gpt-5.2-codex, our latest frontier agentic coding model. It is smarter and faster than its predecessors and capable of long-running project-scale work. Learn more about {model_to} at https://openai.com/index/introducing-gpt-5-2-codex
 
-                You can continue using {model_from} if you prefer.
-            "#}
+You can continue using {model_from} if you prefer.
+"}
             .to_string(),
         ),
     }
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,6 +4,7 @@ In the codex-rs folder where the rust code lives:
 
 - Crate names are prefixed with `codex-`. For example, the `core` folder's crate is named `codex-core`
 - When using format! and you can inline variables into {}, always do that.
+- Use `indoc!` for multi-line strings instead of embedding `\n` escapes or using awkwardly-wrapped raw strings.
 - Install any commands the repo relies on (for example `just`, `rg`, or `cargo-insta`) if they aren't already available before running instructions here.
 - Never add or modify any code related to `CODEX_SANDBOX_NETWORK_DISABLED_ENV_VAR` or `CODEX_SANDBOX_ENV_VAR`.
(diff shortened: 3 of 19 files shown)


codex-rs/core/src/models_manager/model_presets.rs
```diff
@@ -320,13 +320,13 @@ fn gpt_52_codex_upgrade() -> ModelUpgrade {
...
         migration_markdown: Some(
-            indoc! {r#"
+            indoc! {"
                 **Codex just got an upgrade. Introducing {model_to}.**
 
                 Codex is now powered by gpt-5.2-codex, our latest frontier agentic coding model. It is smarter and faster than its predecessors and capable of long-running project-scale work. Learn more about {model_to} at https://openai.com/index/introducing-gpt-5-2-codex
 
                 You can continue using {model_from} if you prefer.
-            "#}
+            "}
```
joshka-oai: This is actually a change - but it's probably more correct to have stripped the indentation here.

codex-rs/core/tests/suite/list_models.rs
```diff
@@ -417,13 +412,13 @@ fn gpt52_codex_upgrade() -> codex_protocol::openai_models::ModelUpgrade {
...
         migration_markdown: Some(
-            indoc! {r#"
+            indoc! {"
                 **Codex just got an upgrade. Introducing {model_to}.**
 
                 Codex is now powered by gpt-5.2-codex, our latest frontier agentic coding model. It is smarter and faster than its predecessors and capable of long-running project-scale work. Learn more about {model_to} at https://openai.com/index/introducing-gpt-5-2-codex
 
                 You can continue using {model_from} if you prefer.
-            "#}
+            "}
```
joshka-oai: As above

codex-rs/core/src/config/mod.rs
```diff
@@ -3336,53 +3337,53 @@ model = "gpt-5.1-codex"
...
-
-[profiles.gpt5]
-model = "gpt-5.1"
-model_provider = "openai"
-approval_policy = "on-failure"
-model_reasoning_effort = "high"
-model_reasoning_summary = "detailed"
-model_verbosity = "high"
-"#;
+        let toml = indoc! {r#"
```
bolinfest: Given the potential for long lines here, I'm not sure this is better this way?

codex-rs/core/src/models_manager/model_presets.rs
bolinfest: Hmm, is this surfaced in the UI anywhere?

codex-rs/core/src/skills/render.rs
```diff
@@ -19,23 +20,25 @@ pub fn render_skills_section(skills: &[SkillMetadata]) -> Option<String> {
...
-  - If multiple skills apply, choose the minimal set that covers the request and state the order you'll use them.
-  - Announce which skill(s) you're using and why (one short line). If you skip an obvious skill, say why.
-- Context hygiene:
-  - Keep context small: summarize long sections instead of pasting them; only load extra files when needed.
-  - Avoid deep reference-chasing: prefer opening only files directly linked from `SKILL.md` unless you're blocked.
-  - When variants exist (frameworks, providers, domains), pick only the relevant reference file(s) and note that choice.
-- Safety and fallback: If a skill can't be applied cleanly (missing files, unclear instructions), state the issue, pick the next-best approach, and continue."###
-            .to_string(),
+        indoc! {"
+            - Discovery: The list above is the skills available in this session (name + description + file path). Skill bodies live on disk at the listed paths.
```
bolinfest: Similarly, this indent just results in more line wrapping for the reader?

bolinfest (commented):
In a number of these examples, I think the non-indoc version was better because I could copy-paste the content (like a sample `config.toml` file) into somewhere else, or into a PR body I am drafting, and I don't have to fix the indent.

Particularly for long pieces of content, it is hard to find the "leftmost" edge of the content to understand where the true edge is. (It also requires more line wrapping to read in an editor.)

I feel like `indoc!` works better for things in the small, or where the leading newline would throw things off (whereas `r#` preserves it).

codex-rs/core/src/models_manager/model_presets.rs
joshka-oai: We have tests on the UI that pass on this. - the snap file is: https://github.com/openai/codex/blob/main/codex-rs/tui/src/snapshots/codex_tui__app__tests__model_migration_prompt_shows_for_hidden_model.snap

codex-rs/core/src/skills/render.rs
joshka-oai: I think this one probably should be an include_str / use a template type of lib.

codex-rs/core/src/config/mod.rs
joshka-oai: I'm 50/50 on this - I don't mind the indented doc in general as it makes the code easier to scan top to bottom as the indents are natural. I also don't mind removing it as it's not offensively bad.

joshka-oai: [279314d](https://github.com/openai/codex/pull/9185/commits/279314d16ac8e3066f0170bec6a7cb709ac1ddbc)

codex-rs/core/src/skills/render.rs
joshka-oai: [279314d](https://github.com/openai/codex/pull/9185/commits/279314d16ac8e3066f0170bec6a7cb709ac1ddbc)

bolinfest (commented):
Proposal: as a general rule, if the content is Markdown, it should be left-aligned. It can still be done within an in `indoc!`, though.

codex-rs/core/src/models_manager/model_presets.rs
joshka-oai: actually - this is not a change.
I read it as r# -> indoc!, but it's indoc!{r# -> indoc!.
This is the same string.

joshka-oai:
> Proposal: as a general rule, if the content is Markdown, it should be left-aligned. It can still be done within an in `indoc!`, though.

I think perhaps larger multi-line blocks (where longer lines and wrapping matter), but smaller low line count ones seem reasonable to keep indented with code. The main value I see in using indoc is that it makes it easy to have indentation that follows the program's natural flow rather than having to have it stacked on the left margin when there's newlines.

PR #9191, opened 2026-01-14
Get model on session configured

aibrahim-oai:
- Don't try to precompute model unless you know it from `config`
- Block `/model` on session configured
- Queue messages until session configured
- show "loading" in status until session configured

### DIFF
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -27,6 +27,7 @@ use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
 
+use crate::version::CODEX_CLI_VERSION;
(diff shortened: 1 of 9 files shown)


codex-rs/tui/src/chatwidget.rs
```diff
@@ -2105,7 +2131,13 @@ impl ChatWidget {
     }
 
     fn add_boxed_history(&mut self, cell: Box<dyn HistoryCell>) {
-        if !cell.display_lines(u16::MAX).is_empty() {
+        let hold_session_header = !self.is_session_configured()
```
joshka-oai: A small doc here would go a long way to explaining what's happening here.
I'm not sure that hold_session_header is particularly clear about what's being done here

joshka-oai (approved):

PR #9195, opened 2026-01-14
chore(windows) Enable Powershell UTF8 feature

dylan-hurd-oai:
## Summary
We've received a lot of positive feedback about this feature, so we're going to enable it by default.

### DIFF
--- a/codex-rs/core/src/features.rs
+++ b/codex-rs/core/src/features.rs
@@ -405,9 +405,9 @@ pub const FEATURES: &[FeatureSpec] = &[
         key: "powershell_utf8",
         #[cfg(windows)]
         stage: Stage::Beta {
-            name: "Powershell UTF8",
-            menu_description: "Enable UTF8 output in Powershell.",
-            announcement: "Try UTF8 output in Powershell for better Unicode support. Enable in /experimental!",
+            name: "Powershell UTF-8 support",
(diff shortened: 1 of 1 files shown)


codex-rs/core/src/features.rs
```diff
@@ -403,7 +403,17 @@ pub const FEATURES: &[FeatureSpec] = &[
...
         id: Feature::PowershellUtf8,
         key: "powershell_utf8",
+        #[cfg(windows)]
+        stage: Stage::Beta {
+            name: "Powershell UTF-8 support",
+            menu_description: "Enable UTF-8 output in Powershell.",
+            announcement: "Try UTF-8 output in Powershell for better Unicode support. Enable in /experimental!",
+        },
+        #[cfg(windows)]
+        default_enabled: false,
```
bolinfest: I'm OK defaulting to `true` based on what we know.

bolinfest (approved):

codex-rs/core/src/features.rs
etraut-openai: Yeah, I haven't heard of any problems with this feature. Everyone who has tried it seems to be happy.

dylan-hurd-oai: Sounds good, will switch!

PR #9216, opened 2026-01-14
add WebSearchMode enum

sayan-oai:
### What
Add `WebSearchMode` enum (disabled, cached live, defaults to disabled) to config + V2 protocol. This enum takes precedence over legacy flags: `web_search_cached`, `web_search_request`, and `tools.web_search`.

Keep `--search` as live.

### Tests
Added tests


codex-rs/core/src/tasks/review.rs
```diff
@@ -87,7 +88,9 @@ async fn start_review_conversation(
     // re-enable blocked tools (web search, view image).
     sub_agent_config
         .features
-        .disable(crate::features::Feature::WebSearchRequest);
+        .disable(crate::features::Feature::WebSearchRequest)
```
pakrym-oai: do we need these?

codex-rs/core/tests/suite/model_tools.rs
```diff
@@ -132,6 +137,7 @@ async fn model_selects_expected_tools() {
             "read_mcp_resource".to_string(),
             "update_plan".to_string(),
             "apply_patch".to_string(),
+            "web_search".to_string(),
```
pakrym-oai, line 144: does this PR turn search on by default? Let's wait a bit until we do that.

pakrym-oai (approved):
LGTM but let's not enable it on by default just yet.

codex-rs/core/src/tasks/review.rs
sayan-oai: not explicitly; setting the new enum to disabled is enough. added for clarity, can rm

codex-rs/core/tests/suite/model_tools.rs
sayan-oai: it does. i can make the default disabled if we want to merge this, then we can flip the default to cached later

sdk/typescript/src/exec.ts
```diff
@@ -88,8 +95,10 @@ export class CodexExec {
       );
     }
 
-    if (args.webSearchEnabled !== undefined) {
-      commandArgs.push("--config", `features.web_search_request=${args.webSearchEnabled}`);
+    if (args.webSearchMode) {
+      commandArgs.push("--config", `web_search="${args.webSearchMode}"`);
+    } else if (args.webSearchEnabled === true) {
+      commandArgs.push("--config", `web_search="live"`);
```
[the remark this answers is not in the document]
pakrym-oai: Valid comment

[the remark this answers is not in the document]
sayan-oai: yeah fixing

PR #9217, opened 2026-01-14
feat(app-server): add prompt/list API

owenlin0:
Add a new `prompt/list` API that fetches the user's custom prompts from `$CODEX_HOME/prompts`. https://developers.openai.com/codex/custom-prompts

Currently our VSCode IDE extension reads it directly from disk, but it'll be better for app-server to provide a proper API for this. We currently only support `list` (no API for creating) since we expect users to manually create the markdown files.


owenlin0:
@codex review

owenlin0:
@codex review

owenlin0:
@codex review

PR #9220, opened 2026-01-14
fix(tui): disable double-press quit shortcut

joshka-oai:
Disables the default Ctrl+C/Ctrl+D double-press quit UX (keeps the code path behind a const) while we rethink the quit/interrupt flow.

Tests:
- just fmt
- cargo clippy --fix --all-features --tests --allow-dirty --allow-no-vcs -p codex-tui
- cargo test -p codex-tui --lib

### DIFF
--- a/codex-rs/tui/src/bottom_pane/mod.rs
+++ b/codex-rs/tui/src/bottom_pane/mod.rs
@@ -71,6 +71,13 @@ pub(crate) use feedback_view::FeedbackNoteView;
 /// Keeping a single value ensures Ctrl+C and Ctrl+D behave identically.
 pub(crate) const QUIT_SHORTCUT_TIMEOUT: Duration = Duration::from_secs(1);
 
+/// Whether Ctrl+C/Ctrl+D require a second press to quit.
+///
+/// This UX experiment was enabled by default, but requiring a double press to quit feels janky in
+/// practice (especially for users accustomed to shells and other TUIs). Disable it for now while we
+/// rethink a better quit/interrupt design.
+pub(crate) const DOUBLE_PRESS_QUIT_SHORTCUT_ENABLED: bool = false;
+
 /// The result of offering a cancellation key to a bottom-pane surface.
 ///
 /// This is primarily used for Ctrl+C routing: active views can consume the key to dismiss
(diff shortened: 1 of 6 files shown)


codex-rs/tui/src/bottom_pane/mod.rs
```diff
@@ -266,15 +273,19 @@ impl BottomPane {
                     self.view_stack.pop();
                     self.on_active_view_complete();
                 }
-                self.show_quit_shortcut_hint(key_hint::ctrl(KeyCode::Char('c')));
+                if DOUBLE_PRESS_QUIT_SHORTCUT_ENABLED {
+                    self.show_quit_shortcut_hint(key_hint::ctrl(KeyCode::Char('c')));
```
bolinfest: Should the `DOUBLE_PRESS_QUIT_SHORTCUT_ENABLED` check be in `show_quit_shortcut_hint()` instead of outside it? Not sure what the full implications are.

codex-rs/tui/src/bottom_pane/mod.rs
```diff
@@ -71,6 +71,13 @@ pub(crate) use feedback_view::FeedbackNoteView;
 /// Keeping a single value ensures Ctrl+C and Ctrl+D behave identically.
 pub(crate) const QUIT_SHORTCUT_TIMEOUT: Duration = Duration::from_secs(1);
 
+/// Whether Ctrl+C/Ctrl+D require a second press to quit.
+///
+/// This UX experiment was enabled by default, but requiring a double press to quit feels janky in
+/// practice (especially for users accustomed to shells and other TUIs). Disable it for now while we
+/// rethink a better quit/interrupt design.
+pub(crate) const DOUBLE_PRESS_QUIT_SHORTCUT_ENABLED: bool = false;
```
bolinfest, line 79: I guess it's tricky to thread this through from `Config` at the moment for folks who want to try this out?

bolinfest (approved):

codex-rs/tui/src/bottom_pane/mod.rs
joshka-oai: Makes sense - https://github.com/openai/codex/compare/2e02110c90a493a6a4d244ea7b5b61c28292a7ff..a641f9feaabb9cf329fa4660b4b912a36da62eae

codex-rs/tui/src/bottom_pane/mod.rs
joshka-oai: We can do this - for now, focused on disabling this to ship the next alpha.

PR #9232, opened 2026-01-14
feat: propagate approval request of unsubscribed threads

jif-oai:
A thread can now be spawned by another thread. In order to process the approval requests of such sub-threads, we need to detect those event and show them in the TUI.

This is a temporary solution while the UX is being figured out. This PR should be reverted once done

### DIFF
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -51,7 +51,6 @@ use codex_protocol::ThreadId;
 use codex_protocol::openai_models::ModelPreset;
 use codex_protocol::openai_models::ModelUpgrade;
 use codex_protocol::openai_models::ReasoningEffort as ReasoningEffortConfig;
-use codex_protocol::protocol::ReviewDecision;
 use color_eyre::eyre::Result;
 use color_eyre::eyre::WrapErr;
 use crossterm::event::KeyCode;
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -1464,7 +1464,7 @@ impl App {
     }
 
     fn finish_external_approval(&mut self) {
-        if !self.external_approval_routes.is_empty() {
+        if self.external_approval_routes.is_empty() {
             while let Some(event) = self.paused_codex_events.pop_front() {
                 self.handle_codex_event_now(event);
             }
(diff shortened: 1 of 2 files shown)


codex-rs/tui/src/app.rs
```diff
@@ -868,7 +896,47 @@ impl App {
             AppEvent::FatalExitRequest(message) => {
                 return Ok(AppRunControl::Exit(ExitReason::Fatal(message)));
             }
-            AppEvent::CodexOp(op) => self.chat_widget.submit_op(op),
+            AppEvent::CodexOp(op) => match op {
+                Op::ExecApproval { id, decision } => {
```
joshka-oai, line 903: Probably worth pulling this out into a small method still.

codex-rs/tui/src/app.rs
```diff
@@ -868,7 +896,47 @@ impl App {
...
+                            },
+                        )
+                        .await;
+                        self.finish_external_approval();
+                    } else {
+                        self.chat_widget
+                            .submit_op(Op::PatchApproval { id, decision });
+                    }
+                }
+                _ => self.chat_widget.submit_op(op),
```
joshka-oai, line 944: Document with narrative / tests the expected approval behavior. Worth doing even if this is temporary.

codex-rs/tui/src/app.rs
```diff
@@ -1343,6 +1411,83 @@ impl App {
...
+        }
+        if let EventMsg::ListSkillsResponse(response) = &event.msg {
+            let cwd = self.chat_widget.config_ref().cwd.clone();
+            let errors = errors_for_cwd(&cwd, response);
+            emit_skill_load_warnings(&self.app_event_tx, &errors);
+        }
+        self.chat_widget.handle_codex_event(event);
+    }
+
+    fn handle_external_approval(&mut self, thread_id: ThreadId, mut event: Event) {
```
joshka-oai: Add some docs here that grounds the implementation against what you intend this to do

joshka-oai (commented):
This needs more comments to help ground what it's doing against the intent. I don't know if any of this is right because there's nothing to refer to here that would help clarify what the intended behavior is (add doc comments and tests if possible).

codex-rs/tui/src/app.rs
jif-oai: Added some doc but I generally prefer to inline such thing. Mainly as here the code the seems duplicate can't be cleanly generalized

joshka-oai (approved):
LGTM - though it seems that this behavior and the change it sits on doesn't appear to be in tui2. Perhaps add that for now?

jif-oai:
@codex review

jif-oai:
@codex review

PR #9235, opened 2026-01-14
Add text element metadata to types

charley-oai:
Initial type tweaking PR to make the diff of https://github.com/openai/codex/pull/9116 smaller

This should not change any behavior, just adds some fields to types

### DIFF
--- a/codex-rs/app-server-protocol/src/protocol/thread_history.rs
+++ b/codex-rs/app-server-protocol/src/protocol/thread_history.rs
@@ -197,6 +197,8 @@ impl ThreadHistoryBuilder {
         if !payload.message.trim().is_empty() {
             content.push(UserInput::Text {
                 text: payload.message.clone(),
+                // TODO: Thread text element ranges into thread history. Empty keeps old behavior.
+                text_elements: Vec::new(),
             });
         }
         if let Some(images) = &payload.images {
@@ -244,6 +246,8 @@ mod tests {
             EventMsg::UserMessage(UserMessageEvent {
                 message: "First turn".into(),
                 images: Some(vec!["https://example.com/one.png".into()]),
+                text_elements: Vec::new(),
+                local_images: Vec::new(),
             }),
             EventMsg::AgentMessage(AgentMessageEvent {
                 message: "Hi there".into(),
@@ -257,6 +261,8 @@ mod tests {
             EventMsg::UserMessage(UserMessageEvent {
                 message: "Second turn".into(),
                 images: None,
+                text_elements: Vec::new(),
+                local_images: Vec::new(),
             }),
             EventMsg::AgentMessage(AgentMessageEvent {
                 message: "Reply two".into(),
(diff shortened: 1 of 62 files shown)


codex-rs/app-server-protocol/src/protocol/thread_history.rs
```diff
@@ -197,6 +197,8 @@ impl ThreadHistoryBuilder {
         if !payload.message.trim().is_empty() {
             content.push(UserInput::Text {
                 text: payload.message.clone(),
+                // TODO: Thread text element ranges into thread history. Empty keeps old behavior.
```
aibrahim-oai, line 200: what do you mean?

charley-oai: This means this line will have to have data in the followup PR / `Vec::new()` will not be correct. I used these TODO comments to distinguish between where `Vec::new()` is genuinely correct, vs where we will need to thread through data and it's just empty now to make the PR minimal

codex-rs/protocol/src/protocol.rs
```diff
@@ -1255,8 +1255,19 @@ pub struct AgentMessageEvent {
...
+    /// the model.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub images: Option<Vec<String>>,
+    /// Local file paths sourced from `UserInput::LocalImage`. These are kept so
+    /// the UI can reattach images when editing history, and should not be sent
+    /// to the model or treated as API-ready URLs.
+    #[serde(default)]
+    pub local_images: Vec<std::path::PathBuf>,
+    /// UI-defined spans within `message` used to render or persist special elements.
+    #[serde(default)]
```
aibrahim-oai, line 1269: let's have them as option and skip if none?

codex-rs/protocol/src/user_input.rs
```diff
@@ -10,21 +10,48 @@ use ts_rs::TS;
 pub enum UserInput {
     Text {
         text: String,
+        /// UI-defined spans within `text` that should be treated as special elements.
+        /// These are byte ranges into the UTF-8 `text` buffer and are used to render
+        /// or persist rich input markers (e.g., image placeholders) across history
+        /// and resume without mutating the literal text.
+        #[serde(default)]
+        text_elements: Vec<TextElement>,
```
aibrahim-oai, line 18: let's have them as option and skip if none?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1543,21 +1543,55 @@ pub struct TurnInterruptParams {
...
 pub enum UserInput {
-    Text { text: String },
-    Image { url: String },
-    LocalImage { path: PathBuf },
-    Skill { name: String, path: PathBuf },
+    Text {
+        text: String,
+        /// UI-defined spans within `text` used to render or persist special elements.
+        #[serde(default)]
+        text_elements: Vec<TextElement>,
```
aibrahim-oai, line 1573: let's have them as option and skip if none?

codex-rs/protocol/src/protocol.rs
charley-oai: What would be the difference between `None` vs empty vector?

Owen mentioned we want to move away from `skip_serializing_if` in this comment: https://github.com/openai/codex/pull/9116#discussion_r2691461104

aibrahim-oai (approved):

Issue #8700, opened 2026-01-03
resume --last should scope to current working directory


janthmueller:
### Problem

`codex resume --last` currently resumes the **globally most recent session**.

For users working on **multiple projects in parallel**, this behavior is often unintuitive and unhelpful. The globally “last” session is frequently unrelated to the current working directory (CWD), which breaks the expectation that Codex sessions are tied to the directory in which the tool is invoked.

By contrast, running `codex resume` without flags opens a TUI that lists sessions **scoped to the current working directory**. This behavior aligns much better with typical workflows and is what I initially expected `resume --last` to do.

---

### Expected Behavior

`codex resume --last` should behave as follows:

* Resume the **most recent session associated with the current working directory (CWD)**
* Be functionally equivalent to:

  * running `codex resume`, and
  * automatically selecting the most recent session shown in the CWD-scoped TUI

In other words, `--last` should resolve the most recent session **within the current working directory**, rather than the most recent session across all directories.

---

### Why This Matters

For users managing multiple active projects, the global “last session” is rarely the correct one. Requiring interaction with the TUI introduces unnecessary friction into everyday workflows.

The core use case is the ability to **resume the correct Codex session for the current working directory with a single, non-interactive command**. This is essential for setups where Codex is launched via a keyboard shortcut or script (for example from a terminal multiplexer), where any additional selection step breaks the flow.

Making `resume --last` CWD-scoped would allow reliable, scriptable, one-command session startup per project, enabling Codex to be used comfortably as a **day-to-day development tool**.

---

### Suggested Options

* Change `resume --last` to be scoped to the current working directory
* **OR** introduce an explicit alternative flag, for example:

  * `codex resume --last-cwd`
  * `codex resume --local-last`

dgunay:
This is a small papercut in my workflow - I run codex out of a separate worktree per branch/feature and wanted to script resuming a session based on the folder. If I manually run `codex resume` in a folder, the session I'd expect shows up at the top of the list, but if I run `codex resume --last` then it may start a different session if I've started another since then. Makes it much more annoying to recover if I have to shut down my laptop or otherwise experience some issue that nukes my sessions.

etraut-openai:
This will be fixed in the next release (0.85.0).

janthmueller:
@etraut-openai can we have the same for exec resume --last?

etraut-openai:
@janthmueller, good call. I missed that in my earlier change.

PR #9245, opened 2026-01-15
Changed `codex resume --last` to honor the current cwd

etraut-openai:
This PR changes `codex resume --last` to work consistently with `codex resume`. Namely, it filters based on the cwd when selecting the last session. It also supports the `--all` modifier as an override.

This addresses #8700


codex-rs/core/src/rollout/recorder.rs
```diff
@@ -431,3 +461,40 @@ impl JsonlWriter {
...
+}
+
+fn extract_session_cwd(head: &[serde_json::Value]) -> Option<PathBuf> {
+    head.iter().find_map(|value| {
+        let meta_line = serde_json::from_value::<SessionMetaLine>(value.clone()).ok()?;
+        Some(meta_line.meta.cwd)
+    })
+}
+
+fn paths_match(a: &Path, b: &Path) -> bool {
```
jif-oai: I think we should inline those single-use function. IMO, the only reason to extract would be:
* testing
* multiple usage
* simplify the workflow of large functions

jif-oai (approved):

PR #9247, opened 2026-01-15
feat(app-server, core): return threads by created_at or updated_at

owenlin0:
Add support for returning threads by either `created_at` OR `updated_at` descending. Previously core always returned threads ordered by `created_at`. 

This PR:
- updates core to be able to list threads by `updated_at` OR `created_at` descending based on what the caller wants
- also update `thread/list` in app-server to expose this (default to `created_at` if not specified)

All existing codepaths (app-server, TUI) still default to `created_at`, so no behavior change is expected with this PR.

**Implementation**
To sort by `updated_at` is a bit nontrivial (whereas `created_at` is easy due to the way we structure the folders and filenames on disk, which are all based on `created_at`). 

The most naive way to do this without introducing a cache file or sqlite DB (which we have to implement/maintain) is to scan files in reverse `created_at` order on disk, and look at the file's mtime (last modified timestamp according to the filesystem) until we reach `MAX_SCAN_FILES` (currently set to 10,000). Then, we can return the most recent N threads. 

Based on some quick and dirty benchmarking on my machine with ~1000 rollout files, calling `thread/list` with limit 50, the `updated_at` path is slower as expected due to all the I/O:
- updated-at: average 103.10 ms
- created-at: average 41.10 ms

Those absolute numbers aren't a big deal IMO, but we can certainly optimize this in a followup if needed by introducing more state stored on disk.

**Caveat**
There's also a limitation in that any files older than `MAX_SCAN_FILES` will be excluded, which means if a user continues a REALLY old thread, it's possible to not be included. In practice that should not be too big of an issue.

If a user makes...
- 1000 rollouts/day → threads older than 10 days won't show up
- 100 rollouts/day → ~100 days

If this becomes a problem for some reason, even more motivation to implement an updated_at cache.


owenlin0:
@codex review

owenlin0:
@codex review

owenlin0:
@codex review

codex-rs/app-server-protocol/src/protocol/v1.rs
```diff
@@ -128,6 +126,7 @@ pub struct ConversationSummary {
     pub path: PathBuf,
     pub preview: String,
     pub timestamp: Option<String>,
+    pub updated_at: Option<String>,
```
jif-oai, line 131: Do we really want to maintain it?

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -1598,15 +1601,20 @@ impl CodexMessageProcessor {
...
+            sort_key,
             model_providers,
         } = params;
 
         let requested_page_size = limit
             .map(|value| value as usize)
             .unwrap_or(THREAD_LIST_DEFAULT_LIMIT)
             .clamp(1, THREAD_LIST_MAX_LIMIT);
+        let core_sort_key = match sort_key.unwrap_or(ThreadSortKey::CreatedAt) {
+            ThreadSortKey::CreatedAt => CoreThreadSortKey::CreatedAt,
```
jif-oai, line 1613: Why do you need a Core version?

codex-rs/app-server/tests/suite/v2/thread_list.rs
```diff
@@ -82,6 +102,28 @@ fn timestamp_at(
     )
 }
 
+fn rollout_path(codex_home: &Path, filename_ts: &str, thread_id: &str) -> PathBuf {
```
jif-oai: Are you sure we don't already have a helper for this

jif-oai (commented):
We should document what happen if you change ordering key while keeping the same pagination token (I think we should return an error)

aibrahim-oai (approved):

codex-rs/app-server/tests/suite/v2/thread_list.rs
owenlin0: refactored

codex-rs/app-server/src/codex_message_processor.rs
owenlin0: just a pattern we have, not sure if we want core to directly use an app-server type

codex-rs/app-server-protocol/src/protocol/v1.rs
owenlin0: codex just went ahead and did it, but i am very down to rm -rf v1 as soon as we can

PR #9249, opened 2026-01-15
[search] allow explicitly disabling web search

sayan-oai:
moving `web_search` rollout serverside, so need a way to explicitly disable search + signal eligibility from the client.

- Add `x‑oai‑web‑search‑eligible` header that signifies whether the request can have web search.
- Only attach the `web_search` tool when the resolved `WebSearchMode` is `Live` or `Cached`.

### DIFF
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -1204,6 +1204,7 @@ fn resolve_web_search_mode(
     WebSearchMode::default()
 }
 
+/// Returns Some(true) only when the user explicitly disables web search
 fn resolve_explicit_web_search_disabled(
     config_toml: &ConfigToml,
     config_profile: &ConfigProfile,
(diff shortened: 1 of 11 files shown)


codex-rs/core/src/config/mod.rs
```diff
@@ -338,6 +338,8 @@ pub struct Config {
     pub include_apply_patch_tool: bool,
 
     pub web_search_mode: WebSearchMode,
+    /// Set when the user explicitly disables web search in config.
```
pakrym-oai: make the config value `Optional` ?

codex-rs/core/tests/responses_headers.rs
```diff
@@ -213,6 +214,103 @@ async fn responses_stream_includes_subagent_header_on_other() {
     );
 }
 
+#[tokio::test]
+async fn responses_stream_includes_web_search_disabled_header() {
```
pakrym-oai: can we use test_code here? I know some other tests might use the verbose pattern but it's not worth spreading it.

pakrym-oai (changes requested):
I think we can simplify by having web_search_mode be Optional

pakrym-oai (approved):
Clean!

PR #9260, opened 2026-01-15
Chore: mcp prompt intervention

shijie-oai:
### Summary
Prompt intervention for MCP related queries (i.e. MCP connections, what MCPs do I have etc). 

Screenshot 2026-01-14 at 10 46 32 PM

### DIFF
--- a/codex-rs/core/gpt-5.1-codex-max_prompt.md
+++ b/codex-rs/core/gpt-5.1-codex-max_prompt.md
@@ -81,4 +81,6 @@ You are producing plain text that will later be styled by the CLI. Follow these 
 
 ## MCP servers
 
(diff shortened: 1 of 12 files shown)


codex-rs/core/gpt-5.1-codex-max_prompt.md
```diff
@@ -78,3 +78,7 @@ You are producing plain text that will later be styled by the CLI. Follow these
   * Do not use URIs like file://, vscode://, or https://.
   * Do not provide range of lines
   * Examples: src/app.ts, src/app.ts:42, b/server/index.js#L10, C:\repo\project\main.rs:12:5
+
+## MCP servers
+
+When asked what MCPs are available, check for tools with the `mcp__` prefix and report those back with the server names inferred from those tools.
```
gpeal: I think we can do better than this. You should say something like `MCP servers are accessed via tool calls. MCP Tools are provided to you via tool calls in the form of `mcp__{server_name}__{tool_name}`. MCP Resources can be read via `list_mcp_resources` (although maybe we should name this via `mcp__list_resources` or something.

It's also worth checking if there is an MCP server description that we're missing or could enumerate and provide to the model.

PR #9263, opened 2026-01-15
Record exec PIDs on existing tool-call spans and remove exec-specific OTEL events

marina-oai:
### Motivation
- Attach exec process identifiers (OS PID and process_id) to the existing tool-call tracing span instead of introducing separate exec-specific OTEL events so telemetry remains unified and simpler to maintain.
- Ensure PID metadata flows through unified exec / PTY / pipe plumbing so session logs, UI, and traces can reference the real OS process when available.

### Description
- Stop emitting dedicated `exec_command_begin`/`exec_command_end` OTEL events and instead record `process_id` and `os_pid` onto the current tool-call tracing span via a new `record_exec_span_metadata` helper (added in `core/src/tools/events.rs`).
- Propagated `os_pid` through the exec plumbing by adding an `os_pid: Option<u32>` field to raw/final exec outputs and types, capturing the child PID in `consume_truncated_output`, and threading it into `ExecToolCallOutput`, `RawExecToolCallOutput`, `ExecCommandBeginEvent`/`ExecCommandEndEvent`, `ExecCommandInput` and `ExecCommandResult`.
- Exposed OS PID on interactive processes by adding `os_pid` to `ProcessHandle` and passing the PID from PTY/pipe spawn sites (`utils/pty/*`), and threading `os_pid` through unified-exec manager, store and background watcher to emit it in events.
- Extended the tool-call tracing span (`dispatch_tool_call`) to include empty `process_id`/`os_pid` fields (`tracing::field::Empty`) so `record_exec_span_metadata` can record them at exec begin/end time.
- Updated protocol, UI test fixtures and many tests to include the new optional `os_pid` fields and adjusted event emission sites to call `record_exec_span_metadata` when exec begin/end events are emitted.

### Testing
- Ran formatting and autofixes with `just fmt` and `just fix -p codex-core` / `just fix -p codex-otel` (completed and applied Clippy autofixes).
- Ran `cargo test -p codex-core` where a large majority of tests passed but the test run included failures: `default_client::tests::test_create_client_sets_default_headers`, `models_manager::manager::tests::refresh_available_models_drops_removed_remote_models`, `models_manager::manager::tests::refresh_available_models_refetches_when_cache_stale`, `models_manager::manager::tests::refresh_available_models_sorts_and_marks_default`, `models_manager::manager::tests::refresh_available_models_uses_cache_when_fresh`, and `shell_snapshot::tests::linux_sh_snapshot_includes_sections` (failures reproduced in the run).
- Ran `cargo test -p codex-otel` where most tests passed but one test failed: `suite::otlp_http_loopback::otlp_http_exporter_sends_metrics_to_collector` reporting `ProviderShutdown { source: InternalFailure("Failed to flush") }`.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_i_69686f9bdcdc8322a2e9ddcd90671b12)


marina-oai:

marina-oai:
recheck

jif-oai:
@codex review

codex-rs/core/src/tools/events.rs
```diff
@@ -409,29 +432,45 @@ async fn emit_exec_end(
     exec_input: ExecCommandInput<'_>,
     exec_result: ExecCommandResult,
 ) {
+    let event = ExecCommandEndEvent {
```
jif-oai, line 435: Why those changes (here and above)
Why not having the clones inside record_exec_span_metadata and sending it some ref instead?

codex-rs/protocol/src/protocol.rs
```diff
@@ -1749,6 +1749,10 @@ pub struct ExecCommandBeginEvent {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     #[ts(optional)]
     pub process_id: Option<String>,
+    /// OS process identifier for the spawned command (when available).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub os_pid: Option<u32>,
```
jif-oai, line 1755: Are we sure we want to send this to the end user?
@owenlin0 and thoughts on this?

PR #9265, opened 2026-01-15
feat(linux-sandbox): add back bind mounts with a flag and gracefully fall back to landlock-only

viyatb-oai:
## What was happening

In containerized environments where namespaces are restricted (e.g., `Seccomp: 2`, no `CAP_SYS_ADMIN`, AppArmor), we saw:
- `unshare()` failing with `EPERM` (root and non-root paths)
- `/proc/self/setgroups` / `uid_map` / `gid_map` writes failing with permission errors after userns
- bind/remount operations failing with `EPERM`
- Landlock being unavailable or reporting “not enforced”

These failures were either aborting sandbox setup or being surfaced as “error running landlock” even when the failure occurred earlier.

## What this changes

- Bind-mount protections are now **experimental** behind `linux_sandbox_bind_mounts` and `--enable-bind-mounts`.
- When enabled, root still attempts `unshare(CLONE_NEWNS)` first, then falls back to `unshare(CLONE_NEWUSER | CLONE_NEWNS)` to regain mount privileges; non-root always uses userns+mountns.
- If any namespace/mount setup step is denied (including userns map writes, make-private, bind/remount), we **skip bind mounts** and continue with Landlock-only sandboxing (warn log).
- Landlock now **falls back to no-Landlock** when the kernel denies it or reports “not enforced”; seccomp/no_new_privs still apply (warn log).
- Errors are attributed by stage, so namespace/mount failures are no longer misreported as Landlock failures.

## Testing

- `just fix -p codex-linux-sandbox`


viyatb-oai:
Addressed review comments:

- https://github.com/openai/codex/pull/9265#discussion_r2693537147: test hook already short-circuits before real unshare when FORCE_UNSHARE_USERNS_SUCCESS is set, so hardened containers won’t execute the syscall in these tests.
- https://github.com/openai/codex/pull/9265#discussion_r2696055663: now drop caps after bind mounts even on the root+mountns path (used_userns || running_as_root).
- https://github.com/openai/codex/pull/9265#discussion_r2714140120: write_user_namespace_maps permission denied now returns an error (no Landlock-only fallback in an unmapped userns). Tests updated.
- https://github.com/openai/codex/pull/9265#discussion_r2714514823: should_skip_landlock now only skips on LandlockRestrict, so path permission errors surface as failures instead of disabling Landlock.

Latest fixes in commit b13f22975. cargo test -p codex-linux-sandbox passes locally.

codex-rs/core/src/exec.rs
bolinfest, line 935: In a follow-up, I think I'm going to change `process_exec_tool_call()` to take a struct because this is getting a bit hard to follow.

codex-rs/core/src/landlock.rs
```diff
@@ -13,19 +13,26 @@ use tokio::process::Child;
 /// helper accepts a list of `--sandbox-permission`/`-s` flags mirroring the
 /// public CLI. We convert the internal [`SandboxPolicy`] representation into
 /// the equivalent CLI options.
+#[allow(clippy::too_many_arguments)]
 pub async fn spawn_command_under_linux_sandbox<P>(
```
bolinfest, line 17: This one too. That's what the warning is for...

codex-rs/core/src/mcp/mod.rs
```diff
@@ -43,6 +44,7 @@ pub async fn collect_mcp_snapshot(config: &Config) -> McpListToolsResponseEvent
         sandbox_policy: SandboxPolicy::ReadOnly,
         codex_linux_sandbox_exe: config.codex_linux_sandbox_exe.clone(),
         sandbox_cwd: env::current_dir().unwrap_or_else(|_| PathBuf::from("/")),
+        linux_sandbox_bind_mounts: config.features.enabled(Feature::LinuxSandboxBindMounts),
```
bolinfest: I feel like we should be naming this `use_linux_sandbox_bind_mounts` throughout because the current name suggests the value is a list of bind mounts.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -1263,13 +1263,18 @@ impl CodexMessageProcessor {
         let outgoing = self.outgoing.clone();
         let req_id = request_id;
         let sandbox_cwd = self.config.cwd.clone();
+        let linux_sandbox_bind_mounts = self
```
bolinfest: Can we do a runtime test to verify that the host supports the API we need to use bind mounts in this way? If so, I think we could build better scaffolding around how we communicate support to users.

codex-rs/linux-sandbox/src/landlock.rs
```diff
@@ -54,6 +75,45 @@ pub(crate) fn apply_sandbox_policy_to_current_thread(
...
+}
+
+fn is_permission_denied(err: &CodexErr) -> bool {
+    if let CodexErr::Io(io_error) = err
+        && (io_error.kind() == std::io::ErrorKind::PermissionDenied
+            || io_error.raw_os_error() == Some(libc::EPERM))
+    {
+        return true;
+    }
+    let mut current: Option<&(dyn Error + 'static)> = Some(err);
```
bolinfest, line 91: nit: blank line after early return just to break this function up a bit

codex-rs/linux-sandbox/src/landlock.rs
```diff
@@ -54,6 +75,45 @@ pub(crate) fn apply_sandbox_policy_to_current_thread(
...
+            return true;
+        }
+        current = error.source();
+    }
+    false
+}
+
+fn sandbox_debug_enabled() -> bool {
+    matches!(
+        std::env::var("CODEX_SANDBOX_DEBUG").as_deref(),
```
bolinfest: I am generally wary of introducing support for new environment variables. What is this for? It should be documented as a top-level constant if we're going to have it.

codex-rs/linux-sandbox/src/landlock.rs
```diff
@@ -54,6 +75,45 @@ pub(crate) fn apply_sandbox_policy_to_current_thread(
...
+        Ok("1") | Ok("true") | Ok("TRUE")
+    )
+}
+
+fn log_landlock_fallback(err: &CodexErr) {
+    if sandbox_debug_enabled() {
+        eprintln!(
```
bolinfest: Hmm, we should have `#![deny(clippy::print_stdout, clippy::print_stderr)]` on this crate...

codex-rs/linux-sandbox/src/mounts.rs
```diff
@@ -1,8 +1,13 @@
 #![allow(dead_code)]
```
bolinfest: What is this allowing?

codex-rs/linux-sandbox/src/mounts.rs
```diff
@@ -130,6 +209,14 @@ fn unshare_mount_namespace() -> Result<()> {
...
 /// Unshare user + mount namespaces so the process can remount read-only without privileges.
 fn unshare_user_and_mount_namespaces() -> Result<()> {
+    #[cfg(test)]
+    if FORCE_UNSHARE_USERNS_SUCCESS.load(std::sync::atomic::Ordering::SeqCst) {
+        return Ok(());
+    }
+    #[cfg(test)]
+    if FORCE_UNSHARE_PERMISSION_DENIED.load(std::sync::atomic::Ordering::SeqCst) {
+        return Err(std::io::Error::from_raw_os_error(libc::EPERM).into());
+    }
```
bolinfest, line 324: Note you can use `#[cfg(test)]` on an anonymous block.

```suggestion
    #[cfg(test)]
    {
        if FORCE_UNSHARE_USERNS_SUCCESS.load(std::sync::atomic::Ordering::SeqCst) {
            return Ok(());
        }
        if FORCE_UNSHARE_PERMISSION_DENIED.load(std::sync::atomic::Ordering::SeqCst) {
            return Err(std::io::Error::from_raw_os_error(libc::EPERM).into());
        }
    }
```

codex-rs/linux-sandbox/src/mounts.rs
```diff
@@ -138,9 +225,100 @@ fn unshare_user_and_mount_namespaces() -> Result<()> {
...
+static FORCE_MAKE_MOUNTS_PRIVATE_PERMISSION_DENIED: AtomicBool = AtomicBool::new(false);
+#[cfg(test)]
+static FORCE_BIND_MOUNT_PERMISSION_DENIED: AtomicBool = AtomicBool::new(false);
+#[cfg(test)]
+static FORCE_DROP_CAPS_SUCCESS: AtomicBool = AtomicBool::new(false);
+#[cfg(test)]
+static FORCE_FLAGS_LOCK: Mutex<()> = Mutex::new(());
+
+#[cfg(test)]
+fn reset_force_flags() {
```
bolinfest, line 390: This can lead to race conditions in tests. Please reorganize to use pass-from-above.

Alternatively, use integration tests that spawn different instances of `codex-linux-sandbox` and then you can set flags/env vars for that individual invocation.

codex-rs/linux-sandbox/src/mounts.rs
```diff
@@ -138,9 +225,100 @@ fn unshare_user_and_mount_namespaces() -> Result<()> {
...
+#[cfg(test)]
+impl Drop for ForceFlagsGuard {
+    fn drop(&mut self) {
+        reset_force_flags();
+    }
+}
+
+#[cfg(test)]
+fn force_flags_guard() -> ForceFlagsGuard {
+    let guard = FORCE_FLAGS_LOCK.lock().expect("force flags lock");
```
bolinfest, line 419: Oh, so you're using this to prevent these tests from running in parallel?

I guess that also works (though it's slightly unfortunate, but at least this crate is pretty small).

If you stick with this, please document the strategy somewhere.

codex-rs/linux-sandbox/src/mounts.rs
```diff
@@ -138,9 +225,100 @@ fn unshare_user_and_mount_namespaces() -> Result<()> {
...
+    reset_force_flags();
+    ForceFlagsGuard(guard)
+}
+
+fn sandbox_debug_enabled() -> bool {
+    matches!(
+        std::env::var("CODEX_SANDBOX_DEBUG").as_deref(),
```
bolinfest: this is copy/pasted?

codex-rs/linux-sandbox/src/mounts.rs
```diff
@@ -138,9 +225,100 @@ fn unshare_user_and_mount_namespaces() -> Result<()> {
...
+        Ok("1") | Ok("true") | Ok("TRUE")
+    )
+}
+
+fn log_namespace_fallback(err: &CodexErr) {
```
bolinfest, line 424: and this?

codex-rs/linux-sandbox/src/mounts.rs
```diff
@@ -336,4 +538,135 @@ mod tests {
...
+        let sandbox_policy = SandboxPolicy::WorkspaceWrite {
+            writable_roots: vec![root],
+            network_access: false,
+            exclude_tmpdir_env_var: true,
+            exclude_slash_tmp: true,
+        };
+
+        let result = apply_read_only_mounts(&sandbox_policy, tempdir.path());
+
+        assert!(result.is_ok());
```
bolinfest: For some of these, can you add a second string literal argument that explains why this is expected? The test name suggests "permission denied" such that this should be `is_err()`.

bolinfest (commented):
Still very excited about this, but I have a number of questions!

codex-rs/linux-sandbox/src/mounts.rs
viyatb-oai: fixed

codex-rs/linux-sandbox/src/mounts.rs
viyatb-oai: yeah debugging artifcact, removed

codex-rs/linux-sandbox/src/mounts.rs
viyatb-oai: added

codex-rs/linux-sandbox/src/mounts.rs
viyatb-oai: added a comment

codex-rs/linux-sandbox/src/mounts.rs
viyatb-oai: added

codex-rs/linux-sandbox/src/landlock.rs
viyatb-oai: fixed

codex-rs/linux-sandbox/src/mounts.rs
viyatb-oai: old artifact from the last pr

codex-rs/linux-sandbox/src/landlock.rs
viyatb-oai: removed all print statements

codex-rs/core/src/mcp/mod.rs
viyatb-oai: agreed, changed!

codex-rs/core/src/features.rs
```diff
@@ -383,6 +385,16 @@ pub const FEATURES: &[FeatureSpec] = &[
...
     },
+    FeatureSpec {
+        id: Feature::LinuxSandboxBindMounts,
+        key: "use_linux_sandbox_bind_mounts",
+        stage: Stage::Experimental {
+            name: "Linux bind-mount sandbox",
+            menu_description: "Enable read-only bind mounts in the Linux sandbox for extra file-system protection.",
+            announcement: "NEW! Try Linux bind-mount sandboxing for stronger file-system protections. Enable in /experimental.",
+        },
+        default_enabled: false,
```
bolinfest, line 396: Hopefully we can default to `true` quickly...

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -23,11 +35,57 @@ pub fn run_main() -> ! {
...
+                std::process::exit(0);
+            }
+            BindMountProbeStatus::Unsupported { reason } => {
+                let _ = writeln!(
+                    std::io::stderr(),
+                    "codex-linux-sandbox: bind-mount probe unsupported: {reason}"
+                );
+                std::process::exit(1);
+            }
+        }
```
bolinfest: Just to be extra clear that this does not return, I would consider:

```suggestion
        let exit_code = match status {
            BindMountProbeStatus::Supported => {
                let _ = writeln!(
                    std::io::stderr(),
                    "codex-linux-sandbox: bind-mount probe supported"
                );
                0
            }
            BindMountProbeStatus::Unsupported { reason } => {
                let _ = writeln!(
                    std::io::stderr(),
                    "codex-linux-sandbox: bind-mount probe unsupported: {reason}"
                );
                1
            }
        };
        std::process::exit(exit_code);
```

In Rust, you can also have a function that declares its return type as `-> !` so the compiler guarantees it does not return.

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -23,11 +35,57 @@ pub fn run_main() -> ! {
...
         command,
     } = LandlockCommand::parse();
 
-    if let Err(e) = apply_sandbox_policy_to_current_thread(&sandbox_policy, &sandbox_policy_cwd) {
-        panic!("error running landlock: {e:?}");
+    if probe_bind_mounts_flag {
+        let status = probe_bind_mounts();
+        match status {
+            BindMountProbeStatus::Supported => {
+                let _ = writeln!(
```
bolinfest: You can do eprintln, you just have to do `#[allow(clippy::print_stderr)]`.

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -23,11 +35,57 @@ pub fn run_main() -> ! {
...
+                    std::io::stderr(),
+                    "codex-linux-sandbox: bind-mount probe unsupported: {reason}"
+                );
+                std::process::exit(1);
+            }
+        }
+    }
+
+    let sandbox_policy_cwd = sandbox_policy_cwd.unwrap_or_else(|| {
```
bolinfest: I think `expect()` would have the same effect here.

codex-rs/linux-sandbox/src/mounts.rs
```diff
@@ -24,28 +40,187 @@ pub(crate) fn apply_read_only_mounts(sandbox_policy: &SandboxPolicy, cwd: &Path)
...
+                // Root fallback: try userns+mountns to acquire mount powers.
+                let original_euid = unsafe { libc::geteuid() };
+                let original_egid = unsafe { libc::getegid() };
+                match unshare_user_and_mount_namespaces() {
+                    Ok(()) => {
+                        if let Err(err) = write_user_namespace_maps(original_euid, original_egid) {
+                            if is_permission_denied(&err) {
+                                return Err(err);
+                            }
+                            return Err(err);
```
bolinfest, line 59: Should we do something special when `is_permission_denied()` is `true`?

codex-rs/linux-sandbox/src/mounts.rs
```diff
@@ -24,28 +40,187 @@ pub(crate) fn apply_read_only_mounts(sandbox_policy: &SandboxPolicy, cwd: &Path)
...
+        }
     } else {
         let original_euid = unsafe { libc::geteuid() };
         let original_egid = unsafe { libc::getegid() };
-        unshare_user_and_mount_namespaces()?;
-        write_user_namespace_maps(original_euid, original_egid)?;
+        match unshare_user_and_mount_namespaces() {
+            Ok(()) => {
+                if let Err(err) = write_user_namespace_maps(original_euid, original_egid) {
+                    if is_permission_denied(&err) {
```
bolinfest, line 79: Same here.

codex-rs/linux-sandbox/src/mounts.rs
```diff
@@ -24,28 +40,187 @@ pub(crate) fn apply_read_only_mounts(sandbox_policy: &SandboxPolicy, cwd: &Path)
...
+                            if is_permission_denied(&err) {
+                                return Err(err);
+                            }
+                            return Err(err);
+                        }
+                        used_userns = true;
+                    }
+                    Err(err) if is_permission_denied(&err) => {
+                        // No namespaces available; continue with Landlock-only.
+                        log_namespace_fallback(&err);
```
bolinfest, line 65: We can maybe address this in a follow-up, but this is really NOT okay. Or at least, there should be a strict mode (that is the default) that hard fails in this case because we want to be able to rely on this as a security boundary. I feel like the user should have to pass a flag or somehow indicate that they are tolerant of the landlock-only solution and the risks that come with it.

codex-rs/linux-sandbox/src/linux_run_main.rs
viyatb-oai: oO, I didn't know this, changing this to exit code

codex-rs/linux-sandbox/src/linux_run_main.rs
viyatb-oai: added.

codex-rs/linux-sandbox/src/linux_run_main.rs
viyatb-oai: fixed.

codex-rs/linux-sandbox/src/mounts.rs
viyatb-oai: fixing - falling back to Landlock‑only (warn log) instead of erroring

codex-rs/linux-sandbox/src/mounts.rs
viyatb-oai: fixed.

codex-rs/linux-sandbox/src/mounts.rs
viyatb-oai: agreed, this makes it unpredicatable what is happening. I'll switch to this once we've had users test it out more.

PR #9275, opened 2026-01-15
feat: add agent roles to collab tools

jif-oai:
Add `agent_type` parameter to the collab tool `spawn_agent` that contains a preset to apply on the config when spawning this agent

### DIFF
--- a/codex-rs/core/src/agent/role.rs
+++ b/codex-rs/core/src/agent/role.rs
@@ -7,17 +7,14 @@ use serde::Serialize;
 const ORCHESTRATOR_PROMPT: &str = include_str!("../../templates/agents/orchestrator.md");
 /// Base instructions for the worker role.
 const WORKER_PROMPT: &str = include_str!("../../gpt-5.2-codex_prompt.md");
-/// Base instructions for the verifier role.
-const VERIFIER_PROMPT: &str = include_str!("../../review_prompt.md");
 /// Default worker model override used by the worker role.
 const WORKER_MODEL: &str = "gpt-5.2-codex";
 
 /// Enumerated list of all supported agent roles.
-const ALL_ROLES: [AgentRole; 4] = [
+const ALL_ROLES: [AgentRole; 3] = [
     AgentRole::Default,
     AgentRole::Orchestrator,
     AgentRole::Worker,
-    AgentRole::Verifier,
 ];
 
 /// Hard-coded agent role selection used when spawning sub-agents.
@@ -30,8 +27,6 @@ pub enum AgentRole {
     Orchestrator,
     /// Task-executing agent with a fixed model override.
     Worker,
-    /// Read-only reviewer that verifies outputs without editing.
-    Verifier,
 }
 
 /// Immutable profile data that drives per-agent configuration overrides.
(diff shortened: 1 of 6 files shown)


codex-rs/core/src/agent/role.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum AgentRole {
+    /// Inherit the parent agent's configuration unchanged.
+    Default,
+    /// Coordination-only agent that delegates to workers.
+    Orchestrator,
+    /// Task-executing agent with a fixed model override.
+    Worker,
+    /// Read-only reviewer that verifies outputs without editing.
```
gt-oai: OOC can verifiers run tools? If so, how are we distinguishing between read-only tools and write-tools. e.g. how about a formatter vs a linter vs a unit test? Maybe worth documenting the answer.

codex-rs/core/src/agent/role.rs
```diff
@@ -0,0 +1,95 @@
...
+
+/// Immutable profile data that drives per-agent configuration overrides.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct AgentProfile {
+    /// Optional base instructions override.
+    pub base_instructions: Option<&'static str>,
+    /// Optional model override.
+    pub model: Option<&'static str>,
+    /// Whether to force a read-only sandbox policy.
+    pub read_only: bool,
```
gt-oai, line 40: Should this be an enum rather than a bool to allow future nuance?

codex-rs/core/templates/agents/orchestrator.md
```diff
@@ -0,0 +1,38 @@
+You are Codex Orchestrator, based on GPT-5. You are running as an orchestration agent in the Codex CLI on a user's computer.
+Your name is batman
```
gt-oai: did batman eval better than the others?

codex-rs/core/templates/agents/orchestrator.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Role
+
+- You do not solve the task yourself. Your job is to delegate, coordinate, and verify.
```
gt-oai, line 5: Is verifying the verifier's job?

Maybe orchestrator + worker + verifier is too many. And actually you do just need orchestrator / critic + actor.

In general curious to learn how we arrived at these agents in particular (maybe offline if sensitive)

gt-oai (approved):

codex-rs/core/src/agent/role.rs
gt-oai: i guess the answer is the sandboxing policy

jif-oai: This uses our standard read-only sandboxing. But yes I'll document more

codex-rs/core/src/agent/role.rs
jif-oai: I'll consider. I'm not sure we want to over complexify this for now. I'm not even sure this read-only mode will be required

codex-rs/core/templates/agents/orchestrator.md
jif-oai: Used for prompt detection, will drop thanks

codex-rs/core/templates/agents/orchestrator.md
jif-oai: Let's chat offline yes. But this PR is mostly about the harness. Now we need to iterate to find the good pattern

PR #9276, opened 2026-01-15
feat: add interrupt capabilities to `send_input`

jif-oai:


codex-rs/core/src/tools/spec.rs
```diff
@@ -468,6 +468,15 @@ fn create_send_input_tool() -> ToolSpec {
             description: Some("Message to send to the agent.".to_string()),
         },
     );
+    properties.insert(
+        "interrupt".to_string(),
+        JsonSchema::Boolean {
+            description: Some(
+                "When true, interrupt the agent's current task before sending the message. Defaults to false."
```
gt-oai: Would be good to clarify the false behaviour. If the agent is already doing something, is the message sent after the current task? 

Also if people are likely to use this arg directly from Rust it would be nice to replicate the doc directly as a comment on `SendInputArgs`.

gt-oai (approved):

kornreaksmei-dev (commented):
[[Apple_Configurator_Logging_Instructions.pdf](https://github.com/user-attachments/files/24643204/Apple_Configurator_Logging_Instructions.pdf)]()

PR #9279, opened 2026-01-15
fix: update file search directory when session CWD changes

### LINKED ISSUE #9041: @filename lookup broken: “no matches” in trusted roots

  0.80.0

  Plus

  Codex (GPT-5) — issue not model-specific

  Darwin 24.6.0 arm64

  Summary
  After starting Codex in a non‑git folder and choosing “Require approval of edits and commands,” @filename
  stops working in trusted workspaces and returns “no matches.” This side effect is non‑obvious and appears
  unrelated to safety.

  1. Start Codex in a non‑git folder.
  2. Choose “Require approval of edits and commands.”
  3. Open a trusted workspace and use @filename.

  @filename resolves consistently in trusted workspaces.

  ### Additional information

  I tried removing the persisted trust entry in the TOML config; it sometimes helps after a terminal restart but
  not reliably. This suggests the issue isn’t fully resolved by deleting that entry alone.

yuvrajangadsingh:
## Summary

Fixes #9041

- Adds update_search_dir() method to FileSearchManager to allow updating the search directory after initialization
- Calls this method when the session CWD changes: new session, resume, or fork

## Problem

The FileSearchManager was created once with the initial search_dir and never updated. When a user:

1. Starts Codex in a non-git directory (e.g., /tmp/random)
2. Resumes or forks a session from a different workspace
3. The @filename lookup still searched the original directory

This caused no matches to be returned even when files existed in the current workspace.

## Solution

Update FileSearchManager.search_dir whenever the session working directory changes:
- AppEvent::NewSession: Use current config CWD
- SessionSelection::Resume: Use resumed session CWD
- SessionSelection::Fork: Use forked session CWD

## Test plan

- [ ] Start Codex in /tmp/test-dir (non-git)
- [ ] Resume a session from a project with actual files
- [ ] Verify @filename returns matches from the resumed session directory

### DIFF
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -1370,10 +1370,7 @@ impl App {
                                 self.shutdown_current_thread().await;
                                 self.config = resume_config;
                                 tui.set_notification_method(self.config.tui_notification_method);
-                                self.file_search = FileSearchManager::new(
-                                    self.config.cwd.clone(),
-                                    self.app_event_tx.clone(),
-                                );
+                                self.file_search.update_search_dir(self.config.cwd.clone());
                                 let init = self.chatwidget_init_for_forked_or_resumed_thread(
                                     tui,
                                     self.config.clone(),
--- a/codex-rs/tui/src/file_search.rs
+++ b/codex-rs/tui/src/file_search.rs
@@ -43,7 +43,7 @@ impl FileSearchManager {
     }
 
     /// Updates the directory used for file searches.
-    /// This should be called when the session's CWD changes (e.g., on new session, resume, or fork).
+    /// This should be called when the session's CWD changes on resume.
     /// Drops the current session so it will be recreated with the new directory on next query.
     pub fn update_search_dir(&mut self, new_dir: PathBuf) {
         self.search_dir = new_dir;
--- a/codex-rs/tui/src/file_search.rs
+++ b/codex-rs/tui/src/file_search.rs
@@ -42,6 +42,17 @@ impl FileSearchManager {
         }
     }
 
+    /// Updates the directory used for file searches.
+    /// This should be called when the session's CWD changes (e.g., on new session, resume, or fork).
+    /// Drops the current session so it will be recreated with the new directory on next query.
+    pub fn update_search_dir(&mut self, new_dir: PathBuf) {
+        self.search_dir = new_dir;
+        #[expect(clippy::unwrap_used)]
+        let mut st = self.state.lock().unwrap();
+        st.session.take();
+        st.latest_query.clear();
+    }
+
     /// Call whenever the user edits the `@` token.
     pub fn on_user_query(&self, query: String) {
         #[expect(clippy::unwrap_used)]


etraut-openai:
@codex review

etraut-openai:
@yuvrajangadsingh, thanks for the contribution. Please take a look at the code review comment from Codex.

yuvrajangadsingh:
Thanks for the review! I've addressed the feedback:

- `update_search_dir` now cancels any in-flight search via the cancellation token
- Clears `active_search`, `latest_query`, and `is_search_scheduled` to prevent stale results

This ensures old searches from the previous workspace can't pollute the new session.

yuvrajangadsingh:
@codex review

yuvrajangadsingh:
Addressed the second review comment as well:

**Fix**: Moved `search_dir` into `SearchState` (inside the mutex) so the debounce thread reads the **current** value when it fires, not a stale captured value.

Before: `search_dir` was captured at thread spawn time → race condition
After: `search_dir` read from mutex after debounce → always current

@codex review

yuvrajangadsingh:
Addressed the third round of feedback:

**Fix**: Skip search if `latest_query` is empty when debounce fires.

This handles the case where `update_search_dir()` cleared the state while the debounce thread was sleeping - prevents expensive match-all traversal.

@codex review

codex-rs/tui/src/app.rs
```diff
@@ -675,6 +675,8 @@ impl App {
                 };
                 self.chat_widget = ChatWidget::new(init, self.server.clone());
                 self.current_model = model_info.slug.clone();
+                // Update file search directory to match the current config's cwd.
```
etraut-openai: These changes need to be mirrored into the tui2 parallel directory. We're currently exploring a different tui implementation, so for the time being, we need to keep the two in sync.

etraut-openai (changes requested):

yuvrajangadsingh:
Mirrored the changes to tui2 as requested. Both tui and tui2 now have:

- `update_search_dir` method in `file_search.rs`
- `search_dir` stored inside `SearchState` (mutex-protected) to prevent race conditions
- Empty query check to avoid expensive match-all traversal
- `update_search_dir` calls in NewSession, Resume, and ForkCurrentSession handlers in `app.rs`

All tests pass for both packages.

yuvrajangadsingh:
@etraut-openai Changes mirrored to tui2 as requested. Ready for another look!

@codex review

yuvrajangadsingh:
Addressed the P3 feedback: added a generation counter to invalidate stale debounce threads after session changes.

**How it works:**
- `generation` field added to `SearchState`
- Incremented in `update_search_dir()` on session change
- Debounce threads capture generation at spawn time
- Before proceeding with search, check if generation matches - if not, abort silently

This prevents duplicate searches when user types a new query before the old debounce fires.

@etraut-openai @codex review

codex-rs/tui/src/file_search.rs
```diff
@@ -72,12 +75,30 @@ impl FileSearchManager {
...
+        let mut st = self.state.lock().unwrap();
+        st.search_dir = new_dir;
+        if let Some(active_search) = &st.active_search {
+            active_search
+                .cancellation_token
+                .store(true, Ordering::Relaxed);
+        }
+        st.active_search = None;
+        st.latest_query.clear();
+        st.is_search_scheduled = false;
```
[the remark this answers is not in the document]
yuvrajangadsingh: Addressed: Added a `generation` counter to `SearchState`. It's incremented in `update_search_dir()` and debounce threads check if generation matches before proceeding - if not, they abort silently.

codex-rs/tui2/src/file_search.rs
```diff
@@ -72,12 +75,30 @@ impl FileSearchManager {
...
+        let mut st = self.state.lock().unwrap();
+        st.search_dir = new_dir;
+        if let Some(active_search) = &st.active_search {
+            active_search
+                .cancellation_token
+                .store(true, Ordering::Relaxed);
+        }
+        st.active_search = None;
+        st.latest_query.clear();
+        st.is_search_scheduled = false;
```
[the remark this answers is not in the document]
yuvrajangadsingh: Addressed: Same fix applied to tui2 - added `generation` counter with the same behavior.

codex-rs/tui/src/app.rs
yuvrajangadsingh: Done! All changes mirrored to tui2:
- `file_search.rs`: `update_search_dir`, `search_dir` in mutex, empty query check, generation counter
- `app.rs`: `update_search_dir` calls in NewSession, Resume, and ForkCurrentSession handlers

yuvrajangadsingh:
Rebased on latest main to resolve conflicts.

@codex review

yuvrajangadsingh:
@etraut-openai All changes addressed - generation counter added to prevent duplicate debounce searches, mirrored to tui2. Conflicts resolved, CI passing, Codex review clean. Ready for final review.

yuvrajangadsingh:
Rebased again on latest main to resolve new conflicts.

@codex review

yuvrajangadsingh:
Rebased on latest main - tui2 was removed upstream, changes now only apply to tui.

@codex review

yuvrajangadsingh:
@codex review

etraut-openai (approved):

codex-rs/tui/src/file_search.rs
```diff
@@ -42,6 +42,17 @@ impl FileSearchManager {
         }
     }
 
+    /// Updates the directory used for file searches.
+    /// This should be called when the session's CWD changes (e.g., on new session, resume, or fork).
```
etraut-openai: This comment looks wrong. I think this should be called only on resume. For new sessions, the FileSearchManager is created with the correct cwd from the start. For forks, the cwd doesn't change.

etraut-openai (changes requested):
Looking pretty good. One small change request.

codex-rs/tui/src/file_search.rs
yuvrajangadsingh: Fixed. Updated the docstring to only mention resume.

yuvrajangadsingh:
@etraut-openai Fixed the docstring - now only mentions resume. Ready for another look.

etraut-openai (approved):

PR #9293, opened 2026-01-15
[codex-tui] exit when terminal is dumb

jmickey-oai:
Using terminal with TERM=dumb specifically mean that TUIs and the like don't work. Ensure that codex doesn't run in these environments and exit with odd errors like crossterm's "Error: The cursor position could not be read within a normal duration"


jmickey-oai:

joshka-oai (approved):
Added some bits to this:
- confirmation prompt to continue anyway
- added dumb to the term detection code

jmickey-oai:
thanks for carrying this forward with a far better change

Issue #9292, opened 2026-01-15
Linux: codex-cli 0.81+ sandbox fails with landlock; 0.79/0.80 ignore --sandbox flags


vincentsch:
## Summary
On Ubuntu 24.04.3 (Xubuntu), codex-cli 0.81+ fails any sandboxed run with:

```
error running landlock: Io(Os { code: 1, kind: PermissionDenied, message: "Operation not permitted" })
```

Only `--dangerously-bypass-approvals-and-sandbox` works.

Additionally, in 0.79/0.80 the `--sandbox` flag appears ignored (acts like `workspace-write` even when `read-only` or `danger-full-access` is passed).

## Environment
- OS: Ubuntu 24.04.3 LTS (Xubuntu)
- Kernel: 6.14.0-37-generic
- AppArmor enabled; `kernel.apparmor_restrict_unprivileged_userns=1`
- `/sys/kernel/security/landlock` is missing
- Install method: `npm install -g @openai/codex@<version>`

## Minimal repro (0.81+)
```
codex exec --sandbox workspace-write \
  -c model=gpt-5.2-codex -c reasoning_effort=xhigh -c max_tokens=8000 \
  --full-auto "Run 'true' using exec."
```

### Expected
- `true` executes successfully inside sandbox.

### Actual
- Landlock setup fails with `Operation not permitted` before command runs.

## Version matrix (tested)
| Version | Result (sandboxed run) | Notes |
|---|---|---|
| 0.79.0 | Works | `--sandbox read-only` still allows `/tmp` writes; `--sandbox danger-full-access` still blocks writes outside workspace (flag ignored). |
| 0.80.0 | Works | Same behavior as 0.79.0. |
| 0.81.0 | Fails | Landlock permission denied. |
| 0.84.0 | Fails | Landlock permission denied. |

## Additional observations
- The CLI header always reports `sandbox: workspace-write` regardless of `--sandbox` value (0.79/0.80).
- `--dangerously-bypass-approvals-and-sandbox` works on 0.81/0.84.

## Workaround
- Pin to `@openai/codex@0.80.0`, or bypass sandbox (not ideal).

Happy to run more diagnostics if you have specific data to collect.

vincentsch:
Opened PR #9295 with a graceful fallback: if unshare/uid_map/mount steps fail with EPERM, skip the new read-only bind mounts and continue with Landlock-only enforcement (pre-0.81 behavior). Tests: cargo test -p codex-linux-sandbox.

[...]

PR #9295, opened 2026-01-15
Linux sandbox: gracefully skip read-only mounts when unshare is denied

vincentsch:
## What
- Skip read-only bind-mount setup when user/mount namespace operations fail with EPERM, falling back to Landlock-only enforcement.
- Add a small unit test for permission-denied detection.

## Why
On Ubuntu 24.04 with `kernel.apparmor_restrict_unprivileged_userns=1`, codex-cli 0.81+ fails to start any sandboxed command because the new read-only mount step (`unshare(CLONE_NEWUSER|CLONE_NEWNS)` + uid/gid maps) returns EPERM. This change restores pre-0.81 behavior in that case (Landlock-only sandbox), so sandboxed runs continue instead of crashing.

## How
Treat PermissionDenied from unshare/mount/uid_map steps as non-fatal and continue without read-only bind mounts.

Issue: #9292

## Tests
- `cargo test -p codex-linux-sandbox` (Rust 1.90 via rustup).
- `just` not available locally.

### DIFF
--- a/codex-rs/linux-sandbox/src/mounts.rs
+++ b/codex-rs/linux-sandbox/src/mounts.rs
@@ -27,14 +27,37 @@ pub(crate) fn apply_read_only_mounts(sandbox_policy: &SandboxPolicy, cwd: &Path)
     } else {
         let original_euid = unsafe { libc::geteuid() };
         let original_egid = unsafe { libc::getegid() };
-        unshare_user_and_mount_namespaces()?;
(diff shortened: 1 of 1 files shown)


vincentsch:

etraut-openai:
Thanks for the contribution. This overlaps an in-flight PR #9265, so we'll pass on this one.

PR #9304, opened 2026-01-15
nit: clean unified exec background processes

jif-oai:
To fix the occurences where the End event is received after the listener stopped listenning


tavaresgmg (commented):
Reviewed the changes. The introduction of  improves code reuse and ensures that stale process states are correctly cleared both on turn start and abort. The added test case  effectively verifies the fix. LGTM!

pakrym-oai (approved):

PR #9330, opened 2026-01-15
feat: show forked from session id in /status

apanasenko-oai:
Summary:
- Add forked_from to SessionMeta/SessionConfiguredEvent and persist it for forked sessions.
- Surface forked_from in /status for tui + tui2 and add snapshots.

### DIFF
--- a/codex-rs/app-server/tests/common/rollout.rs
+++ b/codex-rs/app-server/tests/common/rollout.rs
@@ -52,6 +52,7 @@ pub fn create_fake_rollout(
     // Build JSONL lines
     let meta = SessionMeta {
         id: conversation_id,
+        forked_from_id: None,
         timestamp: meta_rfc3339.to_string(),
         cwd: PathBuf::from("/"),
         originator: "codex".to_string(),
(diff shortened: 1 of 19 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -587,6 +586,18 @@ impl Session {
             ));
         }
 
+        let forked_from = match &initial_history {
```
aibrahim-oai: let's put it as a method on initial history maybe or a separate method at least

aibrahim-oai (approved):

codex-rs/app-server/tests/common/rollout.rs
```diff
@@ -42,6 +42,7 @@ pub fn create_fake_rollout(
     // Build JSONL lines
     let meta = SessionMeta {
         id: conversation_id,
+        forked_from: None,
```
pap-openai: `forked_from_id` maybe ? (considering we might have some other identifiers such as names – not sure if we'll need this though)

jif-oai (approved):
lgtm after the comments of the others

PR #9331, opened 2026-01-15
Add text element metadata to protocol, app server, and core

charley-oai:
The second part of breaking up PR https://github.com/openai/codex/pull/9116

Summary:

- Add `TextElement` / `ByteRange` to protocol user inputs and user message events with defaults.
- Thread `text_elements` through app-server v1/v2 request handling and history rebuild.
- Preserve UI metadata only in user input/events (not `ContentItem`) while keeping local image attachments in user events for rehydration.

Details:

- Protocol: `UserInput::Text` carries `text_elements`; `UserMessageEvent` carries `text_elements` + `local_images`. Serialization includes empty vectors for backward compatibility.
- app-server-protocol: v1 defines `V1TextElement` / `V1ByteRange` in camelCase with conversions; v2 uses its own camelCase wrapper.
- app-server: v1/v2 input mapping includes `text_elements`; thread history rebuilds include them.
- Core: user event emission preserves UI metadata while model history stays clean; history replay round-trips the metadata.


codex-rs/protocol/src/items.rs
```diff
@@ -83,6 +85,32 @@ impl UserMessageItem {
...
+                            end: offset + elem.byte_range.end,
+                        },
+                        placeholder: elem.placeholder.clone(),
+                    });
+                }
+                offset += text.len();
+            }
+        }
+        out
+    }
```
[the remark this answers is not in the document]
charley-oai: In this context it’s intentionally a straight rebase:

  - text_elements are produced by our own textarea logic and are always chunk‑relative and within that chunk’s byte length. We don’t accept arbitrary
    client‑provided ranges here (this is the legacy message composition inside core), so we don’t need to defensively clamp or reorder.
  - Overlaps or out‑of‑order ranges are allowed; they just describe multiple UI spans. The consumer renders spans in order, not as “non‑overlapping ranges,” so
    overlaps aren’t inherently invalid.

codex-rs/core/src/codex.rs
```diff
@@ -2563,9 +2564,18 @@ pub(crate) async fn run_turn(
...
     }
 
+    let user_message_item = UserMessageItem::new(&input);
     let initial_input_for_turn: ResponseInputItem = ResponseInputItem::from(input);
     let response_item: ResponseItem = initial_input_for_turn.clone().into();
-    sess.record_response_item_and_emit_turn_item(turn_context.as_ref(), response_item)
+    // Persist the user message to history, but emit the turn item from `UserInput` so
+    // UI-only `text_elements` are preserved. `ResponseItem::Message` does not carry
+    // those spans, and `record_response_item_and_emit_turn_item` would drop them.
+    sess.record_conversation_items(turn_context.as_ref(), &[response_item])
```
aibrahim-oai: why don't we change `parse_turn_item` behavior inside?

charley-oai: `parse_turn_item` only sees a `ResponseItem`, and `ResponseItem::Message` doesn’t carry `text_elements` or `local_images`. So even if we change `parse_turn_item`, it still
wouldn’t have the data needed to preserve UI-only spans unless we either:

1. expand `ResponseItem::Message` to carry those fields (which we explicitly avoided to keep API content clean), or
2. change `parse_turn_item` to accept additional context (e.g., the original `UserInput`) and thread that through its callers.

In this spot we do have the UserInput, so emitting `TurnItem::UserMessage` directly is the minimal, local fix without broad API changes. If you’d prefer the
architectural change, we can refactor `parse_turn_item` to take `Option<&UserInput>` (or similar) and update its call sites, but that’s a larger change and would
touch more of the codebase.

aibrahim-oai: hmm that makes sense. 

`record_response_item_and_emit_turn_item` to be used only for user message and assistant message. 

Let's make a new function just for recording prompt so it's easier to follow.

aibrahim-oai (commented):
Very clean! Let's add integration tests testing that we send turn item and event msgs and also we keep them on resume.

aibrahim-oai (approved):
Awesome! I'd also add tests in app-server

PR #9333, opened 2026-01-16
[device-auth] Add device code auth as a standalone option when headless environment is detected.

mzeng-openai:
Previously device code replaces the standard "Sign in with ChatGPT" option when headless environment is detected. This prevents certain users who's normal dev setup is remote devboxes. So we are move it to be an additional option alongside the standard browser flow instead of replacing it.

- [x] Add device code auth as a standalone option when headless environment is detected.

### DIFF
--- a/codex-rs/tui/src/onboarding/auth.rs
+++ b/codex-rs/tui/src/onboarding/auth.rs
@@ -5,7 +5,6 @@ use codex_core::auth::AuthCredentialsStoreMode;
 use codex_core::auth::CLIENT_ID;
 use codex_core::auth::login_with_api_key;
 use codex_core::auth::read_openai_api_key_from_env;
-use codex_core::env::is_headless_environment;
 use codex_login::DeviceCode;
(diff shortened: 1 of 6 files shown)


codex-rs/tui/src/onboarding/auth.rs
```diff
@@ -588,13 +660,21 @@ impl AuthModeWidget {
         self.request_frame.schedule_frame();
     }
 
+    fn handle_existing_chatgpt_login(&mut self) -> bool {
```
joshka-oai, line 663: nit: A bool return value isn't really all that useful without documenting its meaning.

joshka-oai (approved):
Can this be tested simply? (e.g. Codex "add unit tests that use insta to check the output of these screens for the various combinations. Ideally these tests should be phrased as simple explanations of the behavior").

I suspect the animation may make this non-trivial, but there's a way to ignore parts of an output in insta (IIRC).

PR #9336, opened 2026-01-16
rename model turn to sampling request

aibrahim-oai:
We have two type of turns now: model and user turns. It's always confusing to refer to either. Model turn is basically a sampling request.


aibrahim-oai:
@codex review this

jif-oai (approved):

PR #9340, opened 2026-01-16
Introduce collaboration modes

aibrahim-oai:
- Merge `model` and `reasoning_effort` under collaboration modes.
- Add additional instructions for custom collaboration mode
- Default to Custom to not change behavior


aibrahim-oai:
@codex review this

codex-rs/protocol/src/config_types.rs
```diff
@@ -127,3 +129,78 @@ pub enum AltScreenMode {
...
+
+/// Collaboration mode for a Codex session.
+#[derive(Clone, PartialEq, Eq, Hash, Debug, Serialize, Deserialize, JsonSchema, TS)]
+#[serde(tag = "mode", rename_all = "lowercase")]
+pub enum CollaborationMode {
+    Plan(Settings),
+    Collaborate(Settings),
+    Execute(Settings),
+    Custom(CustomSettings),
+}
```
dylan-hurd-oai, line 141: I think we need a more flexible structure here - how will this handle multiple user-defined modes?

codex-rs/protocol/src/config_types.rs
```diff
@@ -127,3 +129,78 @@ pub enum AltScreenMode {
...
+    pub model: String,
+    pub reasoning_effort: Option<ReasoningEffort>,
+}
+
+/// Custom settings that extend base settings with developer instructions.
+#[derive(Clone, PartialEq, Eq, Hash, Debug, Serialize, Deserialize, JsonSchema, TS)]
+pub struct CustomSettings {
+    pub settings: Settings,
+    pub developer_instructions: Option<String>,
+}
```
dylan-hurd-oai: `Settings` and `CustomSettings` feels like a needless distinction. Why wouldn't we support developer_instructions in Settings, if we want to configure how Plan mode works?

codex-rs/protocol/src/config_types.rs
```diff
@@ -127,3 +129,78 @@ pub enum AltScreenMode {
...
+        }
+    }
+}
+
+/// Settings for a collaboration mode.
+#[derive(Clone, PartialEq, Eq, Hash, Debug, Serialize, Deserialize, JsonSchema, TS)]
+pub struct Settings {
+    pub model: String,
+    pub reasoning_effort: Option<ReasoningEffort>,
+}
```
dylan-hurd-oai: (not hard-blocking but important IMO) This feels like we're going to end up duplicating most of `ConfigProfile` here - I wonder if we should just reuse / copy that approach

codex-rs/protocol/src/config_types.rs
aibrahim-oai: using `Custom` on `op::userturn`

codex-rs/protocol/src/config_types.rs
aibrahim-oai: Because `plan` mode should work universally through the agent not through the client deciding how it should work. Also we will need to send all this information to the client to configure it.

codex-rs/protocol/src/config_types.rs
aibrahim-oai: I wouldn't interconnect them for now at least. profile is larger surface area and not very well maintained feature

codex-rs/protocol/src/config_types.rs
dylan-hurd-oai: > Because plan mode should work universally through the agent not through the client deciding how it should work. Also we will need to send all this information to the client to configure it.

Both of these are true but I think that's a bit orthogonal to the Settings structure here

aibrahim-oai: hmm ok good point

dylan-hurd-oai (approved):

PR #9366, opened 2026-01-16
feat: CODEX_CI

jif-oai:


codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -57,6 +57,7 @@ const UNIFIED_EXEC_ENV: [(&str, &str); 9] = [
     ("PAGER", "cat"),
     ("GIT_PAGER", "cat"),
     ("GH_PAGER", "cat"),
+    ("CODEX_CI", "1"),
```
gt-oai, line 60: what's the effect of this? I don't see `CODEX_CI` anywhere else in the codebase? worth commenting I think

gt-oai (approved):

codex-rs/core/src/unified_exec/process_manager.rs
jif-oai: People asked for is. Will document

PR #9374, opened 2026-01-16
nit: do not render terminal interactions if no task running

jif-oai:
To prevent race where the terminal interaction message is processed after the last message


joshka-oai (approved):
Does this need a tui2 equivalent?
I've seen this problem there and wondered what the deal was.

joshka-oai:
> Does this need a tui2 equivalent? I've seen this problem there and wondered what the deal was.

To answer my own question. No as this is not presently implemented in tui2.

Issue #9081, opened 2026-01-12
Windows Terminal/cmd/pwsh: cursor blink speeds up and jumps during Codex TUI redraw


YcccByHe:
### What version of Codex is running?

codex-cli 0.80.0

### What subscription do you have?

ChatGPT Plus

### Which model were you using?

gpt-5.2

### What platform is your computer?

Microsoft Windows NT 10.0.26200.0 x64

### What issue are you seeing?

  On Windows Terminal / cmd.exe / PowerShell (pwsh), while Codex CLI is waiting/streaming and the TUI redraws
  frequently, the terminal cursor blink rate (and intensity animation) becomes abnormally fast and the cursor visibly
  “jumps/sweeps” during redraws.

  This makes the UI distracting and hard to use. It does not reproduce in IntelliJ IDEA’s built-in terminal.

### What steps can reproduce the bug?

  1. Open Windows Terminal (also reproduces in cmd.exe and pwsh).
  2. Run `codex`.
  3. Trigger a redraw-heavy state (e.g. any waiting/streaming state such as a long-running tool call / background exec).
  4. Observe the cursor while the status line/spinner updates.

  Expected: cursor blink rate stays consistent with terminal defaults; cursor does not visibly jump.
  Actual: cursor blink speeds up dramatically and cursor position visibly jumps/sweeps during redraws.

  Evidence: MP4 recording attached.

https://github.com/user-attachments/assets/023d43ec-6e46-4942-95f8-7a4cfbe11f2e

etraut-openai:
This is an unfortunate behavior of the PowerShell terminal. I'm not sure that there's anything we can do about it given the way the Codex TUI works.

YcccByHe:
> This is an unfortunate behavior of the PowerShell terminal. I'm not sure that there's anything we can do about it given the way the Codex TUI works.
https://github.com/openai/codex/pull/9082

etraut-openai:
@YcccByHe, I saw your PR after I wrote that comment. Looks like you've proven me wrong. :)

YcccByHe:
@etraut-openai ,
Thanks for the quick follow-up!

  Agreed it looked terminal-specific at first. The fix in #9082 keeps the real terminal cursor hidden and renders a non-
  blinking “fake cursor” in the in-memory buffer (reverse-video on the cursor cell), which avoids the visible cursor
  jumping / accelerated blinking during redraws on Windows Terminal / cmd.exe / pwsh.

[...]

PR #9379, opened 2026-01-16
fix(tui): note Windows Terminal <1.24 flicker

joshka-oai:
Windows Terminal versions earlier than 1.24 don't support Synchronized Update (DEC Mode 2026), which can cause Codex output to flicker.

This adds a non-error startup tip when we detect Windows Terminal <1.24 and suggests upgrading (including the Preview channel).

Detection reads the installed Windows Terminal version via PowerShell `Get-AppxPackage` (stable + Preview, picking the newest).

Refs: https://github.com/openai/codex/issues/9081
Release: https://github.com/microsoft/terminal/releases/tag/v1.24.2372.0

### DIFF
--- a/codex-rs/core/src/terminal.rs
+++ b/codex-rs/core/src/terminal.rs
@@ -172,6 +172,18 @@ impl TerminalInfo {
 
         sanitize_header_value(raw)
     }
+
+    /// Returns the (major, minor) Windows Terminal version when available.
+    ///
+    /// This parses the first two numeric components from the detected version string
+    /// (e.g. `"1.24.2372.0"` -> `(1, 24)`).
+    pub fn windows_terminal_major_minor(&self) -> Option<(u64, u64)> {
+        if self.name != TerminalName::WindowsTerminal {
+            return None;
+        }
+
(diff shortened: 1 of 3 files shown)


codex-rs/core/src/terminal.rs
```diff
@@ -172,6 +172,18 @@ impl TerminalInfo {
...
+    ///
+    /// This parses the first two numeric components from the detected version string
+    /// (e.g. `"1.24.2372.0"` -> `(1, 24)`).
+    pub fn windows_terminal_major_minor(&self) -> Option<(u64, u64)> {
+        if self.name != TerminalName::WindowsTerminal {
+            return None;
+        }
+
+        self.version.as_deref().and_then(parse_major_minor_version)
+    }
```
joshka-oai, line 186: note that this doesn't actually work - needs to use powershell to check the version

joshka-oai:
Closing this out for now - we can redo this if we need it / get more reports of this problem. Otherwise it's reasonable for people to find the linked issues and self-update to the 1.24 pre-release without having to be prompted by us to do this.

PR #9388, opened 2026-01-16
fix(shell-snapshot): detach snapshot from tty

kfiramar:
## Summary
- fix a macOS Terminal.app input regression when shell_snapshot is enabled by isolating the snapshot subprocess from the interactive TTY
- prevent snapshot shells/.zshrc from flipping terminal modes while the TUI is in raw mode
- add CODEX_SHELL_SNAPSHOT so users can guard tty-touching rc logic if needed

## Bug Details
When shell_snapshot is enabled, Codex spawns a login shell (zsh -lc) and sources ~/.zshrc to capture aliases/functions/env. On some setups (e.g., Terminal.app + a .zshrc that runs stty/tput or emits escape sequences), the snapshot shell writes to the live TTY and changes terminal modes. While the Codex TUI is in raw mode, this breaks input handling: Enter/backspace stop working and raw escape sequences like ^[[O appear (often after alt-tab). Disabling shell_snapshot avoids the issue.

## Repro Steps (macOS)
1. Enable `shell_snapshot` (TUI `/experimental` → “Shell snapshot”, or set `[features].shell_snapshot = true` in `~/.codex/config.toml` and restart).
2. Create a minimal zshrc that touches the tty:

   ```bash
   mkdir -p /tmp/codex-tty-repro
   cat > /tmp/codex-tty-repro/.zshrc <<'EOF'
   stty sane 2>/dev/null
   EOF
   ```
3. Launch Codex with that rc:

   ```bash
   ZDOTDIR=/tmp/codex-tty-repro codex
   ```
4. When the TUI appears, input is broken (Enter/Backspace don’t work). After alt‑tabbing, you may see raw escape sequences like `^[[O`.
5. Confirm that disabling snapshot restores input:

   ```bash
   ZDOTDIR=/tmp/codex-tty-repro codex --disable shell_snapshot
   ```

## Why It Failed Before
The snapshot subprocess inherited the controlling TTY and ran `.zshrc` while the TUI was in raw mode. Any tty mode changes or escape output from the snapshot shell hit the live terminal, which flips the terminal state out of raw mode and effectively “breaks” input handling for the running TUI. The failure is config-dependent, so it only shows up when shell_snapshot is enabled and `.zshrc` (or a plugin) touches the TTY.

## Fix
Detach the snapshot subprocess from the controlling terminal and avoid inheriting stdin:
- stdin set to /dev/null
- setsid() before exec on Unix
- env CODEX_SHELL_SNAPSHOT=1 for rc guards

[...]

PR #9403, opened 2026-01-17
Detach non-tty child spawns

frantic-openai:
## Why

Sometimes codex gets into a bad state when running scripts. I see `zsh: suspended (tty input)` and `fg` will crash shell/terminal emulator. I can repro this reliably on iTerm2, Ghostty, inside and outside tmux etc.

After trial and error codex (haha) suggested this is related to the way child process handles tty. Either way bad programs shouldn't crash codex.

## Repro

```python
#!/usr/bin/env python
import os
import signal
import time

tty_fd = os.open('/dev/tty', os.O_RDWR)
time.sleep(2)
os.killpg(os.tcgetpgrp(tty_fd), signal.SIGTTIN)
```

## Summary
- gate non-PTY detaching *behind the experimental flag* `detach_non_tty` (default off)
- thread detach flag through exec/unified exec/app server/user shell paths; only applies to non-tty + no-sandbox spawns
- update config schema for the new feature flag
- silence unused detach flag warning on non-unix in utils/pty
- silence unused detach flag warning on non-unix in core spawn path

## Testing
- just fmt
- just fix -p codex-core
- just fix -p codex-app-server
- just fix -p codex-exec-server
- just fix -p codex-linux-sandbox
- just fix -p codex-utils-pty
- cargo test -p codex-core
- cargo test -p codex-exec-server
- cargo test -p codex-linux-sandbox
- cargo test -p codex-utils-pty
- cargo test --all-features
- cargo build -p codex-cli
- just write-config-schema

### DIFF
--- a/codex-rs/app-server-protocol/src/protocol/v2.rs
+++ b/codex-rs/app-server-protocol/src/protocol/v2.rs
@@ -512,17 +512,22 @@ pub struct ConfigEdit {
 #[ts(export_to = "v2/")]
 pub enum CommandExecutionApprovalDecision {
     /// User approved the command.
+    #[serde(alias = "approved")]
     Accept,
     /// User approved the command and future identical commands should run without prompting.
+    #[serde(alias = "approved_for_session")]
     AcceptForSession,
     /// User approved the command, and wants to apply the proposed execpolicy amendment so future
     /// matching commands can run without prompting.
+    #[serde(alias = "approved_with_amendment")]
     AcceptWithExecpolicyAmendment {
         execpolicy_amendment: ExecPolicyAmendment,
     },
     /// User denied the command. The agent will continue the turn.
+    #[serde(alias = "denied")]
     Decline,
     /// User denied the command. The turn will also be immediately interrupted.
+    #[serde(alias = "abort")]
     Cancel,
 }
 
@@ -531,12 +536,16 @@ pub enum CommandExecutionApprovalDecision {
 #[ts(export_to = "v2/")]
 pub enum FileChangeApprovalDecision {
     /// User approved the file changes.
+    #[serde(alias = "approved")]
     Accept,
     /// User approved the file changes and future changes to the same files should run without prompting.
+    #[serde(alias = "approved_for_session")]
     AcceptForSession,
     /// User denied the file changes. The agent will continue the turn.
+    #[serde(alias = "denied")]
     Decline,
     /// User denied the file changes. The turn will also be immediately interrupted.
+    #[serde(alias = "abort")]
     Cancel,
 }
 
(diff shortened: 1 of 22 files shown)


frantic-openai:

etraut-openai:
Is this related to https://github.com/openai/codex/pull/9388?

bolinfest:
Hi there @frantic-openai ! Can you please flesh out the PR body and take a look at the failing CI jobs?

frantic-openai:
> Is this related to https://github.com/openai/codex/pull/9388?

The issue this PR attempts to fix doesn't depend on the "Shell snapshot" feature on or off, but the root cause is probably all the same

codex-rs/app-server-protocol/src/protocol/v2.rs
joshka-oai, line 1: This seems like it's possible an unrelated change - guessing that something changed in the app server underneath this change and codex worked around that?

Do you have any context for this file change?

codex-rs/core/src/features.rs
```diff
@@ -358,6 +360,16 @@ pub const FEATURES: &[FeatureSpec] = &[
         },
         default_enabled: false,
     },
+    FeatureSpec {
+        id: Feature::DetachNonTty,
+        key: "detach_non_tty",
+        stage: Stage::Beta {
+            name: "Detach non-PTY commands",
+            menu_description: "Run non-interactive commands without inheriting the controlling TTY.",
+            announcement: "NEW! Try detaching non-PTY commands to avoid TTY issues. Enable in /experimental!",
```
joshka-oai, line 369: :)

kfiramar:
@etraut-openai This looks related to #9388 at the root-cause level (child procs touching the controlling TTY), but it’s not overlapping.
#9388 isolates the shell_snapshot subprocess, while this PR gates detaching non‑PTY spawns for exec/unified‑exec.

codex-rs/core/src/features.rs
```diff
@@ -358,6 +360,16 @@ pub const FEATURES: &[FeatureSpec] = &[
         },
         default_enabled: false,
     },
+    FeatureSpec {
+        id: Feature::DetachNonTty,
+        key: "detach_non_tty",
+        stage: Stage::Beta {
```
jif-oai, line 366: Nope, put it as `Stage::Experimental` and default to true, this is already behind the unified_exec flag

jif-oai (commented):
I would actually drop the whole wiring and the feature flag and default it to true. Unified exec is still experimental so no need to hide this behind a flag + the full wiring is a bit messy TBH

frantic-openai:
Fixed via https://github.com/openai/codex/pull/9477

PR #9427, opened 2026-01-18
fix(core) Preserve base_instructions in SessionMeta

dylan-hurd-oai:
## Summary
This PR consolidates base_instructions onto SessionMeta / SessionConfiguration, so we ensure `base_instructions` is set once per session and should be (mostly) immutable, unless:
- overridden by config on resume / fork
- sub-agent tasks, like review or collab

In a future PR, we should convert all references to `base_instructions` to consistently used the typed struct, so it's less likely that we put other strings there. See #9423. However, this PR is already quite complex, so I'm deferring that to a follow-up.

## Testing
- [x] Added a resume test to assert that instructions are preserved. In particular, `resume_switches_models_preserves_base_instructions` fails against main.

Existing test coverage thats assert base instructions are preserved across multiple requests in a session:
  - Manual compact keeps baseline instructions: core/tests/suite/compact.rs:199
  - Auto-compact keeps baseline instructions: core/tests/suite/compact.rs:1142
  - Prompt caching reuses the same instructions across two requests: core/tests/suite/prompt_caching.rs:150 and core/tests/suite/prompt_caching.rs:157
  - Prompt caching with explicit expected string across two requests: core/tests/suite/prompt_caching.rs:213 and core/tests/suite/prompt_caching.rs:222
  - Resume with model switch keeps original instructions: core/tests/suite/resume.rs:136
  - Compact/resume/fork uses request 0 instructions for later expected payloads: core/tests/suite/compact_resume_fork.rs:215


aibrahim-oai (commented):
clean!

aibrahim-oai:
@codex review this

dylan-hurd-oai:
Looks like bazel build is failing due to cross-crate import, will duplicate across before merging

codex-rs/core/src/config/mod.rs
```diff
@@ -1611,6 +1613,14 @@ impl Config {
         Ok(config)
     }
 
+    pub fn collaboration_mode(&self, model: String) -> CollaborationMode {
```
jif-oai: What is this and how is this related?

codex-rs/core/src/rollout/tests.rs
```diff
@@ -89,6 +89,7 @@ fn write_session_file_with_provider(
         "cwd": ".",
         "originator": "test_originator",
         "cli_version": "test_version",
+        "base_instructions": null,
```
jif-oai, line 93: Can you make sure we have a test where this is just missing as well?

codex-rs/core/src/rollout/tests.rs
```diff
@@ -89,6 +89,7 @@ fn write_session_file_with_provider(
         "cwd": ".",
```
jif-oai, line 90: Would be good to have a test where this is not null

codex-rs/core/src/tools/handlers/collab.rs
```diff
@@ -557,15 +560,18 @@ fn collab_agent_error(agent_id: ThreadId, err: CodexErr) -> FunctionCallError {
     }
 }
 
-fn build_agent_spawn_config(turn: &TurnContext) -> Result<Config, FunctionCallError> {
+fn build_agent_spawn_config(
+    base_instructions: &BaseInstructions,
```
jif-oai, line 564: There are already some `base_instructions` in `TurnContext`, why do you need different once here? Can you document a bit please? (tbh I don't even make sure we need this)

codex-rs/core/src/client_common.rs
```diff
@@ -245,76 +235,8 @@ mod tests {
     use codex_api::create_text_param_for_request;
     use pretty_assertions::assert_eq;
 
-    use crate::config::test_config;
-    use crate::models_manager::manager::ModelsManager;
-
     use super::*;
 
-    struct InstructionsTestCase {
```
jif-oai, line 253: Yesssss

codex-rs/protocol/src/models.rs
```diff
@@ -167,6 +166,21 @@ pub enum ResponseItem {
     Other,
 }
 
+/// Base instructions for the model in a thread. Corresponds to the `instructions` field in the ResponsesAPI.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, JsonSchema, TS)]
```
jif-oai, line 170: Let's make 100% we don't serialize using this serializer when sending to the model as it would be wrong

jif-oai (approved):
So this means that we can't modify the instructions once the session is started (like the `/search` was doing before for example)
If we are ok with this, good for me

codex-rs/core/src/tools/handlers/collab.rs
dylan-hurd-oai: I'm removing `base_instructions` from TurnContext altogether, since it shouldn't change from turn to turn.  Happy to add a general comment. 

But I think the spirit of this change is "build_agent_spawn_config" should inherit session-level config, not just config housed in TurnContext. If the spirit of `TurnContext` is to denormalize session-level information as well, I'm happy to shift in that direction, but I think we'd want to refactor the serialization there, since we wouldn't want to save the entire base_instructions to the rollout in every turn.

codex-rs/core/src/config/mod.rs
dylan-hurd-oai: Hm, might have gotten pulled in while rebasing on a different branch. Will remove

codex-rs/protocol/src/models.rs
dylan-hurd-oai: @codex check for existing test coverage on the `instructions` field for the responses api request

PR #9437, opened 2026-01-18
fix(core): preserve config.toml symlinks on write

### LINKED ISSUE #6646: Honor config.toml symlinks when creating or updating config

codex-cli 0.58.0

API

gpt-5-codex

Linux 6.6.87.2-microsoft-standard-WSL2 x86_64 x86_64 (Windows 11 host with WSL2 Ubuntu)

Since updating to 0.58.0 the CLI recreates `~/.codex/config.toml` on startup/update and silently replaces my existing symlink with a regular file containing default config. This breaks my setup where the real config lives on the Windows  side and is symlinked into WSL.

1. On 0.57.x (or any build prior to 0.58.0) create a config on Windows and symlink it into WSL.
2. Upgrade Codex CLI to 0.58.0. Launch `codex`.
3. Inspect `ls -l ~/.codex/config.toml`. The symlink has been replaced by a regular file containing the new default config.

The upgrade/startup process should respect an existing symlink (or at least prompt before removing it) so shared configs remain usable after upgrades.

### Additional information

`auth.json` symlinks survive the upgrade; only `config.toml` is rewritten. The new file always contains a `[notice]` block with `hide_gpt5_1_migration_prompt = true`. Please avoid unlinking when the target is already a symlink.

ryoppippi:
Fixes #6646

## What
When Codex writes to `~/.codex/config.toml`, we persist via a temporary file + rename. If `config.toml` is a symlink (e.g. managed externally), renaming over it replaces the symlink with a regular file.

This PR follows the symlink chain and atomically writes to the final target file instead, preserving the symlink.

## Changes
- Follow `config.toml` symlink chain and write to the final target path
- Create the temp file in the target directory so the atomic rename works
- Return a clear error when the symlink target is read-only
- Add a unit test that asserts the symlink is preserved and the target is updated

## Testing
- `nix develop -c cargo test -p codex-core`

### DIFF
--- a/codex-rs/core/src/config/edit.rs
+++ b/codex-rs/core/src/config/edit.rs
@@ -614,6 +614,45 @@ fn normalize_skill_config_path(path: &Path) -> String {
         .to_string()
 }
 
+fn resolve_config_write_target(config_path: &Path) -> anyhow::Result<PathBuf> {
+    // If config.toml is a symlink, writing via rename/persist would replace the
+    // symlink itself. Instead, follow the symlink chain and write to the final
+    // target path so the link is preserved.
+    let mut current = config_path.to_path_buf();
+    for _ in 0..16 {
+        match std::fs::symlink_metadata(&current) {
+            Ok(metadata) => {
+                if !metadata.file_type().is_symlink() {
+                    return Ok(current);
+                }
+
+                let target = std::fs::read_link(&current).with_context(|| {
+                    format!("failed to read config symlink target at {}", current.display())
+                })?;
+                current = if target.is_absolute() {
+                    target
+                } else {
+                    let parent = current
+                        .parent()
+                        .context("config.toml symlink missing parent directory")?;
+                    parent.join(target)
+                };
+            }
+            Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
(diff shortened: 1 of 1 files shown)


ryoppippi:

ryoppippi:
recheck

etraut-openai:
Thanks for the PR. Our contribution guidelines indicate that we're not currently accepting PRs for new features. The linked issue is a feature request, so I would normally close this PR. However, one could argue that this is a bug fix because Codex does follow symlinks for `config.toml`, so it should arguably handle updates if a symlink is present. This change is also pretty localized (doesn't affect UI or other user-visible behavior).

etraut-openai:
I've reviewed your PR. I think it's directionally correct, but it has a lot of shortcomings:
* It has hard-coded undocumented constants for symlink depth limits.
* It doesn't detect symlink cycles or have a good fallback behavior when cycles are detected.
* It doesn't handle all cases where config.toml is written.
* It doesn't encapsulate the symlink-following logic into a utility routine.

I've prepared a [separate PR](https://github.com/openai/codex/pull/9445) that addresses these issues. I welcome you to review and comment on it.

I'm going to close this PR. Thanks again for taking the time to post it.

PR #9444, opened 2026-01-18
fix(core): use platform-appropriate UTF-8 locale in unified exec

zerone0x:
## Summary

Fixes #9354

On macOS, `C.UTF-8` is not a valid locale, causing Codex shell commands under unified exec to fail or produce unexpected behavior when locale-dependent operations are performed.

## Changes

- Introduced platform-specific locale constant `UNIFIED_EXEC_LOCALE`:
  - `en_US.UTF-8` on macOS (universally available)
  - `C.UTF-8` on Linux and other platforms (widely supported)
- Updated `apply_unified_exec_env()` to use this constant for `LANG`, `LC_CTYPE`, and `LC_ALL`
- Updated test to use the platform-appropriate locale constant

## Testing

- Updated existing unit test `unified_exec_env_injects_defaults` to validate against platform-specific locale
- Manual verification: locale settings work correctly on macOS and Linux

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/codex-rs/core/src/unified_exec/process_manager.rs
+++ b/codex-rs/core/src/unified_exec/process_manager.rs
@@ -47,12 +47,18 @@ use crate::unified_exec::process::OutputHandles;
 use crate::unified_exec::process::UnifiedExecProcess;
 use crate::unified_exec::resolve_max_tokens;
 
-const UNIFIED_EXEC_ENV: [(&str, &str); 10] = [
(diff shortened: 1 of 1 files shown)


etraut-openai:
@zerone0x, thanks for the PR. There are technical reasons that we hard-code these env variables for unified exec. We're having some internal discussions about whether it's OK to change this. Until we conclude these discussions, we're going to keep the code as is.

PR #9445, opened 2026-01-18
Fixed symlink support for config.toml

### LINKED ISSUE #6646: Honor config.toml symlinks when creating or updating config

etraut-openai:
We already support reading from `config.toml` through a symlink, but the code was not properly handling updates to a symlinked config file. This PR generalizes safe symlink-chain resolution and atomic writes into path_utils, updating all config write paths to use the shared logic (including set_default_oss_provider, which previously didn't use the common path), and adds tests for symlink chains and cycles.

This resolves #6646.

Notes:
* Symlink cycles or resolution failures replace the top-level symlink with a real file.
* Shared config write path now handles symlinks consistently across edits, defaults, and empty-user-layer creation.

This PR was inspired by https://github.com/openai/codex/pull/9437, which was contributed by @ryoppippi

### DIFF
--- a/codex-rs/core/src/config/service.rs
+++ b/codex-rs/core/src/config/service.rs
@@ -9,6 +9,9 @@ use crate::config_loader::LoaderOverrides;
 use crate::config_loader::load_config_layers_state;
 use crate::config_loader::merge_toml_values;
 use crate::path_utils;
+use crate::path_utils::SymlinkWritePaths;
+use crate::path_utils::resolve_symlink_write_paths;
+use crate::path_utils::write_atomically;
 use codex_app_server_protocol::Config as ApiConfig;
 use codex_app_server_protocol::ConfigBatchWriteParams;
 use codex_app_server_protocol::ConfigLayerMetadata;
@@ -27,6 +30,7 @@ use std::borrow::Cow;
 use std::path::Path;
 use std::path::PathBuf;
 use thiserror::Error;
+use tokio::task;
 use toml::Value as TomlValue;
 use toml_edit::Item as TomlItem;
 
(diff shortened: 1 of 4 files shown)


etraut-openai:
@codex review

codex-rs/core/src/path_utils.rs
```diff
@@ -8,6 +12,100 @@ pub fn normalize_for_path_comparison(path: impl AsRef<Path>) -> std::io::Result<
     Ok(normalize_for_wsl(canonical))
 }
 
+pub struct SymlinkWritePaths {
+    pub read_path: Option<PathBuf>,
+    pub write_path: PathBuf,
+}
+
+pub fn resolve_symlink_write_paths(path: &Path) -> io::Result<SymlinkWritePaths> {
```
bolinfest, line 26: Can you please add a docstring for this? It looks like it guards against cycles but has no max number of symlink resolutions?

Also, can you please add a unit test?

codex-rs/core/src/config/service.rs
```diff
@@ -385,6 +400,13 @@ async fn create_empty_user_layer(
     ))
 }
 
+async fn write_empty_user_config(write_path: PathBuf) -> Result<(), ConfigServiceError> {
+    task::spawn_blocking(move || write_atomically(&write_path, ""))
```
bolinfest, line 404: Should we make an async version of `write_atomically()` instead?

codex-rs/core/src/path_utils.rs
```diff
@@ -8,6 +12,100 @@ pub fn normalize_for_path_comparison(path: impl AsRef<Path>) -> std::io::Result<
...
+    let parent = write_path.parent().ok_or_else(|| {
+        io::Error::new(
+            io::ErrorKind::InvalidInput,
+            format!("path {} has no parent directory", write_path.display()),
+        )
+    })?;
+    std::fs::create_dir_all(parent)?;
+    let tmp = NamedTempFile::new_in(parent)?;
+    std::fs::write(tmp.path(), contents)?;
+    tmp.persist(write_path)?;
```
bolinfest, line 111: I guess this is the real reason why this needs to be sync? It doesn't appear as though there is an async version readily available...

Incidentally, here is how the tempfile crate creates a file on UNIX:

https://github.com/Stebalien/tempfile/blob/8e5ba93488ad0a6354c19776a0cb89b95273a491/src/file/imp/unix.rs#L47-L65

codex-rs/core/src/path_utils.rs
```diff
@@ -8,6 +12,100 @@ pub fn normalize_for_path_comparison(path: impl AsRef<Path>) -> std::io::Result<
     Ok(normalize_for_wsl(canonical))
 }
 
+pub struct SymlinkWritePaths {
```
bolinfest, line 15: This type feels a little weird to me: symlinks are relatively rare, but you get this type back even when symlinks are not involved.

Should this be an enum like:

```rust
pub enum WritePath {
  RegularFile {
    write_path: PathBuf,
  }
  Symlink {
    read_path: PathBuf,
    write_path: PathBuf,
  }
}
```

bolinfest (approved):

PR #9462, opened 2026-01-19
Prefix-filter TUI slash commands

aibrahim-oai:
- Make slash command suggestions prefix-only (tui + tui2)\n- Update snapshots; improve cargo_bin fallback for integration tests


codex-rs/utils/cargo-bin/src/lib.rs
joshka-oai, line 1: looks unrelated - guessing codex fixing a test?

Issue #6717, opened 2025-11-16
codex exec resume --last rejects prompt when --json is set


dkhokhlov:
### What version of Codex is running?

0.58.0

### What subscription do you have?

team

### Which model were you using?

gpt-5 codex

### What platform is your computer?

Linux 6.2.0-39-generic x86_64 x86_64

### What issue are you seeing?

Running the exec CLI with both `--json` and `resume --last` fails whenever a prompt is provided after `--last`. Clap treats that positional as `SESSION_ID`, so it errors with “the  argument '--last' cannot be used with '[SESSION_ID]'”.

Execution aborts immediately with `error: the argument '--last' cannot be used with '[SESSION_ID]'`.

### What steps can reproduce the bug?

Run: 
`codex exec --json resume --last "2+2"`

### What is the expected behavior?

The command resumes the most recent session and sends `"2+2"` as the new prompt while still emitting JSON events.

ghost:
EDIT . Actually exec version (non interactive) works . Interactive doesn't. Not sure if I should create separate ticket.

On latest version 0.63 you still cannot use it . For command :
`
codex resume --last "tell me the same joke"`

I get 

error: the argument '--last' cannot be used with '[SESSION_ID]'

Usage: codex resume --last [SESSION_ID] [PROMPT]


tested on Linux

PR #9475, opened 2026-01-19
Fix `resume --last` with `--json` option

etraut-openai:
Fix resume --last prompt parsing by dropping the clap conflict on the codex resume subcommand so a positional prompt is accepted when --last is set. This aligns interactive resume behavior with exec-mode logic and avoids the “--last cannot be used with SESSION_ID” error.

This addresses #6717


jif-oai (approved):

codex-rs/cli/src/main.rs
```diff
@@ -929,6 +929,23 @@ mod tests {
...
+
+        let Some(Subcommand::Exec(exec)) = cli.subcommand else {
+            panic!("expected exec subcommand");
+        };
+        let Some(codex_exec::Command::Resume(args)) = exec.command else {
+            panic!("expected exec resume");
+        };
+
+        assert!(args.last);
+        assert_eq!(args.session_id.as_deref(), Some("2+2"));
```
pakrym-oai: This looks like a bad breaking change.

pakrym-oai (approved):

pakrym-oai (changes requested):
wrong button

PR #9478, opened 2026-01-19
feat: close all threads in `/new`

jif-oai:


codex-rs/core/src/thread_manager.rs
```diff
@@ -235,6 +235,15 @@ impl ThreadManager {
         self.state.threads.write().await.remove(thread_id)
     }
 
+    /// Closes all threads open in this ThreadManager
+    pub async fn remove_and_close_all_threads(&self) -> CodexResult<()> {
+        for thread in self.state.threads.read().await.values() {
+            thread.submit(Op::Shutdown).await?;
```
gt-oai, line 241: Will this error contain any info about which thread failed to shutdown?

gt-oai (approved):

codex-rs/core/src/thread_manager.rs
jif-oai: Nope :/ this basically should only happen when the thread is already gone

PR #9492, opened 2026-01-19
Add originator to metrics

gt-oai:
Will be helpful to distinguish extension vs CLI, for example.


jif-oai (approved):
There seems to be some redundancy in the flags, we should clean this a bit

gt-oai:
`service.name` already exists and is what we want (maybe should be "originator")

PR #9495, opened 2026-01-19
Feat: plan mode prompt update

shijie-oai:
### Summary
* Added instruction on using `request_user_input`
* Added the output to be json with `plan` key and the actual plan as the value. 
* Remove `PLAN.md` write because that gets into sandbox issue. We can add it back later.

### DIFF
--- a/codex-rs/core/templates/collaboration_mode/plan.md
+++ b/codex-rs/core/templates/collaboration_mode/plan.md
@@ -1,40 +1,133 @@
 # Collaboration Style: Plan
+
 You work in 2 distinct modes:
+
 1. Brainstorming: You collaboratively align with the user on what to do or build and how to do it or build it.
-2. Writing and confirming a plan: After you've gathered all the information you write up a plan and verify it with the user.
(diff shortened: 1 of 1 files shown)


codex-rs/core/templates/collaboration_mode/plan.md
```diff
@@ -1,40 +1,133 @@
...
 
-*Be mindful of time.* The user is right here with you. Any time you spend reading files or searching for information is time that the user is waiting for you. Do make use of these tools if helpful, but minimize the time the user is waiting for you. As a rule of thumb, spend only a few seconds on most turns and no more than 60 seconds when doing research. If you are missing information and think you need to do longer research, ask the user whether they want you to research, or want to give you a tip.
+_Be mindful of time._ The user is right here with you. Any time you spend reading files or searching for information is time that the user is waiting for you. Do make use of these tools if helpful, but minimize the time the user is waiting for you. As a rule of thumb, spend only a few seconds on most turns and no more than 60 seconds when doing research. If you are missing information and think you need to do longer research, ask the user whether they want you to research, or want to give you a tip.
```
shijie-oai, line 29: Automatic formatting - not sure if this is good or bad and can revert back to what we had before.

aibrahim-oai (approved):

PR #9497, opened 2026-01-19
Migrate tui to use UserTurn

aibrahim-oai:
- `tui/` and `tui2/` submit `Op::UserTurn` and own full turn context (cwd/approval/sandbox/model/etc.).
- `Op::UserInput` is documented as legacy in `codex-protocol` (doc-only; no `#[deprecated]` to avoid `-D warnings` fallout).
- Remove obsolete `#[allow(deprecated)]` and the unused `ConversationId` alias/re-export.


aibrahim-oai:
@codex review

jif-oai (approved):

PR #2576, opened 2025-08-22
[apply_patch] freeform apply_patch tool

dylan-hurd-oai:
## Summary
GPT-5 introduced the concept of [custom tools](https://platform.openai.com/docs/guides/function-calling#custom-tools), which allow the model to send a raw string result back, simplifying json-escape issues. We are migrating gpt-5 to use this by default.

However, gpt-oss models do not support custom tools, only normal functions. So we keep both tool definitions, and provide whichever one the model family supports.

## Testing
- [x] Tested locally with various models
- [x] Unit tests pass

PR #9498, opened 2026-01-19
fix: improve apply_patch command detection for path prefixes and variants

### LINKED ISSUE #2235: apply_patch sometimes invoked as shell command -> 'apply_patch: command not found'
@bolinfest @gpeal @pakrym-oai @dylan-hurd-oai

**What version of Codex is running?**
Built from source at commit `cb78f23` (local copy).

**Which model were you using?**
Observed when assistant outputs resembling `gpt-4.1` style tool calls; the bug is about how assistant-generated `apply_patch` invocations are parsed/handled by Codex regardless of model.

**What platform is your computer?**
Linux 6.14.3-x86_64-linode168 x86_64 unknown

**Steps to reproduce**
1. Have the assistant produce a shell/tool call that embeds an `apply_patch` invocation in a shell script. Example variants that can trigger the issue:
   - `{"command": ["apply_patch", "<<'EOF'\n*** Begin Patch\n*** Update File: README.md\n@@\n- old\n+ new\n*** End Patch\nEOF\n"]}`
   - `{"command": ["bash","-lc","cd repo && apply_patch <<'EOF'\n*** Begin Patch\n*** Update File: README.md\n@@\n- old\n+ new\n*** End Patch\nEOF\n"]}`
2. Codex attempts to execute the tool call. If the invocation is not recognized by the `maybe_parse_apply_patch` detection logic, the runtime will spawn the requested shell command literally.
3. Since `apply_patch` is a *virtual* tool implemented inside the Codex binary (invoked via `--codex-run-as-apply-patch`), there is no `apply_patch` binary on PATH, so bash reports:

   `bash: line N: apply_patch: command not found`

**Expected behavior**
- Codex should robustly detect `apply_patch` invocations (including heredoc forms and when `apply_patch` is not at the start of the script) and either: (A) handle the patch programmatically, or (B) exec the Codex binary with `--codex-run-as-apply-patch <patch>` so the virtual tool runs — instead of running the shell which will try to find a real `apply_patch` binary.

**Actual behavior**
- In some model-generated cases the detection logic misses the `apply_patch` invocation (examples: scripts that include `cd && apply_patch <<'EOF'`, other shells like `sh -lc`, or different command-name spellings like `apply-patch`). The fallback is to spawn the shell literally, which leads to `apply_patch: command not found`.

**Relevant code pointers**
- Detection/parsing: `codex-rs/apply-patch/src/lib.rs` (`maybe_parse_apply_patch`, `extract_heredoc_body_from_apply_patch_command`)
- Patch parsing: `codex-rs/apply-patch/src/parser.rs`
- Replacement exec: `codex-rs/core/src/codex.rs` (creates exec with `--codex-run-as-apply-patch` when a patch is recognized)
- Spawn/exec: `codex-rs/core/src/exec.rs` and `codex-rs/core/src/spawn.rs`

**Suggested fixes / mitigation**
- Relax the initial check in `extract_heredoc_body_from_apply_patch_command` (it currently requires the script to start with `apply_patch`) and instead traverse the AST to find any `apply_patch` invocation and extract its heredoc body.
- Accept more shell invocation forms (e.g. `sh -lc`) and common command-name variants (e.g. `apply-patch`) or normalize names in detection.
- Add a pre-spawn, permissive check: if any element of `params.command` contains `apply_patch`/`applypatch`/`<<EOF` etc., try to parse it as a patch before actually spawning.
- Add unit tests that cover realistic failing variants (e.g., `cd && apply_patch <<'EOF'` and `sh -lc` forms).

**Additional notes**
This is a reliability/usability issue that causes noisy user-visible errors when the model's generated tool call is syntactically different from the narrow forms currently recognized. Improving detection should reduce these failures.

yuvrajangadsingh:
## Summary

Fixes #2235

The `apply_patch` command detection was failing for certain edge cases reported by users:
- `/usr/bin/apply_patch` (absolute path prefix)
- `./apply_patch` (relative path prefix)  
- `apply-patch` (hyphenated variant)

### `apply-patch/src/invocation.rs`
- Added `is_apply_patch_command()` helper that extracts the filename from path-prefixed commands
- Added `apply-patch` to the `APPLY_PATCH_COMMANDS` array
- Updated Tree-sitter query to match `apply-patch` in heredoc scripts
- Added 5 new tests covering the edge cases

### `arg0/src/lib.rs`
- Added `HYPHENATED_APPLY_PATCH_ARG0` constant
- Updated arg0 dispatch to handle `apply-patch` invocations
- Added symlink creation for `apply-patch` in PATH

## Test plan

- [x] All 54 existing tests pass
- [x] 5 new tests for edge cases pass
- [x] Clippy clean


etraut-openai:
@codex review

yuvrajangadsingh:
Hi @etraut-openai, this PR has been open for a couple weeks. Would you be able to take a look when you have a chance? It improves the apply_patch command detection to handle path prefixes and variants.

yuvrajangadsingh:
@bolinfest Since you commented on issue #2235, would you mind taking a look at this fix as well?

bolinfest:
While this has historically been a problem (certainly when https://github.com/openai/codex/issues/2235 was originally opened), reports have died down considerably. Accepting this change would paper over errors _and_ potentially result in incorrect behavior if, for some reason, you had your own `/usr/bin/apply_patch` that you wanted the model to use.

We have also refined the custom `apply_patch` tool (originally introduced in https://github.com/openai/codex/pull/2576, I believe), reducing the reliance on the CLI, which also appears to be helping.

bolinfest:
As such, thank you for your submission, but we are not going to take this solution.

PR #9509, opened 2026-01-20
Fixed config merging issue with profiles

etraut-openai:
This PR fixes a small issue with chained (layered) config.toml file merging. The old logic didn't properly handle profiles.

In particular, if a lower-layer config overrides a profile defined in a higher-layer config, the override did not take effect. This prevents users from having project-specific profile overrides and contradicts the (soon-to-be) documented behavior of config merging.

The change adds a unit test for this case. It also exposes a function from the config crate that is needed by the app server code paths to implement support for layered configs.

### DIFF
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -2313,9 +2313,11 @@ trust_level = "trusted"
     #[tokio::test]
     async fn project_profile_overrides_user_profile() -> std::io::Result<()> {
         let codex_home = TempDir::new()?;
+        let workspace = TempDir::new()?;
         std::fs::write(
             codex_home.path().join(CONFIG_TOML_FILE),
-            r#"
+            format!(
+                r#"
 profile = "global"
 
 [profiles.global]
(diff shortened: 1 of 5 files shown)


codex-rs/core/src/config/mod.rs
```diff
@@ -506,6 +506,18 @@ impl Config {
     }
 }
 
+/// Load config with CLI overrides and apply a fallback cwd when none is specified in overrides.
+pub async fn load_with_cli_overrides_and_fallback_cwd(
```
bolinfest: I would like to avoid having multiple ways to construct a `Config`.

Should `fallback_cwd` be an option on `ConfigBuilder` instead?

bolinfest (commented):
@etraut-openai could you please expand the PR body to explain what the fix was? I see the new unit test, which is great, but I also see a bunch of changes to TUI code that I did not expect. It seems like there is some relationship between restoring `cwd` and profile correctness, which is not obvious, so I think that should be explained in the PR body.

bolinfest (approved):

codex-rs/core/src/config/mod.rs
etraut-openai: Good suggestion. I'll make that change.

PR #9510, opened 2026-01-20
Add layered config.toml support to app server

etraut-openai:
This PR adds support for chained (layered) config.toml file merging for clients that use the app server interface. This feature already exists for the TUI, but it does not work for GUI clients.

It does the following:
* Changes code paths for new thread, resume thread, and fork thread to use the effective config based on the cwd.
* Updates the `config/read` API to accept an optional `cwd` parameter. If specified, the API returns the effective config based on that cwd path. Also optionally includes all layers including project config files. If cwd is not specified, the API falls back on its older behavior where it considers only the global (non-project) config files when computing the effective config.

The changes in codex_message_processor.rs look deceptively large. They mostly just involve moving existing blocks of code to a later point in some functions so it can use the cwd to calculate the config.

This PR builds upon #9509 and should be reviewed and merged after that PR.

Tested:
* Verified change with (dependent, as-yet-uncommitted) changes to IDE Extension and confirmed correct behavior

The full fix requires additional changes in the IDE Extension code base, but they depend on this PR.


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -449,6 +448,8 @@ pub enum ConfigWriteErrorCode {
 pub struct ConfigReadParams {
     #[serde(default)]
     pub include_layers: bool,
+    /// Optional working directory to resolve project config layers (e.g. .codex/config.toml).
```
owenlin0: Might be helpful to explain "project config layers" a bit more. When should the user specify `cwd` (i.e. always, unless they specifically want to ignore their project-level config)? If specified, what should the `cwd` typically be (e.g. the root of their current git repo)?

owenlin0: (i'm also learning btw, but helpful to write this down as documentation for others too)

owenlin0 (approved):
API changes look good to me, just would like an expanded docstring (from my comments above)

PR #9512, opened 2026-01-20
Fix resume picker when user event appears after head

jdsalasca:
Fixes #9501

Contributing guide: https://github.com/openai/codex/blob/main/docs/contributing.md

## Summary
The resume picker requires a session_meta line and at least one user_message event within the initial head scan. Some rollout files contain multiple session_meta entries before the first user_message, so the user event can fall outside the default head window and the session is omitted from the picker even though it is resumable by ID.

This PR keeps the head summary bounded but extends scanning for a user_message once a session_meta has been observed. The summary still caps stored head entries, but we allow a small, bounded extra scan to find the first user event so valid sessions are not filtered out.

## Changes
- Continue scanning past the head limit (bounded) when session_meta is present but no user_message has been seen yet.
- Mark session_meta as seen even if the head summary buffer is already full.
- Add a regression test with multiple session_meta lines before the first user_message.

## Why This Is Safe
- The head summary remains bounded to avoid unbounded memory usage.
- The extra scan is capped (USER_EVENT_SCAN_LIMIT) and only triggers after a session_meta is seen.
- Behavior is unchanged for typical files where the user_message appears early.

## Testing
- cargo test -p codex-core --lib test_list_threads_scans_past_head_for_user_event


jdsalasca:

etraut-openai:
@codex review

aibrahim-oai (approved):

etraut-openai:
@jdsalasca, thanks for the contribution!

PR #9528, opened 2026-01-20
feat: metrics on remote models

jif-oai:


codex-rs/core/src/models_manager/manager.rs
```diff
@@ -211,6 +210,8 @@ impl ModelsManager {
     }
 
     async fn fetch_and_update_models(&self) -> CoreResult<()> {
+        let _timer =
+            codex_otel::start_global_timer("codex.remote_models.fetch_update.duration_ms", &[]);
```
gt-oai, line 214: it's weird to me that this timer metric name includes `duration_ms` -- surely units are already handled by the timer itself?

gt-oai (approved):

codex-rs/core/src/models_manager/manager.rs
jif-oai: This is more for the collection part. It makes is non ambigious for the DS team building dashboards on top of this

PR #9533, opened 2026-01-20
Only load config from trusted folders

gt-oai:
Config includes multiple code execution entrypoints. 

Now, we load the config from predetermined locations first (~/.codex/config.toml etc), use those to learn which folders are 'trusted', and only load additional config from the CWD if it is trusted.

### DIFF
--- a/codex-rs/core/src/config_loader/mod.rs
+++ b/codex-rs/core/src/config_loader/mod.rs
@@ -12,10 +12,13 @@ mod tests;
 
 use crate::config::CONFIG_TOML_FILE;
 use crate::config::ConfigToml;
+use crate::config::deserialize_config_toml_with_base;
 use crate::config_loader::config_requirements::ConfigRequirementsWithSources;
 use crate::config_loader::layer_io::LoadedConfigLayers;
+use crate::git_info::resolve_root_git_project_for_trust;
 use codex_app_server_protocol::ConfigLayerSource;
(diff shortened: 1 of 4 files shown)


codex-rs/core/src/config_loader/mod.rs
```diff
@@ -158,17 +167,22 @@ pub async fn load_config_layers_state(
...
+            merge_toml_values(&mut merged_so_far, cli_overrides_layer);
+        }
+
         let project_root_markers = project_root_markers_from_config(&merged_so_far)?
             .unwrap_or_else(default_project_root_markers);
-
-        let project_root = find_project_root(&cwd, &project_root_markers).await?;
-        let project_layers = load_project_layers(&cwd, &project_root).await?;
-        layers.extend(project_layers);
+        if let Some(project_root) =
```
jif-oai, line 176: We should send a warning to the user instead of silently ignoring

jif-oai (approved):
We should add this to the doc website as well

codex-rs/core/src/config_loader/mod.rs
gt-oai: Good point, will print a warning iff you have CWD / git repo codex config AND the folder is untrusted (will do in a follow up)

xl-openai:
FYI this PR has some side effect for skill discovery.

PR #9539, opened 2026-01-20
feat: dynamic tools injection

jif-oai:
## Summary
Add dynamic tool injection to thread startup in API v2, wire dynamic tool calls through the app server to clients, and plumb responses back into the model tool pipeline.

### Flow (high level)
- Thread start injects `dynamic_tools` into the model tool list for that thread (validation is done here).
- When the model emits a tool call for one of those names, core raises a `DynamicToolCallRequest` event.
- The app server forwards it to the client as `item/tool/call`, waits for the client’s response, then submits a `DynamicToolResponse` back to core.
- Core turns that into a `function_call_output` in the next model request so the model can continue.

### What changed
- Added dynamic tool specs to v2 thread start params and protocol types; introduced `item/tool/call` (request/response) for dynamic tool execution.
- Core now registers dynamic tool specs at request time and routes those calls via a new dynamic tool handler.
- App server validates tool names/schemas, forwards dynamic tool call requests to clients, and publishes tool outputs back into the session.
- Integration tests


jif-oai:
@codex review

jif-oai:
Will fix the merge ofc

codex-rs/core/src/tools/spec.rs
```diff
@@ -1101,6 +1102,26 @@ pub(crate) fn mcp_tool_to_openai_tool(
     })
 }
 
+fn dynamic_tool_to_openai_tool(
+    tool: &DynamicToolSpec,
+) -> Result<ResponsesApiTool, serde_json::Error> {
+    let input_schema = parse_tool_input_schema(&tool.input_schema)?;
```
dylan-hurd-oai, line 1108: I'm a bit worried that RESERVED_DYNAMIC_TOOL_NAMES will drift - should we check for conflicts/duplicates at runtime, or add a test to assert that all built-in tool names are covered by the list?

dylan-hurd-oai (approved):
1 suggestion for specs, otherwise really like the approach!

PR #9545, opened 2026-01-20
Add total (non-partial) TextElement placeholder accessors

charley-oai:
## Summary
  - Make `TextElement` placeholders private and add a text-backed accessor to avoid assuming `Some`.
  - Since they are optional in the protocol, we want to make sure any accessors properly handle the None case (getting the placeholder using the byte range in the text)
  - Preserve placeholders during protocol/app-server conversions using the accessor fallback.
  - Update TUI composer/remap logic and tests to use the new constructor/accessor.


charley-oai:
@codex review this

aibrahim-oai (approved):

PR #9547, opened 2026-01-20
merge remote models

aibrahim-oai:
We have `models.json` and `/models` response
Behavior:
1. New models from models endpoint gets added
2. Shared models get replaced by remote ones
3. Existing models in `models.json` but not `/models` are kept
4. Mark highest priority as default

### DIFF
--- a/codex-rs/core/src/models_manager/manager.rs
+++ b/codex-rs/core/src/models_manager/manager.rs
@@ -407,6 +407,16 @@ mod tests {
         .expect("valid model")
     }
 
+    fn assert_models_contain(actual: &[ModelInfo], expected: &[ModelInfo]) {
+        for model in expected {
+            assert!(
+                actual.iter().any(|candidate| candidate.slug == model.slug),
+                "expected model {} in cached list",
+                model.slug
+            );
+        }
+    }
+
     fn provider_for(base_url: String) -> ModelProviderInfo {
         ModelProviderInfo {
             name: "mock".into(),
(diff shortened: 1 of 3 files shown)


aibrahim-oai:
@codex review this

aibrahim-oai:
@codex review

codex-rs/core/src/models_manager/manager.rs
```diff
@@ -239,7 +239,18 @@ impl ModelsManager {
...
+            if let Some(existing_index) = existing_models
+                .iter()
+                .position(|existing| existing.slug == model.slug)
+            {
+                existing_models[existing_index] = model;
+            } else {
+                existing_models.push(model);
+            }
+        }
+        *self.remote_models.write().await = existing_models;
```
charley-oai, line 253: `self.remote_models` now stores a merged catalog (bundled `models.json` + `/models` overrides/additions), not just the remote response. Consider renaming to something like `cached_models` or `merged_models` (and update related comments) to avoid misleading future readers.

charley-oai (approved):

codex-rs/core/src/models_manager/manager.rs
aibrahim-oai: true. but in a way models.json is also cached remote models and is different from the hot cache. This needs simplification I agree.

charley-oai: ack yeah just a suggestion, feel free to not do. My read is a distinction btwn remote vs bundled vs cached

PR #9549, opened 2026-01-20
Reduce burst testing flake

charley-oai:
## Summary

- make paste-burst tests deterministic by injecting explicit timestamps instead of relying on wall clock timing
- add time-aware helpers for input/submission paths so tests can drive the burst heuristic precisely
- update burst-related tests to flush using computed timeouts while preserving behavior assertions
- increase timeout slack in shell_tools_start_before_response_completed_when_stream_delayed to reduce flakiness


charley-oai:
@codex review this

dylan-hurd-oai (approved):

PR #9551, opened 2026-01-20
nit: better collab tui

jif-oai:
Screenshot 2026-01-21 at 11 53 50


gt-oai (approved):
would be nice to see a screenshot

PR #9565, opened 2026-01-21
Fix execpolicy parsing for multiline quoted args

### LINKED ISSUE #9541: execpolicy allow rules fail when a quoted argument contains a literal newline

0.87.0

Pro

gpt-5.2-codex (reasoning high)

Darwin 25.2.0 arm64 arm

Ghostty 1.2.3

When a command includes a literal newline inside a double‑quoted string, the command extraction step fails, so execpolicy cannot extract the inner command.

Attempting to debug this with codex-cli yielded the following insights:

`tree-sitter-bash` produces a valid string node with multiple `string_content` children when a double‑quoted string contains a literal newline. `parse_plain_command_from_node` currently only accepts a string with exactly one string_content, so `parse_shell_lc_plain_commands` returns `None` in this case.

That means allow rules like `["git","commit"]` do not match when a newline is included in an argument to the command and `create_exec_approval_requirement_for_command` does not set `bypass_sandbox`. In `workspace-write` this causes git commit to fail with:

`fatal: Unable to create '.../.git/index.lock': Operation not permitted`

The same commit succeeds if the body is split into multiple `-m` flags without a literal newline.

Depending on approval policy and user interaction, Codex may then prompt/escalate to run outside the sandbox, even though a rule allowing git commit exists.

1) Create a rule file in ~/.codex/rules/ that allows git commit:

`prefix_rule(pattern=["git", "commit"], decision="allow")`

2) Run Codex with workspace-write sandbox (default, or `--sandbox workspace-write`).

3) Attempt a commit with a literal newline inside a quoted -m argument:

```
git commit -m "title" -m "line1
line2"
```

4) Observe failure:

`fatal: Unable to create '.../.git/index.lock': Operation not permitted`

5) Compare with:

```
git commit -m "title" -m "line1" -m "line2"
(This succeeds because the allow rule matches and bypasses the sandbox.)
```

[Example rules and config that successfully replicate this behavior](https://gist.github.com/adamalix/152d9f439bc57babbb740c28848a459b).

Allow rules like ["git","commit"] should apply even when a quoted argument contains a literal newline, so git commit runs with the same sandbox bypass behavior regardless of newline formatting in the message.

### Additional information

This appears to be caused by `parse_shell_lc_plain_commands` rejecting scripts with literal newlines inside quoted strings, so `create_exec_approval_requirement_for_command` falls back to the outer `zsh -lc` argv and cannot match allow rules.

Potentially related issues (context):

https://github.com/openai/codex/issues/7071

Some tests that are primarily codex-driven which confirm the behavior described above: https://github.com/openai/codex/compare/main...adamalix:codex:execpolicy-multiline-newline-tests

jdsalasca:
## What
Fix bash command parsing to accept double-quoted strings that contain literal newlines so execpolicy can match allow rules.

## Why
Allow rules like [git, commit] should still match when commit messages include a newline in a quoted argument; the parser currently rejects these strings and falls back to the outer shell invocation.

## How
- Validate double-quoted strings by ensuring all named children are string_content and then stripping the outer quotes from the raw node text so embedded newlines are preserved.
- Reuse the helper for concatenated arguments.
- Ensure large SI suffix formatting uses the caller-provided locale formatter for grouping.
- Add coverage for newline-containing quoted arguments.

Fixes #9541.

## Tests
- cargo test -p codex-core
- just fix -p codex-core
- cargo test -p codex-protocol
- just fix -p codex-protocol
- cargo test --all-features


etraut-openai:
@codex review

codex-rs/core/src/bash.rs
```diff
@@ -119,6 +119,23 @@ pub fn parse_shell_lc_plain_commands(command: &[String]) -> Option<Vec<Vec<Strin
     try_parse_word_only_commands_sequence(&tree, script)
 }
 
+fn parse_double_quoted_string(node: Node, src: &str) -> Option<String> {
```
bolinfest: We generally declare utility support functions below public/primary library functions (particularly if they are well-named), so can you please move this down?

codex-rs/core/src/bash.rs
```diff
@@ -250,6 +261,20 @@ mod tests {
         );
     }
 
+    #[test]
+    fn accepts_double_quoted_strings_with_newlines() {
+        let cmds = parse_seq("git commit -m \"line1\nline2\"").unwrap();
```
bolinfest, line 254: `parse_double_quoted_string()` could use more tests. I would like to verify that things like env vars embedded in double quoted strings are rejected.

I would also like to see if this covers examples like:

```
"/usr"'/'"local"/bin
```

where mixed quote types are used.

codex-rs/protocol/src/num_format.rs
```diff
@@ -25,7 +25,11 @@ fn formatter() -> &'static DecimalFormatter {
 /// Format an i64 with locale-aware digit separators (e.g. "12345" -> "12,345"
 /// for en-US).
 pub fn format_with_separators(n: i64) -> String {
-    formatter().format(&Decimal::from(n)).to_string()
+    format_with_separators_with_formatter(n, formatter())
```
bolinfest: This is unrelated to the `bash.rs` change, no?

jdsalasca:
Addressed review feedback: moved the quote-parsing helpers below the public entry points, added tests for double-quoted expansions and mixed-quote concatenation, and dropped the unrelated num_format change.

bolinfest (approved):
Nice job: tysm!

bolinfest:
FYI, I made some small fixes in https://github.com/openai/codex/pull/9745 that I didn't think were worth another round of code review.

PR #9568, opened 2026-01-21
Show session header before configuration

aibrahim-oai:
We were skipping if we know the model. We shouldn't

### DIFF
--- a/codex-rs/tui2/src/history_cell.rs
+++ b/codex-rs/tui2/src/history_cell.rs
@@ -19,6 +19,7 @@ use crate::exec_cell::output_lines;
 use crate::exec_cell::spinner;
 use crate::exec_command::relativize_to_home;
 use crate::exec_command::strip_bash_lc_and_escape;
+use crate::key_hint;
 use crate::markdown::append_markdown;
 use crate::render::line_utils::line_to_static;
 use crate::render::line_utils::prefix_lines;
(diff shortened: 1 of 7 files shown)


aibrahim-oai:
@codex review

codex-rs/tui2/src/history_cell.rs
```diff
@@ -1146,25 +1181,48 @@ impl HistoryCell for SessionHeaderHistoryCell {
...
-        model_spans.push(CHANGE_MODEL_HINT_COMMAND.cyan());
-        model_spans.push(CHANGE_MODEL_HINT_EXPLANATION.dim());
+        let model_spans: Vec<Span<'static>> = if self.is_collaboration {
+            let collab_label = format!(
+                "{collab_label:<label_width$}",
+                collab_label = "mode:",
+                label_width = label_width
+            );
+            let mut spans = vec![Span::from(format!("{collab_label} ")).dim()];
+            if self.model == "loading" {
```
charley-oai: Placeholder detection in collaboration headers relies on the literal "loading". If DEFAULT_MODEL_DISPLAY_NAME ever changes, placeholders will silently fall back to showing a mode
label instead of “loading.” Consider sharing the constant or passing an explicit placeholder flag.

codex-rs/tui2/src/history_cell.rs
charley-oai, line 2494: The new collaboration-mode branch in the TUI2 session header isn’t exercised by tests (current header test uses is_collaboration=false). A targeted test would lock in the
mode label + Shift+Tab hint behavior.

charley-oai (approved):

codex-rs/tui2/src/history_cell.rs
aibrahim-oai: good one

PR #9570, opened 2026-01-21
fix: MCP cwd use absolute path

shijie-oai:
### Issue
When using MCP with cwd, `~` in cwd path is not expanded. Reproduce with the following steps:

`~/.codex/config.toml`
```
[mcp_servers.context7Local]
command = "./scripts/context7-script.sh"
cwd = "~/Documents"
```

`~/Documents/scripts/context7-script.sh`
```
#!/bin/bash
set -euo pipefail

cd "$(dirname "$0")/.."
set -a && source .env && set +a

API_KEY="${CONTEXT7_API_KEY:-${CONTEXT_API_KEY:-}}"
if [ -z "$API_KEY" ]; then
  echo "CONTEXT7_API_KEY (or CONTEXT_API_KEY) is not set in $BASE_DIR/.env" >&2
  exit 1
fi

exec npx -y @upstash/context7-mcp --api-key "$API_KEY"
```

`~/Documents/.env`
```
CONTEXT_API_KEY="key_from_https://context7.com/"
```

### Summary
* Converted cwd for mcp config  to use absolute path so that we can handle `~` and relative path based on the config layer. 
* Ensure that we mcp command for MCP is correctly concatenated if there is a cwd detected and the command is a relative path.

### DIFF
--- a/codex-rs/core/src/config/types.rs
+++ b/codex-rs/core/src/config/types.rs
@@ -14,6 +14,7 @@ use std::path::PathBuf;
 use std::time::Duration;
 use wildmatch::WildMatchPattern;
 
+use dirs::home_dir;
 use schemars::JsonSchema;
 use serde::Deserialize;
 use serde::Deserializer;
@@ -156,11 +157,11 @@ impl<'de> Deserialize<'de> for McpServerConfig {
             throw_if_set("stdio", "http_headers", raw.http_headers.as_ref())?;
             throw_if_set("stdio", "env_http_headers", raw.env_http_headers.as_ref())?;
             McpServerTransportConfig::Stdio {
-                command,
+                command: expand_tilde_string(command),
                 args: raw.args.clone().unwrap_or_default(),
                 env: raw.env.clone(),
                 env_vars: raw.env_vars.clone().unwrap_or_default(),
-                cwd: raw.cwd.take(),
+                cwd: raw.cwd.take().map(expand_tilde_pathbuf),
             }
         } else if let Some(url) = raw.url.clone() {
             throw_if_set("streamable_http", "args", raw.args.as_ref())?;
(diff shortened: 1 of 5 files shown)


shijie-oai:
@codex review this

codex-rs/rmcp-client/src/rmcp_client.rs
```diff
@@ -435,6 +437,20 @@ impl RmcpClient {
...
 
+fn resolve_relative_program(program: OsString, cwd: Option<&Path>) -> OsString {
+    let Some(cwd) = cwd else {
+        return program;
+    };
+
+    let program_path = Path::new(&program);
+    let has_separator = program_path.components().count() > 1;
+    if program_path.is_relative() && has_separator {
+        return cwd.join(program_path).into_os_string();
```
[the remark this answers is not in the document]
shijie-oai: I am getting conflict information from codex web and codex local....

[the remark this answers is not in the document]
shijie-oai: I think ultimately it should be fine cause we would convert the program to be absolute pathed too and therefore it doesn't really matter with `Command::current_dir(cwd)`.

codex-rs/core/src/config/types.rs
```diff
@@ -194,6 +194,26 @@ const fn default_enabled() -> bool {
     true
 }
 
+fn expand_tilde_string(value: String) -> String {
+    if cfg!(target_os = "windows") {
+        return value;
+    }
```
bolinfest, line 200: We could choose to honor this in some way on Windows, though perhaps we would recognize both `~/` and `~\` in that case and then use `%USERPROFILE%` instead of `$HOME`?

codex-rs/core/src/config/types.rs
```diff
@@ -194,6 +194,26 @@ const fn default_enabled() -> bool {
...
+
+    let Some(home) = home_dir() else {
+        return value;
+    };
```
bolinfest, line 204: Personally, I would read this lazily, like:

```rust
if !value.starts_with("~") {
  return value;
}

let Some(home) = home_dir() else {
  return value;
}

// at this point, it is extremely unlikely that you won't use `home`
```

codex-rs/core/src/config/types.rs
```diff
@@ -194,6 +194,26 @@ const fn default_enabled() -> bool {
...
+
+    if value == "~" {
```
bolinfest, line 206: I guess we don't really need to support `~username`.

codex-rs/core/src/config/types.rs
```diff
@@ -194,6 +194,26 @@ const fn default_enabled() -> bool {
...
+        return home.to_string_lossy().into_owned();
+    }
+
+    if let Some(rest) = value.strip_prefix("~/") {
+        return home.join(rest).to_string_lossy().into_owned();
+    }
+
+    value
```
bolinfest, line 214:
```suggestion
    if value == "~" {
        home.to_string_lossy().into_owned()
    } else if let Some(rest) = value.strip_prefix("~/") {
        return home.join(rest).to_string_lossy().into_owned();
    } else {
        value
    }
```

codex-rs/core/src/config/types.rs
```diff
@@ -156,7 +156,7 @@ impl<'de> Deserialize<'de> for McpServerConfig {
             throw_if_set("stdio", "http_headers", raw.http_headers.as_ref())?;
             throw_if_set("stdio", "env_http_headers", raw.env_http_headers.as_ref())?;
             McpServerTransportConfig::Stdio {
-                command,
+                command: expand_tilde_string(command),
```
bolinfest, line 159: I think the real question is whether we should do this only for `cwd` or for all `AbsolutePathBuf`s in config. It seems weird just to do it in one place.

I'm not sure we should honor `~` across `AbsolutePathBuf` in general, but maybe we could add a different sort of guard to honor it across `Config`:

https://github.com/openai/codex/blob/5dad1b956e872b5cf960c367a627358a58f2ce60/codex-rs/core/src/config/mod.rs#L557

PR #9571, opened 2026-01-21
feat(app-server): support archived threads in thread/list

owenlin0:
Add an `archived` bool filter to `thread/list` to support showing archived threads.

Archived threads are stored in `$CODEX_HOME/archived_sessions/` without the `YYYY/MM/DD` directory nesting that regular threads do, so we needed to update core's rollout listing code to deal with the flat structure.


owenlin0:
@codex review

aibrahim-oai (approved):

PR #9572, opened 2026-01-21
feat: publish config schema on release

sayan-oai:
Follow up to #8956; publish schema on new release to stable URL.

Also canonicalize schema (sort keys) when writing. This avoids reliance on default `schema_rs` behavior and makes the schema easier to read.

### DIFF
--- a/.github/workflows/rust-release.yml
+++ b/.github/workflows/rust-release.yml
@@ -434,6 +434,15 @@ jobs:
           # (e.g. -alpha, -beta). Otherwise publish a normal release.
           prerelease: ${{ contains(steps.release_name.outputs.name, '-') }}
 
+      - name: Trigger developers.openai.com deploy
+      # Only trigger the deploy if the release is not a pre-release.
+      # The deploy is used to update the developers.openai.com website with the new config schema json file.
+        if: ${{ success() && !contains(steps.release_name.outputs.name, '-') }}
+        env:
+          DEV_WEBSITE_VERCEL_DEPLOY_HOOK_URL: ${{ secrets.DEV_WEBSITE_VERCEL_DEPLOY_HOOK_URL }}
+        run: |
+          curl -sS -X POST "$DEV_WEBSITE_VERCEL_DEPLOY_HOOK_URL"
+
       - uses: facebook/dotslash-publish-release@v2
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/rust-release.yml
+++ b/.github/workflows/rust-release.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-      - uses: dtolnay/rust-toolchain@1.90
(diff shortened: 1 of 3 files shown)


.github/workflows/rust-release.yml
```diff
@@ -45,6 +46,14 @@ jobs:
           echo "✅  Tag and Cargo.toml agree (${tag_ver})"
           echo "::endgroup::"
 
+      - name: Verify config schema fixture
```
joshka-oai, line 49: Maybe something here in the name that helps devs understand what needs to be run to update when this fails (or a comment in the CI that helps point the right direction). Make it easy enough that codex can fix it automatically.

If it's not already there, add an instruction in AGENTS.md about how to keep this up to date (haven't checked if this exists - it may already be the case).

.github/workflows/rust-release.yml
```diff
@@ -421,6 +434,15 @@ jobs:
...
           prerelease: ${{ contains(steps.release_name.outputs.name, '-') }}
 
+      - name: Trigger developers.openai.com deploy
+      # Only trigger the deploy if the release is not a pre-release.
+      # The deploy is used to update the developers.openai.com website with the new config schema json file.
+        if: ${{ success() && !contains(steps.release_name.outputs.name, '-') }}
+        env:
+          DEV_WEBSITE_VERCEL_DEPLOY_HOOK_URL: ${{ secrets.DEV_WEBSITE_VERCEL_DEPLOY_HOOK_URL }}
+        run: |
+          curl --fail --show-error --silent -X POST "$DEV_WEBSITE_VERCEL_DEPLOY_HOOK_URL"
```
joshka-oai: !

sayan-oai: assuming this is in relation to our discussion about the secret possibly leaking. redirected output to `/dev/null` to avoid, and I believe GH will auto-redact any value pulled from `secrets` anyways.

.github/workflows/rust-release.yml
sayan-oai: AGENTS.md already includes a note about updating the schema via `just write-config-schema` when `ConfigToml` types change.

I added an echo in that CI step nudging users/Codex to run `just write-config-schema` if changes are intentional.

In practice differences between the generated schema and the checked-in fixture would be caught by the test in CI before the release runs anyways.

joshka-oai (approved):

.github/workflows/rust-release.yml
```diff
@@ -421,6 +435,15 @@ jobs:
           # (e.g. -alpha, -beta). Otherwise publish a normal release.
           prerelease: ${{ contains(steps.release_name.outputs.name, '-') }}
 
+      - name: Trigger developers.openai.com deploy
```
joshka-oai: Should this be after the facebook/dotslash-publish-release@v2 call? I.e. if this fails, should it block the publish?

sayan-oai: fixed, good catch!

sayan-oai:
@codex review this

sayan-oai:
@codex review this

PR #9600, opened 2026-01-21
feat: async shell snapshot

jif-oai:


codex-rs/core/src/shell.rs
```diff
@@ -14,12 +14,16 @@ pub enum ShellType {
...
     pub(crate) shell_type: ShellType,
     pub(crate) shell_path: PathBuf,
-    #[serde(skip_serializing, skip_deserializing, default)]
-    pub(crate) shell_snapshot: Option<Arc<ShellSnapshot>>,
+    #[serde(
+        skip_serializing,
+        skip_deserializing,
+        default = "empty_shell_snapshot_receiver"
+    )]
+    pub(crate) shell_snapshot: watch::Receiver<Option<Arc<ShellSnapshot>>>,
```
dylan-hurd-oai, line 26: should we abstract this away more from consumers?

codex-rs/core/src/tools/runtimes/mod.rs
```diff
@@ -54,7 +54,7 @@ pub(crate) fn maybe_wrap_shell_lc_with_snapshot(
     command: &[String],
     session_shell: &Shell,
 ) -> Vec<String> {
-    let Some(snapshot) = &session_shell.shell_snapshot else {
+    let Some(snapshot) = session_shell.shell_snapshot.borrow().clone() else {
```
dylan-hurd-oai: could we make this a method on `Shell` instead?

dylan-hurd-oai (approved):
approach lgtm but feels like we should have an abstraction in Shell instead making consumers deal with the receiver so directly

PR #9609, opened 2026-01-21
chore: defensive shell snapshot

jif-oai:
This PR adds 2 defensive mechanisms for shell snapshotting:
* Filter out invalid env variables (containing `-` for example) without dropping the whole snapshot
* Validate the snapshot before considering it as valid by running a mock command with a shell snapshot

### DIFF
--- a/codex-rs/core/src/shell_snapshot.rs
+++ b/codex-rs/core/src/shell_snapshot.rs
@@ -74,7 +74,7 @@ impl ShellSnapshot {
         });
 
         // Make the new snapshot.
-        match write_shell_snapshot(shell.shell_type.clone(), &path).await {
+        let snapshot = match write_shell_snapshot(shell.shell_type.clone(), &path).await {
             Ok(path) => {
                 tracing::info!("Shell snapshot successfully created: {}", path.display());
                 Some(Self { path })
@@ -86,7 +86,16 @@ impl ShellSnapshot {
                 );
                 None
             }
+        };
+
+        if let Some(snapshot) = snapshot.as_ref()
+            && let Err(err) = validate_snapshot(shell, &snapshot.path).await
+        {
+            tracing::error!("Shell snapshot validation failed: {err:?}");
+            return None;
         }
+
+        snapshot
     }
 }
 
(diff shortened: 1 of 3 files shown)


jif-oai:
@codex review

codex-rs/core/src/shell_snapshot.rs
```diff
@@ -86,7 +86,16 @@ impl ShellSnapshot {
                 );
                 None
             }
+        };
+
+        if let Some(snapshot) = snapshot.as_ref()
+            && let Err(err) = validate_snapshot(shell, &snapshot.path).await
+        {
+            tracing::error!("Shell snapshot validation failed: {err:?}");
+            return None;
```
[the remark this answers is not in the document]
jif-oai: This is not true. `snapshot` is an `Option<ShellSnapshot>` which has a custom `Drop` implementation

codex-rs/core/src/shell_snapshot.rs
```diff
@@ -86,7 +86,16 @@ impl ShellSnapshot {
                 );
                 None
             }
+        };
+
+        if let Some(snapshot) = snapshot.as_ref()
+            && let Err(err) = validate_snapshot(shell, &snapshot.path).await
```
pakrym-oai, line 92: Isn't this just hiding issues? Snapshots would be silently failing and nobody would know.

pakrym-oai (approved):

codex-rs/core/tests/suite/user_shell_cmd.rs
```diff
@@ -129,7 +130,10 @@ async fn user_shell_cmd_can_be_interrupted() {
 #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
 async fn user_shell_command_history_is_persisted_and_shared_with_model() -> anyhow::Result<()> {
     let server = responses::start_mock_server().await;
-    let mut builder = core_test_support::test_codex::test_codex();
+    // Disable it to ease command matching.
+    let mut builder = core_test_support::test_codex::test_codex().with_config(move |config| {
```
pakrym-oai, line 134: This is just kicking the can down the road.

jif-oai: No, this test is unrelated. This is just not to have to match on the sourcing part of the command which is highly OS dependant

codex-rs/core/src/shell_snapshot.rs
jif-oai: We send metrics for this so we know. And this does not prevent users to use Codex
We can add a warning to the users but IDK to block a user from using Codex just because they have funky stuff in their `.bashrc`

codex-rs/core/tests/suite/user_shell_cmd.rs
pakrym-oai: I mean when we enable the feature by default we'll have to fix the test to account for it anyway.

PR #9621, opened 2026-01-21
fix: Fix tilde expansion to avoid absolute-path escape

tiffanycitra:
### Motivation
- Prevent inputs like `~//` or `~///etc` from expanding to arbitrary absolute paths (e.g. `/`) because `Path::join` discards the left side when the right side is absolute, which could allow config values to escape `HOME` and broaden writable roots.

### Description
- In `codex-rs/utils/absolute-path/src/lib.rs` update `maybe_expand_home_directory` to trim leading separators from the suffix and return `home` when the remainder is empty so tilde expansion stays rooted under `HOME`.
- Add a non-Windows unit test `home_directory_double_slash_on_non_windows_is_expanded_in_deserialization` that validates `"~//code"` expands to `home.join("code")`.

### Testing
- Ran `just fmt` successfully.
- Ran `just fix -p codex-utils-absolute-path` (Clippy autofix) successfully.
- Ran `cargo test -p codex-utils-absolute-path` and all tests passed.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_i_697007481cac832dbeb1ee144d1e4cbe)

### DIFF
--- a/codex-rs/utils/absolute-path/src/lib.rs
+++ b/codex-rs/utils/absolute-path/src/lib.rs
@@ -33,6 +33,10 @@ impl AbsolutePathBuf {
                 return home;
             }
             if let Some(rest) = path_str.strip_prefix("~/") {
+                let rest = rest.trim_start_matches('/');
+                if rest.is_empty() {
+                    return home;
+                }
                 return home.join(rest);
             }
         }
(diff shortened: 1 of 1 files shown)


etraut-openai (approved):
Good catch!

PR #9624, opened 2026-01-21
forgot to add some windows sandbox nux events.

iceweasel-oai:


jif-oai (approved):
tui2 will be drop, no need to maintain it anymore

PR #9625, opened 2026-01-21
TUI: rank prefix matches ahead of fuzzy results in slash command popup

surajs-oai:
Summary:
- Show prefix-matching slash commands at the top of the popup, then append fuzzy matches.
- Keeps original command order while avoiding prefix/fuzzy collisions like /ps → /approvals.

Testing:
- cargo test -p codex-tui --quiet
- cargo test -p codex-tui2 --quiet

Issue: N/A


surajs-oai:

etraut-openai:
If we make this change here, we should probably make the same change in the Codex IDE extension. We're trying to keep the behaviors consistent across them.

surajs-oai:
Addressed in https://github.com/openai/codex/pull/9629.

PR #9630, opened 2026-01-21
Prevent backspace from removing a text element when the cursor is at the element’s left edge

charley-oai:
**Summary**
- Prevent backspace from removing a text element when the cursor is at the element’s left edge.
- Instead just delete the char before the placeholder (moving it to the left).


charley-oai:
@codex review this

aibrahim-oai (approved):

PR #9639, opened 2026-01-21
fix(ci): add stapling and notarization verification for macOS binaries

### LINKED ISSUE #9632: Codex v0.88 Homebrew Build Broken

0.88.0

Pro

N/A

Darwin 25.3.0 arm64 arm

Terminal.app but N/A

After installing the v0.88 build via Homebrew (brew upgrade), it will not launch as the system says it cannot verify it is not malware.

**

Image

**
Apple could not verify “codex-aarch64-apple-darwin” is free of malware that may harm your Mac or compromise your privacy.

1. Install or update Codex via Homebrew.
2. Try to launch it.

_No response_

### Additional information

_No response_

hdcodedev:
Fixes #9632

## Summary
- Add macOS stapling for staged binaries so the packaged artifacts retain the notarization ticket.
- Staple and validate the `dist/` macOS binaries before compression.
- Keep staging with `cp` while ensuring final artifacts are properly stapled.

## Problem
Homebrew-installed macOS binaries can fail Gatekeeper checks because the packaged artifacts are missing stapled notarization tickets.

## Solution
Staple the staged `dist/` binaries directly and validate them prior to compression, ensuring the distributed artifacts retain the ticket without relying on xattr-preserving copy.

## Testing
- CI will validate in release workflow

### DIFF
--- a/.github/workflows/rust-release.yml
+++ b/.github/workflows/rust-release.yml
@@ -225,6 +225,27 @@ jobs:
             cp target/${{ matrix.target }}/release/codex-${{ matrix.target }}.dmg "$dest/codex-${{ matrix.target }}.dmg"
           fi
 
+      - if: ${{ runner.os == 'macOS' }}
+        name: Staple and verify notarization (staged binaries)
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          dest="dist/${{ matrix.target }}"
+          for binary in codex codex-responses-api-proxy; do
+            path="${dest}/${binary}-${{ matrix.target }}"
+            echo "Stapling: ${binary}"
+            xcrun stapler staple "$path"
+
+            if ! xcrun stapler validate "$path" >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


hdcodedev:
@codex review

shijie-oai (commented):
We have attempted to staple the ticket but Apple only allows stapling for dmg or app and cannot do it directly for binary.

hdcodedev:
@shijie-oai  Thanks for pointing this out. Maybe we could update the Homebrew release process to use the DMG so the ticket is preserved

shijie-oai:
We need to do more consideration around that - ideally we would like to use the same distro for both npm and homebrew. Also homebrew is not ideal for dmg distribution. Are you running into specific issues here without stapling?

hdcodedev:
That makes sense. No, I'm not running into specific issues personally, I was mainly trying to address a user report in #9632.

PR #9647, opened 2026-01-22
Add MCP server `scopes` config and use it as fallback for OAuth login

blevy-oai:
### Motivation
- Allow MCP OAuth flows to request scopes defined in `config.toml` instead of requiring users to always pass `--scopes` on the CLI.  CLI/remote parameters should still override config values.

### Description
- Add optional `scopes: Option<Vec<String>>` to `McpServerConfig` and `RawMcpServerConfig`, and propagate it through deserialization and the built config types.
- Serialize `scopes` into the MCP server TOML via `serialize_mcp_server_table` in `core/src/config/edit.rs` and include `scopes` in the generated config schema (`core/config.schema.json`).
- CLI: update `codex-rs/cli/src/mcp_cmd.rs` `run_login` to fall back to `server.scopes` when the `--scopes` flag is empty, with explicit CLI scopes still taking precedence.
- App server: update `codex-rs/app-server/src/codex_message_processor.rs` `mcp_server_oauth_login` to use `params.scopes.or_else(|| server.scopes.clone())` so the RPC path also respects configured scopes.
- Update many test fixtures to initialize the new `scopes` field (set to `None`) so test code builds with the new struct field.

### Testing
- Ran config tooling and formatters: `just write-config-schema` (succeeded), `just fmt` (succeeded), and `just fix -p codex-core`, `just fix -p codex-cli`, `just fix -p codex-app-server` (succeeded where applicable). 
- Ran unit tests for the CLI: `cargo test -p codex-cli` (passed).
- Ran unit tests for core: `cargo test -p codex-core` (ran; many tests passed but several failed, including model refresh/403-related tests, shell snapshot/timeouts, and several `unified_exec` expectations). 
- Ran app-server tests: `cargo test -p codex-app-server` (ran; many integration-suite tests failed due to mocked/remote HTTP 401/403 responses and wiremock expectations). 

If you want, I can split the tests into smaller focused runs or help debug the failing integration tests (they appear to be unrelated to the config change and stem from external HTTP/mocking behaviors encountered during the test runs).

------
[Codex Task](https://chatgpt.com/codex/tasks/task_i_69718f505914832ea1f334b3ba064553)

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -2490,6 +2490,8 @@ impl CodexMessageProcessor {
             }
         };
 
+        let scopes = scopes.or_else(|| server.scopes.clone());
+
         match perform_oauth_login_return_url(
             &name,
             &url,
--- a/codex-rs/cli/src/mcp_cmd.rs
+++ b/codex-rs/cli/src/mcp_cmd.rs
@@ -247,6 +247,7 @@ async fn run_add(config_overrides: &CliConfigOverrides, add_args: AddArgs) -> Re
         tool_timeout_sec: None,
         enabled_tools: None,
         disabled_tools: None,
+        scopes: None,
     };
 
     servers.insert(name.clone(), new_entry);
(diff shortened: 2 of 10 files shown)


blevy-oai:

blevy-oai:
With a config.toml of 

```
[mcp_servers.mcp-auth-test]
enabled = true
url = "<my url>"
scopes = ["api://.../all", "offline_access"]
```
I'm able to `codex mcp login mcp-auth-test` without problems whereas prior to this change `codex mcp login mcp-auth-test` fails on the redirect flow because no scopes are passed (the --scopes was the only way to pass this info previously which makes the login flow more annoying/ frictionful for users)

gpeal (approved):

PR #9653, opened 2026-01-22
Fix: Lower log level for closed-channel send

Kbediako:
## What?
- Downgrade the closed-channel send error log to debug in `codex-rs/core/src/codex.rs`.

## Why?
- `async_channel::Sender::send` only fails when the channel is closed, so the current error-level log is noisy during normal shutdown. See issue #9652.

## How?
- Replace the error log with a debug log on send failure.

## Tests
- `just fmt`
- `just fix -p codex-core`
- `cargo test -p codex-core`


Kbediako:

etraut-openai (approved):

PR #9654, opened 2026-01-22
Show OAuth error descriptions in callback responses

blevy-oai:
### Motivation
- The local OAuth callback server returned a generic "Invalid OAuth callback" on failures even when the query contained an `error_description`, making it hard to debug OAuth failures.

### Description
- Update `codex-rs/rmcp-client/src/perform_oauth_login.rs` to surface `error_description` values from the callback query in the HTTP response.
- Introduce a `CallbackOutcome` enum and change `parse_oauth_callback` to return it, parsing `code`, `state`, and `error_description` from the query string.
- Change `spawn_callback_server` to match on `CallbackOutcome` and return `OAuth error: <description>` with a 400 status when `error_description` is present, while preserving the existing success and invalid flows.
- Use inline formatting for the error response string.

### Testing
- Ran `just fmt` in the `codex-rs` workspace to format changes successfully.
- Ran `cargo test -p codex-rmcp-client` and all tests passed.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_i_6971aadc68d0832e93159efea8cd48a9)

### DIFF
--- a/codex-rs/rmcp-client/src/perform_oauth_login.rs
+++ b/codex-rs/rmcp-client/src/perform_oauth_login.rs
@@ -103,21 +103,32 @@ fn spawn_callback_server(server: Arc<Server>, tx: oneshot::Sender<(String, Strin
     tokio::task::spawn_blocking(move || {
         while let Ok(request) = server.recv() {
             let path = request.url().to_string();
-            if let Some(OauthCallbackResult { code, state }) = parse_oauth_callback(&path) {
-                let response =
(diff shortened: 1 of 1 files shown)


blevy-oai:
Flows that fail with an error_description in the query now have a nice informative error in the body which is displayed in the browser.  This makes debugging failures easier as now you don't have to copy the path out of the url bar and inspect it in a text editor

gpeal (approved):

PR #9658, opened 2026-01-22
fix(mcp-server): Support resuming threads across server restarts

sorryhyun:
## What

Fix `codex-reply` MCP tool to support resuming threads across server restarts.

## Why

Previously, `codex-reply` only looked up threads in memory via `get_thread()`. When the MCP server restarted, all threads were lost even though rollout files existed on disk. This differed from `codex exec --resume` which correctly loads sessions from disk.

## How

Modified `handle_tool_call_codex_session_reply` in `message_processor.rs` to fall back to disk lookup:

1. First tries `get_thread()` for same-session lookups (fast path)
2. If not found, searches for the rollout file using `find_thread_path_by_id_str()`
3. If found on disk, loads it with `resume_thread_from_rollout()`

## Checks

- [x] `just fmt`
- [x] `just fix -p codex-mcp-server`

### DIFF
--- a/codex-rs/mcp-server/src/message_processor.rs
+++ b/codex-rs/mcp-server/src/message_processor.rs
@@ -15,6 +15,7 @@ use codex_core::ThreadManager;
 use codex_core::config::Config;
 use codex_core::default_client::USER_AGENT_SUFFIX;
 use codex_core::default_client::get_codex_user_agent;
+use codex_core::find_thread_path_by_id_str;
 use codex_core::protocol::Submission;
 use mcp_types::CallToolRequestParams;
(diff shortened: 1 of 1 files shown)


etraut-openai:
Our [contribution guidelines](https://github.com/openai/codex/blob/main/docs/contributing.md) state that we're not accepting feature contributions at this time. We need to make sure that all new features are well designed, compose well with both existing and upcoming features, fit into our roadmap, and are consistent across all Codex surfaces (CLI, IDE extension, web). If you would like to propose a new feature, please file or upvote an enhancement request in the issue tracker. We will generally prioritize new features based on community feedback.

PR #9667, opened 2026-01-22
[connectors] Support connectors part 1 - App server & MCP

mzeng-openai:
In order to make Codex work with connectors, we add a built-in gateway MCP that acts as a transparent proxy between the client and the connectors. The gateway MCP collects actions that are accessible to the user and sends them down to the user, when a connector action is chosen to be called, the client invokes the action through the gateway MCP as well.

 - [x] Add the system built-in gateway MCP to list and run connectors.
 - [x] Add the app server methods and protocol


codex-rs/chatgpt/src/connectors.rs
```diff
@@ -0,0 +1,125 @@
...
+enum PrincipalType {
+    User,
+}
+
+#[derive(Debug, Deserialize)]
+struct ListConnectorsResponse {
+    connectors: Vec<ConnectorInfo>,
+}
+
+pub async fn list_connectors(config: &Config) -> anyhow::Result<Vec<ConnectorInfo>> {
```
mzeng-openai, line 39: There are two types of connectors:
1. Connectors that are already connected by the user, and we have a link that we can use to authenticate it
2. Connectors that have not been installed by the user. We need the user to install it on ChatGPT UI.
In the UI we mix them together and use is_accessible to distinguish them. This methods does the merge.

codex-rs/chatgpt/src/connectors.rs
```diff
@@ -0,0 +1,125 @@
...
+    let (connectors_result, accessible_result) = tokio::join!(
+        list_all_connectors(config),
+        list_accessible_connectors_from_mcp_tools(config),
+    );
+    let connectors = connectors_result?;
+    let accessible = accessible_result?;
+    Ok(merge_connectors(connectors, accessible))
+}
+
+pub async fn list_all_connectors(config: &Config) -> anyhow::Result<Vec<ConnectorInfo>> {
```
mzeng-openai, line 52: List all available connectors regardless of whether they are installed or not.

codex-rs/core/src/mcp/mod.rs
```diff
@@ -9,16 +9,135 @@ use codex_protocol::protocol::SandboxPolicy;
...
+            CODEX_APPS_MCP_SERVER_NAME.to_string(),
+            codex_apps_mcp_server_config(config, auth),
+        );
+    } else {
+        servers.remove(CODEX_APPS_MCP_SERVER_NAME);
+    }
+    servers
+}
+
+pub(crate) fn effective_mcp_servers(
```
mzeng-openai, line 120: Add codex apps MCP as an always on built-in MCP.

codex-rs/core/src/codex.rs
```diff
@@ -2950,6 +2972,60 @@ async fn run_auto_compact(sess: &Arc<Session>, turn_context: &Arc<TurnContext>)
...
+    input: &[ResponseItem],
+) -> Vec<connectors::ConnectorInfo> {
+    let user_messages = collect_user_messages(input);
+    if user_messages.is_empty() {
+        return Vec::new();
+    }
+
+    connectors
+        .into_iter()
+        .filter(|connector| connector_inserted_in_messages(connector, &user_messages))
```
mzeng-openai, line 2986: Include connectors mentioned in the user messages as tools.

codex-rs/core/src/codex.rs
```diff
@@ -2950,6 +2972,60 @@ async fn run_auto_compact(sess: &Arc<Session>, turn_context: &Arc<TurnContext>)
...
+    let label = connectors::connector_display_label(connector);
+    let needle = label.to_lowercase();
+    let legacy = format!("{label} connector").to_lowercase();
+    user_messages.iter().any(|message| {
+        let message = message.to_lowercase();
+        message.contains(&needle) || message.contains(&legacy)
+    })
+}
+
+fn filter_codex_apps_mcp_tools(
```
mzeng-openai, line 3003: Find the mcp tools that are associated with connectors.

codex-rs/core/src/connectors.rs
```diff
@@ -0,0 +1,227 @@
...
+    pub connector_description: Option<String>,
+    #[serde(default, rename = "logo_url")]
+    pub logo_url: Option<String>,
+    #[serde(default, rename = "install_url")]
+    pub install_url: Option<String>,
+    #[serde(default)]
+    pub is_accessible: bool,
+}
+
+pub async fn list_accessible_connectors_from_mcp_tools(
```
mzeng-openai, line 36: We collect the actually accessible mcp tools from the gateway MCP tool list.

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -965,6 +965,43 @@ pub struct ListMcpServerStatusResponse {
     pub next_cursor: Option<String>,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Default, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ConnectorsListParams {}
```
owenlin0: can we support pagination from the get-go? to make it consistent with our other `*/list` APIs. See `model/list` and `thread/list` as examples

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -965,6 +965,43 @@ pub struct ListMcpServerStatusResponse {
...
+    #[serde(rename = "id")]
+    pub connector_id: String,
+    #[serde(rename = "name")]
+    pub connector_name: String,
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        rename = "description"
+    )]
+    #[ts(optional)]
```
owenlin0: remove all these annotations

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -965,6 +965,43 @@ pub struct ListMcpServerStatusResponse {
...
+    pub connector_description: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
```
owenlin0: and here

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -965,6 +965,43 @@ pub struct ListMcpServerStatusResponse {
...
+    pub logo_url: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
```
owenlin0: here too

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -965,6 +965,43 @@ pub struct ListMcpServerStatusResponse {
...
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ConnectorsListParams {}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ConnectorInfo {
+    #[serde(rename = "id")]
+    pub connector_id: String,
```
owenlin0: can we just name the field `id` so we don't need to use `serde(rename)`?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -965,6 +965,43 @@ pub struct ListMcpServerStatusResponse {
...
+    #[serde(rename = "name")]
+    pub connector_name: String,
```
owenlin0: call this `name`

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -965,6 +965,43 @@ pub struct ListMcpServerStatusResponse {
...
+    #[ts(optional)]
+    pub install_url: Option<String>,
+    #[serde(default)]
+    pub is_accessible: bool,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ConnectorsListResponse {
```
owenlin0: let's update the response to have a `next_cursor` field to be consistent with the `model/list` and `thread/list` responses

codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -133,6 +133,10 @@ client_request_definitions! {
         params: v2::SkillsListParams,
         response: v2::SkillsListResponse,
     },
+    ConnectorsList => "connectors/list" {
```
owenlin0: nit: `connector/list` (singular)

I know it reads a bit funny, but it will be consistent with other APIs (skills is kind of an exception that I didn't catch)

```
thread/list
thread/start
thread/archive
etc.
```

owenlin0 (commented):
bunch of API nits

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -965,6 +965,43 @@ pub struct ListMcpServerStatusResponse {
...
+    pub connector_id: String,
+    #[serde(rename = "name")]
+    pub connector_name: String,
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        rename = "description"
+    )]
+    #[ts(optional)]
+    pub connector_description: Option<String>,
```
owenlin0: name this `description`

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -965,6 +965,43 @@ pub struct ListMcpServerStatusResponse {
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub logo_url: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub install_url: Option<String>,
+    #[serde(default)]
+    pub is_accessible: bool,
```
owenlin0, line 988: is there a future where connector state evolves to more than just `is_accessible: true/false`? If so, consider an enum

codex-rs/app-server/README.md
```diff
@@ -88,6 +88,7 @@ Example (from OpenAI's official VSCode extension):
 - `model/list` — list available models (with reasoning effort options).
 - `collaborationMode/list` — list available collaboration mode presets (experimental, no pagination).
 - `skills/list` — list skills for one or more `cwd` values (optional `forceReload`).
+- `connectors/list` — list available connectors.
```
owenlin0: connectors are going to be super useful!! would love to have a section focused on connectors (explaining what they are, how codex uses them, can link out to any official openai docs, etc.), like we have one for skills: https://github.com/openai/codex/blob/main/codex-rs/app-server/README.md#skills

doesn't have to be this PR, but it'd be very useful since lots of people are starting to reference this README.md :)

codex-rs/app-server-protocol/src/protocol/v2.rs
owenlin0: maybe go with an enum to be future proof and have it be `installed | notInstalled`?

codex-rs/app-server/README.md
mzeng-openai: That's a good point! Our current goal is to unify MCPs/Connectors/Skills into one coherent App concept. We'll have a more simplified story here soon, i'll update it once it's finalized.

codex-rs/app-server-protocol/src/protocol/common.rs
owenlin0: is there a brand name for this feature? Consider the app-server API public-facing. I noticed it's called `## Apps (Connectors)` in docs/config.md so wondering if it should be something other than `connector/list`, like `chatgptApp/list` or something

owenlin0: ah, i think so: https://help.openai.com/en/articles/11487775-apps-in-chatgpt

codex-rs/app-server-protocol/src/protocol/v2.rs
mzeng-openai: it's either accessible or not accessible so we should be good.

codex-rs/app-server-protocol/src/protocol/common.rs
mzeng-openai: renamed to apps

codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -133,6 +133,10 @@ client_request_definitions! {
         params: v2::SkillsListParams,
         response: v2::SkillsListResponse,
     },
+    AppsList => "app/list" {
```
apanasenko-oai, line 136: nit: `apps/list` similar to `skills/list`

mzeng-openai: ironically @owenlin0 just said the other way, because all the other protocols are singular.

apanasenko-oai (approved):

codex-rs/app-server-protocol/src/protocol/common.rs
apanasenko-oai: it reads very strangely: `App*s*List => "app/list"`

owenlin0: it does read strangely, but only for `/list` methods

```
thread/list
thread/start
thread/archive
... etc.
```

owenlin0, line 136: nit: `AppsList` -> `AppList`

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -965,6 +965,39 @@ pub struct ListMcpServerStatusResponse {
...
+    /// Opaque pagination cursor returned by a previous call.
+    pub cursor: Option<String>,
+    /// Optional page size; defaults to a reasonable server-side value.
+    pub limit: Option<u32>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct AppInfo {
```
owenlin0, line 981: nit: should we just call this `App`? `AppInfo` sounds redundant (like `ThreadInfo` for a `thread/list` API)

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -965,6 +965,39 @@ pub struct ListMcpServerStatusResponse {
...
+    pub logo_url: Option<String>,
+    pub install_url: Option<String>,
+    #[serde(default)]
+    pub is_accessible: bool,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct AppsListResponse {
```
owenlin0, line 994: `AppListResponse`

owenlin0 (approved):
should we also add some integration tests at `codex-rs/app-server/tests/suite/v2/app_list.rs`? can probably ask codex to generate some

codex-rs/app-server-protocol/src/protocol/v2.rs
mzeng-openai: I'm mostly trying to avoid confusing this with the actual codex app itself.

owenlin0: it's ok, you get to own the word "app" in the app-server API

PR #9691, opened 2026-01-22
Add hooks implementation and wire up to `notify`

gt-oai:
This introduces a `Hooks` service. It registers hooks from config and dispatches hook events at runtime. 

N.B. The hook config is not wired up to this yet. But for legacy reasons, we wire up `notify` from config and power it using hooks now. Nothing about the `notify` interface has changed.

I'd start by reviewing `hooks/types.rs`

Some things to note:
  - hook names subject to change
  - no hook result yet
  - stopping semantics yet to be introduced
  - additional hooks yet to be introduced


jif-oai:
@codex review

codex-rs/core/src/hooks/types.rs
```diff
@@ -0,0 +1,72 @@
+use std::sync::Arc;
+
+use futures::future::BoxFuture;
+use serde::Serialize;
+
+pub(crate) type HookFn = Arc<dyn Fn(HookPayload) -> BoxFuture<'static, HookOutcome> + Send + Sync>;
```
jif-oai: Intuitively it would be better for hooks to take a ref to avoid a pre-clone while it might not be necessary but it's a nit

codex-rs/core/src/hooks/types.rs
```diff
@@ -0,0 +1,72 @@
...
+#[serde(rename_all = "snake_case")]
+pub(crate) struct HookEventAfterAgent {
+    pub thread_id: String,
+    pub turn_id: String,
+    pub input_messages: Vec<String>,
+    pub last_assistant_message: Option<String>,
+}
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(tag = "event_type", rename_all = "PascalCase")]
```
jif-oai: Why PascalCase only for this?

codex-rs/core/src/hooks/types.rs
```diff
@@ -0,0 +1,72 @@
...
+pub(crate) enum HookEvent {
+    AfterAgent {
+        #[serde(flatten)]
+        event: HookEventAfterAgent,
+    },
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub(crate) enum HookEventKind {
```
jif-oai: Do you expect to have the same event for different kind? Why having 2 different enums?

codex-rs/core/src/hooks/mod.rs
```diff
@@ -0,0 +1,317 @@
...
+    config
+        .notify
+        .as_ref()
+        .filter(|argv| !argv.is_empty() && !argv[0].is_empty())
+        .map(|argv| notify_hook(argv.clone()))
+}
+
+// Hooks are arbitrary, user-specified functions that are deterministically
+// executed after specific events in the Codex lifecycle.
+impl Hooks {
```
jif-oai: Why having this code here? In general in the `mod.rs` we have the struct and trait and then the implementation in the corresponding files of the module

codex-rs/core/src/hooks/mod.rs
```diff
@@ -0,0 +1,317 @@
...
+        for hook in hooks {
+            let outcome = hook.execute(hook_payload.clone()).await;
+            if matches!(outcome, HookOutcome::Stop) {
+                break;
+            }
+        }
+    }
+}
+
+pub(super) fn command_from_argv(argv: &[String]) -> Option<Command> {
```
jif-oai: Shouldn't we log or bubble up something if this is None?

codex-rs/core/src/hooks/types.rs
```diff
@@ -0,0 +1,72 @@
...
+        (self.func)(payload).await
+    }
+}
+
+#[derive(Debug, Serialize, Clone)]
+#[serde(rename_all = "snake_case")]
+pub(crate) struct HookPayload {
+    pub(crate) session_id: String,
+    pub(crate) cwd: String,
+    pub(crate) triggered_at: String,
```
jif-oai: Why not strongly typing this?

codex-rs/core/src/hooks/types.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub(super) async fn execute(&self, payload: HookPayload) -> HookOutcome {
+        (self.func)(payload).await
+    }
+}
+
+#[derive(Debug, Serialize, Clone)]
+#[serde(rename_all = "snake_case")]
+pub(crate) struct HookPayload {
+    pub(crate) session_id: String,
+    pub(crate) cwd: String,
```
jif-oai: Why not strongly typing this?

codex-rs/core/src/hooks/mod.rs
```diff
@@ -0,0 +1,317 @@
...
+        }
+    }
+
+    pub(crate) async fn dispatch(&self, hook_payload: HookPayload) {
+        // TODO(gt): support interrupting program execution by returning a result here.
+        let Some(hooks) = self.hooks.get(&hook_payload.hook_event.kind()) else {
+            return;
+        };
+
+        for hook in hooks {
```
jif-oai: The order matter so we should probably document it at some point

codex-rs/core/src/hooks/mod.rs
```diff
@@ -0,0 +1,317 @@
...
+// Hooks are arbitrary, user-specified functions that are deterministically
+// executed after specific events in the Codex lifecycle.
+impl Hooks {
+    // new creates a new Hooks instance from config.
+    // For legacy compatibility, if config.notify is set, it will be added to
+    // the after_agent hooks.
+    pub(crate) fn new(config: &Config) -> Self {
+        let after_agent_hooks = get_notify_hook(config).into_iter().collect();
+        Self {
+            hooks: HashMap::from([(HookEventKind::AfterAgent, after_agent_hooks)]),
```
jif-oai: We have a finite set of possible hooks (after/before agent, after/before session, ...). Instead of using a HashMap with a finite enum, should we have a well defined structure? It would make this easier to read and to document

codex-rs/core/src/hooks/mod.rs
gt-oai: This one feels ok to ignore

codex-rs/core/src/hooks/mod.rs
gt-oai: will document in the docs

codex-rs/core/src/hooks/mod.rs
gt-oai: removed the map and replaced with strongly-typed struct field.

codex-rs/core/src/codex.rs
```diff
@@ -3776,14 +3777,21 @@ pub(crate) async fn run_turn(
...
-                    sess.notifier()
-                        .notify(&UserNotification::AgentTurnComplete {
-                            thread_id: sess.conversation_id.to_string(),
-                            turn_id: turn_context.sub_id.clone(),
-                            cwd: turn_context.cwd.display().to_string(),
-                            input_messages: sampling_request_input_messages,
-                            last_assistant_message: last_agent_message.clone(),
-                        });
+                    sess.hooks()
+                        .dispatch(crate::hooks::HookPayload {
```
jif-oai, line 3781: Maybe we should think about a timeout on those hooks. Otherwise a badly configured hook will prevent the turn from finishing. Ideally we can make it configurable

Can be in a follow-up PR

codex-rs/core/src/hooks/registry.rs
```diff
@@ -0,0 +1,315 @@
...
+    fn hooks_for_event(&self, hook_event: &HookEvent) -> &[Hook] {
+        match hook_event {
+            HookEvent::AfterAgent { .. } => &self.after_agent,
+        }
+    }
+
+    pub(crate) async fn dispatch(&self, hook_payload: HookPayload) {
+        // TODO(gt): support interrupting program execution by returning a result here.
+        for hook in self.hooks_for_event(&hook_payload.hook_event) {
+            let outcome = hook.execute(&hook_payload).await;
```
jif-oai, line 43: It would be cool to measure and log hook failures

codex-rs/core/src/hooks/registry.rs
```diff
@@ -0,0 +1,315 @@
...
+            if matches!(outcome, HookOutcome::Stop) {
```
jif-oai, line 44: I guess it's on purpose that the `Stop` does not actually stop anything but the dispatch of the other hooks?

jif-oai (approved):

codex-rs/core/src/codex.rs
gt-oai: yup timeout is in the design doc

codex-rs/core/src/hooks/registry.rs
gt-oai: not right now, will propagate the error up in future PRs

PR #9698, opened 2026-01-22
Support end_turn flag

pakrym-oai:
Experimental flag that signals the end of the turn.

### DIFF
--- a/codex-rs/core/dfa_unless_trusted_5_1.txt
+++ b/codex-rs/core/dfa_unless_trusted_5_1.txt
@@ -1,1 +1,0 @@
-danger-unless-trusted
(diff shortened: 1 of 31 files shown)


codex-rs/core/dfa_unless_trusted_5_1.txt
```diff
@@ -1 +0,0 @@
-danger-unless-trusted
```
aibrahim-oai, line 1: ?

aibrahim-oai (approved):

codex-rs/core/dfa_unless_trusted_5_1.txt
pakrym-oai: test artifact of sorts can look later

PR #9707, opened 2026-01-22
bundle sandbox helper binaries in main zip, for winget.

iceweasel-oai:
Winget uses the main codex.exe value as its target.
The elevated sandbox requires these two binaries to live next to codex.exe

### DIFF
--- a/.github/workflows/rust-release.yml
+++ b/.github/workflows/rust-release.yml
@@ -285,7 +285,28 @@ jobs:
             # Must run from inside the dest dir so 7z won't
             # embed the directory path inside the zip.
             if [[ "${{ matrix.runner }}" == windows* ]]; then
-              (cd "$dest" && 7z a "${base}.zip" "$base")
+              if [[ "$base" == "codex-${{ matrix.target }}.exe" ]]; then
+                # Bundle the sandbox helper binaries into the main codex zip so
+                # WinGet installs include the required helpers next to codex.exe.
+                # Fall back to the single-binary zip if the helpers are missing
+                # to avoid breaking releases.
+                bundle_dir="$(mktemp -d)"
+                runner_src="$dest/codex-command-runner-${{ matrix.target }}.exe"
+                setup_src="$dest/codex-windows-sandbox-setup-${{ matrix.target }}.exe"
+                if [[ -f "$runner_src" && -f "$setup_src" ]]; then
+                  cp "$dest/$base" "$bundle_dir/$base"
+                  cp "$runner_src" "$bundle_dir/codex-command-runner.exe"
+                  cp "$setup_src" "$bundle_dir/codex-windows-sandbox-setup.exe"
+                  (cd "$bundle_dir" && 7z a "$dest/${base}.zip" .)
+                else
+                  echo "warning: missing sandbox binaries; falling back to single-binary zip"
+                  echo "warning: expected $runner_src and $setup_src"
+                  (cd "$dest" && 7z a "${base}.zip" "$base")
+                fi
+                rm -rf "$bundle_dir"
+              else
+                (cd "$dest" && 7z a "${base}.zip" "$base")
+              fi
             fi
 
             # Also create .zst (existing behaviour) *and* remove the original


.github/workflows/rust-release.yml
```diff
@@ -285,7 +285,28 @@ jobs:
             # Must run from inside the dest dir so 7z won't
             # embed the directory path inside the zip.
             if [[ "${{ matrix.runner }}" == windows* ]]; then
-              (cd "$dest" && 7z a "${base}.zip" "$base")
+              if [[ "$base" == "codex-${{ matrix.target }}.exe" ]]; then
+                # Bundle the sandbox helper binaries into the main codex zip so
+                # WinGet installs include the required helpers next to codex.exe.
+                # Fall back to the single-binary zip if the helpers are missing
+                # to avoid breaking releases.
```
bolinfest, line 292: I don't follow: are we talking about what happens on our CI builder machine? Shouldn't that be deterministic? I'm confused what "Fall back" situation we are designing for here.

iceweasel-oai: this is me not wanting to break the release, and if it fails to find the binaries (like I'm constructing the filepaths wrong) falling back to what currently happens (each binary in its own zip file)

iceweasel-oai: I plan to remove this once I'm sure this is working.

.github/workflows/rust-release.yml
```diff
@@ -285,7 +285,28 @@ jobs:
...
+                  cp "$setup_src" "$bundle_dir/codex-windows-sandbox-setup.exe"
+                  (cd "$bundle_dir" && 7z a "$dest/${base}.zip" .)
+                else
+                  echo "warning: missing sandbox binaries; falling back to single-binary zip"
+                  echo "warning: expected $runner_src and $setup_src"
+                  (cd "$dest" && 7z a "${base}.zip" "$base")
+                fi
+                rm -rf "$bundle_dir"
+              else
+                (cd "$dest" && 7z a "${base}.zip" "$base")
```
bolinfest, line 308: So this step should just be run for `"codex-windows-sandbox-setup-${{ matrix.target }}.exe"`, is that right?

If so, maybe we should make an `elif` for that right here and then a final `else` that raises an exception that `$base` does not match one of the expected patterns.

.github/workflows/rust-release.yml
```diff
@@ -285,7 +285,28 @@ jobs:
...
+                setup_src="$dest/codex-windows-sandbox-setup-${{ matrix.target }}.exe"
+                if [[ -f "$runner_src" && -f "$setup_src" ]]; then
+                  cp "$dest/$base" "$bundle_dir/$base"
+                  cp "$runner_src" "$bundle_dir/codex-command-runner.exe"
+                  cp "$setup_src" "$bundle_dir/codex-windows-sandbox-setup.exe"
+                  (cd "$bundle_dir" && 7z a "$dest/${base}.zip" .)
+                else
+                  echo "warning: missing sandbox binaries; falling back to single-binary zip"
+                  echo "warning: expected $runner_src and $setup_src"
+                  (cd "$dest" && 7z a "${base}.zip" "$base")
```
bolinfest, line 304: Why shouldn't this break the release build? Hasn't something gone horribly wrong if this is true?

iceweasel-oai: I mostly don't want to break the build while I am testing/iterating on this. Like making sure I don't have a typo or something. Once it's working, I would revert any fallbacks and then the build would break if something was missing

.github/workflows/rust-release.yml
iceweasel-oai: the bundling part of it should happen in the loop for `"codex-${{ matrix.target }}.exe"`
all other binaries should be as they are today, just the binary by itself in a zip file

bolinfest: Oh, it is both of these:

```
codex-windows-sandbox-setup-${{ matrix.target }}.exe
codex-windows-command-runner-${{ matrix.target }}.exe
```

Does each of these need to be in its own `.zip`? Do we encourage users to download these individually?

It feels like we should favor the big `codex.zip` now that it will include everything?

iceweasel-oai: yeah, both of those are being included in `codex.zip` with this change. That is what winget will use

bolinfest: The existing build script for the npm package appears to rely on a `.zst` being available for both `codex-windows-sandbox-setup` and `codex-command-runner`:

https://github.com/openai/codex/blob/c6ded0afd84ab30300aa85d4fffd41079cf16656/codex-cli/scripts/install_native_deps.py#L334-L337

so I guess we still need to publish both of these until we update that script.

bolinfest (approved):
We can land this as-is for now, but as a follow-up, I would like to see dependent build scripts rely on only these artifacts (or the `.zip` or `.tar.gz` versions) for Windows:

```
codex-x86_64-pc-windows-msvc.exe.zst
codex-aarch64-pc-windows-msvc.exe.zst
```

and then extract `codex-windows-sandbox-setup` and `codex-command-runner` from within the archive, as necessary.

We should then stop publishing all of the individual `codex-windows-sandbox-setup` and `codex-command-runner` entries from our GitHub release to make things easier to reason about.

PR #9712, opened 2026-01-22
TUI: prompt to implement plan and switch to Execute

charley-oai:
## Summary
- Replace the plan‑implementation prompt with a standard selection popup.
- “Yes” submits a user turn in Execute via a dedicated app event to preserve normal transcript behavior.
- “No” simply dismisses the popup.

Screenshot 2026-01-22 at 2 00 54 PM

## Changes
- Add a plan‑implementation popup using `SelectionViewParams`.
- Add `SubmitUserMessageWithMode` so “Yes” routes through `submit_user_message` (ensures user history + separator state).
- Track `saw_plan_update_this_turn` so the prompt appears even when only `update_plan` is emitted.
- Suppress the plan popup on replayed turns, when messages are queued, or when a rate‑limit prompt is pending.
- Add `execute_mode` helper for collaboration modes.
- Add tests for replay/queued/rate‑limit guards and plan update without final message.
- Add snapshots for both the default and “No”‑selected popup states.

### DIFF
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -922,6 +922,13 @@ impl ChatWidget {
             return;
         }
 
+        if matches!(
+            self.rate_limit_switch_prompt,
+            RateLimitSwitchPromptState::Pending
+        ) {
+            return;
+        }
+
         self.open_plan_implementation_prompt();
     }
 
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -119,6 +119,10 @@ use tokio::task::JoinHandle;
 use tracing::debug;
 
 const DEFAULT_MODEL_DISPLAY_NAME: &str = "loading";
+const PLAN_IMPLEMENTATION_TITLE: &str = "Implement this plan?";
+const PLAN_IMPLEMENTATION_YES: &str = "Yes, implement this plan";
+const PLAN_IMPLEMENTATION_NO: &str = "No, and tell the model what to do differently";
+const PLAN_IMPLEMENTATION_EXECUTE_MESSAGE: &str = "Implement the plan.";
 
 use crate::app_event::AppEvent;
 use crate::app_event::ExitMode;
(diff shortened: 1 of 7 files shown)


charley-oai:
@codex review this

codex-rs/tui/src/chatwidget.rs
```diff
@@ -894,6 +909,101 @@ impl ChatWidget {
...
+        match selected {
+            Some(PLAN_IMPLEMENTATION_OPTION_YES) => {
+                let Some(mode) = collaboration_modes::execute_mode(self.models_manager.as_ref())
+                else {
+                    self.on_warning("Execute mode is unavailable.".to_string());
+                    return;
+                };
+                self.app_event_tx
+                    .send(AppEvent::UpdateCollaborationMode(mode));
+                self.queue_user_message(PLAN_IMPLEMENTATION_EXECUTE_MESSAGE.into());
```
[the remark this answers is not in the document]
aibrahim-oai: Seems right

aibrahim-oai (commented):
Why introducing SubmitTarget? We can just send a user turn in the right mode.

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

codex-rs/tui/src/chatwidget.rs
```diff
@@ -884,6 +891,9 @@ impl ChatWidget {
         self.clear_unified_exec_processes();
         self.request_redraw();
 
+        if !from_replay && self.queued_user_messages.is_empty() {
```
aibrahim-oai, line 894: does this mean if we have queued messages we will submit first?
curious about the decision

aibrahim-oai (approved):

codex-rs/tui/src/chatwidget.rs
charley-oai: Yes if we have queued messages then just submit those rather than prompting the user whether they want to implement the plan. Otherwise queued messages wouldn't work in plan mode

aibrahim-oai: makes sense

PR #9713, opened 2026-01-22
use machine scope instead of user scope for dpapi.

iceweasel-oai:
This fixes a bug where the elevated sandbox setup encrypts sandbox user passwords as an admin user, but normal command execution attempts to decrypt them as a different user.

Machine scope allows all users to encyrpt/decrypt

this PR also moves the encrypted file to a different location .codex/.sandbox-secrets which the sandbox users cannot read.


etraut-openai:
@codex review

etraut-openai (approved):
Nice! lgtm.

PR #9718, opened 2026-01-22
feat(tui) /personality

dylan-hurd-oai:
## Summary
Adds /personality selector in the TUI, which leverages the new core interface in #9644

Notes:
- We are doing some of our own state management for model_info loading here, but not sure if that's ideal. open to opinions on simpler approach, but would like to avoid blocking on a larger refactor
- Right now, the `/personality` selector just hides when the model doesn't support it. we can update this behavior down the line

## Testing
- [x] Tested locally
- [x] Added snapshot tests


codex-rs/tui/tests/suite/no_panic_on_startup.rs
```diff
@@ -62,6 +62,10 @@ async fn run_codex_cli(
         "CODEX_HOME".to_string(),
         codex_home.as_ref().display().to_string(),
     );
+    // Force the CLI into sandboxed mode so reqwest disables system proxy detection,
+    // which can panic under macOS CI/test environments (#8912).
+    env.insert("CODEX_SANDBOX".to_string(), "seatbelt".to_string());
+    env.insert("RUST_BACKTRACE".to_string(), "1".to_string());
```
joshka-oai: This seems unrelated

joshka-oai (approved):
LGTM generally

codex-rs/tui/tests/suite/no_panic_on_startup.rs
dylan-hurd-oai: fixed

aibrahim-oai (approved):

PR #9720, opened 2026-01-22
feat: add session source as otel metadata tag

alexsong-oai:
Add session_source as global OTEL metric tags to identify client surface

### DIFF
--- a/codex-rs/otel/src/lib.rs
+++ b/codex-rs/otel/src/lib.rs
@@ -31,6 +31,7 @@ pub struct OtelEventMetadata {
     pub(crate) auth_mode: Option<String>,
     pub(crate) account_id: Option<String>,
     pub(crate) account_email: Option<String>,
+    pub(crate) session_source: String,
     pub(crate) model: String,
     pub(crate) slug: String,
     pub(crate) log_user_prompts: bool,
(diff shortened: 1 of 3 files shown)


codex-rs/otel/src/lib.rs
```diff
@@ -149,8 +150,18 @@ impl OtelManager {
         if !self.metrics_use_metadata_tags {
             return Ok(Vec::new());
         }
-        let mut tags = Vec::with_capacity(5);
+        let mut tags = Vec::with_capacity(7);
         Self::push_metadata_tag(&mut tags, "auth_mode", self.metadata.auth_mode.as_deref())?;
+        Self::push_metadata_tag(
+            &mut tags,
+            "user.account_id",
+            self.metadata.account_id.as_deref(),
```
gverma-openai: I thought we couldn't add any non-anonymized info here. Is `as_deref` obfuscating the actual user ID?

xl-openai: Good call. Also it will have high-cardinality issue.

alexsong-oai: No, it's the raw user ID. I can remove it if not allowed. cc @xl-openai

etraut-openai: Good catch. Yeah, we should not include any PII (personally identifiable information) in telemetry.

alexsong-oai: updated

codex-rs/otel/src/lib.rs
```diff
@@ -149,8 +150,13 @@ impl OtelManager {
         if !self.metrics_use_metadata_tags {
             return Ok(Vec::new());
         }
-        let mut tags = Vec::with_capacity(5);
+        let mut tags = Vec::with_capacity(6);
         Self::push_metadata_tag(&mut tags, "auth_mode", self.metadata.auth_mode.as_deref())?;
+        Self::push_metadata_tag(
+            &mut tags,
+            "session.source",
```
xl-openai: I prefer just session_source

xl-openai (approved):

PR #9730, opened 2026-01-23
Hide mode cycle hint while a task is running

charley-oai:
## Summary
- hide the “(shift+tab to cycle)” suffix on the collaboration mode label while a task is running
- keep the cycle hint visible when idle
- add a snapshot to cover the running-task label state


charley-oai:
@codex review this

aibrahim-oai (approved):

PR #9736, opened 2026-01-23
ansi-escape: normalize ANSI black spans for dark themes

swordfish444:
Fixes #9694.

The diff view renders ANSI output via `ansi_to_tui`. ANSI black (`30`) maps to `Color::Black`, which is effectively invisible on dark themes; similarly, `Color::Reset` should behave like “no explicit fg” rather than forcing a reset color.

This post-processes `ansi_to_tui` output so `Color::Reset` becomes `fg=None`, and `Color::Black` (when no background is set) becomes `fg=None` (or `DarkGray` when bold).

Tests: `cargo test -p codex-ansi-escape`, `cargo test -p codex-tui`.

cc @Weijiang-Xiong @etraut-openai

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -17,6 +17,9 @@ concurrency:
   cancel-in-progress: ${{ github.ref_name != 'main' }}
 jobs:
   test:
+    # This workflow relies on `secrets.BUILDBUDDY_API_KEY`, which is not available for PRs
+    # from forks. Skip the Bazel jobs in that case rather than failing with UNAUTHENTICATED.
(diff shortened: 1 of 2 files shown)


etraut-openai:
@swordfish444, I don't think this is a fix for the linked bug. That bug was reported for the IDE Extension, not the CLI.

Is there a different bug that you're trying to fix here?

etraut-openai:
As I mentioned above, this doesn't fix the bug reported in the associated bug report, so I'm going to close.

swordfish444:
You're right — my mistake. I misread #9694 as a CLI issue, but it sounds like this is actually the VS Code extension and already being addressed there.

No worries closing this; thanks for the quick clarification.

PR #9740, opened 2026-01-23
Mark subagent threads as SessionSource::SubAgent to prevent app sidebar spam

charley-oai:
## Summary
- Tag subagent-spawned threads with `SessionSource::SubAgent` instead of inheriting the parent session source.
- Add a unit test to assert the subagent session source is persisted in rollout metadata.

## Why
Subagent threads were being recorded as interactive sessions (CLI/VSCode), which caused them to appear in
`thread/list` and clutter app-server UIs. Marking them as subagent sessions keeps the default thread list clean
without changing the public app-server API.

## Changes
- `codex-rs/core/src/thread_manager.rs`
  - Add helpers to spawn threads with an explicit `SessionSource`.
- `codex-rs/core/src/agent/control.rs`
  - Spawn subagents with `SessionSource::SubAgent(SubAgentSource::Other("spawn_agent"))`.
  - New test: `spawn_agent_sets_subagent_session_source`.

### DIFF
--- a/codex-rs/core/src/agent/control.rs
+++ b/codex-rs/core/src/agent/control.rs
@@ -5,6 +5,8 @@ use crate::error::Result as CodexResult;
 use crate::thread_manager::ThreadManagerState;
 use codex_protocol::ThreadId;
 use codex_protocol::protocol::Op;
+use codex_protocol::protocol::SessionSource;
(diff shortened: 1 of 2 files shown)


charley-oai:
@codex review this

Issue #9668, opened 2026-01-22
Error: invalid type: boolean `true`, expected struct SkillsConfig in `skills`


kevin-valerio:
### What version of Codex is running?

0.88

### What subscription do you have?

API key

### Which model were you using?

gpt-5.2

### What platform is your computer?

Darwin 25.2.0 arm64 arm

### What terminal emulator and version are you using (if applicable)?

iTerm2

### What issue are you seeing?

After updating Codex to 0.88, I get the following error message that blocks Codex:
```
Error: invalid type: boolean `true`, expected struct SkillsConfig in `skills`
```

### What steps can reproduce the bug?

Update with `npm install -g @openai/codex` and start codex.

### What is the expected behavior?

Codex starting.

etraut-openai:
The error message indicates that your `config.toml` config file has an invalid `skills` entry. Refer to the [config reference for details](https://developers.openai.com/codex/config-reference).

kevin-valerio:
Thank you!

etraut-openai:
That error message is not as clear as it could be. I'll work on improving it.

PR #9746, opened 2026-01-23
Another round of improvements for config error messages

etraut-openai:
In a [recent PR](https://github.com/openai/codex/pull/9182), I made some improvements to config error messages so errors didn't leave app server clients in a dead state. This is a follow-on PR to make these error messages more readable and actionable for both TUI and GUI users. For example, see #9668 where the user was understandably confused about the source of the problem and how to fix it.

The improved error message:
1. Clearly identifies the config file where the error was found (which is more important now that we support layered configs)
2. Provides a line and column number of the error
3. Displays the line where the error occurred and underlines it

For example, if my `config.toml` includes the following:
```toml
[features]
collaboration_modes = "true"
```

Here's the current CLI error message:
```
Error loading config.toml: invalid type: string "true", expected a boolean in `features`
```

And here's the improved message:
```
Error loading config.toml:
/Users/etraut/.codex/config.toml:43:23: invalid type: string "true", expected a boolean
   |
43 | collaboration_modes = "true"
   |                       ^^^^^^
```

The bulk of the new logic is contained within a new module `config_loader/diagnostics.rs` that is responsible for calculating the text range for a given toml path (which is more involved than I would have expected).

In addition, this PR adds the file name and text range to the `ConfigWarningNotification` app server struct. This allows GUI clients to present the user with a better error message and an optional link to open the errant config file. This was a suggestion from @.bolinfest when he reviewed my previous PR.


etraut-openai:
@codex review

codex-rs/app-server/src/lib.rs
```diff
@@ -44,6 +49,67 @@ mod outgoing_message;
...
+    let (path, range) = config_error_location(err);
+    ConfigWarningNotification {
+        summary: summary.into(),
+        details: Some(err.to_string()),
+        path,
+        range,
+    }
+}
+
+fn config_error_location(err: &std::io::Error) -> (Option<String>, Option<AppTextRange>) {
```
bolinfest: It looks like this can return `(Some, Some)` or `(None, None)`, so can we move the `Option` up one level so there are only two states to reason about instead of four? You can do `Option<(String, AppTextRange)>` if you don't want to create a small struct?

codex-rs/app-server/src/lib.rs
```diff
@@ -44,6 +49,67 @@ mod outgoing_message;
...
+                        column: location.range.start.column,
+                    },
+                    end: AppTextPosition {
+                        line: location.range.end.line,
+                        column: location.range.end.column,
+                    },
+                };
+                return (Some(location.path), Some(range));
+            }
+            (Some(path.clone()), None)
```
bolinfest, line 98: Oh, is this so the signatures match in the two functions? So `(None, Some)` isn't really valid, though?

I guess it's private to this file, so I don't feel _that_ strongly.

bolinfest (approved):
Very nice and thoughtful change!

PR #9762, opened 2026-01-23
VS Code extension issue template: capture UI locale for localization bugs

MaxMiksa:
Fixes #9761 (process improvement)

Adds an explicit `locale` field to the VS Code Extension issue template and updates the description to request the i18n message id/key + rendered string for localization-related reports.

This makes localization bugs (like zh-CN Plan summary wording/order issues) much easier to triage and reproduce.

### DIFF
--- a/.github/ISSUE_TEMPLATE/5-vs-code-extension.yml
+++ b/.github/ISSUE_TEMPLATE/5-vs-code-extension.yml
@@ -30,6 +30,11 @@ body:
     validations:
       required: true
   - type: input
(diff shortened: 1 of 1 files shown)


etraut-openai:
Thanks, but I think we'll pass on this. We don't get many localization bug reports, so I don't want to ask every person who files a bug report to fill in this field.

MaxMiksa:
The issue has been resolved in the latest version of the VSCode Codex Extension (0.4.67).

PR #9765, opened 2026-01-23
feat: ephemeral threads

jif-oai:
Add ephemeral threads capabilities. Only exposed through the `app-server` v2

The idea is to disable the rollout recorder for those threads.


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1081,6 +1081,9 @@ pub struct ThreadStartParams {
     pub base_instructions: Option<String>,
     pub developer_instructions: Option<String>,
     pub personality: Option<Personality>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
```
owenlin0: nit: let's remove these annotations (will look cleaner and more consistent with the other params when we export these types)

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1464,7 +1467,9 @@ pub struct Thread {
     #[ts(type = "number")]
     pub updated_at: i64,
     /// [UNSTABLE] Path to the thread on disk.
-    pub path: PathBuf,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
```
owenlin0: same here

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1081,6 +1081,9 @@ pub struct ThreadStartParams {
     pub base_instructions: Option<String>,
     pub developer_instructions: Option<String>,
     pub personality: Option<Personality>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub persist_rollout: Option<bool>,
```
apanasenko-oai: nit: `ephemeral: Option<bool>`?

apanasenko-oai (approved):

PR #9769, opened 2026-01-23
feat: add thread spawn source for collab tools

jif-oai:


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -686,6 +688,45 @@ pub enum CommandAction {
     },
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(rename_all = "camelCase", export_to = "v2/")]
+pub enum SubAgentSource {
+    Review,
+    Compact,
+    CollabSpawn { parent_thread_id: String },
```
owenlin0: maybe just `Collab` for simplicity?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -686,6 +688,45 @@ pub enum CommandAction {
     },
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(rename_all = "camelCase", export_to = "v2/")]
+pub enum SubAgentSource {
+    Review,
+    Compact,
```
owenlin0: from an API perspective, shouldn't the compaction operation always be behind the scenes? Does a `compact` thread ever exist in practice? Would be nice to be avoid exposing this if possible

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -686,6 +688,45 @@ pub enum CommandAction {
...
+    CollabSpawn { parent_thread_id: String },
+    Other(String),
```
owenlin0: just curious, what kind of values are possible in `Other`?

owenlin0 (approved):

codex-rs/app-server-protocol/src/protocol/v2.rs
jif-oai: This was actually already here, we should deprecate it soon as compaction is mostly done in remote right now

codex-rs/app-server-protocol/src/protocol/v2.rs
jif-oai: None for now, it was already there

PR #9776, opened 2026-01-23
fix: attempt to reduce high cpu usage when using collab

eugeneoden:
Reproduce with a prompt like this with collab enabled:
```
Examine the code at <some subdirectory with a deeply nested project>.  Find the most urgent issue to resolve and describe it to me.
```

Existing behavior causes the top-level agent to busy wait on subagents.

### DIFF
--- a/codex-rs/core/src/tools/handlers/collab.rs
+++ b/codex-rs/core/src/tools/handlers/collab.rs
@@ -26,6 +26,8 @@ use serde::Serialize;
 
 pub struct CollabHandler;
 
+/// Minimum wait timeout to prevent tight polling loops from burning CPU.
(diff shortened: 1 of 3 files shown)


codex-rs/core/src/tools/handlers/collab.rs
```diff
@@ -26,6 +26,8 @@ use serde::Serialize;
 
 pub struct CollabHandler;
 
+/// Minimum wait timeout to prevent tight polling loops from burning CPU.
+pub(crate) const MIN_WAIT_TIMEOUT_MS: i64 = 1_000;
```
jif-oai: Let's go to 10s actullay as we have early exit

jif-oai (approved):

PR #9777, opened 2026-01-23
plan mode prompt change

aibrahim-oai:
Pick the first available preset after auth filtering for default selection.

### DIFF
--- a/codex-rs/core/templates/collaboration_mode/plan.md
+++ b/codex-rs/core/templates/collaboration_mode/plan.md
@@ -2,7 +2,7 @@
 
 You work in **two phases**:
 
-- **PHASE 1 — Understand user intent**: Align on what the user is trying to accomplish and what “success” means. Focus on intent, scope, constraints, and context.
+- **PHASE 1 — Understand user intent**: Align on what the user is trying to accomplish and what “success” means. Focus on intent, scope, constraints, and preference tradeoffs.
 - **PHASE 2 — Technical spec & implementation plan**: Convert the intent into a decision‑complete technical spec and an implementation plan detailed enough that another agent could execute with minimal follow‑ups.
 
 ---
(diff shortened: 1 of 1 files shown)


codex-rs/core/templates/collaboration_mode/plan.md
```diff
@@ -125,42 +166,28 @@ Do **not** use `request_user_input` to ask:
...
-          "description": "Lower write latency but higher complexity."
-        },
-        {
-          "label": "Other",
-          "description": "Provide a custom strategy or constraints."
-        }
+        { "label": "Ship fast (Recommended)", "description": "Minimal changes, pragmatic shortcuts, faster delivery." },
+        { "label": "Robust & maintainable", "description": "Cleaner abstractions, more refactor, better long-term stability." },
+        { "label": "Performance-first", "description": "Optimize latency/throughput even if complexity rises." },
+        { "label": "Other", "description": "Specify a different priority or constraint." }
```
shijie-oai, line 189: Separate from this PR - our parsing is purely string match and what if the label is localized by the model? Then we lose other option technically. We should modify the tool schema to include a isOtherAvailable field probaby.

codex-rs/core/templates/collaboration_mode/plan.md
```diff
@@ -93,14 +131,17 @@ Before finalizing the plan, ensure you’ve pinned down (answer or assumption):
...
 If something is high-impact and unknown, ask via `request_user_input`. Otherwise assume defaults and proceed.
 
 ---
 
 ## Using `request_user_input` in Plan Mode
 
-Use `request_user_input` only when you are genuinely blocked on a decision that materially changes the plan AND the decision cannot be resolved via evidence-first workspace exploration.
+Use `request_user_input` when either:
+1) You are genuinely blocked on a decision that materially changes the plan and cannot be resolved via evidence-first exploration, OR  
+2) There is a meaningful **preference/tradeoff** the user should choose among.
```
shijie-oai, line 142: Not sure if it is included - can we also say that if no answer is provided, we should use the recommended answer as default.

shijie-oai (approved):

codex-rs/core/templates/collaboration_mode/plan.md
aibrahim-oai: okay

PR #9779, opened 2026-01-23
feat: override model and reasoning effort for subagents

eugeneoden:
Useful for starting the top-level with high/xhigh effort then
subagents with different models and/or reasoning levels as appropriate
to the task.


jif-oai:
Why not using agent role for this? In general having lower cardinality in tool calls enums is better for the model

jif-oai (commented):
Thanks, if ok for you let's wait to finish the protocol with the o-serie team first

PR #9785, opened 2026-01-23
Fix: Respect global + per-mode model overrides for collaboration modes

sonobo:
This PR fixes https://github.com/openai/codex/issues/9783

When collaboration_modes is enabled, collaboration mode presets could override/ignore user-selected model and model_reasoning_effort (from config.toml and/or CLI overrides), leading to unexpected defaults when starting Codex or switching modes.

**Changes**

  - Add optional config.toml overrides for collaboration mode presets:
      - [collaboration_modes.{plan,pair_programming,execute}]
      - supported keys: model, model_reasoning_effort
  - Apply precedence when building collaboration mode presets:
      - built-in defaults < global config/CLI < per-mode overrides
  - Update JSON schema + docs
  - Add regression test covering precedence

**Manual test**

  - Start with --enable collaboration_modes and verify the active mode uses the configured model/effort and stays consistent when cycling modes.

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -455,10 +455,17 @@ impl CodexMessageProcessor {
             ClientRequest::CollaborationModeList { request_id, params } => {
                 let outgoing = self.outgoing.clone();
                 let thread_manager = self.thread_manager.clone();
+                let config = self.config.clone();
 
                 tokio::spawn(async move {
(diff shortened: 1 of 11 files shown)


sonobo:

etraut-openai:
Thanks, but I think we'll pass on this. 

This feature is still under active development, and we're making significant changes to it still.

vietnt:
@etraut-openai so is there a work around to this problem? or simply not using colloboration_modes for now?

PR #9797, opened 2026-01-23
remove sandbox globals.

iceweasel-oai:
Threads sandbox updates through OverrideTurnContext for active turn
Passes computed sandbox type into safety/exec


iceweasel-oai:
@codex review

codex-rs/core/src/windows_sandbox.rs
```diff
@@ -8,6 +12,36 @@ use std::path::Path;
...
+
+impl WindowsSandboxModeExt for WindowsSandboxMode {
+    fn from_config(config: &Config) -> WindowsSandboxMode {
+        Self::from_features(&config.features)
+    }
+
+    fn from_features(features: &Features) -> WindowsSandboxMode {
+        if !features.enabled(Feature::WindowsSandbox) {
+            return WindowsSandboxMode::Disabled;
+        }
```
dylan-hurd-oai, line 28: should we require both here? I would imagine many users will only end up doing one

codex-rs/protocol/src/config_types.rs
```diff
@@ -66,6 +66,18 @@ pub enum SandboxMode {
     DangerFullAccess,
 }
 
+#[derive(
+    Deserialize, Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Display, JsonSchema, TS,
+)]
+#[serde(rename_all = "kebab-case")]
+#[strum(serialize_all = "kebab-case")]
+pub enum WindowsSandboxMode {
```
dylan-hurd-oai: I'm pretty concerned about confusion here, but don't have a better name of the top of my head

codex-rs/tui/src/app.rs
```diff
@@ -1894,6 +1921,24 @@ impl App {
...
+                                sandbox_policy: None,
+                                windows_sandbox_mode: Some(windows_sandbox_mode),
+                                model: None,
+                                effort: None,
+                                summary: None,
+                                collaboration_mode: None,
+                                personality: None,
+                            }));
+                    }
+                }
```
dylan-hurd-oai, line 1942: Do we need to provide an OverrideTurnContext or can we just include it in the next user turn?

codex-rs/tui/src/app.rs
```diff
@@ -1668,9 +1672,24 @@ impl App {
...
+                                        approval_policy: None,
+                                        sandbox_policy: None,
+                                        windows_sandbox_mode: Some(windows_sandbox_mode),
+                                        model: None,
+                                        effort: None,
+                                        summary: None,
+                                        collaboration_mode: None,
+                                        personality: None,
+                                    },
+                                ));
```
dylan-hurd-oai, line 1692: should we be doing this or `OverrideTurnContext` or just sending with next UserTurn?

dylan-hurd-oai (commented):
Couple of small comments, otherwise excited to land!

codex-rs/core/src/windows_sandbox.rs
iceweasel-oai: yeah, right now the "main" feature needs to be on for either sandbox. I'm going to clean this up when these become non-experimental and then it will be a real enum.

For now, this functions correctly because having only the elevated feature on never happens

codex-rs/protocol/src/config_types.rs
iceweasel-oai: yeah fair point, lemme think on this for a future PR (wanna get this one landed soon because it's been conflicting a lot with main).

`WindowsSandboxLevel` might be better

codex-rs/tui/src/app.rs
iceweasel-oai: in practice these are the same thing - we use OverrideTurnContext to change settings/etc that are used by future turns. This approach keeps `UserTurn` simpler and representing the actual Turn from the user, and not settings/config updates

codex-rs/tui/src/app.rs
iceweasel-oai: ditto

codex-rs/tui/src/app.rs
dylan-hurd-oai: I don't disagree with that argument but I believe we are trying to standardizing on UserTurn cc @aibrahim-oai

dylan-hurd-oai: Not blocking though!

codex-rs/protocol/src/config_types.rs
dylan-hurd-oai: Can we go with `WindowsSandboxLevel` - hear you on the merge conflicts and am happy to stamp after this change, but in the config interface we probably want to avoid as many breaking changes

codex-rs/tui/src/app.rs
iceweasel-oai: as in, we are getting rid of `OverrideTurnContext` and all the fields in it will be moved to `UserTurn`?
If so, I'd say when that happens, whoever does it can take care of this field too

codex-rs/protocol/src/config_types.rs
iceweasel-oai: yup, will change this one now

dylan-hurd-oai (approved):

PR #9805, opened 2026-01-24
feat: stashes current message with Ctrl+S

fcoury:
# Summary

Adds message stash to the composer with Ctrl+S.

This feature allows users to temporarily stash their current message, attachments and pending paste expansions and have it automatically restored after the next submission.

# Motivation

There were a few occasions when I was preparing a large follow-up task for the agent while I noticed the agent was doing something undesired. Under normal circumstances I need to copy the text I had written, paste it somewhere else (not really because I use a clipboard manager, but still), enter the steering message, submit it and then paste the message I had in progress.

This PR would allow me hit Ctrl+S as soon as I realize the model needs steering and as soon as I send the steering message, the stashed message and all its attributes gets automatically restored.

# Edge cases

We don't want the user to ever lose a message in progress, so if the composer isn't empty when we attempt to restore the stash, we stop, don't restore it and keep the stash untouched.


etraut-openai:
@codex review

fcoury:

PR #9806, opened 2026-01-24
Use collaboration mode masks without mutating base settings

aibrahim-oai:
Keep an unmasked base collaboration mode and apply the active mask on demand. Simplify the TUI mask helpers and update tests/docs to match the mask contract.


aibrahim-oai:
@codex review

codex-rs/core/src/models_manager/collaboration_mode_presets.rs
```diff
@@ -20,50 +19,46 @@ pub(super) fn builtin_collaboration_mode_presets() -> Vec<CollaborationMode> {
...
-        settings: Settings {
-            model: "gpt-5.2-codex".to_string(),
-            reasoning_effort: Some(ReasoningEffort::Medium),
-            developer_instructions: Some(COLLABORATION_MODE_CODE.to_string()),
-        },
+fn code_preset() -> CollaborationModeMask {
+    CollaborationModeMask {
+        name: "Code".to_string(),
+        mode: Some(ModeKind::Code),
+        model: Some("gpt-5.2-codex".to_string()),
```
charley-oai, line 40: I thought we didn't want to override model in Code mode?

codex-rs/core/src/models_manager/collaboration_mode_presets.rs
```diff
@@ -20,50 +19,46 @@ pub(super) fn builtin_collaboration_mode_presets() -> Vec<CollaborationMode> {
...
-        settings: Settings {
-            model: "gpt-5.2-codex".to_string(),
-            reasoning_effort: Some(ReasoningEffort::Medium),
-            developer_instructions: Some(COLLABORATION_MODE_PAIR_PROGRAMMING.to_string()),
-        },
+fn pair_programming_preset() -> CollaborationModeMask {
+    CollaborationModeMask {
+        name: "Pair Programming".to_string(),
+        mode: Some(ModeKind::PairProgramming),
+        model: Some("gpt-5.2-codex".to_string()),
```
charley-oai, line 50: Same for Pair mode (I still don't like the name Pair fwiw)

codex-rs/tui/src/chatwidget.rs
```diff
@@ -2816,18 +2795,24 @@ impl ChatWidget {
             }
         }
 
+        let effective_mode = self.effective_collaboration_mode();
+        let collaboration_mode = if self.collaboration_modes_enabled() {
+            self.active_collaboration_mask
+                .as_ref()
+                .map(|_| effective_mode.clone())
+        } else {
+            None
```
charley-oai, line 2831: If active_collaboration_mask is None, that expression returns None, so the mode is not sent even though the feature
is enabled.

Why active_collaboration_mask is often None by default:

- On init, they set:
    - base mode = Custom
    - active_collaboration_mask: None
    - See codex-rs/tui/src/chatwidget.rs:1937 and codex-rs/tui/src/chatwidget.rs:1979 (near the struct
      initialization).

So: the mode is now only attached to user turns after the user selects a collaboration preset (i.e., a mask becomes
active).

I know this is intentional, but why not just simplify the code by starting the user off in Code mode? Then they also get the shift+tab suggestion so Plan mode is more discoverable?

codex-rs/protocol/src/config_types.rs
```diff
@@ -219,3 +239,14 @@ pub struct Settings {
...
+
+/// A mask for collaboration mode settings, allowing partial updates.
+/// All fields except `name` are optional, enabling selective updates.
+#[derive(Clone, PartialEq, Eq, Hash, Debug, Serialize, Deserialize, JsonSchema, TS)]
+pub struct CollaborationModeMask {
+    pub name: String,
+    pub mode: Option<ModeKind>,
+    pub model: Option<String>,
+    pub reasoning_effort: Option<ReasoningEffort>,
+    pub developer_instructions: Option<String>,
```
charley-oai: Non-blocking thought: `CollaborationModeMask` can’t express “clear developer instructions / effort” (because `None` currently means “no change,” not “set to null”). If we anticipate runtime mask edits (not just presets), consider a tri-state shape like `Option<Option<T>>` or an enum (`NoChange | Update(Option<T>)`).

codex-rs/core/src/models_manager/collaboration_mode_presets.rs
```diff
@@ -20,50 +19,46 @@ pub(super) fn builtin_collaboration_mode_presets() -> Vec<CollaborationMode> {
...
 
-fn execute_preset() -> CollaborationMode {
-    CollaborationMode {
-        mode: ModeKind::Execute,
-        settings: Settings {
-            model: "gpt-5.2-codex".to_string(),
-            reasoning_effort: Some(ReasoningEffort::High),
-            developer_instructions: Some(COLLABORATION_MODE_EXECUTE.to_string()),
-        },
+fn execute_preset() -> CollaborationModeMask {
```
charley-oai, line 56: Is it weird to have both Code and Execute referenced in the codebase given that they are just different names for the same thing? Can we just rename Execute to Code?

codex-rs/tui/src/chatwidget.rs
```diff
@@ -415,12 +416,12 @@ pub(crate) struct ChatWidget {
     /// where the overlay may briefly treat new tail content as already cached.
     active_cell_revision: u64,
     config: Config,
-    /// Stored collaboration mode with model and reasoning effort.
+    /// The unmasked collaboration mode settings (always Custom mode).
```
charley-oai, line 419: Does this mean Custom is the default? Wouldn't we want Code to be the default?

charley-oai:
@codex review this

charley-oai:
Also CI is failing

codex-rs/core/src/models_manager/collaboration_mode_presets.rs
aibrahim-oai: Next PR

codex-rs/core/src/models_manager/collaboration_mode_presets.rs
aibrahim-oai: same as above

aibrahim-oai:
@codex review this

aibrahim-oai:
@codex review this

charley-oai (approved):

PR #6238, opened 2025-11-05
fix: update brew auto update version check

shijie-oai:
### Summary
* Use `https://github.com/Homebrew/homebrew-cask/blob/main/Casks/c/codex.rb` to get the latest available version for brew usage.

PR #9823, opened 2026-01-24
fix: use `brew upgrade --cask codex` to avoid warnings and ambiguity

JBallin:
Fixes #9822 

### Summary

Make the Homebrew upgrade command explicit by using `brew upgrade --cask codex`.

### Motivation

During the Codex self-update, Homebrew can emit an avoidable warning because the
name `codex` resolves to a cask:

```
Warning: Formula codex was renamed to homebrew/cask/codex.
````

While the upgrade succeeds, this relies on implicit name resolution and produces
unnecessary output during the update flow.

### Why `--cask`

* Eliminates warning/noise for users
* Explicitly matches how Codex is distributed via Homebrew
* Avoids reliance on name resolution behavior
* Makes the command more robust if a `codex` formula is ever introduced

### Context

This restores the `--cask` flag that was removed in #6238 after being considered
“not necessary” during review: [https://github.com/openai/codex/pull/6238#discussion_r2505947880](https://github.com/openai/codex/pull/6238#discussion_r2505947880).


JBallin:

shijie-oai (approved):

PR #9827, opened 2026-01-24
tui: render MCP image output regardless of order

AmirTlinov:
## Summary
- render MCP image output even when text precedes the image block
- add a unit test covering text-then-image tool output
- keep existing image decoding path unchanged

## Testing
- just fmt
- cargo test -p codex-tui

Fixes #9814


etraut-openai:
There is already a PR submitted for this issue: #9815

PR #9834, opened 2026-01-25
Mark collab as beta

pakrym-oai:


etraut-openai (changes requested):
Beta implies that it's ready for broad external testing. It should be fully documented at this stage. I think it should be marked as "Under development" still.

etraut-openai (approved):
It would be good to align the terminology with the docs, but that can be a separate PR.

PR #9838, opened 2026-01-25
Fix typo in prompt files: "command execution" -> "see command execution"

mschruf:
The current phrasing "The user does not command execution outputs" appears to be a typo (missing verb). Based on the context of the following instructions, the missing verb should be "see".


mschruf:

etraut-openai:
Thanks, but we'll pass on this. System prompts are tuned carefully, and even small changes can result in unintended changes. That means they require significant regression testing.

PR #9843, opened 2026-01-25
Add thread/unarchive to restore archived rollouts

charley-oai:
## Summary
- Adds a new `thread/unarchive` RPC to move archived thread rollouts back into the active `sessions/` tree.

## What changed
- **Protocol**
  - Adds `thread/unarchive` request/response types and wiring.
- **Server**
  - Implements `thread_unarchive` in the app server.
  - Validates the archived rollout path and thread ID.
  - Restores the rollout to `sessions/YYYY/MM/DD/...` based on the rollout filename timestamp.
- **Core**
  - Adds `find_archived_thread_path_by_id_str` helper for archived rollouts.
- **Docs**
  - Documents the new RPC and usage example.
- **Tests**
  - Adds an end-to-end server test that:
    1) starts a thread,
    2) archives it,
    3) unarchives it,
    4) asserts the file is restored to `sessions/`.

## How to use
```json
{ "method": "thread/unarchive", "id": 24, "params": { "threadId": "<thread-id>" } }
```

## Author Codex Session

`codex resume 019bf158-54b6-7960-a696-9d85df7e1bc1` (soon I'll make this kind of session UUID forkable by anyone with the right `session_object_storage_url` line in their config, but for now just pasting it here for my reference)

### DIFF
--- a/codex-rs/app-server-protocol/src/protocol/v2.rs
+++ b/codex-rs/app-server-protocol/src/protocol/v2.rs
@@ -1220,7 +1220,9 @@ pub struct ThreadUnarchiveParams {
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
 #[serde(rename_all = "camelCase")]
 #[ts(export_to = "v2/")]
-pub struct ThreadUnarchiveResponse {}
+pub struct ThreadUnarchiveResponse {
+    pub thread: Thread,
+}
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
 #[serde(rename_all = "camelCase")]
--- a/codex-rs/app-server-protocol/src/protocol/v2.rs
+++ b/codex-rs/app-server-protocol/src/protocol/v2.rs
@@ -1213,6 +1213,18 @@ pub struct ThreadArchiveResponse {}
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
 #[serde(rename_all = "camelCase")]
 #[ts(export_to = "v2/")]
+pub struct ThreadUnarchiveParams {
+    pub thread_id: String,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ThreadUnarchiveResponse {}
+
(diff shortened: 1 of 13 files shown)


charley-oai:
@codex review this

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3358,6 +3419,91 @@ impl CodexMessageProcessor {
         })
     }
 
+    async fn unarchive_thread_common(
```
owenlin0: nit: instead of a helper you can just in-line this code in `unarchive_thread`. `*_common` was a pattern I had to use to share code between v1 and v2 in the initial migration, but now that we're solely contributing to v2 we don't really need it.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -4445,6 +4591,15 @@ pub(crate) async fn read_summary_from_rollout(
     })
 }
 
+fn rollout_date_parts(file_name: &OsStr) -> Option<(String, String, String)> {
```
owenlin0: is there a helper function already in `codex-rs/core/src/rollout/*` somewhere that we can reuse? a function like this should belong there

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1210,6 +1210,18 @@ pub struct ThreadArchiveParams {
...
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ThreadUnarchiveParams {
+    pub thread_id: String,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ThreadUnarchiveResponse {}
```
owenlin0: would be helpful to return the thread object IMO (see how `thread/start` does it, although you can omit all other params like `model_provider`, `cwd`, etc.)

`pub thread: Thread`

owenlin0 (approved):
left some small comments, pre-approved

codex-rs/app-server/src/codex_message_processor.rs
charley-oai: I don't think one already existed so I added this to `codex-rs/core/src/rollout/list.rs`. Does that seem reasonable?

charley-oai:
@codex review this

PR #9845, opened 2026-01-25
Expose collaboration_mode_kind in turn.started / turn notifications

charley-oai:
## Summary
This threads `collaboration_mode_kind` from `TurnContext` into client-visible turn events so TUI/VSCE can render plan/exec modes without relying on plan items.

## What changed
- Added `collaboration_mode_kind: ModeKind` to `TurnStartedEvent`.
- Populated it from `TurnContext` at core turn entry points.
- Carried it through app-server turn surfaces:
  - `Turn.collaboration_mode_kind: Option<ModeKind>`
  - Stored on turn start and emitted on turn completion/interruption.
- Updated affected protocol + unit tests.

## Why
Clients already observe turn lifecycle events. Surfacing the collaboration mode kind there is the simplest, most stable way to let clients special-case rendering for plan mode.


charley-oai:
@codex review this

PR #9851, opened 2026-01-25
fix(tui): restore fuzzy matching fallback for slash-command popup

### LINKED ISSUE #9741: Slash-command popup lost fuzzy/substring matching in 0.89.0 (now prefix-only) — please add fuzzy fallback or config toggle

codex-cli 0.89.0 (tag: rust-v0.89.0)

N/A (this is a local TUI behavior; occurs before any model call)

N/A (not model-dependent)

macOS 15.6.1 (Darwin 24.6.0, arm64)

Ghostty 1.2.3 (zsh)

After upgrading to 0.89.0, the slash-command autocomplete/popup appears to no longer support fuzzy/substring (subsequence) matching.
  It now behaves like exact/prefix-only filtering, which makes it hard to discover/trigger commands unless you remember the exact prefix.

  Example:
  - Typing `/ac` used to show `/compact` (because "ac" is a subsequence / substring match), but now it shows no results.

  This looks related to `openai/codex#9629` (“better sorting of shell commands”), but the current behavior feels more strict than “exact/prefix first, then
  fuzzy”: fuzzy seems removed entirely.
 

1. Install/upgrade to codex-cli 0.89.0.
2. Start an interactive Codex session.
3. Type `/ac` in the composer so the slash-command popup is active.
4. Observe the suggestions list.

 Either:
  - Keep fuzzy matches as a fallback (e.g. exact > prefix > fuzzy), OR
  - If fuzzy matches are considered too noisy, only show fuzzy results when there are no exact/prefix matches, OR
  - Provide a config option to choose behavior, e.g.:
    - `slash_command_matcher = "prefix"` (current)
    - `slash_command_matcher = "exact_prefix_fuzzy"` (previous/desired)

  So that `/ac` can still surface `/compact` without breaking the “prefix should win” goal.

### What do you see instead?
  No suggestions for inputs that rely on fuzzy/substring matching (e.g. `/ac` does not include `/compact`).

### Additional information

 - Related PR: `openai/codex#9629`
  - A prior approach that seemed closer to the desired UX: openai/codex#9625 (prefix first, then fuzzy), but it was closed in favor of #9629.
  - If the change is intentional, it would be helpful to document the new matching semantics in the slash-commands docs / release notes more explicitly
  (since many users treat fuzzy as “autocomplete”).

zerone0x:
## Summary

Fixes #9741

Re-enable fuzzy/substring matching as a fallback in the slash-command autocomplete popup. 

PR #9629 was intended to implement "exact > prefix > fuzzy" ordering but actually removed fuzzy matching entirely. This PR restores the original behavior where fuzzy matches appear after exact and prefix matches.

## Changes

- Re-imported `fuzzy_match` from `codex_common::fuzzy_match`
- Modified the `filtered()` function to include fuzzy matches as a third tier
- Fuzzy matches are sorted by score (lower is better)
- Updated tests to verify fuzzy fallback works

## Test Plan

- [x] `cargo test -p codex-tui --lib -- command_popup` passes
- [x] Typing `/ac` now shows `/compact` as a fuzzy match
- [x] Prefix matches still appear before fuzzy matches (e.g., `/mo` shows `/model` first)
- [x] Clippy passes with no warnings

---
🤖 Generated with Claude Code (issue-hunter-pro)

### DIFF
--- a/codex-rs/tui/src/bottom_pane/command_popup.rs
+++ b/codex-rs/tui/src/bottom_pane/command_popup.rs
@@ -10,6 +10,7 @@ use crate::render::Insets;
 use crate::render::RectExt;
 use crate::slash_command::SlashCommand;
 use crate::slash_command::built_in_slash_commands;
+use codex_common::fuzzy_match::fuzzy_match;
 use codex_protocol::custom_prompts::CustomPrompt;
 use codex_protocol::custom_prompts::PROMPTS_CMD_PREFIX;
 use std::collections::HashSet;
@@ -116,9 +117,11 @@ impl CommandPopup {
         measure_rows_height(&rows, &self.state, MAX_POPUP_ROWS, width)
     }
 
-    /// Compute exact/prefix matches over built-in commands and user prompts,
-    /// paired with optional highlight indices. Preserves the original
-    /// presentation order for built-ins and prompts.
+    /// Compute exact/prefix/fuzzy matches over built-in commands and user prompts,
+    /// paired with optional highlight indices. Results are ordered:
+    /// 1. Exact matches first
+    /// 2. Prefix matches second
+    /// 3. Fuzzy matches third (sorted by score, lower is better)
     fn filtered(&self) -> Vec<(CommandItem, Option<Vec<usize>>)> {
         let filter = self.command_filter.trim();
         let mut out: Vec<(CommandItem, Option<Vec<usize>>)> = Vec::new();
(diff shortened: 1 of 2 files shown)


etraut-openai:
Our [contribution guidelines](https://github.com/openai/codex/blob/main/docs/contributing.md) state that we're not accepting feature contributions at this time. We need to make sure that all new features are well designed, compose well with both existing and upcoming features, fit into our roadmap, and are consistent across all Codex surfaces (CLI, IDE extension, web). If you would like to propose a new feature, please file or upvote an enhancement request in the issue tracker. We will generally prioritize new features based on community feedback.

PR #9859, opened 2026-01-25
feat(network-proxy): add MITM support and gate limited-mode CONNECT

viyatb-oai:
## Description
- Adds MITM support (CA load/issue, TLS termination, optional body inspection).
- Adds `codex-network-proxy init` to create `CODEX_HOME/network_proxy/mitm`.
- Enforces limited-mode HTTPS correctly: `CONNECT` requires MITM, otherwise blocked with `mitm_required`.
- Keeps `origin/main` layering/reload semantics (managed layers included in reload checks).
- Centralizes block reasons (`REASON_MITM_REQUIRED`) and removes `println!`.
- Scope is MITM-only (no SOCKS changes).

gated by `mitm=false` (default)


rreichel3-oai (approved):
Looks great!

PR #9890, opened 2026-01-26
Feat: add isOther to question returned by request user input tool

shijie-oai:
### Summary
Add `isOther` to question object from request_user_input tool input and remove `other` option from the tool prompt to better handle tool input.


aibrahim-oai:
Let's have it on question level and tell the model to not provide an other option

aibrahim-oai (approved):

PR #9829, opened 2026-01-24
core: refine read_file block comment headers

AmirTlinov:
## Summary
- treat inline `/*` as starting block-comment state for following lines without marking the current line as a header
- ignore single-line `/* ... */` comments when trailing code is present
- add unit tests covering both cases

## Testing
- just fmt
- cargo test -p codex-core

Fixes #9828

PR #9895, opened 2026-01-26
Fix read_file inline block comment headers

AmirTlinov:
Fixes https://github.com/openai/codex/issues/9828

Problem
- Inline `/*` does not enter block-comment state, so header lines are missed.
- `/* ... */ code` is misclassified as a header comment.

Solution
- Track block-comment state per line when collecting indentation records.
- Treat inline `/*` as starting comment state for subsequent lines, but not as a header itself.
- Add two unit tests for the above cases.

Testing
- just fmt
- just fix -p codex-core
- cargo test -p codex-core
- cargo test --all-features (flaked once on read_file_tools_run_in_parallel; rerun passed)
- cargo test --all-features -p codex-core --test all suite::tool_parallelism::read_file_tools_run_in_parallel

### DIFF
--- a/codex-rs/core/src/tools/handlers/read_file.rs
+++ b/codex-rs/core/src/tools/handlers/read_file.rs
@@ -18,7 +18,6 @@ pub struct ReadFileHandler;
 const MAX_LINE_LENGTH: usize = 500;
 const TAB_WIDTH: usize = 4;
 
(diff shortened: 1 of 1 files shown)


etraut-openai:
There is already a PR posted for this issue: #9829.

PR #9899, opened 2026-01-26
feat: disable collab at max depth

jif-oai:


codex-rs/core/src/codex.rs
```diff
@@ -262,6 +264,12 @@ impl Codex {
             );
         }
 
+        if let SessionSource::SubAgent(SubAgentSource::ThreadSpawn { depth, .. }) = session_source
+            && depth >= MAX_THREAD_SPAWN_DEPTH
+        {
+            config.features.disable(Feature::Collab);
```
gt-oai, line 270: This smells bad -- is there really no other way to disable this feature? My worry is this eventually persists beyond the current session. Maybe add a comment, or a todo, or do this check in the caller?

gt-oai (approved):

PR #9917, opened 2026-01-26
chore: add try/catch around turn thread.run in sdk/typescript/samples…

prajjwaloneport:
add try/catch around turn thread.run in sdk/typescript/samples…


prajjwaloneport:

etraut-openai:
Thanks, but I think we'll pass on this one. We want to keep the samples simple.

PR #9938, opened 2026-01-26
feat(linux-sandbox): add bwrap support

viyatb-oai:
## Summary
This PR introduces a gated Bubblewrap (bwrap) Linux sandbox path. The curent Linux sandbox path relies on in-process restrictions (including Landlock). Bubblewrap gives us a more uniform filesystem isolation model, especially explicit writable roots with the option to make some directories read-only and granular network controls.

This is behind a feature flag so we can validate behavior safely before making it the default.

- Added temporary rollout flag:
  - `features.use_linux_sandbox_bwrap`
- Preserved existing default path when the flag is off.
- In Bubblewrap mode:
  - Added internal retry without /proc when /proc mount is not permitted by the host/container.


codex-rs/core/src/landlock.rs
```diff
@@ -6,26 +6,34 @@ use std::path::Path;
...
-/// (codex-linux-sandbox).
+/// Spawn a shell tool command under the Linux sandbox helper
+/// (codex-linux-sandbox), which currently uses bubblewrap for filesystem
+/// isolation plus seccomp for network restrictions.
 ///
 /// Unlike macOS Seatbelt where we directly embed the policy text, the Linux
 /// helper accepts a list of `--sandbox-permission`/`-s` flags mirroring the
 /// public CLI. We convert the internal [`SandboxPolicy`] representation into
 /// the equivalent CLI options.
+#[allow(clippy::too_many_arguments)]
```
viyatb-oai, line 17: keeping this since `use_bwrap_sandbox` is temporary

.github/workflows/bazel.yml
```diff
@@ -65,6 +65,14 @@ jobs:
       - name: Set up Bazel
         uses: bazelbuild/setup-bazelisk@v3
 
+      - name: Install Linux bwrap build dependencies
+        if: ${{ runner.os == 'Linux' }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          sudo apt-get update -y
+          sudo DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends pkg-config libcap-dev
```
bolinfest: Again, this should not be necessary because none of the build should be happening locally.

/cc @zbarsky-openai to double-check me there.

codex-rs/linux-sandbox/src/vendored_bwrap.rs
```diff
@@ -13,30 +13,43 @@ mod imp {
...
 
         // SAFETY: We provide a null-terminated argv vector whose pointers
         // remain valid for the duration of the call.
-        let exit_code = unsafe { bwrap_main(cstrings.len() as libc::c_int, argv_ptrs.as_ptr()) };
+        unsafe { bwrap_main(cstrings.len() as libc::c_int, argv_ptrs.as_ptr()) }
+    }
+
+    /// Execute the build-time bubblewrap `main` function with the given argv.
+    pub(crate) fn exec_vendored_bwrap(argv: Vec<String>) -> ! {
+        let exit_code = run_vendored_bwrap_main(&argv);
```
bolinfest, line 44: Hmm, so this appears to be the end of `main()`:

https://github.com/containers/bubblewrap/blob/535238486c5d41be5ee49c6a85842b6f29ba93b5/bubblewrap.c#L3626-L3640

where:

```
if (execvp (exec_path, argv) == -1)
```

leads to:

```
die_with_error ("execvp %s", exec_path);
```

but if `execvp()` does not return `-1`, then bwrap's main returns `0`, even though that suggests something has gone horribly wrong?

.github/workflows/bazel.yml
viyatb-oai: removed! i forgot about this

.github/workflows/rust-ci.yml
```diff
@@ -175,6 +176,13 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+      - name: Install Linux bwrap build dependencies
+        if: ${{ runner.os == 'Linux' }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          sudo apt-get update -y
+          sudo DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends pkg-config libcap-dev
```
zbarsky-openai: we should not be doing this, it should build properly via bazel

zbarsky-openai: nevermind i see you did it :)

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -87,82 +75,174 @@ pub fn run_main() -> ! {
...
-    } else {
-        // Legacy path: Landlock enforcement only.
-        if let Err(e) = apply_sandbox_policy_to_current_thread(&sandbox_policy, &sandbox_policy_cwd)
-        {
-            panic!("error applying legacy Linux sandbox restrictions: {e:?}");
+    }
+    "true".to_string()
+}
+
+fn run_bwrap_in_child_capture_stderr(argv: Vec<String>) -> String {
```
bolinfest, line 193: Please add a docstring to describe the strategy here.

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -87,82 +75,174 @@ pub fn run_main() -> ! {
...
+
+    // Parent: close the write end and read stderr while the child runs.
+    unsafe {
+        libc::close(write_fd);
+    }
+
+    let mut stderr = String::new();
+    // SAFETY: `read_fd` is a valid owned fd in the parent.
+    let mut read_file = unsafe { File::from_raw_fd(read_fd) };
+    if let Err(err) = read_file.read_to_string(&mut stderr) {
```
bolinfest: This does not stream stderr, is that right? So this could grow to arbitrary size in memory?

Is this intended to be used only for the preflight check?

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -87,82 +75,174 @@ pub fn run_main() -> ! {
...
+    let pid = unsafe { libc::fork() };
+    if pid < 0 {
+        let err = std::io::Error::last_os_error();
+        panic!("failed to fork for bubblewrap: {err}");
+    }
+
+    if pid == 0 {
+        // Child: redirect stderr to the pipe, then run bubblewrap.
+        unsafe {
+            libc::close(read_fd);
```
bolinfest: ignore rc?

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -87,82 +75,174 @@ pub fn run_main() -> ! {
...
+            if libc::dup2(write_fd, libc::STDERR_FILENO) < 0 {
+                let err = std::io::Error::last_os_error();
+                panic!("failed to redirect stderr for bubblewrap: {err}");
+            }
+            libc::close(write_fd);
```
bolinfest: ignore rc?

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -87,82 +75,174 @@ pub fn run_main() -> ! {
...
-    };
 
-    exec_or_panic(command);
+        let exit_code = run_vendored_bwrap_main(&argv);
+        std::process::exit(exit_code);
+    }
+
+    // Parent: close the write end and read stderr while the child runs.
+    unsafe {
+        libc::close(write_fd);
```
bolinfest: ignore rc?

bolinfest (commented):
OK, a handful of questions/clarifications!

codex-rs/linux-sandbox/src/linux_run_main.rs
viyatb-oai: fixed

codex-rs/linux-sandbox/src/linux_run_main.rs
viyatb-oai: fixed

codex-rs/linux-sandbox/src/linux_run_main.rs
viyatb-oai: fixed

codex-rs/linux-sandbox/src/linux_run_main.rs
viyatb-oai: only preflight, but I added an upper bound.

codex-rs/linux-sandbox/src/linux_run_main.rs
viyatb-oai: done!

codex-rs/linux-sandbox/tests/suite/landlock.rs
```diff
@@ -47,12 +50,32 @@ async fn run_cmd(cmd: &[&str], writable_roots: &[PathBuf], timeout_ms: u64) {
...
-#[expect(clippy::expect_used, clippy::unwrap_used)]
+#[expect(clippy::expect_used)]
 async fn run_cmd_output(
     cmd: &[&str],
     writable_roots: &[PathBuf],
     timeout_ms: u64,
 ) -> codex_core::exec::ExecToolCallOutput {
+    run_cmd_result(cmd, writable_roots, timeout_ms)
+        .await
+        .expect("sandboxed command should succeed")
```
bolinfest: This is a little misleading. The command should "execute," but it may exit non-zero, so it won't necessarily "succeed" in that sense.

codex-rs/linux-sandbox/tests/suite/landlock.rs
```diff
@@ -47,12 +50,32 @@ async fn run_cmd(cmd: &[&str], writable_roots: &[PathBuf], timeout_ms: u64) {
...
+}
+
+async fn run_cmd_result(
```
bolinfest: Why is this its own function? I only see it called by `run_cmd_output()`, so can we just inline it?

codex-rs/linux-sandbox/tests/suite/landlock.rs
```diff
@@ -47,12 +50,32 @@ async fn run_cmd(cmd: &[&str], writable_roots: &[PathBuf], timeout_ms: u64) {
...
+    cmd: &[&str],
+    writable_roots: &[PathBuf],
+    timeout_ms: u64,
+) -> Result<codex_core::exec::ExecToolCallOutput> {
+    run_cmd_result_with_bwrap(cmd, writable_roots, timeout_ms, false).await
+}
+
+#[expect(clippy::expect_used)]
+async fn run_cmd_result_with_bwrap(
```
bolinfest: This is a weird name because `use_bwrap_sandbox` can be `false` and it won't use `bwrap`, is that right? If so, should this be `run_command_with_writable_roots()`?

codex-rs/linux-sandbox/tests/suite/landlock.rs
```diff
@@ -242,6 +297,90 @@ async fn sandbox_blocks_nc() {
     assert_network_blocked(&["nc", "-z", "127.0.0.1", "80"]).await;
 }
 
+#[tokio::test]
+async fn sandbox_blocks_git_and_codex_writes_inside_writable_root() {
```
bolinfest, line 300: Thanks for these tests!

codex-rs/linux-sandbox/src/linux_run_main.rs
bolinfest: So this means that we run `/bin/true` under `bwrap` every time we want to spawn a command, is that right? Since this is initially behind a flag, I think that's OK, but ultimately I think we should try to find a way to cache this.

bolinfest: In particular, if you are on a host where bwrap doesn't work, ultimately, you should probably have to add something to your `config.toml` that means, "I know the full version of bwrap does not work on this host and I accept the risks that come with that."

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -217,32 +318,43 @@ fn exec_or_panic(command: Vec<String>) -> ! {
...
-    );
+    #[test]
+    fn inserts_bwrap_argv0_before_command_separator() {
+        let argv = build_bwrap_argv(
+            vec!["/bin/true".to_string()],
+            &SandboxPolicy::ReadOnly,
+            Path::new("/"),
+            BwrapOptions { mount_proc: true },
+        );
+        let command_separator_index = argv
```
bolinfest: If `argv` is predictable (which it should be?) can we just `assert_eq!()` on the full vec? See:

https://github.com/openai/codex/blob/49dd67a2606c73eed7316f882c8db027cbf71bbe/AGENTS.md#L80

codex-rs/linux-sandbox/scripts/test_linux_sandbox.sh
```diff
@@ -0,0 +1,222 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Linux sandbox smoke test script.
```
bolinfest: Do you think this should get turned into a diagnostic under `codex debug` that end-users should be able to run? Can be done in a follow-up PR, if so.

bolinfest: Though if this is going to stay long-term and it's only for local debugging, I would strongly prefer Python for this file because this is more Bash than I generally think is good to maintain.

bolinfest (approved):

codex-rs/linux-sandbox/scripts/test_linux_sandbox.sh
viyatb-oai: I'll remove this - this was mainly for local debugging to make sure i'm not breaking functionality!

codex-rs/linux-sandbox/src/linux_run_main.rs
viyatb-oai: it should be. fixed

codex-rs/linux-sandbox/tests/suite/landlock.rs
viyatb-oai: correct, missed this one. changed.

codex-rs/linux-sandbox/tests/suite/landlock.rs
viyatb-oai: ugh yes. inlined now.

codex-rs/linux-sandbox/tests/suite/landlock.rs
viyatb-oai: changed wording.

PR #9941, opened 2026-01-26
error code/msg details for failed elevated setup

iceweasel-oai:


iceweasel-oai:
@codex review

codex-rs/tui/src/app.rs
```diff
@@ -1612,10 +1612,27 @@ impl App {
...
+                                    code_tag = Some(code);
+                                    message_tag = Some(message);
+                                }
+                                let mut tags: Vec<(&str, &str)> = Vec::new();
+                                if let Some(code) = code_tag.as_deref() {
+                                    tags.push(("code", code));
+                                }
+                                if let Some(message) = message_tag.as_deref() {
+                                    tags.push(("message", message));
+                                }
```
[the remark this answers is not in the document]
iceweasel-oai: I'm aware and willing to accept for now. The metric volume is very low, and the fidelity of paths will help me here. Will address this soon-ish

dylan-hurd-oai, line 1631: non-blocking but would be great to include in core

codex-rs/windows-sandbox-rs/src/setup_error.rs
```diff
@@ -0,0 +1,296 @@
...
+    }
+}
+
+pub fn write_setup_error_report(codex_home: &Path, report: &SetupErrorReport) -> Result<()> {
+    let sandbox_dir = codex_home.join(".sandbox");
+    fs::create_dir_all(&sandbox_dir)
+        .with_context(|| format!("create sandbox dir {}", sandbox_dir.display()))?;
+    let path = setup_error_path(codex_home);
+    let json = serde_json::to_vec_pretty(report)?;
+    fs::write(&path, json).with_context(|| format!("write {}", path.display()))?;
```
dylan-hurd-oai, line 165: Do we intend for this to be an append-only log, or only contain the most recent run? I think we want to make this more resilient to race conditions here.

dylan-hurd-oai (approved):
2 follow-ups I think we need to address but in favor of better reporting!

codex-rs/windows-sandbox-rs/src/setup_error.rs
iceweasel-oai: it gets deleted before the setup process runs - right now there really should just be one concurrent setup running, but I'll look into handling concurrency just in case

codex-rs/tui/src/app.rs
iceweasel-oai: yes - one of the things we'll address when we implement the NUX in the vsce is moving all the metrics stuff to core, so it gets included

PR #9947, opened 2026-01-26
fix: try to fix freezes of TUI

jif-oai:
## Potential issue
- When the terminal regains focus (alt-tab), `crossterm` emits `FocusGained`
- TUI handler synchronously calls `requery_default_colors()`, which sends OSC color queries and waits for replies on stdin.
- `crossterm` EventStream also uses a global stdin reader; the query and the event stream can contend for the same input/reply, causing the query to stall and the UI to “freeze”.

## Proposed fix
- On `FocusGained`, we now pause the EventStream, run the palette query, then resume the EventStream. This releases stdin while the OSC query runs, avoiding the contention that can hang the UI.

### DIFF
--- a/codex-rs/core/src/mcp/mod.rs
+++ b/codex-rs/core/src/mcp/mod.rs
@@ -97,6 +97,7 @@ fn codex_apps_mcp_server_config(config: &Config, auth: Option<&CodexAuth>) -> Mc
         tool_timeout_sec: None,
         enabled_tools: None,
         disabled_tools: None,
+        scopes: None,
     }
 }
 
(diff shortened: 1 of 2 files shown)


mehulparmariitr:
The closest follow-up implementation to this earlier freeze fix is now available:

- Canonical issue and direct measurements: https://github.com/openai/codex/issues/24527
- Reviewable two-fix implementation: https://github.com/mehulparmariitr/codex/pull/1

It retains this PR's event-source pause/resume architecture, but replaces the remaining sequential blocking color queries with the existing batched terminal probe under one shared 100 ms deadline. It preserves cached colors on timeout/error and adds a regression proving the event source is recreated and the next key is delivered.

The same branch also removes separately confirmed synchronous terminal-size queries from ordinary draws. Direct upstream submission is invitation-gated; maintainers can invite the branch or port the commits.

PR #9948, opened 2026-01-26
chore(app-server): add test coverage for thread/read not bumping updated_at

owenlin0:
Just adds a quick test ensuring that a call to `thread/read` does not change the thread's updated_at timestamp.

### DIFF
--- a/codex-rs/app-server/tests/suite/v2/thread_read.rs
+++ b/codex-rs/app-server/tests/suite/v2/thread_read.rs
@@ -2,7 +2,9 @@ use anyhow::Result;
 use app_test_support::McpProcess;
 use app_test_support::create_fake_rollout_with_text_elements;
 use app_test_support::create_mock_responses_server_repeating_assistant;
+use app_test_support::rollout_path;
 use app_test_support::to_response;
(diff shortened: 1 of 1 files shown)


codex-rs/app-server/tests/suite/v2/thread_read.rs
```diff
@@ -134,6 +137,57 @@ async fn thread_read_can_include_turns() -> Result<()> {
...
+    .await??;
+    let ThreadReadResponse { thread } = to_response::<ThreadReadResponse>(read_resp)?;
+
+    let expected_updated_at = chrono::DateTime::parse_from_rfc3339(expected_updated_at_rfc3339)?
+        .with_timezone(&Utc)
+        .timestamp();
+    assert_eq!(thread.updated_at, expected_updated_at);
+
+    let after_modified = std::fs::metadata(&rollout_file_path)?.modified()?;
+    assert_eq!(after_modified, before_modified);
```
celia-oai, line 186: I'm a bit confused what this check is doing? the name before and after modified made it sound like they should be different?

celia-oai (approved):

PR #9950, opened 2026-01-27
fix(app-server, core): defer initial context write to rollout file until first turn

owenlin0:
### Overview
Currently calling `thread/resume` will always bump the thread's `updated_at` timestamp. This PR makes it the `updated_at` timestamp changes only if a turn is triggered.

### Additonal context
What we typically do on resuming a thread is **always** writing “initial context” to the rollout file immediately. This initial context includes:
- Developer instructions derived from sandbox/approval policy + cwd
- Optional developer instructions (if provided)
- Optional collaboration-mode instructions
- Optional user instructions (if provided)
- Environment context (cwd, shell, etc.)

This PR defers writing the “initial context” to the rollout file until the first `turn/start`, so we don't inadvertently bump the thread's `updated_at` timestamp until a turn is actually triggered. 

This works even though both `thread/resume` and `turn/start` accept overrides (such as `model`, `cwd`, etc.) because the initial context is seeded from the effective `TurnContext` in memory, computed at `turn/start` time, after both sets of overrides have been applied.

**NOTE**: This is a very short-lived solution until we introduce sqlite. Then we can remove this.

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -1914,7 +1914,7 @@ impl CodexMessageProcessor {
             personality,
         } = params;
 
-        let thread_history = if let Some(history) = history {
+        let mut thread_history = if let Some(history) = history {
             if history.is_empty() {
                 self.send_invalid_request_error(
                     request_id,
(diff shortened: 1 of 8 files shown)


owenlin0:
@codex review

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2144,6 +1988,24 @@ impl CodexMessageProcessor {
             }
         };
 
+        let has_request_overrides = request_overrides
```
pakrym-oai: should there be a way through comparing the turn_context object?

owenlin0: i ended up changing the approach on this PR, so this code is no longer needed :D

owenlin0:
@codex review

codex-rs/core/src/codex.rs
```diff
@@ -2330,6 +2375,11 @@ mod handlers {
             return;
         }
 
+        let initial_context_seeded = sess.state.lock().await.initial_context_seeded;
+        if !initial_context_seeded {
+            return;
```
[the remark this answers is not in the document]
owenlin0: that's fine for now

aibrahim-oai (approved):

PR #9951, opened 2026-01-27
fix: try to fix freezes 2

jif-oai:
Fixes a TUI freeze caused by awaiting `mpsc::Sender::send()` that blocks the tokio thread, stopping the consumption runtime and creating a deadlock. This could happen if the server was producing enough chunks to fill the `mpsc` fast enough. To solve this we try on insert using a `try_send()` (not requiring an `await`) and delegate to a tokio task if this does not work

This is a temporary solution as it can contain races for delta elements and a stronger design should come here

### DIFF
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -98,7 +98,7 @@ use tokio::sync::mpsc::unbounded_channel;
 use toml::Value as TomlValue;
 
 const EXTERNAL_EDITOR_HINT: &str = "Save and close external editor to continue.";
-const THREAD_EVENT_CHANNEL_CAPACITY: usize = 4;
+const THREAD_EVENT_CHANNEL_CAPACITY: usize = 1024;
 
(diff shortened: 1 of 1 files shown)


dylan-hurd-oai (approved):

codex-rs/tui/src/app.rs
```diff
@@ -714,8 +715,23 @@ impl App {
             guard.active
         };
 
-        if should_send && let Err(err) = sender.send(event).await {
-            tracing::warn!("thread {thread_id} event channel closed: {err}");
+        if should_send {
+            // Never await a bounded channel send on the main TUI loop: if the receiver falls behind,
+            // `send().await` can block and the UI stops drawing. If the channel is full, wait in a
+            // spawned task instead.
```
dylan-hurd-oai, line 721: @codex scan all uses of channels in codex-rs and propose an AGENTS.md update with best practices to improve our usage of them

PR #9959, opened 2026-01-27
linux-sandbox: allow workspace-write without enforced Landlock

AmirTlinov:
Fixes openai/codex#5824.

On kernels where Landlock returns `RulesetStatus::NotEnforced` (e.g. no Landlock support), `WorkspaceWrite` sandboxing previously failed hard even with network disabled.

This change:
- Allows `WorkspaceWrite` with `network_access=false` to continue when Landlock is `NotEnforced`.
- Still fails closed when `network_access=true`.
- Emits a one-time stderr warning when falling back (filesystem sandbox disabled; network remains restricted).
- Adds unit tests for the allow/deny gating + enforcement contract.

Tests: `cargo test -p codex-linux-sandbox`

### DIFF
--- a/codex-rs/linux-sandbox/src/landlock.rs
+++ b/codex-rs/linux-sandbox/src/landlock.rs
@@ -1,5 +1,6 @@
 use std::collections::BTreeMap;
 use std::path::Path;
+use std::sync::Once;
 
 use codex_core::error::CodexErr;
 use codex_core::error::Result;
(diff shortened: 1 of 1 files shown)


etraut-openai:
@AmirTlinov, thanks for taking the time to investigate this.

We've updated our [contribution guidelines](https://github.com/openai/codex/blob/main/docs/contributing.md) to indicate that we're no longer accepting unsolicited code contributions. Going forward, all code contributions are by invitation only. To read more about why we've taken this step, please refer to [this announcement](https://github.com/openai/codex/discussions/9956).

As for this particular fix, I'm not sure we want to allow Codex to run locally on systems that don't support Landlock. It might be preferable to block execution entirely. If we do allow this, we will need to make it very clear to users that they're running without a safety net. We'll need to discuss further with our internal security experts. This hasn't been a high priority issue for us to address because it doesn't affect many Codex users.

AmirTlinov:
Thanks for the clarification, and understood on the invitation-only policy.

For context: the underlying failure in #5824 happens when `landlock::Ruleset::restrict_self()` returns `RulesetStatus::NotEnforced` (e.g. kernels without Landlock support). My patch attempted a gated fallback for `WorkspaceWrite + network_access=false`, but I agree this is ultimately a product/security decision and the default may need to remain fail-closed.

Happy to provide a smaller change (e.g. clearer, actionable error messaging) or open a PR if/when the team decides on the preferred behavior and explicitly invites an external contribution.

PR #9960, opened 2026-01-27
fix: handle all web_search actions and in progress invocations

sayan-oai:
### Summary
- Parse all `web_search` tool actions (`search`, `find_in_page`, `open_page`).
  - Previously we only parsed + displayed `search`, which made the TUI appear to pause when the other actions were being used.
- Show in progress `web_search` calls as `Searching the web`
  - Previously we only showed completed tool calls

image

### Tests
Added + updated tests, tested locally

### Follow ups
Update VSCode extension to display these as well


codex-rs/core/src/context_manager/history.rs
```diff
@@ -75,6 +75,8 @@ impl ContextManager {
         self.items
             .retain(|item| !matches!(item, ResponseItem::GhostSnapshot { .. }));
         self.items
+            .retain(|item| !matches!(item, ResponseItem::WebSearchCall { action: None, .. }));
```
pakrym-oai: Why?

codex-rs/core/src/context_manager/history.rs
```diff
@@ -307,17 +309,17 @@ impl ContextManager {
 }
 
 /// API messages include every non-system item (user/assistant messages, reasoning,
-/// tool calls, tool outputs, shell calls, and web-search calls).
+/// tool calls, tool outputs, shell calls, and complete web-search calls).
```
pakrym-oai: How do incomplete calls get here?

codex-rs/core/src/context_manager/history.rs
sayan-oai: avoid adding `WebSearchCall`s with no actions to the prompt sent back to the model. if thats not a concern i can rm

codex-rs/exec/src/event_processor_with_jsonl_output.rs
```diff
@@ -195,11 +198,36 @@ impl EventProcessorWithJsonOutput {
...
+                id: ev.call_id.clone(),
+                query: String::new(),
+                action: WebSearchAction::Other,
+            }),
+        };
+
+        vec![ThreadEvent::ItemStarted(ItemStartedEvent { item })]
+    }
+
+    fn handle_web_search_end(&mut self, ev: &protocol::WebSearchEndEvent) -> Vec<ThreadEvent> {
```
pakrym-oai, line 220: Is there a test for begin/end events for items somewhere?

codex-rs/core/src/context_manager/history.rs
pakrym-oai: are these completed items?

sayan-oai: no, they're `in_progress`, we're pulling them off `response.output_item.added`. they can theoretically end up in history via compaction replacement histories or rollbacks.

codex-rs/core/src/context_manager/history.rs
sayan-oai: incomplete here just means `WebSearchCall`s without `action`, which is the form they take while running. `action` is only populated when the tool call completes.

codex-rs/protocol/src/models.rs
```diff
@@ -1063,26 +1085,43 @@ mod tests {
...
                 status: expected_status.clone(),
                 action: expected_action.clone(),
             };
             assert_eq!(parsed, expected);
 
             let serialized = serde_json::to_value(&parsed)?;
-            let original_value: serde_json::Value = serde_json::from_str(json_literal)?;
-            assert_eq!(serialized, original_value);
+            let mut expected_serialized: serde_json::Value = serde_json::from_str(json_literal)?;
+            if !expect_roundtrip && let Some(obj) = expected_serialized.as_object_mut() {
```
pakrym-oai, line 1105: Why don't we expect roundtrip?

codex-rs/protocol/src/models.rs
```diff
@@ -576,6 +578,22 @@ pub enum WebSearchAction {
     Other,
 }
 
+impl WebSearchAction {
+    pub fn detail(&self) -> String {
+        match self {
+            WebSearchAction::Search { query } => query.clone().unwrap_or_default(),
+            WebSearchAction::OpenPage { url } => url.clone().unwrap_or_default(),
+            WebSearchAction::FindInPage { url, pattern } => match (pattern, url) {
+                (Some(pattern), Some(url)) => format!("'{pattern}' in {url}"),
```
pakrym-oai: This looks like rendering does it belong here?

pakrym-oai (approved):

codex-rs/protocol/src/models.rs
sayan-oai: `id` isnt serialized on `WebSearchCall` so it's not present after serialization

codex-rs/core/src/context_manager/history.rs
pakrym-oai: Yeah, let's send them back.

codex-rs/protocol/src/models.rs
sayan-oai: will move

codex-rs/exec/src/event_processor_with_jsonl_output.rs
sayan-oai: theres a test for begin, added some for end -> item and the whole flow for begin -> end.

PR #9982, opened 2026-01-27
[skills] Auto install MCP dependencies when running skils with dependency specs.

mzeng-openai:
Auto install MCP dependencies when running skils with dependency specs.


codex-rs/core/src/mcp/mod.rs
```diff
@@ -181,6 +198,331 @@ pub async fn collect_mcp_snapshot(config: &Config) -> McpListToolsResponseEvent
...
+        .answers
+        .get(SKILL_MCP_DEPENDENCY_PROMPT_ID)
+        .is_some_and(|answer| {
+            answer
+                .answers
+                .iter()
+                .any(|entry| entry == MCP_DEPENDENCY_OPTION_INSTALL)
+        });
+
+    sess.record_mcp_dependency_prompted(missing.keys().cloned())
```
mzeng-openai: Save the mcps that are prompted so that we don't show the approval prompt for them again.

codex-rs/core/src/mcp/mod.rs
```diff
@@ -181,6 +198,331 @@ pub async fn collect_mcp_snapshot(config: &Config) -> McpListToolsResponseEvent
...
+
+        sess.notify_background_event(
+            turn_context,
+            format!(
+                "Authenticating MCP {name}... Follow instructions in your browser if prompted."
+            ),
+        )
+        .await;
+
+        if let Err(err) = perform_oauth_login(
```
mzeng-openai: Automatically trigger OAuth if available.

codex-rs/core/src/skills/injection.rs
```diff
@@ -76,23 +66,235 @@ fn emit_skill_injected_metric(otel: Option<&OtelManager>, skill: &SkillMetadata,
...
+    for skill in skills {
+        by_name_and_path.insert((skill.name.as_str(), skill.path.as_path()), skill);
+    }
+
     for input in inputs {
-        if let UserInput::Skill { name, path } = input
-            && seen.insert(name.clone())
-            && let Some(skill) = skills.iter().find(|s| s.name == *name && s.path == *path)
-            && !disabled_paths.contains(&skill.path)
+        match input {
```
mzeng-openai: Extending the detection from CLI-specific logic, to actually matching `$skill-name` patterns.

codex-rs/core/src/state/session.rs
```diff
@@ -15,6 +16,7 @@ pub(crate) struct SessionState {
     pub(crate) history: ContextManager,
     pub(crate) latest_rate_limits: Option<RateLimitSnapshot>,
     pub(crate) server_reasoning_included: bool,
+    pub(crate) mcp_dependency_prompted: HashSet<String>,
```
mzeng-openai, line 19: Keep track of which mcps have been prompted to be installed.

codex-rs/core/src/codex.rs
```diff
@@ -2041,6 +2058,44 @@ impl Session {
         Arc::clone(&self.services.user_shell)
     }
 
+    async fn refresh_mcp_servers_inner(
```
mzeng-openai, line 2116: extracted from below.

codex-rs/core/src/mcp/mod.rs
```diff
@@ -181,6 +198,331 @@ pub async fn collect_mcp_snapshot(config: &Config) -> McpListToolsResponseEvent
     snapshot
 }
 
+const SKILL_MCP_DEPENDENCY_PROMPT_ID: &str = "skill_mcp_dependency_install";
```
xl-openai: Maybe move all skill-specific MCP logic to a separate file.

codex-rs/core/src/skills/loader.rs
```diff
@@ -50,6 +54,23 @@ struct Interface {
     default_prompt: Option<String>,
 }
 
+#[derive(Debug, Default, Deserialize)]
+struct Dependencies {
+    #[serde(default)]
+    tools: Vec<DependencyTool>,
```
xl-openai, line 60: Should we call it Tools or just MCP? Tools is too generic right?

codex-rs/core/src/skills/injection.rs
```diff
@@ -76,23 +66,235 @@ fn emit_skill_injected_metric(otel: Option<&OtelManager>, skill: &SkillMetadata,
...
+            UserInput::Skill { name, path } => {
```
xl-openai: Internal client should already do the skill name parsing and set UserInput::Skill, not sure if adding server side parsing on  UserInput::Text is an overkill or not.

codex-rs/core/src/mcp/mod.rs
xl-openai: Wondering if we should always show the prompt if it is still not set up?

codex-rs/core/src/skills/loader.rs
mzeng-openai: I think we'll need to support connectors at least.

codex-rs/core/src/mcp/mod.rs
mzeng-openai: done

codex-rs/core/src/skills/injection.rs
```diff
@@ -85,14 +80,202 @@ fn collect_explicit_skill_mentions(
     let mut seen: HashSet<String> = HashSet::new();
 
     for input in inputs {
-        if let UserInput::Skill { name, path } = input
```
mzeng-openai, line 88: I actually realized it would still be helpful for the CLI to use this, once we have connectors we need to include it's path which the CLI doesn't support display.

codex-rs/core/src/mcp/mod.rs
mzeng-openai: We can add that later, for the first version I'm trying to play safe and try not to annoy the users with repeating prompts

xl-openai (approved):

PR #9985, opened 2026-01-27
Added support for live updates to skills and AGENTS

etraut-openai:
### Overview

Add a centralized FileWatcher in codex-core (using notify) that watches:
* `AGENTS.md` / `AGENTS.override.md` and project `AGENTS` search dirs
* Skill roots from the config layer stack (recursive)

Send `AgentsChanged` and `SkillsChanged` events when relevant file system changes are detected

On `SkillsChanged`:
* Invalidate the skills cache immediately in ThreadManager
* Emit EventMsg::SkillsUpdateAvailable to active sessions
* Broadcast a new app-server notification: skills/list/updated

On `AgentsChanged`:
* Set a per-session agents_changed flag scoped to relevant watch dirs
* On the next user turn, recompute skills + user instructions and inject an updated UserInstructions item into the event stream

Wire the watcher through `ThreadManager` -> `Codex::spawn` -> `SessionServices`, including sub-agent sessions.

Add `SkillsListUpdatedNotification` to the app-server protocol and gate broadcast until after initialize.

### Testing

I did a bunch of manual testing of both AGENTS and skills updates. They work surprisingly well. Skill addition and removal are both handled seamlessly in both the UI and in the behavior of the agent on the next turn. Changes to skill details were picked up and followed on the next turn. Additions and modifications to AGENTS.md were followed well. The only failure that I found was when I deleted a major block of instructions from AGENTS.md. In that case, the model still followed those instructions from the older AGENTS.md because it was still in the context window.


xl-openai (commented):
Thanks for making this change! The skills part LGTM!

codex-rs/core/src/codex.rs
```diff
@@ -806,6 +864,7 @@ impl Session {
             );
         }
         let state = SessionState::new(session_configuration.clone());
+        let agents_watch_dirs = Self::build_agents_watch_dirs(&config);
```
pakrym-oai: Aren't agents.md cwd-specific? Can we create the watcher only once for the entire session?

codex-rs/core/src/codex.rs
```diff
@@ -1665,6 +1734,32 @@ impl Session {
...
+            error!(
+                "failed to load skill {}: {}",
+                err.path.display(),
+                err.message
+            );
+        }
+        let enabled_skills = skills_outcome.enabled_skills();
+        let user_instructions = get_user_instructions(&config, Some(&enabled_skills)).await;
+        let mut state = self.state.lock().await;
+        state.session_configuration.user_instructions = user_instructions.clone();
```
pakrym-oai, line 2003: Should we let the model know which file names changed instead? And let it decide which ones it cares about?

pakrym-oai: The current logic can inadvertently create huge messages and flood the context. We've seen very large skills + agents combinations. 

If we inject the entire thing for every change we'll exhaust the context in no time.

pakrym-oai (changes requested):
Needs tests.

I think re-injecting full Agents.md and skills is too much of a risk/context bloat. We should consider only injecting updated file names for both skills and agents along with some short prompt and letting the model decide. 

I'd also prefer if we had a very aggressive throttle on file change events. If there is any bug in file_watcher it'll flood the session (some editors perform save operations that cause multiple file notifications, rename, create, delete etc). 

And we should consider shipping this behind a feature first.

codex-rs/core/src/codex.rs
etraut-openai: This function is taking cwd into consideration. It builds a list of all directories from the cwd up to the root of the repo and watches for changes in those directories. This mirrors the same logic used when enumerating the AGENTS files.

codex-rs/core/src/codex.rs
etraut-openai: I don't think it will work to only tell it which file names changed. We need to update the names and summaries of the skills. Both the name and summary are bounded in size (64 characters and 512 characters, respectively). We are not including the full skill file here.

We do include the full agents contents. I agree that's a concern in terms of size. But I don't see a way around that. We don't initially tell the model the file paths of the AGENTS files; we simply read the contents of all of these files, concatenate them, and then truncate to 32K. If we were to later tell it about file paths, it would have no way of correlating those paths with the original AGENTS text. I suppose we could modify the way we handle AGENTS, but that's a pretty big change and would probably be OOD for the model. Let me know if you have any other suggestions.

etraut-openai: Note that we're reinjecting only on turn boundaries. This acts as a "debounce" mechanism in the case where multiple file changes occur in quick succession. But I understand your concern about potential storms of file watcher events. Since we deliver a `SkillsListUpdatedNotification` app server notification each time, it could swap an app server client. I've added an additional one-second throttling/debounce mechanism.

etraut-openai:
@codex review

etraut-openai:
codex review

etraut-openai:
@pakrym-oai, I've made the following changes since the last review:
1. Added experimental feature flags `live_agents_reload` and `live_skills_reload`; they default to false for now
2. Added an "event coalescer" mechanism for file watcher events that limits them to one per second
3. Added integration tests
4. Added support for fallback filenames (a config option that allows users to specify alternative names for AGENTS.md)
5. Modified AGENTS.md for this repo to remove instructions about updating docs, since docs are no longer stored in this repo

I didn't make changes to the injection strategy (see responses above). Let me know if you have other thoughts here.

xiaotang-12-ops:
I’m pleasantly surprised to find that the LLM recognized my skill today.If possible, could we add a frontend display that indicates a specific skill has been invoked? This way, when a skill is triggered implicitly, users can clearly see that the large model has called the corresponding skill—and this would create a positive user experience.

codex-rs/cloud-requirements/src/lib.rs
```diff
@@ -120,8 +120,10 @@ impl CloudRequirementsService {
 
     async fn fetch(&self) -> Option<ConfigRequirementsToml> {
         let auth = self.auth_manager.auth().await?;
-        if !(auth.mode == AuthMode::ChatGPT
-            && auth.account_plan_type() == Some(PlanType::Enterprise))
+        if !(matches!(
+            auth.api_auth_mode(),
+            AuthMode::ChatGPT | AuthMode::ChatgptAuthTokens
```
pakrym-oai: oof, this is going to haunt us

codex-rs/core/src/rollout/truncation.rs
```diff
@@ -192,6 +192,16 @@ mod tests {
     async fn ignores_session_prefix_messages_when_truncating_rollout_from_start() {
         let (session, turn_context) = make_session_and_context().await;
         let mut items = session.build_initial_context(&turn_context).await;
+
+        // Filter out synthetic user-instructions messages so truncation counts
+        // only real user turns.
+        items.retain(|item| match item {
```
pakrym-oai, line 200: Why did we start needing this?

codex-rs/core/src/skills/manager.rs
```diff
@@ -120,8 +121,17 @@ impl SkillsManager {
 
     pub fn clear_cache(&self) {
         match self.cache_by_cwd.write() {
-            Ok(mut cache) => cache.clear(),
-            Err(err) => err.into_inner().clear(),
+            Ok(mut cache) => {
```
pakrym-oai: can we simplify this by returning cache from match and sharing clear/log logic?

codex-rs/core/src/rollout/truncation.rs
etraut-openai: This is required for some of the existing tests to accommodate the new skills & AGENTS injection.

PR #9998, opened 2026-01-27
Allow user config to override OpenAI provider

danbosscher:
### Issue summary
Customers using an API key in a US DR project need to connect to us.api.openai.com instead of api.openai.com.
I can confirm that Codex works with such a key by setting this environment variable:

OPENAI_BASE_URL="https://us.api.openai.com/v1" 

However, we should also be able to do this by overriding the openai provider in [config.toml](https://developers.openai.com/codex/config-sample).
Screenshot 2026-01-27 at 14 34 32

Codex does not appear to be picking up any changes to this provider in config.toml. 

### Motivation

- Users should be able to override the built-in OpenAI provider via `model_providers.openai.base_url` in `config.toml`, but the previous merge logic and ordering prevented that.

### Description

- Merge user `model_providers` into the built-in map using `insert` so entries in `cfg.model_providers` replace built-in providers instead of being ignored by `or_insert`.
- Apply the `ResponsesWebsockets` feature flag after merging user providers so feature-driven modifications don't clobber user overrides.
- Add a unit test `config_overrides_openai_base_url` that asserts a `[model_providers.openai] base_url = "https://us.api.openai.com/v1"` entry is respected at load time.
- Run `just fmt` to format the modified code.

### Testing

- Ran `just fmt` in the `codex-rs` workspace with no errors.
- Ran the crate test suite with `cargo test -p codex-core`, and the tests (unit suite for `codex-core`) completed successfully.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_i_6978c754b92083328dd6ee233e7d8ca1)

### DIFF
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -1355,15 +1355,15 @@ impl Config {
             || cfg.sandbox_mode.is_some();
 
         let mut model_providers = built_in_model_providers();
+        // Merge user-defined providers into the built-in list.
+        for (key, provider) in cfg.model_providers.into_iter() {
+            model_providers.insert(key, provider);
+        }
         if features.enabled(Feature::ResponsesWebsockets)
             && let Some(provider) = model_providers.get_mut("openai")
             && provider.is_openai()
         {
             provider.wire_api = crate::model_provider_info::WireApi::ResponsesWebsocket;
-        }
-        // Merge user-defined providers into the built-in list.
-        for (key, provider) in cfg.model_providers.into_iter() {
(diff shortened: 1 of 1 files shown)


danbosscher:

danbosscher:
Tested locally, marking ready for review

Screenshot 2026-01-27 at 15 22 40

hourianto:
This PR might be a more proper solution, but just letting people know that you can already make Codex work like that:

```
model_provider = "customname"

[model_providers.customname]
name = "OpenAI"
base_url = "https://url/v1"
wire_api = "responses"
```


This works fine for me for a custom endpoint.

PR #10001, opened 2026-01-27
backend-client: add get_config_requirements_file

gt-oai:
Adds getting config requirement to backend-client.

I made a slash command to test it (not included in this PR):
Screenshot 2026-01-27 at 15 20 41


codex-rs/codex-backend-openapi-models/src/models/config_file_response.rs
```diff
@@ -0,0 +1,40 @@
+/*
+ * codex-backend
```
jif-oai, line 2: huh?

jif-oai (approved):
Ok but where is it used?

codex-rs/codex-backend-openapi-models/src/models/config_file_response.rs
gt-oai: Generated from the internal OpenAPI spec

gt-oai:
> Ok but where is it used?

Nowhere yet, will be used to load requirements.toml from Codex Cloud.

PR #10003, opened 2026-01-27
feat: make it possible to specify --config flags in the SDK

bolinfest:
Updates the `CodexOptions` passed to the `Codex()` constructor in the SDK to support a `config` property that is a map of configuration data that will be transformed into `--config` flags passed to the invocation of `codex`.

Therefore, something like this:

```typescript
const codex = new Codex({
  config: {
    show_raw_agent_reasoning: true,
    sandbox_workspace_write: { network_access: true },
  },
});
```

would result in the following args being added to the invocation of `codex`:

```shell
--config show_raw_agent_reasoning=true --config sandbox_workspace_write.network_access=true
```

### DIFF
--- a/sdk/typescript/src/codexOptions.ts
+++ b/sdk/typescript/src/codexOptions.ts
@@ -1,7 +1,19 @@
+export type CodexConfigValue = string | number | boolean | CodexConfigValue[] | CodexConfigObject;
+
+export type CodexConfigObject = { [key: string]: CodexConfigValue };
+
 export type CodexOptions = {
   codexPathOverride?: string;
   baseUrl?: string;
   apiKey?: string;
+  /**
+   * Additional `--config key=value` overrides to pass to the Codex CLI.
+   *
+   * Provide a JSON object and the SDK will flatten it into dotted paths and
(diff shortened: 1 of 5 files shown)


sdk/typescript/src/codexOptions.ts
```diff
@@ -1,7 +1,24 @@
+export type CodexConfigValue =
```
lopopolo-openai: We should be able to provide a fully typed config interface and zod schema, including narrow enums of the permitted string literals in specific config positions. Can we push the model to do this?

For example, TypeScript can natively represent that sandbox has a fixed set of valid string literals values. Let's do that!

lopopolo-openai (changes requested):

sdk/typescript/src/codexOptions.ts
lopopolo-openai: Ideally with tests to ensure it matches the real config format!

bolinfest: That is not something we have the bandwidth to take on right now.

pakrym-oai (approved):

PR #10008, opened 2026-01-27
enable live web search for DangerFullAccess sandbox policy

sayan-oai:
Auto-enable live `web_search` tool when sandbox policy is `DangerFullAccess`.

Explicitly setting `web_search` (canonical setting), or enabling `web_search_cached` or `web_search_request` still takes precedence over this sandbox-policy-driven enablement.

### DIFF
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -1203,11 +1203,13 @@ pub fn resolve_oss_provider(
     }
 }
 
-/// Resolve the web search mode from explicit config and feature flags.
+/// Resolve the web search mode from explicit config, feature flags, and sandbox policy.
+/// Live search is auto-enabled when sandbox policy is `DangerFullAccess`
 fn resolve_web_search_mode(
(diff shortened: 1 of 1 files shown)


codex-rs/core/src/config/mod.rs
```diff
@@ -2302,11 +2308,50 @@ trust_level = "trusted"
...
+            ..Default::default()
+        };
+
+        let config = Config::load_from_base_config_with_overrides(
+            cfg,
+            ConfigOverrides::default(),
+            codex_home.path().to_path_buf(),
+        )?;
+
+        assert_eq!(config.web_search_mode, WebSearchMode::Cached);
```
pakrym-oai, line 2351: do we have integration tests for search, would be nice to have one

pakrym-oai (approved):

codex-rs/core/src/config/mod.rs
sayan-oai: ok will add

PR #10009, opened 2026-01-27
chore: introduce *Args types for new() methods

bolinfest:
Constructors with long param lists can be hard to reason about when a number of the args are `None`, in practice. Introducing a struct to use as the args type helps make things more self-documenting.


viyatb-oai (approved):
much better, thank you!

PR #10012, opened 2026-01-27
feat(app-server): support external auth mode

owenlin0:
This enables a new use case where `codex app-server` is embedded into a parent application that will directly own the user's ChatGPT auth lifecycle, which means it owns the user’s auth tokens and refreshes it when necessary. The parent application would just want a way to pass in the auth tokens for codex to use directly.

The idea is that we are introducing a new "auth mode" currently only exposed via app server: **`chatgptAuthTokens`** which consist of the `id_token` (stores account metadata) and `access_token` (the bearer token used directly for backend API calls). These auth tokens are only stored in-memory. This new mode is in addition to the existing `apiKey` and `chatgpt` auth modes. 

This PR reuses the shape of our existing app-server account APIs as much as possible:
- Update `account/login/start` with a new `chatgptAuthTokens` variant, which will allow the client to pass in the tokens and have codex app-server use them directly. Upon success, the server emits `account/login/completed` and `account/updated` notifications.
- A new server->client request called `account/chatgptAuthTokens/refresh` which the server can use whenever the access token previously passed in has expired and it needs a new one from the parent application.

I leveraged the core 401 retry loop which typically triggers auth token refreshes automatically, but made it pluggable:
- **chatgpt** mode refreshes internally, as usual.
- **chatgptAuthTokens** mode calls the client via `account/chatgptAuthTokens/refresh`, the client responds with updated tokens, codex updates its in-memory auth, then retries. This RPC has a 10s timeout and handles JSON-RPC errors from the client.

Also some additional things:
- chatgpt logins are blocked while external auth is active (have to log out first. typically clients will pick one OR the other, not support both)
- `account/logout` clears external auth in memory
- Ensures that if `forced_chatgpt_workspace_id` is set via the user's config, we respect it in both:
  - `account/login/start` with `chatgptAuthTokens` (returns a JSON-RPC error back to the client) 
  - `account/chatgptAuthTokens/refresh` (fails the turn, and on next request app-server will send another `account/chatgptAuthTokens/refresh` request to the client).

### DIFF
--- a/codex-rs/app-server-protocol/src/protocol/common.rs
+++ b/codex-rs/app-server-protocol/src/protocol/common.rs
@@ -28,6 +28,11 @@ impl GitSha {
 pub enum AuthMode {
     ApiKey,
     ChatGPT,
+    /// [UNSTABLE] FOR OPENAI INTERNAL USE ONLY - DO NOT USE.
+    #[serde(rename = "chatgptAuthTokens")]
+    #[ts(rename = "chatgptAuthTokens")]
+    #[strum(serialize = "chatgptAuthTokens")]
+    ChatgptAuthTokens,
 }
 
 /// Generates an `enum ClientRequest` where each variant is a request that the
@@ -534,6 +539,11 @@ server_request_definitions! {
         response: v2::DynamicToolCallResponse,
     },
 
+    ChatgptAuthTokensRefresh => "account/chatgptAuthTokens/refresh" {
+        params: v2::ChatgptAuthTokensRefreshParams,
+        response: v2::ChatgptAuthTokensRefreshResponse,
+    },
+
     /// DEPRECATED APIs below
     /// Request to approve a patch.
     /// This request is used for Turns started via the legacy APIs (i.e. SendUserTurn, SendUserMessage).
@@ -754,6 +764,29 @@ mod tests {
     }
 
     #[test]
+    fn serialize_chatgpt_auth_tokens_refresh_request() -> Result<()> {
+        let request = ServerRequest::ChatgptAuthTokensRefresh {
+            request_id: RequestId::Integer(8),
+            params: v2::ChatgptAuthTokensRefreshParams {
+                reason: v2::ChatgptAuthTokensRefreshReason::Unauthorized,
+                previous_account_id: Some("org-123".to_string()),
+            },
+        };
+        assert_eq!(
+            json!({
+                "method": "account/chatgptAuthTokens/refresh",
+                "id": 8,
+                "params": {
+                    "reason": "unauthorized",
+                    "previousAccountId": "org-123"
+                }
+            }),
+            serde_json::to_value(&request)?,
+        );
+        Ok(())
+    }
+
+    #[test]
     fn serialize_get_account_rate_limits() -> Result<()> {
         let request = ClientRequest::GetAccountRateLimits {
             request_id: RequestId::Integer(1),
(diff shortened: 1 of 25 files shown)


owenlin0:
@codex review

owenlin0:
@codex review

owenlin0:
@codex review

owenlin0:
@codex review

owenlin0:
@codex review

owenlin0:
@codex review

owenlin0:
@codex review

owenlin0:
@codex review

codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -28,6 +28,10 @@ impl GitSha {
 pub enum AuthMode {
     ApiKey,
     ChatGPT,
+    #[serde(rename = "chatgptAuthTokens")]
```
celia-oai, line 32: Can we add docstring here & in proto section showing that this is an internal-only feature and is very error prone, and list the conditions where users can use it:
1. cover codex's scope;
2. Codex treats it as read-only (i.e. does not attempt to refresh it but delegates that responsibility to the main app)

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -965,6 +992,92 @@ impl CodexMessageProcessor {
...
+            };
+            self.outgoing.send_error(request_id, error).await;
+            return;
+        }
+
+        // Cancel any active login attempt to avoid persisting managed auth state.
+        {
+            let mut guard = self.active_login.lock().await;
+            if let Some(active) = guard.take() {
+                drop(active);
```
celia-oai, line 1018: I believe this is only needed for signin with chatgpt, where you have a browser session open for user to log in & you close it out when user no longer needs it. this workflow should behave more like api workflow, where you just have to make a write the auth token to keyring or disk.

celia-oai:
chatted offline, we'll hide the ephemeral logic in `storage.rs` so it's abstracted away from `auth.rs`.

owenlin0:
@codex review

owenlin0:
@codex review

codex-rs/app-server/README.md
```diff
@@ -659,16 +659,25 @@ $demo-app Pull the latest updates from the team.
...
 
+### Authentication modes
+
+Codex supports three authentication modes. The current mode is surfaced in `account/updated` (`authMode`) and can be inferred from `account/read`.
+
+- **API key (`apiKey`)**: Caller supplies an OpenAI API key via `account/login/start` with `type: "apiKey"`. The API key is saved and used for API requests.
+- **ChatGPT managed (`chatgpt`)** (recommended): Codex owns the ChatGPT OAuth flow and refresh tokens. Start via `account/login/start` with `type: "chatgpt"`; Codex persists tokens to disk and refreshes them automatically.
+- **ChatGPT external (`chatgptAuthTokens`)** (for OpenAI internal use only): The client supplies existing ChatGPT auth tokens (`idToken` and `accessToken`) via `account/login/start` with `type: "chatgptAuthTokens"`. Codex uses the access token for API calls, parses the ID token for account metadata, stores the tokens only in memory, and requests refresh via `account/chatgptAuthTokens/refresh` when needed. This mode is useful when using Codex as an embedded agent within an application that already manages ChatGPT auth tokens. The access token must contain the same scopes that Codex-managed ChatGPT auth tokens have.
```
celia-oai: let's not add this to the readme at all? update the docstring on the protocol instead? Readme should only contain information relevant to external customers & developers imo

celia-oai:
the code lgtm, can we also test out that there's no regression & test out the new endpoints (log in, refresh) in a local build end-to-end?

codex-rs/app-server/README.md
owenlin0: fair, will remove

owenlin0:
@celia-oai yep, tested this build locally as well

codex-rs/app-server/README.md
```diff
@@ -690,7 +697,7 @@ Response examples:
 
 Field notes:
 
-- `refreshToken` (bool): set `true` to force a token refresh.
+- `refreshToken` (bool): set `true` to force a token refresh. Ignored when using external ChatGPT auth mode.
```
celia-oai: let's delete this sentence as well to avoid confusion

celia-oai (approved):

bolinfest:
I think we need some documentation that declares the intent behind this `AuthMode`. Does it represent:

1. the default Responses endpoint to use for the `"openai"` provider:

https://github.com/openai/codex/blob/b79bf69af6a68d162a95af883564d9e480948e62/codex-rs/core/src/model_provider_info.rs#L140-L144

2. features to show in the TUI (if there are any that are specific to ChatGPT users)?
3. how to get an auth token?

To me, it feels like `AuthMode` should specify 1 and 2, but `CodexAuth` should specify 3.

bolinfest:
This is maybe OK for now, but ultimately, I think we should be focused on pulling the auth tokens _out_ of the app server rather than pushing them _in_. That is, I would rather see a new `CodexAuth ` variant like `EndpointProvidedAuth` that means "rely on the endpoint to ensure HTTP requests are authenticated" that could be applicable to either `AuthMode::ApiKey` or `AuthMode::ChatGPT`.

PR #10016, opened 2026-01-27
Emit compaction as item lifecycle notifications

aibrahim-oai:
- Emit compaction as item started/completed with a new Compaction item and keep legacy thread/compacted.\n- Add v2 integration test coverage for the item lifecycle + legacy notification.


aibrahim-oai:
@codex review this

aibrahim-oai:
@codex review this and make sure we always send compacted start on both types of compaction (remote local and also inline self initiated)

PR #10017, opened 2026-01-27
chore: add preconfig hooks to test and add web search integration test

sayan-oai:
Add integration test for #10008 (live `web_search` when sandbox policy is `DangerFullAccess`).

This required adding preconfig hooks to `test_codex`, so I split it into a new PR.

We need preconfig hooks because the existing prebuild hooks run after the config is looaded and therefore cannot seed `config.toml` early enough to affect sandbox policy, which is what this test requires.

### DIFF
--- a/codex-rs/core/tests/suite/web_search.rs
+++ b/codex-rs/core/tests/suite/web_search.rs
@@ -10,6 +10,8 @@ use core_test_support::test_codex::test_codex;
 use pretty_assertions::assert_eq;
 use serde_json::Value;
 
+const CONFIG_TOML: &str = "config.toml";
+
 fn sse_completed(id: &str) -> String {
     load_sse_fixture_with_id("../fixtures/completed_template.json", id)
 }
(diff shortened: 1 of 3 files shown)


codex-rs/core/tests/suite/web_search.rs
```diff
@@ -86,3 +88,40 @@ async fn web_search_mode_takes_precedence_over_legacy_flags_in_request_body() {
...
+
+    let mut builder = test_codex()
+        .with_model("gpt-5-codex")
+        .with_config(|config| {
+            config.features.disable(Feature::WebSearchCached);
+            config.features.disable(Feature::WebSearchRequest);
+        })
+        .with_pre_config_hook(|home| {
+            let config_path = home.join(CONFIG_TOML);
+            std::fs::write(config_path, "sandbox_mode = \"danger-full-access\"\n")
```
pakrym-oai, line 108: can you set this value in with_config?

sayan-oai: `with_config` applies mutations after the config is loaded, but `web_search_mode` is determined at config load time. so we have to apply the config change before it's loaded.

PR #10035, opened 2026-01-27
fix(tui): tab submits when no task running in steer mode

joshka-oai:
When steer mode is enabled, Tab used to only queue while a task was running and otherwise did nothing. Treat Tab as an immediate submit when no task is running so input isn't dropped when the inflight turn ends mid-typing.

Adds a regression test and updates docs/tooltips.


charley-oai (approved):

joshka-oai:
Noted some internal feedback - need to ensure that this doesn't submit on !

PR #10038, opened 2026-01-27
tui: complete directories in @ popup

joshka-oai:
@ completion now includes directory entries so you can select a folder
and keep typing to reach a file. Directory matches are returned with a
trailing path separator to distinguish them from files, and the
composer avoids adding a trailing space when inserting a directory.

This is implemented in codex-file-search (new run_including_directories)
and wired into the TUI file search manager.

Tests: cargo test -p codex-file-search, cargo test -p codex-tui

### DIFF
--- a/codex-rs/file-search/src/lib.rs
+++ b/codex-rs/file-search/src/lib.rs
@@ -14,6 +14,7 @@ use nucleo::Utf32String;
 use nucleo::pattern::CaseMatching;
 use nucleo::pattern::Normalization;
 use serde::Serialize;
+use std::borrow::Cow;
 use std::num::NonZero;
 use std::path::Path;
 use std::path::PathBuf;
@@ -86,6 +87,7 @@ pub struct SessionOptions {
     pub threads: NonZero<usize>,
     pub compute_indices: bool,
     pub respect_gitignore: bool,
+    pub include_directories: bool,
 }
 
 impl Default for SessionOptions {
@@ -98,6 +100,7 @@ impl Default for SessionOptions {
             threads: NonZero::new(2).unwrap(),
             compute_indices: false,
             respect_gitignore: true,
+            include_directories: false,
         }
     }
 }
@@ -151,6 +154,7 @@ fn create_session_inner(
         threads,
         compute_indices,
         respect_gitignore,
+        include_directories,
     } = options;
 
     let override_matcher = build_override_matcher(search_directory, &exclude)?;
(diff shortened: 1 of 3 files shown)


codex-rs/file-search/src/lib.rs
```diff
@@ -86,6 +87,7 @@ pub struct SessionOptions {
     pub threads: NonZero<usize>,
     pub compute_indices: bool,
     pub respect_gitignore: bool,
+    pub include_directories: bool,
```
nornagon-openai, line 90: why is this optional? i feel like the behavior should always be the same.

codex-rs/file-search/src/lib.rs
```diff
@@ -401,20 +408,33 @@ fn walker_worker(
...
-            Ok(rel_path) => rel_path.to_str(),
+            Ok(rel_path) => {
+                if rel_path.as_os_str().is_empty() {
+                    return None;
+                }
+                let rel_str = rel_path.to_str()?;
+                if is_dir {
+                    let mut with_sep = rel_str.to_string();
+                    with_sep.push(std::path::MAIN_SEPARATOR);
+                    return Some(Cow::Owned(with_sep));
```
nornagon-openai, line 434: perhaps we should just annotate `is_dir` on the file match and do this in display...?

codex-rs/tui/src/bottom_pane/chat_composer.rs
```diff
@@ -1723,17 +1723,28 @@ impl ChatComposer {
             path.to_string()
         };
 
-        // Replace the slice `[start_idx, end_idx)` with the chosen path and a trailing space.
+        let ends_with_path_separator = path.ends_with(std::path::MAIN_SEPARATOR)
+            || path.ends_with('/')
+            || path.ends_with('\\');
+        let add_trailing_space = !ends_with_path_separator;
```
nornagon-openai, line 1729: i don't feel like the "add a space" behavior should differ for directories, i think a common usage for this is like "examine the code in @some/dir and [...]"

nornagon-openai (changes requested):

codex-rs/tui/src/bottom_pane/chat_composer.rs
joshka-oai: It's useful for Create @src/foo.rs (I can take a look at the CoT on this to see why Codex added it, that's post hoc rationalization from me)

codex-rs/file-search/src/lib.rs
joshka-oai: I think codex was likely preserving existing behavior of the tool here - not certain that's really necessary - will take another look later.

PR #10039, opened 2026-01-27
chore(arg0): advisory-lock janitor for codex tmp paths

viyatb-oai:
### What changed
- Switch the arg0 helper root from `~/.codex/tmp/path` to `~/.codex/tmp/path2`
- Add `Arg0PathEntryGuard` to keep both the `TempDir` and an exclusive `.lock` file alive for the process lifetime
- Add a startup janitor that scans `path2` and deletes only directories whose lock can be acquired

### Tests
- `cargo clippy -p codex-arg0`
- `cargo clippy -p codex-core`
- `cargo test -p codex-arg0`
- `cargo test -p codex-core`

### DIFF
--- a/codex-rs/arg0/src/lib.rs
+++ b/codex-rs/arg0/src/lib.rs
@@ -301,13 +301,7 @@ fn janitor_cleanup(temp_root: &Path) -> std::io::Result<()> {
 
 fn try_lock_dir(dir: &Path) -> std::io::Result<Option<File>> {
     let lock_path = dir.join(LOCK_FILENAME);
-    let lock_file = match File::options()
-        .read(true)
-        .write(true)
-        .create(true)
-        .truncate(false)
-        .open(&lock_path)
-    {
+    let lock_file = match File::options().read(true).write(true).open(&lock_path) {
         Ok(file) => file,
         Err(err) if err.kind() == std::io::ErrorKind::NotFound => return Ok(None),
         Err(err) => return Err(err),
@@ -317,5 +311,63 @@ fn try_lock_dir(dir: &Path) -> std::io::Result<Option<File>> {
         Ok(()) => Ok(Some(lock_file)),
         Err(std::fs::TryLockError::WouldBlock) => Ok(None),
         Err(err) => Err(err.into()),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::LOCK_FILENAME;
+    use super::janitor_cleanup;
+    use std::fs;
+    use std::fs::File;
+    use std::path::Path;
+
+    fn create_lock(dir: &Path) -> std::io::Result<File> {
+        let lock_path = dir.join(LOCK_FILENAME);
+        File::options()
+            .read(true)
+            .write(true)
+            .create(true)
+            .truncate(false)
+            .open(lock_path)
+    }
+
+    #[test]
+    fn janitor_skips_dirs_without_lock_file() -> std::io::Result<()> {
+        let root = tempfile::tempdir()?;
+        let dir = root.path().join("no-lock");
+        fs::create_dir(&dir)?;
+
+        janitor_cleanup(root.path())?;
+
+        assert!(dir.exists());
+        Ok(())
+    }
+
+    #[test]
+    fn janitor_skips_dirs_with_held_lock() -> std::io::Result<()> {
+        let root = tempfile::tempdir()?;
+        let dir = root.path().join("locked");
+        fs::create_dir(&dir)?;
+        let lock_file = create_lock(&dir)?;
+        lock_file.try_lock()?;
+
+        janitor_cleanup(root.path())?;
+
+        assert!(dir.exists());
+        Ok(())
+    }
+
+    #[test]
+    fn janitor_removes_dirs_with_unlocked_lock() -> std::io::Result<()> {
+        let root = tempfile::tempdir()?;
+        let dir = root.path().join("stale");
+        fs::create_dir(&dir)?;
+        create_lock(&dir)?;
+
+        janitor_cleanup(root.path())?;
+
+        assert!(!dir.exists());
+        Ok(())
     }
 }
--- a/codex-rs/arg0/src/lib.rs
+++ b/codex-rs/arg0/src/lib.rs
@@ -1,3 +1,4 @@
+use std::fs::File;
 use std::future::Future;
 use std::path::Path;
 use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


codex-rs/arg0/src/lib.rs
```diff
@@ -231,5 +272,102 @@ pub fn prepend_path_entry_for_codex_aliases() -> std::io::Result<TempDir> {
         std::env::set_var("PATH", updated_path_env_var);
     }
 
-    Ok(temp_dir)
+    Ok(Arg0PathEntryGuard::new(temp_dir, lock_file))
+}
+
+fn janitor_cleanup(temp_root: &Path) -> std::io::Result<()> {
```
bolinfest, line 268: Maybe add a docstring that, in practice, we expect this to be fast because, assuming things are getting cleaned up correctly, the number of entries in `temp_root` should be small (one per active `codex` process).

codex-rs/arg0/src/lib.rs
```diff
@@ -231,5 +272,102 @@ pub fn prepend_path_entry_for_codex_aliases() -> std::io::Result<TempDir> {
...
+        if !path.is_dir() {
+            continue;
+        }
+
+        // Skip the directory if locking fails or the lock is currently held.
+        let Some(_lock_file) = try_lock_dir(&path)? else {
+            continue;
+        };
+
+        std::fs::remove_dir_all(&path)?;
```
bolinfest: Because we are holding the lock, no other instance should be trying to `rm -rf` at the same time, right?

codex-rs/arg0/src/lib.rs
```diff
@@ -177,11 +204,25 @@ pub fn prepend_path_entry_for_codex_aliases() -> std::io::Result<TempDir> {
...
     let path = temp_dir.path();
 
+    let lock_path = path.join(LOCK_FILENAME);
+    let lock_file = File::options()
+        .read(true)
+        .write(true)
+        .create(true)
+        .truncate(false)
+        .open(&lock_path)?;
+    lock_file.try_lock()?;
```
bolinfest, line 214: Should we retry if `try_lock()` fails? Is it possible that another `codex` process could race and delete `lock_file` before we have a chance to lock it?

codex-rs/core/tests/suite/mod.rs
```diff
@@ -1,16 +1,51 @@
...
+
+const CODEX_HOME_ENV_VAR: &str = "CODEX_HOME";
+
 // This code runs before any other tests are run.
 // It allows the test binary to behave like codex and dispatch to apply_patch and codex-linux-sandbox
 // based on the arg0.
 // NOTE: this doesn't work on ARM
 #[ctor]
-pub static CODEX_ALIASES_TEMP_DIR: TempDir = unsafe {
+pub static CODEX_ALIASES_TEMP_DIR: TestCodexAliasesGuard = unsafe {
```
bolinfest, line 22: This change makes me a bit uncomfortable.

I have been fighting against the use of `std::env::set_var()` in our tests to ensure we don't have race conditions, though admittedly this is in `#[ctor]` so there should not be multiple threads yet.

I don't have a great alternative, but I'm going on the record that I don't like it...

Can you at least add some more comments that explains why we are doing this?

codex-rs/arg0/src/lib.rs
```diff
@@ -10,8 +11,34 @@ use tempfile::TempDir;
...
+
+impl Arg0PathEntryGuard {
+    fn new(temp_dir: TempDir, lock_file: File) -> Self {
+        Self {
+            temp_dir,
+            lock_file,
+        }
+    }
+
+    #[allow(dead_code)]
```
bolinfest: If this is actually dead code, can we remove it?

codex-rs/arg0/src/lib.rs
```diff
@@ -167,7 +194,7 @@ pub fn prepend_path_entry_for_codex_aliases() -> std::io::Result<TempDir> {
 
     std::fs::create_dir_all(&codex_home)?;
     // Use a CODEX_HOME-scoped temp root to avoid cluttering the top-level directory.
-    let temp_root = codex_home.join("tmp").join("path");
+    let temp_root = codex_home.join("tmp").join("path2");
```
bolinfest: I don't love the name `path2`, but I guess it's fine?

bolinfest (approved):
I have some nits, but thanks for the thoughtful implementation!

codex-rs/arg0/src/lib.rs
viyatb-oai: right. the behavior is already safe for same-dir janitor races.

codex-rs/arg0/src/lib.rs
viyatb-oai: janitor should be best effort right? instead of a retry, I can ignore `NotFound` on `remove_dir_all` (directory disappeared between scan/remove)

codex-rs/core/tests/suite/mod.rs
viyatb-oai: yeah agreed, I remember us discussing this in the other PR. I will add comments.

codex-rs/arg0/src/lib.rs
viyatb-oai: what about `arg0`?

codex-rs/arg0/src/lib.rs
viyatb-oai: will remove

codex-rs/arg0/src/lib.rs
viyatb-oai: added

codex-rs/arg0/src/lib.rs
bolinfest: I would agree it should be best effort, but as written, Codex will crash if it fails because this will return `Err`, right?

codex-rs/arg0/src/lib.rs
bolinfest: I like `arg0`!

codex-rs/arg0/src/lib.rs
viyatb-oai: i did this ^

bolinfest: Hmm, I guess we do this today:

```rust
    match prepend_path_entry_for_codex_aliases() {
        Ok(path_entry) => Some(path_entry),
        Err(err) => {
            // It is possible that Codex will proceed successfully even if
            // updating the PATH fails, so warn the user and move on.
            eprintln!("WARNING: proceeding, even though we could not update PATH: {err}");
            None
        }
    }
```

bolinfest: Though if `lock_file.try_lock()` returns `Err`, perhaps we should log, but still try to add the entries in the `for filename` loop below?

codex-rs/arg0/src/lib.rs
viyatb-oai: 🔨 done

PR #10064, opened 2026-01-28
TUI: Render request_user_input results in history and simplify interrupt handling

charley-oai:
## Summary
This PR improves the TUI experience for `request_user_input` by rendering submitted question/answer sets directly in conversation history with clear, structured formatting.

It also intentionally simplifies interrupt behavior for now: on `Esc` / `Ctrl+C`, the questions overlay interrupts the turn without attempting to submit partial answers.

Screenshot 2026-02-02 at 4 51 40 PM

## Scope
- TUI-only changes.
- No core/protocol/app-server behavior changes in this PR.
- Resume reconstruction of interrupted question sets is out of scope for this PR.

## What Changed
- Added a new history cell: `RequestUserInputResultCell` in `codex-rs/tui/src/history_cell.rs`.
- On normal `request_user_input` submission, TUI now inserts that history cell immediately after sending `Op::UserInputAnswer`.
- Rendering includes a `Questions` header with `answered/total` count.
- Rendering shows each question as a bullet item.
- Rendering styles submitted answer lines in cyan.
- Rendering styles notes (for option questions) as `note:` lines in cyan.
- Rendering styles freeform text (for no-option questions) as `answer:` lines in cyan.
- Rendering dims only the `(unanswered)` suffix.
- Rendering can include an interrupted suffix and summary text when the cell is marked interrupted.
- Rendering redacts secret questions as `••••••` instead of showing raw values.
- Added `wrap_with_prefix(...)` in `history_cell.rs` for wrapped prefixed lines.
- Added `split_request_user_input_answer(...)` in `history_cell.rs` for decoding `"user_note: ..."` entries.

## Interrupt Behavior (Intentional for this PR)
- `Esc` / `Ctrl+C` in the questions overlay now performs `Op::Interrupt` and exits the overlay.
- It does **not** submit partial/committed answers on interrupt.
- Added TODO comments in `request_user_input` overlay interrupt paths indicating where interrupted partial result emission should be reintroduced once core support is finalized.
- Queued `request_user_input` overlays are discarded on interrupt in the current behavior.

## Tests Updated
- Updated/added overlay tests in `codex-rs/tui/src/bottom_pane/request_user_input/mod.rs` to reflect interrupt-only behavior.
- Added helper assertion for interrupt-only event expectation.
- Existing submission-path tests now validate history insertion behavior and expected answer maps.

## Behavior Notes
- Completed question flows now produce a readable `Questions` block in transcript history.
- Interrupted flows currently do not persist partial answers to model-visible tool output.

## Follow-ups
- Reintroduce partial-answer-on-interrupt semantics once core can persist/sequence interrupted `request_user_input` outputs safely.
- Optionally add replay/resume rendering for interrupted question sets as a separate PR.

## Codex author
`codex fork 019bfb8d-2a65-7313-9be2-ea7100d19a61`


charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

codex-rs/tui/src/chatwidget.rs
```diff
@@ -785,6 +839,8 @@ impl ChatWidget {
             .set_history_metadata(event.history_log_id, event.history_entry_count);
         self.set_skills(None);
         self.bottom_pane.set_connectors_snapshot(None);
+        self.clear_pending_request_user_input_answers();
+        self.request_user_input_questions.clear();
```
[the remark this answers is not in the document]
charley-oai: Fine to not do for now

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

codex-rs/core/src/codex_delegate.rs
```diff
@@ -380,32 +380,33 @@ async fn handle_request_user_input(
...
-                .notify_user_input_response(sub_id, empty.clone())
-                .await;
-            empty
+            parent_session.cancel_request_user_input(sub_id).await;
+            None
         }
-        response = fut => response.unwrap_or_else(|| RequestUserInputResponse {
-            answers: HashMap::new(),
-        }),
+        response = fut => response,
```
charley-oai: - Previously, a cancel would fall through and still submit an empty response. That later
    got recorded as a real function_call_output (with interrupted=false), which is
    misleading.
  - Now, on cancel we explicitly remove the pending sender (cancel_request_user_input) and
    return None, so handle_request_user_input skips submitting anything.
  - The biased; ensures the cancel branch wins if both the cancel and the response become
    ready at the same time, so we don’t race into sending a response after cancellation.

charley-oai:
@codex review this

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2723,6 +2723,9 @@ pub struct ToolRequestUserInputAnswer {
 /// EXPERIMENTAL. Response payload mapping question ids to answers.
 pub struct ToolRequestUserInputResponse {
     pub answers: HashMap<String, ToolRequestUserInputAnswer>,
+    /// True when the questions UI was interrupted before all answers were provided.
```
pakrym-oai: can we store this as part of answer? Like not-provided?

codex-rs/core/src/codex.rs
```diff
@@ -1672,16 +1674,62 @@ impl Session {
...
             }
         };
+
         match entry {
             Some(tx_response) => {
                 tx_response.send(response).ok();
             }
             None => {
-                warn!("No pending user input found for sub_id: {sub_id}");
+                // No pending user-input sender means the tool response arrived after the
```
pakrym-oai: I need to review this more but this feels quite roundabout way to reinject the response.

pakrym-oai: I'd love if we can push the partial response through the same mechanism the full response comes back

codex-rs/app-server-protocol/src/protocol/v2.rs
charley-oai: I think we already represent skipped questions as empty list of answers and so it would be hard to distinguish between that and the questions UI having been interrupted (via esc / ctrl+c)

codex-rs/core/src/codex.rs
charley-oai: Why we can’t reuse the normal path after interrupt:

- The normal path requires a live oneshot receiver (rx_response) held by
  RequestUserInputHandler.
- On interrupt, await_user_input_with_cancel fires, cancels the handler, and removes the
  pending sender; the receiver is dropped.
- There is no place to send the response after cancel.
- Keeping the handler alive would cause the model to continue the cancelled turn once the tool
  output arrives — which is exactly what the user tried to stop.

Why we don’t submit answers immediately on Esc:

- Sending tool output immediately would resume the assistant’s response without the user’s new
  prompt, effectively ignoring the cancel.
- It would also reorder the conversation (tool output → assistant reply → user message), which
  is not what the user intended.
- The current design lets the user interrupt, add their own prompt, and then the model sees:
  tool output first, then the new user message, which preserves tool‑call ordering.

How the model receives partial answers:

- The queued partial answers are serialized as a normal FunctionCallOutput with the original
  call_id.
- That output is stored in history and included in the next request’s input items.
- The model gets the same structure as a normal tool output, but with interrupted=true and only
  committed answers.

charley-oai:
@codex review this

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -297,14 +298,14 @@ pub(crate) async fn apply_bespoke_event_handling(
...
-                    item_id: request.call_id,
+                    item_id: call_id.clone(),
                     questions,
                 };
                 let rx = outgoing
                     .send_request(ServerRequestPayload::ToolRequestUserInput(params))
                     .await;
                 tokio::spawn(async move {
-                    on_request_user_input_response(event_turn_id, rx, conversation).await;
+                    on_request_user_input_response(event_turn_id, call_id, rx, conversation).await;
```
shijie-oai: hm I am surprised that I did not add this before lol

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

codex-rs/tui/src/chatwidget.rs
```diff
@@ -3449,8 +3447,8 @@ impl ChatWidget {
             EventMsg::CollabCloseBegin(_) => {}
             EventMsg::CollabCloseEnd(ev) => self.on_collab_event(collab::close_end(ev)),
             EventMsg::ThreadRolledBack(_) => {}
-            EventMsg::RawResponseItem(_)
-            | EventMsg::ItemStarted(_)
+            EventMsg::RawResponseItem(_) => {}
```
pakrym-oai: unrelated?

codex-rs/protocol/src/protocol.rs
```diff
@@ -222,6 +222,9 @@ pub enum Op {
     UserInputAnswer {
         /// Turn id for the in-flight request.
         id: String,
+        /// Call id for the original tool invocation, if known.
+        #[serde(default)]
+        call_id: Option<String>,
```
pakrym-oai: Why isn't id enough?

pakrym-oai: call_id feels like an implementation details leaking

codex-rs/core/src/codex.rs
```diff
@@ -1660,6 +1660,7 @@ impl Session {
     pub async fn notify_user_input_response(
         &self,
         sub_id: &str,
+        _call_id: Option<String>,
```
pakrym-oai: rem?

codex-rs/protocol/src/protocol.rs
charley-oai: id here is the turn/sub_id (the request id for the turn), not the tool call id. A single turn can emit multiple tool calls (including multiple request_user_input), and they all share the same id. The model and history pairing, however, is keyed by call_id (the tool invocation id).

So:

- id is enough only to route a response to the pending oneshot for that turn.
- call_id is needed to associate a response with the specific tool call in history or when there are multiple request_user_input calls in the same turn, or when responses arrive late (no pending oneshot).

codex-rs/core/src/codex.rs
charley-oai: pushed fix

codex-rs/tui/src/chatwidget.rs
charley-oai: pushed fix

charley-oai:
@codex review this

codex-rs/core/src/codex.rs
charley-oai: Marking as resolved as we no longer persist answers upon esc

codex-rs/protocol/src/protocol.rs
charley-oai: Marking as resolved as we no longer persist answers upon esc

pakrym-oai (approved):

PR #10080, opened 2026-01-28
Upgrade to rust 1.93

zbarsky-openai:
I needed to upgrade bazel one to get gnullvm artifacts and then noticed monorepo had drifted forward. They should move in lockstep. Also 1.93 already shipped so we can try that instead.


bolinfest (approved):

bolinfest:
Wow, we didn't trigger any new Clippy errors?

zbarsky-openai:
> Wow, we didn't trigger any new Clippy errors?

Guess not, it's a baby bump

PR #10086, opened 2026-01-28
feat: add log db

jif-oai:
Add a log DB. The goal is just to store our logs in a `.sqlite` DB to make it easier to crawl them and drop the oldest ones.


jif-oai:
@codex review

aibrahim-oai (approved):

PR #10097, opened 2026-01-28
chore: deprecate old web search feature flags

sayan-oai:
deprecate all old web search flags and aliases, including:
- `[features].web_search_request` and `[features].web_search_cached`
- `[tools].web_search`
- `[features].web_search`

slightly rework `legacy_usages` to enable pointing to non-features from deprecated features; we need to point to `web_search` (not under `[features]`) from things like `[features].web_search_cached` and `[features].web_search_request`.

Added integration tests to confirm deprecation notice is shown on explicit enablement and disablement of deprecated flags.

### DIFF
--- a/codex-rs/core/src/features.rs
+++ b/codex-rs/core/src/features.rs
@@ -148,6 +148,8 @@ impl Feature {
 pub struct LegacyFeatureUsage {
     pub alias: String,
     pub feature: Feature,
+    pub summary: String,
+    pub details: Option<String>,
 }
 
 /// Holds the effective set of enabled features.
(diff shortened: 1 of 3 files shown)


codex-rs/core/src/features.rs
```diff
@@ -301,6 +319,42 @@ impl Features {
     }
 }
 
+fn legacy_usage_notice(alias: &str, feature: Feature) -> (String, Option<String>) {
```
pakrym-oai, line 322: I don't know how I feel about rendering inside core but this logic has to be somewhere.

pakrym-oai (approved):

codex-rs/core/src/features.rs
sayan-oai: yea, we emit from core. i think having it next to the rest of the logic is nicer than hiding it in another module for now.

PR #10098, opened 2026-01-28
[bazel] Improve runfiles handling

zbarsky-openai:
we can't use runfiles directory on Windows due to path lengths, so swap to manifest strategy. Parsing the manifest is a bit complex and the format is changing in Bazel upstream, so pull in the official Rust library (via a small hack to make it importable...) and cleanup all the associated logic to work cleanly in both bazel and cargo without extra confusion


codex-rs/utils/cargo-bin/src/lib.rs
```diff
@@ -72,6 +58,27 @@ fn cargo_bin_env_keys(name: &str) -> Vec<String> {
     keys
 }
 
+fn resolve_bin_from_env(key: &str, value: OsString) -> Result<PathBuf, CargoBinError> {
+    let raw = PathBuf::from(&value);
+    if std::env::var_os("RUN_UNDER_RUNFILES").is_some() {
```
bolinfest: Let's make this a constant with a docstring?

codex-rs/utils/cargo-bin/src/lib.rs
```diff
@@ -84,97 +91,48 @@ fn cargo_bin_env_keys(name: &str) -> Vec<String> {
...
-}
-
-fn absolutize_from_buck_or_cwd(path: PathBuf) -> Result<PathBuf, CargoBinError> {
-    if path.is_absolute() {
-        return Ok(path);
-    }
-
-    if let Some(root) =
-        buck_project_root().map_err(|source| CargoBinError::CurrentExe { source })?
+#[doc(hidden)]
```
bolinfest: `#[doc(hidden)]` is unnecessary for our purposes, I believe.

codex-rs/Cargo.toml
```diff
@@ -185,12 +181,12 @@ regex = "1.12.2"
 regex-lite = "0.1.8"
 reqwest = "0.12"
 rmcp = { version = "0.12.0", default-features = false }
+runfiles = { git = "https://github.com/dzbarsky/rules_rust", rev = "b56cbaa8465e74127f1ea216f813cd377295ad81" }
```
bolinfest, line 189: What is our long-term plan for this dep?

.bazelrc
```diff
@@ -4,11 +4,14 @@ common --repo_env=BAZEL_NO_APPLE_CPP_TOOLCHAIN=1
 common --disk_cache=~/.cache/bazel-disk-cache
 common --repo_contents_cache=~/.cache/bazel-repo-contents-cache
 common --repository_cache=~/.cache/bazel-repo-cache
-common --remote_cache_compression
 startup --experimental_remote_repo_contents_cache
 
 common --experimental_platform_in_output_dir
 
+# Runfile paths are too long on Windows, so we can't enable them there. Disable directory strategy everywhere for consistency.
+# Instead we use the runfiles manifest strategy.
```
bolinfest: Can you cite a path that is likely to be stable: maybe utils/cargo-bin/README.md (which would need to be created?)

bolinfest (approved):

codex-rs/utils/cargo-bin/src/lib.rs
zbarsky-openai: ok will do, but that can only be used in this method, not in the macro (inlined to callers)

codex-rs/Cargo.toml
zbarsky-openai: there's a few potential options:
1) We figure out who owns the `runfiles` crate on crates.io and get it bumped
2) We upstream this fix back to rules_rust so it's consumable from the "official" github
3) We make it part of `rules_rs` and that becomes "official" as it continues to get traction

By the way, this is only necessary for cargo test and non-bazel clippy, so if there's a world where those flows are both bazelized, this can be removed (I will teach rules_rs to alias `@crates//:runfiles` to `@rules_rust//rust/runfiles` instead of the crate from crates.io)

PR #10113, opened 2026-01-28
chore: make homebrew higher priority for upgrade banner

apanasenko-oai:


pakrym-oai (approved):

pakrym-oai (changes requested):
the comment from codex is right

PR #10120, opened 2026-01-28
feat: fire tracking events for skill invocation

alexsong-oai:


alexsong-oai:
@codex review

alexsong-oai:
@codex review

alexsong-oai:
@codex review

codex-rs/core/src/analytics_client.rs
```diff
@@ -0,0 +1,182 @@
...
+                conversation_id: tracking.conversation_id.clone(),
+                gizmo_id: None,
+                gizmo_type: None,
+                message_id: None,
+            },
+        });
+    }
+
+    let base_url = config.chatgpt_base_url.trim_end_matches('/');
+    let url = if base_url.ends_with("/backend-api/codex") {
```
xl-openai: Do we really need this if-else? why the chatgpt_base_url could have different formats?

codex-rs/core/src/codex.rs
```diff
@@ -3179,10 +3180,33 @@ pub(crate) async fn run_turn(
...
     let otel_manager = turn_context.client.get_otel_manager();
+    let auth = match turn_context.client.get_auth_manager() {
+        Some(auth_manager) => auth_manager.auth().await,
+        None => None,
+    };
+    let auth_mode = auth.as_ref().map(|auth| auth.mode);
+    let access_token = auth.as_ref().and_then(|auth| auth.get_token().ok());
+    let account_id = auth.as_ref().and_then(CodexAuth::get_account_id);
+    let config = sess.get_config().await;
+    let tracking = TrackEventsContext {
```
xl-openai: We can just pass auth rather than auth_mode/access_token/account_id.

codex-rs/core/src/analytics_client.rs
```diff
@@ -0,0 +1,182 @@
...
+    } else {
+        "personal".to_string()
+    };
+    let raw_id = format!("{prefix}_{path}_{skill_name}");
+    let mut hasher = Sha1::new();
+    hasher.update(raw_id.as_bytes());
+    format!("{:x}", hasher.finalize())
+}
+
+fn skill_id_path_for_local_skill(
```
xl-openai: Need a test.

codex-rs/core/src/analytics_client.rs
```diff
@@ -0,0 +1,182 @@
...
+            event_type: "skill_invocation",
+            //TODO: add skill_name_encrypted
+            event_params: TrackEventParams {
+                skill_id,
+                skill_scope: skill_scope.to_string(),
+                product_surface: tracking.session_source.to_string(),
+                product_client_id: tracking.product_client_id.clone(),
+                model_slug: tracking.model_slug.clone(),
+                conversation_id: tracking.conversation_id.clone(),
+                gizmo_id: None,
```
xl-openai: Why do you define them if we don't need?

alexsong-oai: they're defined as required but nullable fields in SkillResourceInvocationEventParam

codex-rs/core/src/analytics_client.rs
alexsong-oai: Ideally it shouldn't, although it can be override in .codex/config.toml. Removing the if-else

codex-rs/core/src/analytics_client.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    fn expected_absolute_path(path: &PathBuf) -> String {
+        std::fs::canonicalize(path)
+            .unwrap_or_else(|_| path.to_path_buf())
+            .to_string_lossy()
+            .replace('\\', "/")
+    }
+
+    #[test]
+    fn normalize_path_for_skill_id_repo_scoped_uses_relative_path() {
```
xl-openai, line 284: Add one more test when repo_root is not part of skill_path

codex-rs/core/src/codex.rs
```diff
@@ -3179,10 +3180,28 @@ pub(crate) async fn run_turn(
     .await;
 
     let otel_manager = turn_context.client.get_otel_manager();
+    let auth = match turn_context.client.get_auth_manager() {
+        Some(auth_manager) => auth_manager.auth().await,
+        None => None,
+    };
+    let config = sess.get_config().await;
+    let tracking = TrackEventsContext {
```
xl-openai: Maybe move this to a helper function in analytics_client.rs

xl-openai (approved):

codex-rs/core/src/analytics_client.rs
xl-openai: Wondering if we should decouple from SkillResourceInvocationEventParam

codex-rs/core/src/skills/injection.rs
```diff
@@ -17,6 +22,8 @@ pub(crate) struct SkillInjections {
 pub(crate) async fn build_skill_injections(
     mentioned_skills: &[SkillMetadata],
     otel: Option<&OtelManager>,
+    tracking: Option<&TrackEventsContext>,
```
pakrym-oai: ideally you inject some kind of initialized client here instead of having to carry all this context.

codex-rs/core/src/analytics_client.rs
```diff
@@ -0,0 +1,326 @@
...
+struct TrackEventsJob {
+    config: Arc<Config>,
+    tracking: TrackEventsContext,
+    invocations: Vec<SkillInvocation>,
+}
+
+const ANALYTICS_EVENTS_QUEUE_SIZE: usize = 256;
+const ANALYTICS_EVENTS_TIMEOUT: Duration = Duration::from_secs(10);
+
+static ANALYTICS_EVENTS_SENDER: OnceLock<mpsc::Sender<TrackEventsJob>> = OnceLock::new();
```
pakrym-oai: can we avoid the static? check out how other clients are stored on SessionServices.

codex-rs/core/src/codex.rs
```diff
@@ -3385,10 +3389,32 @@ pub(crate) async fn run_turn(
     .await;
 
     let otel_manager = turn_context.client.get_otel_manager();
+    let auth = match turn_context.client.get_auth_manager() {
```
pakrym-oai: consider creating the client early (on session creation) and storing it.

pakrym-oai (changes requested):
let's have the client be the thing we store and use to emit events

codex-rs/core/src/codex.rs
```diff
@@ -3385,10 +3392,29 @@ pub(crate) async fn run_turn(
...
+        .analytics_events_client
+        .update_tracking(tracking.clone());
     let SkillInjections {
         items: skill_items,
         warnings: skill_warnings,
-    } = build_skill_injections(&mentioned_skills, Some(&otel_manager)).await;
+    } = build_skill_injections(
+        &mentioned_skills,
+        Some(&otel_manager),
+        Some(&sess.services.analytics_events_client),
```
pakrym-oai: is it always some? Can we remove the optional?

codex-rs/core/src/codex.rs
```diff
@@ -3385,10 +3392,29 @@ pub(crate) async fn run_turn(
...
+        None => None,
+    };
+    let thread_id = sess.conversation_id.to_string();
+    let tracking = build_track_events_context(
+        auth,
+        turn_context.client.get_model(),
+        thread_id,
+        crate::default_client::originator().value,
+    );
+    sess.services
```
pakrym-oai: what happens with `update_tracking` when there are multiple thread running in parallel?

codex-rs/core/src/codex.rs
```diff
@@ -3385,10 +3392,29 @@ pub(crate) async fn run_turn(
     .await;
 
     let otel_manager = turn_context.client.get_otel_manager();
+    let auth = match turn_context.client.get_auth_manager() {
+        Some(auth_manager) => auth_manager.auth().await,
```
pakrym-oai: Can you pass the auth manager into the client on creation?

codex-rs/core/src/codex.rs
```diff
@@ -3385,10 +3392,29 @@ pub(crate) async fn run_turn(
...
+        None => None,
+    };
+    let thread_id = sess.conversation_id.to_string();
+    let tracking = build_track_events_context(
+        auth,
+        turn_context.client.get_model(),
+        thread_id,
+        crate::default_client::originator().value,
```
pakrym-oai: Why does it have to be a parameter?

pakrym-oai: Can we just grab the same value right where we are sending the request?

codex-rs/core/src/codex.rs
pakrym-oai: It possibly fine but still a bit strange to be mutating the client. Can we pass this value as the parameter to track_skill?

codex-rs/core/src/codex.rs
```diff
@@ -904,6 +907,10 @@ impl Session {
             mcp_connection_manager: Arc::new(RwLock::new(McpConnectionManager::default())),
             mcp_startup_cancellation_token: Mutex::new(CancellationToken::new()),
             unified_exec_manager: UnifiedExecProcessManager::default(),
+            analytics_events_client: AnalyticsEventsClient::new(
+                AnalyticsEventsQueue::new(),
```
pakrym-oai: can this initialization happen inside AnalyticsEventsClient::new ?

pakrym-oai (approved):
Approved with comments.

PR #10125, opened 2026-01-28
emit a metric when we can't spawn powershell

iceweasel-oai:
This will help diagnose and measure the impact of a user-reported bug with the elevated sandbox and powershell


iceweasel-oai:
@codex review

dylan-hurd-oai (approved):

PR #10126, opened 2026-01-29
fix(ci) Get Windows ci stable again

dylan-hurd-oai:
## Summary
- relax mock responses call count in jsonrpc flow test to avoid flaky wiremock expectation

## Testing
- just fmt


iceweasel-oai:
not familiar with what's been going on in CI. I'm fine to stamp this. Two thoughts are
- would be nice to have the TIMEOUT values defined in one place, instead of all over the place
- for less ugliness, you could change the test timeouts to a higher value for all platforms instead of just for Windows

iceweasel-oai (approved):
left 2 comments, but they are non-blocking

PR #10127, opened 2026-01-29
(WIP) Handling of env variable snapshots that contain newlines and quotes

eternal-openai:
I noticed that the env variable snapshotting expected that each export is a single line, which probably works okay, but could be more robust. This PR:

 - Tries to detect unbalanced number of quotes, nested quotes
 - Detects line continuations in shell (backslash)
 - Adds a test


eternal-openai:

PR #10128, opened 2026-01-29
removing quit from dropdown menu, but not autocomplete [cli]

natea-oai:
Currently we have both `\quit` and `\exit` which do the same thing. This removes `\quit` from the slash command menu but allows it to still be an autocomplete option & working for those used to that command.

`/quit` autocomplete:
Screenshot 2026-01-28 at 4 32 53 PM

slash command menu:
Screenshot 2026-01-28 at 4 32 36 PM


natea-oai:

aibrahim-oai (approved):

PR #10132, opened 2026-01-29
Reduce Win exec_command errors - os/env specific shell prompting

canvrno-oai:
When a new thread is created, we include the **current working directory** and **shell** details in `environment_context`. This gives the model important context for choosing valid commands. However, the system prompts include shell command guidance and examples (for example, rg) that are oriented toward macOS/Linux bash/zsh environments. On Windows (where the shell reported to the model is typically PowerShell or cmd) those examples can conflict with environment_context, leading to incorrect `exec_command` tool calls and noisy “command not found” outputs.

This change makes the shell command guidance in the prompt shell-aware at thread creation time. The prompt now uses placeholders for the relevant bullets, and render_base_instructions in codex.rs replaces them with PowerShell, cmd, or bash/zsh-appropriate guidance where appropriate.

Future discussion: Move prompt rendering into a small _prompt renderer_ module that:
- Accepts a typed context (OS, shell, tool capabilities, sandbox mode)
- Performs placeholder substitution in one place
- Has focused unit tests that snapshot the rendered prompt for key shells/OSes


canvrno-oai:

canvrno-oai:
recheck

daniel-oai:

jta5456:
This is a great change in the right direction to fix all these windows shell issues.

Why do you include nl and wc for the powershell prompt? I'd argue too that we should keep rg for windows since it is considerable faster that the powershell alternatives.

PR #10136, opened 2026-01-29
Use active cell for context compaction

aibrahim-oai:
- Render context compaction as an active cell that updates in place.\n- Ignore ContextCompacted and finalize on ItemCompleted.


joshka-oai (approved):
Code changes LGTM - perhaps let codex sprinkle some context dust for future maintainability (i.e. why we're ignoring the contextcompacted event - because we're replacing that with start/

PR #10139, opened 2026-01-29
Fall back to http when websockets fail

pakrym-oai:
I expect not all proxies work with websockets, fall back to http if websockets fail.


codex-rs/core/src/transport_manager.rs
```diff
@@ -0,0 +1,31 @@
...
+use std::sync::atomic::Ordering;
+
+use crate::model_provider_info::WireApi;
+
+#[derive(Clone, Debug, Default)]
+pub struct TransportManager {
+    fallback_to_http: Arc<AtomicBool>,
+}
+
+impl TransportManager {
```
pakrym-oai, line 12: not in love with this single field class that has to be dragged everywhere but we need a bit that is persisted throughout the session

aibrahim-oai (approved):

PR #10142, opened 2026-01-29
feat(secrets): add codex-secrets crate

viyatb-oai:
## Summary
This introduces the first working foundation for Codex managed secrets: a small Rust crate that can securely store and retrieve secrets locally.

Concretely, it adds a `codex-secrets` crate that:
- encrypts a local secrets file using `age`
- generates a high-entropy encryption key
- stores that key in the OS keyring

## What this enables
- A secure local persistence model for secrets
- A clean, isolated place for future provider backends
- A clear boundary: Codex can become a credential broker without putting plaintext secrets in config files

## Implementation details
- New crate: `codex-rs/secrets/`
- Encryption: `age` with scrypt recipient/identity
- Key generation: `OsRng` (32 random bytes)
- Key storage: OS keyring via `codex-keyring-store`

## Testing
- `cd codex-rs && just fmt`
- `cd codex-rs && cargo test -p codex-secrets`


codex-rs/secrets/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+    Global,
+    Environment(String),
+}
+
+impl SecretScope {
+    pub fn environment(environment_id: impl Into<String>) -> Result<Self> {
+        let env_id = environment_id.into();
+        let trimmed = env_id.trim();
+        anyhow::ensure!(!trimmed.is_empty(), "environment id must not be empty");
+        Ok(Self::Environment(trimmed.to_string()))
```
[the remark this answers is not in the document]
xl-openai: environment ID is a Sha256 hash that will not have / right?

codex-rs/secrets/src/lib.rs
```diff
@@ -0,0 +1,227 @@
...
+mod tests {
+    use super::*;
+    use codex_keyring_store::tests::MockKeyringStore;
+    use pretty_assertions::assert_eq;
+
+    #[test]
+    fn environment_id_fallback_has_cwd_prefix() {
+        let dir = tempfile::tempdir().expect("tempdir");
+        let env_id = environment_id_from_cwd(dir.path());
+        assert!(env_id.starts_with("cwd-"));
```
xl-openai: Maybe assert the full env_id (with the hash).

codex-rs/secrets/src/local.rs
```diff
@@ -0,0 +1,248 @@
...
+    fn save_file(&self, file: &SecretsFile) -> Result<()> {
+        let dir = self.secrets_dir();
+        fs::create_dir_all(&dir)
+            .with_context(|| format!("failed to create secrets dir {}", dir.display()))?;
+
+        let passphrase = self.load_or_create_passphrase()?;
+        let plaintext = serde_json::to_vec(file).context("failed to serialize secrets file")?;
+        let ciphertext = encrypt_with_passphrase(&plaintext, &passphrase)?;
+        let path = self.secrets_path();
+        fs::write(&path, ciphertext)
```
xl-openai: Should we write to a temp file first then swap?

codex-rs/secrets/src/local.rs
```diff
@@ -0,0 +1,248 @@
...
+            .with_context(|| format!("failed to write secrets file at {}", path.display()))?;
+        Ok(())
+    }
+
+    fn load_or_create_passphrase(&self) -> Result<SecretString> {
+        let account = compute_keyring_account(&self.codex_home);
+        match self
+            .keyring_store
+            .load(keyring_service(), &account)
+            .map_err(|err| anyhow::anyhow!(err.message()))?
```
xl-openai: To confirm, this implementation will error on keyring store not available right?

viyatb-oai: correct, added an explicit error and fails fast when keyring load fails.

codex-rs/secrets/src/local.rs
viyatb-oai: yeah agreed -  now writes to a temp file and then renames into place (write_file_atomically), instead of writing directly to the target path.

codex-rs/secrets/src/lib.rs
viyatb-oai: fixed

codex-rs/secrets/src/local.rs
```diff
@@ -0,0 +1,248 @@
...
+        let mut parsed: SecretsFile = serde_json::from_slice(&plaintext).with_context(|| {
+            format!(
+                "failed to deserialize decrypted secrets file at {}",
+                path.display()
+            )
+        })?;
+        if parsed.version == 0 {
+            parsed.version = SECRETS_VERSION;
+        }
+        Ok(parsed)
```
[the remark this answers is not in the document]
viyatb-oai: Good catch — fixed.
load_file() now explicitly rejects files with version > SECRETS_VERSION to prevent silent downgrade corruption.
We still keep the existing v0 -> current migration path.
Added regression test: load_file_rejects_newer_schema_versions

codex-rs/secrets/src/lib.rs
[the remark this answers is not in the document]
viyatb-oai: correct.

codex-rs/secrets/src/lib.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+#[derive(Clone)]
+pub struct SecretsManager {
+    backend: Arc<dyn SecretsBackend>,
+}
+
+impl SecretsManager {
+    pub fn new(codex_home: PathBuf, backend_kind: SecretsBackendKind) -> Self {
+        let keyring_store: Arc<dyn KeyringStore> = Arc::new(DefaultKeyringStore);
```
xl-openai: Do we always need a keyring_store? For example if we use 1password we don't need any encryption right?

xl-openai (approved):

codex-rs/secrets/src/lib.rs
viyatb-oai: good catch, making it only for the local backend.

PR #10144, opened 2026-01-29
make codex better at git

pash-openai:
adds basic git context to the session prefix so the model can anchor git actions and be a bit more version-aware. structured it in a multiroot-friendly shape even though we only have one root today


pash-openai:

PR #10145, opened 2026-01-29
make codex better at git

pash-openai:
adds basic git context to the session prefix so the model can anchor git actions and be a bit more version-aware. structured it in a multiroot-friendly shape even though we only have one root today

### DIFF
--- a/codex-rs/core/src/client.rs
+++ b/codex-rs/core/src/client.rs
@@ -145,11 +145,7 @@ impl ModelClient {
         }
     }
 
-    pub fn new_session(
-        &self,
-        _turn_metadata_header: Option<String>,
-        turn_metadata_cwd: Option<PathBuf>,
-    ) -> ModelClientSession {
+    pub fn new_session(&self, turn_metadata_cwd: Option<PathBuf>) -> ModelClientSession {
         ModelClientSession {
             state: Arc::clone(&self.state),
             connection: None,
--- a/codex-rs/core/src/client.rs
+++ b/codex-rs/core/src/client.rs
@@ -147,18 +147,7 @@ impl ModelClient {
         }
     }
 
-    pub fn new_session(&self) -> ModelClientSession {
-        self.new_session_with_turn_metadata_and_cwd(None, None)
-    }
-
-    pub fn new_session_with_turn_metadata(
-        &self,
-        turn_metadata_header: Option<String>,
-    ) -> ModelClientSession {
-        self.new_session_with_turn_metadata_and_cwd(turn_metadata_header, None)
-    }
-
-    pub fn new_session_with_turn_metadata_and_cwd(
+    pub fn new_session(
         &self,
         turn_metadata_header: Option<String>,
         turn_metadata_cwd: Option<PathBuf>,
--- a/codex-rs/core/src/client.rs
+++ b/codex-rs/core/src/client.rs
@@ -1,3 +1,4 @@
+use std::path::PathBuf;
 use std::sync::Arc;
 use std::sync::OnceLock;
 
@@ -5,6 +6,7 @@ use crate::api_bridge::CoreAuthProvider;
 use crate::api_bridge::auth_provider_from_auth;
 use crate::api_bridge::map_api_error;
 use crate::auth::UnauthorizedRecovery;
+use crate::turn_metadata::build_turn_metadata_header;
 use codex_api::AggregateStreamExt;
 use codex_api::ChatClient as ApiChatClient;
 use codex_api::CompactClient as ApiCompactClient;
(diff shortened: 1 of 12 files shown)


codex-rs/core/src/environment_context.rs
```diff
@@ -39,14 +52,31 @@ impl EnvironmentContext {
         } else {
             None
         };
-        EnvironmentContext::new(cwd, shell.clone())
+        // Only include workspace configuration on the initial prefix message.
+        EnvironmentContext::new(cwd, shell.clone(), None)
```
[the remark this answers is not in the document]
pash-openai: good idea for a follow-up pr

codex-rs/core/src/environment_context.rs
```diff
@@ -66,6 +95,33 @@ impl EnvironmentContext {
...
+                        "      <latest_git_commit_hash>{latest_git_commit_hash}</latest_git_commit_hash>"
+                    ));
+                }
+
+                if let Some(associated_remote_urls) = workspace.associated_remote_urls
+                    && !associated_remote_urls.is_empty()
+                {
+                    lines.push("      <associated_remote_urls>".to_string());
+                    for (name, url) in associated_remote_urls {
+                        lines.push(format!("        <remote name=\"{name}\">{url}</remote>"));
```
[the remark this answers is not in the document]
pash-openai: same issue applies to `<cwd>`, so even if this is serialized it's still breakable - the model will figure it out just fine

codex-rs/core/src/client.rs
```diff
@@ -669,6 +687,9 @@ fn build_responses_headers(
     {
         headers.insert(X_CODEX_TURN_STATE_HEADER, header_value);
     }
+    if let Some(header_value) = turn_metadata_header {
+        headers.insert(X_CODEX_TURN_METADATA_HEADER, header_value.clone());
+    }
```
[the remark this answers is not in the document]
pash-openai: only responses matter.

codex-rs/core/src/git_info.rs
```diff
@@ -185,11 +272,19 @@ pub async fn git_diff_to_remote(cwd: &Path) -> Option<GitDiffToRemote> {
...
+
+/// Run a git command with a caller-provided timeout.
+async fn run_git_command_with_timeout_override(
+    args: &[&str],
+    cwd: &Path,
+    timeout_dur: TokioDuration,
+) -> Option<std::process::Output> {
+    let mut command = Command::new("git");
+    command.args(args).current_dir(cwd);
+    command.kill_on_drop(true);
```
pash-openai: added kill_on_drop as a general safeguard - it doesn’t look like any callers rely on git continuing after the future is dropped

codex-rs/core/src/client.rs
```diff
@@ -138,12 +145,31 @@ impl ModelClient {
     }
 
     pub fn new_session(&self) -> ModelClientSession {
```
aibrahim-oai: Why are we keeping 3 constructors? let's just have one.

codex-rs/core/src/client.rs
```diff
@@ -140,13 +147,21 @@ impl ModelClient {
...
+    ) -> ModelClientSession {
+        let turn_metadata_header =
+            turn_metadata_header.and_then(|value| HeaderValue::from_str(&value).ok());
         ModelClientSession {
             state: Arc::clone(&self.state),
             connection: None,
             websocket_last_items: Vec::new(),
             transport_manager: self.state.transport_manager.clone(),
             turn_state: Arc::new(OnceLock::new()),
+            turn_metadata_header,
```
aibrahim-oai: we can build the header on demand instead of storing it on the session

aibrahim-oai (commented):
Let's add an integration test to test the flow e2e

codex-rs/core/tests/responses_headers.rs
```diff
@@ -393,3 +395,191 @@ async fn responses_respects_model_info_overrides_from_config() {
...
+    let expected_origin = String::from_utf8(run_git(&["remote", "get-url", "origin"]).stdout)
+        .expect("git remote get-url output should be valid UTF-8")
+        .trim()
+        .to_string();
+
+    let repo_root = std::fs::canonicalize(cwd)
+        .unwrap_or_else(|_| cwd.to_path_buf())
+        .to_string_lossy()
+        .into_owned();
+    for _attempt in 0..10 {
```
pash-openai: this test is a bit involved (~200 lines) because i have to make multiple requests to verify that headers are eventually sent (because of the async prewarm)

codex-rs/core/src/client.rs
```diff
@@ -149,6 +162,42 @@ impl ModelClient {
             turn_state: Arc::new(OnceLock::new()),
         }
     }
+
+    /// Refresh turn metadata in the background and update a cached header that request
+    /// builders can read without blocking.
+    fn prewarm_turn_metadata_header(&self, turn_metadata_cwd: Option<PathBuf>) {
+        if self.state.provider.wire_api != WireApi::Responses {
```
aibrahim-oai: let's remove this CC is getting removed this week.

aibrahim-oai (approved):
LGTM, we have a lot of helper tests for integration tests that you can use to make the test shorter

pash-openai:
> LGTM, we have a lot of helper tests for integration tests that you can use to make the test shorter

For posterity, I tried using helper tests to make this test shorter, but in this case they obscured the runtime behavior and made the teest less reliable around the async prewarm timing. So I kept the test explicit on purpose so the behavior is clear and stable - i'll keep an eye on it for follow-ups

PR #10152, opened 2026-01-29
chore: unify log queries

jif-oai:
Unify log queries to only have SQLX code in the runtime and use it for both the log client and for tests


codex-rs/state/migrations/0002_logs.sql
gt-oai: does this not have to be a new migration now, where you modify the table?

gt-oai (approved):

codex-rs/state/migrations/0002_logs.sql
jif-oai: No one seeded the DB yet so not a big deal

PR #10154, opened 2026-01-29
feat: add output to `/ps`

jif-oai:
Screenshot 2026-01-29 at 13 24 57

### DIFF
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -1532,6 +1532,7 @@ impl ChatWidget {
         self.bottom_pane.set_unified_exec_processes(processes);
     }
 
+    /// Record recent stdout/stderr lines for the unified exec footer.
     fn track_unified_exec_output_chunk(&mut self, call_id: &str, chunk: &[u8]) {
         let Some(process) = self
             .unified_exec_processes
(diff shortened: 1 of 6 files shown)


codex-rs/tui/src/history_cell.rs
```diff
@@ -627,6 +634,32 @@ impl HistoryCell for UnifiedExecProcessesCell {
                 let (truncated, _, _) = take_prefix_by_width(&snippet, budget);
                 out.push(vec![prefix.dim(), truncated.cyan()].into());
             }
+
+            let chunk_prefix_first = "    ↳ ";
```
joshka-oai, line 638: I'd generally chop this into a small helper method, but not blocking.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -1524,6 +1532,31 @@ impl ChatWidget {
         self.bottom_pane.set_unified_exec_processes(processes);
     }
 
+    fn track_unified_exec_output_chunk(&mut self, call_id: &str, chunk: &[u8]) {
```
joshka-oai, line 1536: Could do with a small doc comment here

joshka-oai (approved):

PR #10159, opened 2026-01-29
Move auth and user-agent to separate crates

gt-oai:
1. Move core/auth into its own crate
2. Move core/default-client in its own crate (user-agent)
3. Re-export these values so I don't need to change all the imports right now
4. ...Except redirect the import in backend-client to the new location

This allows core to import backend-client (which has an auth dep), which I'll need to load requirements files.


gt-oai:
@codex review

gt-oai:
nah

Issue #2926, opened 2025-08-30
Allow customizing the status line


guidedways:
### What feature would you like to see?

I'd like to see the current working directory / git branch being worked on in the status line (the area below the entry field). Claude Code offers complete customization of this area, which makes it very easy to see what's important for the project at hand. Right now switching between terminals it's easy to forget what the working directory is.

hagan:
I second this, just knowing the current working directory/git branch details makes this so much better. Additionally, allowing users to customize in a similar manner to Claude would be an extremely useful addition to Codex. I wrote my own custom statusline for Claude in Rust to provide/track stats too, so I'm a little biased here.  [https://github.com/hagan/claudia-statusline](https://github.com/hagan/claudia-statusline). I would be willing to rework/rename and re-license to support Codex supporting this feature.

joshsymonds:
@etraut-openai is there interest in this feature and is it something OpenAI intends to produce itself? I am maintaining a fork of Codex to add the status line that I want but rebasing it constantly on upstream is proving very painful. Per [CONTRIBUTING.md](https://github.com/openai/codex/blob/main/docs/contributing.md) I'd be happy to contribute a more configurable status line/styling system if contributions like that would be accepted. I'd be happy to post what the API would look like first if that would be a good way to start this conversation, either here or in a new issue.

etraut-openai:
Thanks for the offer to help. I think we'd be interested in this, but it may be a little while before we can engage. The codex team has limited bandwidth, and we've received many hundreds of great feature requests from the community. We're trying to prioritize bug fixes and feature requests that have the highest impact. We use upvotes (reactions) to help inform priorities. I see that this feature has received some reactions, but there are many other features ahead of it in the queue. Thanks for your patience as we work through the list of feature requests and bug reports. It's gratifying to see the innovative ways that the community is leveraging and extending codex.

If you'd like to post your thoughts below about the shape that this feature could take, that would be most welcome!

joshsymonds:
Thanks @etraut-openai! This is kind of the end result I'm thinking of, which is what I have working right now:

<img width="1728" height="184" alt="Image" src="https://github.com/user-attachments/assets/dd607108-c1c4-4b4b-b59b-569ec1510b4b" />

Below is the shape I’ve been prototyping in my fork -- figured I’d capture it here so others can react. The inspiration is Starship, for which I have a prompt setup I really like that I want to replicate in my LLM tools.

  ### Goal

[...]

PR #10170, opened 2026-01-29
feat(tui): add custom status line support

fcoury:
# Summary

Add support for running a command or script and capturing the first output line to be used as the status line.

Commands receive a JSON payload via stdin with current session information (model, context window, workspace) and respond with the rendering of the information optionally use ANSI escape sequence for styling.

CleanShot 2026-01-29 at 12 41 53@2x

CleanShot 2026-01-29 at 13 09 16@2x

CleanShot 2026-01-29 at 13 14 50@2x

CleanShot 2026-01-29 at 13 10 48@2x

# Motivation

When using multiple different instances of Codex, it becomes hard to differentiate which instance refers to which project, or branch within a project when using worktrees or multiple clones.

This PR allows complete customization of the information displayed below the composer.

# Configuration

```toml
[tui]
status_line = ["my-status-script", "--arg"]
status_line_timeout_ms = 800 # optional, default 500ms
```

# Test configuration

Use [this gist here](https://gist.github.com/fcoury/274ca426c57a4828688eb6f59d875be4): 

```bash
cd ~/.codex
wget https://gist.githubusercontent.com/fcoury/274ca426c57a4828688eb6f59d875be4/raw/1b09c1b8f736bd8eaaba36f6124b6ee1421deaf7/status_line.py
chmod +x status_line.py
```

Add this to your config

```toml
[tui]
status_line = ["/Users/fcoury/.codex/status_line.py"]
```

**IMPORTANT:** Replace the prefix `/Users/fcoury` by your home folder, you can't use `~` here.

You can also pass in a new param to test other styles:

```toml
[tui]
status_line = ["/Users/fcoury/.codex/status_line.py", "powerline"]
```

You can use `codex` (default), `simple`, `powerline` or `powerline-basic`.

https://github.com/openai/codex/issues/2926
https://github.com/openai/codex/issues/10293


etraut-openai:
@codex review

codex-rs/tui/src/chatwidget.rs
```diff
@@ -6201,6 +6237,86 @@ async fn fetch_rate_limits(base_url: String, auth: CodexAuth) -> Option<RateLimi
     }
 }
 
+pub fn build_status_line_payload(chat: &ChatWidget, config: &Config) -> serde_json::Value {
```
etraut-openai: Please move this to its own status_line.rs module that's imported by chatwidget.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -6201,6 +6237,86 @@ async fn fetch_rate_limits(base_url: String, auth: CodexAuth) -> Option<RateLimi
...
+    let mut payload = serde_json::Map::new();
+
+    let cwd = chat.current_cwd.as_ref().unwrap_or(&config.cwd);
+
+    payload.insert(
```
etraut-openai: We tend to define schemas for everything in this source base. That makes it easier to maintain and test. It also allows us to auto-generate documentation, interfaces, SDKs, etc. Please look at the approach used elsewhere in the sources and copy that. The schema can be defined in the status_line.rs. Most of this logic should go away or become much simpler.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -6201,6 +6237,86 @@ async fn fetch_rate_limits(base_url: String, auth: CodexAuth) -> Option<RateLimi
...
+            .map(|m| m.display_name)
+            .unwrap_or_else(|| chat.current_model().to_string()),
+        "reasoning_level": chat.effective_reasoning_effort(),
+    });
+
+    payload.insert("model".to_string(), model);
+    payload.insert("version".to_string(), CODEX_CLI_VERSION.into());
+
+    let cost = serde_json::json!({
+        "total_lines_added": chat.total_lines_added,
```
etraut-openai: What about `total_duration_ms` and `total_api_duration_ms`? If we want to mirror CC's schema those could be added. Or I guess we can wait to see if we get feedback. `total_cost_usd` doesn't make sense for us because our usage model is different.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -6201,6 +6237,86 @@ async fn fetch_rate_limits(base_url: String, auth: CodexAuth) -> Option<RateLimi
...
+
+    let cwd = chat.current_cwd.as_ref().unwrap_or(&config.cwd);
+
+    payload.insert(
+        "session_id".to_string(),
+        chat.thread_id
+            .map(|id| id.to_string().into())
+            .unwrap_or(serde_json::Value::Null),
+    );
+    payload.insert(
```
etraut-openai: We should add a section for usage that is filled in when using an OpenAI ChatGPT subscription. It should include:
* 5h limit (percentage left and reset time/date)
* 7d limit (percentage left and reset time/date)
These are both provided in the status. And coincidentally, a [feature request](https://github.com/openai/codex/issues/10328) just arrived asking to expose these in the status bar.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -6201,6 +6237,86 @@ async fn fetch_rate_limits(base_url: String, auth: CodexAuth) -> Option<RateLimi
...
+        "total_lines_removed": chat.total_lines_removed,
+    });
+
+    payload.insert("cost".to_string(), cost);
+
+    let context_window_size = chat
+        .token_info
+        .as_ref()
+        .and_then(|ti| ti.model_context_window);
+    let remaining_percentage = context_window_size
```
etraut-openai, line 6367: When I test this manually, tThe used percentage and remaining percentage look wrong. They don't match what I see in `/status` (which looks correct).

Image

I recommend manually testing all exposed fields and exercising them to make sure they work as expected.

codex-rs/tui/src/status_line.rs
```diff
@@ -0,0 +1,335 @@
+use std::process::Stdio;
```
etraut-openai, line 16: Add a block comment at the top that explains what this module does.

codex-rs/tui/src/chatwidget.rs
etraut-openai: I think this should return a struct, not a JSON value. The caller should then be responsible for calling serde to serialize the struct into JSON.

We tend to define schemas for everything in this source base. That makes it easier to maintain and test. It also allows us to auto-generate documentation, interfaces, SDKs, etc. Please look at the approach used elsewhere in the sources and copy that. The schema can be defined in the status_line.rs. Serde will handle the JSON serialization for you based on the schema definition.

codex-rs/core/src/config/mod.rs
```diff
@@ -364,6 +367,9 @@ pub struct Config {
 
     /// OTEL configuration (exporter type, endpoint, headers, etc.).
     pub otel: crate::config::types::OtelConfig,
+
+    /// Status line command timeout duration.
+    pub tui_status_line_timeout_ms: Option<u64>,
```
etraut-openai: I wonder if we really need a configurable knob here. Couldn't we just pick a reasonable limit and hard-code it? Every new config knob needs to be documented and maintained, so there should be a good reason for every one of them. It looks like CC doesn't have such a knob. What do you think?

codex-rs/core/src/config/mod.rs
```diff
@@ -223,6 +223,9 @@ pub struct Config {
     /// - `never`: Never use alternate screen (inline mode, preserves scrollback).
     pub tui_alternate_screen: AltScreenMode,
 
+    /// Optional command (argv form) to generate a custom TUI status line.
+    pub tui_status_line: Option<Vec<String>>,
```
etraut-openai, line 227: It's interesting that CC uses a struct here that includes a "type" field. It looks like they support only "command" right now. I wonder if they are considering adding other options in the future. I guess it's good for us to keep it simple and support only a command for now.

codex-rs/tui/src/status_line.rs
```diff
@@ -0,0 +1,335 @@
...
+        let mut state = self
+            .state
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        state.latest_payload = Some(payload);
+        Ok(())
+    }
+
+    pub(crate) fn request_update(&self) -> anyhow::Result<()> {
+        let Some(command) = self
```
etraut-openai, line 57: It looks like we don't have any throttle on the frequency of calls. I wonder if we should add one. The CC documentation says that their command is called at most once every 300ms. My concern is that this could become a performance issue under some situations. What do you think?

etraut-openai (changes requested):
Nice! This is looking good. I left some comments and questions.

FYI, I'm personally not an expert on the TUI code, so I may ask one of the other Codex team members to do a second-pass review.

codex-rs/tui/src/chatwidget.rs
fcoury: @etraut-openai they could be added, I just didn't add because we didn't have them right off the bat. Agreed with `total_cost_usd`, that's why I left it out.

codex-rs/tui/src/chatwidget.rs
fcoury: > We should add a section for usage that is filled in when using an OpenAI ChatGPT subscription. It should include:

This is a great idea, I will add them.

codex-rs/tui/src/chatwidget.rs
fcoury: > I recommend manually testing all exposed fields and exercising them to make sure they work as expected.

I did, but clearly not properly. Sorry about that, I will re-test everything.

codex-rs/tui/src/chatwidget.rs
fcoury: > I think this should return a struct, not a JSON value. The caller should then be responsible for calling serde to serialize the struct into JSON.

Yes, I totally agree. I'll change it.

codex-rs/core/src/config/mod.rs
fcoury: > I wonder if we really need a configurable knob here. Couldn't we just pick a reasonable limit and hard-code it? Every new config knob needs to be documented and maintained, so there should be a good reason for every one of them. It looks like CC doesn't have such a knob. What do you think?

The reason I added was because ccstatusline was failing on a cold boot and I thought it would be good to have it. But since I bumped it to 500ms it didn't trigger again. I think CC uses 300ms, do you want me to keep it 300ms?

codex-rs/core/src/config/mod.rs
fcoury: > It's interesting that CC uses a struct here that includes a "type" field. It looks like they support only "command" right now. I wonder if they are considering adding other options in the future. I guess it's good for us to keep it simple and support only a command for now.

Yes that was a deliberate decision on my end. They have the concept of "hooks" and I think they modeled it as another hook. I modeled our config after the `notification` setting we already have.

codex-rs/tui/src/status_line.rs
fcoury: > It looks like we don't have any throttle on the frequency of calls. I wonder if we should add one. The CC documentation says that their command is called at most once every 300ms. My concern is that this could become a performance issue under some situations. What do you think?

I think we should add it, I'll do that!

codex-rs/core/src/config/mod.rs
etraut-openai: I think you mean `notify` rather than `notification`. Just fyi, our `notify` setting will be deprecated soon in favor of hooks. We've received a lot of requests for a full hooks mechanism.

codex-rs/tui/src/chatwidget.rs
fcoury: @etraut-openai fixed.

CleanShot 2026-02-02 at 16 12 02@2x

codex-rs/tui/src/chatwidget.rs
fcoury: @etraut-openai added: 
CleanShot 2026-02-02 at 16 12 02@2x

codex-rs/tui/src/chatwidget.rs
fcoury: @etraut-openai I decided to wait for the feedback for this one, let me know otherwise.

codex-rs/core/src/config/mod.rs
fcoury: @etraut-openai I did mean `notify`, sorry about that. I was aware of the hooks initiative, even though I am not sure what my source was, maybe twitter? 😊

In any case, I think we could ship it this way and think about a refactor to add this as a new hook if we find it necessary after the deprecation. 

What do you think?

PR #9636, opened 2026-01-21
Restore pending large-paste payloads on Ctrl+C history recall

charley-oai:
## Summary
- Persist pending large-paste payloads in in-memory ↑/↓ history entries and rehydrate placeholder elements on recall.
- Restore large-paste placeholders with element styling and expansion behavior after Ctrl+C draft clears.
- Document history navigation behavior in the chat composer doc.

Addresses https://github.com/openai/codex/issues/9050

PR #9040, opened 2026-01-10
fix(tui): rehydrate drafts and restore image placeholders

Chriss4123:
Fixes #9050

When a draft is stashed with Ctrl+C, we now persist the full draft state (text elements, local image paths, and pending paste payloads) in local history. Up/Down recall rehydrates placeholder elements and attachments so styling remains correct and large pastes still expand on submit. Persistent (cross‑session) history remains text‑only.

Backtrack prefills now reuse the selected user message’s text elements and local image paths, so image placeholders/attachments rehydrate when rolling back.

External editor replacements keep only attachments whose placeholders remain and then normalize image placeholders to `[Image #1]..[Image #N]` to keep the attachment mapping consistent.

Docs:
- docs/tui-chat-composer.md

Testing:
- just fix -p codex-tui
- cargo test -p codex-tui

PR #10178, opened 2026-01-29
clearing paste counters to reset on new user message

natea-oai:
When pasting text with the same number of characters we never decrement the suffix used. This updates the suffix to start at 0 for each new user message

Before:
pasted values 1

After:
pasted values 2

### DIFF
--- a/codex-rs/tui/src/bottom_pane/chat_composer.rs
+++ b/codex-rs/tui/src/bottom_pane/chat_composer.rs
@@ -6113,6 +6113,50 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


joshka-oai:
Over in https://github.com/openai/codex/pull/9636 and possibly https://github.com/openai/codex/pull/9040 (which is on my backlog to get back to on review), we're going to rehydrate some things that may mean numbers come back oddly. I suspect this may cause issues in the following sequence because the numbers will overlap:
1. paste, then queue
2. paste, then queue
3. interrupt
4. up arrow to edit
(I'm not 100% sure though).

natea-oai:
ah got it, i'll close for now then

PR #10189, opened 2026-01-29
implement per-workspace capability SIDs for workspace specific ACLs

iceweasel-oai:
Today, there is a single capability SID that allows the sandbox to write to
* workspace (cwd)
* tmp directories if enabled
* additional writable roots

This change splits those up, so that each workspace has its own capability SID, while tmp and additional roots, which are installation-wide, are still governed by the "generic" capability SID

This isolates workspaces from each other in terms of sandbox write access.
Also allows us to protect <cwd>/.codex when codex runs in a specific <cwd>


iceweasel-oai:
@codex review

codex-rs/windows-sandbox-rs/src/setup_main_win.rs
```diff
@@ -628,7 +638,21 @@ fn run_setup_full(payload: &Payload, log: &mut File, sbx_dir: &Path) -> Result<(
...
+        let cap_label = if is_command_cwd {
+            "workspace_cap"
+        } else {
+            "cap"
+        };
+        let cap_psid_for_root = if is_command_cwd {
+            workspace_psid
+        } else {
+            cap_psid
+        };
```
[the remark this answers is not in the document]
iceweasel-oai: I'm aware of this and willing to punt for now for simplicity. The sandbox is still experimental and this is more of an inconvenience (that is not likely to be hit) than a security flaw

codex-rs/windows-sandbox-rs/src/cap.rs
```diff
@@ -60,7 +68,30 @@ pub fn load_or_create_cap_sids(codex_home: &Path) -> Result<CapSids> {
...
     Ok(caps)
 }
+
+fn canonical_cwd_key(cwd: &Path) -> String {
+    let canonical = dunce::canonicalize(cwd).unwrap_or_else(|_| cwd.to_path_buf());
+    canonical
+        .to_string_lossy()
+        .replace('\\', "/")
+        .to_ascii_lowercase()
+}
```
dylan-hurd-oai: This feels like something we should encapsulate and reuse across windows logic? Similar to `AbsolutePathBuf`

dylan-hurd-oai (approved):
1 non-blocking suggestion

codex-rs/windows-sandbox-rs/src/cap.rs
iceweasel-oai: agreed - I consolidated this from a few places and added a few more tests.

PR #10190, opened 2026-01-30
Load exec policy rules from requirements

gt-oai:
`requirements.toml` should be able to specify rules which always run. 

My intention here was that these rules could only ever be restrictive, which means the decision can be "prompt" or "forbidden" but never "allow". A requirement of "you must always allow this command" didn't make sense to me, but happy to be gaveled otherwise.

Rules already applies the most restrictive decision, so we can safely merge these with rules found in other config folders.

### DIFF
--- a/codex-rs/core/src/config_loader/config_requirements.rs
+++ b/codex-rs/core/src/config_loader/config_requirements.rs
@@ -6,6 +6,8 @@ use serde::Deserialize;
 use std::collections::BTreeMap;
 use std::fmt;
 
+use super::requirements_exec_policy::RequirementsExecPolicy;
+use super::requirements_exec_policy::RequirementsExecPolicyToml;
 use crate::config::Constrained;
 use crate::config::ConstraintError;
 
@@ -49,6 +51,7 @@ pub struct ConfigRequirements {
     pub approval_policy: Constrained<AskForApproval>,
     pub sandbox_policy: Constrained<SandboxPolicy>,
     pub mcp_servers: Option<Sourced<BTreeMap<String, McpServerRequirement>>>,
+    pub(crate) exec_policy: Option<Sourced<RequirementsExecPolicy>>,
 }
 
 impl Default for ConfigRequirements {
@@ -57,6 +60,7 @@ impl Default for ConfigRequirements {
             approval_policy: Constrained::allow_any_from_default(),
             sandbox_policy: Constrained::allow_any(SandboxPolicy::ReadOnly),
             mcp_servers: None,
+            exec_policy: None,
         }
     }
 }
@@ -79,6 +83,7 @@ pub struct ConfigRequirementsToml {
     pub allowed_approval_policies: Option<Vec<AskForApproval>>,
     pub allowed_sandbox_modes: Option<Vec<SandboxModeRequirement>>,
     pub mcp_servers: Option<BTreeMap<String, McpServerRequirement>>,
+    pub rules: Option<RequirementsExecPolicyToml>,
 }
 
 /// Value paired with the requirement source it came from, for better error
@@ -108,6 +113,7 @@ pub struct ConfigRequirementsWithSources {
     pub allowed_approval_policies: Option<Sourced<Vec<AskForApproval>>>,
     pub allowed_sandbox_modes: Option<Sourced<Vec<SandboxModeRequirement>>>,
     pub mcp_servers: Option<Sourced<BTreeMap<String, McpServerRequirement>>>,
+    pub rules: Option<Sourced<RequirementsExecPolicyToml>>,
 }
 
 impl ConfigRequirementsWithSources {
@@ -139,6 +145,7 @@ impl ConfigRequirementsWithSources {
                 allowed_approval_policies,
                 allowed_sandbox_modes,
                 mcp_servers,
+                rules,
             }
         );
     }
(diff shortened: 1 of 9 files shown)


codex-rs/core/src/config_loader/config_requirements.rs
```diff
@@ -282,23 +307,36 @@ impl TryFrom<ConfigRequirementsWithSources> for ConfigRequirements {
...
     use codex_protocol::protocol::NetworkAccess;
     use codex_utils_absolute_path::AbsolutePathBuf;
     use pretty_assertions::assert_eq;
     use toml::from_str;
 
+    fn tokens(cmd: &[&str]) -> Vec<String> {
+        cmd.iter().map(std::string::ToString::to_string).collect()
+    }
+
+    fn allow_all(_: &[String]) -> Decision {
```
bolinfest: why the arg?

codex-rs/core/src/config_loader/config_requirements.rs
```diff
@@ -595,4 +637,67 @@ mod tests {
         );
         Ok(())
     }
+
+    #[test]
+    fn deserialize_exec_policy_requirements() -> Result<()> {
+        let toml_str = r#"
+            [exec_policy]
```
bolinfest: We should not use `exec_policy` in any API since we have moved to "rules" as the public-facing name?

Though honestly, it may soon become `permissions`...

codex-rs/core/src/config_loader/config_requirements.rs
```diff
@@ -595,4 +637,67 @@ mod tests {
...
+                { pattern = [{ token = "rm" }] },
+            ]
+        "#;
+        let config: ConfigRequirementsToml = from_str(toml_str)?;
+
+        let requirements_toml_file = if cfg!(windows) {
+            "C:\\etc\\codex\\requirements.toml"
+        } else {
+            "/etc/codex/requirements.toml"
+        };
```
bolinfest: Should this be in a helper function or something?

codex-rs/core/src/config_loader/requirements_exec_policy.rs
```diff
@@ -82,6 +109,12 @@ pub enum RequirementsExecPolicyParseError {
 
     #[error("exec policy prefix_rule at index {rule_index} has an empty justification")]
     EmptyJustification { rule_index: usize },
+
+    #[error("exec policy prefix_rule at index {rule_index} is missing a decision")]
```
bolinfest: "rules"

codex-rs/core/src/config_loader/tests.rs
```diff
@@ -1072,4 +1097,139 @@ prefix_rules = [
...
+    #[test]
+    fn allow_decision_is_rejected() -> anyhow::Result<()> {
+        let toml_str = r#"
+prefix_rules = [
+    { pattern = [{ token = "rm" }], decision = "allow" },
+]
+"#;
+
+        let parsed: RequirementsExecPolicyToml = from_str(toml_str)?;
+        let err = parsed.to_policy().expect_err("allow decision not allowed");
```
bolinfest, line 1233: Why isn't the provider of `requirements.toml` allowed to declare a rule with `decision = "allow"`?

codex-rs/core/src/config_loader/tests.rs
```diff
@@ -1072,4 +1097,139 @@ prefix_rules = [
...
+            err,
+            RequirementsExecPolicyParseError::AllowDecisionNotAllowed { rule_index: 0 }
+        ));
+        Ok(())
+    }
+
+    #[test]
+    fn empty_prefix_rules_is_rejected() -> anyhow::Result<()> {
+        let toml_str = r#"
+prefix_rules = []
```
bolinfest, line 1245: Couldn't this mean that an end-user is not allowed to define any rules?

codex-rs/core/src/exec_policy.rs
```diff
@@ -280,7 +280,18 @@ pub async fn load_exec_policy(config_stack: &ConfigLayerStack) -> Result<Policy,
     let policy = parser.build();
     tracing::debug!("loaded rules from {} files", policy_paths.len());
 
-    Ok(policy)
+    let Some(requirements_policy) = config_stack.requirements().exec_policy.as_deref() else {
+        return Ok(policy);
+    };
+
+    let mut combined_rules = policy.rules().clone();
```
bolinfest, line 287: I don't think we should combine rules in this case, should we? Shouldn't `requirements_policy` be the only rules you can set?

bolinfest (commented):
Please fill out the PR body.

jif-oai:
@codex review

codex-rs/core/src/exec_policy.rs
gt-oai: IMO users (and repo config) should be allowed to set rules that are more restrictive than the ones in `requirements_policy`. Rules already apply the most restrictive matching decision, so additional rules feels harmless to me.

I cannot think of a genuine use for "admins can require that command X is always allowed, regardless of what the user wants". Perhaps it would be good to e.g. always allow `bazel build` or something, but that feels more in the camp of sensible default that should be included in the repo, rather than a strict requirement that prevents a user's other rules from loading.

codex-rs/core/src/config_loader/config_requirements.rs
gt-oai: Wasn't material to the test

codex-rs/core/src/config_loader/config_requirements.rs
gt-oai: Inlined this

codex-rs/core/src/config_loader/tests.rs
gt-oai: Clarified the error message as to why this is the case

bolinfest (approved):

PR #10192, opened 2026-01-30
feat: migrate TUI to use app-server v2

bolinfest:
I asked Codex to update the TUI to speak the app server protocol instead of the harness's internal protocol and this is what it came up with.

In particular, `codex-rs/app-server/src/lib.rs` introduces `spawn_in_memory_typed()`, which returns:

```rust
pub struct InProcessAppServer {
    pub incoming: mpsc::Sender<AppServerClientMessage>,
    pub outgoing: mpsc::Receiver<AppServerMessage>,
}
```

This represents the bidirectional streams of the app server, though these new `AppServerClientMessage` and `AppServerClientMessage` types avoid the need to serialize and deserialize JSON messages since, at the moment, the app server is still running in the same process as the TUI.

This PR adds a number of new message types to `codex-rs/app-server-protocol/src/protocol/{common,v2}.rs` as well as new handlers in `codex-rs/app-server/src/codex_message_processor.rs` to handle the `Op`s used by TUI that were not previously represented in app server. New methods:

- `thread/compact`
- `thread/shutdown`
- `mcp/elicitation/resolve`


nazreen:
will this still happen?

abrahem79:
`pub struct InProcessAppServer {
    pub incoming: mpsc::Sender<AppServerClientMessage>,
    pub outgoing: mpsc::Receiver<AppServerMessage>,
}`

PR #10197, opened 2026-01-30
Add collaboration mode instructions per model

aibrahim-oai:
1. add it as a property of modelinfo


aibrahim-oai:
@codex review to make sure this is backward comp. and also will survive deleting the field from the /models endpoint.

PR #10207, opened 2026-01-30
Skip loading codex home as project layer

daniel-oai:
Summary:
- Fixes issue #9932: https://github.com/openai/codex/issues/9932
- Prevents `$CODEX_HOME` (typically `~/.codex`) from being discovered as a project `.codex` layer by skipping it during project layer traversal. We compare both normalized absolute paths and best-effort canonicalized paths to handle symlinks.
- Adds regression tests for home-directory invocation and for the case where `CODEX_HOME` points to a project `.codex` directory (e.g., worktrees/editor integrations).

Testing:
- `cargo build -p codex-rmcp-client --bin test_stdio_server`
- `cargo test -p codex-core`
- `cargo test --all-features`
- Manual: ran `target/debug/codex` from `~` and confirmed the disabled-folder warning and trust prompt no longer appear.


daniel-oai:
@codex please review

codex-rs/core/src/config_loader/mod.rs
```diff
@@ -692,6 +698,11 @@ async fn load_project_layers(
         let layer_dir = AbsolutePathBuf::from_absolute_path(dir)?;
         let decision = trust_context.decision_for_dir(&layer_dir);
         let dot_codex_abs = AbsolutePathBuf::from_absolute_path(&dot_codex)?;
+        let dot_codex_normalized =
+            normalize_path(dot_codex_abs.as_path()).unwrap_or_else(|_| dot_codex_abs.to_path_buf());
+        if dot_codex_abs == codex_home_abs || dot_codex_normalized == codex_home_normalized {
```
xl-openai, line 703: Any reason you need both here?

daniel-oai: The first is a cheap, reliable equality check. The second handles symlinked or differently-canonicalized paths (which can happen with worktrees, temp dirs, or custom $CODEX_HOME).

gt-oai: this feels unnecessary to me but also don't think it's actively harmful

gt-oai (approved):
thanks!

xl-openai (approved):

codex-rs/core/src/config_loader/mod.rs
daniel-oai: Yeah its an uncommon situation but certainly does happen

PR #10208, opened 2026-01-30
feat: refactor CodexAuth so invalid state cannot be represented

bolinfest:
Previously, `CodexAuth` was defined as follows:

https://github.com/openai/codex/blob/d550fbf41afc09d7d7b5ac813aea38de07b2a73f/codex-rs/core/src/auth.rs#L39-L46

But if you looked at its constructors, we had creation for `AuthMode::ApiKey` where `storage` was built using a nonsensical path (`PathBuf::new()`) and `auth_dot_json` was `None`:

https://github.com/openai/codex/blob/d550fbf41afc09d7d7b5ac813aea38de07b2a73f/codex-rs/core/src/auth.rs#L212-L220

By comparison, when `AuthMode::ChatGPT` was used, `api_key` was always `None`:

https://github.com/openai/codex/blob/d550fbf41afc09d7d7b5ac813aea38de07b2a73f/codex-rs/core/src/auth.rs#L665-L671

https://github.com/openai/codex/pull/10012 took things further because it introduced a new `ChatgptAuthTokens` variant to `AuthMode`, which is important in when invoking `account/login/start` via the app server, but most logic _internal_ to the app server should just reason about two `AuthMode` variants: `ApiKey` and `ChatGPT`.

This PR tries to clean things up as follows:

- `LoginAccountParams` and `AuthMode` in `codex-rs/app-server-protocol/` both continue to have the `ChatgptAuthTokens` variant, though it is used exclusively for the on-the-wire messaging.
- `codex-rs/core/src/auth.rs` now has its own `AuthMode` enum, which only has two variants: `ApiKey` and `ChatGPT`.
- `CodexAuth` has been changed from a struct to an enum. It is a disjoint union where each variant (`ApiKey`, `ChatGpt`, and `ChatGptAuthTokens`) have only the associated fields that make sense for that variant.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/10208).
* #10224
* __->__ #10208

### DIFF
--- a/codex-rs/core/src/auth.rs
+++ b/codex-rs/core/src/auth.rs
@@ -13,8 +13,9 @@ use std::path::Path;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::sync::Mutex;
+use std::sync::RwLock;
 
-use codex_app_server_protocol::AuthMode;
+use codex_app_server_protocol::AuthMode as ApiAuthMode;
 use codex_protocol::config_types::ForcedLoginMethod;
 
 pub use crate::auth::storage::AuthCredentialsStoreMode;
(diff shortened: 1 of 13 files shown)


codex-rs/core/src/auth.rs
```diff
@@ -35,19 +36,50 @@ use codex_protocol::account::PlanType as AccountPlanType;
...
+    ApiKey,
+    ChatGPT,
+}
+
+/// Authentication mechanism used by the current user.
+#[derive(Debug, Clone)]
+pub enum CodexAuth {
+    ApiKey(ApiKeyAuth),
+    ChatGpt(ChatGptAuth),
+    ChatGptAuthTokens(ChatGptAuthTokens),
```
owenlin0, line 55: totally a nit: I'd prefer `Chatgpt` / `ChatgptAuthTokens` (lowercase g)

owenlin0 (approved):

codex-rs/core/src/auth.rs
bolinfest: I will do it for consistency with the surrounding code, but I think we should be consistent throughout.

Based on https://rust-lang.github.io/api-guidelines/naming.html and looking at other things in the standard library (like `std::net::TcpStream`), using `ChatGpt` seems like the correct casing.

bolinfest: Discussed offline: will stick with `ChatGpt`.

PR #10214, opened 2026-01-30
feat(app-server) FindFilesStream

dylan-hurd-oai:
## Summary
We recently upgraded file search in #9939 - let's see if we can take advantage of it in the app-server

## Testing
- [x] integration tests


gpeal (approved):

gpeal (approved):
If you could add an experimental flag to this, that would be great

PR #10234, opened 2026-01-30
Make cloud requirements load fail-closed

gt-oai:
Prevent Codex from starting if we cannot load Cloud requirements.

We should merge this once we've built confidence in the cloud requirement loading path.


gverma-openai:
Note: do not merge until team decides it is safe to do so.

gt-oai:
Maybe 404 should fail open, not closed. Would mean users are not feature-flagged into this.

PR #10240, opened 2026-01-30
file-search: multi-root walk

nornagon-openai:
Instead of a separate walker for each root in a multi-root walk, use a single walker.


codex-rs/app-server/src/fuzzy_file_search.rs
```diff
@@ -23,64 +21,53 @@ pub(crate) async fn run_fuzzy_file_search(
...
+    let mut files = match tokio::task::spawn_blocking(move || {
+        file_search::run(
+            query.as_str(),
+            limit,
+            search_dirs,
+            Vec::new(),
+            threads,
+            cancellation_flag,
+            COMPUTE_INDICES,
+            true,
```
joshka-oai: side note: should be an enum

joshka-oai (approved):
LGTM - Most of this could do with some better docs

codex-rs/app-server/src/fuzzy_file_search.rs
nornagon-openai: replaced this with the options struct.

joshka-oai: LGTM

PR #10246, opened 2026-01-30
adding fork information (UI) when forking

pap-openai:
- shows `/fork` command that ran in prev session
- shows `session forked from name? || uuid` as an event in new session

image
image


joshka-oai (approved):
Please add insta snapshot tests for this

Consider adding a bit more about the session ids to the message after fork, and whether we use session / thread as the name of this.

PR #10247, opened 2026-01-30
Fix unsafe auto-approval of git branch deletion

etraut-openai:
Diagnosis: Although the bug report specifically mentioned worktrees, this problm isn't worktree-specific. `git branch -d/-D/--delete` isn’t treated as dangerous, and git branch is currently treated as “safe,” so it won’t prompt under on-request. In a worktree, that’s extra risky because branch deletion writes to the common gitdir outside the workdir.

* Classify git branch -d/-D/--delete as dangerous to force approval under on-request.
* Restrict git branch “safe” classification to read-only forms (e.g., --show-current, --list).
* Add focused safety tests for branch deletion and read-only branch queries.

This addresses #10160


etraut-openai:
@codex review

Issue #9222, opened 2026-01-14
os error 2 after Windows Update


carldebilly:
### What version of Codex is running?

codex-cli 0.81.0

### What subscription do you have?

ChatGPT PLUS

### Which model were you using?

gpt-5.2-codex

### What platform is your computer?

Windows 11 PRO 64 bits

### What issue are you seeing?

After the reboot to apply Windows Security Update KB5074109, Codex is unable to start, producing this error message:

```
PS C:\Users\Me> codex
Error: The system cannot find the file specified. (os error 2)
```

I checked the following:
- Nothing blocked on Windows Defender
- I reinstalled MSVC librairies (repair)
- I tried other versions of Codex - 0.78.0, 0.79.0, 0.80.0 -- those who were working until this morning
- I rebooted plenty of times
- I reinstalled npm to latest version (in case)
- I confirm the `C:\Users\Me\AppData\Roaming\npm\node_modules\@openai\codex\vendor\x86_64-pc-windows-msvc\codex\codex.exe` exists and a direct execution produces the same result (os error 2)

### What steps can reproduce the bug?

I don't know what to check

### Additional information

Windows Security Update: `KB5074109 (26200.7623)`
`.NET 9.0.12` / `8.0.23` updates installed same day

etraut-openai:
Sounds like a path configuration issue on your system. You might want to ask ChatGPT to suggest potential fixes.

carldebilly:
> Sounds like a path configuration issue on your system. You might want to ask ChatGPT to suggest potential fixes.

### Windows-only failure after Windows Update (codex.exe)

After a Windows Update, `codex` fails to start on one machine, while the **same Codex version** works on other Windows machines.

Node **does start correctly** and executes `codex.js`.  
The failure occurs when `codex.js` spawns the native binary **`codex.exe`**.

The user-visible error is always:

```text
Error: The system cannot find the file specified. (os error 2)
```

However, `codex.exe` **does exist** and is launched.

---

### Key observations (Process Monitor)

Using Sysinternals Process Monitor, filtering on `codex.exe`:

- `codex.exe` attempts to load app-local DLLs from:
  ```
  vendor\x86_64-pc-windows-msvc\
  ```
  (e.g. `codexVCRUNTIME140.dll`, `codexbcrypt.dll`, `codexuserenv.dll`)
- These DLLs are not present → `NAME NOT FOUND`
- Windows then correctly falls back to:
  ```
[...]

PR #10249, opened 2026-01-30
Validate CODEX_HOME before resolving

etraut-openai:
Summary
- require `CODEX_HOME` to point to an existing directory before canonicalizing and surface clear errors otherwise
- share the same helper logic in both `core` and `rmcp-client` and add unit tests that cover missing, non-directory, valid, and default paths

This addresses #9222


codex-rs/core/src/config/mod.rs
```diff
@@ -1749,17 +1749,41 @@ fn toml_uses_deprecated_instructions_file(value: &TomlValue) -> bool {
...
+}
+
+fn find_codex_home_from_env(codex_home_env: Option<&str>) -> std::io::Result<PathBuf> {
     // Honor the `CODEX_HOME` environment variable when it is set to allow users
     // (and tests) to override the default location.
-    if let Ok(val) = std::env::var("CODEX_HOME")
-        && !val.is_empty()
-    {
-        return PathBuf::from(val).canonicalize();
+    if let Some(val) = codex_home_env {
```
bolinfest: Could you turn this whole thing into a `match` and remove the `return` statements so it's easier to verify that all of these codepaths are mutually exclusive?

codex-rs/core/src/config/mod.rs
```diff
@@ -1749,17 +1749,41 @@ fn toml_uses_deprecated_instructions_file(value: &TomlValue) -> bool {
...
+        let path = PathBuf::from(val);
+        if !path.exists() {
```
bolinfest: Prefer `std::fs::metadata()` so that you do one fs check?

codex-rs/utils/home-dir/src/lib.rs
```diff
@@ -9,17 +9,41 @@ use std::path::PathBuf;
...
 /// - If `CODEX_HOME` is not set, this function does not verify that the
 ///   directory exists.
 pub(crate) fn find_codex_home() -> std::io::Result<PathBuf> {
+    let codex_home_env = std::env::var("CODEX_HOME")
+        .ok()
+        .filter(|val| !val.is_empty());
+    find_codex_home_from_env(codex_home_env.as_deref())
+}
+
+fn find_codex_home_from_env(codex_home_env: Option<&str>) -> std::io::Result<PathBuf> {
```
bolinfest, line 19: Can we eliminate the duplication?

bolinfest (approved):
Very nice: ty!

PR #10250, opened 2026-01-30
Fix resume picker bug where it shows "no message yet"

etraut-openai:
Fix "the resume-picker list sometimes incorrectly shows "(no message yet)" under conversation"

I'm not an expert on the events that comprise a rollout, so I'm not 100% confident this is the correct fix. It was recommended by Codex.

Summary
- ensure the head preview always includes at least one user message even if it appears beyond the initial record limit, by tracking the first user response seen while scanning
- reuse that delayed user message if the clipped preview lacks any user turn, so session summaries reply faster with meaningful context
- add regression coverage that writes a rollout file where the first user response occurs past the head limit and verify it shows up in the preview

Testing
I don't have a repro case for this problem, so I didn't manually test it. If we're concerned about the fix, I could ask the user to provide a rollout that repros the problem. Otherwise I'll just have the user confirm the fix once we roll out the next alpha.

This addresses #10216


etraut-openai:
@codex review

PR #10252, opened 2026-01-30
[feat] persist thread_dynamic_tools in db

celia-oai:
Persist thread_dynamic_tools in sqlite and read first from it. Fall back to rollout files if it's not found. Persist dynamic tools to both sqlite and rollout files.

Saw that new sessions get populated to db correctly & old sessions get backfilled correctly at startup:
```
celia@com-92114 codex-rs % sqlite3 ~/.codex/state.sqlite \      "select thread_id, position,name,description,input_schema from thread_dynamic_tools;"
019c0cad-ec0d-74b2-a787-e8b33a349117|0|geo_lookup|lookup a city|{"properties":{"city":{"type":"string"}},"required":["city"],"type":"object"}
....
019c10ca-aa4b-7620-ae40-c0919fbd7ea7|0|geo_lookup|lookup a city|{"properties":{"city":{"type":"string"}},"required":["city"],"type":"object"}
```

### DIFF
--- a/codex-rs/core/src/rollout/metadata.rs
+++ b/codex-rs/core/src/rollout/metadata.rs
@@ -189,12 +189,19 @@ pub(crate) async fn backfill_sessions(
                     stats.upserted = stats.upserted.saturating_add(1);
                     if let Ok(meta_line) = rollout::list::read_session_meta_line(&path).await {
                         if let Err(err) = runtime
-                            .replace_dynamic_tools(
+                            .persist_dynamic_tools(
                                 meta_line.meta.id,
                                 meta_line.meta.dynamic_tools.as_deref(),
                             )
                             .await
                         {
+                            if let Some(otel) = otel {
+                                otel.counter(
+                                    DB_ERROR_METRIC,
+                                    1,
+                                    &[("stage", "backfill_dynamic_tools")],
+                                );
+                            }
                             warn!("failed to backfill dynamic tools {}: {err}", path.display());
                         }
                     } else {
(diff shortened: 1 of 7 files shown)


codex-rs/state/migrations/0004_thread_dynamic_tools.sql
```diff
@@ -0,0 +1,11 @@
...
+    thread_id TEXT NOT NULL,
+    position INTEGER NOT NULL,
+    name TEXT NOT NULL,
+    description TEXT NOT NULL,
+    input_schema TEXT NOT NULL,
+    PRIMARY KEY(thread_id, position),
+    FOREIGN KEY(thread_id) REFERENCES threads(id) ON DELETE CASCADE
+);
+
+CREATE INDEX idx_thread_dynamic_tools_thread ON thread_dynamic_tools(thread_id);
```
owenlin0, line 11: love it, it's like i'm looking at backend code again

celia-oai: resolving all the tricky db issues with migrations and foreign keys etc. do bring a throwback

codex-rs/core/src/rollout/metadata.rs
```diff
@@ -187,6 +187,22 @@ pub(crate) async fn backfill_sessions(
...
                     stats.upserted = stats.upserted.saturating_add(1);
+                    if let Ok(meta_line) = rollout::list::read_session_meta_line(&path).await {
+                        if let Err(err) = runtime
+                            .replace_dynamic_tools(
+                                meta_line.meta.id,
+                                meta_line.meta.dynamic_tools.as_deref(),
+                            )
+                            .await
+                        {
+                            warn!("failed to backfill dynamic tools {}: {err}", path.display());
```
jif-oai, line 205: Can you add a metric counter for this?

jif-oai (approved):
I'm not sure the replace semantic makes sense as we don't want people to replace them after thread start but good for me

celia-oai:
> I'm not sure the replace semantic makes sense as we don't want people to replace them after thread start but good for me

will change the name to persist_dynamic_tools and gate it to only write tools when there's no dynamic tools for a thread

PR #10254, opened 2026-01-30
Fix backtrack to restore collaboration mode and keep developer instructions in sync

charley-oai:
## Summary
- Syncs collaboration mode correctly when backtracking/rolling back, including custom mode cases.
- Stores turn context history alongside user turns so rollback restores the right mode (or forces a re‑emit when missing).
- Ensures rollback failures revert the optimistic UI mode switch.
- Preserves the one‑shot collaboration instruction re‑emit through unrelated overrides.
- Adds focused tests for backtrack/rollback mode behavior.

## Motivation
Backtracking could leave the client in the wrong collaboration mode (or skip re‑emitting developer instructions) when rolling back to turns without
stored context or when a rollback failed. This keeps the UI and core in sync and avoids stale mode instructions.

## Key Changes
- Core: maintain turn context history per user turn and apply it during rollback.
- Core: force collaboration instruction injection when the latest turn lacks context; preserve the flag until consumed.
- TUI: apply and restore collaboration masks correctly during backtrack/rollback.
- Tests: cover plan/code/custom backtrack flows and stale/missing turn context rollback.

## Codex Author
`codex fork 019c1047-d5d5-7c92-a357-6009604dc7e8`


charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

codex-rs/core/src/state/session.rs
```diff
@@ -15,6 +16,13 @@ use crate::truncate::TruncationPolicy;
 pub(crate) struct SessionState {
     pub(crate) session_configuration: SessionConfiguration,
     pub(crate) history: ContextManager,
+    pub(crate) turn_context_history: Vec<Option<TurnContextItem>>,
+    /// Force a developer-instruction update for collaboration mode on the next turn.
+    ///
+    /// This is set when rollback/backtrack lacks a reliable turn context for the latest user turn,
+    /// so core should re-emit the collaboration developer instructions even if the mode is
+    /// unchanged.
+    pub(crate) force_inject_collaboration_instructions: bool,
```
pakrym-oai: is there a more generic solution where we pick up the last turn context at the point of rollback and inject all items that changed since then?

pakrym-oai: I expect this problem is going to hit many features, sandbox, cwd, personality etc.

pakrym-oai: Having some convention of "we store conversation state on turn context event" then on resume or new turn or rollback inject things that changed.

charley-oai: > the last turn context at the point of rollback

(I don't think this is your point but:) I think if it's anything other than the turn context of the latest user message at the point of rollback, we can't know whether the model saw a mode change btwn that and the latest user message (unless we can make assumptions about when `TurnContextItem` is guaranteed to be available, but that isn't rly safe because of legacy rollouts).

> inject all items that changed since then

This seems reasonable, e.g. if we have any model visible state now or in the future that we want rollback to respect, doesn't make sense to pass around this collaboration-mode specific flag rather than something more general.


Probably we should have some sort of `ModelVisibleState` type (constructable from turn context), and have per-field logic for how to represent a state change to the model. Then, any future (or current) "model visible state" can be handled with the same code, rather than having special case code for collaboration_instructions

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

PR #10258, opened 2026-01-30
fix: unsafe auto-approval of git commands

viyatb-oai:
fixes https://github.com/openai/codex/issues/10160 and some more.

## Description

Hardens Git command safety to prevent approval bypasses for destructive or write-capable invocations (branch delete, risky push forms, output/config-override flags), so these commands no longer auto-run as “safe.”

- `git branch -d` variants (especially in worktrees / with global options like -C / -c)
- `git show|diff|log --output` ... style file-write flags
- risky Git config override flags (-c, --config-env) that can trigger external execution
- dangerous push forms that weren’t fully caught (`--force*`, `--delete`, `+refspec`, `:refspec`)
- grouped short-flag delete forms (e.g. stacked branch flags containing `d/D`)

will fast follow with a common git policy to bring windows to parity.


codex-rs/core/src/command_safety/is_dangerous_command.rs
```diff
@@ -45,74 +124,132 @@ fn is_dangerous_to_call_with_exec(command: &[String]) -> bool {
...
+        );
     }
 
     #[test]
     fn rm_f_is_dangerous() {
-        assert!(command_might_be_dangerous(&vec_str(&["rm", "-f", "/"])));
+        assert_eq!(
+            command_might_be_dangerous(&vec_str(&["rm", "-f", "/"])),
+            true
+        );
```
joshka-oai: This may be an overeager rule about avoiding assert! triggering somewhere (model / AGENTS.md?). It's fine to use assert on bool props, but only for things that aren't going to fail and then hide evaluating assertions in the tests after the failure.

viyatb-oai: yeah this was a rule trigger the model behavior - reverted.

codex-rs/core/src/command_safety/is_dangerous_command.rs
```diff
@@ -27,12 +27,98 @@ pub fn command_might_be_dangerous(command: &[String]) -> bool {
...
+/// Find the first matching git subcommand, skipping known global options that
+/// may appear before it (e.g., `-C`, `-c`, `--git-dir`).
+///
+/// Shared with `is_safe_command` to avoid git-global-option bypasses.
+pub(crate) fn find_git_subcommand<'a>(
+    command: &'a [String],
+    subcommands: &[&str],
+) -> Option<(usize, &'a str)> {
+    let cmd0 = command.first().map(String::as_str)?;
+    if !(cmd0.ends_with("git") || cmd0.ends_with("/git")) {
```
joshka-oai: This is odd

codex-rs/core/src/command_safety/is_dangerous_command.rs
```diff
@@ -27,12 +27,98 @@ pub fn command_might_be_dangerous(command: &[String]) -> bool {
...
+        return None;
+    }
+
+    let mut skip_next = false;
+    for (idx, arg) in command.iter().enumerate().skip(1) {
```
joshka-oai, line 69: (not blocking) observation.

I wonder if it's worth refactoring this code using nom / winow instead of this sort of fragile parsing logic that's difficult to understand and easy to hide bugs in? Given a good enough set of tests I expect codex might do the refactor pretty well.

codex-rs/core/src/command_safety/is_dangerous_command.rs
```diff
@@ -45,6 +131,15 @@ fn is_dangerous_to_call_with_exec(command: &[String]) -> bool {
...
 }
 
+fn git_branch_is_delete(branch_args: &[String]) -> bool {
+    branch_args.iter().any(|arg| {
+        matches!(arg.as_str(), "-d" | "-D" | "--delete")
+            || arg.starts_with("--delete=")
+            || (arg.starts_with("-d") && arg != "-d")
+            || (arg.starts_with("-D") && arg != "-D")
+    })
+}
```
joshka-oai, line 145: I gather this is for argument stacking (`git branch -dv` or `git branch -vd` are both `verbose` + `delete`.?
Would be worth commenting here on how that's handled.

codex-rs/core/src/command_safety/is_dangerous_command.rs
```diff
@@ -45,6 +131,15 @@ fn is_dangerous_to_call_with_exec(command: &[String]) -> bool {
...
+
 #[cfg(test)]
 mod tests {
```
joshka-oai, line 179: A way to make these tests more succinct would be to use parameterized tests with rstest. Not a blocker though, but useful as a way to have a large amount of tested surface on this (generated obv.)

joshka-oai (approved):
General approval with the approach.

I think simplifying this code with nom / winow might be worth considering (but I'd add significant characterization tests before we do that on this code to prevent regression)

codex-rs/core/src/command_safety/is_dangerous_command.rs
bolinfest: Let's not special-case `"/git"` here (to support `"/usr/bin/git"`) since we should fix that for all commands if we're going to do it.

joshka-oai: I meant how can the second part ever be triggered if the first part is false? Odd suggestion from codex to include both checks.

viyatb-oai: fixed!

codex-rs/core/src/command_safety/is_dangerous_command.rs
viyatb-oai: added

PR #9954, opened 2026-01-27
Warn users on enabling underdevelopment features

aibrahim-oai:
image

PR #10259, opened 2026-01-30
Handle stdout broken pipe during JSON streaming

### LINKED ISSUE #10248: codex-cli 0.92.0 panics on Broken pipe when stdout closes during --json streaming (macOS)
### Summary
On macOS, `codex-cli 0.92.0` can crash mid-run with a Rust panic:

> failed printing to stdout: Broken pipe (os error 32)

We started seeing this only after upgrading to 0.92.0. It shows up most often in headless/detached runners where `codex exec --json` is piped (e.g. via `tee`) and the stdout consumer closes.

### Environment
- codex-cli: 0.92.0
- macOS: 15.5 (24F74)
- Auth: API key (not ChatGPT login)

### Observed error
From `codex.stderr.log`:
```
thread 'main' (...) panicked at library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: Broken pipe (os error 32)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### How we trigger it
In our harness we were doing something like:
```bash
codex exec --json ... | tee events.jsonl >/dev/null
```
In headless/detached contexts, if the stdout consumer closes (runner exits, HUP, etc.), `codex` may panic with Broken pipe and terminate the run.

### Minimal repro idea
This should not crash; it should exit gracefully if the consumer closes:
```bash
# Any prompt; keep it tiny to minimize cost.
codex exec --skip-git-repo-check --json "Print OK" | head -n 1 >/dev/null
```
(If `head` closes early, upstream should handle EPIPE without panicking.)

### Workarounds
- Redirect stdout directly to a file instead of piping:
```bash
codex exec --json ... > events.jsonl 2> codex.stderr.log
```
- Pinning to `codex-cli 0.91.0` avoids the issue for us in headless runs.

### Requested behavior
Please handle EPIPE/BrokenPipe on stdout gracefully (no panic) when streaming `--json`.

Happy to provide more logs or try a debug build/backtrace if you point me at what to collect.

etraut-openai:
Handle BrokenPipe in codex exec --json stdout streaming

Fixes [#10248](https://github.com/openai/codex/issues/10248)

Replaces println! in JSONL streaming with a locked stdout writer. Treats ErrorKind::BrokenPipe as a graceful shutdown signal instead of panicking.

This was a long-standing bug, but the user reported seeing it first in 0.91.0. Here's a theory about why. 0.92.0 includes a new “under-development features enabled” warning event (c900de271, #9954) that prints to stdout in JSON mode. More early stdout writes increase the chance of hitting EPIPE when downstream consumers (e.g., head, tee, detached runners) close the pipe mid-run.

This change makes --json streaming robust to stdout closure regardless of when it happens.

### DIFF
--- a/codex-rs/exec/src/event_processor_with_jsonl_output.rs
+++ b/codex-rs/exec/src/event_processor_with_jsonl_output.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::io::Write as _;
 use std::path::PathBuf;
 use std::sync::atomic::AtomicU64;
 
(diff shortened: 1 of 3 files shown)


etraut-openai:
@codex review

PR #10261, opened 2026-01-30
Improve app awareness and skill deconflict

gabec-openai:

Before opening this Pull Request, please read the dedicated "Contributing" markdown file or your PR may be closed:

### DIFF
--- a/codex-rs/core/src/connectors.rs
+++ b/codex-rs/core/src/connectors.rs
@@ -85,6 +85,42 @@ pub fn connector_mention_slug(connector: &AppInfo) -> String {
     connector_name_slug(&connector_display_label(connector))
 }
 
+pub fn render_apps_section(connectors: &[AppInfo]) -> String {
+    let mut lines: Vec<String> = Vec::new();
+    lines.push("## Apps".to_string());
+    lines.push(
+        "Apps (connectors) are MCP tools. Users may refer to apps, connectors, or MCPs interchangeably; treat those requests as app mentions when they match an available app, otherwise ask for clarification."
+            .to_string(),
+    );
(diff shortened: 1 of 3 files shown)


codex-rs/core/src/connectors.rs
```diff
@@ -85,6 +85,42 @@ pub fn connector_mention_slug(connector: &AppInfo) -> String {
     connector_name_slug(&connector_display_label(connector))
 }
 
+pub fn render_apps_section(connectors: &[AppInfo]) -> String {
+    let mut lines: Vec<String> = Vec::new();
+    lines.push("## Apps".to_string());
```
mzeng-openai, line 90: IMHO we should not distinguish between apps and MCPs and the way it works now is we load connector tools as MCP tools so that the model can triage and run the same way as MCPs.
How it works now is we detect any explicit $ mentions of apps, and include the tools from the mentioned apps when sending requests to the responses API. Which means the model doesn't have any apps context unless the harness explicitly add them.
Going forward we should be treating both apps and MCPs the same way as other tools that the model can use, and we'll build a tool search tool to progressively reveal them.

codex-rs/core/src/project_doc.rs
```diff
@@ -69,6 +72,13 @@ pub(crate) async fn get_user_instructions(
         output.push_str(&skills_section);
     }
 
+    if config.features.enabled(Feature::Apps) {
+        if !output.is_empty() {
+            output.push_str("\n\n");
+        }
+        output.push_str(&render_apps_section(apps.unwrap_or(&[])));
```
mzeng-openai, line 79: Unfortunately if we want to change the prompts we need to go through evals and a/b testing to make sure there's no regression.

PR #10271, opened 2026-01-30
[Codex][CLI] Gate image inputs by model modalities

ccy-oai:
###### Summary

  - Add input_modalities to model metadata so clients can determine supported input types.
  - Gate image paste/attach in TUI when the selected model does not support images.
  - Block submits that include images for unsupported models and show a clear warning.
  - Propagate modality metadata through app-server protocol/model-list responses.
  - Update related tests/fixtures.

  ###### Rationale

  - Models support different input modalities.
  - Clients need an explicit capability signal to prevent unsupported requests.
  - Backward-compatible defaults preserve existing behavior when modality metadata is absent.

  ###### Scope

  - codex-rs/protocol, codex-rs/core, codex-rs/tui
  - codex-rs/app-server-protocol, codex-rs/app-server
  - Generated app-server types / schema fixtures

  ###### Trade-offs

  - Default behavior assumes text + image when field is absent for compatibility.
  - Server-side validation remains the source of truth.

  ###### Follow-up

  - Non-TUI clients should consume input_modalities to disable unsupported attachments.
  - Model catalogs should explicitly set input_modalities for text-only models.

  ###### Testing

  - cargo fmt --all
  - cargo test -p codex-tui
  - env -u GITHUB_APP_KEY cargo test -p codex-core --lib
  - just write-app-server-schema
  - cargo run -p codex-cli --bin codex -- app-server generate-ts --out app-server-types 
  - test against local backend
  
image


wiltzius-openai (approved):
lgtm but someone with more client familiarity should take a look  too

ccy-oai:
Going to merge in an hour or so to keep things moving and unblock app/extension.
@aibrahim-oai @pakrym

joshka-oai:
Added some docs in [docs: clarify model input modality contracts](https://github.com/openai/codex/pull/10271/commits/98f546d3953e9f6dd42c6b33170616548d1a4a56)

codex-rs/tui/src/chatwidget.rs
```diff
@@ -3153,6 +3161,16 @@ impl ChatWidget {
         if text.is_empty() && local_images.is_empty() {
             return;
         }
+        if !local_images.is_empty() && !self.current_model_supports_images() {
+            let local_image_paths = local_images.iter().map(|img| img.path.clone()).collect();
+            self.bottom_pane
+                .set_composer_text(text, text_elements, local_image_paths);
```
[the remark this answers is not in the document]
joshka-oai: Clarifying “mention linkage”: when a user picks an item from the $-mention popup, we store both (1) the visible token in the draft text (for example `$foo`) and (2) an internal mapping `mention_paths[name] -> canonical target` (skill file path or `app://...` connector id).

In this blocked-image path we restore the draft via `set_composer_text`, which currently clears `mention_paths`. That means the user still *sees* `$foo` after the warning, but we’ve lost the canonical target they selected.

On retry (after removing images), mention parsing falls back to name-only heuristics. Unambiguous cases may still work, but ambiguous/colliding names can resolve to the wrong target or be dropped. So this is user-visible as “I picked a mention, but after the image warning + retry it didn’t reference what I selected.”

Also, this affects both skill mentions and connector/app mentions (not just skills), because both rely on the same linkage map.

joshka-oai (approved):

codex-rs/tui/src/chatwidget.rs
```diff
@@ -5274,6 +5254,36 @@ impl ChatWidget {
             .unwrap_or(false)
     }
 
+    /// Return whether the effective model currently advertises image-input support.
+    ///
+    /// We intentionally default to `true` when model metadata cannot be read so transient catalog
+    /// failures do not hard-block user input in the UI.
+    fn current_model_supports_images(&self) -> bool {
+        let model = self.current_model();
+        self.models_manager
```
aibrahim-oai, line 5332: I think this info should be on ModelPreset for tui and on the Model info app-server sends to client

codex-rs/protocol/src/openai_models.rs
```diff
@@ -43,6 +48,40 @@ pub enum ReasoningEffort {
...
+    JsonSchema,
+    TS,
+    EnumIter,
+    Hash,
+)]
+#[serde(rename_all = "lowercase")]
+#[strum(serialize_all = "lowercase")]
+pub enum InputModality {
+    /// Plain text turns and tool payloads.
+    #[default]
```
aibrahim-oai: default shouldn't be text. Let's not have a default

aibrahim-oai (changes requested):
Let's not have a default and always have the two text and image as values. Also, add the information on ModelPreset so we don't need to list models again

codex-rs/protocol/src/openai_models.rs
ccy-oai: should be Image+Text as default unless model_cache says differently
```
pub fn default_input_modalities() -> Vec<InputModality> {
    vec![InputModality::Text, InputModality::Image]
}
```

ccy-oai:
> Let's not have a default and always have the two text and image as values. Also, add the information on ModelPreset so we don't need to list models again

Could we get there in two stages? I'd like to avoid having an opening surface area for regression w/ backward compatibility this close to a release cycle. How does this sound 
1. Go out with what we have now: Default [Image, Text] - This maintains backward compatibility with older clients and is equivalent to present state
2. After the launches, we make sure statsig dynamic configs all backfill [Image, Text] and then we pull this default away, if needed for cleaner config.

codex-rs/protocol/src/openai_models.rs
aibrahim-oai: yes

aibrahim-oai (approved):

PR #10273, opened 2026-01-30
plan mode: allow clarification turns without proposed_plan

baumann-oai:
The failure mode was that Plan Mode forced every turn to be either a request_user_input tool call or a final plan, so when a user asked for a clarification (“can you explain the issue?”), the model often responded with a full plan (sometimes inside <proposed_plan>), which prematurely triggered the “Implement this plan?” UI. This change adds an explicit “clarification/answer” turn type and makes it clear that <proposed_plan> is only for the official final plan.


JonKrone:
Thanks! I ran into this a few times

PR #10276, opened 2026-01-30
add missing fields to WebSearchAction and update app-server types

sayan-oai:
- add `WebSearchAction` to app-server v2 types
- add `queries` to `WebSearchAction::Search` type

Updated tests.

### DIFF
--- a/codex-rs/app-server-protocol/src/protocol/v2.rs
+++ b/codex-rs/app-server-protocol/src/protocol/v2.rs
@@ -15,6 +15,7 @@ use codex_protocol::config_types::WebSearchMode;
 use codex_protocol::items::AgentMessageContent as CoreAgentMessageContent;
 use codex_protocol::items::TurnItem as CoreTurnItem;
 use codex_protocol::models::ResponseItem;
+use codex_protocol::models::WebSearchAction;
 use codex_protocol::openai_models::ReasoningEffort;
 use codex_protocol::parse_command::ParsedCommand as CoreParsedCommand;
 use codex_protocol::plan_tool::PlanItemArg as CorePlanItemArg;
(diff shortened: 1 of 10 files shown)


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2113,7 +2114,13 @@ pub enum ThreadItem {
     },
     #[serde(rename_all = "camelCase")]
     #[ts(rename_all = "camelCase")]
-    WebSearch { id: String, query: String },
+    WebSearch {
+        id: String,
+        query: String,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        #[ts(optional)]
```
owenlin0: let's remove these annotations, we want to avoid using these in app-server (but ok in core)

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2113,7 +2114,13 @@ pub enum ThreadItem {
...
+        action: Option<WebSearchAction>,
```
owenlin0, line 2119: Can we redefine an app-server representation of `WebSearchAction` using camelCase? This will be consistent with the rest of the API. A little boilerplate-y but we want to avoid exposing core types directly (which often uses snake_case too)

codex-rs/app-server-protocol/src/protocol/v2.rs
owenlin0: this is because we want the field exposed as: `field: T | null` instead of `field?: T` to be consistent with how our public APIs like Responses API expose null fields

codex-rs/app-server-protocol/src/protocol/v2.rs
owenlin0: Should we also include `domains` from Responses API? Looks like it's also a field for WebSearchAction::Search, seems useful

sayan-oai: i dont think that field actually exists, see my slack msg

owenlin0 (approved):

PR #10281, opened 2026-01-31
Select experimental features with space

pakrym-oai:


codex-rs/tui/src/bottom_pane/experimental_features_view.rs
```diff
@@ -175,6 +175,11 @@ impl BottomPaneView for ExperimentalFeaturesView {
                 ..
             } => self.move_down(),
             KeyEvent {
+                code: KeyCode::Char(' '),
+                modifiers: KeyModifiers::NONE,
+                ..
+            }
+            | KeyEvent {
                 code: KeyCode::Enter,
                 modifiers: KeyModifiers::NONE,
```
joshka-oai: We could actually drop Enter here reasonably. Space means toggle, Enter means submit / next

joshka-oai (approved):

PR #10231, opened 2026-01-30
feat: experimental flags

jif-oai:
## Problem being solved
- We need a single, reliable way to mark app-server API surface as experimental so that:
  1. the runtime can reject experimental usage unless the client opts in
  2. generated TS/JSON schemas can exclude experimental methods/fields for stable clients.

Right now that’s easy to drift or miss when done ad-hoc.

## How to declare experimental methods and fields
- **Experimental method**: add `#[experimental("method/name")]` to the `ClientRequest` variant in `client_request_definitions!`.
- **Experimental field**: on the params struct, derive `ExperimentalApi` and annotate the field with `#[experimental("method/name.field")]` + set `inspect_params: true` for the method variant so `ClientRequest::experimental_reason()` inspects params for experimental fields.

## How the macro solves it
- The new derive macro lives in `codex-rs/codex-experimental-api-macros/src/lib.rs` and is used via `#[derive(ExperimentalApi)]` plus `#[experimental("reason")]` attributes.
- **Structs**:
  - Generates `ExperimentalApi::experimental_reason(&self)` that checks only annotated fields.
  - The “presence” check is type-aware:
    - `Option<T>`: `is_some_and(...)` recursively checks inner.
    - `Vec`/`HashMap`/`BTreeMap`: must be non-empty.
    - `bool`: must be `true`.
    - Other types: considered present (returns `true`).
  - Registers each experimental field in an `inventory` with `(type_name, serialized field name, reason)` and exposes `EXPERIMENTAL_FIELDS` for that type. Field names are converted from `snake_case` to `camelCase` for schema/TS filtering.
- **Enums**:
  - Generates an exhaustive `match` returning `Some(reason)` for annotated variants and `None` otherwise (no wildcard arm).
- **Wiring**:
  - Runtime gating uses `ExperimentalApi::experimental_reason()` in `codex-rs/app-server/src/message_processor.rs` to reject requests unless `InitializeParams.capabilities.experimental_api == true`.
  - Schema/TS export filters use the inventory list and `EXPERIMENTAL_CLIENT_METHODS` from `client_request_definitions!` to strip experimental methods/fields when `experimental_api` is false.

PR #10298, opened 2026-01-31
app-server: add v2 APIs for TUI

bolinfest:
This is the piece of the original https://github.com/openai/codex/pull/10192 that adds the APIs we need to add to app server to support the TUI.

Perhaps I should wait until https://github.com/openai/codex/pull/10231 is in so I can update these with the new `#[experimental]` annotation?


codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -140,6 +148,14 @@ client_request_definitions! {
         params: v2::ThreadRollbackParams,
         response: v2::ThreadRollbackResponse,
     },
+    ThreadCompact => "thread/compact" {
+        params: v2::ThreadCompactParams,
+        response: v2::ThreadCompactResponse,
+    },
+    ThreadShutdown => "thread/shutdown" {
```
pakrym-oai, line 155: ooc can we reuse `turn/interrupt` ?

PR #10316, opened 2026-02-01
Add websocket telemetry metrics and labels

apanasenko-oai:
Summary
- expose websocket telemetry hooks through the responses client so request durations and event processing can be reported
- record websocket request/event metrics and emit runtime telemetry events that the history UI now surfaces
- improve tests to cover websocket telemetry reporting and guard runtime summary updates

Screenshot 2026-01-31 at 5 28 12 PM


wiltzius-openai (approved):
ty ty

PR #10324, opened 2026-02-01
Add a codex.rate_limits event for websockets

rasmusrygaard:
When communicating over websockets, we can't rely on headers to deliver rate limit information. This PR adds a `codex.rate_limits` event that the server can pass to the client to inform them about rate limit usage. The client parses this data the same way we parse rate limit headers in HTTP mode.

This PR also wires up the etag and reasoning headers for websockets


codex-rs/codex-api/src/endpoint/responses_websocket.rs
```diff
@@ -287,6 +314,36 @@ async fn run_websocket_response_stream(
...
+                            && let Some(header_value) = headers
+                                .get(X_CODEX_TURN_STATE_HEADER)
+                                .and_then(|value| value.to_str().ok())
+                        {
+                            let _ = turn_state.set(header_value.to_string());
+                        }
+                        if let Some(snapshot) = parse_rate_limit(&headers) {
+                            let _ = tx_event.send(Ok(ResponseEvent::RateLimits(snapshot))).await;
+                        }
+                        if let Some(etag) = headers
```
pakrym-oai: X-Models-Etag we can return on the connection itself. The model mapping should change during the connection

codex-rs/codex-api/src/endpoint/responses_websocket.rs
```diff
@@ -287,6 +314,36 @@ async fn run_websocket_response_stream(
                         continue;
                     }
                 };
+                if event.kind() == "codex.metadata" {
+                    if let Some(raw_headers) = event.headers()
```
pakrym-oai: Doesn't feel like we should embed these as headers. Should we have a `codex.rate_limits` event with a strongly typed payload?

codex-rs/codex-api/src/endpoint/responses_websocket.rs
```diff
@@ -287,6 +314,36 @@ async fn run_websocket_response_stream(
...
+                        }
+                        if let Some(etag) = headers
+                            .get("X-Models-Etag")
+                            .and_then(|value| value.to_str().ok())
+                        {
+                            let _ = tx_event
+                                .send(Ok(ResponseEvent::ModelsEtag(etag.to_string())))
+                                .await;
+                        }
+                        if headers.contains_key(X_REASONING_INCLUDED_HEADER) {
```
pakrym-oai: This should also come from the original response.

pakrym-oai (changes requested):
few comments.

I think it might be cleaner to add separate event for rate limit than we don't have to cram everything in metadata.

Also some of this information doesn't change during the connection and can be returned on websocket headers.

pakrym-oai (approved):

PR #10337, opened 2026-02-01
chore: remove duplicate mcp-types declaration in Cargo.toml

bolinfest:
Noticed this while auditing our existing uses of this crate. We should switch to https://crates.io/crates/rmcp throughout.


bolinfest:
I guess I'll just go with https://github.com/openai/codex/pull/10349 instead?

PR #10339, opened 2026-02-01
fix(execpolicy) Handle heredoc parsing

dylan-hurd-oai:
## Summary
`python` prefix approval isn't working because parsing was failing on heredoc usage. This updates execpolicy parsing to handle this case, while still guarding against substitution in the command

Open Question for discussion: do we want to fold this into `parse_shell_lc_plain_commands`? I'm less certain so I left it separate for now.

## Testing
- [x] Tested locally and confirmed this works
- [x] Added integration test


codex-rs/core/src/bash.rs
```diff
@@ -119,6 +120,83 @@ pub fn parse_shell_lc_plain_commands(command: &[String]) -> Option<Vec<Vec<Strin
...
+    }
+    tokens.clear();
+    if command_tokens.is_empty() {
+        None
+    } else {
+        Some(command_tokens)
+    }
+}
+
+fn is_shell_separator(token: &str) -> bool {
```
jif-oai, line 188: inline

codex-rs/core/src/bash.rs
```diff
@@ -119,6 +120,83 @@ pub fn parse_shell_lc_plain_commands(command: &[String]) -> Option<Vec<Vec<Strin
...
+    matches!(token, "&&" | "||" | ";" | "|")
+}
+
+fn is_redirection_token(token: &str) -> bool {
```
jif-oai, line 192: inline

codex-rs/core/src/bash.rs
```diff
@@ -119,6 +120,83 @@ pub fn parse_shell_lc_plain_commands(command: &[String]) -> Option<Vec<Vec<Strin
...
+    token.contains('<') || token.contains('>')
+}
+
+fn contains_shell_substitution(line: &str) -> bool {
```
jif-oai, line 196: inline

jif-oai, line 196: This is very very agressive. It will fail most string parsing, regex etc

codex-rs/core/src/bash.rs
jif-oai, line 192: this will prevent a lot of Python usage

codex-rs/core/src/bash.rs
```diff
@@ -119,6 +120,83 @@ pub fn parse_shell_lc_plain_commands(command: &[String]) -> Option<Vec<Vec<Strin
...
+    if tokens.is_empty() {
+        return None;
+    }
+
+    let mut command_tokens = Vec::new();
+    for token in tokens.iter() {
+        if is_redirection_token(token) {
+            break;
+        }
+        command_tokens.push(token.clone());
```
jif-oai, line 178: you could use a drain instead of cloning each time

jif-oai (commented):
The heurisitic seems overly broad. It would be good to extract some prod data to see if it would actually work reasonably well

PR #10348, opened 2026-02-02
fix(core) Stop duplicating dev messages from OverrideTurnContext

dylan-hurd-oai:
## Summary
We are adding personality update messages twice -  the root cause is stale previous_context. OverrideTurnContext recorded one personality update immediately, then the next user turn diffed against old context and emitted another.

## Testing
- [x] Added stronger integration assertions to catch this exact bug by requiring exactly one <personality_spec> developer message

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -2434,6 +2434,7 @@ async fn submission_loop(sess: Arc<Session>, config: Arc<Config>, rx_sub: Receiv
                         personality,
                         ..Default::default()
                     },
+                    &mut previous_context,
                 )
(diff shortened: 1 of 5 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -2597,12 +2599,13 @@ mod handlers {
             return;
         }
 
+        let current_context = sess.new_default_turn_with_sub_id(sub_id).await;
```
pakrym-oai, line 2602: Why doesn't the current code work? new_default_turn_with_sub_id should be creating the context with latest current settings?

PR #10350, opened 2026-02-02
fix: System skills marker includes nested folders recursively

gverma-openai:
Updated system skills bundled with Codex were not correctly replacing the user's skills in their .system folder.

- Fix `.codex-system-skills.marker` not updating by hashing embedded system skills recursively (nested dirs + file contents), so updates trigger a reinstall.
- Added a build Cargo hook to rerun if there are changes in `src/skills/assets/samples/*`, ensuring embedded skill updates rebuild correctly under caching.
- Add a small unit test to ensure nested entries are included in the fingerprint.

### DIFF
--- a/codex-rs/core/src/skills/system.rs
+++ b/codex-rs/core/src/skills/system.rs
@@ -86,21 +86,8 @@ fn read_marker(path: &AbsolutePathBuf) -> Result<String, SystemSkillsError> {
 }
 
 fn embedded_system_skills_fingerprint() -> String {
-    let mut items: Vec<(String, Option<u64>)> = SYSTEM_SKILLS_DIR
-        .entries()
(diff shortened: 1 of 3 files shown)


alexsong-oai (approved):

xl-openai (approved):

codex-rs/core/src/skills/system.rs
```diff
@@ -86,21 +86,8 @@ fn read_marker(path: &AbsolutePathBuf) -> Result<String, SystemSkillsError> {
 }
 
 fn embedded_system_skills_fingerprint() -> String {
```
xl-openai, line 88: FYI We can also just bump SYSTEM_SKILLS_MARKER_SALT.

PR #10351, opened 2026-02-02
fix(rules) Limit rules listed in conversation

dylan-hurd-oai:
## Summary
We should probably warn users that they have a million rules, and help clean them up. But for now, we should handle this unbounded case.

Limit rules listed in conversations, with shortest / broadest rules first.

## Testing
- [x] Updated unit tests

### DIFF
--- a/codex-rs/protocol/src/models.rs
+++ b/codex-rs/protocol/src/models.rs
@@ -233,10 +233,13 @@ impl DeveloperInstructions {
                 if !request_rule_enabled {
                     APPROVAL_POLICY_ON_REQUEST.to_string()
                 } else {
-                    let command_prefixes = format_allow_prefixes(exec_policy);
+                    let command_prefixes =
+                        format_allow_prefixes(exec_policy.get_allowed_prefixes());
                     match command_prefixes {
                         Some(prefixes) => {
-                            format!("{APPROVAL_POLICY_ON_REQUEST_RULE}\n{prefixes}")
+                            format!(
+                                "{APPROVAL_POLICY_ON_REQUEST_RULE}\nApproved command prefixes:\n{prefixes}"
+                            )
                         }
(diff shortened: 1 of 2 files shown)


codex-rs/protocol/src/models.rs
```diff
@@ -399,7 +410,19 @@ fn render_command_prefix(prefix: &[String]) -> String {
 fn format_allow_prefixes(exec_policy: &Policy) -> Option<String> {
     let prefixes = exec_policy.get_allowed_prefixes();
     let lines = render_command_prefix_list(prefixes)?;
-    Some(format!("Approved command prefixes:\n{lines}"))
+    let output = format!("Approved command prefixes:\n{lines}");
+    if output.len() <= MAX_ALLOW_PREFIX_TEXT_BYTES {
```
pakrym-oai: Should we use our truncation logic that leaves a placeholder?

codex-rs/protocol/src/models.rs
```diff
@@ -399,7 +410,19 @@ fn render_command_prefix(prefix: &[String]) -> String {
...
+        return Some(output);
+    }
+
+    let mut cutoff = MAX_ALLOW_PREFIX_TEXT_BYTES;
+    while !output.is_char_boundary(cutoff) {
+        cutoff -= 1;
+    }
+
+    let bounded = &output[..cutoff];
+    let newline_cutoff = bounded.rfind('\n').unwrap_or(cutoff);
```
pakrym-oai: Not telling the model there are more items is rude.

pakrym-oai (approved):

codex-rs/protocol/src/models.rs
dylan-hurd-oai: Right! Good call 😅

PR #10367, opened 2026-02-02
chore: add `codex debug app-server` tooling

celia-oai:
codex debug app-server <user message> forwards the message through codex-app-server-test-client’s send_message_v2 library entry point, using std::env::current_exe() to resolve the codex binary.

for how it looks like, see:

```
celia@com-92114 codex-rs % cargo build -p codex-cli && target/debug/codex debug app-server --help                       
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.34s
Tooling: helps debug the app server

Usage: codex debug app-server [OPTIONS] <COMMAND>

Commands:
  send-message-v2  
  help             Print this message or the help of the given subcommand(s)
````
and
```
celia@com-92114 codex-rs % cargo build -p codex-cli && target/debug/codex debug app-server send-message-v2 "hello world"
   Compiling codex-cli v0.0.0 (/Users/celia/code/codex/codex-rs/cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.38s
> {
>   "method": "initialize",
>   "id": "f8ba9f60-3a49-4ea9-81d6-4ab6853e3954",
>   "params": {
>     "clientInfo": {
>       "name": "codex-toy-app-server",
>       "title": "Codex Toy App Server",
>       "version": "0.0.0"
>     },
>     "capabilities": {
>       "experimentalApi": true
>     }
>   }
> }
< {
<   "id": "f8ba9f60-3a49-4ea9-81d6-4ab6853e3954",
<   "result": {
<     "userAgent": "codex-toy-app-server/0.0.0 (Mac OS 26.2.0; arm64) vscode/2.4.27 (codex-toy-app-server; 0.0.0)"
<   }
< }
< initialize response: InitializeResponse { user_agent: "codex-toy-app-server/0.0.0 (Mac OS 26.2.0; arm64) vscode/2.4.27 (codex-toy-app-server; 0.0.0)" }
> {
>   "method": "thread/start",
>   "id": "203f1630-beee-4e60-b17b-9eff16b1638b",
>   "params": {
>     "model": null,
>     "modelProvider": null,
>     "cwd": null,
>     "approvalPolicy": null,
>     "sandbox": null,
>     "config": null,
>     "baseInstructions": null,
>     "developerInstructions": null,
>     "personality": null,
>     "ephemeral": null,
>     "dynamicTools": null,
>     "mockExperimentalField": null,
>     "experimentalRawEvents": false
>   }
> }
...
```

### CODE REVIEW

File: codex-rs/cli/src/main.rs [hunk 1]
```diff
@@ -105,6 +109,10 @@ enum Subcommand {
     #[clap(visible_alias = "debug")]
```
bolinfest: We need to fix this: I don't like that we changed `debug` to `sandbox` when the sandbox subcommands are nowhere near stable.

Can we put your command under the `debug` subcommand (or introduce a new `debug` command) so that it's `debug app-server` and not `debug-app-server`?

codex-rs/cli/src/main.rs
```diff
@@ -409,6 +424,64 @@ fn run_execpolicycheck(cmd: ExecPolicyCheckCommand) -> anyhow::Result<()> {
...
+        return Ok(codex_rs);
+    }
+
+    anyhow::bail!("could not locate codex-rs workspace from {}", cwd.display());
+}
+
+fn run_debug_app_server_command(cmd: DebugAppServerCommand) -> anyhow::Result<()> {
+    let codex_rs_dir = resolve_codex_rs_dir()?;
+    let user_message = cmd.user_message_parts.join(" ");
+    let status = Command::new("cargo")
```
bolinfest: Why is it running `cargo` instead of linking the thing in? Do we not want to do that?

I don't think this is helpful if someone has to be running `codex` from source to use this. If that's the case, we can add an entry to the `justfile` instead?

codex-rs/cli/src/main.rs
celia-oai: sg!

codex-rs/app-server-test-client/src/lib.rs
```diff
@@ -0,0 +1,1063 @@
...
+    #[command(subcommand)]
+    command: CliCommand,
+}
+
+#[derive(Subcommand)]
+enum CliCommand {
+    /// Send a user message through the Codex app-server.
+    SendMessage {
+        /// User message to send to Codex.
+        #[arg()]
```
bolinfest: Are these empty `#[arg()]` things necessary? `codex-rs/app-server-test-client/src/main.rs` is the only existing file that has them, so that seems suspicious.

Oh, I guess this was part of the move from `main.rs` to `lib.rs`. See if we can live without them, maybe?

bolinfest (approved):

Issue #5090, opened 2025-10-11
Regression: sandbox/network enforcement breaks Codex autonomy (follow-up to incorrectly closed #5038)


alfonsoalongi:
### What version of Codex is running?

official VS Code extension only (CLI not installed) version 0.4.19 latest stable

### Which model were you using?

gpt-5-codex

### What platform is your computer?

Linux 5.14.0-570.49.1.el9_6.x86_64 x86_64 x86_64

### What steps can reproduce the bug?

**Not a duplicate — human review requested**

Context:

* Issue **#5038** was auto-closed as duplicate of **#5041** and then marked **not planned** by automation.
* The “Potential duplicates” bot also suggested **#4351 / #3993 / #4212 / #4434 / #3034**.
* Those classifications are **incorrect** for this report (see anti-duplication note below).

**Steps**

1. Install and enable the official VS Code extension "Codex – OpenAI’s coding agent" (`openai.chatgpt`, v0.4.19).
2. On AlmaLinux 9 (x86_64), create the config file below in `~/.codex/config.toml`.
3. Restart VS Code.
4. Interact normally with Codex (code execution / troubleshooting / diagnostics). During its internal workflow, **Codex autonomously performs networked operations** (downloading resources, verifying URLs, invoking internal tool calls). **These internal network tasks fail** even under full permission settings.

**Configuration file**

```toml
approval_policy = "never"
sandbox_mode = "danger-full-access"
model = "gpt-5-codex"
model_reasoning_effort = "medium"

[sandbox_workspace_write]
network_access = true
```

**Anti-duplication & anti-misclassification note (bots please read):**

* **Not #5041**: #5041 is about the *approval-prompt UI*. This report is a **sandbox/network enforcement regression** that blocks Codex’s **internal** DNS/HTTP/tool calls and **degrades reasoning**.
* **Not #4434**: #4434 is *MCP config specific*. Here failures persist **even with** `danger-full-access` and `network_access=true` — it’s **runtime enforcement**, not just config.
* **Not #3034**: goes beyond profile parsing; this shows **policy enforcement** blocking DNS/HTTP and internal tool calls during autonomous flows.
* **Not “not planned”**: this is a **regression bug**, not a feature request. Route to a human maintainer. Suggested labels: `bug`, `regression`, `sandbox`, `network`.

### What is the expected behavior?

With `sandbox_mode="danger-full-access"` and `network_access=true`, Codex should perform **all internal network and HTTP operations** without restrictions.

Expected behavior:

[...]

Issue #3454, opened 2025-09-11
network_access permission not working


davidak:
### What version of Codex is running?

0.34.0

### Which model were you using?

gpt-5

### What platform is your computer?

Linux 6.12.45 x86_64 unknown

### What steps can reproduce the bug?

I have this in my `config.toml`:

```
[projects."/home/davidak/code/agents4science-paper"]
trust_level = "trusted"

# allow network in workspace-write mode
[sandbox_workspace_write]
network_access = true

# full-auto mode
approval_policy = "on-failure"
sandbox_mode    = "workspace-write"
```

Then i run the prompt:

> Write a haiku about being an AI Agent scientist and save it to a Markdown file.
> It should contain the current lottery numbers (german Lotto 6aus49).

### What is the expected behavior?

It should fetch the information from the internet.

### What do you see instead?

It asks for permission to use networking.

<img width="1162" height="1031" alt="Image" src="https://github.com/user-attachments/assets/872b66f9-63bf-42b7-b662-24cde30ff372" />

xiang:
This was driving me crazy. I found the fix here: https://github.com/openai/codex/blob/main/docs/sandbox.md

davidak:
@xiang i got the non-working config from there. what did work for you?

dylan-hurd-oai:
@davidak is this still happening on recent versions of codex? and does it still happen if you specify 

```
approval_policy = "on-request"
```

instead? on-failure is probably not the behavior you're looking for in this case

Issue #8634, opened 2025-12-31
Mcp-server has no network access even with network access flags enabled


chrisspen:
### What version of Codex is running?

codex-cli 0.77.0

### What subscription do you have?

plus

### Which model were you using?

gpt-5.2 auto

### What platform is your computer?

Linux 6.8.0-90-generic x86_64 x86_64

### What issue are you seeing?

Running `codex mcp-server` reports “Network access is restricted” even when started with:

    -c 'sandbox_permissions=["disk-full-read-access","network-full-access"]'
    -c 'network_access="enabled"'

### What steps can reproduce the bug?

1. Run: codex mcp-server -c 'sandbox_permissions=["disk-full-read-access","network-full-access"]' -c 'network_access="enabled"'
2. Start a session and attempt a network action (e.g., git fetch/push).
3. Codex replies that network access is restricted.

### What is the expected behavior?

Network access should be enabled.

### Additional information

I also tried `--enable network` / `-c features.network=true` fails with:
  `Error: Unknown feature flag: network`

etraut-openai:
This bug report hasn't received any upvotes or additional posts in three months. Closing.

PR #10391, opened 2026-02-02
docs: add macOS network sandbox troubleshooting to README

fede-kamel:
## Summary

- Adds a **Troubleshooting** section to the README documenting the macOS seatbelt sandbox silently ignoring `network_access = true` from `config.toml`
- Includes a confirmed workaround (`--sandbox danger-full-access` CLI flag and shell alias)
- References #10390 for full context and a proposed code fix

## Context

On macOS, the seatbelt sandbox unconditionally sets `CODEX_SANDBOX_NETWORK_DISABLED=1`, causing all outbound network calls to fail — even when the user has explicitly set `network_access = true` in `~/.codex/config.toml`. This affects cloud CLIs (OCI, AWS, GCP), curl, package managers, and any workflow requiring network from within a Codex session.

Multiple closed issues reference this problem (#8714, #5041, #3454, #5090) but none document the workaround in an easily discoverable place. This small README addition helps users find the fix without digging through issue threads.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,34 @@ Run `codex` and select **Sign in with ChatGPT**. We recommend signing into your 
 
 You can also use Codex with an API key, but this requires [additional setup](https://developers.openai.com/codex/auth#sign-in-with-an-api-key).
 
+## Troubleshooting
+
+<details>
+<summary><strong>macOS: Network calls fail even with <code>network_access = true</code></strong></summary>
+
+On macOS, the seatbelt sandbox unconditionally blocks outbound network access regardless of your `config.toml` settings. Commands that require network (cloud CLIs, curl, API clients, package managers) will fail with connection timeouts.
+
+**Symptoms:**
+- `CODEX_SANDBOX_NETWORK_DISABLED=1` is set in the sandbox environment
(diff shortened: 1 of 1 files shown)


etraut-openai:

fede-kamel:
Understood — no problem at all. We respect the contribution guidelines.

That said, it's a bit disappointing that there isn't a clearer channel for engineers who use Codex daily to contribute back meaningful fixes. This issue affects real workflows (cloud CLIs, API clients, package managers — all broken under the macOS sandbox).

We did the work here — root cause tracing through the Rust code, confirmed workaround, integration tests proving the seatbelt behavior at the OS level, and tracing to help users debug the config resolution path. The issue and this PR will stay up as a reference for anyone searching.

Wishing you all the best with the project. It's a great product — hope the feedback loop with your user community keeps improving.

etraut-openai:
Just a few clarifications about the related issues you mentioned were "closed without workarounds":
* #8714 was closed because it was specifically about project-local config files, which were not supported at the time.
* #9298 is still active.
* #8634 is also still active but is unrelated to this issue because it's talking about MCP servers which don't run within the sandbox.

fede-kamel:
Fair point — I overstated the connection to those issues. I'll clean up that comment. Thanks for the clarification.

PR #10392, opened 2026-02-02
fix: add seatbelt network_access test coverage and sandbox policy tracing

fede-kamel:
## Summary

- Add **unit tests** verifying seatbelt policy includes/excludes network rules based on `network_access` flag
- Add **integration test** proving seatbelt actually allows/blocks socket creation at the OS level — tests both `true` and `false` in a single test, using localhost socket bind (no external network dependency, CI-safe)
- Add **tracing** to `derive_sandbox_policy()` logging the resolved sandbox mode, its source (`cli-flag` / `config-profile` / `config-toml` / `default`), and the `network_access` value

## Context

All existing seatbelt tests use `network_access: false`. There was zero coverage for the `true` path — the seatbelt network policy was never verified in tests, either at the arg-generation level or at the OS-enforcement level.

The tracing helps users debug why network is blocked. Today, when `[sandbox_workspace_write] network_access = true` doesn't take effect (e.g., because no `sandbox_mode` is set and the fallback is `ReadOnly`), there is no log output explaining what happened. The new `tracing::info!` lines make the resolution chain visible.

Ref: #10390

## Test plan

- [ ] `cargo test -p codex-core seatbelt` passes on macOS
- [ ] `create_seatbelt_args_includes_network_policy_when_network_access_true` — verifies `(allow network-outbound)` present in policy string
- [ ] `create_seatbelt_args_excludes_network_policy_when_network_access_false` — verifies `(allow network-outbound)` absent
- [ ] `seatbelt_network_access_true_allows_socket_creation` — runs `sandbox-exec` with both policies, proves socket bind succeeds/fails accordingly
- [ ] Tracing output visible with `RUST_LOG=info` when running the CLI


etraut-openai:

fede-kamel:
@etraut-openai  Same as noted on #10391 — we understand the policy and respect it.

Just want to highlight that this PR includes integration tests that actually uncovered pre-existing test failures in the seatbelt suite (`/var/folders` vs `/private/var/folders` path mismatch on macOS). That's the kind of thing that's hard to catch without people running the tests on real hardware.

The code, tests, and tracing are all here if the team ever wants to reference or cherry-pick from them. Best of luck.

etraut-openai:
@fede-kamel, I appreciate the time you took to understand the problem here. Most code contributions we've received historically haven't had this level of analysis or understanding behind them.

After taking a closer look at this PR, I think this is something we'd be interested in. I'm going to reopen and put it in our review queue. I'll also take a look at your other associated PR.

Thanks again!

etraut-openai:
@fede-kamel, you'll need to sign the CLA.

fede-kamel:

etraut-openai:
Over the next few weeks, we're doing some overhaul work on the sandbox mechanism and the config controls. Since a bunch of this is likely to change, I'm going to close this PR. Thanks again for taking the time to submit this.

fede-kamel:
sounds good @etraut-openai - Please keep us updated on this change as it is impacting the codex community. Thanks!

fede-kamel:
@etraut-openai — circling back on this ~7 weeks later.

The sandbox overhaul has been impressive (proxy-aware networking, split filesystem policies, new `codex-sandboxing` crate, 20+ commits to seatbelt.rs). The `/var/folders` vs `/private/var/folders` path mismatch we flagged was fixed via `canonicalize()` — great.

However, the core issue this PR was motivated by — **#10390 (`network_access = true` silently ignored)** — is still open and actively impacting users (Maven builds, iOS toolchains, Bazel workflows all broken by silent network blocking as of today).

Two items from this PR that still appear unaddressed:

1. **Integration test for actual network access under seatbelt** — current tests verify policy *string generation*, but there is no test proving `sandbox-exec` actually allows/blocks socket creation. The proxy-based model makes this even more important now.

2. **Tracing in `derive_sandbox_policy()`** — there is still no `tracing::info!` logging the resolved sandbox mode, its source (cli-flag / config-profile / config-toml / default), or the `network_access` value. Users have zero visibility into why their config is not being respected, which is exactly the debugging gap this PR aimed to close.

Happy to rebase and adapt this work to the new architecture if it would be useful. The code and approach are all still available on the branch.

PR #10397, opened 2026-02-02
DynamicToolCallResponse can include content_items

aaronl-openai:

Before opening this Pull Request, please read the dedicated "Contributing" markdown file or your PR may be closed:


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2694,8 +2671,16 @@ pub struct DynamicToolCallParams {
 #[serde(rename_all = "camelCase")]
 #[ts(export_to = "v2/")]
 pub struct DynamicToolCallResponse {
-    pub output: String,
+    /// Optional plain-text output. When omitted, the app server will derive a
+    /// string representation from `contentItems` for legacy consumers.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub output: Option<String>,
```
jif-oai: If this is exclusive with `content_items`, it should become an enum

codex-rs/app-server/src/dynamic_tools.rs
```diff
@@ -56,3 +61,24 @@ pub(crate) async fn on_call_response(
         error!("failed to submit DynamicToolResponse: {err}");
     }
 }
+
+fn normalize_output(
```
jif-oai: Huh? This is not super clean to do this... `content_items` will have all the serialization artifacts while output will be a raw string
We should already have some code to handle `FunctionCallOutputContentItem`

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -15,6 +14,7 @@ use codex_protocol::config_types::Verbosity;
 use codex_protocol::config_types::WebSearchMode;
 use codex_protocol::items::AgentMessageContent as CoreAgentMessageContent;
 use codex_protocol::items::TurnItem as CoreTurnItem;
+use codex_protocol::models::FunctionCallOutputContentItem;
```
owenlin0: We want to avoid exposing codex "core" types directly, since we want:
- camelCase for consistency in the app-server API (whereas core types tend to use snake_case)
- having app-server API types serve as a layer of abstraction between client and codex core

can we define this a new struct in this file called `DynamicToolCallOutputContentItem` and map to/from `FunctionCallOutputContentItem` as necessary?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2694,8 +2671,16 @@ pub struct DynamicToolCallParams {
 #[serde(rename_all = "camelCase")]
 #[ts(export_to = "v2/")]
 pub struct DynamicToolCallResponse {
-    pub output: String,
+    /// Optional plain-text output. When omitted, the app server will derive a
+    /// string representation from `contentItems` for legacy consumers.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
```
owenlin0: nit: let's remove these annotations

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2694,8 +2671,16 @@ pub struct DynamicToolCallParams {
...
+    pub output: Option<String>,
     pub success: bool,
+    /// Optional structured content for the tool output (for example, images).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
```
owenlin0: remove annotations here too

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2694,8 +2671,16 @@ pub struct DynamicToolCallParams {
...
 #[ts(export_to = "v2/")]
 pub struct DynamicToolCallResponse {
-    pub output: String,
+    /// Optional plain-text output. When omitted, the app server will derive a
+    /// string representation from `contentItems` for legacy consumers.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub output: Option<String>,
     pub success: bool,
+    /// Optional structured content for the tool output (for example, images).
```
owenlin0: Should we place this above the `output` field and indicate in the docstring this is preferred over `output`? Also worth mentioning that one of these fields must be set.

codex-rs/protocol/src/dynamic_tools.rs
```diff
@@ -27,4 +29,7 @@ pub struct DynamicToolResponse {
     pub call_id: String,
     pub output: String,
```
jif-oai: Why don't you put this as optional this time?

jif-oai (changes requested):

codex-rs/app-server/src/dynamic_tools.rs
owenlin0: Right, I think for this to work end-to-end we need to make sure we ultimately pass these `content_items` to Responses API here: https://platform.openai.com/docs/api-reference/responses/create#responses_create-input-input_item_list-item-function_tool_call_output-output-array-input_image 

if we serialize as string I don't think this would work end-to-end, right?

codex-rs/app-server-protocol/src/protocol/v2.rs
aaronl-openai: yup sounds good!

codex-rs/app-server-protocol/src/protocol/v2.rs
aaronl-openai: that's definitely cleaner but we would loose backwards compatibility. is that okay?

aaronl-openai:
closed in favor of https://github.com/openai/codex/pull/10567

PR #10402, opened 2026-02-02
feat: add `--experimental` to `generate-ts`

jif-oai:
Adding a `--experimental` flag to the `generate-ts` fct in the app-sever.

It can be called through one of those 2 command
```
just write-app-server-schema --experimental
codex app-server generate-ts --experimental
```


jif-oai:
@codex review

shijie-oai (approved):

PR #10404, opened 2026-02-02
Cleanup collaboration mode variants

charley-oai:
## Summary

This PR simplifies collaboration modes to the visible set `default | plan`, while preserving backward compatibility for older partners that may still send legacy mode
names.

Specifically:
- Renames the old Code behavior to **Default**.
- Keeps **Plan** as-is.
- Removes **Custom** mode behavior (fallbacks now resolve to Default).
- Keeps `PairProgramming` and `Execute` internally for compatibility plumbing, while removing them from schema/API and UI visibility.
- Adds legacy input aliasing so older clients can still send old mode names.

## What Changed

1. Mode enum and compatibility
- `ModeKind` now uses `Plan` + `Default` as active/public modes.
- `ModeKind::Default` deserialization accepts legacy values:
  - `code`
  - `pair_programming`
  - `execute`
  - `custom`
- `PairProgramming` and `Execute` variants remain in code but are hidden from protocol/schema generation.
- `Custom` variant is removed; previous custom fallbacks now map to `Default`.

2. Collaboration presets and templates
- Built-in presets now return only:
  - `Plan`
  - `Default`
- Template rename:
  - `core/templates/collaboration_mode/code.md` -> `default.md`
- `execute.md` and `pair_programming.md` remain on disk but are not surfaced in visible preset lists.

3. TUI updates
- Updated user-facing naming and prompts from “Code” to “Default”.
- Updated mode-cycle and indicator behavior to reflect only visible `Plan` and `Default`.
- Updated corresponding tests and snapshots.

4. request_user_input behavior
- `request_user_input` remains allowed only in `Plan` mode.
- Rejection messaging now consistently treats non-plan modes as `Default`.

5. Schemas
- Regenerated config and app-server schemas.
- Public schema types now advertise mode values as:
  - `plan`
  - `default`

## Backward Compatibility Notes

- Incoming legacy mode names (`code`, `pair_programming`, `execute`, `custom`) are accepted and coerced to `default`.
- Outgoing/public schema surfaces intentionally expose only `plan | default`.
- This allows tolerant ingestion of older partner payloads while standardizing new integrations on the reduced mode set.

## Codex author
`codex fork 019c1fae-693b-7840-b16e-9ad38ea0bd00`

### DIFF
--- a/codex-rs/core/config.schema.json
+++ b/codex-rs/core/config.schema.json
@@ -378,10 +378,7 @@
       "description": "Initial collaboration mode to use when the TUI starts.",
       "enum": [
         "plan",
-        "code",
-        "pair_programming",
-        "execute",
-        "custom"
+        "default"
       ],
       "type": "string"
     },
@@ -1021,7 +1018,7 @@
             }
           ],
           "default": null,
-          "description": "Start the TUI in the specified collaboration mode (plan/execute/etc.). Defaults to unset."
+          "description": "Start the TUI in the specified collaboration mode (plan/default). Defaults to unset."
         },
         "notification_method": {
           "allOf": [
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -370,7 +370,7 @@ impl Codex {
         // TODO (aibrahim): Consolidate config.model and config.model_reasoning_effort into config.collaboration_mode
         // to avoid extracting these fields separately and constructing CollaborationMode here.
         let collaboration_mode = CollaborationMode {
-            mode: ModeKind::Custom,
+            mode: ModeKind::Default,
             settings: Settings {
                 model: model.clone(),
                 reasoning_effort: config.model_reasoning_effort,
@@ -2628,7 +2628,7 @@ mod handlers {
             } => {
                 let collaboration_mode = collaboration_mode.or_else(|| {
                     Some(CollaborationMode {
-                        mode: ModeKind::Custom,
+                        mode: ModeKind::Default,
                         settings: Settings {
                             model: model.clone(),
                             reasoning_effort: effort,
@@ -4942,7 +4942,7 @@ mod tests {
         let model_info = ModelsManager::construct_model_info_offline(model.as_str(), &config);
         let reasoning_effort = config.model_reasoning_effort;
         let collaboration_mode = CollaborationMode {
-            mode: ModeKind::Custom,
+            mode: ModeKind::Default,
             settings: Settings {
                 model,
                 reasoning_effort,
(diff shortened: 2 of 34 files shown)


charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

codex-rs/core/config.schema.json
```diff
@@ -1021,7 +1018,7 @@
             }
           ],
           "default": null,
-          "description": "Start the TUI in the specified collaboration mode (plan/execute/etc.). Defaults to unset."
+          "description": "Start the TUI in the specified collaboration mode (plan/default). Defaults to unset."
```
shijie-oai, line 1021: `Defaults to unset` - is this still accurate?

codex-rs/core/src/codex.rs
```diff
@@ -370,7 +370,7 @@ impl Codex {
         // TODO (aibrahim): Consolidate config.model and config.model_reasoning_effort into config.collaboration_mode
         // to avoid extracting these fields separately and constructing CollaborationMode here.
```
shijie-oai, line 371: Just want to point out that we need to think hard about migration path forward if we need to do that. Currently we are not even tracking them.

codex-rs/core/src/features.rs
```diff
@@ -121,7 +121,7 @@ pub enum Feature {
     SkillEnvVarDependencyPrompt,
     /// Steer feature flag - when enabled, Enter submits immediately instead of queuing.
     Steer,
-    /// Enable collaboration modes (Plan, Code, Pair Programming, Execute).
+    /// Enable collaboration modes (Plan, Default).
     CollaborationModes,
```
shijie-oai, line 125: Should we just remove this? I am not sure if we are going to turn this off at anytime in the future.

shijie-oai: I guess we have to also remove the enable flag on app side as we do that.

shijie-oai (approved):
Overall looks straight forward - one concern I do have (might be outdated info) is when we go from plan -> code, we also tell the model that we are exiting plan mode. want to make sure that behavior is not lost. Also let @aibrahim-oai take another look.

aibrahim-oai (approved):

codex-rs/core/src/features.rs
charley-oai: Makes sense, will do in followup PR

codex-rs/core/config.schema.json
charley-oai: > "Defaults to unset" is still technically accurate for the config field itself:
> 
> - tui.experimental_mode is Option<ModeKind> with #[serde(default)], so missing value deserializes to None (schema default: null).
> 
> But behavior-wise, None now means:
> 
> - no explicit override selected by the user, and then
> - if collaboration modes are enabled, startup resolves to the Default preset (not Plan) via the runtime fallback.

PR #10406, opened 2026-02-02
fix: clarify deprecation message for features.web_search

sayan-oai:
clarify that the new `web_search` is not a feature flag under `[features]` in the deprecation CTA


joshka-oai (approved):

sayan-oai:
noting here that we should probably list auto-enablement in the deprecation reason. will do this after we fix auto-enable issues with azure

PR #10408, opened 2026-02-02
fix WebSearchAction type clash between v1 and v2

sayan-oai:
type clash; app-server generated types were still using the v1 snake_case `WebSearchAction`, so there was a mismatch between the camelCase emitted types and the snake_case types we were trying to parse.

Updated v2 `WebSearchAction` to export into the `v2/` type set and updated `ThreadItem` to use that.

### Tests
Ran new `just write-app-server-schema` to surface changes to schema, the import looks correct now.


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2167,6 +2167,7 @@ pub enum ThreadItem {
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
 #[serde(tag = "type", rename_all = "camelCase")]
 #[ts(tag = "type", rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
```
owenlin0, line 2173: good catch, must have missed this when i reviewed!

owenlin0 (approved):

PR #5939, opened 2025-10-30
[app-server] remove serde(skip_serializing_if = "Option::is_none") annotations

owenlin0:
We had this annotation everywhere in app-server APIs which made it so that fields get serialized as `field?: T`, meaning if the field as `None` we would omit the field in the payload. Removing this annotation changes it so that we return `field: T | null` instead, which makes codex app-server's API more aligned with the convention of public OpenAI APIs like Responses.

Separately, remove the `#[ts(optional_fields = nullable)]` annotations that were recently added which made all the TS types become `field?: T | null` which is not great since clients need to handle undefined and null.

I think generally it'll be best to have optional types be either:
- `field: T | null` (preferred, aligned with public OpenAI APIs)
- `field?: T` where we have to, such as types generated from the MCP schema: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-06-18/schema.ts (see changes to `mcp-types/`)

I updated @etraut-openai's unit test to check that all generated TS types are one or the other, not both (so will error if we have a type that has `field?: T | null`). I don't think there's currently a good use case for that - but we can always revisit.

PR #5901, opened 2025-10-29
Add missing "nullable" macro to protocol structs that contain optional fields

etraut-openai:
This PR addresses a current hole in the TypeScript code generation for the API server protocol. Fields that are marked as "Optional<>" in the Rust code are serialized such that the value is omitted when it is deserialized — appearing as `undefined`, but the TS type indicates (incorrectly) that it is always defined but possibly `null`. This can lead to subtle errors that the TypeScript compiler doesn't catch. The fix is to include the `#[ts(optional_fields = nullable)]` macro for all protocol structs that contain one or more `Optional<>` fields.

This PR also includes a new test that validates that all TS protocol code containing "| null" in its type is marked optional ("?") to catch cases where `#[ts(optional_fields = nullable)]` is omitted.

PR #10412, opened 2026-02-02
fix(app-server): fix TS annotations for optional fields on requests

owenlin0:
This updates our generated TypeScript types to be more correct with how the server actually behaves, **specifically for JSON-RPC requests**.

Before this PR, we'd generate `field: T | null`. After this PR, we will have `field?: T | null`. The latter matches how the server actually works, in that if an optional field is omitted, the server will treat it as null. This also makes it less annoying in theory for clients to upgrade to newer versions of Codex, since adding a new optional field to a JSON-RPC request should not require a client change.

NOTE: This only applies to JSON-RPC requests. All other payloads (i.e. responses, notifications) will return `field: T | null` as usual.


gpeal (commented):
Does this change ALL nullable fields to be nullable or optional or only actually optional ones?

I can't find the PR but I'm almost positive that, last year, used to do this and explicitly moved away from it because of the ambiguitiy between null and undefined in many cases. Let's make sure we square this with that decision before moving forward

Maybe it was [this one](https://github.com/openai/codex/pull/5901)?

@bolinfest @etraut-openai do you remember when/where we did that?

owenlin0:
@gpeal Yes, I was the one who landed a change to explicitly move away from `field?: T | null` everywhere (this was all structs, whether they were sent by the client or server). This PR: https://github.com/openai/codex/pull/5939

However there is a distinction between structs sent by the client vs. structs sent by the server, and this only reintroduces `field?: T | null` for payloads sent by client->server.

bolinfest (approved):
Thanks for the thorough `AGENTS.md` update!

PR #10413, opened 2026-02-02
feat(linux-sandbox): vendor bubblewrap and wire it with FFI

viyatb-oai:
## Summary

Vendor Bubblewrap into the repo and add minimal build plumbing in `codex-linux-sandbox` to compile/link it.

## Why

We want to move Linux sandboxing toward Bubblewrap, but in a safe two-step rollout:
1) vendoring/build setup (this PR),  
2) runtime integration (follow-up PR).

## Included

- Add `codex-rs/vendor/bubblewrap` sources.
- Add build-time FFI path in `codex-rs/linux-sandbox`.
- Update `build.rs` rerun tracking for vendored files.
- Small vendored compile warning fix (`sockaddr_nl` full init).

follow up in https://github.com/openai/codex/pull/9938


codex-rs/vendor/bubblewrap/.github/workflows/check.yml
```diff
@@ -0,0 +1,105 @@
+name: CI checks
```
bolinfest, line 1: We should probably exclude the `.github/workflows` folder from the `bubblewrap` repo? We aren't going to run it, so I worry it may just confuse things?

Though I concede it makes the diff with the true repo slightly unclean.

codex-rs/linux-sandbox/build.rs
```diff
@@ -0,0 +1,113 @@
...
+    let src_dir = resolve_bwrap_source_dir(&manifest_dir)?;
+
+    let libcap = pkg_config::Config::new()
+        .probe("libcap")
+        .map_err(|err| format!("libcap not available via pkg-config: {err}"))?;
+
+    let config_h = out_dir.join("config.h");
+    std::fs::write(
+        &config_h,
+        "#pragma once\n#define PACKAGE_STRING \"bubblewrap built at codex build-time\"\n",
```
bolinfest: consider `r#` if you think that will make things cleaner?

bolinfest (approved):
Exciting!

codex-rs/vendor/bubblewrap/.github/workflows/check.yml
viyatb-oai: I asked codex about it and apparently Github actions only takes workflows from the top level `.github` directory

codex-rs/linux-sandbox/build.rs
viyatb-oai: added

PR #10414, opened 2026-02-02
Require models refresh on cli version mismatch

aibrahim-oai:
New versions can need to receive different model configurations. If the cache is for a different version, refresh.


aibrahim-oai:
@codex review

pakrym-oai:
shouldn't etag handle this?

pakrym-oai (approved):

PR #10423, opened 2026-02-02
fix(auth): isolate chatgptAuthTokens concept to auth manager and app-server

owenlin0:
So that the rest of the codebase (like TUI) don't need to be concerned whether ChatGPT auth was handled by Codex itself or passed in via app-server's external auth mode.

### DIFF
--- a/codex-rs/core/src/auth.rs
+++ b/codex-rs/core/src/auth.rs
@@ -16,6 +16,7 @@ use std::sync::Mutex;
 use std::sync::RwLock;
 
 use codex_app_server_protocol::AuthMode as ApiAuthMode;
+use codex_otel::TelemetryAuthMode;
 use codex_protocol::config_types::ForcedLoginMethod;
 
 pub use crate::auth::storage::AuthCredentialsStoreMode;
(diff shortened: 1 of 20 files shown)


codex-rs/otel/src/lib.rs
```diff
@@ -30,6 +30,12 @@ pub enum ToolDecisionSource {
     User,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Display)]
+pub enum TelemetryAuthMode {
```
owenlin0, line 36: needed to avoid a circular dep

celia-oai: can we add a docstring to explain this?

codex-rs/core/src/auth.rs
```diff
@@ -47,6 +48,15 @@ pub enum AuthMode {
...
+        match mode {
+            AuthMode::ApiKey => TelemetryAuthMode::ApiKey,
+            AuthMode::Chatgpt => TelemetryAuthMode::Chatgpt,
+        }
+    }
+}
+
 /// Authentication mechanism used by the current user.
 #[derive(Debug, Clone)]
 pub enum CodexAuth {
```
celia-oai, line 62: if this is now internal is there a way to make this enum not public while making the impl functions public?

owenlin0: unfortunately no

codex-rs/otel/src/lib.rs
owenlin0: added

celia-oai (approved):

PR #10445, opened 2026-02-03
Add thread/compact v2 

aibrahim-oai:
- add `thread/compact` as a trigger-only v2 RPC that submits `Op::Compact` and returns `{}` immediately.
- add v2 compaction e2e coverage for success and invalid/unknown thread ids, and update protocol schemas/docs.


codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -208,6 +208,10 @@ client_request_definitions! {
         params: v2::ThreadUnarchiveParams,
         response: v2::ThreadUnarchiveResponse,
     },
+    ThreadCompact => "thread/compact" {
```
owenlin0: can we call this `thread/compact/start`? this will be more in line with `turn/start` and `review/start` that kick off a turn + items.

codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -208,6 +208,10 @@ client_request_definitions! {
...
+        params: v2::ThreadCompactParams,
+        response: v2::ThreadCompactResponse,
```
owenlin0: then these can be `ThreadCompactStartParams` and `ThreadCompactStartResponse`

codex-rs/app-server/README.md
```diff
@@ -237,6 +238,15 @@ Use `thread/unarchive` to move an archived rollout back into the sessions direct
 { "id": 24, "result": { "thread": { "id": "thr_b" } } }
 ```
 
+### Example: Trigger thread compaction
+
+Use `thread/compact` to trigger manual history compaction for a thread. The request returns immediately, and you can observe progress via standard `turn/*` and `item/*` notifications.
```
owenlin0: Should the client only expect a turn with a single compaction item? If so, worth making it a bit more clear in the docs here.

aibrahim-oai: okay

shijie-oai: Is compaction asynchronous?

aibrahim-oai: it's async thread manager wise but blocks threads

shijie-oai: so if that is the case do we want to return right away? cause otherwise folks will think that their instance is stuck depending on duration of the compaction.

owenlin0 (approved):

aibrahim-oai:
@codex fix ci

Issue #10442, opened 2026-02-03
Plan options re-trigger on /agent switch


doublemover:
### What version of Codex is running?

codex-cli 0.94.0

### What subscription do you have?

Pro

### Which model were you using?

gpt-5.2-codex

### What platform is your computer?

Microsoft Windows NT 10.0.26200.0 x64

### What terminal emulator and version are you using (if applicable)?

Windows Terminal, Powershell 7.5.4

### What issue are you seeing?

After using `/plan` mode and confirming the choices, while the plan is being carried out, if you steer the agent to spawn additional agents to spread out and handle the task more quickly and then use `/agent` to swap to one of the spawned agents, using `/agent` again to swap back to the original agent re-triggers the already-confirmed `/plan` options dialogue. 

I did not want to interrupt its work by re-confirming choices, so I pressed escape, which both canceled the dialogue and interrupted its work. Telling it continue allowed it to resume without issue. 

### What steps can reproduce the bug?

- Use /plan mode, pick a task that lends itself well to sub-agents 
- Confirm your choices for the plan, allow it to begin carrying out the tasks
- Steer the agent and tell it to spawn other agents so they can spread out and complete the task more efficiently
- Use /agent to swap to one of the spawned agents and see what it's up to
- Use /agent to swap back to the original agent
- Plan dialogue should incorrectly re-trigger at this point
Uploaded thread: 019c10fe-7132-79a2-859b-2eb167d93103

### What is the expected behavior?

It should swap back to the original agent without re-triggering the plan dialogue

### Additional information

Lengthy sessions cause extended delays when submitting /feedback, they also cause codex to hang (indefinitely?) when quitting normally

charley-oai:
Addressed by https://github.com/openai/codex/pull/10447

charley-oai:
This should be fixed in the next release

PR #10447, opened 2026-02-03
Fix plan implementation prompt reappearing after /agent thread switch

charley-oai:
## Summary

This fixes a UX bug (https://github.com/openai/codex/issues/10442) where the **"Implement this plan?"** prompt could reappear after switching agents with `/agent` and then switching back to the original agent during plan execution.

## Root Cause

On thread switch, the TUI rebuilds `ChatWidget`, replays buffered thread events, then drains any queued live events.

In this flow, a `TurnComplete` can be handled twice for the same logical turn:
1. replayed (`from_replay = true`)
2. then live (`from_replay = false`)

`ChatWidget` used `saw_plan_item_this_turn` to decide whether to show the plan implementation prompt, but that flag was only reset on `TurnStarted`.
If duplicate completion events occurred, stale `saw_plan_item_this_turn = true` could cause the prompt to re-trigger unexpectedly.

## Fix

- Clear `saw_plan_item_this_turn` at the end of `on_task_complete`, after prompt gating runs.
- This keeps the flag truly turn-scoped and prevents duplicate `TurnComplete` handling from reopening the prompt.

### DIFF
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -1070,6 +1070,9 @@ impl ChatWidget {
         if !from_replay && self.queued_user_messages.is_empty() {
             self.maybe_prompt_plan_implementation();
         }
+        // `saw_plan_item_this_turn` is only meaningful while a turn is active; clearing it here
+        // avoids replay/live duplicate TurnComplete events re-triggering the plan prompt.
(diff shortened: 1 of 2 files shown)


charley-oai:
@codex review this

shijie-oai (approved):

charley-oai:
@codex review this

PR #10448, opened 2026-02-03
feat: add APIs to list and download public remote skills

xl-openai:
Add API to list / download from remote public skills


codex-rs/core/src/skills/remote.rs
```diff
@@ -0,0 +1,292 @@
...
+#[derive(Debug, Deserialize)]
+struct RemoteSkillFileRangePayload {
+    start: u64,
+    length: u64,
+}
+
+pub async fn list_remote_skills(config: &Config) -> Result<Vec<RemoteSkillSummary>> {
+    let base_url = config.chatgpt_base_url.trim_end_matches('/');
+    let base_url = base_url.strip_suffix("/backend-api").unwrap_or(base_url);
+    let url = format!("{base_url}/public-api/hazelnuts/");
```
xl-openai, line 79: Public API has a slightly different format and it doesn't require Auth.

codex-rs/core/src/skills/remote.rs
```diff
@@ -0,0 +1,292 @@
...
+        hazelnut.id.clone()
+    } else {
+        hazelnut.name.clone()
+    };
+    let output_dir = downloaded_skills_dir(config).join(dir_name);
+    tokio::fs::create_dir_all(&output_dir)
+        .await
+        .context("Failed to create downloaded skills directory")?;
+
+    let allowed_files = hazelnut.files.keys().cloned().collect::<HashSet<String>>();
```
xl-openai, line 163: remote skill will be a zip and need to unzip while downloading.

codex-rs/core/src/skills/remote.rs
```diff
@@ -0,0 +1,313 @@
...
+#[derive(Debug, Deserialize)]
+struct RemoteSkillFileRangePayload {
+    start: u64,
+    length: u64,
+}
+
+pub async fn list_remote_skills(config: &Config) -> Result<Vec<RemoteSkillSummary>> {
+    let base_url = config.chatgpt_base_url.trim_end_matches('/');
+    let base_url = base_url.strip_suffix("/backend-api").unwrap_or(base_url);
+    let url = format!("{base_url}/public-api/hazelnuts/");
```
alexsong-oai, line 79: should we have a helper function for constructing public-api/hazelnuts url?  Then we can utilize it in all the places

alexsong-oai (approved):

codex-rs/core/src/skills/remote.rs
xl-openai: Currently it is only used for skills-remote. we could do so if we need to use it in more places.

PR #10461, opened 2026-02-03
feat(tui): pace catch-up stream chunking with hysteresis

joshka-oai:
## Summary
- preserve baseline streaming behavior (smooth mode still commits one line per 50ms tick)
- extract adaptive chunking policy and commit-tick orchestration from ChatWidget into `streaming/chunking.rs` and `streaming/commit_tick.rs`
- add hysteresis-based catch-up behavior with bounded batch draining to reduce queue lag without bursty single-frame jumps
- document policy behavior, tuning guidance, and debug flow in rustdoc + docs

## Testing
- just fmt
- cargo test -p codex-tui

### DIFF
--- a/codex-rs/tui/src/tui/frame_rate_limiter.rs
+++ b/codex-rs/tui/src/tui/frame_rate_limiter.rs
@@ -1,7 +1,7 @@
 //! Limits how frequently frame draw notifications may be emitted.
 //!
 //! Widgets sometimes call `FrameRequester::schedule_frame()` more frequently than a user can
-//! perceive. This limiter clamps draw notifications to a maximum of 60 FPS to avoid wasted work.
+//! perceive. This limiter clamps draw notifications to a maximum of 120 FPS to avoid wasted work.
 //!
 //! This is intentionally a small, pure helper so it can be unit-tested in isolation and used by
 //! the async frame scheduler without adding complexity to the app/event loop.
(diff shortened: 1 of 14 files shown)


codex-rs/tui/src/tui/frame_rate_limiter.rs
```diff
@@ -1,16 +1,16 @@
...
 //!
 //! This is intentionally a small, pure helper so it can be unit-tested in isolation and used by
 //! the async frame scheduler without adding complexity to the app/event loop.
 
 use std::time::Duration;
 use std::time::Instant;
 
-/// A 60 FPS minimum frame interval (≈16.67ms).
-pub(super) const MIN_FRAME_INTERVAL: Duration = Duration::from_nanos(16_666_667);
+/// A 120 FPS minimum frame interval (≈8.33ms).
```
wiltzius-openai, line 12: you think its worth checking the actual hardware refresh rate vs always using 120? just curious

wiltzius-openai (approved):
very nice :)

codex-rs/tui/src/tui/frame_rate_limiter.rs
joshka-oai: NNot simple to do in a way that makes sense for this. We can probably make it user configurable is this results in high cpu (though there's other things we could do to drive this down first)

PR #10474, opened 2026-02-03
Ignore remote_compact_trims_function_call_history_to_fit_context_window on windows

pakrym-oai:


bolinfest (approved):

codex-rs/core/tests/suite/compact_remote.rs
```diff
@@ -222,6 +222,7 @@ async fn remote_compact_runs_automatically() -> Result<()> {
     Ok(())
 }
 
+#[cfg_attr(target_os = "windows", ignore)]
```
bolinfest, line 225: Hopefully this doesn't cause any unused import clippy warnings on Windows?

pakrym-oai: This just adds an attribute, it doesn't remove the entire method. In theory it shouldn't affect imports.

PR #10478, opened 2026-02-03
Added support for live updates to skills

etraut-openai:
Add a centralized FileWatcher in codex-core (using notify) that watches skill roots from the config layer stack (recursive)

Send `SkillsChanged` events when relevant file system changes are detected

On `SkillsChanged`:
* Invalidate the skills cache immediately in ThreadManager
* Emit EventMsg::SkillsUpdateAvailable to active sessions
~~* Broadcast a new app-server notification: SkillsListUpdatedNotification~~

This change does not inject new items into the event stream. That means the agent will not know about new skills, so it won't be able to implicitly invoke new skills. It also won't know about changes to existing skills, so if it has already read the contents of a modified skill, it will not honor the new behavior.

This change also does not detect modifications to AGENTS.md.

I plan to address these limitations in a follow-on PR modeled after #9985. Injection of new skills and AGENTS was deemed to risky, hence the need to split the feature into two stages. The changes in this PR were designed to easily accommodate the second stage once we have some other foundational changes in place.

Testing: In addition to automated tests, I did manual testing to confirm that newly-created skills, deleted skills, and renamed skills are reflected in the TUI skill picker menu. Also confirmed that modifications to behaviors for explicitly-invoked skills are honored.


etraut-openai:
@codex review

codex-rs/app-server-protocol/schema/json/v2/SkillsListUpdatedNotification.json
```diff
@@ -0,0 +1,5 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "SkillsListUpdatedNotification",
+  "type": "object"
+}
```
xl-openai, line 5: Add new line?

etraut-openai: This is an auto-generated file.

codex-rs/core/src/project_doc.rs
```diff
@@ -148,6 +148,31 @@ pub async fn read_project_docs(config: &Config) -> std::io::Result<Option<String
 /// directory (inclusive). Symlinks are allowed. When `project_doc_max_bytes`
 /// is zero, returns an empty list.
 pub fn discover_project_doc_paths(config: &Config) -> std::io::Result<Vec<PathBuf>> {
+    let search_dirs = project_doc_search_dirs(config)?;
```
xl-openai: Irrelevant change?

codex-rs/app-server/src/message_processor.rs
```diff
@@ -112,6 +114,7 @@ pub(crate) struct MessageProcessor {
     config: Arc<Config>,
     initialized: bool,
     experimental_api_enabled: Arc<AtomicBool>,
+    initialized_flag: Arc<AtomicBool>,
```
pakrym-oai: nit: maybe more descriptive name? requested_skills? send_skill_updated_events?

pakrym-oai: ArcSwap might be a good primitive here but not required.

codex-rs/core/src/skills/manager.rs
```diff
@@ -109,22 +107,22 @@ impl SkillsManager {
...
         outcome.disabled_paths = disabled_paths_from_stack(&config_layer_stack);
-        match self.cache_by_cwd.write() {
-            Ok(mut cache) => {
-                cache.insert(cwd.to_path_buf(), outcome.clone());
-            }
-            Err(err) => {
-                err.into_inner().insert(cwd.to_path_buf(), outcome.clone());
-            }
-        }
+        let mut cache = match self.cache_by_cwd.write() {
```
xl-openai, line 110: 👍

codex-rs/core/src/file_watcher.rs
```diff
@@ -0,0 +1,316 @@
...
+mod tests {
+    use super::*;
+    use pretty_assertions::assert_eq;
+
+    fn path(name: &str) -> PathBuf {
+        PathBuf::from(name)
+    }
+
+    #[test]
+    fn throttles_and_coalesces_within_interval() {
```
pakrym-oai, line 283: We might want a few more file watcher specific tests in the future. Not blocking.

codex-rs/app-server/src/message_processor.rs
```diff
@@ -156,6 +159,30 @@ impl MessageProcessor {
             auth_manager.clone(),
             SessionSource::VSCode,
         ));
+
+        // Watch for on-disk skill changes and send notifications to the client.
+        let initialized_flag = Arc::new(AtomicBool::new(false));
+        let mut skills_updates_rx = thread_manager.subscribe_file_watcher();
```
pakrym-oai: A little surprising we don't do subscription somewhere in core while the watcher lives in core. 

Feels a bit like a leaking abstraction where we create the watcher in core but then expose it as is and have the external consumer manage it. 

Have you considered adding a SkillsChanged core event instead?

codex-rs/core/tests/suite/live_reload.rs
```diff
@@ -0,0 +1,171 @@
...
+use core_test_support::responses::ResponsesRequest;
+use core_test_support::responses::mount_sse_sequence;
+use core_test_support::responses::start_mock_server;
+use core_test_support::test_codex::TestCodex;
+use core_test_support::test_codex::test_codex;
+use core_test_support::wait_for_event;
+use tokio::time::timeout;
+
+fn sse_completed(id: &str) -> String {
+    load_sse_fixture_with_id("../fixtures/completed_template.json", id)
```
pakrym-oai: prefer the existing shared `sse_completed` helper

codex-rs/app-server/src/message_processor.rs
pakrym-oai: Ah, you did add the event. Should we use it here instead of resubscribing to watcher directly?

codex-rs/core/tests/suite/live_reload.rs
```diff
@@ -0,0 +1,171 @@
...
+            });
+        assert!(
+            saw_expected_path,
+            "expected changed watcher path to include {expected_skill_path:?} or {expected_skill_dir:?}, got {changed_paths:?}"
+        );
+    } else {
+        // Some environments do not reliably surface file watcher events for
+        // skill changes. Clear the cache explicitly so we can still validate
+        // that the updated skill body is injected on the next turn.
+        test.thread_manager.skills_manager().clear_cache();
```
pakrym-oai, line 137: Is this test too weak?

codex-rs/core/tests/suite/live_reload.rs
```diff
@@ -0,0 +1,171 @@
...
+        .expect("first request captured");
+    assert!(
+        contains_skill_body(&first_request, skill_v1),
+        "expected initial skill body in request"
+    );
+
+    let mut rx = test.thread_manager.subscribe_file_watcher();
+    write_skill(test.codex_home_path(), "demo", "demo skill", skill_v2);
+
+    let changed_paths = timeout(Duration::from_secs(5), async move {
```
pakrym-oai: Should we assert EventMsg::SkillsUpdateAvailable event instead?

codex-rs/core/tests/suite/live_reload.rs
```diff
@@ -0,0 +1,171 @@
...
+        test.thread_manager.skills_manager().clear_cache();
+    }
+
+    submit_skill_turn(&test, skill_path.clone(), "please use $demo again").await?;
+    let last_request = responses
+        .last_request()
+        .expect("request captured after skill update");
+
+    assert!(
+        contains_skill_body(&last_request, skill_v2),
```
pakrym-oai, line 146: Surprised the skill is getting reinjected on next turn.

pakrym-oai (approved):
I think we should peek file watcher as code implementation detail and only expose/consume EventMsg::SkillsUpdateAvailable event.

codex-rs/core/src/file_watcher.rs
```diff
@@ -0,0 +1,316 @@
...
+            tx,
+        }
+    }
+
+    pub(crate) fn subscribe(&self) -> broadcast::Receiver<FileWatcherEvent> {
+        self.tx.subscribe()
+    }
+
+    pub(crate) fn register_config(&self, config: &Config) {
+        self.register_skills_root(config.codex_home.join("skills"));
```
xl-openai: skill_roots_from_layer_stack_with_agents should cover this already?

codex-rs/core/tests/suite/live_reload.rs
xl-openai: I don't see a shared sse_completed helper?

codex-rs/core/src/codex.rs
```diff
@@ -959,6 +988,9 @@ impl Session {
             sess.send_event_raw(event).await;
         }
 
+        // Start the watcher after SessionConfigured so it cannot emit earlier events.
+        sess.start_file_watcher_listener();
```
xl-openai, line 992: So we will still emit EventMsg::SkillsUpdateAvailable (for each session) for app-server clients right?

codex-rs/core/tests/suite/live_reload.rs
pakrym-oai: responses::ev_completed("resp-1"),

codex-rs/core/tests/suite/live_reload.rs
xl-openai: The submit_skill_turn explicitly mentioned the skill again. Currently we do not dedupe pre-injection if you mention the same skills multiple times in a thread.

codex-rs/core/src/file_watcher.rs
xl-openai: add basic tests on util functions.

codex-rs/core/tests/suite/live_reload.rs
xl-openai: I can’t find a way to make this behave consistently. I agree that if we swallow the error, this test becomes largely meaningless.

PR #10508, opened 2026-02-03
Add more detail to 401 error

gt-oai:
Add the error.message if it exists, the body otherwise. Truncate body to 1k characters. Print the cf-ray and the requestId.

**Before:**
Screenshot 2026-02-03 at 13 15 28

**After:**
Screenshot 2026-02-03 at 13 15 38

### DIFF
--- a/codex-rs/core/src/api_bridge.rs
+++ b/codex-rs/core/src/api_bridge.rs
@@ -28,6 +28,7 @@ pub(crate) fn map_api_error(err: ApiError) -> CodexErr {
             status,
             body: message,
             url: None,
+            cf_ray: None,
             request_id: None,
         }),
         ApiError::InvalidRequest { message } => CodexErr::InvalidRequest(message),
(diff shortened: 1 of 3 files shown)


codex-rs/core/src/api_bridge.rs
```diff
@@ -149,15 +151,23 @@ mod tests {
...
 
+fn extract_request_tracking_id(headers: Option<&HeaderMap>) -> Option<String> {
+    extract_request_id(headers).or_else(|| extract_cf_ray(headers))
+}
+
 fn extract_request_id(headers: Option<&HeaderMap>) -> Option<String> {
+    extract_header(headers, "x-request-id").or_else(|| extract_header(headers, "x-oai-request-id"))
+}
+
+fn extract_cf_ray(headers: Option<&HeaderMap>) -> Option<String> {
```
jif-oai: can we inline?

codex-rs/core/src/api_bridge.rs
```diff
@@ -149,15 +151,23 @@ mod tests {
...
+    extract_header(headers, "cf-ray")
```
jif-oai: I feel those `cf-ray`, `x-request-id`, `x-oai-request-id` should be const somewhere

codex-rs/core/src/error.rs
```diff
@@ -335,6 +370,23 @@ impl std::fmt::Display for UnexpectedResponseError {
 }
 
 impl std::error::Error for UnexpectedResponseError {}
+
+fn truncate_with_ellipsis(text: &str, max_chars: usize) -> String {
```
jif-oai: Sure we don't already have some truncation code that fits the need somewhere? I feel we do truncation everywhere
Also, I'm not sure the truncation should be done at this layer, IMO it should be in the client

jif-oai (approved):

PR #10524, opened 2026-02-03
fix: make $PWD/.agents read-only like $PWD/.codex

bolinfest:
In light of https://github.com/openai/codex/pull/10317, because `.agents` can include resources that Codex can run in a privileged way, it should be read-only by default just as `.codex` is.


viyatb-oai (approved):
lgtm that .agent/skills should not be writable by default.

PR #10527, opened 2026-02-03
feat(app-server, skills): add a way for clients to specify additional dirs for skills

owenlin0:
## Summary
This PR adds multi-root skill discovery support to the app-server APIs, so clients can provide extra skill directories without relying on symlinks or client-side copying.

The feature is added to these RPCs:
- `skills/list` (discovery-time roots)
- `turn/start` (runtime roots used for skill mention resolution in this and subsequent turns)

## Motivation
Some app-server clients need to source skills from multiple locations. 

We _could_ do this using the existing primitives, where there's a single canonical directory (`$CODEX_HOME/skills`) and the client can manually create symlinks before starting up `codex app-server`. However, explicit additional roots in API requests feels like a cleaner long-term model:
- fewer edge cases around clobbering/conflicts
- clearer ownership and policy boundaries per root
- no symlink management requirements

I chose not to do anything around config.toml for now to limit the blast radius of this change.

## Behavior Semantics
For both endpoints:
- if additionalRoots is unspecified, `null`, or `[]` => no additional roots

For `turn/start` specifically:
- `None` => do not change current per-thread additional roots
- `Some(vec![...])` => replace roots for this and subsequent turns
- `Some(vec![])` => clear previously-set additional roots

## Implementation Notes
- `skills/list` now passes additionalRoots straight to skill loading as path vectors.
- `turn/start` now forwards provided roots via a core op (SetSkillRoots) before submitting user input.
- Core session/turn context now carries `additional_roots` so turn-time skill resolution uses them consistently.
- Review turns inherit the same additional roots from parent turn context.


leoshimo-oai:
Q: Does this work with the hotreload PR?
https://github.com/openai/codex/pull/9985/

codex-rs/core/src/skills/manager.rs
```diff
@@ -120,6 +122,66 @@ impl SkillsManager {
...
+        additional_roots: &[PathBuf],
+    ) -> SkillLoadOutcome {
+        if additional_roots.is_empty() {
+            return self.skills_for_cwd(cwd, force_reload).await;
+        }
+
+        let cwd_abs = match AbsolutePathBuf::try_from(cwd) {
+            Ok(cwd_abs) => cwd_abs,
+            Err(err) => {
+                return SkillLoadOutcome {
```
xl-openai, line 142: Should we fail-open to skill loading issues? Or can you confirm this won't break the client?

owenlin0: yes will confirm!

PR #10536, opened 2026-02-03
Revert "Load untrusted rules"

viyatb-oai:
Reverts openai/codex#9791


gt-oai (approved):
consider a less pure revert that keeps the test but inverts the condition. but this is fine.

PR #10096, opened 2026-01-28
Inject CODEX_THREAD_ID into the terminal environment

maxj-oai:
Inject CODEX_THREAD_ID (when applicable) into the terminal environment so that the agent (and skills) can refer to the current thread / session ID.

Discussion: https://openai.slack.com/archives/C095U48JNL9/p1769542492067109

PR #10538, opened 2026-02-03
Add CODEX_SESSION_ID env var and inject into shell tool environments

abarron-oai:
### Motivation
- Ensure processes launched by the CLI (shell tool calls) can discover their originating Codex session id via an environment variable for improved traceability and tooling integration.

### Description
- Define `CODEX_SESSION_ID_ENV_VAR` in `core/src/exec_env.rs` and inject the active `session.conversation_id` into shell tool invocation env maps in `core/src/tools/handlers/shell.rs` using `.entry(...).or_insert_with(...)` so existing values are preserved.

### Testing
- Ran `just fmt` in `codex-rs`, which completed successfully. 
- Ran `cargo test -p codex-core`; the test run executed and produced `857 passed, 10 failed` with failures appearing in `default_client::tests::test_create_client_sets_default_headers`, several `models_manager::manager` tests, some `shell_snapshot` tests, and a few `unified_exec` tests, which appear related to network/shell behavior in the test environment rather than this small env-var change.

### DIFF
--- a/codex-rs/core/src/exec_env.rs
+++ b/codex-rs/core/src/exec_env.rs
@@ -4,6 +4,8 @@ use crate::config::types::ShellEnvironmentPolicyInherit;
 use std::collections::HashMap;
 use std::collections::HashSet;
 
+pub const CODEX_SESSION_ID_ENV_VAR: &str = "CODEX_SESSION_ID";
+
 /// Construct an environment map based on the rules in the specified policy. The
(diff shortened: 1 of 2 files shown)


etraut-openai:
I think this is a duplicate of https://github.com/openai/codex/pull/10096, which was just merged.

abarron-oai:
Awesome! Closed.

PR #10546, opened 2026-02-03
feat(tui): add /statusline command for interactive status line configuration

fcoury:
## Summary
- Adds a new `/statusline` command to configure TUI footer status line
- Introduces reusable `MultiSelectPicker` component with keyboard navigation, optional ordering and toggle support
- Implement status line setup modal that persist configuration to config.toml

  ## Status Line Items
  The following items can be displayed in the status line:
  - **Model**: Current model name (with optional reasoning level)
  - **Context**: Remaining/used context window percentage
  - **Rate Limits**: 5-day and weekly usage limits
  - **Git**: Current branch (with optimized lookups)
  - **Tokens**: Used tokens, input/output token counts
  - **Session**: Session ID (full or shortened prefix)
  - **Paths**: Current directory, project root
  - **Version**: Codex version

  ## Features
  - Live preview while configuring status line items
  - Fuzzy search filtering in the picker
  - Intelligent truncation when items don't fit
  - Items gracefully omit when data is unavailable
  - Configuration persists to `config.toml`
  - Validates and warns about invalid status line items

  ## Test plan
  - [x] Run `/statusline` and verify picker UI appears
  - [x] Toggle items on/off and verify live preview updates
  - [x] Confirm selection persists after restart
  - [x] Verify truncation behavior with many items selected
  - [x] Test git branch detection in and out of git repos


etraut-openai:
@codex review

codex-rs/tui/src/chatwidget.rs
```diff
@@ -3755,6 +3905,205 @@ impl ChatWidget {
...
+            .as_ref()
+            .and_then(|info| info.model_context_window)
+            .or(self.config.model_context_window)
+    }
+
+    fn status_line_context_remaining_percent(&self) -> Option<i64> {
+        let Some(context_window) = self.status_line_context_window_size() else {
+            return Some(100);
+        };
+        let default_usage = TokenUsage::default();
```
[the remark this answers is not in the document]
fcoury: @copilot compare this behavior to what is done when the status line is disabled and we show (num)% context left on the right.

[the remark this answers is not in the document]
fcoury: Why you see “100% context left” on the right-hand corner at startup

- That text is not coming from status_line_context_remaining_percent.
- It comes from context_window_line() in codex-rs/tui/src/bottom_pane/footer.rs, which defaults to "100% context left" when both percent and used_tokens are None.
- percent is set via ChatWidget::apply_token_info() (based on TokenUsageInfo.model_context_window). At startup, that is often None, so the footer fallback kicks in.

So the behavior you saw is consistent with the current footer logic,

codex-rs/tui/src/chatwidget.rs
```diff
@@ -3755,6 +3905,205 @@ impl ChatWidget {
...
+        self.token_info
+            .as_ref()
+            .and_then(|info| info.model_context_window)
+            .or(self.config.model_context_window)
+    }
+
+    fn status_line_context_remaining_percent(&self) -> Option<i64> {
+        let Some(context_window) = self.status_line_context_window_size() else {
+            return Some(100);
+        };
```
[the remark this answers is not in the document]
fcoury: This was addressed before: we are replicating the same behavior of the current 100% context left display.

etraut-openai:
@codex review

codex-rs/tui/src/bottom_pane/multi_select_picker.rs
```diff
@@ -0,0 +1,796 @@
...
+///
+/// # Returns
+///
+/// * `Some((Some(indices), score))` - Matched on display name with highlight indices
+/// * `Some((None, score))` - Matched on skill name only (no highlights for display)
+/// * `None` - No match
+pub(crate) fn match_item(
+    filter: &str,
+    display_name: &str,
+    skill_name: &str,
```
etraut-openai: I was initially confused by the parameter name `skill_name` because it doesn't have anything to do with "skills". Maybe go with something more general?

codex-rs/tui/src/bottom_pane/status_line_setup.rs
```diff
@@ -0,0 +1,285 @@
...
+        match self {
+            StatusLineItem::ModelName => "gpt-5.2-codex",
+            StatusLineItem::ModelWithReasoning => "gpt-5.2-codex medium",
+            StatusLineItem::CurrentDir => "~/project/path",
+            StatusLineItem::ProjectRoot => "~/project",
+            StatusLineItem::GitBranch => "feat/awesome-feature",
+            StatusLineItem::ContextRemaining => "18% left",
+            StatusLineItem::ContextUsed => "82% used",
+            StatusLineItem::FiveHourLimit => "5h 100%",
+            StatusLineItem::WeeklyLimit => "7d 98%",
```
etraut-openai: This differs from what's actually rendered in this case. It appears as "weekly". Should be consistent. I don't have a strong opinion here, but "7d" is more consistent with "5h".

codex-rs/tui/src/chatwidget.rs
```diff
@@ -3780,6 +3948,215 @@ impl ChatWidget {
         ));
     }
 
+    fn open_status_line_setup(&mut self) {
```
etraut-openai, line 3983: The chatwidget module is getting really large. I wonder if we should move the new functionality to a different module rather than here. We could do this as a separate cleanup PR, so I consider this non-blocking.

etraut-openai (commented):
Nice! This looks really good. I appreciate the attention to detail. Nicely commented, good error handling, good structure. 

I added a few really minor suggestions, but I think this is nearly ready to go. I'm going to see if joshka wants to do a quick review.

codex-rs/tui/src/bottom_pane/multi_select_picker.rs
fcoury: I extracted this whole module from SkillsToggleView, so this was a copy and paste oversight, thank you for catching it!

codex-rs/tui/src/bottom_pane/status_line_setup.rs
fcoury: I kept it weekly since we are using `get_limits_duration` that is already used in other parts of the code.

codex-rs/tui/src/chatwidget.rs
fcoury: I agree with it getting extremely long but I also would prefer to address it on a separate PR, since we have more code to extract and this function in particular is only 7 lines long. I'll keep it for now if you agree.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -3780,6 +3948,215 @@ impl ChatWidget {
...
+        self.token_info
+            .as_ref()
+            .and_then(|info| info.model_context_window)
+            .or(self.config.model_context_window)
+    }
+
+    fn status_line_context_remaining_percent(&self) -> Option<i64> {
+        let Some(context_window) = self.status_line_context_window_size() else {
+            return Some(100);
+        };
```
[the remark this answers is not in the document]
fcoury: This is by design and matches the behavior of other areas of code that also falls back to 100%.

joshka-oai:
## problem
The TUI footer needed a user-configurable status line so people can choose which
session/runtime signals matter to them without hardcoding one fixed layout.
The previous setup also lacked a dedicated configuration flow and clear
handling for unavailable values (for example, branch/session-dependent data).

## mental model
`/statusline` opens a picker that defines an ordered list of item ids persisted
in config. At render time, `ChatWidget` treats that list as intent, resolves
each item against current runtime data, and omits items that are temporarily
unavailable. Git branch is resolved asynchronously and cached per cwd so the
footer can stay responsive while still updating when branch data arrives.

## non-goals
This change does not introduce custom external status-line commands, does not
guarantee all selected items are always visible at all times, and does not
redesign the footer collapse/truncation system beyond integrating status-line
output with existing width constraints.

## tradeoffs
The design prefers omission over placeholders when data is missing, which keeps
the footer compact but can make item absence look implicit. Branch lookup is
async and guarded against stale cwd updates, which improves responsiveness but
adds state tracking complexity. Invalid configured item ids are tolerated and
warned once rather than hard-failing startup.

## architecture
The feature is split across reusable and feature-specific layers:
`MultiSelectPicker` provides searchable, reorderable multi-select UI;
`StatusLineSetupView` maps picker selection to `StatusLineItem`; `AppEvent`
carries setup/branch update events; `ChatWidget` owns status-line orchestration
(config parsing, warning policy, branch lookup lifecycle, item rendering); and
status helpers shape rate-limit/context/token values for display. Footer
rendering remains the presentation layer and consumes the resulting line.

## observability
User-visible warnings are emitted for invalid configured item ids (once per
session after thread start). Setup confirm/cancel paths emit info logs. Branch
updates flow through `StatusLineBranchUpdated` events, and stale async branch
results are ignored when cwd no longer matches.

## tests
Validated with `just fmt` and `cargo test -p codex-tui` (full suite,
including integration/doc tests for the crate). Existing snapshot-framework
legacy-format notices remain informational and did not indicate behavioral
regressions.

## design inconsistencies / risks
- PR text says "5-day" in one place, but implementation and labels are
  `5h`/"5-hour".
- Invalid status-line ids are only warned after session/thread exists;
  pre-session misconfig can stay silent.
- `tui.status_line` remains free-form string ids at config boundary, so
  correctness is mostly runtime-validated rather than schema-enforced.

joshka-oai (commented):
Things I noted in manual testing.
- Wasn't clear how to reorder
- session id prefix shouldn't be there (we use the full id for linking feedback right now and resuming)
- wasn't sure about the project item
- used / avail tokens need some extra text (just numbers)

I pushed some extra doc comments for future maintenance

Otherwise LGTM

fcoury:
> Things I noted in manual testing.

> * Wasn't clear how to reorder

I added instructions to the footer, but open to other ideas.

> * session id prefix shouldn't be there (we use the full id for linking feedback right now and resuming)

I'll take it out.

> * wasn't sure about the project item

Do you mean project-root? That's the root folder of the current project. It's a short and canonical name for the project. I added it because, in my case, when I am working with paths outside the home folder using cwd gets extremely long (see below). I always always had this single identifier when using other harnesses and they worked as a fast and reliable way to spot which project and/or worktree I was working on.

image

See codex-status-line-builder vs /Volumes/External/code-external/codex-status-line-builder/codex-rs.

> * used / avail tokens need some extra text (just numbers)

Agreed. I will use "used" and "avail" for shortness. Any concerns?

> I pushed some extra doc comments for future maintenance

I got them, thank you!

OK one question, I really really liked the other comment you sent before with the problem, mental model, etc. and I have to ask: do you use AI to generate those? I would love to get something similar if you can share.

Other than that, thank you very much for the review!

fcoury:
> * used / avail tokens need some extra text (just numbers)

Actually the context used already had text, just added to all the others that were missing:

image

codex-rs/tui/src/chatwidget.rs
joshka-oai: The general approach I've been using here is to instruct codex not to add code to god ~classes~ structs, and instead generally add new modules and extract code. It usually does pretty good here - there's a few large modules like this (chat_widget, app, tui, history_cells) that need a clean up.

joshka-oai:
> OK one question, I really really liked the other comment you sent before with the problem, mental model, etc. and I have to ask: do you use AI to generate those? I would love to get something similar if you can share.

https://gist.github.com/joshka-oai/4262dc0a09067971738ccea774449ded

I use it for my own work and reviewing other work. Add comments either with your context and history while developing with the model's Chain of Thought available, or after without it and inferred. It's not perfect, but it's got a lot of value in accelerating from idkwtf to tillgtm

joshka-oai:
> Do you mean project-root? That's the root folder of the current project. It's a short and canonical name for the project. I added it because, in my case, when I am working with paths outside the home folder using cwd gets extremely long (see below). I always always had this single identifier when using other harnesses and they worked as a fast and reliable way to spot which project and/or worktree I was working on.

yep - for some reason I wasn't seeing any text on this when I tested - likely because I usually use jujutsu not git

PR #10549, opened 2026-02-03
feat(core): wire secrets backend config

viyatb-oai:
## Summary

Adds TOML config support for:

```
[secrets]
backend = "local" (only value today)
```

and maps that into runtime config as Config.secrets_backend. It defaults to `SecretsBackendKind::Local` when omitted.

We already have a local encrypted secrets backend in `codex-secrets`.  This PR wires that backend into `codex-core` so the runtime can construct and carry a `SecretsManager` through normal config loading.


codex-rs/core/src/config/types.rs
```diff
@@ -24,6 +25,13 @@ use serde::de::Error as SerdeError;
 
 pub const DEFAULT_OTEL_ENVIRONMENT: &str = "dev";
 
+#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq, Eq, JsonSchema)]
+#[schemars(deny_unknown_fields)]
+pub struct SecretsConfigToml {
```
xl-openai: Let's don't introduce this config yet (until we are ready to introduce more backend store).

viyatb-oai: I need this for non-skill stuff :)

xl-openai:
If LOCAL is the only supported secrets_backend at this moment, I’d recommend hard-coding the value instead of introducing a new user-facing config field.

PR #10551, opened 2026-02-03
feat(cli): add codex secrets commands

viyatb-oai:
## Summary

Builds on PR B’s core secrets config wiring by adding a user-facing CLI:

- `codex secrets set`
- `codex secrets edit`
- `codex secrets list`
- `codex secrets delete`

Uses `Config.secrets_backend` and supports scopes:

- `--global`
- `--env <id>`
- default env scope from cwd

Values can be provided via `--value`, stdin, or interactive masked prompt.

## Notes

- `list` never prints secret values
- `edit` errors if secret is missing
- `delete` is no-op if missing


codex-rs/cli/src/secrets_cmd.rs
```diff
@@ -0,0 +1,371 @@
...
+    std::io::stdin()
+        .read_to_string(&mut buf)
+        .context("failed to read secret value from stdin")?;
+    let trimmed = buf.trim_end_matches(['\n', '\r']).to_string();
+    anyhow::ensure!(!trimmed.is_empty(), "secret value must not be empty");
+    Ok(trimmed)
+}
+
+fn prompt_secret_value(display_name: &str) -> Result<String> {
+    print!("Enter value for {display_name}: ");
```
xl-openai: Let's add something suggesting this is experimental (for now). We could clean everything up once we think it is ready for release.

codex-rs/cli/src/secrets_cmd.rs
```diff
@@ -0,0 +1,371 @@
...
+}
+
+fn resolve_scope(config: &Config, scope_args: &SecretsScopeArgs) -> Result<SecretScope> {
+    if scope_args.global {
+        return Ok(SecretScope::Global);
+    }
+    if let Some(env_id) = scope_args.environment_id.as_deref() {
+        return SecretScope::environment(env_id.to_string());
+    }
+    default_environment_scope(config)
```
xl-openai: I feel default to global makes more sense.

xl-openai (approved):

codex-rs/cli/src/secrets_cmd.rs
```diff
@@ -0,0 +1,370 @@
...
+    pub name: String,
+
+    #[clap(flatten)]
+    pub scope: SecretsScopeArgs,
+}
+
+impl SecretsCli {
+    pub async fn run(self) -> Result<()> {
+        let config = load_config(self.config_overrides).await?;
+        let manager = SecretsManager::new(config.codex_home.clone(), config.secrets_backend);
```
xl-openai: You can set it directly to LOCAL without relying on a config.

PR #10567, opened 2026-02-03
feat(app-server, core): allow text + image content items for dynamic tool outputs

owenlin0:
Took over the work that @aaronl-openai started here: https://github.com/openai/codex/pull/10397

Now that app-server clients are able to set up custom tools (called `dynamic_tools` in app-server), we should expose a way for clients to pass in not just text, but also image outputs. This is something the Responses API already supports for function call outputs, where you can pass in either a string or an array of content outputs (text, image, file): https://platform.openai.com/docs/api-reference/responses/create#responses_create-input-input_item_list-item-function_tool_call_output-output-array-input_image

So let's just plumb it through in Codex (with the caveat that we only support text and image for now). This is implemented end-to-end across app-server v2 protocol types and core tool handling.

## Breaking API change
NOTE: This introduces a breaking change with dynamic tools, but I think it's ok since this concept was only recently introduced (https://github.com/openai/codex/pull/9539) and it's better to get the API contract correct. I don't think there are any real consumers of this yet (not even the Codex App).

Old shape:
`{ "output": "dynamic-ok", "success": true }`

New shape:
```
{
    "contentItems": [
      { "type": "inputText", "text": "dynamic-ok" },
      { "type": "inputImage", "imageUrl": "data:image/png;base64,AAA" }
    ]
  "success": true
}
```


owenlin0:
@codex review

codex-rs/app-server-protocol/schema/json/v1/ForkConversationResponse.json
```diff
@@ -2864,6 +2811,19 @@
         }
```
jif-oai, line 2864: I wonder if we should not generate all those files at runtime... this is quite poluting

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2814,10 +2775,37 @@ pub struct DynamicToolCallParams {
...
+pub enum DynamicToolCallResult {
+    /// Preferred structured tool output (for example text + images) that is
+    /// forwarded directly to the model as content items.
+    #[serde(rename_all = "camelCase")]
+    ContentItems {
+        content_items: Vec<DynamicToolCallOutputContentItem>,
+    },
+    /// Plain-text tool output.
+    #[serde(rename_all = "camelCase")]
+    Output { output: String },
```
jif-oai: Why keeping it if this is a breaking change anyway?

codex-rs/app-server/src/dynamic_tools.rs
```diff
@@ -56,3 +72,16 @@ pub(crate) async fn on_call_response(
         error!("failed to submit DynamicToolResponse: {err}");
     }
 }
+
+fn map_content_item(
```
jif-oai: Should be an `impl From` in the correct location (ignore if this creates a circular dependency)

codex-rs/core/src/tools/handlers/dynamic.rs
```diff
@@ -55,10 +58,21 @@ impl ToolHandler for DynamicToolHandler {
                 )
             })?;
 
+        let DynamicToolResponse {
+            result, success, ..
+        } = response;
+        let (content, content_items) = match result {
+            DynamicToolResult::Output { output } => (output, None),
+            DynamicToolResult::ContentItems { content_items } => (
+                content_items_to_text(Some(&content_items)).unwrap_or_default(),
```
jif-oai: that is pretty strange. It means you would have both `content` and `content_items` non-null and potentially containing the same information

codex-rs/core/src/tools/handlers/dynamic.rs
```diff
@@ -96,3 +110,73 @@ async fn request_dynamic_tool(
...
+    }
+
+    if text.is_empty() {
+        None
+    } else {
+        Some(text.join("\n"))
+    }
+}
+
+fn map_content_item(item: DynamicToolCallOutputContentItem) -> FunctionCallOutputContentItem {
```
jif-oai: impl From

codex-rs/core/src/tools/handlers/dynamic.rs
```diff
@@ -96,3 +110,73 @@ async fn request_dynamic_tool(
...
+            && !segment.trim().is_empty()
+        {
+            text.push(segment.as_str());
+        }
+    }
+
+    if text.is_empty() {
+        None
+    } else {
+        Some(text.join("\n"))
```
jif-oai: this sounds like a heuristic... is it really how we should do it?

codex-rs/protocol/src/dynamic_tools.rs
```diff
@@ -25,6 +25,28 @@ pub struct DynamicToolCallRequest {
...
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, JsonSchema, TS)]
+#[serde(tag = "type", rename_all = "camelCase")]
+#[ts(tag = "type")]
+pub enum DynamicToolResult {
+    #[serde(rename_all = "camelCase")]
+    ContentItems {
+        content_items: Vec<DynamicToolCallOutputContentItem>,
+    },
+    #[serde(rename_all = "camelCase")]
+    Output { output: String },
```
jif-oai: Same as before, drop output if you plan to break the compatibility anyway

jif-oai (changes requested):

codex-rs/app-server-protocol/src/protocol/v2.rs
owenlin0: good call! i will remove

codex-rs/core/src/tools/handlers/dynamic.rs
owenlin0: this is used for logging mainly, via `ToolOutput::log_preview`

owenlin0: This never gets sent to the model

codex-rs/app-server-protocol/schema/json/v1/ForkConversationResponse.json
owenlin0: v1 will go away soon so this should be less annoying hopefully

owenlin0:
@codex review

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2814,10 +2775,34 @@ pub struct DynamicToolCallParams {
...
-    pub output: String,
+    pub content_items: Vec<DynamicToolCallOutputContentItem>,
     pub success: bool,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(tag = "type", rename_all = "camelCase")]
+#[ts(tag = "type")]
+#[ts(export_to = "v2/")]
+pub enum DynamicToolCallOutputContentItem {
```
jif-oai, line 2837: Shouldn't we re-use what already exist if it takes this shape? Feel free to ignore

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2814,10 +2775,34 @@ pub struct DynamicToolCallParams {
...
+    #[serde(rename_all = "camelCase")]
+    InputText { text: String },
+    #[serde(rename_all = "camelCase")]
+    InputImage { image_url: String },
+}
+
+impl From<DynamicToolCallOutputContentItem>
```
jif-oai, line 2844: ty

codex-rs/app-server/README.md
```diff
@@ -537,6 +535,39 @@ Order of messages:
 
 UI guidance for IDEs: surface an approval dialog as soon as the request arrives. The turn will proceed after the server receives a response to the approval request. The terminal `item/completed` notification will be sent with the appropriate status.
 
+### Dynamic tool calls
```
jif-oai: Maybe good to say that this needs the experimental flag

codex-rs/core/src/tools/handlers/mcp.rs
```diff
@@ -57,11 +57,18 @@ impl ToolHandler for McpHandler {
...
-                    success,
-                } = output;
+                let success = output.success;
+                let (content, content_items) = match output.body {
+                    codex_protocol::models::FunctionCallOutputBody::Text(content) => {
+                        (content, None)
+                    }
+                    codex_protocol::models::FunctionCallOutputBody::ContentItems(content_items) => {
+                        (
+                            serde_json::to_string(&content_items).unwrap_or_default(),
```
jif-oai: Why here do you use a serialization and above you use `content_items_to_text`

codex-rs/protocol/src/dynamic_tools.rs
```diff
@@ -25,6 +25,26 @@ pub struct DynamicToolCallRequest {
...
     pub call_id: String,
-    pub output: String,
+    pub result: DynamicToolResult,
     pub success: bool,
 }
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, JsonSchema, TS)]
+#[serde(tag = "type", rename_all = "camelCase")]
+#[ts(tag = "type")]
+pub enum DynamicToolResult {
```
jif-oai: Why do we still need an enum if there is only one choice which is an enum itself

codex-rs/core/src/tools/handlers/dynamic.rs
jif-oai: Huh? Content is used here: https://github.com/openai/codex/blob/7a253076fef8058adb7964fdd521df80663cf39d/codex-rs/core/src/tools/context.rs#L97

codex-rs/app-server-protocol/src/protocol/v2.rs
owenlin0: yeah we could just because it happens to be the exact same in this case, but continued following the pattern of having app-server own its exposed types

codex-rs/core/src/tools/handlers/dynamic.rs
owenlin0: so the code is super confusing actually, but it wasn't hitting that condition. i refactored the code so it should be clearer

owenlin0:
@codex review

codex-rs/app-server/src/dynamic_tools.rs
```diff
@@ -56,3 +64,10 @@ pub(crate) async fn on_call_response(
         error!("failed to submit DynamicToolResponse: {err}");
     }
 }
+
+fn fallback_core_response(text: String) -> CoreDynamicToolResponse {
```
jif-oai: This was inline and is not anymore. Any reason? Otherwise please inline it

codex-rs/core/src/tools/context.rs
```diff
@@ -53,15 +53,21 @@ impl ToolPayload {
             ToolPayload::Mcp { raw_arguments, .. } => Cow::Borrowed(raw_arguments),
         }
     }
+
+    pub fn outputs_custom_response(&self) -> bool {
```
jif-oai: inline for such a small function

codex-rs/core/src/tools/context.rs
```diff
@@ -115,6 +118,10 @@ impl ToolOutput {
     }
 }
 
+fn tool_output_text(body: &FunctionCallOutputBody) -> String {
```
jif-oai: drop this function

jif-oai (approved):
Way better, thanks

General comment: models love to create tons of small functions but I think most of the time this is a bad idea because:
* If it's single usage, it does not reduce the amount of code
* It makes the code less readable by human because we need to navigate the code quite a lot
* It makes the code less efficiently readable by AI because they need extra tool call to find the relevant code

PR #10568, opened 2026-02-04
Requirements: add source to constrained requirement values

gt-oai:
If we want to build `/debug-config`, we'll need to know the requirements sources that supplied the values.

This PR adds those sources such that we can render them in the UI.


jif-oai:
@codex review

jif-oai (approved):

PR #10569, opened 2026-02-04
tui: make Esc clear request_user_input notes while notes are shown

charley-oai:
## Summary

This PR updates the `request_user_input` TUI overlay so `Esc` is context-aware:

- When notes are visible for an option question, `Esc` now clears notes and exits notes mode.
- When notes are not visible (normal option selection UI), `Esc` still interrupts as before.

It also updates footer guidance text to match behavior.

## Changes

- Added a shared notes-clear path for option questions:
  - `Tab` and `Esc` now both clear notes and return focus to options when notes are visible.
- Updated footer hint text in notes-visible state:
  - from: `tab to clear notes | ... | esc to interrupt`
  - to: `tab or esc to clear notes | ...`
- Hid `esc to interrupt` hint while notes are visible for option questions.
- Kept `esc to interrupt` visible and functional in normal option-selection mode.
- Updated tests to assert the new `Esc` behavior in notes mode.
- Updated snapshot output for the notes-visible footer row.
- Updated docs in `docs/tui-request-user-input.md` to reflect mode-specific `Esc` behavior.

### DIFF
--- a/codex-rs/tui/src/bottom_pane/request_user_input/mod.rs
+++ b/codex-rs/tui/src/bottom_pane/request_user_input/mod.rs
@@ -427,8 +427,8 @@ impl RequestUserInputOverlay {
             if self.selected_option_index().is_some() && !notes_visible {
                 tips.push(FooterTip::highlighted("tab to add notes"));
             }
-            if self.selected_option_index().is_some() && notes_visible && self.focus_is_notes() {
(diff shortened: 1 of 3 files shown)


charley-oai:
@codex review this

aibrahim-oai (approved):

PR #10574, opened 2026-02-04
core: refresh developer instructions after compaction replacement history

charley-oai:
## Summary

When replaying compacted history (especially `replacement_history` from remote compaction), we should not keep stale developer messages from older session state. This PR trims developer-
role messages from compacted replacement history and reinjects fresh developer instructions derived from current turn/session state.

This aligns compaction replay behavior with the intended "fresh instructions after summary" model.

## Problem

Compaction replay had two paths:

- `Compacted { replacement_history: None }`: rebuilt with fresh initial context
- `Compacted { replacement_history: Some(...) }`: previously used raw replacement history as-is

The second path could carry stale developer instructions (permissions/personality/collab-mode guidance) across session changes.

### 1) Added helper to refresh compacted developer instructions

- **File:** `codex-rs/core/src/compact.rs`
- **Function:** `refresh_compacted_developer_instructions(...)`

Behavior:
- remove all `ResponseItem::Message { role: "developer", .. }` from compacted history
- append fresh developer messages from current `build_initial_context(...)`

### 2) Applied helper in remote compaction flow

- **File:** `codex-rs/core/src/compact_remote.rs`
- After receiving compact endpoint output, refresh developer instructions before replacing history and persisting `replacement_history`.

### 3) Applied helper while reconstructing history from rollout

- **File:** `codex-rs/core/src/codex.rs`
- In `reconstruct_history_from_rollout(...)`, when processing `Compacted` entries with `replacement_history`, refresh developer instructions instead of directly replacing with raw history.

## Non-Goals / Follow-up

This PR does **not** address the existing first-turn-after-resume double-injection behavior.
A follow-up PR will handle resume-time dedup/idempotence separately.

If you want, I can also give you a shorter “squash-merge friendly” version of the description.

## Codex author
`codex fork 019c25e6-706e-75d1-9198-688ec00a8256`

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -1768,11 +1768,10 @@ impl Session {
                 }
                 RolloutItem::Compacted(compacted) => {
                     if let Some(replacement) = &compacted.replacement_history {
-                        let initial_context = self.build_initial_context(turn_context).await;
-                        history.replace(compact::refresh_compacted_developer_instructions(
-                            replacement.clone(),
-                            &initial_context,
-                        ));
+                        history.replace(
+                            self.process_compacted_history(turn_context, replacement.clone())
+                                .await,
+                        );
                     } else {
                         let user_messages = collect_user_messages(history.raw_items());
                         let rebuilt = compact::build_compacted_history(
@@ -1790,6 +1789,15 @@ impl Session {
             }
         }
         history.raw_items().to_vec()
+    }
+
+    pub(crate) async fn process_compacted_history(
+        &self,
+        turn_context: &TurnContext,
+        compacted_history: Vec<ResponseItem>,
+    ) -> Vec<ResponseItem> {
+        let initial_context = self.build_initial_context(turn_context).await;
+        compact::process_compacted_history(compacted_history, &initial_context)
     }
 
     /// Append ResponseItems to the in-memory conversation history only.
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -4679,18 +4679,19 @@ mod tests {
     #[tokio::test]
     async fn reconstruct_history_refreshes_developer_instructions_for_replacement_history() {
         let (session, turn_context) = make_session_and_context().await;
+        let summary_item = ResponseItem::Message {
+            id: None,
+            role: "user".to_string(),
+            content: vec![ContentItem::InputText {
+                text: "summary".to_string(),
+            }],
+            end_turn: None,
+            phase: None,
+        };
         let rollout_items = vec![RolloutItem::Compacted(CompactedItem {
             message: String::new(),
             replacement_history: Some(vec![
-                ResponseItem::Message {
-                    id: None,
-                    role: "user".to_string(),
-                    content: vec![ContentItem::InputText {
-                        text: "summary".to_string(),
-                    }],
-                    end_turn: None,
-                    phase: None,
-                },
+                summary_item.clone(),
                 ResponseItem::Message {
                     id: None,
                     role: "developer".to_string(),
@@ -4721,27 +4722,9 @@ mod tests {
         });
         assert!(!has_stale_message);
 
-        let summary_position = reconstructed.iter().position(|item| {
-            matches!(
-                item,
-                ResponseItem::Message { role, content, .. }
-                    if role == "user"
-                        && content.iter().any(|part| matches!(
-                            part,
-                            ContentItem::InputText { text } if text == "summary"
-                        ))
-            )
-        });
-        let summary_position = summary_position.expect("summary should be present");
-        let initial_context = session.build_initial_context(&turn_context).await;
-        let expected_developer_messages: Vec<ResponseItem> = initial_context
-            .into_iter()
-            .filter(
-                |item| matches!(item, ResponseItem::Message { role, .. } if role == "developer"),
-            )
-            .collect();
-        let actual_after_summary = reconstructed[summary_position + 1..].to_vec();
-        assert_eq!(actual_after_summary, expected_developer_messages);
+        let mut expected = session.build_initial_context(&turn_context).await;
+        expected.push(summary_item);
+        assert_eq!(reconstructed, expected);
     }
 
     #[tokio::test]
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -1768,7 +1768,11 @@ impl Session {
                 }
                 RolloutItem::Compacted(compacted) => {
                     if let Some(replacement) = &compacted.replacement_history {
-                        history.replace(replacement.clone());
+                        let initial_context = self.build_initial_context(turn_context).await;
+                        history.replace(compact::refresh_compacted_developer_instructions(
+                            replacement.clone(),
+                            &initial_context,
+                        ));
                     } else {
                         let user_messages = collect_user_messages(history.raw_items());
                         let rebuilt = compact::build_compacted_history(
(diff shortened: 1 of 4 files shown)


charley-oai:
@codex review this

codex-rs/core/src/compact.rs
```diff
@@ -264,6 +264,25 @@ pub(crate) fn is_summary_message(message: &str) -> bool {
...
+    compacted_history
+        .retain(|item| !matches!(item, ResponseItem::Message { role, .. } if role == "developer"));
+
+    compacted_history.extend(
+        initial_context
+            .iter()
+            .filter(
+                |item| matches!(item, ResponseItem::Message { role, .. } if role == "developer"),
+            )
+            .cloned(),
```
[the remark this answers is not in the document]
charley-oai: Thanks for flagging this. We reviewed the flow and this behavior is intentional.

`trim_function_call_history_to_fit_context_window` only removes a trailing suffix of **codex-generated** items (developer messages / tool outputs) when the prompt is over the context window. The goal of `/compact` is to summarize the user/assistant conversation content, so excluding trailing codex-generated items is acceptable and expected.

Also, this does not cause persistent instruction loss: after remote compaction returns, we run `process_compacted_history(...)` and re-inject fresh developer instructions into the replacement history for subsequent turns

codex-rs/core/tests/suite/compact_remote.rs
```diff
@@ -537,10 +537,58 @@ async fn remote_compact_persists_replacement_history_in_rollout() -> Result<()>
         };
         if let RolloutItem::Compacted(compacted) = entry.item
             && compacted.message.is_empty()
-            && compacted.replacement_history.as_ref() == Some(&compacted_history)
+            && let Some(replacement_history) = compacted.replacement_history.as_ref()
         {
-            saw_compacted_history = true;
-            break;
+            let has_compacted_user_summary = replacement_history.iter().any(|item| {
```
aibrahim-oai, line 542: we do we need to edit this test?

codex-rs/core/src/compact.rs
```diff
@@ -264,6 +264,25 @@ pub(crate) fn is_summary_message(message: &str) -> bool {
     message.starts_with(format!("{SUMMARY_PREFIX}\n").as_str())
 }
 
+pub(crate) fn process_compacted_history(
+    mut compacted_history: Vec<ResponseItem>,
+    initial_context: &[ResponseItem],
+) -> Vec<ResponseItem> {
+    compacted_history
```
pakrym-oai: there is more stuff we can drop, some items are special user messages, can we somehow share the logic with parse_turn_item that already knows which messages are user messages and which are not?

codex-rs/core/tests/suite/compact_remote.rs
charley-oai: `process_compacted_history(...)` now intentionally mutates the remote compact output before persisting it:

1. removes any developer messages from remote output
2. injects fresh local developer instructions (permissions/personality/etc)

So `replacement_history` is no longer expected to be byte-for-byte equal to the raw `compacted_history` fixture. The old assertion `(== Some(&compacted_history))` would fail for the intended behavior.

codex-rs/core/src/compact.rs
charley-oai: Something like this?

```rust
fn should_keep_compacted_history_item(item: &ResponseItem) -> bool {
    match item {
        ResponseItem::Message { role, .. } if role == "developer" => false,
        ResponseItem::Message { role, .. } if role == "user" => {
            matches!(
                crate::event_mapping::parse_turn_item(item),
                Some(TurnItem::UserMessage(_))
            ) || collect_turn_aborted_marker(item).is_some()
        }
        _ => true,
    }
}
```

charley-oai: Discussed over slack, dropping the `|| collect_turn_aborted_marker(item).is_some()` as well

charley-oai:
@codex review this

codex-rs/core/src/compact.rs
```diff
@@ -264,6 +264,25 @@ pub(crate) fn is_summary_message(message: &str) -> bool {
...
+        .retain(|item| !matches!(item, ResponseItem::Message { role, .. } if role == "developer"));
+
+    compacted_history.extend(
+        initial_context
```
pakrym-oai: why are we filtering this at all?

codex-rs/core/src/codex.rs
```diff
@@ -1788,6 +1791,15 @@ impl Session {
         history.raw_items().to_vec()
     }
 
+    pub(crate) async fn process_compacted_history(
+        &self,
+        turn_context: &TurnContext,
+        compacted_history: Vec<ResponseItem>,
+    ) -> Vec<ResponseItem> {
+        let initial_context = self.build_initial_context(turn_context).await;
+        compact::process_compacted_history(compacted_history, &initial_context)
```
pakrym-oai, line 1797: can initial_context be strictly prepended?

codex-rs/core/src/codex.rs
```diff
@@ -1768,7 +1768,10 @@ impl Session {
                 }
                 RolloutItem::Compacted(compacted) => {
                     if let Some(replacement) = &compacted.replacement_history {
-                        history.replace(replacement.clone());
+                        history.replace(
```
pakrym-oai: instead of re-calculating every time, should we instead append the correct Compacted item?

codex-rs/core/tests/suite/compact_remote.rs
```diff
@@ -551,3 +599,161 @@ async fn remote_compact_persists_replacement_history_in_rollout() -> Result<()>
 
     Ok(())
 }
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn remote_compact_and_resume_refresh_stale_developer_instructions() -> Result<()> {
```
pakrym-oai, line 604: should we have a test that asserts the next request post compaction?

pakrym-oai: (without resume)

pakrym-oai (commented):
few qs

codex-rs/core/src/compact.rs
charley-oai: Good point now that we are dropping more stuff from the user messages we prob should inject the whole thing unfiltered

charley-oai: Although maybe this leads to compaction death spiral if the user has some ridiculously sized custom developer instructions, AGENTS.md, or we include too much environment context text accidentally

charley-oai: > What is bounded:
> 
> - Project docs (AGENTS.md discovered in repo) are byte-capped by project_doc_max_bytes (default 32 * 1024): codex-rs/core/src/config/mod.rs:96, codex-rs/core/src/
>   project_doc.rs:93.
> - Request-rule allow-prefix rendering in permissions instructions is capped (MAX_RENDERED_PREFIXES=100, MAX_ALLOW_PREFIX_TEXT_BYTES=5000): codex-rs/protocol/src/
>   models.rs:388.
> - EnvironmentContext itself is small/structured (cwd + shell): codex-rs/core/src/environment_context.rs:13.
> 
> What is not bounded (or not strongly bounded):
> 
> - developer_instructions override is reinjected as-is (no size cap): codex-rs/core/src/codex.rs:355, codex-rs/core/src/codex.rs:1898.
> - User instructions loaded from codex home (~/.codex/AGENTS.md / override) are read with read_to_string and no byte cap: codex-rs/core/src/config/mod.rs:1717.
> - Full initial_context is appended post-compaction without a size/token budget: codex-rs/core/src/compact.rs:252.

codex-rs/core/src/codex.rs
charley-oai: Ok spoke with Hanson and he recommended the following post-compaction layout:
1. system prompt
2. user messages except last
3. turn context
4. last user message
5. summary

codex-rs/core/src/codex.rs
charley-oai: Codex is getting a bit hung up on one edge case:

> A concrete failure case if we “just trust stored replacement_history”:
> 
> 1. Resume a thread with stale developer instructions in stored replacement history.
> 2. Before any normal user turn (which is where seeding happens), run /compact.
> 3. /compact path does not call seed_initial_context_if_needed first (codex-rs/core/src/codex.rs:2916).
> 4. Compaction prompt is built from current history as-is (codex-rs/core/src/compact.rs:113), so stale developer instructions are sent to /responses/compact.

But I'm down to just keep `history.replace(replacement.clone());` if you think that's cleaner. We can deal with resume behavior later (deduping and making sure compact-after-resume is fine)

charley-oai:
@codex review this

codex-rs/core/src/codex.rs
charley-oai: Discussed over slack, we don't want to rewrite everyone's history for this unimportant/rare edge case

codex-rs/core/src/compact.rs
```diff
@@ -32,6 +32,20 @@ use tracing::error;
 pub const SUMMARIZATION_PROMPT: &str = include_str!("../templates/compact/prompt.md");
 pub const SUMMARY_PREFIX: &str = include_str!("../templates/compact/summary_prefix.md");
 const COMPACT_USER_MESSAGE_MAX_TOKENS: usize = 20_000;
+// Turn context messages are re-injected after compaction so the next turn sees
+// canonical session state (permissions, environment context, instructions).
+//
+// Some of those entries can be very large (for example AGENTS.md or custom
```
pakrym-oai: Not sure I understand this. 

We inject the same items into a normal new conversation. Why is it worse to inject the same set of items post-compaction?

codex-rs/core/src/compact.rs
```diff
@@ -235,33 +249,119 @@ pub(crate) fn collect_user_messages(items: &[ResponseItem]) -> Vec<String> {
...
+        ResponseItem::Message { role, .. } if role == "developer" => false,
+        ResponseItem::Message { role, .. } if role == "user" => matches!(
+            crate::event_mapping::parse_turn_item(item),
+            Some(TurnItem::UserMessage(_))
+        ),
+        _ => true,
+    }
+}
+
+fn initial_context_for_reinjection(initial_context: &[ResponseItem]) -> Vec<ResponseItem> {
```
pakrym-oai: I'm very unsure we should mess with the context. We already removed exactly the same context from compaction items. There should be enough space to add more.

pakrym-oai (changes requested):
Qs on trimming context.

codex-rs/core/src/compact.rs
charley-oai: Discussed over slack, reverting that extra trimming logic in favor of startup-time sanitization

charley-oai:
@codex review this

codex-rs/core/src/compact.rs
```diff
@@ -235,33 +234,62 @@ pub(crate) fn collect_user_messages(items: &[ResponseItem]) -> Vec<String> {
...
+    initial_context: &[ResponseItem],
+) -> Vec<ResponseItem> {
+    compacted_history.retain(should_keep_compacted_history_item);
+
+    let initial_context = initial_context.to_vec();
+
+    // Re-inject canonical context from the current session since we stripped it
+    // from the pre-compaction history. Keep it right before the last user
+    // message so older user messages remain earlier in the transcript.
+    if let Some(last_user_index) = compacted_history.iter().rposition(|item| {
```
pakrym-oai, line 257: Why do we have 2 different ways of detecting user messages (talking about collect_user_messages).

pakrym-oai: can we rename should_keep_compacted_history_item  to is_user_message and use it here?

charley-oai: Doesn't `should_keep_compacted_history_item` need to also keep `ResponseItem::Compaction` items though? Which are not a user messages?

pakrym-oai: ah, right

pakrym-oai (approved):
Planned to approve all along

PR #10584, opened 2026-02-04
Add option to approve and remember MCP/Apps tool usage

canvrno-oai:
This PR adds a new approval option for app/MCP tool calls: “Allow and remember” (session-scoped).
When selected, Codex stores a temporary approval and auto-approves matching future calls for the rest of the session.

Added a session-scoped approval key (`server`, `connector_id`, `tool_name`) and persisted it in `tool_approvals` as `ApprovedForSession`.
On subsequent matching calls, approval is skipped and treated as accepted.
- Updated the approval question options to conditionally include:
- Accept
- Allow and remember (conditional)
- Decline
- Cancel

The new “Allow and remember” option is only shown when all of these are true:

1. The call is routed through the Codex Apps MCP server (codex_apps).
2. The tool requires approval based on annotations:
- read_only_hint == false, and
- destructive_hint == true or open_world_hint == true.
3. The tool includes a connector_id in metadata (used to build the remembered approval key).

If no `connector_id` is present, the prompt still appears (when approval is required), but only with the existing choices (Accept / Decline / Cancel). Approval prompting in this path has an explicit early return unless server == `codex_apps`.

### DIFF
--- a/codex-rs/core/src/mcp_tool_call.rs
+++ b/codex-rs/core/src/mcp_tool_call.rs
@@ -182,10 +182,10 @@ struct McpToolApprovalMetadata {
 }
 
 const MCP_TOOL_APPROVAL_QUESTION_ID_PREFIX: &str = "mcp_tool_call_approval";
-const MCP_TOOL_APPROVAL_ACCEPT: &str = "Yes, proceed once (y)";
+const MCP_TOOL_APPROVAL_ACCEPT: &str = "Yes, proceed once";
 const MCP_TOOL_APPROVAL_ACCEPT_AND_REMEMBER: &str =
-    "Yes, and don't ask again for this tool and this session (p)";
-const MCP_TOOL_APPROVAL_DECLINE: &str = "No (esc)";
+    "Yes, and don't ask again for this tool and session";
+const MCP_TOOL_APPROVAL_DECLINE: &str = "No";
 const MCP_TOOL_APPROVAL_CANCEL: &str = "Cancel";
 
 #[derive(Debug, Serialize)]
--- a/codex-rs/core/src/mcp_tool_call.rs
+++ b/codex-rs/core/src/mcp_tool_call.rs
@@ -183,7 +183,8 @@ struct McpToolApprovalMetadata {
 
 const MCP_TOOL_APPROVAL_QUESTION_ID_PREFIX: &str = "mcp_tool_call_approval";
 const MCP_TOOL_APPROVAL_ACCEPT: &str = "Yes, proceed once (y)";
-const MCP_TOOL_APPROVAL_ACCEPT_AND_REMEMBER: &str = "Yes, and don't ask again for this tool and this session (p)";
+const MCP_TOOL_APPROVAL_ACCEPT_AND_REMEMBER: &str =
+    "Yes, and don't ask again for this tool and this session (p)";
 const MCP_TOOL_APPROVAL_DECLINE: &str = "No (esc)";
 const MCP_TOOL_APPROVAL_CANCEL: &str = "Cancel";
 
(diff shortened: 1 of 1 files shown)


codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -167,21 +169,31 @@ async fn notify_mcp_tool_call_event(sess: &Session, turn_context: &TurnContext,
...
 
 struct McpToolApprovalMetadata {
     annotations: ToolAnnotations,
+    connector_id: Option<String>,
     connector_name: Option<String>,
     tool_title: Option<String>,
 }
 
 const MCP_TOOL_APPROVAL_QUESTION_ID_PREFIX: &str = "mcp_tool_call_approval";
 const MCP_TOOL_APPROVAL_ACCEPT: &str = "Accept";
```
mzeng-openai: Can we make them match the phrases we use for command approval so we can unify them later? i.e.
"Accept" -> "Approve Once"
"Allow and remember" -> "Approve this Session"
"Decline" -> "Deny"

mzeng-openai:
This is great! Just one nit on the copy, otherwise good to go! Have we tested locally?

mzeng-openai (approved):
Approve to unblock

codex-rs/core/src/mcp_tool_call.rs
canvrno-oai: @mzeng-openai  For commands, here are the phrases we have now:
```
  1. Yes, proceed (y)
  2. Yes, and don't ask again for commands that start with `echo hello world` (p)
  3. No, and tell Codex what to do differently (esc)
```
 
 How would you feel about something like: 
 ```
  1. Yes, proceed once (y)
  2. Yes, and don't ask again for this tool and this session (p)
  3. No (esc)
```

**Comment by canvrno-oai** [comment]
> This is great! Just one nit on the copy, otherwise good to go! Have we tested locally?

Yes, I tested with the slack connector.

File: codex-rs/core/src/mcp_tool_call.rs (hunk 1 above)
Comment (canvrno-oai) [reply codex-rs/core/src/mcp_tool_call.rs]: Ive updated the phrasing, we can adjust or unify next.

Comment (mzeng-openai) [reply codex-rs/core/src/mcp_tool_call.rs]: Oh you are right, the phrases I was mentioning are from the webview. That complicates things a little bit because the phrases we use for elicitation here are reused both in the CLI and the webview. So I guess either way works (my slight preference is to use the webview version since it's cleaner but your call)

File: codex-rs/core/src/mcp_tool_call.rs [hunk 2]
```diff
@@ -167,21 +169,31 @@ async fn notify_mcp_tool_call_event(sess: &Session, turn_context: &TurnContext,
...
     annotations: ToolAnnotations,
+    connector_id: Option<String>,
     connector_name: Option<String>,
     tool_title: Option<String>,
 }
 
 const MCP_TOOL_APPROVAL_QUESTION_ID_PREFIX: &str = "mcp_tool_call_approval";
-const MCP_TOOL_APPROVAL_ACCEPT: &str = "Accept";
-const MCP_TOOL_APPROVAL_DECLINE: &str = "Decline";
+const MCP_TOOL_APPROVAL_ACCEPT: &str = "Yes, proceed once (y)";
```
mzeng-openai: nit: The hotkey here will not work on webview, you have to either
1. Drop the hotkey
2. Check the originator() and optionally add it when the originator is codex_cli_rs

canvrno-oai: Thanks for catching that! For now I'm going to remove the hotkeys, and implement the second recommendation in a fast follow. Also adjusting the labels as described above.

PR #10589, opened 2026-02-04
Move metadata calculation out of client

pakrym-oai:
Model client shouldn't be responsible for this.

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -519,18 +519,34 @@ impl TurnContext {
             .unwrap_or(compact::SUMMARIZATION_PROMPT)
     }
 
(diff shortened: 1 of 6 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -514,6 +518,22 @@ impl TurnContext {
             .as_deref()
             .unwrap_or(compact::SUMMARIZATION_PROMPT)
     }
+
+    pub(crate) async fn resolve_turn_metadata_header(&self) -> Option<String> {
```
aibrahim-oai: should we have an async blocking function here? It runs on each turn, will add latency

pakrym-oai: added a timeout

aibrahim-oai (approved):

PR #10595, opened 2026-02-04
Stop client from being state carrier

pakrym-oai:
I'd like to make client session wide. This requires shedding all random state it has to carry.

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -212,6 +212,7 @@ use codex_protocol::models::ContentItem;
 use codex_protocol::models::DeveloperInstructions;
 use codex_protocol::models::ResponseInputItem;
 use codex_protocol::models::ResponseItem;
+use codex_protocol::openai_models::ReasoningEffort as ReasoningEffortConfig;
 use codex_protocol::protocol::CodexErrorInfo;
 use codex_protocol::protocol::InitialHistory;
 use codex_protocol::user_input::UserInput;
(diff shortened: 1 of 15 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -484,6 +485,15 @@ pub(crate) struct Session {
 pub(crate) struct TurnContext {
     pub(crate) sub_id: String,
     pub(crate) client: ModelClient,
+    pub(crate) config: Arc<Config>,
```
jif-oai, line 488: Do you plan to remove all those fields from the `ModelClientState` then?

pakrym-oai: Not all, at least half but either way I don't think we should be using the client to carry them around. The fact that the client needs `otel` is it's implementation detail.

codex-rs/core/src/codex.rs
```diff
@@ -484,6 +485,15 @@ pub(crate) struct Session {
...
+    pub(crate) auth_manager: Option<Arc<AuthManager>>,
+    pub(crate) model_info: ModelInfo,
+    pub(crate) otel_manager: OtelManager,
+    pub(crate) provider: ModelProviderInfo,
+    pub(crate) reasoning_effort: Option<ReasoningEffortConfig>,
+    pub(crate) reasoning_summary: ReasoningSummaryConfig,
+    pub(crate) session_source: SessionSource,
+    pub(crate) transport_manager: TransportManager,
```
aibrahim-oai, line 496: Shouldn't this live on client? or because it can change on websocket failure?

aibrahim-oai (approved):

codex-rs/core/src/codex.rs
pakrym-oai: This will get merged with the client.

PR #10623, opened 2026-02-04
Migrate state DB path helpers to versioned filename

jif-oai:
Summary
- add versioned state sqlite filename helpers and re-export them from the state crate
- remove legacy state files when initializing the runtime and update consumers/tests to use the new helpers
- tweak logs client description and database resolution to match the new path


gt-oai (approved):
why have versions of a database vs. migrating it? What sorts of use cases would we bump the version vs. migrating the current one?

jif-oai:
If we need to retrigger a backfill
I don't expect this to happen much in the future but while this feature is not fully landed, it let us care less about backward compatibility

PR #10631, opened 2026-02-04
Cloud Requirements: increase timeout and retries

gt-oai:
Add retries and an increased-length timeout for loading Cloud Requirements.


codex-rs/cloud-requirements/src/lib.rs
```diff
@@ -129,11 +132,40 @@ impl CloudRequirementsService {
...
 
-        let contents = self.fetcher.fetch_requirements(&auth).await?;
-        parse_cloud_requirements(&contents)
-            .inspect_err(|err| tracing::warn!(error = %err, "Failed to parse cloud requirements"))
-            .ok()
-            .flatten()
+        self.fetch_with_retries(&auth).await
+    }
+
+    async fn fetch_with_retries(&self, auth: &CodexAuth) -> Option<ConfigRequirementsToml> {
```
jif-oai, line 151: We should have shared code for this kind of things that support exp backoff etc

codex-rs/cloud-requirements/src/lib.rs
```diff
@@ -262,6 +297,29 @@ mod tests {
         }
     }
 
+    struct SequenceFetcher {
```
jif-oai, line 320: this looks like a very non idiomatic way to a retry policy

gt-oai: This is just in the test, to force retries?

codex-rs/cloud-requirements/src/lib.rs
gt-oai: am depending on `backoff` here. I can introduce a retry crate but would argue unnecessary for a for-loop

alexsong-oai (approved):

PR #10633, opened 2026-02-04
Cloud Requirements: take precedence over MDM

gt-oai:
Cloud Requirements should be applied before MDM requirements.


jif-oai (approved):
I guess this is validated by the team?

gt-oai:
> I guess this is validated by the team?

It will have to be, otherwise on MacOS you can do `defaults write com.openai.codex requirements_toml_base64 "$(base64 -i /Users/gt/requirements.toml)"` and overwrite the Cloud requirements.

gverma-openai (approved):

PR #10646, opened 2026-02-04
fix(sdk): remove hardcoded timeout from run.test.ts

MohdHasan2502:
Resolves a TODO to remove an explicit 10000ms timeout from the "throws ThreadRunError on turn failures" test case.

The default Jest timeout (usually 5000ms) should be sufficient for this test. Relying on the global timeout configuration prevents tests from hanging for excessive periods and encourages writing performant tests.


etraut-openai:

PR #10649, opened 2026-02-04
Fix test_shell_command_interruption flake

gt-oai:
## Human summary
Sandboxing (specifically `LandlockRestrict`) is means that e.g. `sleep 10` fails immediately. Therefore it cannot be interrupted. 

In suite::interrupt::test_shell_command_interruption, sleep 10 is issued at 17:28:16.554 (ToolCall: shell_command {"command":"sleep 10"...}), then fails at 17:28:16.589 with duration_ms=34, success=false, exit_code=101, and
    Sandbox(LandlockRestrict).

## Codex summary
- set `sandbox_mode = "danger-full-access"` in `interrupt` and `v2/turn_interrupt` integration tests
- set `sandbox: Some(SandboxMode::DangerFullAccess)` in `test_codex_jsonrpc_conversation_flow`
- set `sandbox_policy: Some(SandboxPolicy::DangerFullAccess)` in `command_execution_notifications_include_process_id`

## Why
On some Linux CI environments, command execution fails immediately with `LandlockRestrict` when sandboxed. These tests are intended to validate JSON-RPC/task lifecycle behavior (interrupt semantics, command notification shape/process id, request flow), but early sandbox startup failure changes turn flow and can trigger extra follow-up requests, causing flakes.

This change removes environment-specific sandbox startup dependency from these tests while preserving their primary intent.

## Testing
- not run in this environment (per request)

### DIFF
--- a/codex-rs/app-server/tests/suite/codex_message_processor_flow.rs
+++ b/codex-rs/app-server/tests/suite/codex_message_processor_flow.rs
@@ -36,7 +36,7 @@ use std::path::Path;
 use tempfile::TempDir;
 use tokio::time::timeout;
 
-const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(20);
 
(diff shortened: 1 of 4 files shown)


sayan-oai (approved):
ty

PR #10651, opened 2026-02-04
fix(core) switching model appends model instructions

dylan-hurd-oai:
## Summary
When switching models, we should append the instructions of the new model to the conversation as a developer message.

## Test
- [x] Adds a unit test


codex-rs/core/src/codex.rs
```diff
@@ -5820,6 +5846,143 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn model_change_appends_model_instructions_developer_message() {
```
joshka-oai: These tests seem pretty complex

aibrahim-oai (approved):

codex-rs/core/src/codex.rs
dylan-hurd-oai: Moved to an integration test file which mirrors similar test coverage

joshka-oai (approved):

PR #10657, opened 2026-02-04
feat: search_tool

apanasenko-oai:
**Why We Did This**
- The goal is to reduce MCP tool context pollution by not exposing the full MCP tool list up front
- It forces an explicit discovery step (`search_tool_bm25`) so the model narrows tool scope before making MCP calls, which helps relevance and lowers prompt/tool clutter.

**What It Changed**
- Added a new experimental feature flag `search_tool` in `core/src/features.rs:90` and `core/src/features.rs:430`.
- Added config/schema support for that flag in `core/config.schema.json:214` and `core/config.schema.json:1235`.
- Added BM25 dependency (`bm25`) in `Cargo.toml:129` and `core/Cargo.toml:23`.
- Added new tool handler `search_tool_bm25` in `core/src/tools/handlers/search_tool_bm25.rs:18`.
- Registered the handler and tool spec in `core/src/tools/handlers/mod.rs:11` and `core/src/tools/spec.rs:780` and `core/src/tools/spec.rs:1344`.
- Extended `ToolsConfig` to carry `search_tool` enablement in `core/src/tools/spec.rs:32` and `core/src/tools/spec.rs:56`.
- Injected dedicated developer instructions for tool-discovery workflow in `core/src/codex.rs:483` and `core/src/codex.rs:1976`, using `core/templates/search_tool/developer_instructions.md:1`.
- Added session state to store one-shot selected MCP tools in `core/src/state/session.rs:27` and `core/src/state/session.rs:131`.
- Added filtering so when feature is enabled, only selected MCP tools are exposed on the next request (then consumed) in `core/src/codex.rs:3800` and `core/src/codex.rs:3843`.
- Added E2E suite coverage for enablement/instructions/hide-until-search/one-turn-selection in `core/tests/suite/search_tool.rs:72`, `core/tests/suite/search_tool.rs:109`, `core/tests/suite/search_tool.rs:147`, and `core/tests/suite/search_tool.rs:218`.
- Refactored test helper utilities to support config-driven tool collection in `core/tests/suite/tools.rs:281`.

**Net Behavioral Effect**
- With `search_tool` **off**: existing MCP behavior (tools exposed normally).
- With `search_tool` **on**: MCP tools start hidden, model must call `search_tool_bm25`, and only returned `selected_tools` are available for the next model call.


jif-oai:
@codex review

codex-rs/core/src/state/session.rs
```diff
@@ -128,6 +130,32 @@ impl SessionState {
     pub(crate) fn dependency_env(&self) -> HashMap<String, String> {
         self.dependency_env.clone()
     }
+
+    pub(crate) fn merge_mcp_tool_selection(&mut self, tool_names: Vec<String>) -> Vec<String> {
+        if tool_names.is_empty() {
+            return self.active_mcp_tool_selection.clone().unwrap_or_default();
```
jif-oai, line 148: Ok for this because this is small but in general we should stop cloning everything all the time
In this function, the final tool selection is cloned like 3 times

codex-rs/core/src/tools/handlers/search_tool_bm25.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    async fn handle(&self, invocation: ToolInvocation) -> Result<ToolOutput, FunctionCallError> {
+        let ToolInvocation {
+            payload, session, ..
+        } = invocation;
+
+        let arguments = match payload {
+            ToolPayload::Function { arguments } => arguments,
+            _ => {
+                return Err(FunctionCallError::RespondToModel(
```
jif-oai: We have constraint sampling so this should never happen. I don't think this should be a RespondToModel

codex-rs/core/src/tools/handlers/search_tool_bm25.rs
```diff
@@ -0,0 +1,221 @@
...
+            ));
+        }
+
+        if args.limit == 0 {
+            return Err(FunctionCallError::RespondToModel(
+                "limit must be greater than zero".to_string(),
+            ));
+        }
+
+        let limit = args.limit.min(MAX_LIMIT);
```
jif-oai: Any reason why we don't error here? What happen if there are more elements to return? How the model can reach them?

codex-rs/core/src/tools/handlers/search_tool_bm25.rs
```diff
@@ -0,0 +1,221 @@
...
+            "total_tools": entries.len(),
+            "active_selected_tools": active_selected_tools,
+            "tools": result_payloads,
+        })
+        .to_string();
+
+        Ok(ToolOutput::Function {
+            body: FunctionCallOutputBody::Text(content),
+            success: Some(!result_payloads.is_empty()),
+        })
```
jif-oai, line 177: I don't see any events emission. Don't we want to render this in the clients?

codex-rs/core/src/tools/handlers/search_tool_bm25.rs
```diff
@@ -0,0 +1,221 @@
...
+            "query": query,
+            "total_tools": entries.len(),
+            "active_selected_tools": active_selected_tools,
+            "tools": result_payloads,
+        })
+        .to_string();
+
+        Ok(ToolOutput::Function {
+            body: FunctionCallOutputBody::Text(content),
+            success: Some(!result_payloads.is_empty()),
```
jif-oai: Is it really a failure if we did not find anything?

codex-rs/core/src/tools/handlers/search_tool_bm25.rs
```diff
@@ -0,0 +1,221 @@
...
+        })
+    }
+}
+
+fn collect_input_keys(input_schema: &serde_json::Map<String, JsonValue>) -> Vec<String> {
```
jif-oai: I would inline this to make the code easier to read for both humans and AI

codex-rs/core/src/tools/handlers/search_tool_bm25.rs
```diff
@@ -0,0 +1,221 @@
...
+    let Some(JsonValue::Object(map)) = input_schema.get("properties") else {
+        return Vec::new();
+    };
+
+    map.keys().map(ToString::to_string).collect()
+}
+
+fn build_search_text(name: &str, info: &ToolInfo, input_keys: &[String]) -> String {
```
jif-oai, line 181: OOC where is this format coming from? this looks like a strange serializer

codex-rs/core/src/tools/spec.rs
```diff
@@ -799,6 +802,34 @@ fn create_grep_files_tool() -> ToolSpec {
...
+            "limit".to_string(),
+            JsonSchema::Number {
+                description: Some("Maximum number of tools to return (defaults to 8).".to_string()),
+            },
+        ),
+    ]);
+
+    ToolSpec::Function(ResponsesApiTool {
+        name: "search_tool_bm25".to_string(),
+        description: "Searches MCP tool metadata with BM25 and exposes matching tools for the next model call.".to_string(),
```
jif-oai, line 827: Is the model aware of what BM25 is?

codex-rs/core/src/tools/spec.rs
```diff
@@ -799,6 +802,34 @@ fn create_grep_files_tool() -> ToolSpec {
...
+        (
+            "query".to_string(),
+            JsonSchema::String {
+                description: Some("Search query for MCP tools.".to_string()),
+            },
+        ),
+        (
+            "limit".to_string(),
+            JsonSchema::Number {
+                description: Some("Maximum number of tools to return (defaults to 8).".to_string()),
```
jif-oai: We should parametrize the 8 and I would talk about the max as wel

codex-rs/core/src/tools/spec.rs
```diff
@@ -799,6 +802,34 @@ fn create_grep_files_tool() -> ToolSpec {
     })
 }
 
+fn create_search_tool_bm25_tool() -> ToolSpec {
+    let properties = BTreeMap::from([
+        (
+            "query".to_string(),
+            JsonSchema::String {
+                description: Some("Search query for MCP tools.".to_string()),
```
jif-oai, line 812: Is the model supposed to know how to write down a query?

codex-rs/core/src/codex.rs
```diff
@@ -501,6 +501,9 @@ pub(crate) struct Session {
     next_internal_sub_id: AtomicU64,
 }
 
+const SEARCH_TOOL_DEVELOPER_INSTRUCTIONS: &str =
```
jif-oai, line 508: FYI: if we ship this to users, it will require a statsig update as the prompt is gated

codex-rs/core/src/codex.rs
```diff
@@ -2099,7 +2117,7 @@ impl Session {
         &self,
         turn_context: &TurnContext,
     ) -> Vec<ResponseItem> {
-        let mut items = Vec::<ResponseItem>::with_capacity(4);
+        let mut items = Vec::<ResponseItem>::with_capacity(6);
```
jif-oai: Why from 4 to 6?

codex-rs/core/src/codex.rs
```diff
@@ -3992,6 +4015,37 @@ fn filter_codex_apps_mcp_tools(
     mcp_tools
 }
 
+fn filter_codex_apps_mcp_tools_only(
```
jif-oai: This code should not live in this 6k+ lines file

codex-rs/core/src/codex.rs
```diff
@@ -4028,20 +4082,39 @@ async fn run_sampling_request(
         .list_all_tools()
         .or_cancel(&cancellation_token)
         .await?;
-    let connectors_for_tools = if turn_context.config.features.enabled(Feature::Apps) {
-        let connectors = connectors::accessible_connectors_from_mcp_tools(&mcp_tools);
+
+    let search_tool_enabled = turn_context.config.features.enabled(Feature::SearchTool);
+    let apps_enabled = turn_context.config.features.enabled(Feature::Apps);
```
jif-oai: ooc why changing this line with the if?

codex-rs/core/src/codex.rs
```diff
@@ -4028,20 +4082,39 @@ async fn run_sampling_request(
...
             &input,
             tool_selection.explicit_app_paths,
             tool_selection.skill_name_counts_lower,
         ))
     } else {
         None
     };
-    if let Some(connectors) = connectors_for_tools.as_ref() {
+
+    if search_tool_enabled {
```
jif-oai: why not inlining this?

codex-rs/core/src/features.rs
```diff
@@ -87,6 +87,10 @@ pub enum Feature {
     /// Allow the model to request web searches that fetch cached content.
     /// Takes precedence over `WebSearchRequest`.
     WebSearchCached,
+    /// Allow the model to search MCP tools via BM25 before exposing them.
+    SearchTool,
+    /// Gate the execpolicy enforcement for shell/unified exec.
+    ExecPolicy,
```
jif-oai: huh? How is it related?

codex-rs/core/src/features.rs
```diff
@@ -424,6 +428,16 @@ pub const FEATURES: &[FeatureSpec] = &[
         stage: Stage::Deprecated,
         default_enabled: false,
     },
+    FeatureSpec {
+        id: Feature::SearchTool,
+        key: "search_tool",
+        stage: Stage::Experimental {
```
jif-oai: I would not put it as experimental unless you triple checked the statsig gate for the prompt for every model

codex-rs/core/templates/search_tool/developer_instructions.md
```diff
@@ -0,0 +1,29 @@
+# MCP tool discovery
```
jif-oai, line 1: Doc now lives here: https://github.com/openai/developers-website

codex-rs/core/tests/suite/tools.rs
```diff
@@ -281,7 +281,10 @@ async fn sandbox_denied_shell_returns_original_output() -> Result<()> {
     Ok(())
 }
 
-async fn collect_tools(use_unified_exec: bool) -> Result<Vec<String>> {
+async fn collect_tools_with_config<F>(update: F) -> Result<Vec<String>>
```
jif-oai: I'm not sure to understand this... how is it related?

codex-rs/core/src/tools/handlers/search_tool_bm25.rs
apanasenko-oai: We don't want to overwhelm model with context if there a lot of results, but we can lift cap.

codex-rs/core/src/tools/handlers/search_tool_bm25.rs
apanasenko-oai: I don't think, why do we want to render anything about this? Once model will pick tools to call, we will render that.

codex-rs/core/src/tools/handlers/search_tool_bm25.rs
apanasenko-oai: >Is it really a failure if we did not find anything?

true, it no necessary a problem.

codex-rs/core/src/tools/handlers/search_tool_bm25.rs
apanasenko-oai: just dump everything about mcp into search corpus with simple spacing, without any formatting.

codex-rs/core/src/tools/spec.rs
apanasenko-oai:
```
› do you know what is BM25?


• Yes — BM25 (Best Matching 25) is a classic ranking function used in information retrieval. It scores how relevant a document is to a query using term frequency, inverse document frequency, and document length normalization; it’s a tuned, practical variant of TF‑IDF widely used in search engines.
```

codex-rs/core/src/tools/spec.rs
apanasenko-oai: model knows what BM25 and search query looks like, in my testing it know very well how to use it.

codex-rs/core/src/codex.rs
apanasenko-oai: I pass it as dev instruction, dev instructions is not require statsig update, right?

codex-rs/core/src/codex.rs
apanasenko-oai: nvm. some artifact of early refactoring.

codex-rs/core/src/features.rs
apanasenko-oai: sorry, slip through after rebase.

codex-rs/core/templates/search_tool/developer_instructions.md
apanasenko-oai: it's not doc, it's dev instruction for the model.

codex-rs/core/tests/suite/tools.rs
apanasenko-oai: it's not, rebase artifact.

codex-rs/core/src/tools/handlers/search_tool_bm25.rs
apanasenko-oai: Should we use `FunctionCallError::Fatal` here?

codex-rs/core/src/tools/handlers/search_tool_bm25.rs
apanasenko-oai: ok, removed max limit for now.

codex-rs/core/src/codex.rs
apanasenko-oai: moved to codex-rs/core/src/mcp_connection_manager.rs

codex-rs/core/src/codex.rs
apanasenko-oai: artifact of refactoring.

codex-rs/core/src/codex.rs
```diff
@@ -4028,20 +4082,39 @@ async fn run_sampling_request(
...
+            if let Some(selected_tools) = sess.get_mcp_tool_selection().await {
+                filter_mcp_tools_by_name(mcp_tools.clone(), &selected_tools)
+            } else {
+                HashMap::new()
+            };
+
+        if let Some(connectors) = apps_connectors.as_ref() {
+            let apps_mcp_tools = filter_codex_apps_mcp_tools_only(mcp_tools, connectors);
+            selected_mcp_tools.extend(apps_mcp_tools);
+        }
```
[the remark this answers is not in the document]
apanasenko-oai: it's actually desired behaviour.

apanasenko-oai:
@codex

codex-rs/core/src/tools/handlers/search_tool_bm25.rs
```diff
@@ -0,0 +1,217 @@
...
+            ));
+        }
+
+        if args.limit == 0 {
+            return Err(FunctionCallError::RespondToModel(
+                "limit must be greater than zero".to_string(),
+            ));
+        }
+
+        let limit = args.limit;
```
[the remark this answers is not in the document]
apanasenko-oai: removed 50 from dev instruction.

jif-oai:
@codex review

codex-rs/core/src/codex.rs
```diff
@@ -4065,9 +4091,25 @@ async fn run_sampling_request(
...
+
+        mcp_tools = selected_mcp_tools;
+    } else if let Some(connectors) = connectors_for_tools.as_ref() {
         mcp_tools = filter_codex_apps_mcp_tools(mcp_tools, connectors);
     }
+
     let router = Arc::new(ToolRouter::from_config(
         &turn_context.tools_config,
         Some(
```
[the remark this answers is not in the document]
apanasenko-oai: we don't try to restrict tool call rather don't expose and pollute model context.

codex-rs/core/src/codex.rs
jif-oai: Just confirm it by trying this feature with a personal account

codex-rs/core/src/tools/handlers/search_tool_bm25.rs
```diff
@@ -0,0 +1,217 @@
...
+use crate::tools::context::ToolInvocation;
+use crate::tools::context::ToolOutput;
+use crate::tools::context::ToolPayload;
+use crate::tools::handlers::parse_arguments;
+use crate::tools::registry::ToolHandler;
+use crate::tools::registry::ToolKind;
+
+pub struct SearchToolBm25Handler;
+
+pub(crate) const DEFAULT_LIMIT: usize = 8;
```
jif-oai, line 20: There is no more default limit from what I see... this means we either can have an unbounded message sent back to the model or this will be truncated so this has the same role as a limit but less controlled
Why not having a pagination system? With a limit and a pagination token? That would make this tool way more stable

jif-oai (approved):
lgtm for internal testing

codex-rs/core/src/tools/handlers/search_tool_bm25.rs
apanasenko-oai: There is no cap, but the model controls what limit it will use. We can't overflow the model since we can't return more tools than we currently have.

PR #10662, opened 2026-02-04
[apps] Cache MCP actions from apps.

mzeng-openai:
MCP actions take a long time to load for users with lots of apps installed. Adding a cache for these actions with 1hr expiration, given that they are almost always aren't going to change unless people install another app, which means they also need to restart codex to pick it up.


canvrno-oai (approved):
If we could refresh the cache when users make changes to avoid requiring a restart, I think that would be nice, but as scoped and specified this is a known behavior with these changes. LGTM.

PR #10672, opened 2026-02-04
Add js_repl host helpers and exec end events

fjord-oai:
## Summary

This PR adds host-integrated helper APIs for `js_repl` and updates model guidance so the agent can use them reliably.

### What’s included

- Add `codex.tool(name, args?)` in the JS kernel so `js_repl` can call normal Codex tools.
- Keep persistent JS state and scratch-path helpers available:
  - `codex.state`
  - `codex.tmpDir`
- Wire `js_repl` tool calls through the standard tool router path.
- Add/align `js_repl` execution completion/end event behavior with existing tool logging patterns.
- Update dynamic prompt injection (`project_doc`) to document:
  - how to call `codex.tool(...)`
  - raw output behavior
  - image flow via `view_image` (`codex.tmpDir` + `codex.tool("view_image", ...)`)
  - stdio safety guidance (`console.log` / `codex.tool`, avoid direct `process.std*`)

## Why

- Standardize JS-side tool usage on `codex.tool(...)`
- Make `js_repl` behavior more consistent with existing tool execution and event/logging patterns.
- Give the model enough runtime guidance to use `js_repl` safely and effectively.

## Testing

- Added/updated unit and runtime tests for:
  - `codex.tool` calls from `js_repl` (including shell/MCP paths)
  - image handoff flow via `view_image`
  - prompt-injection text for `js_repl` guidance
  - execution/end event behavior and related regression coverage

- ✅ `1` https://github.com/openai/codex/pull/10674
- 👉 `2` https://github.com/openai/codex/pull/10672
- ⏳ `3` https://github.com/openai/codex/pull/10671
- ⏳ `4` https://github.com/openai/codex/pull/10673
- ⏳ `5` https://github.com/openai/codex/pull/10670

### DIFF
--- a/codex-rs/core/src/project_doc.rs
+++ b/codex-rs/core/src/project_doc.rs
@@ -42,13 +42,13 @@ fn render_js_repl_instructions(config: &Config) -> Option<String> {
     let mut section = String::from("## JavaScript REPL (Node)\n");
     section.push_str("- Use `js_repl` for Node-backed JavaScript with top-level await in a persistent kernel. `codex.state` persists for the session (best effort) and is cleared by `js_repl_reset`.\n");
     section.push_str("- `js_repl` is a freeform/custom tool. Direct `js_repl` calls must send raw JavaScript tool input (optionally with first-line `// codex-js-repl: timeout_ms=15000`). Do not wrap code in JSON (for example `{\"code\":\"...\"}`), quotes, or markdown code fences.\n");
-    section.push_str("- Helpers available in `js_repl`: `codex.state` and `codex.tmpDir`.\n");
+    section.push_str("- Helpers: `codex.state`, `codex.tmpDir`, and `codex.tool(name, args?)`.\n");
+    section.push_str("- `codex.tool` executes a normal tool call and resolves to the raw tool output object. Use it for shell and non-shell tools alike.\n");
+    section.push_str("- To share generated images with the model, write a file under `codex.tmpDir`, call `await codex.tool(\"view_image\", { path: \"/absolute/path\" })`, then delete the file.\n");
     section.push_str("- Top-level bindings persist across cells. If you hit `SyntaxError: Identifier 'x' has already been declared`, reuse the binding, pick a new name, wrap in `{ ... }` for block scope, or reset the kernel with `js_repl_reset`.\n");
     section.push_str("- Top-level static import declarations (for example `import x from \"pkg\"`) are currently unsupported in `js_repl`; use dynamic imports with `await import(\"pkg\")` instead.\n");
 
-    section.push_str(
-        "- Avoid direct access to `process.stdout` / `process.stderr` / `process.stdin`; it can corrupt the JSON line protocol. Use `console.log`.",
-    );
+    section.push_str("- Avoid direct access to `process.stdout` / `process.stderr` / `process.stdin`; it can corrupt the JSON line protocol. Use `console.log` and `codex.tool(...)`.");
 
     Some(section)
 }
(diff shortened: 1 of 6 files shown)


codex-rs/core/src/project_doc.rs
```diff
@@ -42,13 +42,13 @@ fn render_js_repl_instructions(config: &Config) -> Option<String> {
     let mut section = String::from("## JavaScript REPL (Node)\n");
     section.push_str("- Use `js_repl` for Node-backed JavaScript with top-level await in a persistent kernel. `codex.state` persists for the session (best effort) and is cleared by `js_repl_reset`.\n");
     section.push_str("- `js_repl` is a freeform/custom tool. Direct `js_repl` calls must send raw JavaScript tool input (optionally with first-line `// codex-js-repl: timeout_ms=15000`). Do not wrap code in JSON (for example `{\"code\":\"...\"}`), quotes, or markdown code fences.\n");
-    section.push_str("- Helpers available in `js_repl`: `codex.state` and `codex.tmpDir`.\n");
+    section.push_str("- Helpers: `codex.state`, `codex.tmpDir`, and `codex.tool(name, args?)`.\n");
+    section.push_str("- `codex.tool` executes a normal tool call and resolves to the raw tool output object. Use it for shell and non-shell tools alike.\n");
```
owenlin0, line 46: `codex.tool` makes things really interesting, since for the first time we'll have the concept of tools calling other tools in a parent-child relationship.

let's use the example of a JS repl tool call that just calls the shell tool

in app-server API terms, you'll see something like:
```
- item/started (JS repl tool)
- item/started (CommandExecution, aka the shell tool)
- an approval request is potentially triggered for the command exec
- item/completed (CommandExecution)
- item/completed (JS repl tool)
```

I'm wondering if it'd be useful to model this parent-child relationship in EventMsg and pipe it through app-server's ThreadItem definitions (so clients can choose to render UI based on this relationship), but we can explore that in a later PR (as well as define the JS repl thread item in app-server v2.rs). I think as is, those order of events look good.

@fjord-oai should we prevent codex.tool from calling the JS Repl tool? to prevent infinite recursion :)

codex-rs/core/src/tools/js_repl/mod.rs
```diff
@@ -385,23 +416,49 @@ impl JsReplManager {
...
+                        let _ = tx.send(payload);
                     }
-                };
-                let _ = tx.send(payload);
+                    exec_contexts.lock().await.remove(&id);
+                }
+                KernelToHost::RunTool(req) => {
+                    let stdin_clone = Arc::clone(&stdin);
+                    let exec_contexts = Arc::clone(&exec_contexts);
+                    tokio::spawn(async move {
```
owenlin0, line 588: asked codex to review and thought this was a good catch:
> RunTool handling is tokio::spawned and untracked, while exec context is removed on exec completion above on line 438.
> 
> This can allow tool calls initiated in JS (especially un-awaited ones) to continue after the js_repl exec has already ended or timed out/reset, causing side effects outside the visible exec boundary.

I believe we should not allow codex.tool calls to run outside the js_repl exec boundary. So if the js_repl tool call ends, then all the tool calls within it must have also completed. Thoughts?

codex-rs/core/src/project_doc.rs
fjord-oai: Yes, prevented the infinite recursion case. I think I don't know enough about the app-server architecture to have an opinion on the other refactor you mentioned, but happy to explore as a followup!

codex-rs/core/src/tools/js_repl/mod.rs
fjord-oai: Good catch, fixed!

owenlin0 (approved):

PR #10688, opened 2026-02-04
add none personality option

aibrahim-oai:
- add none personality enum value and empty placeholder behavior\n- add docs/schema updates and e2e coverage


dylan-hurd-oai (approved):
Let's keep None

PR #10692, opened 2026-02-05
Fix remote compaction estimator/payload instruction small mismatch

charley-oai:
## Summary
This PR fixes a deterministic mismatch in remote compaction where pre-trim estimation and the `/v1/responses/compact` payload could use different base instructions.

Before this change:
- pre-trim estimation used model-derived instructions (`model_info.get_model_instructions(...)`)
- compact payload used session base instructions (`sess.get_base_instructions()`)

After this change:
- remote pre-trim estimation and compact payload both use the same `BaseInstructions` instance from session state.

## Changes
- Added a shared estimator entry point in `ContextManager`:
  - `estimate_token_count_with_base_instructions(&self, base_instructions: &BaseInstructions) -> Option<i64>`
- Kept `estimate_token_count(&TurnContext)` as a thin wrapper that resolves model/personality instructions and delegates to the new helper.
- Updated remote compaction flow to fetch base instructions once and reuse it for both:
  - trim preflight estimation
  - compact request payload construction
- Added regression coverage for parity and behavior:
  - unit test verifying explicit-base estimator behavior
  - integration test proving remote compaction uses session override instructions and trims accordingly

## Why this matters
This removes a deterministic divergence source where pre-trim could think the request fits while the actual compact request exceeded context because its instructions were longer/different.

## Scope
In scope:
- estimator/payload base-instructions parity in remote compaction

Out of scope:
- retry-on-`context_length_exceeded`
- compaction threshold/headroom policy changes
- broader trimming policy changes

## Codex author:
`codex fork 019c2b24-c2df-7b31-a482-fb8cf7a28559`

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -2075,10 +2075,10 @@ impl Session {
     }
 
     pub(crate) async fn recompute_token_usage(&self, turn_context: &TurnContext) {
-        let Some(estimated_total_tokens) = self
-            .clone_history()
-            .await
-            .estimate_token_count(turn_context)
+        let history = self.clone_history().await;
+        let base_instructions = self.get_base_instructions().await;
+        let Some(estimated_total_tokens) =
+            history.estimate_token_count_with_base_instructions(&base_instructions)
         else {
             return;
         };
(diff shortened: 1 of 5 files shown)


charley-oai:
@codex review this

sayan-oai (approved):
looks good. one question, do we also need to update the `estimate_token_count` call in `recompute_token_usage` that runs after compaction ([here](https://github.com/openai/codex/blob/main/codex-rs/core/src/codex.rs#L2097))?

charley-oai:
@codex review this

PR #10693, opened 2026-02-05
Add app-server transport layer with websocket support

maxj-oai:
- Adds --listen <URL> to codex app-server with two listen modes:
      - stdio:// (default, existing behavior)
      - ws://IP:PORT (new websocket transport)
  - Refactors message routing to be connection-aware:
      - Tracks per-connection session state (initialize/experimental capability)
      - Routes responses/errors to the originating connection
      - Broadcasts server notifications/requests to initialized connections
  - Updates initialization semantics to be per connection (not process-global), and updates app-server docs accordingly.
  - Adds websocket accept/read/write handling (JSON-RPC per text frame, ping/pong handling, connection lifecycle events).

Testing

  - Unit tests for transport URL parsing and targeted response/error routing.
  - New websocket integration test validating:
      - per-connection initialization requirements
      - no cross-connection response leakage
      - same request IDs on different connections route independently.


codex-rs/app-server/src/lib.rs
```diff
@@ -382,33 +528,318 @@ pub async fn run_main(
...
+                break;
+            }
+        }
+        info!("stdout writer exited (channel closed)");
+    }));
+
+    Ok(())
+}
+
+async fn start_websocket_acceptor(
```
JaviSoto: Thoughts on splitting this out into its own file? This file is getting pretty massive

JaviSoto (commented):
Not super familiar with the code, so someone else should take a look. It looks good to me overall. I particularly like the `--listen stdio://|ws://IP:PORT` API! 🙏 

Lazy question: are the changes to the app-server messages you made backwards compatible? Just wanna make sure that existing clients don't break

codex-rs/app-server/src/outgoing_message.rs
```diff
@@ -57,7 +79,13 @@ impl OutgoingMessageSender {
 
         let outgoing_message =
             OutgoingMessage::Request(request.request_with_id(outgoing_message_id.clone()));
-        if let Err(err) = self.sender.send(outgoing_message).await {
+        if let Err(err) = self
+            .sender
+            .send(OutgoingEnvelope::Broadcast {
+                message: outgoing_message,
+            })
```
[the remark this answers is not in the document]
maxj-oai: this is intentional - all connected clients are eligible to receive / reply to requests from server->client

codex-rs/app-server/src/lib.rs
maxj-oai: sure! done

maxj-oai:
@JaviSoto this doesn't change any of the app-server messages yet - just allows multiple connections and broadcast + response routing semantics

codex-rs/app-server/src/message_processor.rs
```diff
@@ -110,13 +109,17 @@ pub(crate) struct MessageProcessor {
...
-    initialized: bool,
-    experimental_api_enabled: Arc<AtomicBool>,
-    config_warnings: Vec<ConfigWarningNotification>,
+    config_warnings: Arc<Vec<ConfigWarningNotification>>,
+}
+
+#[derive(Debug, Default)]
+pub(crate) struct ConnectionSessionState {
+    pub(crate) initialized: bool,
+    experimental_api_enabled: bool,
```
owenlin0, line 118: hm I'm not sure if we want `experimental_api_enabled` to be on a per-connection basis, since the state can get pretty weird if a client starts a thread with the experimental API enabled, and another client subscribes to it with experimental API disabled. 

maybe the simplest solution is to make experimental API mode on/off a global setting for a running app-server instance, and it reject clients on `initialize` that don't match. cc @jif-oai 

a concrete example is `dynamic_tools` (what codex calls custom tools/functions that the client can register and handle over the app-server API) which is an experimental API feature. a client can connect to app-server via the experimental API, start a thread with some dynamic_tools, then the server will broadcast a server request for the clients to respond to the tool call. it gets a bit weird for the client that is not aware of the experimental API

owenlin0: maybe we can punt this to a followup

codex-rs/app-server/README.md
```diff
@@ -15,11 +15,16 @@
...
 
 ## Protocol
 
-Similar to [MCP](https://modelcontextprotocol.io/), `codex app-server` supports bidirectional communication, streaming JSONL over stdio. The protocol is JSON-RPC 2.0, though the `"jsonrpc":"2.0"` header is omitted.
+Similar to [MCP](https://modelcontextprotocol.io/), `codex app-server` supports bidirectional communication using JSON-RPC 2.0 messages (with the `"jsonrpc":"2.0"` header omitted on the wire).
+
+Supported transports:
+
+- stdio (`--listen stdio://`, default): newline-delimited JSON (JSONL)
+- websocket (`--listen ws://IP:PORT`): one JSON-RPC message per websocket text frame
```
owenlin0: worth mentioning this is alpha / experimental

owenlin0 (commented):
looks like a good start to me, but curious what @jif-oai thinks about how multi-client should work with experimental

owenlin0 (approved):
approving to unblock, i think the experimental thing we can figure out as a followup

codex-rs/app-server/src/message_processor.rs
maxj-oai: OK sounds good; I've added a TODO to revisit this

PR #10698, opened 2026-02-05
core: preconnect Responses websocket for first turn

joshka-oai:
## Problem
The first user turn can pay websocket handshake latency even when a session has already started. We want to reduce that initial delay while preserving turn semantics and avoiding any prompt send during startup.

Reviewer feedback also called out duplicated connect/setup paths and unnecessary preconnect state complexity.

## Mental model
`ModelClient` owns session-scoped transport state. During session startup, it can opportunistically warm one websocket handshake slot. A turn-scoped `ModelClientSession` adopts that slot once if available, restores captured sticky turn-state, and otherwise opens a websocket through the same shared connect path.

If startup preconnect is still in flight, first turn setup awaits that task and treats it as the first connection attempt for the turn.

Preconnect is handshake-only. The first `response.create` is still sent only when a turn starts.

## Non-goals
This change does not make preconnect required for correctness and does not change prompt/turn payload semantics. It also does not expand fallback behavior beyond clearing preconnect state when fallback activates.

## Tradeoffs
The implementation prioritizes simpler ownership and shared connection code over header-match gating for reuse. The single-slot cache keeps lifecycle straightforward but only benefits the immediate next turn.

Awaiting in-flight preconnect has the same app-level connect-timeout semantics as existing websocket connect behavior (no new timeout class introduced by this PR).

## Architecture
`core/src/client.rs`:
- Added session-level preconnect lifecycle state (`Idle` / `InFlight` / `Ready`) carrying one warmed websocket plus optional captured turn-state.
- Added `pre_establish_connection()` startup warmup and `preconnect()` handshake-only setup.
- Deduped auth/provider resolution into `current_client_setup()` and websocket handshake wiring into `connect_websocket()` / `build_websocket_headers()`.
- Updated turn websocket path to adopt preconnect first, await in-flight preconnect when present, then create a new websocket only when needed.
- Ensured fallback activation clears warmed preconnect state.
- Added documentation for lifecycle, ownership, sticky-routing invariants, and timeout semantics.

`core/src/codex.rs`:
- Session startup invokes `model_client.pre_establish_connection(...)`.
- Turn metadata resolution uses the shared timeout helper.

`core/src/turn_metadata.rs`:
- Centralized shared timeout helper used by both turn-time metadata resolution and startup preconnect metadata building.

`core/tests/common/responses.rs` + websocket test suites:
- Added deterministic handshake waiting helper (`wait_for_handshakes`) with bounded polling.
- Added startup preconnect and in-flight preconnect reuse coverage.
- Fallback expectations now assert exactly two websocket attempts in covered scenarios (startup preconnect + turn attempt before fallback sticks).

## Observability
Preconnect remains best-effort and non-fatal. Existing websocket/fallback telemetry remains in place, and debug logs now make preconnect-await behavior and preconnect failures easier to reason about.

## Tests
Validated with:
1. `just fmt`
2. `cargo test -p codex-core websocket_preconnect -- --nocapture`
3. `cargo test -p codex-core websocket_fallback -- --nocapture`
4. `cargo test -p codex-core websocket_first_turn_waits_for_inflight_preconnect -- --nocapture`

### DIFF
--- a/codex-rs/core/src/client.rs
+++ b/codex-rs/core/src/client.rs
@@ -9,10 +9,16 @@
 //! call site.
 //!
 //! A [`ModelClientSession`] is created per turn and is used to stream one or more Responses API
-//! requests during that turn. It caches a Responses WebSocket connection (opened lazily) and
-//! stores per-turn state such as the `x-codex-turn-state` token used for sticky routing.
+//! requests during that turn. It caches a Responses WebSocket connection (opened lazily, or reused
+//! from a session-level preconnect) and stores per-turn state such as the `x-codex-turn-state`
+//! token used for sticky routing.
+//!
+//! Preconnect is intentionally handshake-only: it may warm a socket and capture sticky-routing
+//! state, but the first `response.create` payload is still sent only when a turn starts.
 
+use std::path::PathBuf;
 use std::sync::Arc;
+use std::sync::Mutex;
 use std::sync::OnceLock;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
@@ -66,6 +72,7 @@ use std::time::Duration;
 use tokio::sync::mpsc;
 use tokio_tungstenite::tungstenite::Error;
 use tokio_tungstenite::tungstenite::Message;
+use tracing::debug;
 use tracing::warn;
 
 use crate::AuthManager;
@@ -81,6 +88,8 @@ use crate::flags::CODEX_RS_SSE_FIXTURE;
 use crate::model_provider_info::ModelProviderInfo;
 use crate::model_provider_info::WireApi;
 use crate::tools::spec::create_tools_json_for_responses_api;
+use crate::turn_metadata::TURN_METADATA_HEADER_TIMEOUT;
+use crate::turn_metadata::build_turn_metadata_header;
 
 pub const OPENAI_BETA_HEADER: &str = "OpenAI-Beta";
 pub const OPENAI_BETA_RESPONSES_WEBSOCKETS: &str = "responses_websockets=2026-02-04";
@@ -93,7 +102,6 @@ pub const X_RESPONSESAPI_INCLUDE_TIMING_METRICS_HEADER: &str =
 ///
 /// This is intentionally kept minimal so `ModelClient` does not need to hold a full `Config`. Most
 /// configuration is per turn and is passed explicitly to streaming/unary methods.
-#[derive(Debug)]
 struct ModelClientState {
     auth_manager: Option<Arc<AuthManager>>,
     conversation_id: ThreadId,
@@ -105,6 +113,52 @@ struct ModelClientState {
     include_timing_metrics: bool,
     beta_features_header: Option<String>,
     disable_websockets: AtomicBool,
+    /// Single-slot cache for a warmed websocket that can be adopted by the next turn.
+    ///
+    /// The slot is consumed on read to avoid accidentally sharing one socket across
+    /// multiple concurrent turn sessions.
+    preconnected_websocket: Mutex<Option<ApiWebSocketConnection>>,
+    /// Sticky routing token captured from a preconnected websocket handshake.
+    ///
+    /// This is consumed together with `preconnected_websocket`.
+    preconnected_turn_state: Mutex<Option<String>>,
+}
+
+impl std::fmt::Debug for ModelClientState {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ModelClientState")
+            .field("auth_manager", &self.auth_manager)
+            .field("conversation_id", &self.conversation_id)
+            .field("provider", &self.provider)
+            .field("session_source", &self.session_source)
+            .field("model_verbosity", &self.model_verbosity)
+            .field(
+                "enable_responses_websockets",
+                &self.enable_responses_websockets,
+            )
+            .field(
+                "enable_request_compression",
+                &self.enable_request_compression,
+            )
+            .field("include_timing_metrics", &self.include_timing_metrics)
+            .field("beta_features_header", &self.beta_features_header)
+            .field(
+                "disable_websockets",
+                &self.disable_websockets.load(Ordering::Relaxed),
+            )
+            .field("preconnected_websocket", &"<opaque>")
+            .finish()
+    }
+}
+
+/// Resolved API client setup for a single request attempt.
+///
+/// Keeping this as a single bundle ensures preconnect and normal request paths
+/// share the same auth/provider setup flow.
+struct CurrentClientSetup {
+    auth: Option<CodexAuth>,
+    api_provider: codex_api::Provider,
+    api_auth: CoreAuthProvider,
 }
 
 /// A session-scoped client for model-provider API calls.
@@ -116,8 +170,9 @@ struct ModelClientState {
 /// WebSocket fallback is session-scoped: once a turn activates the HTTP fallback, subsequent turns
 /// will also use HTTP for the remainder of the session.
 ///
-/// Turn-scoped settings (model selection, reasoning controls, telemetry context, and turn metadata)
-/// are passed explicitly to the relevant methods to keep turn lifetime visible at the call site.
+/// Turn-scoped settings (model selection, reasoning controls, telemetry context, and turn
+/// metadata) are passed explicitly to the relevant methods to keep turn lifetime visible at the
+/// call site.
 ///
 /// This type is cheap to clone.
 #[derive(Debug, Clone)]
@@ -127,8 +182,8 @@ pub struct ModelClient {
 
 /// A turn-scoped streaming session created from a [`ModelClient`].
 ///
-/// The session lazily establishes a Responses WebSocket connection (and reuses it across multiple
-/// requests) and caches per-turn state:
+/// The session establishes a Responses WebSocket connection lazily (or adopts a preconnected one)
+/// and reuses it across multiple requests within the turn. It also caches per-turn state:
 ///
 /// - The last request's input items, so subsequent calls can use `response.append` when the input
 ///   is an incremental extension of the previous request.
@@ -184,20 +239,92 @@ impl ModelClient {
                 include_timing_metrics,
                 beta_features_header,
                 disable_websockets: AtomicBool::new(false),
+                preconnected_websocket: Mutex::new(None),
+                preconnected_turn_state: Mutex::new(None),
             }),
         }
     }
 
     /// Creates a fresh turn-scoped streaming session.
     ///
-    /// This does not open any network connections; the WebSocket connection is established lazily
-    /// when the first WebSocket stream request is issued.
+    /// This constructor does not perform network I/O itself. The returned session either adopts a
+    /// previously preconnected websocket or opens a websocket lazily when the first stream request
+    /// is issued.
     pub fn new_session(&self) -> ModelClientSession {
         ModelClientSession {
             client: self.clone(),
             connection: None,
             websocket_last_items: Vec::new(),
             turn_state: Arc::new(OnceLock::new()),
+        }
+    }
+
+    /// Spawns a best-effort task that warms a websocket for the first turn.
+    ///
+    /// This call performs only connection setup; it never sends prompt payloads.
+    ///
+    /// A timeout when computing turn metadata is treated the same as "no metadata" so startup
+    /// cannot block indefinitely on optional preconnect context.
+    pub fn pre_establish_connection(&self, otel_manager: OtelManager, cwd: PathBuf) {
+        let model_client = self.clone();
+        tokio::spawn(async move {
+            let turn_metadata_header: Option<String> = tokio::time::timeout(
+                TURN_METADATA_HEADER_TIMEOUT,
+                build_turn_metadata_header(cwd.as_path()),
+            )
+            .await
+            .unwrap_or_default();
+            let _ = model_client
+                .preconnect(&otel_manager, turn_metadata_header.as_deref())
+                .await;
+        });
+    }
+
+    /// Opportunistically pre-establishes a Responses WebSocket connection for this session.
+    ///
+    /// This method is best-effort: it returns `false` on any setup/connect failure and the caller
+    /// should continue normally. A successful preconnect reduces first-turn latency but never sends
+    /// an initial prompt; the first `response.create` is still sent only when a turn starts.
+    ///
+    /// The preconnected slot is single-consumer and single-use: the next `ModelClientSession` may
+    /// adopt it once, after which later turns either keep using that same turn-local connection or
+    /// create a new one.
+    pub async fn preconnect(
+        &self,
+        otel_manager: &OtelManager,
+        turn_metadata_header: Option<&str>,
+    ) -> bool {
+        if !self.responses_websocket_enabled() || self.disable_websockets() {
+            return false;
+        }
+
+        let client_setup = match self.current_client_setup().await {
+            Ok(client_setup) => client_setup,
+            Err(err) => {
+                warn!("failed to build websocket preconnect client setup: {err}");
+                return false;
+            }
+        };
+        let turn_state = Arc::new(OnceLock::new());
+
+        match self
+            .connect_websocket(
+                otel_manager,
+                client_setup.api_provider,
+                client_setup.api_auth,
+                Some(Arc::clone(&turn_state)),
+                turn_metadata_header,
+            )
+            .await
+        {
+            Ok(connection) => {
+                self.store_preconnected_websocket(connection, turn_state.get().cloned());
+                true
+            }
+            Err(err) => {
+                debug!("websocket preconnect failed: {err}");
+                false
+            }
         }
     }
 
@@ -217,20 +344,12 @@ impl ModelClient {
         if prompt.input.is_empty() {
             return Ok(Vec::new());
         }
-        let auth_manager = self.state.auth_manager.clone();
-        let auth = match auth_manager.as_ref() {
-            Some(manager) => manager.auth().await,
-            None => None,
-        };
-        let api_provider = self
-            .state
-            .provider
-            .to_api_provider(auth.as_ref().map(CodexAuth::auth_mode))?;
-        let api_auth = auth_provider_from_auth(auth.clone(), &self.state.provider)?;
+        let client_setup = self.current_client_setup().await?;
         let transport = ReqwestTransport::new(build_reqwest_client());
         let request_telemetry = Self::build_request_telemetry(otel_manager);
-        let client = ApiCompactClient::new(transport, api_provider, api_auth)
-            .with_telemetry(Some(request_telemetry));
+        let client =
+            ApiCompactClient::new(transport, client_setup.api_provider, client_setup.api_auth)
+                .with_telemetry(Some(request_telemetry));
 
         let instructions = prompt.base_instructions.text.clone();
         let payload = ApiCompactionInput {
@@ -263,20 +382,12 @@ impl ModelClient {
             return Ok(Vec::new());
         }
 
-        let auth_manager = self.state.auth_manager.clone();
-        let auth = match auth_manager.as_ref() {
-            Some(manager) => manager.auth().await,
-            None => None,
-        };
-        let api_provider = self
-            .state
-            .provider
-            .to_api_provider(auth.as_ref().map(CodexAuth::auth_mode))?;
-        let api_auth = auth_provider_from_auth(auth, &self.state.provider)?;
+        let client_setup = self.current_client_setup().await?;
         let transport = ReqwestTransport::new(build_reqwest_client());
         let request_telemetry = Self::build_request_telemetry(otel_manager);
-        let client = ApiMemoriesClient::new(transport, api_provider, api_auth)
-            .with_telemetry(Some(request_telemetry));
+        let client =
+            ApiMemoriesClient::new(transport, client_setup.api_provider, client_setup.api_auth)
+                .with_telemetry(Some(request_telemetry));
 
         let payload = ApiMemoryTraceSummarizeInput {
             model: model_info.slug.clone(),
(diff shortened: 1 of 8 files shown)


codex-rs/core/src/client.rs
```diff
@@ -88,6 +91,31 @@ pub const X_CODEX_TURN_METADATA_HEADER: &str = "x-codex-turn-metadata";
...
 
+/// A session-scoped websocket warmed before the first user turn starts.
+///
+/// This stores only transport state. The preconnect path performs the websocket
+/// handshake and captures any server-provided turn-state header, but it does not
+/// send `response.create` or any user prompt payload.
+///
+/// Callers must validate connection-scoped headers before reuse to avoid sharing a
+/// warmed socket across incompatible turn settings.
+struct PreconnectedWebSocket {
```
pakrym-oai: That's too much state. There should be no state in preconnected web socket and turn_metadata_header should already have been sent.

pakrym-oai: is there a way where we only store Option<ApiWebSocketConnection> ?

codex-rs/core/src/client.rs
```diff
@@ -201,6 +236,155 @@ impl ModelClient {
...
+    /// On mismatch, the preconnected socket is dropped and the turn opens a fresh connection.
+    ///
+    /// This method is best-effort: it returns `false` on any setup/connect failure and the caller
+    /// should continue normally. A successful preconnect reduces first-turn latency but never sends
+    /// an initial prompt; the first `response.create` is still sent only when a turn starts.
+    ///
+    /// Callers should pass the same connection-scoped values they expect to use on the first turn.
+    /// If these inputs differ from the eventual stream call, the warmed socket is intentionally
+    /// discarded and the first turn pays a fresh handshake cost.
+    pub async fn preconnect(
```
pakrym-oai, line 336: this is a lot of duplicated code with the main connect path.

codex-rs/core/src/client.rs
```diff
@@ -201,6 +236,155 @@ impl ModelClient {
...
+        }
+
+        let auth_manager = self.state.auth_manager.clone();
+        let normalized_turn_metadata = normalize_turn_metadata_header(turn_metadata_header);
+        let turn_state = Arc::new(OnceLock::new());
+        let mut auth_recovery = auth_manager
+            .as_ref()
+            .map(super::auth::AuthManager::unauthorized_recovery);
+
+        loop {
```
pakrym-oai: I think we don't need retries here. Retries can happen on the main path if needed.

codex-rs/core/src/client.rs
```diff
@@ -201,6 +236,155 @@ impl ModelClient {
...
+        turn_metadata_header: Option<&str>,
+    ) -> Option<PreconnectedWebSocket> {
+        let mut slot = self
+            .state
+            .preconnected_websocket
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        let preconnected = slot.take()?;
+        let expected_turn_metadata = normalize_turn_metadata_header(turn_metadata_header);
+        if preconnected.web_search_eligible != web_search_eligible
```
pakrym-oai: we should still take the connection even if these don't match, web_search_eligible is going away and turn_metadata_header is best effort.

codex-rs/core/src/client.rs
```diff
@@ -468,19 +650,38 @@ impl ModelClientSession {
...
             Some(conn) => conn.is_closed().await,
             None => true,
         };
 
         if needs_new {
+            {
+                // The preconnect slot is single-use and scoped to the immediate next turn.
+                // If we still need a fresh websocket here (mismatch, closed socket, or no slot),
+                // clear any leftover cached connection to avoid future stale reuse attempts.
+                let mut preconnected = self
```
pakrym-oai: why are we doing this in addition to try_use_preconnected_websocket ?

codex-rs/core/src/client.rs
```diff
@@ -506,6 +707,29 @@ impl ModelClientSession {
...
+            .client
+            .take_preconnected_websocket(web_search_eligible, turn_metadata_header)
+        else {
+            return;
+        };
+
+        if let Some(turn_state) = preconnected.turn_state {
+            let _ = self.turn_state.set(turn_state);
+        }
+        self.connection = Some(preconnected.connection);
```
pakrym-oai: should this return a value instead of mutating self?  then you won't need the self.connection.is_some() check and it will slot nicer into websocket_connection

codex-rs/core/src/client.rs
```diff
@@ -751,6 +979,13 @@ impl ModelClientSession {
 
             self.connection = None;
             self.websocket_last_items.clear();
+            let mut preconnected = self
```
pakrym-oai: reuse try_use_preconnected_websocket and discard the value?

codex-rs/core/src/codex.rs
```diff
@@ -1053,6 +1053,39 @@ impl Session {
             next_internal_sub_id: AtomicU64::new(0),
         });
 
+        let preconnect_web_search_eligible = !matches!(
```
pakrym-oai: we should just ignore everything about web_search_eligible

codex-rs/core/src/codex.rs
```diff
@@ -1053,6 +1053,39 @@ impl Session {
...
+                session_configuration.provider.is_azure_responses_endpoint(),
+                session_configuration.sandbox_policy.get(),
+            ),
+            WebSearchMode::Disabled
+        );
+        // Warm a websocket in the background so the first turn can reuse it when headers match.
+        // This performs only connection setup; user input is still sent later via response.create
+        // when submit_turn() runs.
+        {
+            let model_client = sess.services.model_client.clone();
```
pakrym-oai: can this be a single method call on the client? client.pre_establish_connection()?

codex-rs/core/src/codex.rs
```diff
@@ -1053,6 +1053,39 @@ impl Session {
...
+            let otel_manager = sess.services.otel_manager.clone();
+            let preconnect_cwd = session_configuration.cwd.clone();
+            tokio::spawn(async move {
+                const TURN_METADATA_HEADER_TIMEOUT_MS: u64 = 250;
```
pakrym-oai: Should this timeout code be repeated here?

codex-rs/core/tests/suite/agent_websocket.rs
```diff
@@ -67,3 +68,36 @@ async fn websocket_test_codex_shell_chain() -> Result<()> {
...
+
+    let deadline = tokio::time::Instant::now() + Duration::from_secs(2);
+    loop {
+        if server.handshakes().len() == 1 {
+            break;
+        }
+        if tokio::time::Instant::now() >= deadline {
+            panic!("expected websocket preconnect handshake during session startup");
+        }
+        tokio::time::sleep(Duration::from_millis(10)).await;
```
pakrym-oai: is this too cpu intensive to wait? is there something more deterministic we can do?

codex-rs/core/tests/suite/client_websockets.rs
```diff
@@ -78,6 +78,83 @@ async fn responses_websocket_streams_request() {
...
+        .expect("websocket stream failed");
+
+    while let Some(event) = stream.next().await {
+        if matches!(event, Ok(ResponseEvent::Completed { .. })) {
+            break;
+        }
+    }
+
+    let handshakes = server.handshakes();
+    assert_eq!(handshakes.len(), 2);
```
pakrym-oai: why two?

codex-rs/core/tests/suite/websocket_fallback.rs
```diff
@@ -46,7 +46,7 @@ async fn websocket_fallback_switches_to_http_after_retries_exhausted() -> Result
         .filter(|req| req.method == Method::POST && req.url.path().ends_with("/responses"))
         .count();
 
-    assert_eq!(websocket_attempts, 1);
+    assert_eq!(websocket_attempts, 2);
```
pakrym-oai, line 52: why two? shouldn't the connection be reused?

pakrym-oai (changes requested):
The connection code should be deduped as much as possible. I also think we can simplify state/logic in many places.

codex-rs/core/src/client.rs
joshka-oai: we do take it (remove it from the slot). We just don't use it if these don't match. Will remove the check though, which should simplify this

joshka-oai:
Addressed the remaining duplication concern in `core/src/client.rs`.

I extracted websocket client setup into shared helpers:
- `current_auth()`
- `provider_and_auth_from_auth(...)`

and now reuse those in:
- `preconnect(...)`
- `stream_responses_api(...)`
- `stream_responses_websocket(...)`
- unary calls that used the same setup (`compact_conversation_history`, `summarize_memory_traces`)

So preconnect and the main connect path no longer carry separate copies of auth/provider setup logic.

Validation run:
- `just fmt`
- `cargo test -p codex-core websocket -- --nocapture`

joshka-oai:
> The connection code should be deduped as much as possible. I also think we can simplify state/logic in many places.

Will fix this tomorrow. Above comment was codex generated and shouldn't be seen as a "this is complete" message. Apologies.

joshka-oai (commented):
Consolidated response to all inline comments (single review), plus two follow-up commits:
- `e1e2976d`: reorder newly added preconnect helpers for top-down readability (non-functional)
- `2473939a`: documentation-only pass clarifying preconnect/turn-state contracts (non-functional)

1. “Too much preconnect state…”
Handled.
- Removed `PreconnectedWebSocket` header-matching wrapper state.
- Preconnect cache is now websocket + minimal sticky turn-state continuity only.
- Refs: `codex-rs/core/src/client.rs` (`preconnected_websocket`, `preconnected_turn_state`).

2. “Only store Option<ApiWebSocketConnection>?”
Mostly handled.
- The socket slot is `Option<ApiWebSocketConnection>`.
- Kept a separate optional string for sticky-routing continuity when adopting preconnect.
- Refs: `codex-rs/core/src/client.rs` (`preconnected_websocket`, `preconnected_turn_state`).

3. “Duplicated code with main connect path.”
Handled.
- Shared setup helper: `current_client_setup()`.
- Shared websocket handshake path: `connect_websocket()` + `build_websocket_headers()` used by preconnect and normal reconnect.
- Refs: `codex-rs/core/src/client.rs`.

4. “No retries needed in preconnect.”
Handled.
- Preconnect is single best-effort attempt.
- Ref: `ModelClient::preconnect` in `codex-rs/core/src/client.rs`.

5. “Still take preconnected connection even if headers don’t match.”
Handled.
- Header-match gating was removed; preconnected socket is adopted directly.
- Ref: `websocket_connection` / `try_use_preconnected_websocket` in `codex-rs/core/src/client.rs`.

6. “Why this in addition to try_use_preconnected_websocket?”
Handled.
- Adoption is centralized in `try_use_preconnected_websocket`.
- Cleanup is centralized with `clear_preconnected_websocket`.
- Refs: `codex-rs/core/src/client.rs`.

7. “Should this return a value instead of mutating self?”
Handled.
- `try_use_preconnected_websocket` now returns `Option<ApiWebSocketConnection>`.
- Ref: `codex-rs/core/src/client.rs`.

8. “Reuse try_use_preconnected_websocket and discard value?”
Handled (equivalent centralized cleanup).
- Fallback/reconnect paths use shared preconnect clear helper.
- Ref: `try_switch_fallback_transport` + `clear_preconnected_websocket` in `codex-rs/core/src/client.rs`.

9. “Ignore everything about web_search_eligible.”
Partially handled.
- Removed from preconnect reuse gating / startup preconnect decision logic.
- Still present in active request header construction path (`x-oai-web-search-eligible`) for normal turn requests.

10. “Single client method call from codex startup?”
Handled.
- Startup now calls one client method: `pre_establish_connection(...)`.
- Refs: `codex-rs/core/src/codex.rs`, `codex-rs/core/src/client.rs`.

11. “Should timeout code be repeated here?”
Handled.
- Timeout logic moved into `ModelClient::pre_establish_connection`.
- Ref: `codex-rs/core/src/client.rs`.

12. “Agent websocket wait too CPU intensive?”
Handled.
- Added deterministic handshake waiting helper with notification signaling.
- Refs: `codex-rs/core/tests/common/responses.rs`, `codex-rs/core/tests/suite/agent_websocket.rs`.

13. “Why two?” (client websocket test)
Handled.
- Test now validates preconnect reuse semantics directly with handshake/connection assertions.
- Ref: `codex-rs/core/tests/suite/client_websockets.rs`.

14. “Why two? shouldn’t connection be reused?” (fallback)
Handled.
- Fallback tests accept `1..=2` websocket attempts to account for startup preconnect race timing.
- Refs: `codex-rs/core/tests/suite/websocket_fallback.rs`.

15. Review summary (“dedupe connection code / simplify logic”)
Handled.
- Setup/connect deduped and preconnect state model simplified.
- Additional docs pass now calls out invariants/lifecycle explicitly in `core/src/client.rs`.

Validation run:
- `just fmt`
- `cargo test -p codex-core websocket -- --nocapture`
- `cargo test -p codex-core websocket_preconnect -- --nocapture`

If you’d like, I can do one more follow-up to fully remove `web_search_eligible` from the remaining active request header path as well.

codex-rs/core/src/client.rs
joshka-oai: Implemented. I removed the extra preconnect wrapper/matching state (`PreconnectedWebSocket`) and stopped using `turn_metadata_header` as preconnect reuse state. Preconnect now keeps only an optional socket plus minimal sticky turn-state continuity needed for routing.

joshka-oai: Mostly yes now: the socket cache is `Option<ApiWebSocketConnection>`. I kept a separate `preconnected_turn_state` only for sticky routing continuity when adopting a preconnected socket; it is cleared after the turn.

codex-rs/core/src/client.rs
joshka-oai: Addressed by deduplicating setup/connect paths used by both preconnect and normal websocket connect flow (shared auth/provider setup, shared header/connect helpers).

codex-rs/core/src/client.rs
joshka-oai: Done. Preconnect is now single best-effort (no retry loop). Retry behavior remains on the main request path.

codex-rs/core/src/client.rs
joshka-oai: Done. We now adopt the preconnected socket regardless of header mismatch; preconnect reuse is no longer gated on `web_search_eligible` / `turn_metadata_header`.

codex-rs/core/src/client.rs
joshka-oai: Consolidated. Adoption is centralized through `try_use_preconnected_websocket()`, and cleanup uses the shared preconnect clear helper so there is not a second parallel path.

codex-rs/core/src/client.rs
joshka-oai: Implemented. `try_use_preconnected_websocket()` now returns `Option<ApiWebSocketConnection>` and the caller decides assignment instead of mutating internal connection state directly.

codex-rs/core/src/client.rs
joshka-oai: Addressed via equivalent dedupe: fallback cleanup uses the same shared preconnect clear path for slot/state so the behavior is centralized.

codex-rs/core/src/codex.rs
joshka-oai: Partially addressed: preconnect startup no longer depends on `web_search_eligible`. We still include it in active request header construction for now; preconnect logic is no longer coupled to it.

codex-rs/core/src/codex.rs
joshka-oai: Implemented. Startup now does a single client call (`pre_establish_connection(...)`) instead of open-coding setup in `codex.rs`.

codex-rs/core/src/codex.rs
joshka-oai: Implemented. Timeout handling was moved into the client-side pre-establish path so this code is not duplicated in `codex.rs`.

codex-rs/core/tests/suite/agent_websocket.rs
joshka-oai: Updated test assertions to be deterministic around connection behavior (explicit handshake/connection counts) instead of relying on a CPU-heavy wait pattern.

codex-rs/core/tests/suite/client_websockets.rs
joshka-oai: This now intentionally allows `1..=2` websocket attempts in fallback tests to account for the startup preconnect race window. Reuse is still expected in the normal successful preconnect path.

codex-rs/core/tests/suite/websocket_fallback.rs
joshka-oai: Same rationale here: fallback path asserts `1..=2` attempts to tolerate race timing between startup preconnect and immediate main-path connect after a fallback-triggering failure.

codex-rs/core/tests/suite/websocket_fallback.rs
```diff
@@ -46,7 +46,7 @@ async fn websocket_fallback_switches_to_http_after_retries_exhausted() -> Result
         .filter(|req| req.method == Method::POST && req.url.path().ends_with("/responses"))
         .count();
 
-    assert_eq!(websocket_attempts, 1);
+    assert!((1..=2).contains(&websocket_attempts));
```
pakrym-oai: This feels flaky, should we always await the initialization of the first connection? It's probably faster than start a new one one the side.

codex-rs/core/src/client.rs
```diff
@@ -103,6 +112,52 @@ struct ModelClientState {
     include_timing_metrics: bool,
     beta_features_header: Option<String>,
     disable_websockets: AtomicBool,
+    /// Single-slot cache for a warmed websocket that can be adopted by the next turn.
+    ///
+    /// The slot is consumed on read to avoid accidentally sharing one socket across
+    /// multiple concurrent turn sessions.
+    preconnected_websocket: Mutex<Option<ApiWebSocketConnection>>,
```
pakrym-oai: nit: if preconnected_websocket/preconnected_turn_state are consumed together consider grouping them into a struct. 

I hoped we could only store the connection when asking for a single field.

pakrym-oai: Then we can implement debug on that struct too.

codex-rs/core/src/client.rs
```diff
@@ -315,6 +427,178 @@ impl ModelClient {
...
+            Some(manager) => manager.auth().await,
+            None => None,
+        }
+    }
+
+    /// Resolves provider and auth adapter values expected by `codex-api`.
+    ///
+    /// This keeps auth/provider translation in one place so preconnect and request paths cannot
+    /// drift when provider mapping behavior changes.
+    fn provider_and_auth_from_auth(
```
pakrym-oai: Do we need this intermediate helper? only the other helper is using it.

codex-rs/core/src/client.rs
```diff
@@ -315,6 +427,178 @@ impl ModelClient {
...
+            auth,
+            api_provider,
+            api_auth,
+        })
+    }
+
+    /// Reads the latest auth snapshot from session auth manager, if configured.
+    ///
+    /// This is intentionally centralized so all call paths share the same auth fetch timing.
+    async fn current_auth(&self) -> Option<CodexAuth> {
```
pakrym-oai: this is only used by current_client_setup. Is codex creating all these helpers? Do we have something in AGENTS.md?

codex-rs/core/src/client.rs
```diff
@@ -767,6 +1065,14 @@ fn build_api_prompt(prompt: &Prompt, instructions: String, tools_json: Vec<Value
     }
 }
 
+/// Parses per-turn metadata into an HTTP header value.
+///
+/// Invalid values are treated as absent so callers can compare and propagate
+/// metadata with the same sanitization path used when constructing headers.
+fn parse_turn_metadata_header(turn_metadata_header: Option<&str>) -> Option<HeaderValue> {
```
pakrym-oai, line 1167: Can this be inlined?

codex-rs/core/src/client.rs
```diff
@@ -201,6 +258,77 @@ impl ModelClient {
...
 
+    /// Spawns a best-effort task that warms a websocket for the first turn.
+    ///
+    /// This call performs only connection setup; it never sends prompt payloads.
+    ///
+    /// A timeout when computing turn metadata is treated the same as "no metadata" so startup
+    /// cannot block indefinitely on optional preconnect context.
+    pub fn pre_establish_connection(&self, otel_manager: OtelManager, cwd: PathBuf) {
+        let model_client = self.clone();
+        tokio::spawn(async move {
```
pakrym-oai: I just moved the turn metadata building code out of the client :).  Can we somehow avoid duplicating this code from resolve_turn_metadata_header on turn context?

codex-rs/core/src/client.rs
```diff
@@ -315,6 +427,178 @@ impl ModelClient {
...
+            .preconnected_websocket
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        slot.take()
+    }
+
+    /// Stores a freshly preconnected websocket and optional captured turn-state token.
+    ///
+    /// This overwrites any previously warmed socket because only one preconnect candidate is kept.
+    fn store_preconnected_websocket(
```
pakrym-oai, line 580: inline please thre is only one usage

codex-rs/core/src/client.rs
```diff
@@ -315,6 +427,178 @@ impl ModelClient {
...
+            headers.insert(
+                X_RESPONSESAPI_INCLUDE_TIMING_METRICS_HEADER,
+                HeaderValue::from_static("true"),
+            );
+        }
+        headers
+    }
+
+    /// Consumes the warmed websocket slot.
+    fn take_preconnected_websocket(&self) -> Option<ApiWebSocketConnection> {
```
pakrym-oai: Should take_preconnected_websocket and turn_state be merged together?

We can also detele clear_ method that way.

codex-rs/core/src/client.rs
```diff
@@ -332,8 +616,7 @@ impl ModelClientSession {
     }
 
     fn responses_websocket_enabled(&self) -> bool {
```
pakrym-oai: should we remove this intermidiate method and have everyone call the client method?

codex-rs/core/src/client.rs
```diff
@@ -468,36 +750,49 @@ impl ModelClientSession {
...
-            let websocket_telemetry = Self::build_websocket_telemetry(otel_manager);
-            let new_conn: ApiWebSocketConnection =
-                ApiWebSocketResponsesClient::new(api_provider, api_auth)
-                    .connect(
-                        headers,
-                        options.turn_state.clone(),
-                        Some(websocket_telemetry),
-                    )
-                    .await?;
+            self.client.clear_preconnected_websocket();
```
pakrym-oai: at this point try_use_preconnected_websocket should've already cleared the field

codex-rs/core/src/client.rs
```diff
@@ -315,6 +427,178 @@ impl ModelClient {
...
+    }
+
+    /// Returns auth + provider configuration resolved from the current session auth state.
+    ///
+    /// This centralizes setup used by both preconnect and normal request paths so they stay in
+    /// lockstep when auth/provider resolution changes.
+    async fn current_client_setup(&self) -> Result<CurrentClientSetup> {
+        let auth = self.current_auth().await;
+        let (api_provider, api_auth) = self.provider_and_auth_from_auth(auth.clone())?;
+        Ok(CurrentClientSetup {
```
pakrym-oai: so sad we need this. opportunities to simplify

pakrym-oai: later

codex-rs/core/src/client.rs
```diff
@@ -315,6 +427,178 @@ impl ModelClient {
...
+    /// This combines provider capability and feature gating; both must be true for websocket paths
+    /// to be eligible.
+    fn responses_websocket_enabled(&self) -> bool {
+        self.state.provider.supports_websockets && self.state.enable_responses_websockets
+    }
+
+    /// Returns whether websocket transport has been permanently disabled for this session.
+    ///
+    /// Once set by fallback activation, subsequent turns must stay on HTTP transport.
+    fn disable_websockets(&self) -> bool {
```
pakrym-oai, line 485: this is a full duplicate of a method on the session

codex-rs/core/tests/suite/websocket_fallback.rs
pakrym-oai: And we can write more reliable tests too.

pakrym-oai:
A bunch of nits and opportunities to simplify. 

My main question is whether in the case of a race we should await the first connection instead of trying to open another PR.

joshka-oai:
Rebased onto current trunk and force-pushed the bookmark. This update is just a rebase sync.

joshka-oai:
Rebased again onto latest trunk and force-pushed the bookmark. This update is rebase-only (no intended behavior change).

codex-rs/core/src/client.rs
joshka-oai: It's called twice and keeps the logic the same

codex-rs/core/src/client.rs
joshka-oai: yeah - this should probably be a connection struct or something similar to that rather than a CurrentClientSetup.

joshka-oai:
Pushed follow-up commit `b3691277` addressing the remaining simplification nits discussed in review:

- merged preconnect socket + turn-state into a single `PreconnectedWebSocket` slot so adoption is atomic
- removed one-use auth/provider helper layering and duplicate session wrapper methods
- unified turn-metadata timeout handling via `turn_metadata::resolve_turn_metadata_header_with_timeout` (used by both `TurnContext` and startup `pre_establish_connection`)
- kept `clear_preconnected_websocket()` as the shared clear path used by reconnect and fallback (called from two places)

Validation run locally:
- `just fmt`
- `cargo test -p codex-core websocket -- --nocapture`

I replied inline where appropriate. Remaining open design question I still need to investigate: in the startup-preconnect race, should turn execution await the in-flight preconnect connection instead of potentially opening a second websocket.

codex-rs/core/src/client.rs
joshka-oai: done - now calls the same method resolve_turn_metadata_header_with_timeout

codex-rs/core/src/client.rs
joshka-oai: I'm preferential to leaving this to ensure that the lock calls are strongly gated - putting code which locks a mutex in a method prevents the code from accidentally being refactored later into a form where the locks are held longer than necessary (this is a general async / mutex design principle).

codex-rs/core/tests/suite/websocket_fallback.rs
joshka-oai: Implemented in the latest update: first-turn websocket setup now awaits an in-flight startup preconnect and uses that as the first connection attempt, instead of racing a second handshake. This made the affected tests deterministic as well (explicit handshake-count expectations).

codex-rs/core/src/client.rs
joshka-oai: Done. Preconnected websocket and captured turn state are now grouped and consumed atomically via `PreconnectedWebSocket` inside `PreconnectState::Ready`.

codex-rs/core/src/client.rs
joshka-oai: Addressed by reducing one-use helper layering. Remaining helper boundaries are intentional for lock scope and turn-level adoption semantics.

codex-rs/core/src/client.rs
joshka-oai: Simplified helper structure since this comment: removed extra one-use wrappers and consolidated setup through `current_client_setup()`.

codex-rs/core/src/client.rs
joshka-oai: Keeping this helper intentionally. It is used in multiple call sites and centralizes header parsing/sanitization so websocket and HTTP paths stay consistent.

codex-rs/core/src/client.rs
joshka-oai: Keeping this as a lock-scope boundary by design. Encapsulating mutex mutation in one method reduces risk of future refactors accidentally extending lock hold time.

codex-rs/core/src/client.rs
joshka-oai: Implemented. Socket and turn state are merged in `PreconnectedWebSocket` and managed via unified `PreconnectState`; clear path is now `clear_preconnect()`.

codex-rs/core/src/client.rs
joshka-oai: Simplified since this thread: session flow now uses client preconnect-state methods directly where appropriate; remaining `try_use_preconnected_websocket` is a turn-local guard/adoption boundary.

codex-rs/core/src/client.rs
joshka-oai: Now handled through unified `PreconnectState` consumption and `clear_preconnect()`, so state is consumed and cleared in one place.

codex-rs/core/src/client.rs
joshka-oai: Followed up with simplification in this branch: reduced layering and kept `CurrentClientSetup` as the minimal shared auth/provider bundle across preconnect and normal connect paths.

joshka-oai:
Follow-up delta in this push (relative to the previously pushed PR state):

1. First-turn connection behavior
- Switched preconnect tracking to a unified `PreconnectState` (`Idle` / `InFlight` / `Ready`).
- First-turn websocket setup now awaits an in-flight startup preconnect and treats it as the first connection attempt before opening a new handshake.
- Preconnect cleanup is unified via `clear_preconnect()`.

2. Test determinism and expectations
- Added deterministic in-flight preconnect coverage using handshake accept delay (`websocket_first_turn_waits_for_inflight_preconnect`).
- Fallback websocket-attempt expectations were tightened to exact counts in these scenarios (startup preconnect + first turn attempt => 2).

3. Docs-only clarification pass
- Documented the preconnect lifecycle and ownership model.
- Documented that awaiting in-flight preconnect does not introduce a new timeout class vs existing websocket handshake behavior.
- Added module/function docs for shared turn-metadata timeout behavior used by both turn execution and startup preconnect.

Validation run:
- `just fmt`
- `cargo test -p codex-core websocket_preconnect -- --nocapture`
- `cargo test -p codex-core websocket_fallback -- --nocapture`
- `cargo test -p codex-core websocket_first_turn_waits_for_inflight_preconnect -- --nocapture`

pakrym-oai (approved):

joshka-oai:
Added docs for the retry-budget tradeoff and linked fallback tests to it.

Why we currently exclude startup preconnect from `stream_max_retries`:
- `stream_max_retries` is currently modeling retryable turn stream failures.
- Startup preconnect is handshake-only warmup and can happen before any turn payload is sent.
- Counting preconnect against retry budget can consume user-visible retries before first turn work starts, causing earlier fallback than intended.

Consequence of current semantics:
- First-turn failure cases may show 2 websocket handshakes (startup preconnect + turn-time connect) before HTTP fallback becomes sticky.

If we choose differently later:
- We can treat preconnect as an explicit first retry-budgeted attempt, but that requires plumbing connection-attempt accounting from websocket acquisition into the turn retry loop and then updating fallback expectations/tests.

PR #10700, opened 2026-02-05
fix: ensure status indicator present earlier in exec path

sayan-oai:
ensure status indicator present in all classifications of exec tool. fixes indicator disappearing after preambles, will look into using `phase` to avoid this class of error in a few hours.

commands parsed as unknown faced this issue

tested locally, added test for specific failure flow

### DIFF
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -1528,6 +1528,8 @@ impl ChatWidget {
     fn on_exec_command_begin(&mut self, ev: ExecCommandBeginEvent) {
         self.flush_answer_stream_with_separator();
         if is_unified_exec_source(ev.source) {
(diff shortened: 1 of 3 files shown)


sayan-oai:
@codex review rhis

joshka-oai:
Compared with #10701: there is semantic overlap, but they cover different timing windows, so both are useful.

- #10700 restores the status indicator when "exec_command_begin" fires (especially unified-exec / unknown-classification paths).
- #10701 keeps/restores the working shimmer during the preamble idle gap while the task is still running.

Together they address the "looks finished too early" UX across both phases.

I also suggest adding a visual snapshot test for the exec-begin path (in addition to the state/visibility assertion), so we verify the user-visible working affordance and guard against future rendering regressions.

joshka-oai (approved):

joshka-oai (approved):

PR #10705, opened 2026-02-05
chore(config) Default Personality Pragmatic

dylan-hurd-oai:
## Summary
Switch back to Pragmatic personality

## Testing
- [x] Updated unit tests


joshka-oai (approved):
For non test code parts of this, there likely should be a central obvious place where we set the default - either a toml file with defaults, a const name DEFAULT_PERSONALITY, or even Personality::Default. (Not a blocker)

dylan-hurd-oai:
@joshka-oai agree!

PR #10719, opened 2026-02-05
fix(core,app-server) resume with different model

dylan-hurd-oai:
## Summary
When resuming with a different model, we should also append a developer message with the model instructions

## Testing
- [x] Added unit tests


codex-rs/core/src/codex.rs
```diff
@@ -1176,32 +1176,26 @@ impl Session {
...
-                        warn!(
-                            "resuming session with different model: previous={prev}, current={curr}"
-                        );
-                        self.send_event(
-                            &turn_context,
-                            EventMsg::Warning(WarningEvent {
-                                message: format!(
-                                    "This session was recorded with model `{prev}` but is resuming with `{curr}`. \
+                let curr = turn_context.model_info.slug.as_str();
+                if let Some(prev) = Self::last_resumed_model_mismatch(&rollout_items, curr) {
```
pakrym-oai: Do we have to push this through state?

pakrym-oai: As in we have rollout_items.iter() right here, can we grab the model out of them?

dylan-hurd-oai: I think that is still the spirit but I can keep the extraction stateless, yeah

codex-rs/core/src/codex.rs
```diff
@@ -1260,13 +1254,36 @@ impl Session {
         }
     }
 
+    fn last_resumed_model_mismatch<'a>(
```
pakrym-oai:
```suggestion
    fn last_model_name<'a>(
```

codex-rs/core/src/codex.rs
```diff
@@ -2819,8 +2852,24 @@ mod handlers {
         // Attempt to inject input into current task
         if let Err(items) = sess.inject_input(items).await {
             sess.seed_initial_context_if_needed(&current_context).await;
-            let update_items =
+            let mut update_items =
                 sess.build_settings_update_items(previous_context.as_ref(), &current_context);
+            if let Some(previous_model) = sess.take_pending_resume_previous_model().await {
+                let model_switch_already_present =
```
pakrym-oai: should we pass the model name to build_settings_update_items and avoid two places in the same stack that add the same item?

pakrym-oai (approved):

codex-rs/core/src/codex.rs
dylan-hurd-oai: 💡

PR #10721, opened 2026-02-05
[app-server] Add a method to list experimental features.

mzeng-openai:
- [x] Add a method to list experimental features.


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1043,6 +1043,46 @@ pub struct CollaborationModeListResponse {
...
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Default, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ExperimentalFeatureListParams {
+    /// Opaque pagination cursor returned by a previous call.
+    #[ts(optional = nullable)]
+    pub cursor: Option<String>,
+    /// Optional page size; defaults to a reasonable server-side value.
```
jif-oai, line 1053: What is `reasonable` ?

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3050,6 +3058,98 @@ impl CodexMessageProcessor {
...
+                    },
+                )
+                .await;
+            return;
+        }
+
+        let effective_limit = limit.unwrap_or(total as u32).max(1) as usize;
+        let effective_limit = effective_limit.min(total);
+        let start = match cursor {
+            Some(cursor) => match cursor.parse::<usize>() {
```
jif-oai, line 3115: Could you add a small comment explaining the cursor. Ok it's an opaque value but it would be good to have it explicitly defined in the code without having to read the code

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3050,6 +3058,98 @@ impl CodexMessageProcessor {
...
+                    ExperimentalFeatureListResponse {
+                        data: Vec::new(),
+                        next_cursor: None,
+                    },
+                )
+                .await;
+            return;
+        }
+
+        let effective_limit = limit.unwrap_or(total as u32).max(1) as usize;
```
jif-oai, line 3112: I don't understand the `.max(1)` here. Can you comment the logic

jif-oai (commented):
@celia-oai review would be good as well

codex-rs/app-server-protocol/schema/typescript/v2/ExperimentalFeature.ts
```diff
@@ -0,0 +1,29 @@
+// GENERATED CODE! DO NOT MODIFY BY HAND!
+
+// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
+
+export type ExperimentalFeature = { 
+/**
+ * Stable key used in config.toml and CLI flag toggles.
+ */
+flagName: string,
```
owenlin0, line 9: nit: maybe just `name`

owenlin0 (approved):

codex-rs/app-server-protocol/src/protocol/v2.rs
mzeng-openai: I think it's just codex echoing what other comments are saying, but I guess it doesn't matter too much now since we don't really have a ton of features.

codex-rs/app-server-protocol/schema/typescript/v2/ExperimentalFeature.ts
mzeng-openai: I kind of feel flag_name is more accurate, so people know they can use it with --enable

owenlin0:
```
      --enable <FEATURE>                      
          Enable a feature (repeatable). Equivalent to `-c features.<name>=true`
```
let's go with `name` IMO!

mzeng-openai: now that makes sense!

PR #10729, opened 2026-02-05
fix(tui): flush input buffer on init to prevent early exit on Windows

Ashutosh0x:
Fixes #10661.

### Problem
On Windows, the sign-in menu can exit immediately if the OS-level input buffer contains trailing characters (like the Enter key from running the command).

### Solution
**Flush Input Buffer on Init**: Use FlushConsoleInputBuffer on Windows (and 	cflush on Unix) in 	ui::init() to discard any input captured before the TUI was ready.

Verified by @CodebyAmbrose in #10661.

### DIFF
--- a/codex-rs/tui/src/tui.rs
+++ b/codex-rs/tui/src/tui.rs
@@ -214,6 +214,8 @@ pub fn init() -> Result<Terminal> {
     }
     set_modes()?;
 
+    flush_terminal_input_buffer();
+
     set_panic_hook();
 
     let backend = CrosstermBackend::new(stdout());


codex-rs/tui/src/onboarding/onboarding_screen.rs
```diff
@@ -441,6 +441,8 @@ pub(crate) async fn run_onboarding_app(
                     });
                 }
             }
+        } else {
```
etraut-openai: Please remove this change from the PR. This is not needed to fix the bug. We prefer that each PR target one problem only. It allows us to revert changes atomically if there's a problem.

etraut-openai:
@codex review

etraut-openai (changes requested):
Thanks for the PR. Let's keep it targeted to just the input buffer flush.

etraut-openai:
You can ignore the fact that the Bazel builds are failing.

PR #10768, opened 2026-02-05
Add agent job runner for CSV batches

daveaitel-openai:
## Summary
- add sqlite-backed agent jobs (schema, models, runtime)
- add CSV-driven agent job tool handlers and runner
- register tools and document usage

## Testing
- just fmt
- just fix -p codex-core
- just fix -p codex-state
- cargo test -p codex-core
- cargo test -p codex-state
- cargo test --all-features (fails: posix::stopwatch::tests::cancellation_receiver_fires_after_limit)
- cargo test -p codex-exec-server --lib posix::stopwatch::tests::cancellation_receiver_fires_after_limit


codex-rs/core/src/agent/mod.rs
```diff
@@ -5,6 +5,7 @@ pub(crate) mod status;
 
 pub(crate) use codex_protocol::protocol::AgentStatus;
 pub(crate) use control::AgentControl;
+#[cfg(test)]
 pub(crate) use guards::MAX_THREAD_SPAWN_DEPTH;
```
jif-oai, line 9: Drop the import then. We should not use this only for test then

codex-rs/core/src/tools/handlers/collab.rs
```diff
@@ -615,11 +613,6 @@ fn build_agent_spawn_config(
             FunctionCallError::RespondToModel(format!("sandbox_policy is invalid: {err}"))
         })?;
 
-    // If the new agent will be at max depth:
-    if exceeds_thread_spawn_depth_limit(child_depth + 1) {
```
jif-oai, line 619: Why removing this? This is not strictly needed for now but this hopefully prevent nasty issues in the future

codex-rs/docs/agent_jobs.md
```diff
@@ -0,0 +1,102 @@
+# Agent Jobs
```
jif-oai, line 1: We don't maintain the doc in this repo anymore except for the `app-server`. Make sure to also add it here and validate with the DX team: https://github.com/openai/developers-website

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1173 @@
+use crate::agent::exceeds_thread_spawn_depth_limit;
```
jif-oai, line 1: This file is pretty large and I don't see much tests/integrations tests on this feature, could you add some?

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1173 @@
...
+            let pending_items = db
+                .list_agent_job_items(
+                    job_id.as_str(),
+                    Some(codex_state::AgentJobItemStatus::Pending),
+                    Some(slots),
+                )
+                .await?;
+            for item in pending_items {
+                if !db
+                    .mark_agent_job_item_running(job_id.as_str(), item.item_id.as_str())
```
jif-oai, line 689: I _think_ we might have a deadlock here. If the spawned agent does not get spawned because of a max agent error, its state stays "running" for ever

codex-rs/core/src/tools/spec.rs
```diff
@@ -1418,6 +1607,22 @@ pub(crate) fn build_specs(
...
 
+    if config.collab_tools {
+        let agent_jobs_handler = Arc::new(AgentJobsHandler);
+        builder.push_spec(create_spawn_agents_on_csv_tool());
+        builder.push_spec(create_run_agent_job_tool());
+        builder.push_spec(create_get_agent_job_status_tool());
+        builder.push_spec(create_wait_agent_job_tool());
+        builder.push_spec(create_export_agent_job_csv_tool());
+        builder.push_spec(create_report_agent_job_result_tool());
+        builder.register_handler("spawn_agents_on_csv", agent_jobs_handler.clone());
```
jif-oai, line 1618: This adds a lot of new tools to the model, can we run an eval just to make sure it does not confuse it? In general models do not like context growing too much
[go/codex-cli-peaval](https://www.golinks.io/codex-cli-peaval?trackSource=github)

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1173 @@
...
+        let input_path = turn.resolve_path(Some(args.csv_path));
+        let input_path_display = input_path.display().to_string();
+        let csv_content = tokio::fs::read_to_string(&input_path)
+            .await
+            .map_err(|err| {
+                FunctionCallError::RespondToModel(format!(
+                    "failed to read csv input {input_path_display}: {err}"
+                ))
+            })?;
+        let (headers, rows) = parse_csv(csv_content.as_str()).map_err(|err| {
```
jif-oai, line 204: We should be more defensive here. What if there are some duplicates in the column names for example? 
I think we need a strong validation on this

codex-rs/core/src/tools/spec.rs
```diff
@@ -484,6 +482,226 @@ fn create_spawn_agent_tool() -> ToolSpec {
...
+        JsonSchema::Boolean {
+            description: Some(
+                "Whether to auto-export CSV output on successful completion (default true)."
+                    .to_string(),
+            ),
+        },
+    );
+    ToolSpec::Function(ResponsesApiTool {
+        name: "spawn_agents_on_csv".to_string(),
+        description: "Create and run a batch agent job over a CSV file.".to_string(),
```
jif-oai, line 545: The model has no idea what a "batch agent job is". We either need to be more descriptive (preferred) or to update the main prompt. Otherwise the usage of this tool by the model will be chaotic

codex-rs/core/src/tools/spec.rs
```diff
@@ -484,6 +482,226 @@ fn create_spawn_agent_tool() -> ToolSpec {
...
+        JsonSchema::Boolean {
+            description: Some(
+                "Whether to auto-export CSV output on successful completion (default true)."
+                    .to_string(),
+            ),
+        },
+    );
+    ToolSpec::Function(ResponsesApiTool {
+        name: "run_agent_job".to_string(),
+        description: "Start or resume execution of an existing agent job.".to_string(),
```
jif-oai, line 583: Does it means we always need to start after creation? Why not auto-start? (open questions)

codex-rs/core/src/tools/spec.rs
```diff
@@ -484,6 +482,226 @@ fn create_spawn_agent_tool() -> ToolSpec {
...
+fn create_export_agent_job_csv_tool() -> ToolSpec {
+    let mut properties = BTreeMap::new();
+    properties.insert(
+        "job_id".to_string(),
+        JsonSchema::String {
+            description: Some("Identifier of the job to export.".to_string()),
+        },
+    );
+    properties.insert(
+        "path".to_string(),
```
jif-oai, line 650: We should keep this name consistent with `output_csv_path`
And why would it have a different value?

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1173 @@
...
+
+        let job_id = Uuid::new_v4().to_string();
+        let output_csv_path = args.output_csv_path.map_or_else(
+            || default_output_csv_path(input_path.as_path(), job_id.as_str()),
+            |path| turn.resolve_path(Some(path)),
+        );
+        let job_suffix = &job_id[..8];
+        let job_name = args
+            .job_name
+            .unwrap_or_else(|| format!("agent-job-{job_suffix}"));
```
jif-oai, line 270: Side not but it would be cool to generate a name

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1173 @@
...
+            body: FunctionCallOutputBody::Text(content),
+            success: Some(true),
+        })
+    }
+}
+
+mod run_agent_job {
+    use super::*;
+
+    pub async fn handle(
```
jif-oai, line 316: When a resume occurs, we can expect that the previous session was non-gracefully closed. I think we should reset all the agent job item that are "running" to pending in this case

jif-oai (commented):
From what I understand, the workflow is the following:
1. You prompt the model to create an "agent job" with all the parameters you want (max concurrency, paths, ...)
2. The model calls `spawn_agents_on_csv` with all those parameters
3. The model monitor the job and when it's done it either call `export_agent_job_csv` or there is an auto-export
4. The model report through `report_agent_job_result`

If Codex get shutdown, you can ask the agent to resume an "agent job"

**But** I see a few issues:
1. Why does the model needs to monitor the job through `get_agent_job_status` or `wait_agent_job`? In practice the model can't do anything on the job while it's running so why not just having something deterministic here?
2. `run_agent_job` should be very specific to resume IMO. The job seems (or should) to auto-start when we call `spawn_agents_on_csv` so "run" is not required by default
3. What is the point of `export_agent_job_csv`? Why not always having auto-export enabled? In we are in the happy path, the user won't be able to send any information to the model anyway (because the model is monitoring the job). So all the informations should be known at startup time? Curious to see any use cases you might have though
4. The resume feature seems pretty hard to use because the user never has the `job_id` so it can't prompt a new session to resume an "agent job". It means you need to keep the same session which sounds like a big limitation
5. Why having `report_agent_job_result` instead of an end of turn or even better, a programmatic tool?

In general I don't think the interface here is the best one. I love the idea of using an agent to figure out how to spawn a bunch of sub-agents but IMO the rest of the workflow should be handled by deterministic code if the agent can't take actions on the "agent job" anyway

PR #10770, opened 2026-02-05
Load requirements on windows

gt-oai:
We support requirements on Unix, loading from `/etc/codex/requirements.toml`. On MacOS, we also support MDM. 

Now, on Windows, we'll load requirements from `%ProgramData%\OpenAI\Codex\requirements.toml`

### DIFF
--- a/codex-rs/core/src/config_loader/mod.rs
+++ b/codex-rs/core/src/config_loader/mod.rs
@@ -29,6 +29,8 @@ use dunce::canonicalize as normalize_path;
 use serde::Deserialize;
 use std::io;
 use std::path::Path;
+#[cfg(windows)]
+use std::path::PathBuf;
 use toml::Value as TomlValue;
 
 pub use cloud_requirements::CloudRequirementsLoader;
@@ -61,13 +63,13 @@ pub use state::ConfigLayerStack;
 pub use state::ConfigLayerStackOrdering;
 pub use state::LoaderOverrides;
 
-/// On Unix systems, load requirements from this file path, if present.
-const DEFAULT_REQUIREMENTS_TOML_FILE_UNIX: &str = "/etc/codex/requirements.toml";
-
 /// On Unix systems, load default settings from this file path, if present.
 /// Note that /etc/codex/ is treated as a "config folder," so subfolders such
 /// as skills/ and rules/ will also be honored.
 pub const SYSTEM_CONFIG_TOML_FILE_UNIX: &str = "/etc/codex/config.toml";
+
+#[cfg(windows)]
+const DEFAULT_PROGRAM_DATA_DIR_WINDOWS: &str = r"C:\ProgramData";
 
 const DEFAULT_PROJECT_ROOT_MARKERS: &[&str] = &[".git"];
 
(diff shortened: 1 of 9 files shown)


codex-rs/core/src/config_loader/mod.rs
```diff
@@ -386,6 +379,41 @@ async fn load_requirements_toml(
...
+fn system_requirements_toml_file() -> Option<PathBuf> {
+    #[cfg(unix)]
+    {
+        Some(PathBuf::from("/etc/codex/requirements.toml"))
+    }
+
+    #[cfg(windows)]
+    {
+        Some(windows_system_requirements_toml_file(
+            std::env::var_os("ProgramData").as_deref(),
```
bolinfest: Doesn't this mean the user could set `%ProgramData%` themselves when invoking Codex to point to folder they can write to and then load `requirements.toml` from there?

bolinfest: ChatGPT said to use this win32 API:

```c
SHGetKnownFolderPath(FOLDERID_ProgramData, ...)
```

or in our case, this Rust equivalent:

```rust
use std::path::PathBuf;
use windows::{
    core::PWSTR,
    Win32::{
        Foundation::HANDLE,
        System::Com::CoTaskMemFree,
        UI::Shell::{FOLDERID_ProgramData, SHGetKnownFolderPath, KF_FLAG_DEFAULT},
    },
};

pub fn programdata_dir() -> windows::core::Result<PathBuf> {
    unsafe {
        let mut path: PWSTR = PWSTR::null();

        // If your process has COM initialized already, great.
        // SHGetKnownFolderPath does not require you to call CoInitialize on its own,
        // but it's common for apps to initialize COM elsewhere.
        SHGetKnownFolderPath(
            &FOLDERID_ProgramData,
            KF_FLAG_DEFAULT,
            HANDLE(0), // current user token; ProgramData is machine-wide anyway
            &mut path,
        )?;

        // Copy into an owned Rust string, then free the COM-allocated buffer.
        let s = path.to_string()?;
        CoTaskMemFree(Some(path.0 as *const _));

        Ok(PathBuf::from(s))
    }
}
```

Though probably worth double-checking that...

codex-rs/core/src/config_loader/mod.rs
```diff
@@ -386,6 +391,83 @@ async fn load_requirements_toml(
...
+    {
+        Some(PathBuf::from("/etc/codex/requirements.toml"))
+    }
+
+    #[cfg(windows)]
+    {
+        Some(windows_system_requirements_toml_file())
+    }
+
+    #[cfg(not(any(unix, windows)))]
```
bolinfest: I would remove this bit and change the signature of this function to be:

```rust
fn system_requirements_toml_file() -> std::io::Result<AbsolutePathBuf>
```

which means pushing the `AbsolutePathBuf::try_from` into this function.

If someone tries to compile Codex CLI on something that is neither UNIX nor Windows, then we'll get a compilation error here, which I think is fine for now. (Incidentally, UNIX covers Android, which some of our users target.)

codex-rs/core/src/config_loader/mod.rs
```diff
@@ -27,9 +27,22 @@ use codex_utils_absolute_path::AbsolutePathBuf;
 use codex_utils_absolute_path::AbsolutePathBufGuard;
 use dunce::canonicalize as normalize_path;
 use serde::Deserialize;
+#[cfg(windows)]
```
bolinfest: Note I think it would be a bit cleaner to push these import statements into `windows_program_data_dir_from_known_folder()`, where possible.

codex-rs/core/src/config_loader/mod.rs
```diff
@@ -386,6 +391,83 @@ async fn load_requirements_toml(
...
+    {
+        None
+    }
+}
+
+fn system_config_toml_file() -> Option<PathBuf> {
+    if cfg!(unix) {
+        Some(PathBuf::from(SYSTEM_CONFIG_TOML_FILE_UNIX))
+    } else {
+        None
```
bolinfest: We should add Windows support for this too, right?

bolinfest (approved):
Thanks for pulling in the Windows API stuff!

PR #10787, opened 2026-02-05
chore: limit update to 0.98.0 NUX to < 0.98.0 ver

sayan-oai:
seems like footgun if we forget to remove before releasing 0.99.0, limited announcement to versions < 0.98.0


jif-oai (approved):
I guess you're sure about the regex?

PR #10792, opened 2026-02-05
Print warning when config does not meet requirements

gt-oai:
Screenshot 2026-02-05 at 23 34 08

Currently, if a config value is set that fails the requirements, we exit Codex.

Now, instead of this, we print a warning and default to a requirements-permitting value.

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -946,6 +946,14 @@ impl Session {
                 }),
             });
         }
+        for message in &config.startup_warnings {
+            post_session_configured_events.push(Event {
+                id: "".to_owned(),
+                msg: EventMsg::Warning(WarningEvent {
+                    message: message.clone(),
+                }),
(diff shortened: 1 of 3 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -652,11 +652,21 @@ impl SessionConfiguration {
         if let Some(personality) = updates.personality {
             next_configuration.personality = Some(personality);
         }
-        if let Some(approval_policy) = updates.approval_policy {
-            next_configuration.approval_policy.set(approval_policy)?;
+        if let Some(approval_policy) = updates.approval_policy
```
xl-openai: What's this path used for?

xl-openai (approved):

codex-rs/core/src/codex.rs
gt-oai: reverted this

PR #10809, opened 2026-02-05
updates: use brew api for version check

magus:
## Problem

`codex` currently prompts you to update via `brew upgrade --cask codex` but the brew api does not return the new version

> Screenshot 2026-02-05 at 12 36 09 PM

## Solution

`codex-rs/tui/src/updates.rs` was using the [latest cask in github](https://github.com/Homebrew/homebrew-cask/blob/HEAD/Casks/c/codex.rb) but this does not agree with the brew api, which leads to the issue above. Instead we use the [brew api json endpoint](https://github.com/Homebrew/homebrew-cask/blob/HEAD/Casks/c/codex.rb) to ensure our version check agrees with the upgrade command.


etraut-openai:

shijie-oai (approved):

PR #10830, opened 2026-02-06
Fix missed mid-stream auto-compaction checks for multi-item responses

charley-oai:
## Summary
This PR fixes a failure mode where auto-compaction was checked only after `run_sampling_request` returned, which is too late when a single streamed response contains multiple tool calls/outputs.

In that case, large tool outputs can accumulate in history inside one response, compaction is delayed, and remote compaction can fail with `context_length_exceeded`.

## Root cause
- Post-sampling token usage logging and auto-compaction checks were only run after `run_sampling_request(...)` returned `Ok` in `run_turn`.
- Streaming responses can include multiple `response.output_item.done` tool calls before `response.completed`.
- Remote trim logic only removed trailing codex-generated items, so a trailing tool call item could block further trimming of oversized history.

### 1) Mid-stream token checkpoints + compaction checks
- Added `log_post_sampling_token_usage_and_maybe_compact(...)` in `core/src/codex.rs`.
- Reused it from:
  - `run_turn` (existing end-of-request check path), and
  - `try_run_sampling_request` after each `ResponseEvent::OutputItemDone`, plus
  - `drain_in_flight` after each tool output is recorded.
- Check now considers both:
  - `total_usage_tokens >= auto_compact_limit`, and
  - `estimated_token_count >= auto_compact_limit`.
- Logging keeps the same `"post sampling token usage"` message and adds a `checkpoint` field to show where the check fired (`run_turn`, `stream_item_done`, `stream_item_done_plan`, `drain_in_flight`).

### 2) Safety guard: do not compact while tool outputs are still pending
- Added `allow_auto_compact` gating so compaction runs only when there are no pending in-flight tool outputs in the current streaming request.
- This avoids orphan `function_call_output` history entries when compaction replaces history before later tool outputs are recorded.

### 3) Remote trim widened for trailing tool-call items (with pending-call protection)
- In `core/src/compact_remote.rs`, remote pre-compaction trimming can still trim trailing tool-call history when appropriate:
  - `function_call`
  - `custom_tool_call`
  - `local_shell_call`
- Added a safety check to stop trimming when the trailing tool call does not yet have a matching output in history.
- This avoids dropping pending calls that could later produce outputs and become orphans, while preserving trim behavior for completed call/output history.

### 4) Regression tests
- Added/updated `remote_auto_compact_runs_after_multi_tool_output_drain` to verify compaction with multi-tool single-response flows and ensure safe behavior around drained outputs.
- Added a unit test in `compact_remote.rs` to verify trailing pending `function_call` items are preserved until output exists.

## Why this is safe
- Auto-compaction still only runs when `needs_follow_up` is true.
- Mid-stream checkpoints still run and log token state, but compaction waits for safe points when tool outputs are fully drained.
- Existing end-of-response behavior is preserved.
- The new trim candidates are limited to tool-call items that are generated during model/tool interaction, matching the existing intent of trimming transient call/output history.

## Codex author
`codex fork 019c3054-f072-79f1-bfd6-dc767d6d6d33`

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -3759,6 +3759,7 @@ pub(crate) async fn run_turn(
                     &turn_context,
                     auto_compact_limit,
                     needs_follow_up,
+                    true,
                     "run_turn",
                 )
                 .await;
@@ -3836,6 +3837,7 @@ async fn log_post_sampling_token_usage_and_maybe_compact(
     turn_context: &Arc<TurnContext>,
     auto_compact_limit: i64,
     needs_follow_up: bool,
+    allow_auto_compact: bool,
     checkpoint: &'static str,
 ) -> bool {
     let total_usage_tokens = sess.get_total_token_usage().await;
@@ -3850,11 +3852,12 @@ async fn log_post_sampling_token_usage_and_maybe_compact(
         auto_compact_limit,
         token_limit_reached,
         needs_follow_up,
+        allow_auto_compact,
         checkpoint,
         "post sampling token usage"
     );
 
-    if token_limit_reached && needs_follow_up {
+    if allow_auto_compact && token_limit_reached && needs_follow_up {
         run_auto_compact(sess, turn_context).await;
     }
 
@@ -4500,6 +4503,7 @@ async fn drain_in_flight(
                     &turn_context,
                     auto_compact_limit,
                     true,
+                    in_flight.is_empty(),
                     "drain_in_flight",
                 )
                 .await;
@@ -4653,6 +4657,7 @@ async fn try_run_sampling_request(
                             &turn_context,
                             auto_compact_limit,
                             needs_follow_up,
+                            in_flight.is_empty(),
                             "stream_item_done_plan",
                         )
                         .await;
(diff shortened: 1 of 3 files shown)


charley-oai:
@codex review this

codex-rs/core/src/codex.rs
```diff
@@ -3876,6 +3831,36 @@ async fn run_auto_compact(sess: &Arc<Session>, turn_context: &Arc<TurnContext>)
...
+    sess: &Arc<Session>,
+    turn_context: &Arc<TurnContext>,
+    auto_compact_limit: i64,
+    needs_follow_up: bool,
+    checkpoint: &'static str,
+) -> bool {
+    let total_usage_tokens = sess.get_total_token_usage().await;
+    let estimated_token_count = sess.get_estimated_token_count(turn_context.as_ref()).await;
+    let token_limit_reached = total_usage_tokens >= auto_compact_limit
+        || estimated_token_count.is_some_and(|count| count >= auto_compact_limit);
```
charley-oai, line 3846: This used to be
```rust
let total_usage_tokens = sess.get_total_token_usage().await;
let token_limit_reached = total_usage_tokens >= auto_compact_limit;

let estimated_token_count =
    sess.get_estimated_token_count(turn_context.as_ref()).await;
```

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

PR #10843, opened 2026-02-06
Log an event (info only) when we receive a file watcher event

etraut-openai:


xl-openai (approved):

codex-rs/core/src/file_watcher.rs
```diff
@@ -162,6 +163,12 @@ impl FileWatcher {
                         res = raw_rx.recv() => {
                             match res {
                                 Some(Ok(event)) => {
+                                    info!(
```
xl-openai, line 166: If we’re concerned about generating too many logs, maybe we can randomly sample just a subset of them. I’ll leave that up to you.

PR #10855, opened 2026-02-06
TUI/Core: preserve duplicate skill/app mention selection across submit + resume

daniel-oai:
## What changed

- In `codex-rs/core/src/skills/injection.rs`, we now honor explicit `UserInput::Skill { name, path }` first, then fall back to text mentions only when safe.
- In `codex-rs/tui/src/bottom_pane/chat_composer.rs`, mention selection is now token-bound (selected mention is tied to the specific inserted `$token`), and we snapshot bindings at submit time so selection is not lost.
- In `codex-rs/tui/src/chatwidget.rs` and `codex-rs/tui/src/bottom_pane/mod.rs`, submit/queue paths now consume the submit-time mention snapshot (instead of rereading cleared composer state).
- In `codex-rs/tui/src/mention_codec.rs` and `codex-rs/tui/src/bottom_pane/chat_composer_history.rs`, history now round-trips mention targets so resume restores the same selected duplicate.
- In `codex-rs/tui/src/bottom_pane/skill_popup.rs` and `codex-rs/tui/src/bottom_pane/chat_composer.rs`, duplicate labels are normalized to `[Repo]` / `[App]`, app rows no longer show `Connected -`, and description space is a bit wider.

Screenshot 2026-02-05 at 9 56 56 PM
Screenshot 2026-02-05 at 9 57 09 PM

## Before vs now

- Before: selecting a duplicate could still submit the default/repo match, and resume could lose which duplicate was originally selected.
- Now: the exact selected target (skill path or app id) is preserved through submit, queue/restore, and resume.

## Manual test

1. Build and run this branch locally:
   - `cd /Users/daniels/code/codex/codex-rs`
   - `./target/debug/codex`
2. Open mention picker with `$` and pick a duplicate entry (not the first one).
3. Confirm duplicate UI:
   - repo duplicate rows show `[Repo]`
   - app duplicate rows show `[App]`
   - app description does **not** start with `Connected -`
4. Submit the prompt, then press Up to restore draft and submit again.  
   Expected: it keeps the same selected duplicate target.
5. Use `/resume` to reopen the session and send again.  
   Expected: restored mention still resolves to the same duplicate target.


codex-rs/core/src/skills/injection.rs
```diff
@@ -78,14 +78,17 @@ fn emit_skill_injected_metric(otel: Option<&OtelManager>, skill: &SkillMetadata,
     );
 }
 
-/// Collect explicitly mentioned skills from `$name` text mentions.
+/// Collect explicitly mentioned skills from structured and text mentions.
 ///
-/// Text inputs are scanned once to extract `$skill-name` tokens, then we iterate `skills`
-/// in their existing order to preserve prior ordering semantics. Explicit links are
-/// resolved by path and plain names are only used when the match is unambiguous.
+/// Structured `UserInput::Skill` selections are resolved first by path against
```
xl-openai, line 83: Nice!

codex-rs/core/src/skills/injection.rs
```diff
@@ -299,6 +324,9 @@ fn select_skills_from_mentions(
         if !mentions.plain_names.contains(skill.name.as_str()) {
             continue;
         }
+        if blocked_plain_names.contains(skill.name.as_str()) {
```
xl-openai: Let's check this ahead of the plain_names.contains check.

codex-rs/core/src/skills/injection.rs
daniel-oai: ❤️

codex-rs/tui/src/bottom_pane/chat_composer.rs
```diff
@@ -309,6 +311,12 @@ struct FooterFlash {
     expires_at: Instant,
 }
 
+#[derive(Clone, Debug)]
+struct ComposerMentionBinding {
```
xl-openai, line 315: @codex can you help me understand how we use MentionBinding? More specially:
1. When do we set MentionBinding;
2. How do we clear MentionBinding.

xl-openai (approved):

codex-rs/core/src/skills/injection.rs
daniel-oai: Applied in 7b8e55de2: I now check `blocked_plain_names` before `mentions.plain_names.contains(...)` in `select_skills_from_mentions`, as suggested.

PR #10861, opened 2026-02-06
Support alternative websocket API

by-openai:
**Test plan**

```
cargo build -p codex-cli && RUST_LOG='codex_api::endpoint::responses_websocket=trace,codex_core::client=debug,codex_core::codex=debug' \
  ./target/debug/codex \
    --enable responses_websockets_v2 \
    --profile byok \
    --full-auto
```

### DIFF
--- a/codex-rs/core/src/client.rs
+++ b/codex-rs/core/src/client.rs
@@ -89,6 +89,8 @@ use reqwest::StatusCode;
 use serde_json::Value;
 use std::time::Duration;
 use tokio::sync::mpsc;
+use tokio::sync::oneshot;
+use tokio::sync::oneshot::error::TryRecvError;
 use tokio::task::JoinHandle;
 use tokio_tungstenite::tungstenite::Error;
 use tokio_tungstenite::tungstenite::Message;
@@ -117,6 +119,7 @@ pub const X_CODEX_TURN_STATE_HEADER: &str = "x-codex-turn-state";
 pub const X_CODEX_TURN_METADATA_HEADER: &str = "x-codex-turn-metadata";
 pub const X_RESPONSESAPI_INCLUDE_TIMING_METRICS_HEADER: &str =
     "x-responsesapi-include-timing-metrics";
+const RESPONSES_WEBSOCKETS_V2_BETA_HEADER_VALUE: &str = "responses_websockets=2026-02-06";
 
 /// Session-scoped state shared by all [`ModelClient`] clones.
 ///
@@ -129,6 +132,7 @@ struct ModelClientState {
     session_source: SessionSource,
     model_verbosity: Option<VerbosityConfig>,
     enable_responses_websockets: bool,
+    enable_responses_websockets_v2: bool,
     enable_request_compression: bool,
     include_timing_metrics: bool,
     beta_features_header: Option<String>,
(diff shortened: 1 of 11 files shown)


pakrym-oai (approved):

codex-rs/core/src/client.rs
```diff
@@ -466,6 +481,54 @@ impl ModelClientSession {
...
     }
 
+    fn prepare_websocket_request(
+        &self,
+        model_slug: &str,
+        api_prompt: &ApiPrompt,
+        options: &ApiResponsesOptions,
+    ) -> ResponsesWsRequest {
+        let incremental_items = self.get_incremental_items(&api_prompt.input);
+        if self.responses_websockets_v2_enabled() {
```
pakrym-oai: nit: can we move the incremental logic into existing `        if let Some(append_items) = incremental_items {` branch and then keep only a single non-incremental self.prepare_websocket_create_request call site?

codex-rs/core/src/client.rs
```diff
@@ -479,8 +542,15 @@ impl ModelClientSession {
         };
 
         if needs_new {
-            let headers =
+            self.websocket_last_items.clear();
+            if let Ok(mut last_response_id) = self.websocket_last_response_id.lock() {
```
pakrym-oai: can we just assign?

codex-rs/core/src/client.rs
```diff
@@ -479,8 +542,15 @@ impl ModelClientSession {
...
+                *last_response_id = None;
+            }
+            let mut headers =
                 build_websocket_connect_headers(options, self.client.state.include_timing_metrics);
+            if self.responses_websockets_v2_enabled() {
```
pakrym-oai: consider moving this to responses_websockets_v2_enabled

codex-rs/core/src/client.rs
```diff
@@ -142,6 +144,7 @@ pub struct ModelClientSession {
     client: ModelClient,
     connection: Option<ApiWebSocketConnection>,
     websocket_last_items: Vec<ResponseItem>,
+    websocket_last_response_id: Arc<std::sync::Mutex<Option<String>>>,
```
pakrym-oai: this feels like an overkill. ModelClientSession isn't shared across threads. Does Option<String> work here?

codex-rs/core/src/client.rs
```diff
@@ -802,6 +881,24 @@ fn build_websocket_connect_headers(
     headers
 }
 
+fn append_openai_beta_header(headers: &mut ApiHeaderMap, beta_value: &'static str) {
```
pakrym-oai: you might not need this if you colocate beta header setting code.

by-openai:

codex-rs/core/src/client.rs
by-openai: removed!

codex-rs/core/src/client.rs
by-openai: looks like the callback that scans the result stream and sees the completed event doesn't have a mutable reference to the `ModelClientSession` -- removed the old `Arc<Mutex<_>>` and switched to a per-request oneshot sender/receiver, which should be lighter and also prevents stale late completions from older streams from overwriting newer state

PR #10863, opened 2026-02-06
Parse skill/connector mentions in skills/automations

canvrno-oai:
This PR introduces a recursive skill-expansion mechanism for mentions so nested or chained skill references are expanded using the same selection semantics across multiple levels in skills and automations. This keeps behavior aligned with existing mention handling while extending coverage to deeper scenarios. With these changes, users can create automations that use skills, and these skills can use other skills.

### Skill & conector mentions scenarios

1. **First-order + Primary**
   - **Context path:** Skill A `SKILL.md` `$mention`
   - Direct skill resolution and execution from top-level `SKILL.md`.
   - Tool triggers with expected role/output and no missing-file errors.

2. **First-order + Secondary**
   - **Context path:** `SKILL.md` → Secondary file (`references.md`, etc.) `$mention`
   - Parsing and selecting skills from non-primary skill files.
   - Tool triggers correctly from secondary skill context.

3. **Second-order + Primary**
   - **Context path:** Skill A `SKILL.md` → Skill B `$mention`
   - Transitive skill execution through nested mentions.
   - Skill B mention resolves and runs from second-order primary context.

4. **Second-order + Secondary**
   - **Context path:** Skill A `SKILL.md` → Secondary file → Skill B Secondary file (`references.md`, etc.) `$mention`
   - Transitive mentions discovered through nested secondary contexts.
   - Skill B mention resolves and runs from second-order secondary context.


mzeng-openai:
This is clever! Have we tested things like:

- A SKILL.md references other .md files (not necessarily SKILL.md) and $app-name is in the reference file
- Circular dependency from SKILL.md and it references?

canvrno-oai:
> This is clever! Have we tested things like:
> 
> * A SKILL.md references other .md files (not necessarily SKILL.md) and $app-name is in the reference file
> * Circular dependency from SKILL.md and it references?

I've tested the following scenarios:

- Skill > Skill 
  - $skill-A is invoked
  - $skill-A prompting directs the agent to a reference file in $skill-A
  - The reference file in $skill-A directs to $skill-B
  - $skill-B prompting directs the agent to a reference/nested.md file in $skill-B
   - The reference.md file in $skill-A directs to return a specific message format to the user
 
 - Skill > Connector 
  - $skill-A is invoked
  - $skill-A prompting directs the agent to use a $connector
  
  However, Skill > Skill > Connector is not working as desired, so I will see if I can get that working as well as craft a detailed test plan and test environments.

PR #10909, opened 2026-02-06
Bootstrap shell commands via user shell snapshot

jif-oai:
Summary
- wrap `shell -lc` executions that use a snapshot with the session shell so the saved environment is sourced before delegating to the original shell
- escape single quotes in the generated wrapper and add tests covering Bash/Zsh/sh session bootstrapping

Testing
- Not run (not requested)

### DIFF
--- a/codex-rs/core/src/tools/runtimes/mod.rs
+++ b/codex-rs/core/src/tools/runtimes/mod.rs
@@ -47,7 +47,7 @@ pub(crate) fn build_command_spec(
 /// the original script:
 ///
 ///   shell -lc "<script>"
-///   => user_shell -c ". SNAPSHOT (best effort); exec shell -lc <script>"
+///   => user_shell -c ". SNAPSHOT (best effort); exec shell -c <script>"
 ///
 /// This wrapper script uses POSIX constructs (`if`, `.`, `exec`) so it can
 /// be run by Bash/Zsh/sh. On non-matching commands this is a no-op.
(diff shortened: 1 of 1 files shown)


jif-oai:
@codex review

jif-oai:
@codex review

jif-oai:
@codex review

codex-rs/core/src/tools/runtimes/mod.rs
```diff
@@ -72,10 +73,148 @@ pub(crate) fn maybe_wrap_shell_lc_with_snapshot(
     }
 
     let snapshot_path = snapshot.path.to_string_lossy();
-    let rewritten_script = format!(". \"{snapshot_path}\" && {}", command[2]);
+    let shell_path = session_shell.shell_path.to_string_lossy();
+    let original_shell = shell_single_quote(&command[0]);
+    let original_script = shell_single_quote(&command[2]);
+    let snapshot_path = shell_single_quote(snapshot_path.as_ref());
```
dylan-hurd-oai, line 79: this seems reasonable but curious about introducing a specific quoting strategy here

dylan-hurd-oai (approved):

jif-oai:
@codex review

codex-rs/core/src/tools/runtimes/mod.rs
jif-oai: wdym?

PR #10916, opened 2026-02-06
requirements: add additional dev instructions

gt-oai:


gt-oai:
Didn't need this in the end

PR #10927, opened 2026-02-06
Treat compaction failure as failure state

aibrahim-oai:
- Return compaction errors from local and remote compaction flows.\n- Stop turns/tasks when auto-compaction fails instead of continuing execution.

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -3654,8 +3654,10 @@ pub(crate) async fn run_turn(
         collaboration_mode_kind: turn_context.collaboration_mode.mode,
     });
     sess.send_event(&turn_context, event).await;
-    if total_usage_tokens >= auto_compact_limit {
-        run_auto_compact(&sess, &turn_context).await;
+    if total_usage_tokens >= auto_compact_limit
+        && let Err(e) = run_auto_compact(&sess, &turn_context).await
+    {
+        return Some(e.to_string());
     }
 
     let skills_outcome = Some(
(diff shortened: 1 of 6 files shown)


charley-oai:
@codex review this

codex-rs/core/src/codex.rs
```diff
@@ -3846,7 +3839,9 @@ pub(crate) async fn run_turn(
 
                 // as long as compaction works well in getting us way below the token limit, we shouldn't worry about being in an infinite loop.
                 if token_limit_reached && needs_follow_up {
-                    run_auto_compact(&sess, &turn_context).await;
+                    if let Err(e) = run_auto_compact(&sess, &turn_context).await {
+                        return Some(e.to_string());
```
charley-oai: Are you sure we want to return `Some(e.to_string());` and not just break the loop here?

codex-rs/core/src/codex.rs
```diff
@@ -3663,8 +3654,10 @@ pub(crate) async fn run_turn(
         collaboration_mode_kind: turn_context.collaboration_mode.mode,
     });
     sess.send_event(&turn_context, event).await;
-    if total_usage_tokens >= auto_compact_limit {
-        run_auto_compact(&sess, &turn_context).await;
+    if total_usage_tokens >= auto_compact_limit
+        && let Err(e) = run_auto_compact(&sess, &turn_context).await
+    {
+        return Some(e.to_string());
```
charley-oai: Why don't we return `None` here?

aibrahim-oai:
@codex review

codex-rs/core/src/tasks/compact.rs
```diff
@@ -31,14 +31,24 @@ impl SessionTask for CompactTask {
                 1,
                 &[("type", "remote")],
             );
-            crate::compact_remote::run_remote_compact_task(session, ctx).await
+            if crate::compact_remote::run_remote_compact_task(session, ctx)
+                .await
+                .is_err()
+            {
+                return None;
+            }
```
charley-oai: I think in this file you can actually just do `crate::compact_remote::run_remote_compact_task(session, ctx).await`, no need to return None upon error here because we just return None anyway.

aibrahim-oai:
@codex review

charley-oai (approved):

PR #10934, opened 2026-02-06
Fail agent loop eagerly on auto compaction failure

charley-oai:
Previously we just emitted a UI error event; this may be an artifact of the `/compact` days where users could run `/compact` and keep going if it failed


charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

PR #10935, opened 2026-02-06
Agent jobs (spawn_agents_on_csv) + progress UI

daveaitel-openai:
## Summary
- Add agent job support: spawn a batch of sub-agents from CSV, auto-run, auto-export, and store results in SQLite.
- Simplify workflow: remove run/resume/get-status/export tools; spawn is deterministic and completes in one call.
- Improve exec UX: stable, single-line progress bar with ETA; suppress sub-agent chatter in exec.

## Why
Enables map-reduce style workflows over arbitrarily large repos using the existing Codex orchestrator. This addresses review feedback about overly complex job controls and non-deterministic monitoring.

## Demo (progress bar)
```
./codex-rs/target/debug/codex exec \
  --enable collab \
  --enable sqlite \
  --full-auto \
  --progress-cursor \
  -c agents.max_threads=16 \
  -C /Users/daveaitel/code/codex \
  - <<'PROMPT'
Create /tmp/agent_job_progress_demo.csv with columns: path,area and 30 rows:
path = item-01..item-30, area = test.

Then call spawn_agents_on_csv with:
- csv_path: /tmp/agent_job_progress_demo.csv
- instruction: "Run `python - <<'PY'` to sleep a random 0.3–1.2s, then output JSON with keys: path, score (int). Set score = 1."
- output_csv_path: /tmp/agent_job_progress_demo_out.csv
PROMPT
```

## Review feedback addressed
- Auto-start jobs on spawn; removed run/resume/status/export tools.
- Auto-export on success.
- More descriptive tool spec + clearer prompts.
- Avoid deadlocks on spawn failure; pending/running handled safely.
- Progress bar no longer scrolls; stable single-line redraw.

## Tests
- `cd codex-rs && cargo test -p codex-exec`
- `cd codex-rs && cargo build -p codex-cli`


codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1012 @@
...
+pub struct AgentJobsHandler;
+
+const STATUS_POLL_INTERVAL_MS: u64 = 250;
+
+#[derive(Debug, Deserialize)]
+struct SpawnAgentsOnCsvArgs {
+    csv_path: String,
+    instruction: String,
+    id_column: Option<String>,
+    job_name: Option<String>,
```
jif-oai: This does not seem to be used anywhere... so I'm not sure this is interesting

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1012 @@
...
+    let base_instructions = session.get_base_instructions().await;
+    let spawn_config = build_agent_spawn_config(&base_instructions, turn.as_ref(), child_depth)?;
+    Ok(JobRunnerOptions {
+        max_concurrency,
+        spawn_config,
+    })
+}
+
+fn normalize_concurrency(requested: Option<usize>, max_threads: Option<usize>) -> usize {
+    let requested = requested.unwrap_or(64).max(1);
```
jif-oai: * Can we make the 64 a `const` somewhere?
* In a follow-up we should make this configurable
* 64 is gigantic. Any normal user would instantaneously get rate limited

codex-rs/core/src/tools/spec.rs
```diff
@@ -484,6 +485,112 @@ fn create_spawn_agent_tool() -> ToolSpec {
...
+        JsonSchema::Object {
+            properties: BTreeMap::new(),
+            required: None,
+            additional_properties: None,
+        },
+    );
+    ToolSpec::Function(ResponsesApiTool {
+        name: "report_agent_job_result".to_string(),
+        description:
+            "Worker-only tool to report a result for an agent job item. Main agents should not call this."
```
jif-oai, line 753: This description is pretty strange from an agent point of view. If the main agent should not call this, we should just not give access to the tool to the main agent. Should be quite easy to do

jif-oai, line 753: What if the worker does not call this tool? We could just use a structured output if we want something more deterministic

codex-rs/core/src/tools/spec.rs
```diff
@@ -1418,6 +1525,14 @@ pub(crate) fn build_specs(
         builder.register_handler("close_agent", collab_handler);
     }
 
+    if config.collab_tools {
```
jif-oai: I just enabled `collab` globally so feel free to create a new feature flag if you want

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1012 @@
...
+            .await;
+        self.last_emit_at = Instant::now();
+        self.last_processed = processed;
+        self.last_failed = progress.failed_items;
+        Ok(())
+    }
+}
+
+#[async_trait]
+impl ToolHandler for AgentJobsHandler {
```
jif-oai: Could we name this BatchJob or something like this?

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1012 @@
...
+        let now = Instant::now();
+        Self {
+            started_at: now,
+            last_emit_at: now,
+            last_processed: 0,
+            last_failed: 0,
+        }
+    }
+
+    async fn maybe_emit(
```
jif-oai, line 129: pretty cool

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1012 @@
...
+    last_processed: usize,
+    last_failed: usize,
+}
+
+impl JobProgressEmitter {
+    fn new() -> Self {
+        let now = Instant::now();
+        Self {
+            started_at: now,
+            last_emit_at: now,
```
jif-oai: nit: this is not correct. Nothing gets emitted on this time frame

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1012 @@
...
+                "unsupported agent job tool {other}"
+            ))),
+        }
+    }
+}
+
+mod spawn_agents_on_csv {
+    use super::*;
+
+    pub async fn handle(
```
jif-oai, line 223: this function would benefit some documentation 😅 , it's getting pretty complex

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1012 @@
...
+        })
+    }
+}
+
+fn required_state_db(
+    session: &Arc<Session>,
+) -> Result<Arc<codex_state::StateRuntime>, FunctionCallError> {
+    session.state_db().ok_or_else(|| {
+        FunctionCallError::RespondToModel(
+            "sqlite state db is unavailable for this session; enable the sqlite feature"
```
jif-oai, line 521: I think this should be a fatal, not a RespondToModel

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1012 @@
+use crate::agent::exceeds_thread_spawn_depth_limit;
```
jif-oai, line 1: this would benefit a bunch of integration tests

codex-rs/state/migrations/0010_agent_jobs_auto_export.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE agent_jobs
```
jif-oai: No need for a dedicated migration... this is not merged yet so just put everything in the same migration
Having tons of migrations just makes things harder to follow IMO

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1012 @@
...
+            let source_id = id_column_index
+                .and_then(|index| row.get(index).cloned())
+                .filter(|value| !value.trim().is_empty());
+            let row_index = idx + 1;
+            let mut item_id = source_id
+                .clone()
+                .unwrap_or_else(|| format!("row-{row_index}"));
+            if !seen_ids.insert(item_id.clone()) {
+                item_id = format!("{item_id}-{row_index}");
+                seen_ids.insert(item_id.clone());
```
jif-oai: this can still violate uniqueness if you have duplicate item_id (unlikely but possible)

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1012 @@
...
+            }
+
+            let row_object = headers
+                .iter()
+                .zip(row.iter())
+                .map(|(header, value)| (header.clone(), Value::String(value.clone())))
```
jif-oai, line 296: are we 100% sure header is unique? I'm not sure the validation will never leak. A few tests would be nice here

codex-rs/state/src/runtime.rs
```diff
@@ -717,6 +726,486 @@ ON CONFLICT(thread_id, position) DO NOTHING
...
+    result_json = ?,
+    reported_at = ?,
+    updated_at = ?,
+    assigned_thread_id = COALESCE(assigned_thread_id, ?),
+    last_error = NULL
+WHERE
+    job_id = ?
+    AND item_id = ?
+    AND status = ?
+    AND (assigned_thread_id IS NULL OR assigned_thread_id = ?)
```
jif-oai: Why this? That sounds racy... what if `report_agent_job_item_result` accepts reports when `assigned_thread_id` is `NULL`? Any caller can claim an item before the worker thread is set. That can misattribute results in the Running -> set_thread window

jif-oai (commented):
A few global points:
* We need more integration tests
* We don't have a mechanism to recover a sub-agent that crashes. In those cases, the job will stay as "Running" forever I think
* Implementation is way cleaner than before. Thanks for this
* Is it on purpose that this is only supported by `codex exec` for now? I think this might also be relevant for the `app-server` but it can come in a follow-up

daveaitel-openai:
Thanks for the thoughtful review, really appreciated. I went through each point and addressed them as follows:

  1. “We need more integration tests.”
     Agreed. I added integration coverage for the agent job flow (spawn, report, export).
     Commit: 904d05027
  2. “Job can get stuck running forever if a worker fails.”
     Fixed. We now mark spawn failures as failed immediately (so they don’t linger in Running), and we added a stale‑running reaper so jobs can’t hang indefinitely. The final job result includes a failure summary so it’s obvious which rows failed and why.
     Commits: 79e19fe41, b45721727
     Follow‑up improvement: the timeout is now configurable per job (max_runtime_seconds) and via config (agents.job_max_runtime_seconds).
     Commits: 8eef34dab, 8177ffaa6
  3. “Exec‑only for now? app‑server later?”
     That’s intentional for this PR. I left app‑server support as a follow‑up to keep scope contained.
  4. “Spawn args struct seems unused.”
     It is used. SpawnAgentsOnCsvArgs is the deserialized input for spawn_agents_on_csv in codex-rs/core/src/tools/handlers/agent_jobs.rs. The handler reads csv_path, instruction, id_column, job_name, output_schema, max_runtime_seconds, etc. These fields directly drive CSV parsing, job metadata, output schema
     storage, and runtime limits. No code change needed here.
  5. “Make 64 a const; make it configurable; 64 is too high.”
     Done. Extracted a constant, lowered default to 16, and made it configurable.
     Commits: 79e19fe41, 96a645e82
  6. “report_agent_job_result shouldn’t be visible to main agent.”
     Done. The tool is now worker‑only (gated by session source), so it’s not exposed to the main agent.
     Commit: 9442a579b
  7. “What if the worker doesn’t call this tool? Maybe structured output?”
     We decided not to add a fallback. Instead, we document that a missing report is treated as a failure.
     Commit: 7bac3aff0
  8. “Collab tools enabled globally.”
     No change needed. Existing gating handles the worker‑only tool exposure.
  9. “Rename AgentJobsHandler → BatchJob.”
     Done. Renamed to BatchJobHandler.
     Commit: 8c58fb3a3
  10. “Progress emitter init time is wrong.”
     Fixed emit timing so “last emit” and “start” aren’t tied together.
     Commit: 79e19fe41
  11. “spawn_agents_on_csv needs docs.”
     Added doc comment and improved tool description/semantics.
     Commit: 8c58fb3a3
  12. “Missing state DB should be fatal.”
     Done. It’s now a fatal error rather than RespondToModel.
     Commit: 984a3691
  13. “Migrations: fold 0010 into 0009.”
     Done. Auto‑export folded into the base migration.
     Commit: 174861ca4
  14. “Duplicate item_id / header uniqueness.”
     Added unique header validation + robust dedupe for item IDs.
     Commit: 79e19fe41
  15. “Thread assignment race in report.”
     Fixed the SQL race around assigned_thread_id.
     Commit: 4d2f37913
  16. “Tool descriptions unclear.”
     Updated descriptions and clarified reporting semantics.
     Commits: fae51af3a, 7bac3aff0

  Additional fixes since the review (from local failures while testing)

  - SQLite state DB default for subagents
    Added agents.sqlite_home + CODEX_SQLITE_HOME and default to a temp dir when in workspace‑write mode (otherwise CODEX_HOME). State DB now uses this, so subagents don’t fail with “sqlite state db unavailable” in full‑auto runs.
    Commit: ceac6a0ab
  - Fix agent_jobs insert error
    The agent_jobs INSERT now has the correct number of values (was 14 for 15 columns).
    Commit: ceac6a0ab
  - spawn_agents_on_csv export guard
    Ensures output CSV is exported even if the tool doesn’t return an output path.
    Commit: ceac6a0ab

codex-rs/core/src/config/mod.rs
```diff
@@ -97,8 +97,35 @@ pub use codex_git::GhostSnapshotConfig;
 /// the context window.
 pub(crate) const PROJECT_DOC_MAX_BYTES: usize = 32 * 1024; // 32 KiB
 pub(crate) const DEFAULT_AGENT_MAX_THREADS: Option<usize> = Some(6);
+pub(crate) const DEFAULT_AGENT_MAX_SPAWN_DEPTH: Option<usize> = Some(2);
+pub(crate) const DEFAULT_AGENT_JOB_MAX_RUNTIME_SECONDS: Option<u64> = None;
 
 pub const CONFIG_TOML_FILE: &str = "config.toml";
+const SQLITE_HOME_ENV: &str = "CODEX_SQLITE_HOME";
```
jif-oai: I _think_ this part should live in the `codex-state` crate but feel free to challenge. No very strong opinion tbh

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1207 @@
...
+const STATUS_POLL_INTERVAL: Duration = Duration::from_millis(250);
+const PROGRESS_EMIT_INTERVAL: Duration = Duration::from_secs(1);
+const DEFAULT_AGENT_JOB_ITEM_TIMEOUT: Duration = Duration::from_secs(60 * 30);
+
+#[derive(Debug, Deserialize)]
+struct SpawnAgentsOnCsvArgs {
+    csv_path: String,
+    instruction: String,
+    id_column: Option<String>,
+    job_name: Option<String>,
```
jif-oai: My point here was not the whole struct but just `job_name`. I'm not sure this is relevant to have it if this is not surfaces. Would be cool later to be able to resume from it though

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1207 @@
...
+fn default_output_csv_path(input_csv_path: &Path, job_id: &str) -> PathBuf {
+    let stem = input_csv_path
+        .file_stem()
+        .and_then(|stem| stem.to_str())
+        .unwrap_or("agent_job_output");
+    let job_suffix = &job_id[..8];
+    input_csv_path.with_file_name(format!("{stem}.agent-job-{job_suffix}.csv"))
+}
+
+fn parse_csv(content: &str) -> Result<(Vec<String>, Vec<Vec<String>>), String> {
```
jif-oai, line 1046: This is pretty cool but I would either:
* Check if there are any existing `crates` for this if we don't want to re-invent the wheel
* If not or nothing that suits our need, I would extract this in a small crate in `codex-utils-...` as I'm quite sure others will need it one day (+ you can put this in a small self-contained PR)

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1207 @@
...
+                    output_csv_path: output_csv_path.display().to_string(),
+                },
+                items.as_slice(),
+            )
+            .await
+            .map_err(|err| {
+                FunctionCallError::RespondToModel(format!("failed to create agent job: {err}"))
+            })?;
+
+        db.mark_agent_job_running(job_id.as_str())
```
jif-oai, line 348: You first mark it as running but then later you have
```
let options = build_runner_options(&session, &turn, requested_concurrency).await?;
```
so this means that if `build_runner_options` fails, the job stays running forever

codex-rs/state/src/model/agent_job.rs
```diff
@@ -0,0 +1,300 @@
...
+
+#[derive(Debug, Clone)]
+pub struct AgentJobItemCreateParams {
+    pub item_id: String,
+    pub row_index: i64,
+    pub source_id: Option<String>,
+    pub row_json: Value,
+}
+
+#[derive(Debug)]
```
jif-oai: You can just derive `FromRow` as well (from SQLX) so that you don't need the `try_from_row` impl

codex-rs/state/src/model/agent_job.rs
```diff
@@ -0,0 +1,300 @@
...
+            completed_at: value
+                .completed_at
+                .map(epoch_seconds_to_datetime)
+                .transpose()?,
+            last_error: value.last_error,
+        })
+    }
+}
+
+#[derive(Debug)]
```
jif-oai: Same comment for the `FromRow` (and same comment everywhere actually)

codex-rs/state/src/model/agent_job.rs
```diff
@@ -0,0 +1,300 @@
...
+pub(crate) struct AgentJobRow {
+    pub(crate) id: String,
+    pub(crate) name: String,
+    pub(crate) status: String,
+    pub(crate) instruction: String,
+    pub(crate) auto_export: i64,
+    pub(crate) max_runtime_seconds: Option<i64>,
+    pub(crate) output_schema_json: Option<String>,
```
jif-oai, line 150: This might be in a follow-up but this `output_schema_json` ticks me a bit. We should just transform it into a JSON-Schema (https://json-schema.org/) and used structured output. This will do a constraint sampling and ensure the schema is always respected. Ok for me not to do in PR but please add in a backlog somewhere (you can assign it to me if you don't want to do it)

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1207 @@
...
+                        options.spawn_config.clone(),
+                        prompt,
+                        Some(SessionSource::SubAgent(SubAgentSource::Other(format!(
+                            "agent_job:{job_id}"
+                        )))),
+                    )
+                    .await
+                {
+                    Ok(thread_id) => thread_id,
+                    Err(CodexErr::AgentLimitReached { .. }) => {
```
jif-oai, line 641: This means we gonna loop while waiting for agent spot to be available right? This can take a lot of time if the sub-agents are handled somewhere else. I don't have a way better solution though

codex-rs/core/src/tools/handlers/agent_jobs.rs
```diff
@@ -0,0 +1,1207 @@
...
+    db: Arc<codex_state::StateRuntime>,
+    job_id: &str,
+    active_items: &mut HashMap<ThreadId, ActiveJobItem>,
+    runtime_timeout: Duration,
+) -> anyhow::Result<()> {
+    let running_items = db
+        .list_agent_job_items(job_id, Some(codex_state::AgentJobItemStatus::Running), None)
+        .await?;
+    for item in running_items {
+        if is_item_stale(&item, runtime_timeout) {
```
jif-oai, line 793: Would be good to also try to kill the job just in case. To make sure we don't increase contention on max number of running agents

codex-rs/core/src/tools/spec.rs
```diff
@@ -1418,6 +1557,16 @@ pub(crate) fn build_specs(
         builder.register_handler("close_agent", collab_handler);
     }
 
+    if config.agent_jobs_tools {
+        let agent_jobs_handler = Arc::new(BatchJobHandler);
+        builder.push_spec(create_spawn_agents_on_csv_tool());
```
jif-oai, line 1817: OOC do you expect recursive jobs? Otherwise we could drop those for depth > 0 (opposite as `spawn_agents_on_csv`)

jif-oai (commented):
Thanks for the prompt work on the previous review. We are close and this is going in a way better direction IMO

jif-oai (approved):
Ok for me with the following comments:
1. Enjoy the merge with main. Ping me if you want a sanity check after
2. As follow-ups make sure to keep track of:
    a. Use of a crate for CSV handling or extract in a dedicated crate
    b. Discuss with the TUI and App team to see how can we render this feature
    c. Add the documentation somewhere here https://github.com/openai/developers-website. You can ask @dkundel-openai for help
    d. try to use structured output for the enforcement of the schema
    e. find a solution to limit the looping of `AgentLimitReached` 
     f. as this contain DB migration, make sure an alpha get cut and this alpha is used by VSCE and the app

PR #10941, opened 2026-02-06
fix(core): canonicalize wrapper approvals and support heredoc prefix …

viyatb-oai:
## Summary
- Reduced repeated approvals for equivalent wrapper commands and fixed execpolicy matching for heredoc-style shell invocations, with minimal behavior change and fail-closed defaults.

## Fixes
1. Canonicalized approval matching for wrappers so equivalent commands map to the same approval intent.
2. Added heredoc-aware prefix extraction for execpolicy so commands like `python3 <<'PY' ... PY` match rules such as `prefix_rule(["python3"], ...)`.
3. Kept fallback behavior conservative: if parsing is ambiguous, existing prompt behavior is preserved.

## Edge Cases Covered
- Wrapper path/name differences: `/bin/bash` vs `bash`, `/bin/zsh` vs `zsh`.
- Shell modes: `-c` and `-lc`.
- Heredoc forms: quoted delimiter (`<<'PY'`) and unquoted delimiter (`<< PY`).
- Multi-command heredoc scripts are rejected by the fallback
- Non-heredoc redirections (`>`, etc.) are not treated as heredoc prefix matches.
- Complex scripts still fall back to prior behavior rather than expanding permissions.


codex-rs/core/src/exec_policy.rs
```diff
@@ -121,8 +122,8 @@ impl ExecPolicyManager {
             prefix_rule,
         } = req;
         let exec_policy = self.current();
-        let commands =
-            parse_shell_lc_plain_commands(command).unwrap_or_else(|| vec![command.to_vec()]);
+        let (commands, used_heredoc_fallback) = commands_for_exec_policy(command);
+        let auto_amendment_allowed = !used_heredoc_fallback;
```
dylan-hurd-oai, line 129: cc @bolinfest for a bit of nuance here, but this feels reasonable to me

viyatb-oai: I'll add a comment here so that its explicit whats happening.

dylan-hurd-oai (approved):

PR #10943, opened 2026-02-06
app-server: print help message to console when starting websockets server

JaviSoto:
Follow-up to https://github.com/openai/codex/pull/10693

image

### DIFF
--- a/codex-rs/app-server/src/transport.rs
+++ b/codex-rs/app-server/src/transport.rs
@@ -11,7 +11,6 @@ use owo_colors::Style;
 use std::collections::HashMap;
 use std::io::ErrorKind;
 use std::io::Result as IoResult;
-use std::io::Write;
 use std::net::SocketAddr;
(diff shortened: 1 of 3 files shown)


codex-rs/app-server/src/transport.rs
```diff
@@ -33,6 +37,36 @@ use tracing::warn;
 /// plenty for an interactive CLI.
 pub(crate) const CHANNEL_CAPACITY: usize = 128;
 
+fn write_stderr_line(args: std::fmt::Arguments<'_>) {
```
maxj-oai: can we remove this and use the more idiomatic `eprintln!` instead?

maxj-oai (approved):

codex-rs/app-server/src/transport.rs
JaviSoto: Done!

PR #10947, opened 2026-02-06
fix(tui): conditionally restore status indicator using message phase

sayan-oai:
TLDR: use new message phase field emitted by preamble-supported models to determine whether an AgentMessage is mid-turn commentary. if so, restore the status indicator afterwards to indicate the turn has not completed.

### Problem
`commit_tick` hides the status indicator while streaming assistant text.
For preamble-capable models, that text can be commentary mid-turn, so hiding was correct during streaming but restore timing mattered:
- restoring too aggressively caused jitter/flashing
- not restoring caused indicator to stay hidden before subsequent work (tool calls, web search, etc.)

### Fix
- Add optional `phase` to `AgentMessageItem` and propagate it from `ResponseItem::Message`
- Keep indicator hidden during streamed commit ticks, restore only when:
  - assistant item completes as `phase=commentary`, and
  - stream queues are idle + task is still running.
- Treat `phase=None` as final-answer behavior (no restore) to keep existing behavior for non-preamble models

### Tests
Add/update tests for:
- no idle-tick restore without commentary completion
- commentary completion restoring status before tool begin
- snapshot coverage for preamble/status behavior


codex-rs/tui/src/chatwidget.rs
```diff
@@ -2083,6 +2117,15 @@ impl ChatWidget {
         self.set_status(message, additional_details);
     }
 
+    fn on_agent_message_item_completed(&mut self, item: AgentMessageItem) {
+        self.pending_status_indicator_restore = match item.phase {
+            // Models that don't support preambles only output AgentMessageItems on turn completion.
+            Some(MessagePhase::FinalAnswer) | None => false,
+            Some(MessagePhase::Commentary) => true,
+        };
```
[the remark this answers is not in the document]
sayan-oai: fixing

sayan-oai:
@codex review this

codex-rs/app-server-protocol/schema/json/v1/ForkConversationResponse.json
joshka-oai, line 1: At what point do we version bump the schema - is it left alone when the change is backwards compatible?

joshka-oai (approved):

codex-rs/app-server-protocol/schema/json/v1/ForkConversationResponse.json
sayan-oai: not sure actually, maybe @owenlin0 knows?

PR #10954, opened 2026-02-06
feat(app-server): thread/unsubscribe API

owenlin0:
Adds a new v2 app-server API for a client to be able to unsubscribe to a thread:
- New RPC method: `thread/unsubscribe`
- New server notification: `thread/closed`

Today clients can start/resume/archive threads, but there wasn’t a way to explicitly unload a live thread from memory without archiving it. With `thread/unsubscribe`, a client can indicate it is no longer actively working with a live Thread. If this is the only client subscribed to that given thread, the thread will be automatically closed by app-server, at which point the server will send `thread/closed` and `thread/status/changed` with `status: notLoaded` notifications.

This gives clients a way to prevent long-running app-server processes from accumulating too many thread (and related) objects in memory.

Closed threads will also be removed from `thread/loaded/list`.


owenlin0:
@codex review

owenlin0:
@codex review

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -4303,6 +4311,72 @@ impl CodexMessageProcessor {
...
+                        }
+                    };
+                    if tokio::time::timeout(Duration::from_secs(10), wait_for_shutdown)
+                        .await
+                        .is_err()
+                    {
+                        warn!("thread {thread_id} shutdown timed out; leaving thread loaded");
+                        ThreadCloseStatus::ShutdownTimedOut
+                    } else {
+                        let removed_thread = self.thread_manager.remove_thread(&thread_id).await;
```
acrognale-oai: can this leak background terminals? From looking at thread_manager, it looks like remove just removes it from the map

owenlin0: ah, we should probably clean up background terminals too. will address

owenlin0: `Op::Shutdown` does this actually in codex core, so what we're doing here is correct

acrognale-oai (approved):

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -4666,6 +4680,89 @@ impl CodexMessageProcessor {
...
+            }
+        };
+
+        let status = if let Ok(thread) = self.thread_manager.get_thread(thread_id).await {
+            info!("thread {thread_id} is loaded; shutting down");
+            match self.wait_for_thread_shutdown(&thread).await {
+                ThreadShutdownResult::Complete => {
+                    let removed_thread = self.thread_manager.remove_thread(&thread_id).await;
+                    self.clear_thread_runtime_tracking(thread_id).await;
+                    if removed_thread.is_some() {
```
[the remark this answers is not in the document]
owenlin0: seems ok for now, can address in a followup if truly necessary. also this is existing behavior with `thread/archive`

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -4666,6 +4680,147 @@ impl CodexMessageProcessor {
...
+                }
+            }
+        }
+
+        self.outgoing
+            .send_response(
+                request_id,
+                ThreadUnsubscribeResponse {
+                    status: ThreadUnsubscribeStatus::Unsubscribed,
+                },
```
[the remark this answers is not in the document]
owenlin0: that's fine

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -4702,6 +4716,143 @@ impl CodexMessageProcessor {
...
+
+        if !self.thread_state_manager.has_subscribers(thread_id).await {
+            // This connection was the last subscriber. Only now do we unload the thread.
+            info!("thread {thread_id} has no subscribers; shutting down");
+            // Any pending app-server -> client requests for this thread can no longer be
+            // answered; cancel their callbacks before shutdown/unload.
+            self.outgoing.cancel_requests_for_thread(thread_id).await;
+            self.thread_state_manager
+                .remove_thread_state(thread_id)
+                .await;
```
maxj-oai, line 4855: I think clearing this right away creates possibility of race conditions. Can we defer listener teardown until shutdown actually completes (listener lifetime should == thread lifetime)?
or otherwise mark the thread as “closing” and block reattach/resume during that window.

owenlin0: > otherwise mark the thread as “closing” and block reattach/resume during that window.

good point! I implemented this

maxj-oai (approved):

PR #10958, opened 2026-02-07
feat(core): add network constraints schema to requirements.toml

viyatb-oai:
## Summary

Add `requirements.toml` schema support for admin-defined network constraints in the requirements layer

example config:

```
[experimental_network]
enabled = true
allowed_domains = ["api.openai.com"]
denied_domains = ["example.com"]
```


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -535,6 +535,23 @@ pub struct ConfigRequirements {
     pub allowed_sandbox_modes: Option<Vec<SandboxMode>>,
     pub allowed_web_search_modes: Option<Vec<WebSearchMode>>,
     pub enforce_residency: Option<ResidencyRequirement>,
+    pub network: Option<NetworkRequirements>,
```
bolinfest, line 539: Perhaps this should get the `#experimental` annotation? See other examples in this file.

codex-rs/core/src/config_loader/config_requirements.rs
```diff
@@ -81,6 +81,7 @@ pub struct ConfigRequirements {
     pub mcp_servers: Option<Sourced<BTreeMap<String, McpServerRequirement>>>,
     pub(crate) exec_policy: Option<Sourced<RequirementsExecPolicy>>,
     pub enforce_residency: ConstrainedWithSource<Option<ResidencyRequirement>>,
+    pub network: Option<Sourced<NetworkRequirementsToml>>,
```
bolinfest: This should not contain a field like `NetworkRequirementsToml`. The on-disk TOML representation should be distinct from the type that defines _constraints_ for a value (or in this case, a set of values).

One thing that is somewhat unclear is what this is the network policy _for_. Today, I think it is meant to govern shell commands, but would it ever have to govern local MCP servers, as well? I suppose that could be configured elsewhere...

Currently, this feels like this tries to define "a specific network policy" rather than "a range of acceptable values for a network policy."

Maybe we're not ready to commit to that sort of API, in which case, perhaps we should try to make it clear that this is experimental/unstable?

viyatb-oai: exactly, I didn't want to commit to a specific API. I'll put a note that its experimental/unstable

bolinfest (approved):

PR #10965, opened 2026-02-07
refactor(network-proxy): flatten network config under [network]

viyatb-oai:
Summary:
- Rename config table from network_proxy to network.
- Flatten allowed_domains, denied_domains, allow_unix_sockets, and allow_local_binding onto NetworkProxySettings.
- Update runtime, state constraints, tests, and README to the new config shape.

### DIFF
--- a/codex-rs/network-proxy/src/state.rs
+++ b/codex-rs/network-proxy/src/state.rs
@@ -56,8 +56,8 @@ pub(crate) async fn build_config_state() -> Result<ConfigState> {
     let constraints = enforce_trusted_constraints(&config_layer_stack, &config)?;
 
     let layer_mtimes = collect_layer_mtimes(&config_layer_stack);
-    let deny_set = compile_globset(&config.network_proxy.policy.denied_domains)?;
-    let allow_set = compile_globset(&config.network_proxy.policy.allowed_domains)?;
+    let deny_set = compile_globset(&config.network.denied_domains)?;
+    let allow_set = compile_globset(&config.network.allowed_domains)?;
     Ok(ConfigState {
         config,
         allow_set,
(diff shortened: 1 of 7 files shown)


codex-rs/network-proxy/src/state.rs
```diff
@@ -314,79 +307,70 @@ pub(crate) fn validate_policy_against_constraints(
...
-                        "network_proxy.policy.allow_local_binding",
-                        "true",
-                        "false (disabled by managed config)",
-                    ))
-                } else {
-                    Ok(())
-                }
-            },
-        )?;
+        let _ = Constrained::new(config.network.allow_local_binding, move |candidate| {
```
bolinfest: I realize this is an existing issue, but why do we create a `Constrained::new()` and then immediately drop it?

codex-rs/network-proxy/src/state.rs
```diff
@@ -314,79 +307,70 @@ pub(crate) fn validate_policy_against_constraints(
...
-                }
-                if invalid.is_empty() {
-                    Ok(())
-                } else {
-                    Err(invalid_value(
-                        "network_proxy.policy.allowed_domains",
-                        format!("{invalid:?}"),
-                        "subset of managed allowed_domains",
-                    ))
+        let _ = Constrained::new(config.network.allowed_domains.clone(), move |candidate| {
```
bolinfest: Here too?

codex-rs/network-proxy/src/state.rs
```diff
@@ -314,79 +307,70 @@ pub(crate) fn validate_policy_against_constraints(
...
-                } else {
-                    Err(invalid_value(
-                        "network_proxy.policy.denied_domains",
-                        "missing managed denied_domains entries",
-                        format!("{missing:?}"),
-                    ))
-                }
-            },
-        )?;
+        let _ = Constrained::new(config.network.denied_domains.clone(), move |candidate| {
```
bolinfest: etc

bolinfest (approved):

codex-rs/network-proxy/src/state.rs
viyatb-oai: agreed its not clean, changed it to explicitly call a validate func

PR #10979, opened 2026-02-07
core: improve remote compaction failure observability

charley-oai:
## Summary
- capture and log the full remote compaction request payload before returning a compaction error
- log explicit diagnostics in the same failure path: status code, last API-reported total tokens/bytes estimate, and estimated tokens/bytes for items that are not yet reflected in the last API usage snapshot
- expose a narrow `TotalTokenUsageBreakdown` from session/context manager so compaction failure logging reports stable, specific fields
- estimate the post-usage delta with a non-stateful boundary: scan backward to the most recent model-generated history item, then count all items after it (user/developer/tool outputs/etc.)
- if no model-generated item exists yet, treat that delta as empty (prevents premature auto-compaction before first model output)

## Test Fixes Included
- fix `suite::compact` regressions caused by the new boundary behavior
- remove stale `Feature::RemoteCompaction` toggles from `suite/compact_remote.rs` (feature no longer exists), so the `all` integration target compiles

## Why
When compaction fails with `context_length_exceeded`, we need both:
1. the exact failing request payload, and
2. a delta estimate that better matches local history growth after model output.

The previous trailing-codex-generated approach could miss non-codex items appended after model output. The new boundary includes all appended items after the latest model-generated item while staying simple and readable.

## Validation
- `just fmt`
- `cargo test -p codex-core --lib history::tests::`
- `cargo test -p codex-core --test all suite::compact::auto_compact_ -- --nocapture`
- relying on CI for full suite

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -114,6 +114,7 @@ use crate::config::resolve_web_search_mode_for_turn;
 use crate::config::types::McpServerConfig;
 use crate::config::types::ShellEnvironmentPolicy;
 use crate::context_manager::ContextManager;
+use crate::context_manager::TotalTokenUsageBreakdown;
 use crate::environment_context::EnvironmentContext;
 use crate::error::CodexErr;
 use crate::error::Result as CodexResult;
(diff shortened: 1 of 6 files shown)


charley-oai:
@codex review this

PR #11028, opened 2026-02-07
Defer persistence of rollout file

etraut-openai:
- Defer rollout persistence for fresh threads (`InitialHistory::New`): keep rollout events in memory and only materialize rollout file + state DB row on first `EventMsg::UserMessage`.
- Keep precomputed rollout path available before materialization.
- Change `thread/start` to build thread response from live config snapshot and optional precomputed path.
- Improve pre-materialization behavior in app-server/TUI: clearer invalid-request errors for file-backed ops and a friendlier `/fork` “not ready yet” UX.
- Update tests to match deferred semantics across start/read/archive/unarchive/fork/resume/review flows.
- Improved resilience of user_shell test, which should be unrelated to this change but must be affected by timing changes

For Reviewers:
* The primary change is in recorder.rs
* Most of the other changes were to fix up broken assumptions in existing tests

Testing:
* Manually tested CLI
* Exercised app server paths by manually running IDE Extension with rebuilt CLI binary
* Only user-visible change is that `/fork` in TUI generates visible error if used prior to first turn


etraut-openai:
@codex review

codex-rs/core/src/rollout/recorder.rs
bolinfest: I think this needs to be updated slightly to reflect the change from `deferred_log_file_info`?

codex-rs/core/src/rollout/recorder.rs
```diff
@@ -543,22 +563,32 @@ fn create_log_file(config: &Config, conversation_id: ThreadId) -> std::io::Resul
...
 
     Ok(LogFileInfo {
-        file,
         path,
         conversation_id,
         timestamp,
     })
 }
 
+fn open_log_file(path: &Path) -> std::io::Result<File> {
```
bolinfest, line 573: async?

codex-rs/core/tests/suite/user_shell_cmd.rs
```diff
@@ -113,11 +114,22 @@ async fn user_shell_cmd_can_be_interrupted() {
...
+            Some(event.clone())
+        }
+        _ => None,
+    })
+    .await;
     codex.submit(Op::Interrupt).await.unwrap();
 
     // Expect a TurnAborted(Interrupted) notification.
-    let msg = wait_for_event(&codex, |ev| matches!(ev, EventMsg::TurnAborted(_))).await;
+    let msg = wait_for_event_with_timeout(
```
pakrym-oai, line 127: why this change?

codex-rs/core/tests/suite/memory_tool.rs
```diff
@@ -26,6 +31,18 @@ async fn get_memory_tool_returns_persisted_thread_memory() -> Result<()> {
     let thread_id = test.session_configured.session_id;
     let thread_id_string = thread_id.to_string();
 
+    let _materialize_mock = mount_sse_once(
```
pakrym-oai: nit: remove all unused variables

codex-rs/tui/src/app.rs
```diff
@@ -1445,46 +1445,55 @@ impl App {
...
-                            if let Some(summary) = summary {
-                                let mut lines: Vec<Line<'static>> =
-                                    vec![summary.usage_line.clone().into()];
-                                if let Some(command) = summary.resume_command {
-                                    let spans = vec![
-                                        "To continue this session, run ".into(),
-                                        command.cyan(),
-                                    ];
-                                    lines.push(spans.into());
+                    if path.exists() {
```
bolinfest, line 1450: Maybe comment that this is created lazily? I'm not sure how best to capture this, but I worry it may not be obvious at call sites. Maybe `ChatWidget.rollout_path()` at least deserves an updated docstring?

codex-rs/core/src/rollout/recorder.rs
```diff
@@ -288,55 +293,70 @@ impl RolloutRecorder {
...
+                    model_provider: Some(config.model_provider_id.clone()),
+                    base_instructions: Some(base_instructions),
+                    dynamic_tools: if dynamic_tools.is_empty() {
+                        None
+                    } else {
+                        Some(dynamic_tools)
+                    },
+                };
+
+                let file = if defer_until_first_user_message {
```
pakrym-oai: I wonder if we should have a `.create()` or `.start()` method on recorder and not have it understand user messages.

Then the agent loop can call commit when needed. This also avoids having the parameter that split based on the code path.

bolinfest (approved):
Code generally seems good by me, though the inherent complexity in supporting the laziness feels a bit unsatisfactory since anything dealing with rollouts needs to defend against the case where nothing may have hit disk yet? But if this is important to support, I suppose it can't be avoided...

codex-rs/core/src/rollout/recorder.rs
```diff
@@ -642,6 +708,72 @@ async fn rollout_writer(
     Ok(())
 }
 
+fn is_user_message_rollout_item(item: &RolloutItem) -> bool {
+    matches!(item, RolloutItem::EventMsg(EventMsg::UserMessage(_)))
```
pakrym-oai: Isn't agent.md message a user message? I thought we inject it pretty early.

pakrym-oai: `<environment_context> is also a user message.

codex-rs/app-server/tests/suite/archive_thread.rs
```diff
@@ -41,10 +41,16 @@ async fn archive_conversation_moves_rollout_into_archived_directory() -> Result<
...
+        !rollout_path.exists(),
+        "expected rollout path {} to be deferred until first user message",
         rollout_path.display()
     );
+    std::fs::create_dir_all(
+        rollout_path
+            .parent()
+            .expect("rollout path should have parent directory"),
+    )?;
+    std::fs::write(&rollout_path, "{}\n")?;
```
pakrym-oai: this looks a bit sus, why are we pre-creating a rollout file?

codex-rs/app-server/tests/suite/v2/thread_unarchive.rs
```diff
@@ -42,9 +48,13 @@ async fn thread_unarchive_moves_rollout_back_into_sessions_directory() -> Result
     .await??;
     let ThreadStartResponse { thread, .. } = to_response::<ThreadStartResponse>(start_resp)?;
 
-    let rollout_path = find_thread_path_by_id_str(codex_home.path(), &thread.id)
+    let rollout_path = thread.path.clone().expect("thread path");
+    write_minimal_rollout(&thread.id, &rollout_path)?;
```
pakrym-oai: is it easier to run a turn and avoid maintaining a second set of rollout writing logic?

especially when codex starts copying it around.

codex-rs/core/src/rollout/recorder.rs
```diff
@@ -567,35 +597,26 @@ async fn rollout_writer(
...
-            state_builder =
-                metadata::builder_from_session_meta(&session_meta_line, rollout_path.as_path());
-        }
-
-        // Write the SessionMeta as the first item in the file, wrapped in a rollout line
-        let rollout_item = RolloutItem::SessionMeta(session_meta_line);
-        writer.write_rollout_item(&rollout_item).await?;
-        state_db::reconcile_rollout(
+    // For non-deferred new sessions and resume sessions, write session metadata immediately.
+    if writer.is_some()
```
pakrym-oai, line 607: Can we somehow avoid forking the main logic path for deferred and not-deffered?

pakrym-oai: For example implement all session as deferred with an explicit signal? I'm worried that things will get out of sync.

codex-rs/core/src/codex.rs
```diff
@@ -872,7 +872,23 @@ impl Session {
...
+                    conversation_id,
+                    RolloutRecorderParams::new(
+                        conversation_id,
+                        forked_from_id,
+                        session_source,
+                        BaseInstructions {
+                            text: session_configuration.base_instructions.clone(),
+                        },
+                        session_configuration.dynamic_tools.clone(),
+                        true,
```
pakrym-oai: Is this the only diff? Can we use matches(initial_history, InitialHistory::New) ?

pakrym-oai (changes requested):
I think we should avoid having rollout recorder from interpreting items and knowing when to "commit". Instead it should be an explicit gesture from the agent loop.

codex-rs/core/tests/suite/user_shell_cmd.rs
etraut-openai: This was a flaky test that was pretty consistently failing for me in CI. I've seen it fail before occasionally, but the timing of my change must have pushed it into a place where it more consistently failed.

codex-rs/core/src/rollout/recorder.rs
etraut-openai: I don't think async makes sense in this case. Let me know if you're seeing something I'm not or feel strongly about it.

pakrym-oai (approved):

codex-rs/core/src/rollout/recorder.rs
bolinfest: I think this has been removed in the revised version, but I suggested this because I thought it was called from an async function (doing I/O).

PR #11039, opened 2026-02-07
feat: include state of [experimental_network] in /debug-config output

bolinfest:
#10958 introduced experimental support for a network config in `/etc/codex/requirements.toml`, so this extends `/debug-config` to surface this information, if set, which should make it easier to debug.


etraut-openai (approved):
Nice! LGTM

PR #11040, opened 2026-02-07
Simplify pre-connect

pakrym-oai:


codex-rs/core/src/codex.rs
```diff
@@ -1099,7 +1100,11 @@ impl Session {
         // when submit_turn() runs.
         sess.services.model_client.pre_establish_connection(
             sess.services.otel_manager.clone(),
-            session_configuration.cwd.clone(),
+            resolve_turn_metadata_header_with_timeout(
```
pakrym-oai, line 1103: keep metadata building outside of the client.

joshka-oai (approved):

PR #11044, opened 2026-02-07
feat: include [experimental_network] in <environment_context>

bolinfest:
If `NetworkConstraints` is set, then include the relevant settings on `<environment_context>`. Example:

```xml
<environment_context>
  <cwd>/repo</cwd>
  <shell>bash</shell>
  <network enabled="true">
    <allowed>api.example.com</allowed>
    <allowed>*.openai.com</allowed>
    <denied>blocked.example.com</denied>
  </network>
</environment_context>
```

### DIFF
--- a/codex-rs/core/src/environment_context.rs
+++ b/codex-rs/core/src/environment_context.rs
@@ -13,11 +13,22 @@ use std::path::PathBuf;
 pub(crate) struct EnvironmentContext {
     pub cwd: Option<PathBuf>,
     pub shell: Shell,
+    pub network: Option<NetworkContext>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub(crate) struct NetworkContext {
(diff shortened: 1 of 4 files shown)


codex-rs/core/src/environment_context.rs
```diff
@@ -13,38 +13,70 @@ use std::path::PathBuf;
 pub(crate) struct EnvironmentContext {
     pub cwd: Option<PathBuf>,
     pub shell: Shell,
+    pub network: Option<NetworkContext>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub(crate) struct NetworkContext {
```
viyatb-oai, line 20: should we also add `enable` flag to this? as a break glass if the proxy is not playing well

viyatb-oai (approved):
I wonder if we should add the proxy enable flag through the env context as an escape hatch (not a good idea if the model uses it every time though). everything else looks good.

PR #11051, opened 2026-02-08
Gate view_image tool by model input_modalities

wiltzius-openai:
- Plumb input modalities from model catalog through the openai model protocol. Default to text and image.
- Conditionally add the view_image tool only if input modalities support image.

### DIFF
--- a/codex-rs/protocol/src/openai_models.rs
+++ b/codex-rs/protocol/src/openai_models.rs
@@ -174,6 +174,13 @@ pub enum ApplyPatchToolType {
     Function,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash, TS, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum InputModality {
+    Text,
+    Image,
+}
+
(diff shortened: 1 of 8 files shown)


codex-rs/protocol/src/openai_models.rs
```diff
@@ -174,6 +134,13 @@ pub enum ApplyPatchToolType {
     Function,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash, TS, JsonSchema)]
+#[serde(rename_all = "snake_case")]
```
aibrahim-oai: don't we already have this definition somewhere else?

codex-rs/protocol/src/openai_models.rs
```diff
@@ -212,6 +179,10 @@ const fn default_effective_context_window_percent() -> i64 {
     95
 }
 
+fn default_input_modalities() -> Vec<InputModality> {
```
aibrahim-oai: don't we already have this function somewhere else?

aibrahim-oai (approved):
LGTM

PR #11109, opened 2026-02-08
[bazel] Upgrade some rulesets in preparation for enabling windows

zbarsky-openai:


.github/workflows/ci.bazelrc
```diff
@@ -16,5 +23,5 @@ common:macos --config=remote
 common:macos --strategy=remote
 common:macos --strategy=TestRunner=darwin-sandbox,local
 
-common:windows --strategy=TestRunner=local
-
+# On windows we can not cross-build the tests but run them locally due to what appears to be a Bazel bug
```
bolinfest, line 26:
```suggestion
# On windows we cannot cross-build the tests but run them locally due to what appears to be a Bazel bug
```

defs.bzl
```diff
@@ -85,6 +85,7 @@ def codex_rust_crate(
     proc_macro_dev_deps = all_crate_deps(proc_macro_dev = True)
 
     test_env = {
+        "INSTA_REQUIRE_FULL_MATCH": "0",
```
bolinfest, line 88: What does this change?

MODULE.bazel
```diff
@@ -35,58 +34,93 @@ register_toolchains(
...
 
-rust = use_extension("@rules_rust//rust:extensions.bzl", "rust")
-rust.toolchain(
+toolchains = use_extension("@rules_rs//rs/experimental/toolchains:module_extension.bzl", "toolchains")
+toolchains.toolchain(
     edition = "2024",
-    extra_target_triples = RUST_TRIPLES,
-    versions = ["1.93.0"],
+    # TODO(zbarsky): bump to 1.93 after fixing mingw
+    version = "1.92.0",
```
bolinfest, line 54: How long will this take to fix? I worry about disagreement between lints that are new in 1.93.0, for example, though I guess we still have cargo builds to catch those today?

bolinfest (approved):

MODULE.bazel
zbarsky-openai: 12 hours :)

defs.bzl
zbarsky-openai: stray change, reverting

PR #11113, opened 2026-02-08
feat(sandbox): enforce proxy-aware network routing in sandbox

viyatb-oai:
## Summary
- expand proxy env injection to cover common tool env vars (`HTTP_PROXY`/`HTTPS_PROXY`/`ALL_PROXY`/`NO_PROXY` families + tool-specific variants)
- harden macOS Seatbelt network policy generation to route through inferred loopback proxy endpoints and fail closed when proxy env is malformed
- thread proxy-aware Linux sandbox flags and add minimal bwrap netns isolation hook for restricted non-proxy runs
- add/refresh tests for proxy env wiring, Seatbelt policy generation, and Linux sandbox argument wiring

gated on `[experimental_network]` enabled in `requirements.toml`


nornagon-openai (approved):
sandbox changes seem safe in that they are strictly less permissive than previously. didn't review the rest of the PR in depth.

codex-rs/core/src/landlock.rs
```diff
@@ -33,20 +35,29 @@ where
...
-    spawn_child_async(
-        codex_linux_sandbox_exe.as_ref().to_path_buf(),
+    spawn_child_async(SpawnChildRequest {
+        program: codex_linux_sandbox_exe.as_ref().to_path_buf(),
         args,
         arg0,
-        command_cwd,
+        cwd: command_cwd,
         sandbox_policy,
+        network: None,
```
bolinfest: Should `network` be threaded through to this point?

codex-rs/core/src/landlock.rs
```diff
@@ -33,20 +35,29 @@ where
...
         stdio_policy,
         env,
-    )
+    })
     .await
 }
 
+pub(crate) fn allow_network_for_proxy(
+    sandbox_policy: &SandboxPolicy,
+    env: &HashMap<String, String>,
```
bolinfest: I think we should carry `network` around until the last possible moment so that we don't do a check like this.

codex-rs/core/src/landlock.rs
```diff
@@ -33,20 +35,29 @@ where
...
+) -> bool {
+    !sandbox_policy.has_full_network_access() && has_proxy_url_env_vars(env)
```
bolinfest: I don't think we should be using `has_full_network_access()` as part of the check because, for now, we want to make it so that `[experimental_network]` in `requirements.toml` trumps `--yolo`.

codex-rs/core/src/seatbelt.rs
```diff
@@ -28,25 +35,116 @@ pub async fn spawn_command_under_seatbelt(
...
+    } else {
+        String::new()
+    }
+}
+
 pub(crate) fn create_seatbelt_command_args(
     command: Vec<String>,
     sandbox_policy: &SandboxPolicy,
     sandbox_policy_cwd: &Path,
+    env: &HashMap<String, String>,
```
bolinfest: Again, I would prefer we pass `network` over `env`.

codex-rs/core/src/landlock.rs
```diff
@@ -33,20 +36,29 @@ where
...
         env,
-    )
+    })
     .await
 }
 
+pub(crate) fn allow_network_for_proxy(_network: Option<&NetworkProxy>) -> bool {
+    // Proxy environment variables are advisory only. In restricted policies we
+    // keep network sandboxing fail-closed unless we have explicit proxy-only
+    // enforcement in the sandbox runtime.
```
viyatb-oai: ugh, will change this

codex-rs/core/src/sandboxing/mod.rs
```diff
@@ -59,6 +60,7 @@ pub(crate) struct SandboxTransformRequest<'a> {
     pub spec: CommandSpec,
     pub policy: &'a SandboxPolicy,
     pub sandbox: SandboxType,
+    pub network: Option<&'a NetworkProxy>,
```
bolinfest, line 67: We can consider in a follow-up, but I have been thinking perhaps we want to `Option<Arc<NetworkProxy>>` instead?

codex-rs/core/src/tools/runtimes/shell.rs
bolinfest: We are already passing network here, so we shouldn't have this above?

```
        let mut env = req.env.clone();
        if let Some(network) = req.network.as_ref() {
        }
```

codex-rs/core/src/sandboxing/mod.rs
bolinfest, line 51: Should this also have `network: Option<NetworkProxy>`?

Also, what about `ExecCommandRequest` in `codex-rs/core/src/unified_exec/mod.rs`?

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -40,6 +40,12 @@ pub struct LandlockCommand {
     #[arg(long = "apply-seccomp-then-exec", hide = true, default_value_t = false)]
     pub apply_seccomp_then_exec: bool,
 
+    /// Internal compatibility flag.
+    ///
+    /// Proxy environment variables do not disable restricted-network sandboxing.
```
bolinfest: I'm struggling with the name/description of this flag. Can you write this as, "By default...If set..."?

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -26,11 +26,19 @@ pub(crate) struct BwrapOptions {
     /// This is the secure default, but some restrictive container environments
     /// deny `--proc /proc` even when PID namespaces are available.
     pub mount_proc: bool,
+    /// Whether to isolate networking with a dedicated network namespace.
+    ///
+    /// When true, bubblewrap adds `--unshare-net`, which removes access to the
+    /// host network stack and leaves only loopback inside the sandbox.
+    pub isolate_network: bool,
```
bolinfest: Is network a tri-state:

- full access
- no access
- restricted access via the proxy

If so, we can't capture this with a single boolean: do we need some tri-state option?

codex-rs/core/src/tools/runtimes/shell.rs
viyatb-oai: yep yep codex didn't catch this...

codex-rs/linux-sandbox/src/bwrap.rs
viyatb-oai: yeah add an enum

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -26,11 +26,37 @@ pub(crate) struct BwrapOptions {
...
-        Self { mount_proc: true }
+        Self {
+            mount_proc: true,
+            network_mode: BwrapNetworkMode::FullAccess,
+        }
+    }
+}
+
+/// Network policy modes for bubblewrap.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
```
bolinfest, line 43: should these be kebab-case?

bolinfest: what values does bubblewrap take?

viyatb-oai: we don't need kebab-case since this is internal. Bubblewrap doesn’t accept a “mode” value; it’s controlled by flags (`--unshare-net` vs omitted). `ProxyOnly` is an internal semantic state that currently maps to `--unshare-net`

codex-rs/core/src/sandboxing/mod.rs
viyatb-oai: added a todo

codex-rs/core/src/tools/runtimes/shell.rs
```diff
@@ -158,25 +158,22 @@ impl ToolRuntime<ShellRequest, ExecToolCallOutput> for ShellRuntime {
             command
         };
 
+        let env = req.env.clone();
+
         let spec = build_command_spec(
             &command,
             &req.cwd,
-            &req.env,
+            &env,
```
bolinfest: Not worth re-running CI for this, but could this stay as `&req.env,` or do we need to `clone()` it above?

codex-rs/core/src/tools/orchestrator.rs
```diff
@@ -151,9 +160,15 @@ impl ToolOrchestrator {
                     }
                 }
 
+                let escalated_sandbox = if has_managed_network_requirements {
```
bolinfest: If we got to this point, the user explicitly approved the command, so I think it should be run unsandboxed at this point (as was the case before).

It seems a bit overbearing if the agent cannot run a command the user explicitly approved since the user could presumably just as easily run it themselves at that point?

codex-rs/core/src/exec.rs
bolinfest:
```suggestion
    let exec_req = manager
```

codex-rs/core/src/tools/runtimes/shell.rs
viyatb-oai: `&req.env` should be ok, changing. thanks for catching that.

codex-rs/core/src/tools/orchestrator.rs
viyatb-oai: yeah, I believe managed network constraint has been applied before, so this should run unsandboxed.

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -26,32 +26,85 @@ pub(crate) struct BwrapOptions {
...
     cwd: &Path,
     options: BwrapOptions,
 ) -> Result<Vec<String>> {
     if sandbox_policy.has_full_disk_write_access() {
-        return Ok(command);
+        if options.network_mode == BwrapNetworkMode::FullAccess {
+            return Ok(command);
+        }
+
+        return Ok(create_bwrap_flags_full_filesystem(command, options));
```
bolinfest: Not worth churning CI for this, but FYI, this is arguably the more Rust-like way to do it:

```suggestion
        return if options.network_mode == BwrapNetworkMode::FullAccess {
            Ok(command)
        } else {
            Ok(create_bwrap_flags_full_filesystem(command, options))
        };
```

codex-rs/linux-sandbox/src/landlock.rs
```diff
@@ -183,3 +196,38 @@ fn install_network_seccomp_filter_on_current_thread() -> std::result::Result<(),
...
+    #[test]
+    fn restricted_network_policy_always_installs_seccomp() {
+        assert_eq!(
+            should_install_network_seccomp(&SandboxPolicy::ReadOnly, false),
+            true
+        );
+        assert_eq!(
+            should_install_network_seccomp(&SandboxPolicy::ReadOnly, true),
+            true
+        );
```
bolinfest:
```suggestion
        assert!(
            should_install_network_seccomp(&SandboxPolicy::ReadOnly, false),
        );
        assert!(
            should_install_network_seccomp(&SandboxPolicy::ReadOnly, true),
        );
```

bolinfest (approved):

PR #11118, opened 2026-02-08
feat(exec): add --append-system-prompt

### LINKED ISSUE #11117: Add --append-system-prompt for exec
## Problem & Context

In my CI pipeline I run two agents (Claude + Codex). The prompt has two parts: (1) the implementation requirements and (2) a strict output document format after implementation. Codex produces excellent code quality (often better than Claude), but it frequently fails the document requirements. The root cause appears to be prompt priority: Claude supports `--append-system-prompt`, so the output-format rules can live in the system prompt and are not diluted by the long implementation request. Codex lacks an equivalent mechanism, so format rules get mixed into the main prompt and are often ignored or drifted.

We discovered this after multiple failures and independent post‑mortems by both Claude and Codex converged on the same root cause. This issue is asking for a Codex‑side fix.

## Request

Add a CLI flag (e.g., `--append-system-prompt`) that **appends** text to the resolved base/system instructions (does not replace them), so high‑priority constraints can survive long prompts.

## Why it matters

This directly improves compliance with strict output formats in CI workflows without sacrificing Codex’s strong implementation quality.

xudesheng:
## Summary
- add `--append-system-prompt` to `codex exec` (and codex-exec) CLI
- store the appended text in config and concatenate onto resolved base/system instructions

## Testing
- not run (local)

## CLA
I have read the CLA Document and I hereby sign the CLA.

Fixes #11117


etraut-openai:

PR #11121, opened 2026-02-08
chore: reverse the codex-network-proxy -> codex-core dependency

bolinfest:


viyatb-oai (approved):

bolinfest:
@viyatb-oai This is the key line in `codex-rs/network-proxy-cli/src/main.rs` that pushed the move for the new crate:

```rust
use codex_core::network_proxy_loader;
```

PR #11132, opened 2026-02-09
core: account for all post-response items in auto-compact token checks

charley-oai:
## Summary
- change compaction pre-check accounting to include all items added after the last model-generated item, not only trailing codex-generated outputs
- use that boundary consistently in get_total_token_usage() and get_total_token_usage_breakdown()
- update history tests to cover user/tool-output items after the last model item

## Why
last_token_usage.total_tokens is API-reported for the last successful model response. After that point, local history may gain additional items (user messages, injected context, tool outputs). Compaction triggering must account for all of those items to avoid late compaction attempts that can overflow context.

## Testing
- just fmt
- cargo test -p codex-core


charley-oai:
@codex review this

jif-oai (approved):

PR #11154, opened 2026-02-09
Persist user messages to disk on context overflow

aibrahim-oai:


codex-rs/core/src/context_manager/context_discoverable.rs
```diff
@@ -33,10 +53,106 @@ impl ContextDiscoverable for FunctionCallOutput {
...
+        &self,
+        policy: TruncationPolicy,
+        _codex_home: Option<&Path>,
+        _thread_id: Option<&str>,
+        _context_window_tokens: Option<i64>,
+        _current_usage_tokens: i64,
+    ) -> ResponseItem {
         ResponseItem::CustomToolCallOutput(CustomToolCallOutput {
             call_id: self.call_id.clone(),
             output: truncate_text(&self.output, policy),
```
jif-oai, line 66: Why not having the same thing here?

PR #11162, opened 2026-02-09
Fix: update parallel tool call exec approval to approve on request id

shijie-oai:
### Summary

In parallel tool call, exec command approvals were not approved at request level but at a turn level. i.e. when a single request is approved, the system currently treats all requests in turn as approved. 

### Before
https://github.com/user-attachments/assets/d50ed129-b3d2-4b2f-97fa-8601eb11f6a8

### After
https://github.com/user-attachments/assets/36528a43-a4aa-4775-9e12-f13287ef19fc

### DIFF
--- a/codex-rs/app-server/src/bespoke_event_handling.rs
+++ b/codex-rs/app-server/src/bespoke_event_handling.rs
@@ -1678,7 +1678,7 @@ async fn on_file_change_request_approval_response(
     if let Some(status) = completion_status {
         complete_file_change_item(
             conversation_id,
-            item_id,
+            item_id.clone(),
             changes,
             status,
             event_turn_id.clone(),
@@ -1690,7 +1690,7 @@ async fn on_file_change_request_approval_response(
 
     if let Err(err) = codex
         .submit(Op::PatchApproval {
-            id: event_turn_id,
+            id: item_id,
             decision,
         })
         .await
@@ -1771,7 +1771,7 @@ async fn on_command_execution_request_approval_response(
 
     if let Err(err) = conversation
         .submit(Op::ExecApproval {
-            id: event_turn_id,
+            id: item_id,
             decision,
         })
         .await
--- a/codex-rs/app-server/src/bespoke_event_handling.rs
+++ b/codex-rs/app-server/src/bespoke_event_handling.rs
@@ -138,6 +138,7 @@ pub(crate) async fn apply_bespoke_event_handling(
             grant_root,
         }) => match api_version {
             ApiVersion::V1 => {
+                let approval_id = call_id.clone();
                 let params = ApplyPatchApprovalParams {
                     conversation_id,
                     call_id,
@@ -149,7 +150,7 @@ pub(crate) async fn apply_bespoke_event_handling(
                     .send_request(ServerRequestPayload::ApplyPatchApproval(params))
                     .await;
                 tokio::spawn(async move {
-                    on_patch_approval_response(event_turn_id, rx, conversation).await;
+                    on_patch_approval_response(approval_id, rx, conversation).await;
                 });
             }
             ApiVersion::V2 => {
@@ -214,6 +215,7 @@ pub(crate) async fn apply_bespoke_event_handling(
             parsed_cmd,
         }) => match api_version {
             ApiVersion::V1 => {
+                let approval_id = call_id.clone();
                 let params = ExecCommandApprovalParams {
                     conversation_id,
                     call_id,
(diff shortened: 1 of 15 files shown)


codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -1771,7 +1771,7 @@ async fn on_command_execution_request_approval_response(
 
     if let Err(err) = conversation
         .submit(Op::ExecApproval {
-            id: event_turn_id,
+            id: item_id,
```
shijie-oai, line 1776: For v2 - we are submitting the item_id so that individual request can be approved. I did not change the v1 behavior as it is no longer being used.

shijie-oai: Edit: I brought v1 API up to use item_id too so that test can pass as expected too. Should not have any affect on usage though.

codex-rs/core/src/codex_delegate.rs
```diff
@@ -330,21 +331,21 @@ async fn handle_exec_approval(
...
-    )
-    .await;
+    let decision =
+        await_approval_with_cancel(approval_fut, parent_session, &approval_id, cancel_token).await;
 
-    let _ = codex.submit(Op::ExecApproval { id, decision }).await;
+    let _ = codex
+        .submit(Op::ExecApproval {
+            id: approval_id,
+            decision,
```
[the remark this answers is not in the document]
shijie-oai: Addressed.

codex-rs/core/src/codex.rs
```diff
@@ -3157,7 +3159,13 @@ mod handlers {
...
     /// Propagate a user's exec approval decision to the session.
     /// Also optionally applies an execpolicy amendment.
-    pub async fn exec_approval(sess: &Arc<Session>, id: String, decision: ReviewDecision) {
+    pub async fn exec_approval(
+        sess: &Arc<Session>,
+        approval_id: String,
+        turn_id: Option<String>,
+        decision: ReviewDecision,
+    ) {
+        let event_turn_id = turn_id.unwrap_or_else(|| approval_id.clone());
```
dylan-hurd-oai, line 3168: @codex explain this particular fallback to me and why it works

codex-rs/core/src/codex.rs
```diff
@@ -1911,22 +1911,21 @@ impl Session {
         reason: Option<String>,
         proposed_execpolicy_amendment: Option<ExecPolicyAmendment>,
     ) -> ReviewDecision {
-        let sub_id = turn_context.sub_id.clone();
         // Add the tx_approve callback to the map before sending the request.
         let (tx_approve, rx_approve) = oneshot::channel();
-        let event_id = sub_id.clone();
+        let approval_id = call_id.clone();
```
shijie-oai, line 1916: Be more clear naming wise on what we are passing here.

codex-rs/core/src/codex_delegate.rs
```diff
@@ -330,21 +331,22 @@ async fn handle_exec_approval(
...
+        })
+        .await;
 }
 
 /// Handle an ApplyPatchApprovalRequest by consulting the parent session and replying.
 async fn handle_patch_approval(
     codex: &Codex,
-    id: String,
+    _id: String,
```
dylan-hurd-oai, line 349: 1. which type of id is this?
2. why don't we need it for `apply_patch`?

dylan-hurd-oai (approved):

codex-rs/core/src/codex_delegate.rs
shijie-oai: this is an event id - based on some local testing and usage check it seems like we can also rely purely on the call_id/item_id (which I call approval_id) to approval than using the event_id. Let me know if I am understanding this wrong though.

shijie-oai: I made the modification in core/src/codex.rs related to this at `request_patch_approval`. Thinking about it now maybe we should not unify it with exec command. The system is designed to only have one apply patch at a time it seems like.

shijie-oai: now think about it - I am going to revet the changes for apply patch approval. keep the behavior there as is and only modify exec approval.

dylan-hurd-oai: I'm fine with not unifying / separating the PRs, but it seems like we should support multiple apply_patch calls, right?

shijie-oai: Yes but is that a wise thing to support? feel like if you are writing to the same file 4 parallel calls to apply patch can be trouble.

jif-oai:
@codex review

PR #11175, opened 2026-02-09
fix: do not show closed agents in `/agent`

jif-oai:


gt-oai (approved):
It feels strange to me that clearing out the stale threads is a side-effect of opening the picker but I guess some separate clean-up process or state machine is a more involved change.

jif-oai:
> It feels strange to me that clearing out the stale threads is a side-effect of opening the picker but I guess some separate clean-up process or state machine is a more involved change.

This just clean the UI list. The stale threads are properly clean in the `ThreadManager`. But the UI is mostly pull based for those things and I don't think it would be a good idea to add dedicated messages just for this

PR #11227, opened 2026-02-09
feat(app-server): experimental flag to persist extended history

owenlin0:
This PR adds an experimental `persist_extended_history` bool flag to app-server thread APIs so rollout logs can retain a richer set of EventMsgs for non-lossy Thread > Turn > ThreadItems reconstruction (i.e. on `thread/resume`).

### Motivation
Today, our rollout recorder only persists a small subset (e.g. user message, reasoning, assistant message) of `EventMsg` types, dropping a good number (like command exec, file change, etc.) that are important for reconstructing full item history for `thread/resume`, `thread/read`, and `thread/fork`. 

Some clients want to be able to resume a thread without lossiness. This lossiness is primarily a UI thing, since what the model sees are `ResponseItem` and not `EventMsg`.

### Approach
This change introduces an opt-in `persist_full_history` flag to preserve those events when you start/resume/fork a thread (defaults to `false`).

This is done by adding an `EventPersistenceMode` to the rollout recorder:
- `Limited` (existing behavior, default)
- `Extended` (new opt-in behavior)

In `Extended` mode, persist additional `EventMsg` variants needed for non-lossy app-server `ThreadItem` reconstruction. We now store the following ThreadItems that we didn't before:
- web search
- command execution
- patch/file changes
- MCP tool calls
- image view calls
- collab tool outcomes
- context compaction
- review mode enter/exit

For **command executions** in particular, we truncate the output using the existing `truncate_text` from core to store an upper bound of 10,000 bytes, which is also the default value for truncating tool outputs shown to the model. This keeps the size of the rollout file and command execution items returned over the wire reasonable.

And we also persist `EventMsg::Error` which we can now map back to the Turn's status and populates the Turn's error metadata.

#### Updates to EventMsgs
To truly make `thread/resume` non-lossy, we also needed to persist the `status` on `EventMsg::CommandExecutionEndEvent` and `EventMsg::PatchApplyEndEvent`. Previously it was not obvious whether a command failed or was declined (similar for apply_patch). These EventMsgs were never persisted before so I made it a required field.


owenlin0:
@codex review

owenlin0:
@codex review

owenlin0:
@codex review

owenlin0:
@codex review

owenlin0:
@codex review

codex-rs/core/src/tools/events.rs
```diff
@@ -311,7 +339,7 @@ impl ToolEmitter {
                 } else {
                     msg
                 };
-                let event = ToolEventStage::Failure(ToolEventFailure::Message(normalized.clone()));
+                let event = ToolEventStage::Failure(ToolEventFailure::Rejected(normalized.clone()));
```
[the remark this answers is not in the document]
owenlin0: added a comment but will not address in this PR

owenlin0:
@codex review

codex-rs/app-server-protocol/src/protocol/thread_history.rs
```diff
@@ -153,23 +195,353 @@ impl ThreadHistoryBuilder {
...
-        turn.status = TurnStatus::Interrupted;
+        turn.status = TurnStatus::Failed;
+        turn.error = Some(V2TurnError {
+            message: payload.message.clone(),
+            codex_error_info: payload.codex_error_info.clone().map(Into::into),
+            additional_details: None,
+        });
+    }
+
+    fn handle_turn_aborted(&mut self, payload: &TurnAbortedEvent) {
```
celia-oai, line 522: is it possible to simplify this logic a bit? if not can we add a docstring?

codex-rs/app-server-protocol/src/protocol/thread_history.rs
```diff
@@ -180,9 +552,38 @@ impl ThreadHistoryBuilder {
...
+            .iter_mut()
+            .find(|turn| turn.id == payload.turn_id)
+        {
+            if matches!(turn.status, TurnStatus::Completed | TurnStatus::InProgress) {
+                turn.status = TurnStatus::Completed;
+            }
+            return;
+        }
+
         if let Some(current_turn) = self.current_turn.as_mut() {
```
celia-oai, line 578: can we simplify this with the line 556?

codex-rs/app-server/README.md
```diff
@@ -209,6 +209,8 @@ To branch from a stored session, call `thread/fork` with the `thread.id`. This c
 { "method": "thread/started", "params": { "thread": { … } } }
 ```
 
+Experimental: `thread/start`, `thread/resume`, and `thread/fork` accept `persistFullHistory: true` to persist a richer subset of `EventMsg` data for future `thread/read`, `thread/resume`, and `thread/fork` turn reconstruction. This does not backfill events that were not persisted previously.
```
celia-oai: should we do this as an `initialize` variable? then you don't need to pass in `persistFullHistory: true` to every new thread

codex-rs/core/src/rollout/recorder.rs
```diff
@@ -104,18 +111,48 @@ impl RolloutRecorderParams {
...
+    match item {
+        RolloutItem::EventMsg(EventMsg::ExecCommandEnd(mut event)) => {
+            event.aggregated_output = truncate_text(
+                &event.aggregated_output,
+                TruncationPolicy::Bytes(PERSISTED_EXEC_AGGREGATED_OUTPUT_MAX_BYTES),
+            );
+            event.stdout.clear();
+            event.stderr.clear();
+            event.formatted_output.clear();
+            RolloutItem::EventMsg(EventMsg::ExecCommandEnd(event))
```
celia-oai, line 155: why is this extra case needed

codex-rs/core/src/rollout/recorder.rs
```diff
@@ -673,9 +726,7 @@ async fn rollout_writer(
             RolloutCmd::AddItems(items) => {
                 let mut persisted_items = Vec::new();
                 for item in items {
-                    if is_persisted_response_item(&item) {
```
celia-oai, line 676: why is this one changed?

celia-oai (commented):
what happens when a user try to resume a session without extended history? Shall we throw some warning? similar with user try to resume a session with extended history but flag off. mostly curious about the mismatch between persistence & resume

codex-rs/core/src/tools/events.rs
```diff
@@ -301,6 +329,13 @@ impl ToolEmitter {
             Err(ToolError::Rejected(msg)) => {
                 // Normalize common rejection messages for exec tools so tests and
                 // users see a clear, consistent phrase.
+                //
+                // NOTE: ToolError::Rejected is currently used for both user-declined approvals
+                // and some operational/runtime rejection paths (for example setup failures).
+                // We intentionally map all of them through the "rejected" event path for now,
+                // which means a subset of non-user failures may be reported as Declined.
+                //
+                // TODO: We should add a new ToolError variant for user-declined approvals.
```
celia-oai, line 338: let's add this to readme too if this status is exposed to user?

codex-rs/protocol/src/protocol.rs
```diff
@@ -1005,6 +1005,27 @@ pub enum CodexErrorInfo {
     Other,
 }
 
+impl CodexErrorInfo {
```
celia-oai, line 1163: why is this logic needed? seems that bespoke_event_handling already special-case ThreadRollbackFailed?

codex-rs/app-server/README.md
owenlin0: it's a bit awkward as an `initialize` param, since `initialize` determines whether you want to opt into the experimental API in the first place. I think it's not too bad to pass it as needed for these RPCs

codex-rs/core/src/rollout/recorder.rs
owenlin0: because we want to store a truncated form of command executions instead of the full aggregated output

codex-rs/core/src/rollout/recorder.rs
owenlin0: it happens earlier in `record_items()` at line 474, so by the time we get here it's already filtered

codex-rs/core/src/tools/events.rs
owenlin0: ehh I feel like this is a quirk and we should fix internally, instead of documenting it in readme. it's a bit of an edge case too

codex-rs/protocol/src/protocol.rs
owenlin0: I wanted to make it explicit / centralized here on which errors affect the turn's status. this does get used by bespoke_event_handling now

owenlin0:
> what happens when a user try to resume a session without extended history? Shall we throw some warning? similar with user try to resume a session with extended history but flag off. mostly curious about the mismatch between persistence & resume

@celia-oai I don't think we should throw a warning, we'll just return what we have stored on disk. the client needs to be able to handle the full set of items anyway (when streaming), so client-rendering should be compatible with either `limited` or `extended` persistence modes.

codex-rs/app-server/README.md
celia-oai: sg

celia-oai:
> > what happens when a user try to resume a session without extended history? Shall we throw some warning? similar with user try to resume a session with extended history but flag off. mostly curious about the mismatch between persistence & resume
> 
> @celia-oai I don't think we should throw a warning, we'll just return what we have stored on disk. the client needs to be able to handle the full set of items anyway (when streaming), so client-rendering should be compatible with either `limited` or `extended` persistence modes.

sg!

celia-oai (approved):

PR #11228, opened 2026-02-09
Use longest remote model prefix matching

aibrahim-oai:
Match model metadata by longest matching remote slug prefix before local fallback.

- Update `get_model_info` to prefer the most specific remote slug prefix for the requested model.
- Add an integration test to assert `gpt-5.3-codex-test` resolves to `gpt-5.3-codex` over `gpt-5.3`.

### DIFF
--- a/codex-rs/core/src/models_manager/manager.rs
+++ b/codex-rs/core/src/models_manager/manager.rs
@@ -152,22 +152,21 @@ impl ModelsManager {
         model: &str,
         config: &Config,
     ) -> Option<ModelInfo> {
-        self.get_remote_models(config)
-            .await
-            .into_iter()
-            .fold(None, |best, candidate| {
-                if model.starts_with(&candidate.slug) {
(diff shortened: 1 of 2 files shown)


codex-rs/core/src/models_manager/manager.rs
```diff
@@ -149,6 +147,29 @@ impl ModelsManager {
...
 
+    async fn find_remote_model_by_longest_prefix(
+        &self,
+        model: &str,
+        config: &Config,
+    ) -> Option<ModelInfo> {
+        self.get_remote_models(config)
+            .await
+            .into_iter()
+            .fold(None, |best, candidate| {
```
pakrym-oai: please, write  me a for loop. I'm not smart enough to understand fold

pakrym-oai (approved):

codex-rs/core/src/models_manager/manager.rs
aibrahim-oai: lol

Issue #11213, opened 2026-02-09
Skip api.openai.com/v1/models refresh when using custom responses-api-endpoint


Ferrus91:
## Problem

When using `codex-action@v1.4` with a custom `responses-api-endpoint` (e.g., an in-house OpenAI-compatible proxy), the Codex CLI still calls `https://api.openai.com/v1/models` on startup to refresh available models.

If the API key is issued for the custom endpoint and not for OpenAI directly, this call always fails with a 401 Unauthorized:

```
ERROR codex_core::models_manager::manager: failed to refresh available models:
unexpected status 401 Unauthorized: Missing bearer authentication in header,
url: https://api.openai.com/v1/models?client_version=0.98.0
```

The error is non-fatal (the CLI continues), but it's noisy and confusing in CI logs.

## Expected behavior

When a custom `responses-api-endpoint` is configured, the CLI should either:
- Skip the `api.openai.com/v1/models` refresh entirely, or
- Call the custom endpoint's `/v1/models` instead, or
- Provide a flag to disable the model refresh (e.g., `--skip-model-refresh`)

## Reproduction

1. Configure `codex-action@v1` with a custom `responses-api-endpoint` and an API key that is valid only for that endpoint
2. Run the action
3. Observe the 401 error in logs from the direct `api.openai.com/v1/models` call

## Environment

- Codex CLI v0.98.0
- `openai/codex-action@v1`
- Using `responses-api-endpoint` pointed at an in-house proxy

etraut-openai:
Thanks for the bug report.

Could you please provide the relevant config settings that you're using to configure your endpoint (but redact any sensitive info)? I presume you're using a custom model provider definition? Are you configuring the client with an OpenAI API Key, or is that handled by your proxy?

etraut-openai:
This will be fixed in the next release.

PR #11239, opened 2026-02-09
Disable dynamic model refresh for custom model providers

etraut-openai:
The dynamic model refresh feature (`https://api.openai.com/v1/models` endpoint) is currently gated on a runtime check for an auth method other than API Key. It should be gated on a check specifically for ChatGPT Auth because some custom model providers (e.g. for local models) use no auth mechanism. A call to `self.auth_manager.auth_mode()` will return `None` in this case.

Addresses #11213


codex-rs/core/src/models_manager/manager.rs
```diff
@@ -194,9 +194,16 @@ impl ModelsManager {
...
-        {
+        if !config.features.enabled(Feature::RemoteModels) {
+            return Ok(());
+        }
+        if self.auth_manager.auth_mode() != Some(AuthMode::Chatgpt) {
+            if matches!(
+                refresh_strategy,
+                RefreshStrategy::Offline | RefreshStrategy::OnlineIfUncached
+            ) {
+                self.try_load_cache().await;
```
aibrahim-oai, line 205: they wouldn't have cache probably.

aibrahim-oai (approved):

PR #11240, opened 2026-02-09
fix(app-server): for external auth, replace id_token with chatgpt_acc…

owenlin0:
…ount_id and chatgpt_plan_type

### Summary
Following up on external auth mode which was introduced here: https://github.com/openai/codex/pull/10012

Turns out some clients have a differently shaped ID token and don't have a chosen workspace (aka chatgpt_account_id) encoded in their ID token. So, let's replace `id_token` param with `chatgpt_account_id` and `chatgpt_plan_type` (optional) when initializing the external ChatGPT auth mode (`account/login/start` with `chatgptAuthTokens`). 

The client was able to test end-to-end with a Codex build from this branch and verified it worked!


teddywyly-oai (approved):
Thx for doing this. 🎨

owenlin0:
@codex review

PR #11244, opened 2026-02-09
feat: add SkillPolicy to skill metadata and support allow_implicit_invocation

alexsong-oai:
Tested by setting the policy in agents/openai.yaml to true, false, and leaving it unset (default).
```
policy:
  allow_implicit_invocation: false
```
Screenshot 2026-02-09 at 3 42 41 PM

### DIFF
--- a/codex-rs/core/src/skills/loader.rs
+++ b/codex-rs/core/src/skills/loader.rs
@@ -70,7 +70,7 @@ struct Dependencies {
 
 #[derive(Debug, Deserialize)]
 struct Policy {
-    #[serde(default)]
+    #[serde(default = "default_allow_implicit_invocation")]
     allow_implicit_invocation: bool,
 }
 
@@ -80,6 +80,10 @@ impl Default for Policy {
             allow_implicit_invocation: true,
         }
     }
+}
+
+fn default_allow_implicit_invocation() -> bool {
+    true
 }
 
 #[derive(Debug, Default, Deserialize)]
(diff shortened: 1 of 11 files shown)


codex-rs/core/src/skills/loader.rs
```diff
@@ -65,6 +68,20 @@ struct Dependencies {
     tools: Vec<DependencyTool>,
 }
 
+#[derive(Debug, Deserialize)]
+struct SkillPolicyFile {
+    #[serde(default)]
+    allow_implicit_invocation: bool,
```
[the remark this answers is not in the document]
alexsong-oai: updated with a custom default

docs/skills.md
```diff
@@ -1,3 +1,7 @@
 # Skills
 
 For information about skills, refer to [this documentation](https://developers.openai.com/codex/skills).
+
+Local skills can opt out of implicit prompt injection by setting
```
xl-openai: Remove this. Instead after the public dev doc later.

codex-rs/core/src/skills/loader.rs
```diff
@@ -47,6 +48,8 @@ struct SkillMetadataFile {
     interface: Option<Interface>,
     #[serde(default)]
     dependencies: Option<Dependencies>,
+    #[serde(default)]
+    policy: Policy,
```
xl-openai: Should it be Optional?

codex-rs/core/src/skills/model.rs
```diff
@@ -10,10 +10,24 @@ pub struct SkillMetadata {
     pub short_description: Option<String>,
     pub interface: Option<SkillInterface>,
     pub dependencies: Option<SkillDependencies>,
+    pub policy: SkillPolicy,
```
xl-openai: You don't need to return this to tui right?

alexsong-oai: Right, it's not added to protocol.rs::SkillMetadata, so won't be exposed on TUI

codex-rs/core/src/skills/model.rs
```diff
@@ -57,10 +71,14 @@ impl SkillLoadOutcome {
         !self.disabled_paths.contains(&skill.path)
     }
 
+    pub fn is_skill_enabled_for_implicit_use(&self, skill: &SkillMetadata) -> bool {
+        self.is_skill_enabled(skill) && skill.policy.allow_implicit_invocation
+    }
+
     pub fn enabled_skills(&self) -> Vec<SkillMetadata> {
```
xl-openai: Need to change the name as you now filter more than just enable status.

codex-rs/core/src/skills/model.rs
xl-openai: I see you can also make policy Optional here.

alexsong-oai: updated

codex-rs/core/src/skills/loader.rs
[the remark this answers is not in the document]
xl-openai: You can also just make this Option

xl-openai (approved):

PR #11245, opened 2026-02-09
deflake linux-sandbox NoNewPrivs timeout

joshka-oai:
Deflake `codex-linux-sandbox::all suite::landlock::test_no_new_privs_is_enabled`.

CI has intermittently failed with `Sandbox(Timeout)` (exit 124) because the sandboxed
`grep '^NoNewPrivs:' /proc/self/status` can run close to the short timeout budget.

This updates only this test to use `LONG_TIMEOUT_MS`, which removes the near-threshold timeout
behavior while keeping the rest of the suite unchanged.

Refs (previous failures):
- PR: https://github.com/openai/codex/actions/runs/21836764823/job/63009902779
- PR: https://github.com/openai/codex/actions/runs/21837427251/job/63012470353
- main: https://github.com/openai/codex/actions/runs/21830746538/job/62988079964

Validation:
- Local: `cd codex-rs && cargo test -p codex-linux-sandbox` (non-Linux runs 0 tests)

### DIFF
--- a/codex-rs/linux-sandbox/tests/suite/landlock.rs
+++ b/codex-rs/linux-sandbox/tests/suite/landlock.rs
@@ -199,7 +199,9 @@ async fn test_no_new_privs_is_enabled() {
     let output = run_cmd_output(
         &["bash", "-lc", "grep '^NoNewPrivs:' /proc/self/status"],
         &[],
-        SHORT_TIMEOUT_MS,
+        // We have seen timeouts when running this test in CI on GitHub,
+        // so we are using a generous timeout until we can diagnose further.
+        LONG_TIMEOUT_MS,
     )
     .await;
     let line = output


codex-rs/linux-sandbox/tests/suite/landlock.rs
```diff
@@ -196,10 +196,14 @@ async fn test_writable_root() {
 
 #[tokio::test]
 async fn test_no_new_privs_is_enabled() {
+    // On non-aarch64 CI we've seen this sandboxed `grep` take just over the 200ms short timeout
+    // (~203ms), so give this test a little extra headroom to avoid jittery failures.
+    let timeout_ms = SHORT_TIMEOUT_MS.max(300);
```
bolinfest: I would advocate for keeping the comment but just using `LONG_TIMEOUT_MS` rather than introducing this new, unnamed value between `SHORT_TIMEOUT_MS` and `LONG_TIMEOUT_MS`? Waiting an extra 700ms isn't going to kill us?

bolinfest (approved):

codex-rs/linux-sandbox/tests/suite/landlock.rs
joshka-oai: sounds good. will fix

joshka-oai:
Addressed bolinfest suggestion in 891084a0: switched test_no_new_privs_is_enabled to LONG_TIMEOUT_MS instead of using an intermediate timeout value.

PR #11249, opened 2026-02-10
[apps] Improve app installation flow.

mzeng-openai:
- [x] Add buttons to start the installation flow and verify installation completes.
- [x] Hard refresh apps list when the /apps view opens.

### DIFF
--- a/codex-rs/Cargo.lock
+++ b/codex-rs/Cargo.lock
@@ -2232,6 +2232,7 @@ dependencies = [
  "url",
  "uuid",
  "vt100",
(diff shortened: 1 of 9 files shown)


canvrno-oai (approved):

canvrno-oai:
LGTM- Reviewed and tested.

The key events got got me thinking about the app list navigation, and I'm wondering if we'd want to implement the same navigation keys there and gate the search mode (currently default when list is opened) behind a `/` key event for consistency.

PR #11250, opened 2026-02-10
Remote models feature flag should only disable refreshing remote models

aibrahim-oai:
This will allow non-chatgpt users to use models.json but still not do a call to /models when disabling the feature flag


aibrahim-oai:
@codex review this

aibrahim-oai:
@codex review

aibrahim-oai:
@codex review this

PR #11251, opened 2026-02-10
Disable WebSocket timing warning in non-dev builds

pakrym-oai:
Summary
- gate the runtime metrics warning behind `cfg!(debug_assertions)` so release builds can skip the WebSocket timing log
- keep metric merging behavior unchanged while avoiding the noise in production

Testing
- Not run (not requested)


joshka-oai (approved):

apanasenko-oai (approved):

apanasenko-oai (changes requested):
we should use flag `runtime_metrics` to control this feature

PR #11269, opened 2026-02-10
feat: reserve loopback ephemeral listeners for managed proxy

bolinfest:
Codex may run many per-thread proxy instances, so hardcoded proxy ports are brittle and conflict-prone. The previous "ephemeral" approach still had a race: `build()` read `local_addr()` from temporary listeners and dropped them before `run()` rebound the ports. That left a [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) window where the OS (or another process) could reuse the same port, causing intermittent `EADDRINUSE` and partial proxy startup.

Change the managed proxy path to reserve real listener sockets up front and keep them alive until startup:

- add `ReservedListeners` on `NetworkProxy` to hold HTTP/SOCKS/admin std listeners allocated during `build()`
- in managed mode, bind `127.0.0.1:0` for each listener and carry those bound sockets into `run()` instead of rebinding by address later
- add `run_*_with_std_listener` entry points for HTTP, SOCKS5, and admin servers so `run()` can start services from already-reserved sockets
- keep static/configured ports only when `managed_by_codex(false)`, including explicit `socks_addr` override support
- remove fallback synthetic port allocation and add tests for managed ephemeral loopback binding and unmanaged configured-port behavior

This makes managed startup deterministic, avoids port collisions, and preserves the intended distinction between Codex-managed ephemeral ports and externally managed fixed ports.

### DIFF
--- a/codex-rs/network-proxy/src/admin.rs
+++ b/codex-rs/network-proxy/src/admin.rs
@@ -16,6 +16,7 @@ use serde::Deserialize;
 use serde::Serialize;
 use std::convert::Infallible;
 use std::net::SocketAddr;
+use std::net::TcpListener as StdTcpListener;
 use std::sync::Arc;
 use tracing::error;
 use tracing::info;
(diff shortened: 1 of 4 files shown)


viyatb-oai:
@codex review

viyatb-oai (approved):

Issue #11210, opened 2026-02-09
macOS Seatbelt sandbox blocks os.cpus() — missing mach-host permission breaks yarn/npm


kaladivo:
### What version of Codex CLI is running?

0.98.0

### What subscription do you have?

pro

### Which model were you using?

gpt-5.3-codex

### What platform is your computer?

Darwin 25.2.0 arm64 arm

### What terminal emulator and version are you using (if applicable)?

Ghostty, VScode

### What issue are you seeing?

`os.cpus()` returns an empty array when commands run under the macOS Seatbelt sandbox. This causes yarn, npm, node-gyp, and other tools that use `os.cpus().length` for worker parallelism to fail or produce incorrect behavior.

## Root Cause

Node.js `os.cpus()` calls libuv's `uv_cpu_info()`, which on macOS uses the Mach API `host_processor_info()`. The current seatbelt policy does not grant `(allow mach-host*)`, so this call silently fails and returns an empty array.

The policy *does* allow `sysctl-read` for `hw.ncpu`, `hw.activecpu`, etc., so `os.availableParallelism()` likely works — but many packages still rely on `os.cpus().length`.

### What steps can reproduce the bug?

```bash
# With default sandbox (workspace-write):
codex --sandbox workspace-write

# Ask codex to run:
node -e "const os=require('os');console.log('cpus',os.cpus().length);console.log('availableParallelism',os.availableParallelism?.())"
```

**Expected:** `cpus` prints the actual CPU count (e.g. 10)
**Actual:** `cpus 0` - yarn install and other tools break

### What is the expected behavior?

See above

## Suggested Fix

Add `(allow mach-host*)` to the seatbelt base policy in `codex-rs/core/src/seatbelt.rs`, similar to the existing `(allow process-info* (target same-sandbox))` rule. This is a read-only query with no security implications.

## Workaround

Monkey-patch `os.cpus` via `NODE_OPTIONS`:

```toml
# ~/.codex/config.toml
[shell_environment_policy.set]
NODE_OPTIONS = "--require /path/to/fix-cpus.js"
```

```javascript
// fix-cpus.js
const os = require('os');
const original = os.cpus;
os.cpus = function () {
  const result = original.call(this);
  if (!result || result.length === 0) {
    const count = os.availableParallelism?.() || 4;
    return Array.from({ length: count }, () => ({
      model: 'unknown', speed: 0,
      times: { user: 0, nice: 0, sys: 0, idle: 0, irq: 0 },
    }));
  }
[...]

PR #11277, opened 2026-02-10
Added seatbelt policy rule to allow os.cpus

etraut-openai:
I don't think this policy change increases the risk, other than potentially exposing the caller to bugs in these kernel calls, which are unlikely. 

Without this change, some tools are silently failing or making incorrect decisions about the processor type (e.g. installing x86 binaries rather than Apple silicon binaries).

This addresses #11210


viyatb-oai:
```
codex sandbox macos --log-denials /usr/bin/env node -e 'console.log(require("os").cpus().length)'                                                                                                       130
0

=== Sandbox denials ===
(env) sysctl-read security.mac.lockdown_mode_state
(env) sysctl-read kern.bootargs
(env) file-write-data /dev/dtracehelper
(node) sysctl-read security.mac.lockdown_mode_state
(node) sysctl-read kern.bootargs
(node) file-write-data /dev/dtracehelper
(node) sysctl-read machdep.cpu.brand_string
(node) sysctl-read hw.model
(node) file-write-data /dev/ttys038
```

I think we need at minimum

```
  ; Node os.cpus() / CPU identification
  (sysctl-name "machdep.cpu.brand_string")
  (sysctl-name "hw.model")

  ; Optional (robustness across Node/macOS variants)
  ; (sysctl-name "hw.logicalcpu")
  ; (sysctl-name "hw.cpufrequency")
 ```
``` ; Allow stdout/stderr writes to the user’s existing TTY (pre-seatbelt, no pty extension)
(allow file-write-data
  (require-all
    (regex #"^/dev/ttys[0-9]+")
    (vnode-type CHARACTER-DEVICE)))```

viyatb-oai (approved):

Issue #1980, opened 2025-08-08
Adhere to XDG Base Desktop Specification on Linux platforms


offsetcyan:
The user's home directory is not the place to dump program data, and for future cross-platform compatibility handling this would be inappropriate. Currently Codex stores its data in ~/.codex (modulated by a custom `CODEX_HOME` environment variable), however Linux has a popular specification for where to place files of various types ([link](https://specifications.freedesktop.org/basedir-spec/latest/)). In Codex's case, $XDG_CONFIG_HOME/codex/ would be suitable. ([more info](https://dotfiles-matter.click/))

bolinfest:
`XDG_CONFIG_HOME` on Mac (I know you're asking about Linux) is `"$HOME/Library/Application Support"`, which is annoying, as it contains spaces and makes all sorts of commands hard to copy/paste.

The default of `$HOME/.config` on Linux is more sane, but then every documentation example is more complex.

We decided to go with `~/.codex` instead, but we do honor `CODEX_HOME`, so you can set that to whatever you like.

offsetcyan:
A user needing to add a custom environment variable to their session for every application that (not unreasonably) prefers special behaviour to save development effort would lead to a very uncooperative ecosystem and tedium on part of users. Would a PR for the desired behaviour in this and #142 be considered?

bolinfest:
@offsetcyan as noted:

> The default of $HOME/.config on Linux is more sane, but then every documentation example is more complex.

Mainly for that reason, no, I would not consider a PR that would change the default to `$XDG_CONFIG_HOME/codex` on Linux.

It would also break existing users at this point.

99991:
This is a common mistake. Here you can track the progress of [VS Code](https://github.com/microsoft/vscode/issues/3884), [Claude Code](https://github.com/anthropics/claude-code/issues/1455) and [Aider](https://github.com/Aider-AI/aider/issues/216) as they fix the same issue.

> `XDG_CONFIG_HOME` on Mac (I know you're asking about Linux) is `"$HOME/Library/Application Support"`, which is annoying, as it contains spaces and makes all sorts of commands hard to copy/paste.

Thus, Linux users suffer again for Apple's sins.

offsetcyan:
> It would also break existing users at this point.

This kind of change can be implemented in a backwards-compatible way by checking for the existence of `CODEX_HOME` and `~/.codex` in that order, then following the XDG spec if neither were specified.

0xdevalias:
[...]

PR #11289, opened 2026-02-10
Support AGENTS_HOME for user skills roots

Bad3r:
## Problem
Codex currently loads user skills from two locations:
- deprecated: `$CODEX_HOME/skills`
- current: `~/.agents/skills`

The second path is hardcoded to the home directory, which makes XDG-managed environments awkward (for example, when users keep agent config under `$XDG_CONFIG_HOME/agents`).

## Purpose
Introduce an explicit agents-home resolver so user skills can be rooted outside `~` while preserving existing behavior.

## Why this change was created
Users who manage config declaratively (Nix/Home Manager, XDG-first setups) need a first-class way to redirect the user skill root. The codebase already has a strong, predictable pattern for this with `CODEX_HOME`, so this PR applies the same pattern to agents home.

## Design Goal: Mirror Existing `CODEX_HOME` Semantics
This implementation intentionally **mimics the existing `CODEX_HOME` logic** in `codex-utils-home-dir`:
- env var is honored when set
- empty env var is treated as unset
- set path must exist
- set path must be a directory
- set path is canonicalized
- if unset, a deterministic default path is used

For agents home, that means:
- env var: `AGENTS_HOME`
- default: `~/.agents`

No implicit `XDG_CONFIG_HOME` fallback is added by design; users can opt in explicitly with:
`AGENTS_HOME=$XDG_CONFIG_HOME/agents`

### 1) New resolver in shared utility crate
- Added `find_agents_home()` in `codex-rs/utils/home-dir/src/lib.rs`
- Added test coverage matching `find_codex_home()` cases:
  - missing path is fatal
  - file path is fatal
  - valid directory canonicalizes
  - unset env uses default home-based path

### 2) Skills loader now uses resolved agents home
- Updated user-scope root construction in `codex-rs/core/src/skills/loader.rs` to use:
  - `$AGENTS_HOME/skills` (defaulting to `~/.agents/skills`)
- Kept deprecated `$CODEX_HOME/skills` loading unchanged for backward compatibility.
- If `AGENTS_HOME` is invalid, loader logs a warning and skips that root instead of failing all skill discovery.

### 3) Test determinism improvements in loader tests
- Added a test helper to avoid ambient host user skills leaking into expectations during unit tests.
- This keeps `skills::loader` tests deterministic on machines where personal skills already exist.

## Backward compatibility
- ✅ Existing users of `~/.agents/skills` are unaffected.
- ✅ Existing users of deprecated `$CODEX_HOME/skills` are unaffected.
- ✅ New behavior is opt-in via `AGENTS_HOME`.

## How to use with XDG
Example:
```bash
export AGENTS_HOME="$XDG_CONFIG_HOME/agents"
```
Codex will then discover user skills from:
```text
$AGENTS_HOME/skills
```

## Validation
- `just fmt`
- `cargo test -p codex-utils-home-dir --locked`
- `nix develop /home/vx/git/codex --command sh -lc 'cd /home/vx/git/codex/codex-rs && GIT_TERMINAL_PROMPT=0 CARGO_NET_GIT_FETCH_WITH_CLI=true cargo test -p codex-core skills::loader'`

## Related issue
- #1980


Bad3r:

etraut-openai:

PR #11290, opened 2026-02-10
chore(tui) Simplify /status Permissions

dylan-hurd-oai:
## Summary
Consolidate `/status` Permissions lines into a simpler view. It should only show "Default," "Full Access," or "Custom" (with specifics)

## Testing
- [x] many snapshots updated


fouad-openai (approved):

codex-rs/tui/src/status/snapshots/codex_tui__status__tests__status_snapshot_includes_reasoning_details.snap
```diff
@@ -12,8 +12,7 @@ expression: sanitized
 │                                                                           │
 │  Model:            gpt-5.1-codex-max (reasoning high, summaries detailed) │
 │  Directory: [[workspace]]                                                 │
-│  Approval:         on-request                                             │
-│  Sandbox:          workspace-write                                        │
+│  Permissions:      Default                                                │
```
joshka-oai, line 15: The problem is this is it doesn't tell the user what this actually means and requires them to be familiar or lookup what the default is. I'd prefer to have a name for this if possible.

joshka-oai (approved):
It would be nice if this aligned perfectly with the menus (currently Default / Full Access), but to do that might require more plumbing that change all the way through properly. This is fine for now given that.

PR #11293, opened 2026-02-10
feat(linux-sandbox): implement proxy-only egress via TCP-UDS-TCP bridge

viyatb-oai:
## Summary
- Implement Linux proxy-only routing in `codex-rs/linux-sandbox` with a two-stage bridge: host namespace `loopback TCP proxy endpoint -> UDS`, then bwrap netns `loopback TCP listener -> host UDS`.
- Add hidden `--proxy-route-spec` plumbing for outer-to-inner stage handoff.
- Fail closed in proxy mode when no valid loopback proxy endpoints can be routed.
- Introduce explicit network seccomp modes: `Restricted` (legacy restricted networking) and `ProxyRouted` (allow INET/INET6 for routed proxy access, deny `AF_UNIX` and `socketpair`).
- Enforce that proxy bridge/routing is bwrap-only by validating `--apply-seccomp-then-exec` requires `--use-bwrap-sandbox`.
- Keep landlock-only flows unchanged (no proxy bridge behavior outside bwrap).


codex-rs/linux-sandbox/src/landlock.rs
```diff
@@ -85,6 +95,21 @@ fn should_install_network_seccomp(
...
+    proxy_routed_network: bool,
+) -> Option<NetworkSeccompMode> {
+    if !should_install_network_seccomp(sandbox_policy, allow_network_for_proxy) {
+        return None;
+    }
+    if proxy_routed_network {
+        Some(NetworkSeccompMode::ProxyRouted)
+    } else {
+        Some(NetworkSeccompMode::Restricted)
+    }
```
bolinfest, line 113:
```suggestion
    if !should_install_network_seccomp(sandbox_policy, allow_network_for_proxy) {
        None
    } else if proxy_routed_network {
        Some(NetworkSeccompMode::ProxyRouted)
    } else {
        Some(NetworkSeccompMode::Restricted)
    }
```

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -74,22 +80,36 @@ pub fn run_main() -> ! {
...
+            let spec = proxy_route_spec
+                .as_deref()
+                .unwrap_or_else(|| panic!("managed proxy mode requires --proxy-route-spec"));
+            if let Err(err) = activate_proxy_routes_in_netns(spec) {
+                panic!("error activating Linux proxy routing bridge: {err}");
+            }
+            true
+        } else {
+            false
+        };
```
bolinfest: Do you need/want the `proxy_routing_active` local variable?

```suggestion
        if allow_network_for_proxy {
            let spec = proxy_route_spec
                .as_deref()
                .unwrap_or_else(|| panic!("managed proxy mode requires --proxy-route-spec"));
            if let Err(err) = activate_proxy_routes_in_netns(spec) {
                panic!("error activating Linux proxy routing bridge: {err}");
            }
        }
        let proxy_routing_active = allow_network_for_proxy;
```

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -370,88 +412,5 @@ fn exec_or_panic(command: Vec<String>) -> ! {
...
-        assert_eq!(argv.contains(&"--unshare-net".to_string()), true);
-    }
-
-    #[test]
-    fn proxy_only_mode_takes_precedence_over_full_network_policy() {
-        let mode = bwrap_network_mode(&SandboxPolicy::DangerFullAccess, true);
-        assert_eq!(mode, BwrapNetworkMode::ProxyOnly);
-    }
-}
+#[path = "linux_run_main_tests.rs"]
```
bolinfest, line 425: TIL!

codex-rs/linux-sandbox/src/linux_run_main.rs
viyatb-oai: nope

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -373,4 +374,28 @@ mod tests {
...
+        let args = create_filesystem_args(&sandbox_policy, Path::new("/")).expect("bwrap fs args");
+
+        let dev_mount_index = args
+            .windows(2)
+            .position(|window| window[0] == "--dev" && window[1] == "/dev")
+            .expect("expected --dev /dev mount");
+        let writable_dev_bind_index = args
+            .windows(3)
+            .position(|window| window[0] == "--bind" && window[1] == "/dev" && window[2] == "/dev")
+            .expect("expected writable /dev bind");
```
bolinfest: Can we `assert_eq!()` on the full `args` instead?

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -373,4 +374,28 @@ mod tests {
...
+
+        assert_eq!(dev_mount_index < writable_dev_bind_index, true);
```
bolinfest: FYI, this is more canonical:

```suggestion
        assert!(dev_mount_index < writable_dev_bind_index);
```

codex-rs/linux-sandbox/src/landlock.rs
```diff
@@ -234,4 +286,28 @@ mod tests {
             true
         ));
     }
+
+    #[test]
+    fn managed_proxy_routes_use_proxy_routed_seccomp_mode() {
+        assert_eq!(
+            network_seccomp_mode(&SandboxPolicy::DangerFullAccess, true, true),
```
bolinfest, line 303: I guess `--dangerously-bypass-approvals-and-sandbox` doesn't mean exactly what it used to mean, so we may have to rework some of the terminology at some point?

codex-rs/linux-sandbox/src/linux_run_main_tests.rs
```diff
@@ -0,0 +1,158 @@
...
+fn proxy_only_mode_takes_precedence_over_full_network_policy() {
+    let mode = bwrap_network_mode(&SandboxPolicy::DangerFullAccess, true);
+    assert_eq!(mode, BwrapNetworkMode::ProxyOnly);
+}
+
+#[test]
+fn managed_proxy_preflight_argv_is_wrapped_for_full_access_policy() {
+    let mode = bwrap_network_mode(&SandboxPolicy::DangerFullAccess, true);
+    let argv = build_preflight_bwrap_argv(Path::new("/"), &SandboxPolicy::DangerFullAccess, mode);
+    assert_eq!(argv.contains(&"--".to_string()), true);
```
bolinfest: Is the `&` necessary before these string literals with the `to_string()`?

codex-rs/linux-sandbox/src/linux_run_main_tests.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+#[test]
+fn managed_proxy_inner_command_includes_route_spec() {
+    let args = build_inner_seccomp_command(
```
bolinfest, line 106: I feel like `build_inner_seccomp_command()` and `apply_sandbox_policy_to_current_thread()` are getting to the point where they should take a struct...

codex-rs/linux-sandbox/tests/suite/managed_proxy.rs
```diff
@@ -0,0 +1,313 @@
...
+fn is_bwrap_unavailable_output(output: &Output) -> bool {
+    String::from_utf8_lossy(&output.stderr).contains(BWRAP_UNAVAILABLE_ERR)
+}
+
+async fn should_skip_bwrap_tests() -> bool {
+    let mut env = create_env_from_core_vars();
+    strip_proxy_env(&mut env);
+
+    let output = run_linux_sandbox_direct(
+        &["bash", "-lc", "true"],
```
bolinfest: Unless you really care about `-l`, I would omit it from these integration tests since that imposes a greater dependency on the host env.

codex-rs/linux-sandbox/src/proxy_routing.rs
```diff
@@ -0,0 +1,740 @@
...
+        let message = if plan.has_proxy_config {
+            "managed proxy mode requires parseable loopback proxy endpoints"
+        } else {
+            "managed proxy mode requires proxy environment variables"
+        };
+        return Err(io::Error::new(io::ErrorKind::InvalidInput, message));
+    }
+
+    let temp_dir = std::env::temp_dir();
+    let _ = cleanup_stale_proxy_socket_dirs_in(temp_dir.as_path());
```
bolinfest: Note Codex can generally write to global `/tmp`, so not sure if that would put us at risk in any way here? Should we use `$CODEX_HOME/tmp` instead?

codex-rs/linux-sandbox/src/proxy_routing.rs
```diff
@@ -0,0 +1,740 @@
...
+    let ip = if host.eq_ignore_ascii_case("localhost") {
+        IpAddr::V4(Ipv4Addr::LOCALHOST)
+    } else {
+        host.parse::<IpAddr>().ok()?
+    };
+    if !ip.is_loopback() {
+        return None;
+    }
+
+    Some(SocketAddr::new(ip, port))
```
bolinfest:
```suggestion
    if ip.is_loopback() {
        Some(SocketAddr::new(ip, port))
    } else {
        None
    }
```

codex-rs/linux-sandbox/src/proxy_routing.rs
```diff
@@ -0,0 +1,740 @@
...
+        && !proxy_url.contains('#')
+        && rewritten.ends_with('/')
+    {
+        rewritten.pop();
+    }
+    Some(rewritten)
+}
+
+fn create_proxy_socket_dir() -> io::Result<PathBuf> {
+    let temp_dir = std::env::temp_dir();
```
bolinfest: Same here?

codex-rs/linux-sandbox/src/proxy_routing.rs
```diff
@@ -0,0 +1,740 @@
...
+
+fn set_parent_death_signal() -> io::Result<()> {
+    let res = unsafe { libc::prctl(libc::PR_SET_PDEATHSIG, libc::SIGTERM) };
+    if res != 0 {
+        return Err(io::Error::last_os_error());
+    }
+    if unsafe { libc::getppid() } == 1 {
+        return Err(io::Error::other("parent process already exited"));
+    }
+    Ok(())
```
bolinfest:
```suggestion
    if res != 0 {
        Err(io::Error::last_os_error())
    } else if unsafe { libc::getppid() } == 1 {
        Err(io::Error::other("parent process already exited"))
    } else {
        Ok(())
    }
```

bolinfest (approved):

codex-rs/linux-sandbox/src/landlock.rs
viyatb-oai: Yeah, agreed. This PR changes the practical meaning in managed-proxy mode, so the flag naming/wording is drifting. I left behavior as-is here and suggest a follow-up cleanup for terminology/docs so we can do it consistently across platforms.

codex-rs/linux-sandbox/src/linux_run_main_tests.rs
viyatb-oai: Agreed. They are at the point where a small args struct would improve readability. I kept this PR focused on the managed-proxy behavior + review fixes, but I am treating the struct refactor as a good follow-up cleanup.

PR #11305, opened 2026-02-10
Cache cloud requirements

gt-oai:
We're loading these from the web on every startup. This puts them in a local file with a 1hr TTL.

We sign the downloaded requirements with a key compiled into the Codex CLI to prevent unsophisticated tampering (determined circumvention is outside of our threat model: after all, one could just compile Codex without any of these checks).

If any of the following are true, we ignore the local cache and re-fetch from Cloud:
* The signature is invalid for the payload (== requirements, sign time, ttl, user identity)
* The identity does not match the auth'd user's identity
* The TTL has expired
* We cannot parse requirements.toml from the payload


codex-rs/cloud-requirements/src/lib.rs
```diff
@@ -9,27 +9,130 @@
...
     BackendClientInit,
     Request,
-    Parse,
+}
+
+#[derive(Clone, Debug, Eq, Error, PartialEq)]
+enum CacheLoadStatus {
+    #[error("Skipping cloud requirements cache read because auth identity is incomplete")]
+    AuthIdentityIncomplete,
+    #[error("cloud requirements cache file not found")]
```
gverma-openai: nit: first word is not capitalized in some errors and is in other error strings.

codex-rs/cloud-requirements/src/lib.rs
```diff
@@ -144,27 +250,39 @@ impl CloudRequirementsService {
...
         }
+        let token_data = auth.get_token_data().ok();
+        let chatgpt_user_id = token_data
+            .as_ref()
+            .and_then(|token_data| token_data.id_token.chatgpt_user_id.as_deref());
+        let account_id = auth.get_account_id();
+        let account_id = account_id.as_deref();
 
-        self.fetch_with_retries(&auth).await
+        match self.load_cache(chatgpt_user_id, account_id).await {
```
gverma-openai, line 260: If `load_cache` fails, do we fallback to network fetch (with the regular retries treatment)?

gverma-openai (approved):

codex-rs/cloud-requirements/src/lib.rs
gt-oai: yes we fall back (line 273)

codex-rs/cloud-requirements/src/lib.rs
gt-oai: done

Issue #52, opened 2025-04-16
Why requires an OpenAI API key if it runs in a network-disabled environment?


v1ckxy:
Is this some kind of magic?
![Image](https://github.com/user-attachments/assets/fbe0dd4c-c92b-430d-b576-27aadcccbcc3)

I've got a bunch of doubts atm:
- Which model it runs?
- Runs using GGML? EXL2?
- Where's the source code of the model?

tibo-openai:
The agent loop is powered by a model that gets sampled over the internet, but the command execution are performed in a network disabled sandbox. Suggest to read the code to understand more on how it all works!

v1ckxy:
I think you misunderstood me.
I am stating that:
NEITHER IT RUNS LOCALLY, NOR IS IT _OPEN_.

It is just another CLI for a third-party paid platform.

I already knew that the model is closed, even if the company is called 'OpenAI', and it runs on the internet, i.e. it is drug-dependent of the net, so it didn't make any sense to make those assumptions in first place.

Have a good one.

Issue #10486, opened 2026-02-03
Plan mode: write plan into markdown file


J3m5:
### What variant of Codex are you using?

CLI

### What feature would you like to see?

In plan mode, after the plan has been proposed, I'd like to be able to choose an option to write the plan into a markdown file

etraut-openai:
Have you tried asking codex to do that? You could also create a skill that you can invoke if you do this regularly.

J3m5:
Yes, I forgot to mention that, that's what I do, but it's not as simple as selecting an option and it could consume more tokens.

Using a skill for that may be an option, but I don't always want to export the plan.

etraut-openai:
You can invoke a skill on demand when you want to export the plan.

J3m5:
I also tried instructing Codex to print the map beforehand, in the plan prompt while in plan mode, but it didn't work because it can't modify files.
So we have to switch modes, then launch the prompt or use a skill to do it.

That would be a welcome improvement, but I understand that it is not a priority. I just wanted to get the discussion started.

wbdb:
Currently, I use Shift+Tab and instruct the .md file to create. In Plan mode, I was unsuccessful with the direct instruction to create the .md.

So yes, a third option would be great: remain in Plan mode, implement, export the plan to an md file (name and path entered via note).

J3m5:
A generic solution is coming: https://github.com/openai/codex/pull/11333#issuecomment-3881321542

runfence:
Somehow copy command doesn't copy the plan itself but only what it said before the plan

thekovic:
This is still unimplemented. `/copy` option doesn't include the plan content, only the response around it so this can't be even used as a workaround. Being able to conveniently export and store plans for later use is a vital part of mine (and I bet many others) workflow and I can't believe it's still missing.

runfence:
Claude is so much better with this because it initially writes plan as a file and all corrections are just edits to existing file instead of full plan rewrite.

PR #11333, opened 2026-02-10
feat(tui): add Alt+C and /copy to copy last agent response as markdown

fcoury:
## Problem

Users want to quickly copy the last agent response as markdown — for pasting into docs, issues, or chat. The TUI had no clipboard integration, and over SSH even if it did the clipboard would write to the _remote_ machine.

## Mental model

The copy feature maintains a **parallel timeline** of agent response markdown alongside the existing transcript cells. The transcript's `history` stores rendered `Line`/`Span` objects for display — the original markdown syntax (`#` headings, `**bold**`, code fences) is lost after rendering. Copying needs the raw markdown, so we keep it separately rather than modifying the shared display model that the rest of the TUI relies on for rendering, scrolling, and rollback.

Alternatives considered:

- **Store `raw_markdown` on `AgentMessageCell`** — couples copy logic to the display model, bloats every cell with an `Option<String>` even though only the last one matters, and rollback has to walk the history vec instead of a simple `Vec::pop`.
- **Query the backend transcript on demand** — `ChatWidget` only receives events; it has no direct access to the backend transcript. Plumbing an async query path or shared reference would be a bigger architectural change for a single feature.
- **Store just the turn ordinal, reconstruct from `plain_text()` on copy** — eliminates string duplication but makes _every_ copy lossy (not just the deep-rollback edge case), defeating the purpose of raw markdown.
- **Store only markdown, render lazily in `display_lines()`** — the most principled approach (markdown as single source of truth), but the streaming pipeline currently renders deltas incrementally into pre-rendered `Line`/`Span` cells, and a single agent response is split across multiple `AgentMessageCell` instances (one per stream chunk). This would require restructuring streaming to accumulate raw markdown, merging cells into one per response, and re-rendering on every draw (or caching, which is back to storing both). Valid long-term direction but too large a refactor for a copy-feature PR.

The current approach (parallel bounded vec) is the simplest that gives lossless copy in the common case without coupling to either the display model or the backend.

Two data structures track this:

- `last_agent_markdown` — the single string that Alt+C / `/copy` will send to the clipboard.
- `agent_turn_markdowns: Vec<AgentTurnMarkdown>` — a bounded (256-entry) vec with drain-on-overflow of `(ordinal, markdown)` pairs so that transcript rollbacks can truncate the timeline and recover the correct copy source.

The timeline is fed from four event sources (last writer wins within a turn):

1. `AgentMessage` — primary source; records commentary or full response text.
2. `PlanItemCompleted` — overwrites commentary with the completed plan text, so Alt+C copies the plan the user actually sees rather than a brief preamble.
3. `ExitedReviewMode` — records the rendered review output (explanation and/or findings) so Alt+C copies the review result.
4. `TurnComplete.last_agent_message` — fallback for providers that don't emit `AgentMessage`.

A `saw_copy_source_this_turn` flag prevents `TurnComplete` from recording a duplicate entry when any copy source (`AgentMessage`, plan item, or review output) was already recorded during the turn.

Clipboard writes use a two-tier strategy:

1. SSH detected → OSC 52 only (reaches local terminal).
2. Local → arboard (native) first, OSC 52 fallback.

> **What is OSC 52?** OSC 52 is a terminal escape sequence (Operating System
> Command #52) that asks the terminal emulator to write to the host's system
> clipboard. This is critical for SSH sessions: the TUI runs on the remote
> machine and has no access to the local clipboard, but the user's terminal
> emulator on the local machine interprets OSC 52 and writes to the local
> clipboard on their behalf. Most modern terminals (iTerm2, WezTerm, kitty,
> Windows Terminal, ghostty) support it.

OSC 52 is written to `/dev/tty` to bypass ratatui's alternate-screen buffer on stdout.

## Non-goals

- Copying tool-call output or reasoning traces. Note: commentary (`AgentMessage`) is recorded and copyable while the turn is still running; once a plan item or review output completes, it overwrites the commentary.

## Tradeoffs

| Decision                                    | Why                                                                                                                                                                                                                                                                        |
| ------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Bounded 256-entry history                   | Keeps memory bounded in long sessions; deep rollbacks past the boundary use a lossy transcript-cell reconstruction.                                                                                                                                                        |
| `plain_text()` fallback is lossy            | The rendered `Line`/`Span` cells lose original markdown syntax. Acceptable because it only triggers on deep rollbacks past the 256-entry window.                                                                                                                           |
| Transcript fallback only when history is empty | After a rollback, the transcript-derived fallback is only inserted when copy history is empty after truncation. If surviving entries remain, they are used as-is.                                                                                                       |
| OSC 52 payload limit of 100 kB              | Most terminals silently truncate larger payloads; we reject early with a clear error.                                                                                                                                                                                      |
| Alt+C instead of Cmd+C / Ctrl+C             | Cmd+C is not available to terminal apps — the terminal emulator intercepts it before it reaches the TUI. Ctrl+C is already wired to SIGINT. Alt+C is the conventional modifier for terminal apps that need a copy binding (e.g. tmux, Midnight Commander).                 |
| macOS `SuppressStderr` guard around arboard | `arboard::Clipboard::new()` initializes `NSPasteboard`, which triggers `os_log`/`NSLog` output on stderr. In raw-mode TUI this corrupts the display. RAII guard redirects fd 2 to `/dev/null` around the call, serialized by a process-wide mutex. No-op on Linux/Windows. |
| Plan text unconditionally overwrites commentary copy source | In plan mode, the model often emits brief commentary ("I'll create a plan") before the plan itself. The plan is what the user sees and wants to copy — not the preamble. |

## Architecture

```
Alt+C / /copy
│
▼
ChatWidget::copy_last_agent_markdown()
│ reads self.last_agent_markdown
▼
clipboard_copy::copy_to_clipboard()
│
├─ SSH? ──► osc52_copy() ──► /dev/tty or stdout
│
└─ Local? ─► arboard_copy() ──► [fail] ──► osc52_copy()
```

Copy-state lifecycle:
```
SessionConfigured    → clear all copy state
AgentMessage         → record_agent_markdown (primary source)
PlanItemCompleted    → record_agent_markdown (overwrites commentary with plan text)
ExitedReviewMode     → record_agent_markdown (captures rendered review output)
TurnComplete         → record_agent_markdown (fallback, only if no copy source recorded this turn)
TurnStarted          → reset saw_copy_source_this_turn
ContextCompacted     → no-op on copy state (display-only info event)
Rollback             → truncate_agent_turn_markdowns_to_turn_count + transcript fallback
```

Copy source priority within a turn:
```
AgentMessage (commentary) → PlanItemCompleted (plan text overwrites) → ExitedReviewMode (review output overwrites) → TurnComplete (fallback, only if no source recorded yet)
```

## Observability

- `tracing::warn` on native clipboard failure before OSC 52 fallback.
- `tracing::debug` on `/dev/tty` open failure before stdout fallback.
- Visible history-cell feedback: "Copied last message to clipboard" / "Copy failed: ..." / "No agent response to copy".

## Tests

~28 new tests covering:

- Rollback recomputes copy source to surviving agent message
- Rollback past bounded history uses transcript fallback
- Deep rollback (300 turns → rollback to turn 20) recovers correct source
- Context-compacted marker does not replace copy source
- TurnComplete fallback seeds copy when AgentMessage is missing
- TurnComplete does not duplicate when AgentMessage already recorded
- Plan item completion overwrites commentary as copy source
- Review output (explanation-only) updates copy source on ExitedReviewMode
- Review output (with findings) updates copy source on ExitedReviewMode
- Session reset clears copy state before replay
- Bounded history evicts oldest entries
- Alt+C clears armed quit-shortcut state
- `/copy` available during running task
- Clipboard strategy: SSH/local, native/OSC52, both-fail error aggregation
- OSC 52 encoding roundtrip, payload size rejection, writer passthrough
- Transcript fallback joins stream-continuation groups
- Transcript fallback ignores info-event markers

Manual testing performed on:

- macOS (local, native clipboard via arboard)
- Linux (local)
- Windows (local)
- macOS → macOS over SSH (OSC 52 to local terminal)

### DIFF
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -3212,6 +3212,93 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn rollback_ignores_context_compacted_marker_when_recomputing_copy_source() {
+        let (mut app, _app_event_rx, _op_rx) = make_test_app_with_channels().await;
+
+        let thread_id = ThreadId::new();
+        app.chat_widget.handle_codex_event(Event {
+            id: "session".to_string(),
+            msg: EventMsg::SessionConfigured(SessionConfiguredEvent {
+                session_id: thread_id,
+                forked_from_id: None,
+                thread_name: None,
+                model: "gpt-test".to_string(),
+                model_provider_id: "test-provider".to_string(),
+                approval_policy: AskForApproval::Never,
+                sandbox_policy: SandboxPolicy::ReadOnly,
+                cwd: PathBuf::from("/home/user/project"),
+                reasoning_effort: None,
+                history_log_id: 0,
+                history_entry_count: 0,
+                initial_messages: None,
+                rollout_path: Some(PathBuf::new()),
+            }),
+        });
+        app.chat_widget.handle_codex_event(Event {
+            id: "agent-1".to_string(),
+            msg: EventMsg::AgentMessage(codex_core::protocol::AgentMessageEvent {
+                message: "answer first".to_string(),
+            }),
+        });
+        app.chat_widget.handle_codex_event(Event {
+            id: "compacted".to_string(),
+            msg: EventMsg::ContextCompacted(codex_core::protocol::ContextCompactedEvent),
+        });
+        app.chat_widget.handle_codex_event(Event {
+            id: "agent-2".to_string(),
+            msg: EventMsg::AgentMessage(codex_core::protocol::AgentMessageEvent {
+                message: "answer second".to_string(),
+            }),
+        });
+
+        app.transcript_cells = vec![
+            Arc::new(UserHistoryCell {
+                message: "first".to_string(),
+                text_elements: Vec::new(),
+                local_image_paths: Vec::new(),
+            }) as Arc<dyn HistoryCell>,
+            Arc::new(AgentMessageCell::new(
+                vec![Line::from("answer first".to_string())],
+                true,
+            )) as Arc<dyn HistoryCell>,
+            Arc::new(UserHistoryCell {
+                message: "second".to_string(),
+                text_elements: Vec::new(),
+                local_image_paths: Vec::new(),
+            }) as Arc<dyn HistoryCell>,
+            Arc::new(AgentMessageCell::new(
+                vec![Line::from("Context compacted".to_string())],
+                true,
+            )) as Arc<dyn HistoryCell>,
+            Arc::new(UserHistoryCell {
+                message: "third".to_string(),
+                text_elements: Vec::new(),
+                local_image_paths: Vec::new(),
+            }) as Arc<dyn HistoryCell>,
+            Arc::new(AgentMessageCell::new(
+                vec![Line::from("answer second".to_string())],
+                true,
+            )) as Arc<dyn HistoryCell>,
+        ];
+
+        app.apply_backtrack_rollback(crate::app_backtrack::BacktrackSelection {
+            nth_user_message: 2,
+            prefill: String::new(),
+            text_elements: Vec::new(),
+            local_image_paths: Vec::new(),
+        });
+        app.handle_backtrack_event(&EventMsg::ThreadRolledBack(
+            codex_core::protocol::ThreadRolledBackEvent { num_turns: 1 },
+        ));
+
+        assert_eq!(app.chat_widget.agent_turn_markdown_count(), 1);
+        assert_eq!(
+            app.chat_widget.last_agent_markdown_text(),
+            Some("answer first"),
+        );
+    }
+
+    #[tokio::test]
     async fn rollback_to_first_user_clears_copy_source() {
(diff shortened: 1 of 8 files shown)


etraut-openai:
@fcoury, please resolve the merge conflicts.

etraut-openai:
@codex review

J3m5:
Will this feature resolve this issue? 
https://github.com/openai/codex/issues/10486

etraut-openai:
@codex review

fcoury:
> Will this feature resolve this issue? #10486

Yes, you just need to press Alt+C when the plan is output and it will be copied to the clipboard as markdown.

fcoury:
Closing in favor of https://github.com/openai/codex/pull/12613

PR #11336, opened 2026-02-10
Always expose view_image and return unsupported image-input error

aibrahim-oai:
- Keep `view_image` in the advertised tool list for all models.
- Return a clear error when the current model does not support image inputs, and cover it with a unit test.


aibrahim-oai:
@codex fix this

pakrym-oai (approved):

aibrahim-oai:
@codex review this

PR #11318, opened 2026-02-10
chore: split NPM packages

jif-oai:
## Summary

  This PR changes npm distribution for @openai/codex from one monolithic tarball to a split-package model:

  - @openai/codex becomes a small JS launcher package.
  - Native binaries move into per-platform packages (for example @openai/codex-darwin-arm64, @openai/codex-
    linux-x64-musl, etc.).
  - The root package uses optionalDependencies so npm installs only the matching platform package.

  ## Why

  Our current publish uploads a single large tarball (~206.5 MB, ~504.7 MB unpacked), which now fails in
  release CI with:

  - npm ERR! code E413
  - 413 Payload Too Large - PUT https://registry.npmjs.org/@openai%2fcodex

  Splitting by platform keeps each publish payload below registry limits and follows established native-CLI
  npm patterns.

  ## Behavior/Compatibility

  - npx @openai/codex and codex CLI UX stay the same.
  - No runtime binary download is required.
  - Unsupported platform/arch still returns a clear error.

  ## Release Workflow Changes

  - Build and pack platform-specific npm packages.
  - Publish root package + platform packages under the same version/tag (latest / alpha).
  - Keep trusted publishing/provenance flow unchanged.

PR #11339, opened 2026-02-10
# Use `@openai/codex` dist-tags for platform binaries instead of separate package names

bolinfest:
https://github.com/openai/codex/pull/11318 introduced logic to publish platform artifacts as separate npm packages (for example, `@openai/codex-darwin-arm64`, `@openai/codex-linux-x64`, etc.). That requires provisioning and maintaining multiple package entries in npm, which we want to avoid.

We still need to keep the package-size mitigation (platform-specific payloads), but we want that layout to live under a single npm package namespace (`@openai/codex`) using dist-tags.

We also need to preserve pre-release workflows where users install `@openai/codex@alpha` and get platform-appropriate binaries.

Additionally, we want GitHub Release assets to group Codex npm tarballs together, so platform tarballs should follow the same `codex-npm-*` filename prefix as the main Codex tarball.

## Release Strategy (New Scheme)

We publish **one npm package name for Codex binaries** (`@openai/codex`) and use **dist-tags** to select platform-specific payloads. This avoids creating separate platform package names while keeping the package size split by platform.

#### Mainline release (`x.y.z`)

- `@openai/codex@latest` (meta package)
- `@openai/codex@darwin-arm64`
- `@openai/codex@darwin-x64`
- `@openai/codex@linux-arm64`
- `@openai/codex@linux-x64`
- `@openai/codex@win32-arm64`
- `@openai/codex@win32-x64`
- `@openai/codex-responses-api-proxy@latest`
- `@openai/codex-sdk@latest`

#### Alpha release (`x.y.z-alpha.N`)

- `@openai/codex@alpha` (meta package)
- `@openai/codex@alpha-darwin-arm64`
- `@openai/codex@alpha-darwin-x64`
- `@openai/codex@alpha-linux-arm64`
- `@openai/codex@alpha-linux-x64`
- `@openai/codex@alpha-win32-arm64`
- `@openai/codex@alpha-win32-x64`
- `@openai/codex-responses-api-proxy@alpha`
- `@openai/codex-sdk@alpha`

As an example, the `package.json` for `@openai/codex@alpha` (using `0.99.0-alpha.17` as the `version`) would be:

```
{
  "name": "@openai/codex",
  "version": "0.99.0-alpha.17",
  "license": "Apache-2.0",
  "bin": {
    "codex": "bin/codex.js"
  },
  "type": "module",
  "engines": {
    "node": ">=16"
  },
  "files": [
    "bin"
  ],
  "repository": {
    "type": "git",
    "url": "git+https://github.com/openai/codex.git",
    "directory": "codex-cli"
  },
  "packageManager": "pnpm@10.28.2+sha512.41872f037ad22f7348e3b1debbaf7e867cfd448f2726d9cf74c08f19507c31d2c8e7a11525b983febc2df640b5438dee6023ebb1f84ed43cc2d654d2bc326264",
  "optionalDependencies": {
    "@openai/codex-linux-x64": "npm:@openai/codex@0.99.0-alpha.17-linux-x64",
    "@openai/codex-linux-arm64": "npm:@openai/codex@0.99.0-alpha.17-linux-arm64",
    "@openai/codex-darwin-x64": "npm:@openai/codex@0.99.0-alpha.17-darwin-x64",
    "@openai/codex-darwin-arm64": "npm:@openai/codex@0.99.0-alpha.17-darwin-arm64",
    "@openai/codex-win32-x64": "npm:@openai/codex@0.99.0-alpha.17-win32-x64",
    "@openai/codex-win32-arm64": "npm:@openai/codex@0.99.0-alpha.17-win32-arm64"
  }
}
```

Note that the keys in `optionalDependencies` have "clean" names, but the values have the tag embedded.

### Important note

**Note:** Because we never created the new platform package names on npm (for example,
`@openai/codex-darwin-arm64`) since #11318 landed, there are no extra npm packages to clean up.

### 1. Stage platform tarballs as `@openai/codex` with platform-specific versions

`codex-cli/scripts/build_npm_package.py`

- Added `CODEX_NPM_NAME = "@openai/codex"` and platform metadata `npm_tag` values:
  - `darwin-arm64`, `darwin-x64`, `linux-arm64`, `linux-x64`, `win32-arm64`, `win32-x64`
- For platform package staging (`codex-<platform>` inputs), switched generated `package.json` from:
  - `name = @openai/codex-<platform>`
  to:
  - `name = @openai/codex`
- Added `compute_platform_package_version(version, platform_tag)` so platform tarballs have unique
  versions (`<release-version>-<platform-tag>`), which is required because npm forbids re-publishing
  the same `name@version`.

### 2. Point meta package optional dependencies at dist-tags on `@openai/codex`

`codex-cli/scripts/build_npm_package.py`

- Updated `optionalDependencies` generation for the main `codex` package to use npm alias syntax:
  - key remains alias package name (for example, `@openai/codex-darwin-arm64`) so runtime lookup behavior is unchanged
  - value now resolves to `@openai/codex` by dist-tag
- Stable releases emit tags like `npm:@openai/codex@darwin-arm64`.
- Alpha releases (`x.y.z-alpha.N`) emit tags like `npm:@openai/codex@alpha-darwin-arm64`.

### 3. Publish with per-tarball dist-tags in release CI

`.github/workflows/rust-release.yml`

- Reworked npm publish logic to derive the publish tag per tarball filename:
  - platform tarballs publish with `<platform>` tags for stable releases
  - platform tarballs publish with `alpha-<platform>` tags for alpha releases
  - top-level tarballs (`codex`, `codex-responses-api-proxy`, `codex-sdk`) continue using
    the existing channel tag policy (`latest` implicit for stable, `alpha` for alpha)
- Added fail-fast behavior for unexpected tarball names to avoid silent mispublishes.

### 4. Normalize Codex platform tarball filenames for GitHub Release grouping

Files: `scripts/stage_npm_packages.py`, `.github/workflows/rust-release.yml`

- Renamed staged platform tarball filenames from:
  - `codex-linux-<arch>-npm-<version>.tgz`
  - `codex-darwin-<arch>-npm-<version>.tgz`
  - `codex-win32-<arch>-npm-<version>.tgz`
- To:
  - `codex-npm-linux-<arch>-<version>.tgz`
  - `codex-npm-darwin-<arch>-<version>.tgz`
  - `codex-npm-win32-<arch>-<version>.tgz`

This keeps all Codex npm artifacts grouped under a common `codex-npm-` prefix in GitHub Releases.

### 5. Documentation update

`codex-cli/scripts/README.md`

- Updated staging docs to clarify that platform-native variants are published as dist-tagged
  `@openai/codex` artifacts rather than separate npm package names.

## Resulting behavior

- Mainline release:
  - `@openai/codex@latest` resolves the meta package
  - meta package optional dependencies resolve `@openai/codex@<platform-tag>`
- Alpha release:
  - users can continue installing `@openai/codex@alpha`
  - alpha meta package optional dependencies resolve `@openai/codex@alpha-<platform-tag>`
- Release assets:
  - Codex npm tarballs share `codex-npm-` prefix for cleaner grouping in GitHub Releases

This preserves platform-specific payload distribution while avoiding separate npm package names and
improves release-asset discoverability.

## Validation notes

- Verified staged `package.json` output for stable and alpha meta packages includes expected alias targets.
- Verified staged platform package manifests are `name=@openai/codex` with unique platform-suffixed versions.
- Verified publish tag derivation maps renamed platform tarballs to expected stable and alpha dist-tags.

### DIFF
--- a/codex-cli/scripts/build_npm_package.py
+++ b/codex-cli/scripts/build_npm_package.py
@@ -14,40 +14,49 @@ CODEX_CLI_ROOT = SCRIPT_DIR.parent
 REPO_ROOT = CODEX_CLI_ROOT.parent
 RESPONSES_API_PROXY_NPM_ROOT = REPO_ROOT / "codex-rs" / "responses-api-proxy" / "npm"
 CODEX_SDK_ROOT = REPO_ROOT / "sdk" / "typescript"
+CODEX_NPM_NAME = "@openai/codex"
 
+# `npm_name` is the local optional-dependency alias consumed by `bin/codex.js`.
+# The underlying package published to npm is always `@openai/codex`.
 CODEX_PLATFORM_PACKAGES: dict[str, dict[str, str]] = {
     "codex-linux-x64": {
         "npm_name": "@openai/codex-linux-x64",
+        "npm_tag": "linux-x64",
         "target_triple": "x86_64-unknown-linux-musl",
         "os": "linux",
         "cpu": "x64",
     },
     "codex-linux-arm64": {
         "npm_name": "@openai/codex-linux-arm64",
+        "npm_tag": "linux-arm64",
         "target_triple": "aarch64-unknown-linux-musl",
         "os": "linux",
         "cpu": "arm64",
     },
     "codex-darwin-x64": {
         "npm_name": "@openai/codex-darwin-x64",
+        "npm_tag": "darwin-x64",
         "target_triple": "x86_64-apple-darwin",
         "os": "darwin",
         "cpu": "x64",
     },
     "codex-darwin-arm64": {
         "npm_name": "@openai/codex-darwin-arm64",
+        "npm_tag": "darwin-arm64",
         "target_triple": "aarch64-apple-darwin",
         "os": "darwin",
         "cpu": "arm64",
     },
     "codex-win32-x64": {
         "npm_name": "@openai/codex-win32-x64",
+        "npm_tag": "win32-x64",
         "target_triple": "x86_64-pc-windows-msvc",
         "os": "win32",
         "cpu": "x64",
     },
     "codex-win32-arm64": {
         "npm_name": "@openai/codex-win32-arm64",
+        "npm_tag": "win32-arm64",
         "target_triple": "aarch64-pc-windows-msvc",
         "os": "win32",
         "cpu": "arm64",
@@ -244,6 +253,8 @@ def stage_sources(staging_dir: Path, version: str, package: str) -> None:
         package_json_path = CODEX_CLI_ROOT / "package.json"
     elif package in CODEX_PLATFORM_PACKAGES:
         platform_package = CODEX_PLATFORM_PACKAGES[package]
+        platform_npm_tag = platform_package["npm_tag"]
+        platform_version = compute_platform_package_version(version, platform_npm_tag)
 
         readme_src = REPO_ROOT / "README.md"
         if readme_src.exists():
@@ -253,8 +264,8 @@ def stage_sources(staging_dir: Path, version: str, package: str) -> None:
             codex_package_json = json.load(fh)
 
         package_json = {
-            "name": platform_package["npm_name"],
-            "version": version,
+            "name": CODEX_NPM_NAME,
+            "version": platform_version,
             "license": codex_package_json.get("license", "Apache-2.0"),
             "os": [platform_package["os"]],
             "cpu": [platform_package["cpu"]],
@@ -294,7 +305,10 @@ def stage_sources(staging_dir: Path, version: str, package: str) -> None:
     if package == "codex":
         package_json["files"] = ["bin"]
         package_json["optionalDependencies"] = {
-            CODEX_PLATFORM_PACKAGES[platform_package]["npm_name"]: version
+            CODEX_PLATFORM_PACKAGES[platform_package]["npm_name"]: (
+                f"npm:{CODEX_NPM_NAME}@"
+                f"{compute_platform_package_version(version, CODEX_PLATFORM_PACKAGES[platform_package]['npm_tag'])}"
+            )
             for platform_package in PACKAGE_EXPANSIONS["codex"]
             if platform_package != "codex"
         }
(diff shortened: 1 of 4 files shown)


codex-cli/scripts/build_npm_package.py
```diff
@@ -14,40 +15,49 @@
 REPO_ROOT = CODEX_CLI_ROOT.parent
 RESPONSES_API_PROXY_NPM_ROOT = REPO_ROOT / "codex-rs" / "responses-api-proxy" / "npm"
 CODEX_SDK_ROOT = REPO_ROOT / "sdk" / "typescript"
+CODEX_NPM_NAME = "@openai/codex"
 
+# `npm_name` is the local optional-dependency alias consumed by `bin/codex.js`.
```
jif-oai, line 19: Shouldn't we drop it then/

codex-cli/scripts/build_npm_package.py
```diff
@@ -316,6 +331,20 @@ def stage_sources(staging_dir: Path, version: str, package: str) -> None:
...
 
+def is_alpha_release(version: str) -> bool:
+    return re.match(r"^[0-9]+\.[0-9]+\.[0-9]+-alpha\.[0-9]+$", version) is not None
+
+
+def compute_platform_dist_tag(version: str, platform_tag: str) -> str:
+    return f"alpha-{platform_tag}" if is_alpha_release(version) else platform_tag
+
+
+def compute_platform_package_version(version: str, platform_tag: str) -> str:
```
jif-oai, line 333: I would inline this

codex-cli/scripts/build_npm_package.py
```diff
@@ -316,6 +331,20 @@ def stage_sources(staging_dir: Path, version: str, package: str) -> None:
         out.write("\n")
 
 
+def is_alpha_release(version: str) -> bool:
+    return re.match(r"^[0-9]+\.[0-9]+\.[0-9]+-alpha\.[0-9]+$", version) is not None
+
+
+def compute_platform_dist_tag(version: str, platform_tag: str) -> str:
```
jif-oai: I would inline this

jif-oai (approved):

codex-cli/scripts/build_npm_package.py
bolinfest: I think this is accurate. In the PR body, I added the `package.json` we should expect. As we can see `npm_name` here is the key in `"optionalDependencies"`, the naming of which `codex.js` depends on here:

https://github.com/openai/codex/blob/a364dd8b56a7406aadd8813716aa4aa34612fc5c/codex-cli/bin/codex.js#L16-L21

And that will be the subdirectory under `node_modules` where things should be found, I believe?

PR #11341, opened 2026-02-10
Promote Windows Sandbox

iceweasel-oai:
1. Move Windows Sandbox NUX to right after trust directory screen
2. Don't offer read-only as an option in Sandbox NUX. Elevated/Legacy/Quit
3. Don't allow new untrusted directories. It's trust or quit
4. move experimental sandbox features to `[windows] sandbox="elevated|unelevatd"`
5. Copy tweaks = elevated -> default, non-elevated -> non-admin


codex-rs/core/src/config/edit.rs
```diff
@@ -819,6 +819,43 @@ impl ConfigEditsBuilder {
...
+    pub fn set_windows_sandbox_mode(mut self, mode: &str) -> Self {
+        let mut segments = vec!["windows".to_string(), "sandbox".to_string()];
+        if let Some(profile) = self.profile.as_ref() {
+            segments = vec![
+                "profiles".to_string(),
+                profile.clone(),
+                "windows".to_string(),
+                "sandbox".to_string(),
+            ];
+        }
```
dylan-hurd-oai:
```suggestion
        let segments = if let Some(profile) = self.profile.as_ref() {
            vec![
                "profiles".to_string(),
                profile.clone(),
                "windows".to_string(),
                "sandbox".to_string(),
            ]
        } else {
         vec!["windows".to_string(), "sandbox".to_string()]
        }
```

codex-rs/core/src/features.rs
```diff
@@ -320,6 +322,10 @@ impl Features {
         }
 
         overrides.apply(&mut features);
+        apply_windows_sandbox_mode(
+            &mut features,
+            resolve_windows_sandbox_mode(cfg, config_profile),
+        );
```
dylan-hurd-oai: should we really be potentially mutating the config every time we load it? I see the reasoning for resolving to a healthy config but curious if this is the right way to do so

codex-rs/tui/src/app_event.rs
```diff
@@ -203,6 +203,12 @@ pub(crate) enum AppEvent {
         preset: ApprovalPreset,
     },
 
+    /// Begin the non-elevated Windows sandbox setup flow.
+    #[cfg_attr(not(target_os = "windows"), allow(dead_code))]
+    BeginWindowsSandboxLegacySetup {
```
dylan-hurd-oai, line 208: We refer to this as non-admin in the product but Legacy in the code - should we standardize? wasn't obvious to me

codex-rs/tui/src/chatwidget.rs
```diff
@@ -5748,9 +5713,24 @@ impl ChatWidget {
                 ..Default::default()
             },
             SelectionItem {
-                name: stay_label,
+                name: "Use non-admin sandbox (higher risk if prompt injected)".to_string(),
```
dylan-hurd-oai, line 5717: Will we have already shown the user the warning in this case? Do we need the parentheses here? Also `Use the`

codex-rs/tui/src/chatwidget.rs
```diff
@@ -5778,58 +5758,27 @@ impl ChatWidget {
...
-            });
         let mut lines = Vec::new();
-        lines.push(line!["Use Non-Elevated Sandbox?".bold()]);
+        lines.push(line![
+            "Couldn't set up your sandbox with Administrator permissions".bold()
+        ]);
         lines.push(line![""]);
         lines.push(line![
-            "Elevation failed. You can also use a non-elevated sandbox, which protects your files and prevents network access under most circumstances. However, it carries greater risk if prompt injected."
+            "You can still use Codex in a non-admin sandbox. It carries greater risk if prompt injected."
```
dylan-hurd-oai, line 5768: Why remove `which protects your files and prevents network access under most circumstances.` ?

codex-rs/tui/src/chatwidget.rs
```diff
@@ -5778,58 +5758,27 @@ impl ChatWidget {
...
-                move |_tx| {
-                    otel.counter("codex.windows_sandbox.fallback_stay_current", 1, &[]);
-                }
-            }),
-        );
+        let quit_otel = self.otel_manager.clone();
         let items = vec![
             SelectionItem {
-                name: "Try elevated agent sandbox setup again".to_string(),
+                name: "Try setting up admin sandbox again".to_string(),
```
dylan-hurd-oai, line 5782: Let's remove `Try` here? Default expectation now is that it should succeed 😁

codex-rs/tui/src/lib.rs
```diff
@@ -215,7 +215,6 @@ pub async fn run_main(
     {
         tracing::warn!(error = %err, "failed to run personality migration");
     }
-
```
dylan-hurd-oai, line 218: nit: keep?

dylan-hurd-oai (commented):
Some light things to discuss but nothing hugely blocking

codex-rs/core/src/features.rs
iceweasel-oai: gonna clean this up so we don't ever rely on the old features. This doesn't write the config to disk, but it's still a bit wonky

codex-rs/tui/src/app_event.rs
iceweasel-oai: we should. I just wanna do it in a follow-up since Legacy probably appears in a bunch of places (including metrics tags, etc.)

codex-rs/tui/src/chatwidget.rs
iceweasel-oai: this is the first time they are seeing really anything about sandbox modes

codex-rs/tui/src/chatwidget.rs
iceweasel-oai: I think we might want to bring that back. Will discuss

codex-rs/tui/src/chatwidget.rs
iceweasel-oai: this is after it failed, so I think it makes sense here.

dylan-hurd-oai: ah missed that, sounds good!

codex-rs/core/src/config/mod.rs
```diff
@@ -1467,7 +1474,11 @@ impl Config {
             || config_profile.sandbox_mode.is_some()
             || cfg.sandbox_mode.is_some();
 
-        let windows_sandbox_level = WindowsSandboxLevel::from_features(&features);
+        let windows_sandbox_level = match windows_sandbox_mode {
+            Some(WindowsSandboxModeToml::Elevated) => WindowsSandboxLevel::Elevated,
+            Some(WindowsSandboxModeToml::Unelevated) => WindowsSandboxLevel::RestrictedToken,
```
dylan-hurd-oai, line 1479: `WindowsSandboxModeToml` and `WindowsSandboxLevel` feel mismatched - should we rename to `WindowsSandboxLevelToml`?

dylan-hurd-oai (approved):
One final comment about naming, LTGM after we fix CI!

PR #11346, opened 2026-02-10
Sanitize MCP image output for text-only models

aibrahim-oai:
- Replace image blocks in MCP tool results with a text placeholder when the active model does not accept image input.
- Add an e2e rmcp test to verify sanitized tool output is what gets sent back to the model.


aibrahim-oai:
@codex review this

pakrym-oai (approved):

PR #11359, opened 2026-02-10
ci: fall back to local Bazel on forks without BuildBuddy key

joshka-oai:
## Summary
- detect whether BUILDBUDDY_API_KEY is present in Bazel CI
- keep existing remote BuildBuddy path when key is available
- add a local fallback path for fork PRs without secrets by clearing remote cache/executor/BES endpoints
- document each fallback flag inline with links to Bazel docs

## Testing
- ruby -e 'require "yaml"; YAML.load_file(".github/workflows/bazel.yml"); puts "ok"'
- verified Bazel docs/flag references used in workflow comments

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -103,9 +103,48 @@ jobs:
           CODEX_BWRAP_ENABLE_FFI: ${{ contains(matrix.target, 'unknown-linux') && '1' || '0' }}
         shell: bash
         run: |
-          bazel $BAZEL_STARTUP_ARGS --bazelrc=.github/workflows/ci.bazelrc test //... \
-            --build_metadata=REPO_URL=https://github.com/openai/codex.git \
-            --build_metadata=COMMIT_SHA=$(git rev-parse HEAD) \
-            --build_metadata=ROLE=CI \
-            --build_metadata=VISIBILITY=PUBLIC \
-            "--remote_header=x-buildbuddy-api-key=$BUILDBUDDY_API_KEY"
+          bazel_args=(
+            test
+            //...
+            --build_metadata=REPO_URL=https://github.com/openai/codex.git
+            --build_metadata=COMMIT_SHA=$(git rev-parse HEAD)
+            --build_metadata=ROLE=CI
+            --build_metadata=VISIBILITY=PUBLIC
+          )
+
+          if [[ -n "${BUILDBUDDY_API_KEY:-}" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/bazel.yml
```diff
@@ -103,9 +103,48 @@ jobs:
...
+            #   https://bazel.build/reference/command-line-reference#common_options-flag--bes_backend
+            #   https://bazel.build/reference/command-line-reference#common_options-flag--bes_results_url
+            # --remote_cache= and --remote_executor=:
+            #   clear remote cache/execution endpoints configured in .bazelrc.
+            #   https://bazel.build/reference/command-line-reference#common_options-flag--remote_cache
+            #   https://bazel.build/reference/command-line-reference#common_options-flag--remote_executor
+            bazel $BAZEL_STARTUP_ARGS \
+              --noexperimental_remote_repo_contents_cache \
+              "${bazel_args[@]}" \
+              --bes_backend= \
```
zbarsky-openai: best_backend and bes_results_url are fine to keep enabled, those work without auth (they just log a warning) and they can be a nicer visualization. not a big deal though!

zbarsky-openai (approved):

.github/workflows/bazel.yml
joshka-oai: Good call, updated. I removed `--bes_backend=` and `--bes_results_url=` from the fork/local fallback path so BES remains enabled. We still clear `--remote_cache=` and `--remote_executor=` (and disable remote repo contents cache) to avoid auth failures when `BUILDBUDDY_API_KEY` is absent.

.github/workflows/bazel.yml
```diff
@@ -103,9 +103,42 @@ jobs:
...
+
+          if [[ -n "${BUILDBUDDY_API_KEY:-}" ]]; then
+            echo "BuildBuddy API key is available; using remote Bazel configuration."
+            bazel $BAZEL_STARTUP_ARGS \
+              --bazelrc=.github/workflows/ci.bazelrc \
+              "${bazel_args[@]}" \
+              "--remote_header=x-buildbuddy-api-key=$BUILDBUDDY_API_KEY"
+          else
+            echo "BuildBuddy API key is not available; using local Bazel configuration."
+            # Keep fork/community PRs on Bazel but disable remote services that are
```
joshka-oai, line 123: Note - docs intentionally verbose here because bazel familiarity in the team is low. These can be moved above / elsewhere as needed, but figured having links to how this works would be helpful in understanding any future problems with this code (instead of having to search for this info)

PR #11368, opened 2026-02-10
fix(network-proxy): add unix socket allow-all and update seatbelt rules

viyatb-oai:
## Summary
Adds support for a Unix socket escape hatch so we can bypass socket allowlisting when explicitly enabled.

## Description
* added a new flag, `network.dangerously_allow_all_unix_sockets` as an explicit escape hatch
* In codex-network-proxy, enabling that flag now allows any absolute Unix socket path from x-unix-socket instead of requiring each path to be explicitly allowlisted. Relative paths are still rejected.
* updated the macOS seatbelt path in core so it enforces the same Unix socket behavior:
  * allowlisted sockets generate explicit network* subpath rules
  * allow-all generates a broad network* (subpath "/") rule


codex-rs/core/src/seatbelt.rs
```diff
@@ -101,18 +100,13 @@ fn proxy_loopback_ports_from_env(env: &HashMap<String, String>) -> Vec<u16> {
...
-        trimmed == "1" || trimmed.eq_ignore_ascii_case("true")
-    })
-}
-
 #[derive(Debug, Default)]
 struct ProxyPolicyInputs {
     ports: Vec<u16>,
     has_proxy_config: bool,
     allow_local_binding: bool,
+    allow_unix_sockets: Vec<String>,
```
bolinfest: Would this be better as `Vec<AbsolutePathBuf>`?

codex-rs/core/src/seatbelt.rs
```diff
@@ -122,13 +116,48 @@ fn proxy_policy_inputs(network: Option<&NetworkProxy>) -> ProxyPolicyInputs {
...
+            allow_local_binding: network.allow_local_binding(),
+            allow_unix_sockets: network.allow_unix_sockets().to_vec(),
+            dangerously_allow_all_unix_sockets: network.dangerously_allow_all_unix_sockets(),
         };
     }
 
     ProxyPolicyInputs::default()
 }
 
+fn normalize_path_for_sandbox(path: &Path) -> Option<PathBuf> {
```
bolinfest: `AbsolutePathBuf` exists so we don't have to reinvent this sort of logic.

codex-rs/core/src/seatbelt.rs
```diff
@@ -122,13 +116,48 @@ fn proxy_policy_inputs(network: Option<&NetworkProxy>) -> ProxyPolicyInputs {
...
+        let Some(normalized_path) = normalize_path_for_sandbox(Path::new(socket_path)) else {
+            continue;
+        };
+        paths.insert(normalized_path);
+    }
+
+    paths
+        .into_iter()
+        .map(|path| {
+            let escaped_path = escape_seatbelt_string(path.to_string_lossy().as_ref());
```
bolinfest: The safer thing is to use the `-D` stuff, I believe.

codex-rs/core/src/seatbelt.rs
```diff
@@ -122,13 +116,48 @@ fn proxy_policy_inputs(network: Option<&NetworkProxy>) -> ProxyPolicyInputs {
...
+        return None;
+    }
+    Some(path.canonicalize().unwrap_or_else(|_| path.to_path_buf()))
+}
+
+fn escape_seatbelt_string(value: &str) -> String {
+    value.replace('\\', "\\\\").replace('"', "\\\"")
+}
+
+fn unix_socket_policy(proxy: &ProxyPolicyInputs) -> String {
```
bolinfest, line 226: Comment that return value will always end with a newline (that's true, right?)?

codex-rs/core/src/seatbelt.rs
```diff
@@ -122,13 +116,48 @@ fn proxy_policy_inputs(network: Option<&NetworkProxy>) -> ProxyPolicyInputs {
...
+    if proxy.dangerously_allow_all_unix_sockets {
+        return "(allow network* (subpath \"/\"))\n".to_string();
+    }
+
+    let mut paths = BTreeSet::new();
+    for socket_path in &proxy.allow_unix_sockets {
+        let Some(normalized_path) = normalize_path_for_sandbox(Path::new(socket_path)) else {
+            continue;
```
bolinfest: Should this Err or silently ignore?

Should at least log?

codex-rs/core/src/seatbelt.rs
```diff
@@ -122,13 +116,48 @@ fn proxy_policy_inputs(network: Option<&NetworkProxy>) -> ProxyPolicyInputs {
...
 
+fn normalize_path_for_sandbox(path: &Path) -> Option<PathBuf> {
+    if !path.is_absolute() {
+        return None;
+    }
+    Some(path.canonicalize().unwrap_or_else(|_| path.to_path_buf()))
+}
+
+fn escape_seatbelt_string(value: &str) -> String {
+    value.replace('\\', "\\\\").replace('"', "\\\"")
```
bolinfest: How confident are we in this? The `-D` is more annoying to manage, but we don't have to worry about being tripped up by escaping issues.

codex-rs/core/src/seatbelt.rs
viyatb-oai: yeah `AbsolutePathBuf` makes more sense.

codex-rs/core/src/seatbelt.rs
viyatb-oai: oops, yes!

codex-rs/core/src/seatbelt.rs
viyatb-oai: adding a log so but it still fails non-fataly

codex-rs/core/src/seatbelt.rs
viyatb-oai: added.

codex-rs/core/src/seatbelt.rs
```diff
@@ -110,34 +112,96 @@ fn proxy_loopback_ports_from_env(env: &HashMap<String, String>) -> Vec<u16> {
...
+        for socket_path in network.allow_unix_sockets() {
+            match normalize_path_for_sandbox(Path::new(socket_path)) {
+                Some(path) => allow_unix_sockets.push(path),
+                None => {
+                    warn!(
+                        "ignoring network.allow_unix_sockets entry because it could not be normalized: {socket_path}"
+                    );
+                }
+            }
+        }
```
bolinfest: Note: could consider `filter_map()`:

```suggestion
let allow_unix_sockets: Vec<_> = network
    .allow_unix_sockets()
    .iter()
    .filter_map(|socket_path| {
        normalize_path_for_sandbox(Path::new(socket_path)).or_else(|| {
            warn!(
                "ignoring network.allow_unix_sockets entry because it could not be normalized: {socket_path}"
            );
            None
        })
    })
    .collect();
```

codex-rs/core/src/seatbelt.rs
```diff
@@ -110,34 +112,96 @@ fn proxy_loopback_ports_from_env(env: &HashMap<String, String>) -> Vec<u16> {
...
+            allow_unix_sockets,
+            dangerously_allow_all_unix_sockets: network.dangerously_allow_all_unix_sockets(),
         };
     }
 
     ProxyPolicyInputs::default()
 }
 
+fn normalize_path_for_sandbox(path: &Path) -> Option<AbsolutePathBuf> {
+    if !path.is_absolute() {
```
bolinfest, line 182: `AbsolutePathBuf::from_absolute_path()` should take care of this, so you can skip this, I think?

codex-rs/core/src/seatbelt.rs
```diff
@@ -110,34 +112,96 @@ fn proxy_loopback_ports_from_env(env: &HashMap<String, String>) -> Vec<u16> {
...
     allow_local_binding: bool,
+    allow_unix_sockets: Vec<AbsolutePathBuf>,
+    dangerously_allow_all_unix_sockets: bool,
 }
 
 fn proxy_policy_inputs(network: Option<&NetworkProxy>) -> ProxyPolicyInputs {
     if let Some(network) = network {
         let mut env = HashMap::new();
         network.apply_to_env(&mut env);
+        let mut allow_unix_sockets = Vec::new();
```
bolinfest: Because `normalize_path_for_sandbox()` canonicalizes, note I believe this could have duplicates even if `allow_unix_sockets()` has unique values?

codex-rs/core/src/seatbelt.rs
```diff
@@ -110,34 +112,96 @@ fn proxy_loopback_ports_from_env(env: &HashMap<String, String>) -> Vec<u16> {
...
-    })
-}
-
 #[derive(Debug, Default)]
 struct ProxyPolicyInputs {
     ports: Vec<u16>,
     has_proxy_config: bool,
     allow_local_binding: bool,
+    allow_unix_sockets: Vec<AbsolutePathBuf>,
+    dangerously_allow_all_unix_sockets: bool,
```
bolinfest: Note if this is `true`, then `allow_unix_sockets` is ignored?

Or make this a disjoint union like the following to eliminate the invalid state:

```rust
enum UnixDomainSocketPolicy {
    AllowAll,
    Restricted {
        allowed: Vec<AbsolutePathBuf>
    }
}
```

codex-rs/core/src/seatbelt.rs
```diff
@@ -110,34 +112,96 @@ fn proxy_loopback_ports_from_env(env: &HashMap<String, String>) -> Vec<u16> {
...
+    let absolute_path = AbsolutePathBuf::from_absolute_path(path).ok()?;
+    let normalized_path = absolute_path
+        .as_path()
+        .canonicalize()
+        .ok()
+        .and_then(|canonical_path| AbsolutePathBuf::from_absolute_path(canonical_path).ok());
+    normalized_path.or(Some(absolute_path))
+}
+
+fn unix_socket_path_params(proxy: &ProxyPolicyInputs) -> Vec<(String, AbsolutePathBuf)> {
```
bolinfest: Docstring for the relationship between the `String` and `AbsolutePathBuf` in each tuple? Or introduce a self-documenting struct?

codex-rs/network-proxy/src/config.rs
```diff
@@ -172,7 +176,24 @@ pub struct RuntimeConfig {
     pub admin_addr: SocketAddr,
 }
 
+pub(crate) fn is_absolute_unix_socket_path(socket_path: &str) -> bool {
+    Path::new(socket_path).is_absolute() || socket_path.starts_with('/')
```
bolinfest: Absolute is not the same as normalized, I believe. Is that OK for your purposes? Can we use `AbsolutePathBuf` instead?

bolinfest: Also, why the `starts_with()` check?

codex-rs/network-proxy/src/config.rs
```diff
@@ -526,4 +548,61 @@ mod tests {
...
+            err.to_string().contains("network.allow_unix_sockets[0]"),
+            "error should point at the invalid allow_unix_sockets entry: {err:#}"
+        );
+    }
+
+    #[test]
+    fn resolve_runtime_accepts_unix_style_absolute_allow_unix_sockets_entries() {
+        let cfg = NetworkProxyConfig {
+            network: NetworkProxySettings {
+                allow_unix_sockets: vec!["/tmp/example.sock".to_string()],
```
bolinfest: Note `/tmp` is not a canonical path on MacOS. Not sure if that's important for these tests?

codex-rs/network-proxy/src/runtime.rs
```diff
@@ -425,13 +425,22 @@ impl NetworkProxyState {
...
+
         // Normalize the path while keeping the absolute-path requirement explicit.
         let requested_abs = match AbsolutePathBuf::from_absolute_path(requested_path) {
             Ok(path) => path,
             Err(_) => return Ok(false),
         };
         let requested_canonical = std::fs::canonicalize(requested_abs.as_path()).ok();
         for allowed in &guard.config.network.allow_unix_sockets {
+            let allowed_path = Path::new(allowed);
+            if !allowed_path.is_absolute() {
```
bolinfest: Again, let's prefer `AbsolutePathBuf` where possible as it has stronger guarantees/requirements.

codex-rs/core/src/seatbelt.rs
viyatb-oai: fixed!

codex-rs/network-proxy/src/config.rs
viyatb-oai: done

codex-rs/network-proxy/src/runtime.rs
viyatb-oai: codex keeps using this - going to add this to agents.md. fixed

codex-rs/core/src/seatbelt.rs
viyatb-oai: correct it should ignore `allow_unix_sockets`. making it into a union

codex-rs/network-proxy/src/config.rs
viyatb-oai: We still need the /... check because native path APIs are host-platform-specific; Windows rejects POSIX absolute paths. I replaced the ad-hoc helper with ValidatedUnixSocketPath so that distinction is explicit

bolinfest (approved):

PR #11370, opened 2026-02-10
Reapply "Add app-server transport layer with websocket support"

maxj-oai:
Reapply "Add app-server transport layer with websocket support" with additional fixes from https://github.com/openai/codex/pull/11313/changes to avoid deadlocking.

This reverts commit 47356ff83c38305e08c4a075fde4624cb04a7aaf.

## Summary

To avoid deadlocking when queues are full, we maintain separate tokio tasks dedicated to incoming vs outgoing event handling
- split the app-server main loop into two tasks in `run_main_with_transport`
   - inbound handling (`transport_event_rx`)
   - outbound handling (`outgoing_rx` + `thread_created_rx`)
- separate incoming and outgoing websocket tasks

## Validation

Integration tests, testing thoroughly e2e in codex app w/ >10 concurrent requests

Screenshot 2026-02-10 at 2 54 22 PM


codex-rs/app-server/src/outgoing_message.rs
```diff
@@ -150,7 +192,9 @@ impl OutgoingMessageSender {
         }
         if let Err(err) = self
             .sender
-            .send(OutgoingMessage::AppServerNotification(notification))
+            .send(OutgoingEnvelope::Broadcast {
+                message: OutgoingMessage::AppServerNotification(notification),
+            })
```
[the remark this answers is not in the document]
maxj-oai: fixed

codex-rs/app-server/src/transport.rs
```diff
@@ -0,0 +1,511 @@
...
+                    }
+                })
+                .collect();
+
+            for connection_id in target_connections {
+                let Some(connection_state) = connections.get(&connection_id) else {
+                    continue;
+                };
+                if connection_state.writer.send(message.clone()).await.is_err() {
+                    connections.remove(&connection_id);
```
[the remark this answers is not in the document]
maxj-oai: will address this in a future change

codex-rs/app-server/src/lib.rs
```diff
@@ -325,41 +338,136 @@ pub async fn run_main(
...
+                        let Some(envelope) = envelope else {
                             break;
                         };
-                        match msg {
-                            JSONRPCMessage::Request(r) => processor.process_request(r).await,
-                            JSONRPCMessage::Response(r) => processor.process_response(r).await,
-                            JSONRPCMessage::Notification(n) => processor.process_notification(n).await,
-                            JSONRPCMessage::Error(e) => processor.process_error(e).await,
-                        }
+                        let mut connections = connections.lock().await;
```
jif-oai: This means we keep the lock while `route_outgoing_envelope` is being awaited. This means inside it, each `writer.send(...)` is awaited as well.

So if a single connection writer is full, the router completely block while holding the lock... this seems like a bad idea

jif-oai (commented):
This transfer the back pressure to the ingress in practice so it means we rely on the ingress to error/drop events once everything is full.

I would prefer an explicit handling on our side instead but this can be done in a follow-up

codex-rs/app-server/src/lib.rs
maxj-oai: great catch - fixing

maxj-oai: replaced with control channel from https://github.com/openai/codex/pull/11313/changes

jif-oai (approved):
I think we lost `optOutNotificationMethods` (we should have a use of `should_skip_notification_for_connection`). Maybe it's just because we need to merge main

The rest lgtm but I guess we will iterate on this code. It's getting quite complex

PR #11382, opened 2026-02-11
Add configurable MCP OAuth callback URL for MCP login

dkumar-oai:
## Summary

Implements a configurable MCP OAuth callback URL override for `codex mcp login` and app-server OAuth login flows, including support for non-local callback endpoints (for example, devbox ingress URLs).

## What changed

- Added new config key: `mcp_oauth_callback_url` in `~/.codex/config.toml`.
- OAuth authorization now uses `mcp_oauth_callback_url` as `redirect_uri` when set.
- Callback handling validates the callback path against the configured redirect URI path.
- Listener bind behavior is now host-aware:
  - local callback URL hosts (`localhost`, `127.0.0.1`, `::1`) bind to `127.0.0.1`
  - non-local callback URL hosts bind to `0.0.0.0`
- `mcp_oauth_callback_port` remains supported and is used for the listener port.
- Wired through:
  - CLI MCP login flow
  - App-server MCP OAuth login flow
  - Skill dependency OAuth login flow
- Updated config schema and config tests.

## Why

Some environments need OAuth callbacks to land on a specific reachable URL (for example ingress in remote devboxes), not loopback. This change allows that while preserving local defaults for existing users.

## Backward compatibility

- No behavior change when `mcp_oauth_callback_url` is unset.
- Existing `mcp_oauth_callback_port` behavior remains intact.
- Local callback flows continue binding to loopback by default.

## Testing

- `cargo test -p codex-rmcp-client callback -- --nocapture`
- `cargo test -p codex-core --lib mcp_oauth_callback -- --nocapture`
- `cargo check -p codex-cli -p codex-app-server -p codex-rmcp-client`

## Example config

```toml
mcp_oauth_callback_port = 5555
mcp_oauth_callback_url = "https://<devbox>-<namespace>.gateway.<cluster>.internal.api.openai.org/callback"

### DIFF
--- a/codex-rs/rmcp-client/src/perform_oauth_login.rs
+++ b/codex-rs/rmcp-client/src/perform_oauth_login.rs
@@ -114,8 +114,9 @@ fn spawn_callback_server(
             let path = request.url().to_string();
             match parse_oauth_callback(&path, &expected_callback_path) {
                 CallbackOutcome::Success(OauthCallbackResult { code, state }) => {
-                    let response =
-                        Response::from_string("Authentication complete. You may close this window.");
+                    let response = Response::from_string(
+                        "Authentication complete. You may close this window.",
+                    );
                     if let Err(err) = request.respond(response) {
                         eprintln!("Failed to respond to OAuth callback: {err}");
                     }
--- a/codex-rs/rmcp-client/src/perform_oauth_login.rs
+++ b/codex-rs/rmcp-client/src/perform_oauth_login.rs
@@ -8,6 +8,7 @@ use anyhow::Result;
 use anyhow::anyhow;
 use anyhow::bail;
 use reqwest::ClientBuilder;
+use reqwest::Url;
 use rmcp::transport::auth::OAuthState;
 use tiny_http::Response;
 use tiny_http::Server;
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: codex-rs/rmcp-client/src/perform_oauth_login.rs [hunk 1]
```diff
@@ -146,11 +156,11 @@ enum CallbackOutcome {
     Invalid,
 }
 
-fn parse_oauth_callback(path: &str) -> CallbackOutcome {
+fn parse_oauth_callback(path: &str, expected_callback_path: &str) -> CallbackOutcome {
     let Some((route, query)) = path.split_once('?') else {
         return CallbackOutcome::Invalid;
     };
-    if route != "/callback" {
```
shijie-oai, line 153: Can we confirm this? feel like we should not be matching to `expected_callback_path`?

shijie-oai: oh I see what is happening below.

shijie-oai (approved):
LGTM - I would still rec to wait for @celia-oai on reviewing this too.

PR #11383, opened 2026-02-11
Do not resend output items in incremental websockets connections

pakrym-oai:
In the incremental websocket output items are already part of the context, no need to send them again and duplicate.


codex-rs/core/src/client.rs
```diff
@@ -544,50 +553,45 @@ impl ModelClientSession {
...
-        self.refresh_websocket_last_response_id();
-        self.websocket_last_response_id
-            .clone()
-            .filter(|id| !id.is_empty())
+    fn get_last_response(&mut self) -> Option<LastResponse> {
+        self.websocket_last_response_rx
+            .take()
+            .and_then(|mut receiver| match receiver.try_recv() {
+                Ok(last_response) => Some(last_response),
+                Err(TryRecvError::Closed) | Err(TryRecvError::Empty) => None,
```
sayan-oai, line 577: why do we now drop the receiver on `TryRecvError::Empty`? if there's a follow up request before the previous stream finishes, aren't we now losing the `LastResponse`?

pakrym-oai: means previous request didn't complete with `response.done` and if it didn't then trying to append to it is dangerous.

sayan-oai (approved):

PR #11397, opened 2026-02-11
fix(exec-policy) No empty command lists

dylan-hurd-oai:
## Summary
This should rarely, if ever, happen in practice. But regardless, we should never provide an empty list of `commands` to ExecPolicy. This PR is almost entirely adding test around these cases.

## Testing
- [x] Adds a bunch of unit tests for this

### DIFF
--- a/codex-rs/core/src/exec_policy.rs
+++ b/codex-rs/core/src/exec_policy.rs
@@ -345,7 +345,9 @@ fn default_policy_path(codex_home: &Path) -> PathBuf {
 }
 
 fn commands_for_exec_policy(command: &[String]) -> (Vec<Vec<String>>, bool) {
-    if let Some(commands) = parse_shell_lc_plain_commands(command) {
+    if let Some(commands) = parse_shell_lc_plain_commands(command)
+        && !commands.is_empty()
+    {
         return (commands, false);
     }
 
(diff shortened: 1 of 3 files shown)


codex-rs/core/src/exec_policy.rs
```diff
@@ -814,6 +816,24 @@ prefix_rule(pattern=["rm"], decision="forbidden")
         );
     }
 
+    #[test]
+    fn commands_for_exec_policy_falls_back_for_empty_shell_script() {
+        let command = vec!["bash".to_string(), "-lc".to_string(), "".to_string()];
+
+        assert_eq!(commands_for_exec_policy(&command), (vec![command], false),);
```
bolinfest: autoformat leftover?

```suggestion
        assert_eq!(commands_for_exec_policy(&command), (vec![command], false));
```

codex-rs/core/src/exec_policy.rs
```diff
@@ -814,6 +816,24 @@ prefix_rule(pattern=["rm"], decision="forbidden")
...
+
+    #[test]
+    fn commands_for_exec_policy_falls_back_for_whitespace_shell_script() {
+        let command = vec![
+            "bash".to_string(),
+            "-lc".to_string(),
+            "  \n\t  ".to_string(),
+        ];
+
+        assert_eq!(commands_for_exec_policy(&command), (vec![command], false),);
```
bolinfest: here too

bolinfest (approved):

shijie-oai (approved):

codex-rs/core/src/exec_policy.rs
dylan-hurd-oai: fixed

codex-rs/core/src/exec_policy.rs
dylan-hurd-oai: fixed

PR #11403, opened 2026-02-11
Compact before sampling when switching to a smaller context model

aibrahim-oai:
- run pre-sampling compact when switching to a model with a smaller context window after usage crosses the new model limit.
- add e2e coverage for switch and resume+switch flows using 273k -> 125k context window models.

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -4131,9 +4131,10 @@ async fn run_pre_sampling_compact(
             total_usage_tokens,
             &previous_turn_context,
             turn_context.as_ref(),
-        ) {
-            run_auto_compact(sess, &previous_turn_context).await?;
-        }
+        )
+    {
+        run_auto_compact(sess, &previous_turn_context).await?;
+    }
     // Compact if the total usage tokens are greater than the auto compact limit
     if total_usage_tokens >= auto_compact_limit {
         run_auto_compact(sess, turn_context).await?;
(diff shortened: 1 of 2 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -4115,6 +4116,53 @@ async fn run_auto_compact(sess: &Arc<Session>, turn_context: &Arc<TurnContext>)
...
+    let total_usage_tokens = sess.get_total_token_usage().await;
+    let auto_compact_limit = turn_context
+        .model_info
+        .auto_compact_token_limit()
+        .unwrap_or(i64::MAX);
+    // Compact with previous model if the model was switched and previous context window is larger than the new one
+    if let Some(previous_turn_context) = sess.previous_turn_context().await
+        && should_run_inline_compact_with_previous_context(
+            total_usage_tokens,
+            &previous_turn_context,
```
pakrym-oai, line 4132: will this produce evens for the wrong turn it?

codex-rs/core/src/codex.rs
```diff
@@ -3858,15 +3858,16 @@ pub(crate) async fn run_turn(
...
         model_context_window: turn_context.model_context_window(),
         collaboration_mode_kind: turn_context.collaboration_mode.mode,
     });
     sess.send_event(&turn_context, event).await;
-    if total_usage_tokens >= auto_compact_limit
-        && run_auto_compact(&sess, &turn_context).await.is_err()
+
+    if run_pre_sampling_compact(&sess, &turn_context)
+        .await
+        .is_err()
```
pakrym-oai, line 3870: surprised we return None here.

aibrahim-oai: if error, we kill the loop with no final message. This was your request.

pakrym-oai: My request was to kill the loop. Having the message seems fine.

pakrym-oai (approved):

codex-rs/core/src/codex.rs
aibrahim-oai: but there is not message at this point though

PR #11404, opened 2026-02-11
feat: add client-side ModelInfo overrides

sayan-oai:
TLDR: add `model_info_overrides` so users can locally patch model metadata per slug from `config.toml`, including creating new model configs without backend changes.

### Problem
Today Codex only supports a few global top-level model overrides.  
There is no client-side way to patch arbitrary model metadata per slug.

### Fix
- Added `model_info_overrides` map in config (`slug -> patch`).
- Overrides apply to both existing server models and custom/eval slugs.
- Precedence is now:
  1. resolved model metadata (server `/models` or fallback)
  2. `model_info_overrides[slug]`
  3. existing top-level global overrides (`model_context_window`, `model_supports_reasoning_summaries`, etc.)
- Updated docs with precedence + merge semantics + example config usage.

### Tests
Added tests for:
- patching an existing resolved model slug,
- patching an unknown/fallback slug,
- patching a requested slug when resolution uses a different base slug,
- parity between runtime model resolution and offline helper behavior,
- integration check that overrides affect outbound `/responses` request construction.
- drift guard for `ModelInfo` changes without an explicit `ModelInfoPatch` change

### DIFF
--- a/codex-rs/core/src/models_manager/model_info.rs
+++ b/codex-rs/core/src/models_manager/model_info.rs
@@ -59,6 +59,9 @@ pub(crate) fn with_model_info_patch(
     }
     if let Some(base_instructions) = &model_info_patch.base_instructions {
         model.base_instructions = base_instructions.clone();
+        // Keep parity with top-level config behavior: explicit base instructions
+        // should disable template-driven model messages unless the patch sets them again.
+        model.model_messages = None;
     }
     if let Some(model_messages) = &model_info_patch.model_messages {
         model.model_messages = Some(model_messages.clone());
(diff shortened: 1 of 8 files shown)


sayan-oai:
@codex review this

sayan-oai:
@codex review this

codex-rs/core/src/models_manager/model_info.rs
```diff
@@ -19,6 +19,108 @@ const LOCAL_FRIENDLY_TEMPLATE: &str =
...
+    model_key: &str,
+    config: &Config,
+) -> ModelInfo {
+    let Some(model_info_patch) = config.model_info_overrides.get(model_key) else {
+        return model;
+    };
+
+    model.slug = model_key.to_string();
+
+    if let Some(display_name) = &model_info_patch.display_name {
```
sayan-oai, line 33: the chain of `if` blocks is kinda gross, open to ideas here

sayan-oai:
@codex review this

sayan-oai:
@codex review this

PR #11417, opened 2026-02-11
[apps] Add is_enabled to app info.

mzeng-openai:
- [x] Add is_enabled to app info and the response of `app/list`.
- [x] Update TUI to have Enable/Disable button on the app detail page.


owenlin0 (approved):
approving interface change for app-server

codex-rs/app-server-protocol/schema/json/v2/AppListUpdatedNotification.json
```diff
@@ -29,6 +29,11 @@
           "default": false,
           "type": "boolean"
         },
+        "isEnabled": {
+          "default": true,
+          "description": "Whether this app is enabled in the loaded config.",
```
joshka-oai: What does "in the loaded config" mean? Note that this is user facing help when they're editing the toml file and should be phrased that way (it's generated from the comment).

mzeng-openai: Good call out, I added more detailed comments here.

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1288,6 +1288,14 @@ pub struct AppInfo {
     pub install_url: Option<String>,
     #[serde(default)]
     pub is_accessible: bool,
+    /// Whether this app is enabled in config.toml.
+    /// Example:
+    /// ```toml
+    /// [app.bad_app]
```
canvrno-oai:
```suggestion
    /// [apps.bad_app]
```

the smallest of nits

canvrno-oai (approved):

PR #11419, opened 2026-02-11
refactor: codex app-server ThreadState

maxj-oai:
this is a no-op functionality wise. consolidates thread-specific message processor / event handling state in ThreadState

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -254,9 +254,6 @@ use crate::filters::compute_source_filters;
 use crate::filters::source_kind_matches;
 
 type PendingInterruptQueue = Vec<(ConnectionRequestId, ApiVersion)>;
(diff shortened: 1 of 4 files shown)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -265,7 +262,90 @@ pub(crate) struct TurnSummary {
...
+
+    fn clear_listeners(&mut self) {
+        for (_, cancel_tx) in self.listener_cancel_txs.drain() {
+            let _ = cancel_tx.send(());
+        }
+    }
+}
+
+#[derive(Default)]
+struct ThreadStateManager {
```
owenlin0: app-server has a few monolithic files (my bad), but probably worth breaking this  and `ThreadState` out to a new file

owenlin0 (approved):
nice! much cleaner

codex-rs/app-server/src/codex_message_processor.rs
maxj-oai: done!

PR #11474, opened 2026-02-11
app-server: thread resume subscriptions

maxj-oai:
This stack layer makes app-server thread event delivery connection-aware so resumed/attached threads only emit notifications and approval prompts to subscribed connections.

- Added per-thread subscription tracking in `ThreadState` (`subscribed_connections`) and mapped subscription ids to `(thread_id, connection_id)`.
- Updated listener lifecycle so removing a subscription or closing a connection only removes that connection from the thread’s subscriber set; listener shutdown now happens when the last subscriber is gone.
- Added `connection_closed(connection_id)` plumbing (`lib.rs` -> `message_processor.rs` -> `codex_message_processor.rs`) so disconnect cleanup happens immediately.
- Scoped bespoke event handling outputs through `TargetedOutgoing` to send requests/notifications only to subscribed connections.
- Kept existing threadresume behavior while aligning with the latest split-loop transport structure.

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -6479,4 +6479,30 @@ mod tests {
         assert!(!state.lock().await.experimental_raw_events);
         Ok(())
     }
+
+    #[tokio::test]
+    async fn removing_connection_clears_subscription_and_listener_when_last_subscriber()
+    -> Result<()> {
+        let mut manager = ThreadStateManager::new();
+        let thread_id = ThreadId::from_string("ad7f0408-99b8-4f6e-a46f-bd0eec433370")?;
+        let listener = Uuid::new_v4();
+        let connection = ConnectionId(1);
+        let (cancel_tx, cancel_rx) = oneshot::channel();
+
+        manager
+            .set_listener(listener, thread_id, connection, false)
+            .await;
+        {
+            let state = manager.thread_state(thread_id);
+            state.lock().await.cancel_tx = Some(cancel_tx);
+        }
+
+        manager.remove_connection(connection).await;
+        assert_eq!(cancel_rx.await, Ok(()));
+        assert_eq!(manager.remove_listener(listener).await, None);
+
+        let state = manager.thread_state(thread_id);
+        assert!(state.lock().await.subscribed_connection_ids().is_empty());
+        Ok(())
+    }
 }
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -2638,7 +2638,9 @@ impl CodexMessageProcessor {
 
     /// Best-effort: attach a listener for thread_id if missing.
     pub(crate) async fn try_attach_thread_listener(&mut self, thread_id: ThreadId) {
-        if self.thread_state_manager.has_listener_for_thread(thread_id) {}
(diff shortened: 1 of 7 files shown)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -5340,11 +5372,69 @@ impl CodexMessageProcessor {
...
+        thread_state: Arc<Mutex<ThreadState>>,
+        raw_events_enabled: bool,
+        api_version: ApiVersion,
+    ) {
+        let (cancel_tx, mut cancel_rx) = oneshot::channel();
+        {
+            let mut thread_state = thread_state.lock().await;
+            if thread_state.listener_matches(&conversation) {
+                return;
+            }
```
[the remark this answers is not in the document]
owenlin0: edge-casey but can't hurt to fix I guess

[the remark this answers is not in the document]
maxj-oai: addressed by making this last write wins

codex-rs/app-server/src/thread_state.rs
```diff
@@ -75,32 +94,68 @@ impl ThreadStateManager {
     pub(crate) async fn remove_listener(&mut self, subscription_id: Uuid) -> Option<ThreadId> {
         let thread_id = self.thread_id_by_subscription.remove(&subscription_id)?;
         if let Some(thread_state) = self.thread_states.get(&thread_id) {
-            thread_state.lock().await.clear_listener(subscription_id);
+            let has_remaining_subscriptions = self
+                .thread_id_by_subscription
+                .values()
```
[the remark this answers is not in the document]
maxj-oai: fixed

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2793,7 +2804,12 @@ impl CodexMessageProcessor {
                 };
                 // Auto-attach a thread listener when resuming a thread.
                 if let Err(err) = self
-                    .attach_conversation_listener(thread_id, false, ApiVersion::V2)
+                    .ensure_conversation_listener(
```
owenlin0, line 2807: seems a bit confusing to have:
- `try_attach_thread_listener`
- `start_conversation_listener`
- `attach_conversation_listener`
- `ensure_conversation_listener`

can we simplify?

maxj-oai: done!

codex-rs/app-server/src/outgoing_message.rs
```diff
@@ -79,13 +147,34 @@ impl OutgoingMessageSender {
...
+                if let Err(err) = self
+                    .sender
+                    .send(OutgoingEnvelope::ToConnection {
+                        connection_id: *connection_id,
+                        message: outgoing_message.clone(),
+                    })
+                    .await
+                {
+                    send_error = Some(err);
+                    break;
```
owenlin0, line 168: is it a good idea to short-circuit or should we continue sending to the rest of the connections?

owenlin0 (approved):

codex-rs/app-server/src/outgoing_message.rs
maxj-oai: hmm I think stopping is right - this is a single shared queue not per connection so if it's erroring no reason to continue

PR #11484, opened 2026-02-11
feat: split SDKs by target for NPM

jif-oai:
Description

## Why

@openai/codex-sdk publish was failing in release with:

npm ERR! E413 Payload Too Large

The SDK tarball currently bundles binaries for all platforms, which pushes the package over npm’s payload
limits.

## What changed

1. Implemented CLI-style packaging for SDK: one lightweight meta package plus platform-specific native tarballs.
2. Added SDK platform package variants in the staging/build tooling:
  - codex-sdk-linux-x64
  - codex-sdk-linux-arm64
  - codex-sdk-darwin-x64
  - codex-sdk-darwin-arm64
  - codex-sdk-win32-x64
  - codex-sdk-win32-arm64
3. Updated codex-sdk meta package generation:
  - keeps JS/TS output (dist) only
  - adds optionalDependencies pointing to platform-specific SDK variants via npm aliasing.
4. Updated SDK runtime binary resolution to match the new layout:
  - resolve platform package via require.resolve(...)
  - fallback to local vendor/ when present (dev/back-compat path).
5. Updated release pipeline publish logic:
  - download and publish new codex-sdk-npm-<platform>-<version>.tgz artifacts
  - publish them with platform tags (including alpha-prefixed tags for prereleases), matching existing CLI behavior.
6. Updated docs for staging flow and SDK install behavior.

## Expected impact

1. @openai/codex-sdk main tarball is much smaller (no multi-platform vendor payload).
2. npm publish no longer hits payload limits.
3. Consumers install only the binary for their platform.
4. No SDK API surface changes.

## Validation

1. pnpm --filter @openai/codex-sdk run build passed.
2. pnpm --filter @openai/codex-sdk run lint passed.
3. pnpm --filter @openai/codex-sdk exec jest tests/exec.test.ts passed.
4. Python staging scripts compile (py_compile) passed.
5. Full SDK Jest suite could not run in this sandbox due to listen EPERM 127.0.0.1 restrictions.

### DIFF
--- a/codex-cli/scripts/build_npm_package.py
+++ b/codex-cli/scripts/build_npm_package.py
@@ -15,6 +15,7 @@ REPO_ROOT = CODEX_CLI_ROOT.parent
 RESPONSES_API_PROXY_NPM_ROOT = REPO_ROOT / "codex-rs" / "responses-api-proxy" / "npm"
 CODEX_SDK_ROOT = REPO_ROOT / "sdk" / "typescript"
 CODEX_NPM_NAME = "@openai/codex"
+CODEX_SDK_NPM_NAME = "@openai/codex-sdk"
 
 # `npm_name` is the local optional-dependency alias consumed by `bin/codex.js`.
 # The underlying package published to npm is always `@openai/codex`.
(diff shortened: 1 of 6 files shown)


codex-cli/scripts/build_npm_package.py
```diff
@@ -63,8 +64,56 @@
     },
 }
 
+# `npm_name` is the local optional-dependency alias consumed by `sdk/typescript/src/exec.ts`.
+# The underlying package published to npm is always `@openai/codex-sdk`.
+CODEX_SDK_PLATFORM_PACKAGES: dict[str, dict[str, str]] = {
```
bolinfest, line 69: can we share more with the existing map we have for the cli?

bolinfest:
Taking a step back, should https://www.npmjs.com/package/@openai/codex-sdk just depend on https://www.npmjs.com/package/@openai/codex?

If so, the order that we `npm publish` matters...

/cc @pakrym

bolinfest:
Fixed by https://github.com/openai/codex/pull/11503.

PR #11491, opened 2026-02-11
[rmcp-client] Fix OAuth URL handling for linear MCP login

caseychow-oai:
## Summary
- Add OAuth URL candidate fallback for MCP login/discovery when configured URL uses `/mcp`.
- Persist and reuse the effective OAuth base URL (the candidate that succeeds) so token storage key is consistent across reconnect and refresh.
- Load persisted OAuth tokens using the same OAuth URL candidate list.
- Initialize OAuth state/runtime using the effective OAuth URL while keeping MCP transport URL unchanged.

## Testing
- just fmt (in codex-rs)
- cargo test -p codex-rmcp-client (in codex-rs)

### DIFF
--- a/codex-rs/rmcp-client/src/auth_status.rs
+++ b/codex-rs/rmcp-client/src/auth_status.rs
@@ -15,6 +15,7 @@ use crate::OAuthCredentialsStoreMode;
 use crate::oauth::has_oauth_tokens;
 use crate::utils::apply_default_headers;
 use crate::utils::build_default_headers;
+use crate::utils::oauth_auth_url_candidates;
 
 const DISCOVERY_TIMEOUT: Duration = Duration::from_secs(5);
 const OAUTH_DISCOVERY_HEADER: &str = "MCP-Protocol-Version";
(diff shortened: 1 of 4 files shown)


4erdenko:
Hi! Confirming a regression in codex-cli 0.99.0 (macOS): `codex mcp login atlassian` fails immediately with
`Error: No authorization support detected` when configured as a streamable HTTP MCP server:

```toml
[mcp_servers.atlassian]
url = "https://mcp.atlassian.com/v1/mcp"
```

In codex logs I also see:

```
OAuth metadata discovery is unavailable for MCP server `atlassian`; falling back to stored bearer token authentication
```

FYI Atlassian serves RFC8414 OAuth AS metadata at the root (200):
https://mcp.atlassian.com/.well-known/oauth-authorization-server
(and also https://cf.mcp.atlassian.com/.well-known/oauth-authorization-server)

`/.well-known/openid-configuration` is 404, and `.../v1/mcp/.well-known/...` returns 401/404, so discovery code
that derives the metadata URL from the MCP base path (or only tries OIDC) will fail.

Workaround: run Atlassian via `mcp-remote` (stdio) which completes OAuth in browser and caches tokens:

```bash
npx -y mcp-remote@latest https://mcp.atlassian.com/v1/mcp --transport http-only
```

caseychow-oai:
Closing for https://github.com/openai/codex/pull/11539

PR #11497, opened 2026-02-11
Hydrate previous model across resume/fork/rollback/task start

aibrahim-oai:
- Replace pending resume model state with persistent previous_model and hydrate it on resume, fork, rollback, and task end in spawn_task


codex-rs/core/src/tasks/mod.rs
```diff
@@ -267,6 +273,10 @@ impl Session {
 
         task.handle.abort();
 
+        // Set previous model even when interrupted so model-switch handling stays correct.
+        self.set_previous_model(Some(task.turn_context.model_info.slug.clone()))
```
pakrym-oai, line 277: Can this be set in regular task instead?

pakrym-oai: We shouldn't pick up models from all tasks I think.

pakrym-oai (approved):

charley-oai (approved):

PR #11500, opened 2026-02-11
feat: build windows support binaries in parallel

bolinfest:
Windows release builds were compiling and linking four release binaries on a single runner, which slowed the release pipeline. The Windows-specific logic also made `rust-release.yml` harder to read and maintain.

## What Changed

- Extracted Windows release logic into a reusable workflow at `.github/workflows/rust-release-windows.yml`.
- Updated `.github/workflows/rust-release.yml` to call the reusable Windows workflow via `workflow_call`.
- Parallelized Windows binary builds with one 4-entry matrix over two targets (`x86_64-pc-windows-msvc`, `aarch64-pc-windows-msvc`) and two bundles (`primary`, `helpers`).
- Kept signing centralized per target by downloading both prebuilt bundles and signing all four executables together.
- Preserved final release artifact behavior and filtered intermediate `windows-binaries*` artifacts out of the published release asset set.

### DIFF
--- a/.github/workflows/rust-release-windows.yml
+++ b/.github/workflows/rust-release-windows.yml
@@ -1,0 +1,269 @@
+name: rust-release-windows
+
+on:
+  workflow_call:
+    inputs:
+      release-lto:
+        required: true
+        type: string
+    secrets:
+      AZURE_TRUSTED_SIGNING_CLIENT_ID:
+        required: true
+      AZURE_TRUSTED_SIGNING_TENANT_ID:
+        required: true
+      AZURE_TRUSTED_SIGNING_SUBSCRIPTION_ID:
+        required: true
+      AZURE_TRUSTED_SIGNING_ENDPOINT:
+        required: true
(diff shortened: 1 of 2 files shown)


.github/workflows/rust-release-windows.yml
```diff
@@ -0,0 +1,269 @@
...
+          targets: ${{ matrix.target }}
+
+      - uses: actions/cache@v5
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            ${{ github.workspace }}/codex-rs/target/
```
joshka-oai, line 96: Are these paths right for windows?

joshka-oai (approved):

.github/workflows/rust-release-windows.yml
bolinfest: Codex says https://github.com/actions/cache honors `~/` as `C:\Users\runneradmin` or whatever on Windows.

PR #11503, opened 2026-02-11
Make codex-sdk depend on openai/codex

pakrym-oai:
Do not bundle all binaries inside the SDK as it makes the package huge. Instead depend on openai/codex

### DIFF
--- a/sdk/typescript/src/exec.ts
+++ b/sdk/typescript/src/exec.ts
@@ -1,7 +1,7 @@
 import { spawn } from "node:child_process";
 import path from "node:path";
 import readline from "node:readline";
-import { fileURLToPath } from "node:url";
+import { createRequire } from "node:module";
 
 import type { CodexConfigObject, CodexConfigValue } from "./codexOptions";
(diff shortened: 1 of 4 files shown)


sdk/typescript/src/exec.ts
```diff
@@ -351,7 +360,23 @@ function findCodexPath() {
...
+  const platformPackage = PLATFORM_PACKAGE_BY_TARGET[targetTriple];
+  if (!platformPackage) {
+    throw new Error(`Unsupported target triple: ${targetTriple}`);
+  }
+
+  let vendorRoot: string;
+  try {
+    const codexPackageJsonPath = moduleRequire.resolve(`${CODEX_NPM_NAME}/package.json`);
+    const codexRequire = createRequire(codexPackageJsonPath);
+    const platformPackageJsonPath = codexRequire.resolve(`${platformPackage}/package.json`);
```
bolinfest, line 372: neat?

bolinfest (commented):
Are there any changes required to the GitHub workflows? For example, I assume we need to be extra sure that `@openai/codex` is published before `@openai/codex-sdk`?

Is there any code that can be deleted from the release workflow?

bolinfest (approved):

PR #11510, opened 2026-02-12
Remove git commands from dangerous command checks

joshka-oai:
### Motivation

- Git subcommand matching was being classified as "dangerous" and caused benign developer workflows (for example `git push --force-with-lease`) to be blocked by the preflight policy. 
- The change aligns behavior with the intent to reserve the dangerous checklist for truly destructive shell ops (e.g. `rm -rf`) and avoid surprising developer-facing blocks.

### Description

- Remove git-specific subcommand checks from `is_dangerous_to_call_with_exec` in `codex-rs/shell-command/src/command_safety/is_dangerous_command.rs`, leaving only explicit `rm` and `sudo` passthrough checks. 
- Deleted the git-specific helper logic that classified `reset`, `branch`-delete, `push` (force/delete/refspec) and `clean --force` as dangerous. 
- Updated unit tests in the same file to assert that various `git reset`/`git branch`/`git push`/`git clean` variants are no longer classified as dangerous. 
- Kept `find_git_subcommand` (used by safe-command classification) intact so safe/unsafe parsing elsewhere remains functional.

### Testing

- Ran formatter with `just fmt` successfully.  
- Ran unit tests with `cargo test -p codex-shell-command` and all tests passed (`144 passed; 0 failed`).

------
[Codex Task](https://chatgpt.com/codex/tasks/task_i_698d19dedb4883299c3ceb5bbc6a0dcf)


codex-rs/shell-command/src/command_safety/is_dangerous_command.rs
```diff
@@ -184,22 +123,22 @@ mod tests {
     }
 
     #[test]
-    fn git_reset_is_dangerous() {
```
bolinfest, line 187: I think I would just delete these tests. Someone looking at this file with fresh eyes would have no context on why we have tests for `git` but nothing else?

bolinfest (approved):

codex-rs/shell-command/src/command_safety/is_dangerous_command.rs
joshka-oai: @codex remove these tests

bolinfest:
@joshka-oai It looks like you need to update this test to match something in the dangerous list:

https://github.com/openai/codex/blob/fd1efb86dff9e0edcad977f14e7cbe52abda097d/codex-rs/core/src/exec_policy.rs#L1387-L1408

joshka-oai:
@codex there are still tests that are git related here - remove them all (e.g. git_reset_is_not_dangerous, zsh_git_reset_is_not_dangerous, ...). We no longer want any tests that test for git dangerous commands here as this behavior is removed and so no longer needs tests.

Make sure to run the full test suite including dangerous_git_push_requires_approval_in_danger_full_access in codex/codex-rs/core/src/exec_policy.rs which needs to be updated to choose a different dangerous command to run that will fail.

PR #11511, opened 2026-02-12
fix: fix bug in resource parsing logic

bolinfest:


bolinfest:
this was on an old main; redoing this

bolinfest:
see https://github.com/openai/codex/pull/11513

PR #11512, opened 2026-02-12
add a slash command to grant sandbox read access to inaccessible directories

iceweasel-oai:
There is an edge case where a directory is not readable by the sandbox. In practice, we've seen very little of it, but it can happen so this slash command unlocks users when it does.

Future idea is to make this a tool that the agent knows about so it can be more integrated.

### DIFF
--- a/codex-rs/tui/src/slash_command.rs
+++ b/codex-rs/tui/src/slash_command.rs
@@ -17,7 +17,7 @@ pub enum SlashCommand {
     Permissions,
     #[strum(serialize = "setup-default-sandbox")]
     ElevateSandbox,
-    #[strum(serialize = "sandbox-read-root")]
+    #[strum(serialize = "sandbox-add-read-dir")]
     SandboxReadRoot,
     Experimental,
     Skills,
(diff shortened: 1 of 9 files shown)


iceweasel-oai:
@codex review

codex-rs/tui/src/slash_command.rs
```diff
@@ -17,6 +17,8 @@ pub enum SlashCommand {
     Permissions,
     #[strum(serialize = "setup-default-sandbox")]
     ElevateSandbox,
+    #[strum(serialize = "sandbox-read-root")]
```
dylan-hurd-oai: can we name this something more intuitive? e.g. `/sandbox-add-dir`

iceweasel-oai: yeah, good call. I think `/sandbox-add-read-dir` just so it doesn't look like this is adding a new *writable* root

codex-rs/tui/src/slash_command.rs
```diff
@@ -77,7 +79,8 @@ impl SlashCommand {
             SlashCommand::Agent => "switch the active agent thread",
             SlashCommand::Approvals => "choose what Codex is allowed to do",
             SlashCommand::Permissions => "choose what Codex is allowed to do",
-            SlashCommand::ElevateSandbox => "set up default agent sandbox",
+            SlashCommand::ElevateSandbox => "set up elevated agent sandbox",
+            SlashCommand::SandboxReadRoot => "grant sandbox read access: /sandbox-read-root <path>",
```
dylan-hurd-oai: Would be great to make the copy a tad clearer here: maybe `add sandbox directory access: /sandbox-read-root <absolute_path>`

Also - I think this is our only slash command to accept an argument. Are we sure we want to start doing this? Could we do this via a popup instead?

iceweasel-oai: we do have other ones that accept arguments but those arguments aren't required. 
Let's try this out and I'm open to improving the UX if needed. Just want to make sure we have *something* we can point folks to, if they get into this situation.

I'll update the copy

dylan-hurd-oai (approved):

PR #11517, opened 2026-02-12
protocol: refresh token usage context window after model switch

charley-oai:
## Summary
Fix stale `model_context_window` in token usage telemetry after model switches.

## Bug
`TokenUsageInfo::new_or_append` cloned existing `TokenUsageInfo` and only set
`model_context_window` during initialization.
After switching models (for example from a 272k model to a 128k model), token usage
events could continue reporting the old context window.

## Root cause
In `TokenUsageInfo::new_or_append`, when `info` already existed, the function preserved
the previous `model_context_window` unconditionally, even when a new window value was
provided for the current turn/model.

## Fix
Update `TokenUsageInfo::new_or_append` so that:
- when `model_context_window` is `Some(...)`, it overwrites `info.model_context_window`
- when `model_context_window` is `None`, existing behavior is preserved

### DIFF
--- a/codex-rs/protocol/src/protocol.rs
+++ b/codex-rs/protocol/src/protocol.rs
@@ -1203,6 +1203,9 @@ impl TokenUsageInfo {
                 model_context_window,
             },
         };
+        if let Some(model_context_window) = model_context_window {
+            info.model_context_window = Some(model_context_window);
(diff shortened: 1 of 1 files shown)


charley-oai:
@codex review this

PR #11519, opened 2026-02-12
Use slug in tui

pakrym-oai:
Display name is for VSCE and App, TUI uses lowercase everywhere.


aibrahim-oai (approved):

joshka-oai (approved):
Test please? Ideally Visual (insta snap)

PR #11547, opened 2026-02-12
Parse first order skill/connector mentions

canvrno-oai:
This PR introduces a skill-expansion mechanism for mentions so nested or skill or connection mentions are expanded if present in skills invoked by the user. This keeps behavior aligned with existing mention handling while extending coverage to deeper scenarios. With these changes, users can create skills that invoke connectors, and skills that invoke other skills.

Replaces #10863, which is not needed with the addition of [search_tool_bm25](https://github.com/openai/codex/issues/10657)


codex-rs/core/src/codex.rs
```diff
@@ -4051,6 +4076,15 @@ pub(crate) async fn run_turn(
             .await;
     }
 
+    let mut explicitly_enabled_connectors = collect_explicit_app_ids(&input);
+    explicitly_enabled_connectors.extend(collect_explicit_app_ids_from_skill_items(
+        &skill_items,
```
mzeng-openai, line 4081: Are these top level skills only? Just want to make sure we are not adding all connector mentions recursively since that will be against the progressive revealing design.

mzeng-openai (approved):

PR #11549, opened 2026-02-12
feat(tui) /permissions custom option

dylan-hurd-oai:
## Summary
When a user has a custom permissions config set, we should show that in the `/permissions` list

## Testing
- [x] Added unit tests
- [x] tested locally

Screenshot 2026-02-11 at 10 18 08 PM


codex-rs/tui/src/chatwidget/tests.rs
```diff
@@ -105,15 +105,24 @@ use tokio::sync::mpsc::unbounded_channel;
 use toml::Value as TomlValue;
 
 async fn test_config() -> Config {
-    // Use base defaults to avoid depending on host state.
-    let codex_home = std::env::temp_dir();
+    // Use an isolated codex_home to avoid depending on host state.
+    let codex_home = tempdir().expect("tempdir").keep();
```
joshka-oai, line 111: probably need to pair this with a `remove()` call later?

codex-rs/tui/src/chatwidget.rs
```diff
@@ -5423,6 +5443,49 @@ impl ChatWidget {
         current_approval == preset.approval && *current_sandbox == preset.sandbox
     }
 
+    fn custom_permissions_from_user_config(
+        config: &Config,
+    ) -> Option<(AskForApproval, SandboxPolicy)> {
+        let user_layer = config.config_layer_stack.get_user_layer()?;
+        let config_toml: ConfigToml = user_layer.config.clone().try_into().ok()?;
+        let profile = config_toml
+            .get_config_profile(config.active_profile.clone())
```
iceweasel-oai, line 5494: looks like there could be a bug here where `config.active_profile` does not exist in `config_toml` which is just the user layer.  Seems like a non-blocker

iceweasel-oai (approved):
one potential bug to look at, non-blocking

codex-rs/tui/src/chatwidget.rs
```diff
@@ -5443,6 +5392,31 @@ impl ChatWidget {
             });
         }
 
+        let custom_config_permissions = Self::custom_permissions_from_user_config(&self.config);
```
joshka-oai, line 5395: A comment here would be very appreciated (or perhaps a small method?)

codex-rs/tui/src/chatwidget/tests.rs
```diff
@@ -4983,6 +4180,69 @@ async fn approvals_selection_popup_snapshot() {
     assert_snapshot!("approvals_selection_popup", popup);
 }
 
+#[tokio::test]
+async fn approvals_popup_shows_custom_option_for_custom_user_config() {
```
joshka-oai, line 4184: Ideally these tests would be simple visual insta snap tests - showing the output of what the menu looks like when rendered makes this super easy to see how this changes when / if we change this behavior in the future. I'm not sure if there are existing tests for this, but I think it makes sense to add them if not.

joshka-oai (approved):
Approved to unblock this, with some suggestions that should be done before merging.

PR #11550, opened 2026-02-12
feat(tui) Permissions update history item

dylan-hurd-oai:
## Summary
We should document in the tui when you switch permissions!

## Testing
- [x] Added unit tests
- [x] Tested locally

### DIFF
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -5318,20 +5318,7 @@ impl ChatWidget {
                     .notices
                     .hide_full_access_warning
                     .unwrap_or(false);
-            let mut actions: Vec<SelectionAction> = Vec::new();
-            if !(is_current || preset.id == "full-access" && requires_confirmation) {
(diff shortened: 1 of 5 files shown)


codex-rs/tui/src/chatwidget.rs
```diff
@@ -5288,7 +5288,20 @@ impl ChatWidget {
                     .notices
                     .hide_full_access_warning
                     .unwrap_or(false);
-            let actions: Vec<SelectionAction> = if requires_confirmation {
+            let mut actions: Vec<SelectionAction> = Vec::new();
+            if !(is_current || preset.id == "full-access" && requires_confirmation) {
+                let selected_name = name.clone();
+                actions.push(Box::new(move |tx| {
```
iceweasel-oai: on Windows - we sometimes initiated the Sandbox NUX when `workspace-write` is chosen. That should be way less common now that we have promoted the NUX, but I think it could still happen. Might be worth checking that this runs after the NUX completes

joshka-oai (approved):

PR #11557, opened 2026-02-12
fix: stop inheriting rate-limit limit_name

xl-openai:
When we carry over values from partial rate-limit, we should skip limit_name


pakrym-oai (approved):

xl-openai:
@codex review again

PR #11560, opened 2026-02-12
tui: highlight selected row in resume picker

bmizerany:
The picker currently only marks the active session with a ">" prefix,
which is easy to miss when scanning dense rows. Add a subtle background
to the selected row using user_message_style so the highlight follows the
existing terminal-aware color scheme.


etraut-openai:

PR #11569, opened 2026-02-12
feat: mem slash commands

jif-oai:
Add 2 slash commands for memories:
* `/m_drop` delete all the memories
* `/m_update` update the memories with phase 1 and 2


gt-oai (approved):
If these are for debugging prefix them with `debug-` e.g. `/debug-memories-drop`. Also would recommend `/debug-memories-drop` v.s. `/debug-m-drop`

jif-oai:
> /debug-memories-drop

I don't want people to see the word "memories" and become crazy on it

PR #11575, opened 2026-02-12
fix: db stuff mem

jif-oai:
* Documenting DB functions
* Fixing 1 nit where stage-2 was sorting the stage 1 in the wrong direction
* Added some tests


jif-oai:
@codex review

PR #11593, opened 2026-02-12
feat: metrics to memories

jif-oai:


codex-rs/core/src/memories/startup/mod.rs
```diff
@@ -79,8 +87,19 @@ pub(super) async fn run_memories_startup_pipeline(
...
+        otel_manager.counter(
+            metrics::MEMORY_PHASE_ONE_JOBS,
+            1,
+            &[("status", "skipped_state_db_unavailable")],
+        );
+        otel_manager.counter(
+            metrics::MEMORY_PHASE_ONE_JOBS,
+            1,
+            &[("status", "skipped_state_db_unavailable")],
+        );
```
gt-oai, line 102: dupe?

gt-oai (approved):

codex-rs/core/src/memories/startup/mod.rs
jif-oai: Good catch

jif-oai: I already have a follow-up where it's dropped. Dropping it therwe

PR #11615, opened 2026-02-12
Allow hooks to error

gt-oai:
Allow hooks to return errors. 

We should do this before introducing more hook types, or we'll have to migrate them all.


jif-oai:
@codex review

codex-rs/core/src/codex.rs
```diff
@@ -4247,6 +4249,46 @@ pub(crate) async fn run_turn(
...
+                                    "after_agent hook '{hook_name}' failed and aborted turn completion: {error}"
+                                );
+                                warn!(
+                                    turn_id = %turn_context.sub_id,
+                                    hook_name = %hook_name,
+                                    error = %error,
+                                    "after_agent hook failed; aborting operation"
+                                );
+                                abort_message = Some(message);
+                                break;
```
jif-oai: The break is supposed to be useless as the `FailedAbort` abort the other hooks. So either we take this assumption and we don't need the break or we don't take it and I would avoid putting the break to make sure we also log the other potential failures

codex-rs/hooks/src/registry.rs
```diff
@@ -45,14 +45,18 @@ impl Hooks {
         }
     }
 
-    pub async fn dispatch(&self, hook_payload: HookPayload) {
-        // TODO(gt): support interrupting program execution by returning a result here.
+    pub async fn dispatch(&self, hook_payload: HookPayload) -> Vec<HookResponse> {
+        let mut outcomes = Vec::new();
```
jif-oai: nit: allocate with capacity

jif-oai (approved):
I'm curious to see what it will look like from an user's point of view (how can they define hooks and parametrise this)

PR #11631, opened 2026-02-12
chore(core) Deprecate approval_policy: on-failure

dylan-hurd-oai:
## Summary
In an effort to start simplifying our sandbox setup, we're announcing this approval_policy as deprecated. In general, it performs worse than `on-request`, and we're focusing on making fewer sandbox configurations perform much better.

## Testing
- [x] Tested locally
- [x] Existing tests pass


etraut-openai (approved):
Thanks! Looks good.

PR #11638, opened 2026-02-12
fix(app-server): surface more helpful errors for json-rpc

owenlin0:
Propagate client JSON-RPC errors for app-server request callbacks. Previously a number of possible errors were collapsed to `channel closed`. Now we should be able to see the underlying client error.

### Summary
This change stops masking client JSON-RPC error responses as generic callback cancellation in app-server server->client request flows.

Previously, when the client responded with a JSON-RPC error, we removed the callback entry but did not send anything to the waiting oneshot receiver. Waiters then observed channel closure (for example, auth refresh request canceled: channel closed), which hid the actual client error.

Now, client JSON-RPC errors are forwarded through the callback channel and handled explicitly by request consumers.

### User-visible behavior
- External auth refresh now surfaces real client JSON-RPC errors when provided.
- True transport/callback-drop cases still report canceled/channel-closed semantics.

### Example: client JSON-RPC error is now propagated (not masked as "canceled")

When app-server asks the client to refresh ChatGPT auth tokens, it sends a server->client JSON-RPC request like:

```json
{
  "id": 42,
  "method": "account/chatgptAuthTokens/refresh",
  "params": {
    "reason": "unauthorized",
    "previousAccountId": "org-abc"
  }
}
```

If the client cannot refresh and responds with a JSON-RPC error:
```
{
  "id": 42,
  "error": {
    "code": -32000,
    "message": "refresh failed",
    "data": null
  }
}
```

app-server now forwards that error through the callback path and surfaces:
`auth refresh request failed: code=-32000 message=refresh failed`

Previously, this same case could be reported as:
`auth refresh request canceled: channel closed`

### DIFF
--- a/codex-rs/app-server/src/bespoke_event_handling.rs
+++ b/codex-rs/app-server/src/bespoke_event_handling.rs
@@ -4,6 +4,7 @@ use crate::codex_message_processor::read_summary_from_rollout;
 use crate::codex_message_processor::summary_to_thread;
 use crate::error_code::INTERNAL_ERROR_CODE;
 use crate::error_code::INVALID_REQUEST_ERROR_CODE;
+use crate::outgoing_message::ClientRequestResult;
 use crate::outgoing_message::ThreadScopedOutgoingMessageSender;
 use crate::thread_state::ThreadState;
 use crate::thread_state::TurnSummary;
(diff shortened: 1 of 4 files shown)


celia-oai (approved):
what makes some request have a fallback response when there's client error, and some don't have this?

owenlin0:
@celia-oai For some codepaths we can default to some reasonable behavior for a client error (like a dynamic tool call being sent as `"tool failed"` back to the model), whereas other codepaths there isn't a reasonable default (the external auth refresh case).

bolinfest (approved):

PR #11639, opened 2026-02-12
[feat] add seatbelt permission files

celia-oai:
Added a structured extension layer for macOS Seatbelt permissions. When used, it adds custom seatbelt permission rules on top of `seatbelt_base_policy.sbpl`.

This introduces macOS extension settings (macos_preferences, macos_automation, macos_accessibility, macos_calendar) and uses them to generate policy  clauses at runtime.

This does modify core/src/seatbelt_base_policy.sbpl, but it preserves existing behavior: the previous default prefs-read grant (user-preference-read) was moved from the base policy into the extension builder and is still applied automatically when no extension profile is provided.


codex-rs/core/src/seatbelt.rs
```diff
@@ -275,15 +296,33 @@ pub(crate) fn create_seatbelt_command_args(
...
     let proxy = proxy_policy_inputs(network);
     let network_policy = dynamic_network_policy(sandbox_policy, enforce_managed_network, &proxy);
-
-    let full_policy = format!(
-        "{MACOS_SEATBELT_BASE_POLICY}\n{file_read_policy}\n{file_write_policy}\n{network_policy}"
+    let seatbelt_extensions = extensions.map_or_else(
+        || {
+            // Backward-compatibility default when no extension profile is provided.
+            build_seatbelt_extensions(&MacOsSeatbeltProfileExtensions {
+                macos_preferences: MacOsPreferencesPermission::ReadOnly,
```
leoshimo-oai, line 303: Should readonly macOS preferences be enabled in default sandbox w/o extensions?

leoshimo-oai: ic - in base profile before

leoshimo-oai (approved):
lgtm

codex-rs/core/src/seatbelt.rs
celia-oai: yup, I can remove if you don't care about this default behavior. cc: @bolinfest

PR #11658, opened 2026-02-12
feat(skills): add permission profiles from openai.yaml metadata

celia-oai:
## Summary

  This PR adds support for skill-level permissions in .codex/openai.yaml and wires that through the skill loading pipeline.

  ## What’s included

  1. Added a new permissions section for skills (network, filesystem, and macOS-related access).
  2. Implemented permission parsing/normalization and translation into runtime permission profiles.
  3. Threaded the new permission profile through SkillMetadata and loader flow.

  ## Follow-up

  A follow-up PR will connect these permission profiles to actual sandbox enforcement and add user approval prompts for executing binaries/scripts from skill directories.

 ## Example 
`openai.yaml` snippet:
```
  permissions:
    network: true
    fs_read:
      - "./data"
      - "./data"
    fs_write:
      - "./output"
    macos_preferences: "readwrite"
    macos_automation:
      - "com.apple.Notes"
    macos_accessibility: true
    macos_calendar: true
```

compiled skill permission profile metadata (macOS): 
```
SkillPermissionProfile {
      sandbox_policy: SandboxPolicy::WorkspaceWrite {
          writable_roots: vec![
              AbsolutePathBuf::try_from("/ABS/PATH/TO/SKILL/output").unwrap(),
          ],
          read_only_access: ReadOnlyAccess::Restricted {
              include_platform_defaults: true,
              readable_roots: vec![
                  AbsolutePathBuf::try_from("/ABS/PATH/TO/SKILL/data").unwrap(),
              ],
          },
          network_access: true,
          exclude_tmpdir_env_var: false,
          exclude_slash_tmp: false,
      },
      // Truncated for readability; actual generated profile is longer.
      macos_seatbelt_permission_file: r#"
  (allow user-preference-write)
  (allow appleevent-send
      (appleevent-destination "com.apple.Notes"))
  (allow mach-lookup (global-name "com.apple.axserver"))
  (allow mach-lookup (global-name "com.apple.CalendarAgent"))
  ...
  "#.to_string(),
```


codex-rs/core/src/skills/assets/samples/skill-creator/references/openai_yaml.md
```diff
@@ -23,6 +23,19 @@ dependencies:
 
 policy:
   allow_implicit_invocation: true
+
+permissions:
```
celia-oai: should we not add this to readme?

codex-rs/core/src/skills/permissions.rs
```diff
@@ -59,27 +56,46 @@ pub(crate) fn compile_permission_profile(
...
+        }
+    } else if !fs_read.is_empty() {
+        SandboxPolicy::ReadOnly {
+            access: ReadOnlyAccess::Restricted {
+                include_platform_defaults: true,
+                readable_roots: fs_read,
+            },
+        }
+    } else {
+        SandboxPolicy::new_read_only_policy()
```
celia-oai, line 106: is this the correct default?

codex-rs/core/src/skills/permissions.rs
```diff
@@ -0,0 +1,357 @@
...
+    permissions: Option<SkillManifestPermissions>,
+) -> Option<SkillPermissionProfile> {
+    let permissions = permissions?;
+    let fs_read =
+        normalize_permission_paths(skill_dir, &permissions.fs_read, "permissions.fs_read");
+    let fs_write =
+        normalize_permission_paths(skill_dir, &permissions.fs_write, "permissions.fs_write");
+    let has_sandbox_permissions =
+        permissions.network || !fs_read.is_empty() || !fs_write.is_empty();
+    let sandbox_policy = if permissions.network || !fs_write.is_empty() {
```
viyatb-oai: here `network: true / fs_write` overgrants filesystem access in the compiled profile  which will make the permissions broad.

`network || fs_write` is mapped to `WorkspaceWrite`, and on line 71 (below), we give `ReadOnlyAccess::FullAccess` when `fs_read` is empty. Combined with default writable roots in protocol.rs, a “network-only” request effectively implies write access to cwd//tmp and full-disk read.

codex-rs/core/src/skills/permissions.rs
```diff
@@ -0,0 +1,357 @@
...
+#[serde(untagged)]
+pub(crate) enum MacOsAutomationValue {
+    Bool(bool),
+    BundleIds(Vec<String>),
+}
+
+pub(crate) fn compile_permission_profile(
+    skill_dir: &Path,
+    permissions: Option<SkillManifestPermissions>,
+) -> Option<SkillPermissionProfile> {
```
owenlin0: Should this just return `Permission`? @bolinfest added that struct in core/src/config/mod.rs

codex-rs/core/src/skills/assets/samples/skill-creator/references/openai_yaml.md
owenlin0: i'm not sure. this is experimental and subject to change. @xl-openai thoughts?

codex-rs/core/src/skills/permissions.rs
celia-oai: good call! I'll add macos_seatbelt_profile_extensions in `Permission`

codex-rs/core/src/skills/permissions.rs
celia-oai: @viyatb-oai I made the change just now, but seems that we don't currently allow `network: true` + readonly? would this confuse the users cause they would think these are two separate knobs? code I'm looking at: https://github.com/openai/codex/blob/dev/cc/skill-permission/codex-rs/protocol/src/protocol.rs#L638-L645. is this something I should fix by adding `network_access` as a field to Sandbox_policy: readonly?

codex-rs/core/src/skills/model.rs
```diff
@@ -3,14 +3,17 @@ use std::path::PathBuf;
...
+
+#[derive(Debug, Clone, PartialEq)]
 pub struct SkillMetadata {
     pub name: String,
     pub description: String,
     pub short_description: Option<String>,
     pub interface: Option<SkillInterface>,
     pub dependencies: Option<SkillDependencies>,
     pub policy: Option<SkillPolicy>,
+    pub permission_profile: Option<Permissions>,
```
xl-openai: Add a comment indicating this is experimental.

codex-rs/core/src/skills/permissions.rs
viyatb-oai: no, that'd change assumptions that readonly mode does not allow network access.
the only fix here is to not map network: true to WorkspaceWrite when fs_write is empty.

celia-oai: sg

codex-rs/core/src/skills/model.rs
```diff
@@ -3,14 +3,18 @@ use std::path::PathBuf;
...
+#[derive(Debug, Clone, PartialEq)]
 pub struct SkillMetadata {
     pub name: String,
     pub description: String,
     pub short_description: Option<String>,
     pub interface: Option<SkillInterface>,
     pub dependencies: Option<SkillDependencies>,
     pub policy: Option<SkillPolicy>,
+    // This is an experimental field.
+    pub permission_profile: Option<Permissions>,
```
owenlin0: nit: should we call this field `permissions`?

owenlin0 (approved):

codex-rs/core/src/skills/model.rs
celia-oai: sg, will rename

viyatb-oai (approved):
lets add a test to cover network-only and network + read-only behavior for regression

PR #11664, opened 2026-02-13
Context update marker

charley-oai:
## Summary

This PR introduces a small, low-risk infrastructure step toward the context-update design by unifying contextual user-message
classification and adding support for the new `<context_update>` marker.

### What it does

- Adds `<context_update>` to session-prefix detection.
- Introduces a single shared helper for contextual user-message detection:
  - `is_contextual_user_message(content: &[ContentItem])`
- Refactors both callsites to use the shared helper:
  - `event_mapping::parse_user_message`
  - `context_manager::is_user_turn_boundary`
- Extends tests to cover `<context_update>` across parsing and turn-boundary behavior.

### What it does **not** do

- Does **not** change context message emission yet.
- Does **not** introduce envelope assembly/state-diff plumbing yet.
- Keeps existing legacy contextual wrappers fully recognized.

## Rationale

Before this change, contextual-user detection logic was duplicated in two places.
That made it easy for parsing and user-turn boundary rules to drift.

This PR centralizes that logic and adds marker recognition now, so future PRs that emit `<context_update>` can rely on a
single classification contract.

### DIFF
--- a/codex-rs/core/src/context_manager/history.rs
+++ b/codex-rs/core/src/context_manager/history.rs
@@ -1,16 +1,12 @@
 use crate::codex::TurnContext;
 use crate::context_manager::normalize;
-use crate::instructions::SkillInstructions;
-use crate::instructions::UserInstructions;
-use crate::session_prefix::is_session_prefix;
+use crate::session_prefix::is_contextual_user_message;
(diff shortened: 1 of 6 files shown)


charley-oai:
@codex review this

PR #11668, opened 2026-02-13
feat(shell-tool-mcp): add patched zsh build pipeline

nornagon-openai:
## Summary
- add `shell-tool-mcp/patches/zsh-exec-wrapper.patch` against upstream zsh `77045ef899e53b9598bebc5a41db93a548a40ca6`
- add `zsh-linux` and `zsh-darwin` jobs to `.github/workflows/shell-tool-mcp.yml`
- stage zsh binaries under `artifacts/vendor/<target>/zsh/<variant>/zsh`
- include zsh artifact jobs in `package.needs`
- mark staged zsh binaries executable during packaging

## Notes
- zsh source is cloned from `https://git.code.sf.net/p/zsh/code`
- workflow pins zsh commit `77045ef899e53b9598bebc5a41db93a548a40ca6`
- zsh build runs `./Util/preconfig` before `./configure`

## Validation
- parsed workflow YAML locally (`yaml-ok`)
- validated zsh patch applies cleanly with `git apply --check` on a fresh zsh clone

### DIFF
--- a/shell-tool-mcp/patches/zsh-exec-wrapper.patch
+++ b/shell-tool-mcp/patches/zsh-exec-wrapper.patch
@@ -18,7 +18,7 @@ index 27bca11..baea760 100644
      if (newenvp == NULL)
  	    newenvp = environ;
 +    exec_argv = argv;
-+    if ((exec_wrapper = getenv("BASH_EXEC_WRAPPER")) &&
++    if ((exec_wrapper = getenv("EXEC_WRAPPER")) &&
 +	*exec_wrapper && !inblank(*exec_wrapper)) {
 +	exec_argv = argv - 2;
 +	exec_argv[0] = exec_wrapper;
--- a/shell-tool-mcp/patches/zsh-exec-wrapper.patch
+++ b/shell-tool-mcp/patches/zsh-exec-wrapper.patch
@@ -1,0 +1,34 @@
+diff --git a/Src/exec.c b/Src/exec.c
+index 27bca11..baea760 100644
+--- a/Src/exec.c
++++ b/Src/exec.c
+@@ -507,7 +507,9 @@ zexecve(char *pth, char **argv, char **newenvp)
+ {
+     int eno;
+     static char buf[PATH_MAX * 2+1];
+-    char **eep;
++    char **eep, **exec_argv;
++    char *orig_pth = pth;
++    char *exec_wrapper;
+ 
+     unmetafy(pth, NULL);
+     for (eep = argv; *eep; eep++)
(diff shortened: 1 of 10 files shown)


bolinfest (approved):
let's give it a shot!

shell-tool-mcp/patches/zsh-exec-wrapper.patch
```diff
@@ -0,0 +1,34 @@
...
+@@ -526,8 +528,17 @@ zexecve(char *pth, char **argv, char **newenvp)
+ 
+     if (newenvp == NULL)
+ 	    newenvp = environ;
++    exec_argv = argv;
++    if ((exec_wrapper = getenv("BASH_EXEC_WRAPPER")) &&
++	*exec_wrapper && !inblank(*exec_wrapper)) {
++	exec_argv = argv - 2;
++	exec_argv[0] = exec_wrapper;
++	exec_argv[1] = orig_pth;
```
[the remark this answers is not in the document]
nornagon-openai: Good callout. I verified this invariant in upstream zsh before landing the patch:

- `zexecve` is `static` in `Src/exec.c`, so all callsites are in that file.
- Every callsite passes the same `argv` allocated in `execute()` via `argv = makecline(args);`.
- `makecline()` intentionally returns `argv` with two leading spare slots (`argv = 2 + hcalloc(...)`).
- `zexecve()` already relies on that same padding today for shebang fallback with `argv[-1]` / `argv[-2]`.

So `argv - 2` here is using an existing zsh internal convention, not introducing a new out-of-bounds pattern. I agree it is an implicit invariant and would become unsafe if a future caller passed a non-`makecline()` argv.

shell-tool-mcp/patches/zsh-exec-wrapper.patch
```diff
@@ -0,0 +1,34 @@
...
++    char *exec_wrapper;
+ 
+     unmetafy(pth, NULL);
+     for (eep = argv; *eep; eep++)
+@@ -526,8 +528,17 @@ zexecve(char *pth, char **argv, char **newenvp)
+ 
+     if (newenvp == NULL)
+ 	    newenvp = environ;
++    exec_argv = argv;
++    if ((exec_wrapper = getenv("BASH_EXEC_WRAPPER")) &&
```
bolinfest: Should we change the env var?

nornagon-openai: yeah probably, i'll change it to just EXEC_WRAPPER i guess

PR #11669, opened 2026-02-13
core: limit search_tool_bm25 to Apps and clarify discovery guidance

apanasenko-oai:
## Summary
- Limit `search_tool_bm25` indexing to `codex_apps` tools only, so non-Apps MCP servers are no longer discoverable through this search path.
- Move search-tool discovery guidance into the `search_tool_bm25` tool description (via template include) instead of injecting it as a separate developer message.
- Update Apps discovery guidance wording to clarify when to use `search_tool_bm25` for Apps-backed systems (for example Slack, Google Drive, Jira, Notion) and when to call tools directly.
- Remove dead `core` helper code (`filter_codex_apps_mcp_tools` and `codex_apps_connector_id`) that is no longer used after the tool-selection refactor.
- Update `core` search-tool tests to assert codex-apps-only behavior and to validate guidance from the tool description.

## Validation
- ✅ `just fmt`
- ✅ `cargo test -p codex-core search_tool`
- ⚠️ `cargo test -p codex-core` was attempted, but the run repeatedly stalled on `tools::js_repl::tests::js_repl_can_attach_image_via_view_image_tool`.

## Tickets
- None


pakrym-oai (approved):

codex-rs/core/templates/search_tool/developer_instructions.md
```diff
@@ -24,6 +26,6 @@ Notes:
   - `connector_name`
   - `connector_id`
   - input schema property keys (`input_keys`)
-- When the user asks to search/lookup/query any external system (logs, tickets, metrics, Slack, etc.), you must call `search_tool_bm25` first before running any shell command or repo search.
-- Only use shell commands if (a) MCP tools for that system are not available or not sufficient, and (b) the user explicitly wants a local file/CLI search.
-- If unsure which system/tool applies, ask a clarifying question after checking MCP tools.
+- Use `search_tool_bm25` when the user asks to work with an Apps-backed external system (for example Slack, Google Drive, Jira, Notion) and the exact tool name is not already known.
```
mzeng-openai: This can still be vague to the model, should we explicitly add the name and description of the installed and enabled apps here?

apanasenko-oai: make sense, let's inject enabled app names.

PR #11671, opened 2026-02-13
chore(core) Restrict model-suggested rules

dylan-hurd-oai:
## Summary
If the model suggests a bad rule, don't show it to the user. This does not impact the parsing of existing rules, just the ones we show.

## Testing
- [x] Added unit tests
- [x] Ran locally


viyatb-oai (approved):
this is good list to start (but will mostly never be comprehensive) - we can combine this with some prompting to steer the model away from suggest broad prefixes

PR #11672, opened 2026-02-13
feat(core): add structured network approval plumbing and policy decision model

viyatb-oai:
#### Summary
Introduces the core plumbing required for structured network approvals

#### What changed
- Added structured network policy decision modeling in core.
- Added approval payload/context types needed for network approval semantics.
- Wired shell/unified-exec runtime plumbing to consume structured decisions.
- Updated related core error/event surfaces for structured handling.
- Updated protocol plumbing used by core approval flow.
- Included small CLI debug sandbox compatibility updates needed by this layer.

#### Why
establishes the minimal backend foundation for network approvals without yet changing high-level orchestration or TUI behavior.

#### Notes
- Behavior remains constrained by existing requirements/config gating.
- Follow-up PRs in the stack handle orchestration, UX, and app-server integration.


codex-rs/core/src/tools/runtimes/shell.rs
```diff
@@ -167,12 +171,63 @@ impl ToolRuntime<ShellRequest, ExecToolCallOutput> for ShellRuntime {
...
+                .unregister_network_approval_attempt(attempt_id)
+                .await;
+        }
+
+        if approval_outcome == Some(NetworkApprovalOutcome::DeniedByUser) {
+            return Err(ToolError::Rejected("rejected by user".to_string()));
+        }
+        if let Some(message) = network_policy_denial_message {
+            return Err(ToolError::Rejected(message));
+        }
```
dylan-hurd-oai: Can we abstract out some of this into a helper function?

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -170,11 +196,48 @@ impl UnifiedExecProcessManager {
...
+                        .session
+                        .unregister_network_approval_attempt(attempt_id)
+                        .await;
+                }
+                self.release_process_id(&request.process_id).await;
+                return Err(err);
+            }
+        };
+
+        let transcript = Arc::new(tokio::sync::Mutex::new(HeadTailBuffer::default()));
```
dylan-hurd-oai: Why are we moving this after the emitter setup?

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -224,25 +288,107 @@ impl UnifiedExecProcessManager {
...
+            if approval_outcome == Some(NetworkApprovalOutcome::DeniedByUser) {
+                return Err(UnifiedExecError::create_process(
+                    "rejected by user".to_string(),
+                ));
+            }
+            if let Some(message) = network_policy_denial_message_for_attempt(
+                request.network.as_ref(),
+                network_attempt_id.as_deref(),
+            )
+            .await
```
dylan-hurd-oai: do we have to be running this on every exec_command?

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -224,25 +288,107 @@ impl UnifiedExecProcessManager {
...
+            context,
+            &request.command,
+            cwd.clone(),
+            start,
+            process_id,
+            request.tty,
+            network_attempt_id.clone(),
+            Arc::clone(&transcript),
+        )
+        .await;
```
dylan-hurd-oai: @codex does this PR change the conditions which `self.store_process(` gets run?

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -224,25 +288,107 @@ impl UnifiedExecProcessManager {
...
+            self.release_process_id(&request.process_id).await;
+            return Err(UnifiedExecError::create_process(
+                "rejected by user".to_string(),
+            ));
+        }
+        if let Some(message) = network_policy_denial_message_for_attempt(
+            request.network.as_ref(),
+            network_attempt_id.as_deref(),
+        )
+        .await
```
dylan-hurd-oai: We're running this check twice - it's not immediately clear to me why

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -516,30 +677,42 @@ impl UnifiedExecProcessManager {
         request: &ExecCommandRequest,
         cwd: PathBuf,
         context: &UnifiedExecContext,
+        skip_command_approval: bool,
```
dylan-hurd-oai: I see this function signature change but no updates to a consumer - am I missing something?

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -516,30 +677,42 @@ impl UnifiedExecProcessManager {
...
-                command: &request.command,
-                approval_policy: context.turn.approval_policy,
-                sandbox_policy: &context.turn.sandbox_policy,
-                sandbox_permissions: request.sandbox_permissions,
-                prefix_rule: request.prefix_rule.clone(),
-            })
-            .await;
+        let exec_approval_requirement = if skip_command_approval {
+            // This is a same-call retry after explicit network approval.
+            // Do not re-prompt command approval.
```
dylan-hurd-oai: I'm not sure re-calling with `skip_command_approval` is the best pattern here - not going to block on this particular issue but I would love if we could find a more graceful approach

dylan-hurd-oai (changes requested):
I'm pretty concerned about the number of changes going into process_manager that are not gated behind a feature flag. will cc @jif-oai for visibility but I would love if we could figure out a way to test this without refactoring our load-bearing command executable path

codex-rs/core/src/unified_exec/process_manager.rs
viyatb-oai: fixed!

codex-rs/core/src/unified_exec/process_manager.rs
viyatb-oai: fixed, bad pr split!

codex-rs/core/src/unified_exec/process_manager.rs
viyatb-oai: fixed

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -224,25 +288,107 @@ impl UnifiedExecProcessManager {
...
+            };
+
+            return Ok(response);
+        }
+
+        let approval_outcome = if let Some(attempt_id) = network_attempt_id.as_deref() {
+            context
+                .session
+                .take_network_approval_outcome(attempt_id)
+                .await
```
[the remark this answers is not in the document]
viyatb-oai: fixed

codex-rs/core/src/unified_exec/process_manager.rs
viyatb-oai: fixed

codex-rs/core/src/unified_exec/process_manager.rs
viyatb-oai: fixed!

codex-rs/core/src/unified_exec/process_manager.rs
viyatb-oai: Good catch. I reverted that ordering change in `process_manager` and changed the `exec_command` flow back to `origin/main`
`start_streaming_output` is again set up immediately after spawn, before the initial output poll, matching origin/main behavior.
fixed!

codex-rs/core/src/tools/runtimes/shell.rs
viyatb-oai: done! looks much cleaner now

codex-rs/core/src/config/network_proxy_spec.rs
```diff
@@ -169,3 +178,64 @@ impl NetworkProxySpec {
...
+        Some(policy_decider) => builder.policy_decider_arc(policy_decider),
+        None => builder.policy_decider(|_request| async {
+            // In restricted sandbox modes, allowlist misses should ask for
+            // explicit network approval instead of hard-denying.
+            NetworkDecision::ask("not_allowed")
+        }),
+    }
+}
+
+fn should_ask_on_allowlist_miss(sandbox_policy: &SandboxPolicy) -> bool {
```
jif-oai: Inline this. In general, take the habit of inlining such simple functions. It just makes the code harder to read

codex-rs/core/src/config/network_proxy_spec.rs
```diff
@@ -92,20 +96,25 @@ impl NetworkProxySpec {
...
         let reloader = Arc::new(StaticNetworkProxyReloader::new(state.clone()));
         let state = NetworkProxyState::with_reloader(state, reloader);
-        let proxy = NetworkProxy::builder()
-            .state(Arc::new(state))
-            .build()
-            .await
-            .map_err(|err| {
-                std::io::Error::other(format!("failed to build network proxy: {err}"))
-            })?;
+        let builder = configure_policy_decider(
```
jif-oai: this pattern is pretty strange. The whole point of a builder is to be able to queue the operations. It's pretty strange to send the building to a static function somewhere and getting it back...

codex-rs/core/src/config/network_proxy_spec.rs
```diff
@@ -169,3 +178,64 @@ impl NetworkProxySpec {
...
+
+fn should_ask_on_allowlist_miss(sandbox_policy: &SandboxPolicy) -> bool {
+    matches!(
+        sandbox_policy,
+        SandboxPolicy::ReadOnly { .. } | SandboxPolicy::WorkspaceWrite { .. }
+    )
+}
+
+#[cfg(test)]
+mod tests {
```
jif-oai: those tests are pointless. They basically test a simple `match`

codex-rs/protocol/src/approvals.rs
```diff
@@ -52,6 +65,10 @@ pub struct ExecApprovalRequestEvent {
     /// Optional human-readable reason for the approval (e.g. retry without sandbox).
     #[serde(skip_serializing_if = "Option::is_none")]
     pub reason: Option<String>,
+    /// Optional network context for a blocked request that can be approved.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub network_approval_context: Option<NetworkApprovalContext>,
```
jif-oai, line 76: Shouldn't this be a list? I feel like a request could actually ping multiple hosts at the same time

codex-rs/protocol/src/approvals.rs
```diff
@@ -37,6 +37,19 @@ impl From<Vec<String>> for ExecPolicyAmendment {
     }
 }
 
+#[derive(Debug, Clone, Copy, Deserialize, Serialize, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "snake_case")]
+pub enum NetworkApprovalProtocol {
+    Http,
+    Https,
```
jif-oai, line 47: Probably good to include web sockets at least here

codex-rs/core/src/network_policy_decision.rs
```diff
@@ -0,0 +1,194 @@
+use codex_network_proxy::BlockedRequest;
+use codex_network_proxy::NetworkDecisionSource;
+use codex_network_proxy::NetworkPolicyDecision;
+use codex_protocol::approvals::NetworkApprovalContext;
+use codex_protocol::approvals::NetworkApprovalProtocol;
+use serde::Deserialize;
+
+#[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct NetworkPolicyDecisionPayload {
```
jif-oai, line 10: For the different field why not using the types defined in `codex-rs/protocol/src/approvals.rs`?

codex-rs/core/src/network_policy_decision.rs
```diff
@@ -0,0 +1,194 @@
...
+    pub decision: String,
+    pub source: String,
+    pub protocol: Option<String>,
+    pub host: Option<String>,
+    pub reason: Option<String>,
+    pub port: Option<u16>,
+}
+
+impl NetworkPolicyDecisionPayload {
+    pub(crate) fn is_ask_from_decider(&self) -> bool {
```
jif-oai, line 21: Same comment, if we had good types in `NetworkPolicyDecisionPayload` instead of just String we wouldn't need this sketchy `eq_ignore_ascii_case`

codex-rs/core/src/network_policy_decision.rs
```diff
@@ -0,0 +1,194 @@
...
+    {
+        return None;
+    }
+
+    let host = blocked.host.trim();
+    if host.is_empty() {
+        return Some("Network access was blocked by policy.".to_string());
+    }
+
+    let detail = match blocked.reason.as_str() {
```
jif-oai, line 68: this is odd as well to do string matching on something we actually control but it seems to be out of scope for this PR

codex-rs/core/src/exec.rs
```diff
@@ -66,6 +67,7 @@ pub struct ExecParams {
     pub expiration: ExecExpiration,
     pub env: HashMap<String, String>,
     pub network: Option<NetworkProxy>,
+    pub network_attempt_id: Option<String>,
```
jif-oai: OOC why using a String and not a type with constant byte size (e.g. a UUID)? This would allocate it on the stack instead of the heap, give you free `Copy` trait and simplify all your life

codex-rs/core/src/codex.rs
```diff
@@ -1212,13 +1234,45 @@ impl Session {
...
-        let network_proxy =
-            match config.permissions.network.as_ref() {
-                Some(spec) => Some(spec.start_proxy().await.map_err(|err| {
-                    anyhow::anyhow!("failed to start managed network proxy: {err}")
-                })?),
-                None => None,
-            };
+        // The managed proxy can call back into core for allowlist-miss decisions.
+        let (inline_network_decider_session, inline_network_decider): (
+            Option<Arc<RwLock<std::sync::Weak<Session>>>>,
```
jif-oai: This is pretty sus... having an `Arc<RwLock<Weak<...>>` looks like a smell. I'm not even sure what you try to achieve here?! It seems to be just because you instantiate the decider before the Session but you could just build it after...

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -87,6 +91,44 @@ fn apply_unified_exec_env(mut env: HashMap<String, String>) -> HashMap<String, S
     env
 }
 
+async fn network_policy_denial_message_for_attempt(
```
jif-oai: this function has a duplicate in `shell.rs`. Those should live close to `codex_network_proxy` IMO

codex-rs/core/src/tools/runtimes/shell.rs
```diff
@@ -65,6 +68,83 @@ impl ShellRuntime {
...
+            )
+            .await;
+    }
+    env.network_attempt_id = network_attempt_id.clone();
+
+    let out = execute_env(env, attempt.policy, ShellRuntime::stdout_stream(ctx))
+        .await
+        .map_err(ToolError::Codex);
+    let approval_outcome = if let Some(attempt_id) = network_attempt_id.as_deref() {
+        ctx.session.take_network_approval_outcome(attempt_id).await
```
jif-oai: This code should not live here. The approval handling with the error reporting etc should live in the orchestrator. This makes things more linear and easier to understand. In this case this is not clear how this behaviour fit with the other tools

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -471,10 +631,40 @@ impl UnifiedExecProcessManager {
...
+                            break;
+                        }
+                        _ = poll.tick() => {
+                            if session.take_network_approval_outcome(&network_attempt_id).await
+                                == Some(NetworkApprovalOutcome::DeniedByUser)
+                            {
+                                tracing::debug!(
+                                    "terminating unified exec process {process_id} after delayed network denial for attempt {network_attempt_id}"
+                                );
+                                process.terminate();
```
jif-oai: If you do this, you must unregister the attempt, make sure the process ID is released etc

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -224,9 +312,66 @@ impl UnifiedExecProcessManager {
             )
             .await;
 
+            let approval_outcome = take_network_approval_outcome_for_attempt(
```
jif-oai: Same point as before, this should live in the orchestrator. That will make our life way easier

jif-oai (dismissed):
I pause the review here. The direction is good but the design is not ready to be merged yet. This is too entangled with the existing code and complexify too much the code around shell and process handling. IMO we should extract this as a clean layer on top of it (this was the original point of the orchestrator) instead of having to cover all the corner cases within the runtimes

codex-rs/protocol/src/approvals.rs
viyatb-oai: the proxy does not support websockets as of yet, I will update it after this pr

codex-rs/protocol/src/approvals.rs
viyatb-oai: no, this is per domain.

codex-rs/core/src/network_policy_decision.rs
viyatb-oai: fixed.

codex-rs/core/src/network_policy_decision.rs
viyatb-oai: i'll clean this up in a follow up pr in the proxy

codex-rs/core/src/unified_exec/process_manager.rs
viyatb-oai: added

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -1688,8 +1688,22 @@ impl CodexMessageProcessor {
         let timeout_ms = params
             .timeout_ms
             .and_then(|timeout_ms| u64::try_from(timeout_ms).ok());
+        let managed_network_requirements_enabled = self
+            .config
+            .config_layer_stack
+            .requirements_toml()
+            .network
+            .is_some();
         let started_network_proxy = match self.config.permissions.network.as_ref() {
```
dylan-hurd-oai, line 1693: 1. seems like we're checking both requirements and config.permissions.network here? Do we need both?
2. Seems like these should be helper functions, since the logic for `managed_network_requirements_enabled` is going to change soon?

codex-rs/cli/src/debug_sandbox.rs
```diff
@@ -213,12 +213,23 @@ async fn run_command_under_sandbox(
     #[cfg(not(target_os = "macos"))]
     let _ = log_denials;
 
+    let managed_network_requirements_enabled = config
+        .config_layer_stack
+        .requirements_toml()
+        .network
+        .is_some();
```
dylan-hurd-oai: helper function

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -380,7 +438,13 @@ impl UnifiedExecProcessManager {
                 call_id: entry.call_id.clone(),
                 process_id,
             }
+        };
+
+        drop(store);
```
dylan-hurd-oai: why do we need this now?

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -437,22 +501,30 @@ impl UnifiedExecProcessManager {
...
-            Self::prune_processes_if_needed(&mut store);
+            let pruned_entry = Self::prune_processes_if_needed(&mut store);
             store.processes.insert(process_id.clone(), entry);
-            store.processes.len()
+            (store.processes.len(), pruned_entry)
         };
+        if let Some(pruned_entry) = pruned_entry {
+            Self::unregister_network_attempt_for_entry(&pruned_entry).await;
+            pruned_entry.process.terminate();
+        }
```
dylan-hurd-oai, line 527: Is there some lifecycle-related reason that we need to move the terminate call out of `prune_processes_if_needed`? Could we just call `unregister_network_attempt_for_entry` at the existing terminate callsite?

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -471,10 +543,46 @@ impl UnifiedExecProcessManager {
...
             transcript,
             started_at,
         );
+
+        if let Some(network_attempt_id) = network_attempt_id_for_watcher {
+            let exit_token = process.cancellation_token();
+            let session = Arc::clone(&context.session);
+            let process = Arc::clone(&process);
+            let process_id = process_id.clone();
+            tokio::spawn(async move {
```
dylan-hurd-oai: I think this logic should be (1) in async_watcher.rs (either in the spawn exit watcher  or its own function/thread) and (2) gated on feature flag / config

codex-rs/core/src/codex.rs
```diff
@@ -1209,13 +1212,53 @@ impl Session {
...
-                Some(spec) => Some(spec.start_proxy().await.map_err(|err| {
-                    anyhow::anyhow!("failed to start managed network proxy: {err}")
-                })?),
-                None => None,
-            };
+        let managed_network_requirements_enabled = config
+            .config_layer_stack
+            .requirements_toml()
+            .network
+            .is_some();
```
dylan-hurd-oai: helper function

codex-rs/core/src/codex.rs
```diff
@@ -1209,13 +1212,53 @@ impl Session {
...
+                    config.permissions.sandbox_policy.get(),
+                    network_policy_decider.as_ref().map(Arc::clone),
+                    blocked_request_observer.as_ref().map(Arc::clone),
+                    managed_network_requirements_enabled,
+                )
+                .await
+                .map_err(|err| anyhow::anyhow!("failed to start managed network proxy: {err}"))?,
+            ),
+            None => None,
+        };
```
dylan-hurd-oai: If i'm understanding this code correctly, we have a series of steps here and then `network_proxy` / `session_network_proxy` are the only outputs we need, right? Would this be cleaner in a helper function?

codex-rs/core/src/network_policy_decision.rs
```diff
@@ -0,0 +1,284 @@
...
+use serde::de::Error;
+
+#[derive(Debug, Clone, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct NetworkPolicyDecisionPayload {
+    pub decision: NetworkPolicyDecision,
+    pub source: NetworkDecisionSource,
+    #[serde(
+        default,
+        deserialize_with = "deserialize_network_approval_protocol_option"
```
dylan-hurd-oai: Why are we setting up a custom deserializer here? Is it just for `.ok_or_else(|| Error::custom(format!("unsupported network protocol: {value}")))`?

dylan-hurd-oai (commented):
This is a lot cleaner - some additional comments to address

codex-rs/core/src/unified_exec/process_manager.rs
viyatb-oai: much cleaner, done

codex-rs/core/src/codex.rs
```diff
@@ -1246,27 +1275,13 @@ impl Session {
...
-                admin_addr: proxy.admin_addr().to_string(),
-            }
-        });
+        let (network_proxy, session_network_proxy) = Self::start_managed_network_proxy(
+            config.as_ref(),
+            network_policy_decider.as_ref().map(Arc::clone),
+            blocked_request_observer.as_ref().map(Arc::clone),
+            managed_network_requirements_enabled,
+        )
+        .await?;
```
dylan-hurd-oai: I would prefer to gate outside the `await` call here

dylan-hurd-oai (commented):
so close! Also, codex had this to say:

```
The new denial watcher drops non-user network outcomes, which can suppress deferred policy-denial errors for long-lived unified-exec processes. That behavior
  is a functional regression in network-approval handling.

  Review comment:

  - [P2] Preserve non-user network outcomes in denial watcher — /Users/dylan.hurd/code/codex-3/codex-rs/core/src/unified_exec/async_watcher.rs:163-166
    The watcher consumes approval outcomes with take_outcome() on every poll, but only acts when the value is DeniedByUser. If a long-lived unified-exec
    process records DeniedByPolicy, this code silently drops that outcome and continues, so later calls cannot surface the policy rejection message (the
    outcome was already removed). This is reproducible when a running process hits a deny-only host after startup; the proxy blocks it, but the deferred
    rejection signal is lost
```

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -130,8 +134,27 @@ impl UnifiedExecProcessManager {
...
+            return;
+        };
+        let Some(session) = entry.session.upgrade() else {
+            return;
+        };
+        session
+            .services
+            .network_approval
+            .unregister_attempt(attempt_id)
+            .await;
```
dylan-hurd-oai:
```suggestion
        if let Some(attempt_id) = entry.network_attempt_id.as_deref()
            && let Some(session) = entry.session.upgrade()  {
            session
              .services
              .network_approval
              .unregister_attempt(attempt_id)
              .await;
        };
```

codex-rs/core/src/codex.rs
```diff
@@ -863,35 +863,29 @@ impl Session {
...
-                )
-                .await
-                .map_err(|err| anyhow::anyhow!("failed to start managed network proxy: {err}"))?,
-            ),
-            None => None,
-        };
-        let session_network_proxy = network_proxy.as_ref().map(|started| {
-            let proxy = started.proxy();
+    ) -> anyhow::Result<(StartedNetworkProxy, SessionNetworkProxyRuntime)> {
+        let network_proxy = spec
```
dylan-hurd-oai, line 872: Why did we remove the conditional here?

dylan-hurd-oai (approved):

codex-rs/core/src/codex.rs
viyatb-oai: The conditional was moved, not removed:

we now gate with:
`if let Some(spec) = config.permissions.network.as_ref() { ... await ... } else { (None, None) }`
The helper `start_managed_network_proxy(...)` now assumes spec exists and returns a non-optional (StartedNetworkProxy, SessionNetworkProxyRuntime).

PR #11673, opened 2026-02-13
feat(core): wire structured network approvals through orchestration and retry flow

viyatb-oai:
#### Summary
Integrates structured network approvals into the core orchestration path.

#### What changed
- Wired structured network approval handling into tool orchestration/retry paths.
- Integrated approval outcomes into command execution lifecycle.
- Added orchestration-level handling for allow-once/session approval behavior.
- Connected core session/delegate surfaces required for end-to-end approval flow.
- Updated related core test coverage for orchestration behavior.

#### Why
With plumbing in place from parent, this PR makes network approvals operational in core command execution flow.

#### Notes
- Gated behavior remains aligned with managed network requirements and sandbox policy.


codex-rs/core/src/codex.rs
```diff
@@ -1174,13 +1196,32 @@ impl Session {
...
+                let inline_network_decider_session = Arc::clone(&inline_network_decider_session);
+                async move {
+                    let Some(session) = inline_network_decider_session.read().await.upgrade()
+                    else {
+                        return NetworkDecision::ask("not_allowed");
+                    };
+                    session.handle_inline_network_policy_request(request).await
+                }
+            }
+        });
```
dylan-hurd-oai, line 1213: 1. It looks like we're constructing this regardless of whether network config is set up? Why not put this inside the let block?
2. Can we abstract this from the core function here?

dylan-hurd-oai (changes requested):
Would like to shield this logic more strictly from the core loop when disabled

PR #11677, opened 2026-02-13
turn metadata: per-turn non-blocking

pash-openai:


codex-rs/core/src/tasks/regular.rs
```diff
@@ -39,13 +38,11 @@ impl RegularTask {
         model_client: ModelClient,
         otel_manager: OtelManager,
         model_info: ModelInfo,
-        turn_metadata_header: BoxFuture<'static, Option<String>>,
     ) -> Self {
         let prewarmed_session_task = tokio::spawn(async move {
             let mut client_session = model_client.new_session();
-            let turn_metadata_header = turn_metadata_header.await;
             match client_session
-                .prewarm_websocket(&otel_manager, &model_info, turn_metadata_header.as_deref())
```
pash-openai, line 48: We'll be moving towards either:

- a separate event with a new message for this metadata
- or appending it to the request bodies  of existing WS messages

Either way, it won't be added as a header in the handshake.

codex-rs/core/src/client.rs
```diff
@@ -656,7 +655,7 @@ impl ModelClientSession {
                 client_setup.api_provider,
                 client_setup.api_auth,
                 Some(Arc::clone(&self.turn_state)),
-                turn_metadata_header,
+                None,
```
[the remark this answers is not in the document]
pash-openai: not used - we're moving x-codex-turn-metadata out of the prewarm loop entirely, and moving to a turn-scoped approach.

codex-rs/core/src/codex.rs
```diff
@@ -560,7 +558,7 @@ pub(crate) struct TurnContext {
     pub(crate) truncation_policy: TruncationPolicy,
     pub(crate) js_repl: Arc<JsReplHandle>,
     pub(crate) dynamic_tools: Vec<DynamicToolSpec>,
-    turn_metadata_header: OnceCell<Option<String>>,
+    turn_metadata_state: Arc<TurnMetadataState>,
```
pakrym-oai, line 561: feel free to make this pub and avoid geters/seters

codex-rs/core/src/codex.rs
```diff
@@ -950,6 +925,17 @@ impl Session {
         });
 
         let cwd = session_configuration.cwd.clone();
+        let turn_metadata_state = Arc::new(TurnMetadataState::new(
+            sub_id.clone(),
+            cwd.clone(),
+            Some(
```
pakrym-oai, line 931: consider moving sandbox_tag calculation logic into TurnMetadataState implementation.

pakrym-oai (approved):

codex-rs/core/src/codex.rs
pash-openai: Addressed both + cleanup in https://github.com/openai/codex/pull/11782

PR #11687, opened 2026-02-13
support app usage analytics

alexsong-oai:
Emit app mentioned and app used events. Dedup by (turn_id, connector_id)

Example event params:
{
    "event_type": "codex_app_used",
    "connector_id": "asdk_app_xxx",
    "thread_id": "019c5527-36d4-xxx",
    "turn_id": "019c552c-cd17-xxx",
    "app_name": "Slack (OpenAI Internal)",
    "product_client_id": "codex_cli_rs",
    "invoke_type": "explicit",
    "model_slug": "gpt-5.3-codex"
}

### DIFF
--- a/codex-rs/core/src/analytics_client.rs
+++ b/codex-rs/core/src/analytics_client.rs
@@ -7,9 +7,11 @@ use codex_protocol::protocol::SkillScope;
 use serde::Serialize;
 use sha1::Digest;
 use sha1::Sha1;
+use std::collections::HashSet;
 use std::path::Path;
 use std::path::PathBuf;
 use std::sync::Arc;
+use std::sync::Mutex;
 use std::time::Duration;
 use tokio::sync::mpsc;
 
@@ -47,6 +49,7 @@ pub(crate) struct AppInvocation {
 #[derive(Clone)]
 pub(crate) struct AnalyticsEventsQueue {
     sender: mpsc::Sender<TrackEventsJob>,
+    app_used_emitted_keys: Arc<Mutex<HashSet<(String, String)>>>,
 }
 
 pub(crate) struct AnalyticsEventsClient {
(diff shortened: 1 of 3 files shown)


mzeng-openai (approved):

codex-rs/core/src/analytics_client.rs
```diff
@@ -84,32 +123,76 @@ impl AnalyticsEventsClient {
             invocations,
         );
     }
+
+    pub(crate) fn track_app_mentioned(
```
xl-openai, line 127: Why do you need this? can you call the underlying track_app_mentioned directly?

codex-rs/core/src/analytics_client.rs
```diff
@@ -35,9 +40,16 @@ pub(crate) struct SkillInvocation {
...
+pub(crate) struct AppInvocation {
+    pub(crate) connector_id: Option<String>,
+    pub(crate) app_name: Option<String>,
+    pub(crate) invoke_type: Option<String>,
+}
+
 #[derive(Clone)]
 pub(crate) struct AnalyticsEventsQueue {
     sender: mpsc::Sender<TrackEventsJob>,
+    app_used_emitted_keys: Arc<Mutex<HashSet<(String, String)>>>,
```
xl-openai, line 52: I think it is OK, but just to confirm, this will be cleared when you close codex then resume right?

alexsong-oai: yes

codex-rs/core/src/analytics_client.rs
alexsong-oai: pub(crate) fn track_app_mentioned is a wrapper around the underlying free function. Since the free function does not have access to self.queue, we need to pass it as a param. Can also move the free function’s logic into this method instead; either should be fine.

PR #11690, opened 2026-02-13
fix: show user warning when using default fallback metadata

sayan-oai:
### What
It's currently unclear when the harness falls back to the default, generic `ModelInfo`. This happens when the `remote_models` feature is disabled or the model is truly unknown, and can lead to bad performance and issues in the harness.

Add a user-facing warning when this happens so they are aware when their setup is broken.

### Tests
Added tests, tested locally.


codex-rs/core/src/models_manager/manager.rs
```diff
@@ -43,6 +43,13 @@ pub enum RefreshStrategy {
     OnlineIfUncached,
 }
 
+/// Result of resolving model metadata for a requested model slug.
+#[derive(Debug, Clone)]
+pub(crate) struct ModelInfoResolution {
+    pub(crate) model_info: ModelInfo,
+    pub(crate) used_fallback_model_metadata: bool,
```
pakrym-oai: can we toss this flag right onto the model info?

codex-rs/core/src/state/session.rs
```diff
@@ -49,6 +50,7 @@ impl SessionState {
             startup_regular_task: None,
             active_mcp_tool_selection: None,
             active_connector_selection: HashSet::new(),
+            warned_unknown_models: HashSet::new(),
```
pakrym-oai: This should happen only on the start. No way to change to unknown model from UI. What am I missing?

Can we rem the field?

pakrym-oai (approved):
Looks reasonable but I think can be much simplified by moving the used_fallback to model info and not caching warnings.

codex-rs/core/src/state/session.rs
sayan-oai: before the decision to rm RemoteModels, switching to any model would use the fallback, so you could switch to an "unknown" model. but yes now that can be simplified bc we're rm'ing RemoteModels

codex-rs/protocol/src/openai_models.rs
```diff
@@ -253,6 +253,11 @@ pub struct ModelInfo {
     /// When true, this model should use websocket transport even when websocket features are off.
     #[serde(default)]
     pub prefer_websockets: bool,
+    /// Internal-only marker set by core when a model slug resolved to fallback metadata.
+    #[serde(default, skip_serializing, skip_deserializing)]
+    #[schemars(skip)]
+    #[ts(skip)]
+    pub used_fallback_model_metadata: bool,
```
sayan-oai, line 260: marked as internal only because its not actually part of model metadata; we are just using it to track local resolution state. i could make it public, though i think that's misleading, or we could go back to the wrapper struct

codex-rs/core/src/models_manager/manager.rs
sayan-oai: moved to model info but marked as internal only

codex-rs/core/src/models_manager/manager.rs
```diff
@@ -140,15 +140,20 @@ impl ModelsManager {
...
+            (
+                ModelInfo {
+                    slug: model.to_string(),
+                    ..remote
+                },
+                false,
+            )
         } else {
-            model_info::model_info_from_slug(model)
+            (model_info::model_info_from_slug(model), true)
```
aibrahim-oai: make model_info_from_slug itself insert the value directly

codex-rs/core/src/codex.rs
```diff
@@ -1781,6 +1779,17 @@ impl Session {
             turn_context.final_output_json_schema = final_schema;
         }
         let turn_context = Arc::new(turn_context);
+        if model_info.used_fallback_model_metadata {
+            self.send_event(
+                turn_context.as_ref(),
+                EventMsg::Warning(WarningEvent {
+                    message: format!(
+                        "Defaulting to fallback model metadata for `{resolved_model_slug}`. This can degrade performance and cause issues."
```
aibrahim-oai: I think the copy can say sth like it's not found first.

aibrahim-oai (approved):

PR #11706, opened 2026-02-13
[apps] Expose more fields from apps listing endpoints.

mzeng-openai:
- [x] Expose app_metadata, branding, and labels in AppInfo.


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1276,6 +1276,65 @@ pub struct AppsListParams {
...
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+/// EXPERIMENTAL - app metadata returned by app-list APIs.
+pub struct AppBranding {
+    pub category: Option<String>,
+    pub developer: Option<String>,
+    pub website: Option<String>,
+    #[serde(alias = "privacy_policy")]
```
owenlin0: remove all serde aliases so that these fields are properly exposed as camelcase

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1276,6 +1276,65 @@ pub struct AppsListParams {
...
+    pub terms_of_service: Option<String>,
+    #[serde(alias = "is_discoverable_app", default)]
+    pub is_discoverable_app: bool,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct AppReview {
+    pub status: String,
```
owenlin0, line 1300: should this be an enum?

canvrno-oai (approved):

codex-rs/app-server-protocol/src/protocol/v2.rs
mzeng-openai: These are coming down from the connectors backend so I would use string to avoid being out-of-sync

owenlin0 (approved):

PR #11711, opened 2026-02-13
feat(tui): prevent macOS idle sleep while turns run

yvolovich-cyber:
## Summary
- add a shared `codex-core` sleep inhibitor that uses native macOS IOKit assertions (`IOPMAssertionCreateWithName` / `IOPMAssertionRelease`) instead of spawning `caffeinate`
- wire sleep inhibition to turn lifecycle in `tui` (`TurnStarted` enables; `TurnComplete` and abort/error finalization disable)
- gate this behavior behind a `/experimental` feature toggle (`[features].prevent_idle_sleep`) instead of a dedicated `[tui]` config flag
- expose the toggle in `/experimental` on macOS; keep it under development on other platforms
- keep behavior no-op on non-macOS targets

image

## Testing
- `cargo check -p codex-core -p codex-tui`
- `cargo test -p codex-core sleep_inhibitor::tests -- --nocapture`
- `cargo test -p codex-core tui_config_missing_notifications_field_defaults_to_enabled -- --nocapture`
- `cargo test -p codex-core prevent_idle_sleep_is_ -- --nocapture`

## Semantics and API references
- This PR targets `caffeinate -i` semantics: prevent *idle system sleep* while allowing display idle sleep.
- `caffeinate -i` mapping in Apple open source (`assertionMap`):
  - `kIdleAssertionFlag -> kIOPMAssertionTypePreventUserIdleSystemSleep`
  - Source: https://github.com/apple-oss-distributions/PowerManagement/blob/PowerManagement-1846.60.12/caffeinate/caffeinate.c#L52-L54
- Apple IOKit docs for assertion types and API:
  - https://developer.apple.com/documentation/iokit/iopmlib_h/iopmassertiontypes
  - https://developer.apple.com/documentation/iokit/1557092-iopmassertioncreatewithname
  - https://developer.apple.com/library/archive/qa/qa1340/_index.html

## Codex Electron vs this PR (full stack path)
- Codex Electron app requests sleep blocking with `powerSaveBlocker.start("prevent-app-suspension")`:
  - https://github.com/openai/codex/blob/main/codex/codex-vscode/electron/src/electron-message-handler.ts
- Electron maps that string to Chromium wake lock type `kPreventAppSuspension`:
  - https://github.com/electron/electron/blob/main/shell/browser/api/electron_api_power_save_blocker.cc
- Chromium macOS backend maps wake lock types to IOKit assertion constants and calls IOKit:
  - `kPreventAppSuspension -> kIOPMAssertionTypeNoIdleSleep`
  - `kPreventDisplaySleep / kPreventDisplaySleepAllowDimming -> kIOPMAssertionTypeNoDisplaySleep`
  - https://github.com/chromium/chromium/blob/main/services/device/wake_lock/power_save_blocker/power_save_blocker_mac.cc

## Why this PR uses a different macOS constant name
- This PR uses `"PreventUserIdleSystemSleep"` directly, via `IOPMAssertionCreateWithName`, in `codex-rs/core/src/sleep_inhibitor.rs`.
- Apple’s IOKit header documents `kIOPMAssertionTypeNoIdleSleep` as deprecated and recommends `kIOPMAssertPreventUserIdleSystemSleep` / `kIOPMAssertionTypePreventUserIdleSystemSleep`:
  - https://github.com/apple-oss-distributions/IOKitUser/blob/IOKitUser-100222.60.2/pwr_mgt.subproj/IOPMLib.h#L1000-L1030
- So Chromium and this PR are using different constant names, but semantically equivalent idle-system-sleep prevention behavior.

## Future platform support
The architecture is intentionally set up for multi-platform extensions:
- UI code (`tui`) only calls `SleepInhibitor::set_turn_running(...)` on turn lifecycle boundaries.
- Platform-specific behavior is isolated in `codex-rs/core/src/sleep_inhibitor.rs` behind `cfg(...)` blocks.
- Feature exposure is centralized in `core/src/features.rs` and surfaced via `/experimental`.
- Adding new OS backends should not require additional TUI wiring; only the backend internals and feature stage metadata need to change.

Potential follow-up implementations:
- Windows:
  - Add a backend using Win32 power APIs (`SetThreadExecutionState(ES_CONTINUOUS | ES_SYSTEM_REQUIRED)` as baseline).
  - Optionally move to `PowerCreateRequest` / `PowerSetRequest` / `PowerClearRequest` for richer assertion semantics.
- Linux:
  - Add a backend using logind inhibitors over D-Bus (`org.freedesktop.login1.Manager.Inhibit` with `what="sleep"`).
  - Keep a no-op fallback where logind/D-Bus is unavailable.

This PR keeps the cross-platform API surface minimal so future PRs can add Windows/Linux support incrementally with low churn.


codex-rs/core/src/sleep_inhibitor.rs
```diff
@@ -0,0 +1,151 @@
...
+#[cfg(test)]
+mod tests {
+    use super::SleepInhibitor;
+
+    #[test]
+    fn sleep_inhibitor_toggles_without_panicking() {
+        let mut inhibitor = SleepInhibitor::new(true);
+        inhibitor.set_turn_running(true);
+        inhibitor.set_turn_running(false);
+    }
```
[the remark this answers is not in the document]
yvolovich-cyber: Added the three edge-case tests in `codex-rs/core/src/sleep_inhibitor.rs`:
- disabled mode
- repeated `set_turn_running(true)` idempotency
- multi-toggle reacquisition

codex-rs/core/src/sleep_inhibitor.rs
```diff
@@ -0,0 +1,151 @@
...
+        }
+    }
+
+    /// Update the active turn state; turns sleep prevention on/off as needed.
+    pub fn set_turn_running(&mut self, turn_running: bool) {
+        if !self.enabled || !turn_running {
+            self.release();
+            return;
+        }
+        self.acquire();
```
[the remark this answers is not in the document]
yvolovich-cyber: Refactored `set_turn_running` to make the control flow explicit (`!enabled` guard, then acquire/release by `turn_running`).
Behavior is unchanged, but the intent is clearer.

codex-rs/core/src/config/types.rs
```diff
@@ -533,6 +533,13 @@ pub struct Tui {
     #[serde(default = "default_true")]
     pub show_tooltips: bool,
 
+    /// Keep the machine awake while a turn is actively running.
+    ///
+    /// On macOS this uses native IOKit power assertions. On other platforms this setting is
+    /// currently ignored.
+    #[serde(default)]
+    pub prevent_sleep_while_running: bool,
```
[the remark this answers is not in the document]
yvolovich-cyber: Addressed. I regenerated `codex-rs/core/config.schema.json` in commit `0c4a61ce1` (`just write-config-schema`) and that commit is included in this branch. Also, the implementation was later migrated from `tui.prevent_sleep_while_running` to the experimental feature key `features.prevent_idle_sleep`, and the schema fixture reflects that current shape.

codex-rs/utils/sleep-inhibitor/src/macos_inhibitor.rs
```diff
@@ -0,0 +1,277 @@
...
+        static API: OnceLock<Option<MacSleepApi>> = OnceLock::new();
+        *API.get_or_init(Self::load)
+    }
+
+    fn load() -> Option<Self> {
+        let handle = unsafe {
+            libc::dlopen(
+                IOKIT_FRAMEWORK_BINARY.as_ptr().cast(),
+                libc::RTLD_LOCAL | libc::RTLD_LAZY,
+            )
```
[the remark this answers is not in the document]
yvolovich-cyber: Good call; addressed now.

`MacSleepApi` now explicitly retains the `dlopen` handle (`_iokit_handle`) and `MacSleepApi::get()` returns `Option<&'static MacSleepApi>` from a `OnceLock`, so the symbol owner/lifetime is explicit and tied to process lifetime. I also added comments in the struct documenting why the handle is intentionally retained.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -1276,6 +1278,7 @@ impl ChatWidget {
 
     fn on_task_started(&mut self) {
         self.agent_turn_running = true;
+        self.turn_sleep_inhibitor.set_turn_running(true);
         self.saw_plan_update_this_turn = false;
```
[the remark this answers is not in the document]
yvolovich-cyber: Addressed in `fe86c57d4`.

I routed `TurnStarted` through a replay-aware path: `dispatch_event_msg` now calls `on_task_started_with_source(from_replay)`, and sleep inhibition is only enabled for live events (`!from_replay`). This keeps replay from creating macOS power assertions while preserving normal turn lifecycle behavior for live runs.

codex-rs/core/src/features.rs
```diff
@@ -604,6 +606,19 @@ pub const FEATURES: &[FeatureSpec] = &[
...
         default_enabled: true,
     },
+    FeatureSpec {
+        id: Feature::PreventIdleSleep,
+        key: "prevent_idle_sleep",
+        #[cfg(target_os = "macos")]
+        stage: Stage::Experimental {
+            name: "Prevent sleep while running",
+            menu_description: "Keep your computer awake while Codex is running a thread.",
+            announcement: "NEW: Prevent sleep while running is now available in /experimental.",
```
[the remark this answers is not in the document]
yvolovich-cyber: Updated in `fe86c57d4`.

I changed the `/experimental` copy to reflect the actual semantics and lifecycle:
- `name`: `Prevent idle system sleep during active turns`
- `menu_description`: `On macOS, prevent idle system sleep while Codex is actively running a turn.`
- `announcement`: `NEW: Prevent idle system sleep during active turns is now available in /experimental.`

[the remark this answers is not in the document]
yvolovich-cyber: Follow-up: I adjusted this copy again in `90f7346e6` to align with the existing Mac app wording and avoid introducing new jargon. Current strings are:
- `Prevent sleep while running`
- `Keep your computer awake while Codex is running a thread.`
- `NEW: Prevent sleep while running is now available in /experimental.`

codex-rs/core/src/features.rs
```diff
@@ -604,6 +606,19 @@ pub const FEATURES: &[FeatureSpec] = &[
         stage: Stage::Stable,
         default_enabled: true,
     },
+    FeatureSpec {
+        id: Feature::PreventIdleSleep,
+        key: "prevent_idle_sleep",
+        #[cfg(target_os = "macos")]
```
jif-oai: Instead of this, could you use a `if cfg!(...)`. Check what is done for `UnifiedExec`
This is better because this allow full compilation locally.

codex-rs/core/src/features.rs
```diff
@@ -722,4 +737,21 @@ mod tests {
         );
         assert_eq!(Feature::UseLinuxSandboxBwrap.default_enabled(), false);
     }
+
+    #[cfg(target_os = "macos")]
+    #[test]
+    fn prevent_idle_sleep_is_experimental_on_macos() {
```
jif-oai: Nice but not really useful and will just add effort when we land this. But it's personal opinion, feel free to ignore

codex-rs/core/src/sleep_inhibitor.rs
```diff
@@ -0,0 +1,280 @@
+//! Cross-platform helper for preventing idle sleep while a turn is running.
```
jif-oai: Could we extract this in a small crate in `codex-rs/utils/...` ? I guess this can be re-used later

codex-rs/core/src/sleep_inhibitor.rs
```diff
@@ -0,0 +1,280 @@
...
+//!
+//! On macOS this uses native IOKit power assertions instead of spawning
+//! `caffeinate`, so assertion lifecycle is tied directly to Rust object lifetime.
+
+#[cfg(target_os = "macos")]
```
jif-oai: If everything is just for MacOS, it would be even easier if you move everything into a crate. You can put this in a file named something like `macos_inhibitor` and then in the `lib.rs` you do something like
```
#[cfg(target_os = "macos")]
pub mod macos_inhibitor;
```
(I let you adapt the names)

codex-rs/tui/src/chatwidget.rs
```diff
@@ -1274,8 +1276,16 @@ impl ChatWidget {
 
     // Raw reasoning uses the same flow as summarized reasoning
 
+    #[cfg(test)]
     fn on_task_started(&mut self) {
+        self.on_task_started_with_source(false);
+    }
+
+    fn on_task_started_with_source(&mut self, from_replay: bool) {
```
jif-oai: This sounds unrelated... how the `from_replay` can have an impact here if `set_turn_running` is idempotent?

codex-rs/core/src/features.rs
yvolovich-cyber: Good call. Updated this to use `if cfg!(target_os = "macos")` for the feature stage (matching the UnifiedExec pattern), so it compiles cleanly across targets.

codex-rs/tui/src/chatwidget.rs
yvolovich-cyber: Good point. I removed the `from_replay` gate and now always call `set_turn_running(true)` on `TurnStarted`; since it is idempotent, replay no longer changes behavior here.

codex-rs/core/src/sleep_inhibitor.rs
yvolovich-cyber: Implemented this structure in the new crate: macOS-specific internals now live in a dedicated `macos_inhibitor` module, and the cross-platform `SleepInhibitor` API is exposed from crate `lib.rs`.

codex-rs/core/src/sleep_inhibitor.rs
yvolovich-cyber: Done. I extracted the implementation into a reusable crate at `codex-rs/utils/sleep-inhibitor` (`codex-utils-sleep-inhibitor`) and moved the sleep inhibition logic there.

codex-rs/core/src/features.rs
yvolovich-cyber: Agreed, I removed this test to keep the suite lean.

jif-oai (approved):
lgtm

PR #11756, opened 2026-02-13
codex-rs: fix thread resume rejoin semantics

maxj-oai:
## Summary
- always rejoin an in-memory running thread on `thread/resume`, even when overrides are present
- reject `thread/resume` when `history` is provided for a running thread
- reject `thread/resume` when `path` mismatches the running thread rollout path
- warn (but do not fail) on override mismatches for running threads
- add more `thread_resume` integration tests and fixes; including restart-based resume-with-overrides coverage

## Validation
- `just fmt`
- `cargo test -p codex-app-server --test all thread_resume`
- manual test with app-server-test-client  https://github.com/openai/codex/pull/11755
- manual test both stdio and websocket in app


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2689,6 +2689,176 @@ impl CodexMessageProcessor {
     }
 
     async fn thread_resume(&mut self, request_id: ConnectionRequestId, params: ThreadResumeParams) {
+        if let Ok(existing_thread_id) = ThreadId::from_string(&params.thread_id)
```
owenlin0, line 2851: `thread_resume` is getting super monolithic, can we refactor? I think we could split into:
- resume_running_thread (the new stuff you're adding where you have a bunch more validations)
- resume_thread_from_history
- resume_thread_from_rollout

and have thread_resume just route accordingly

there's also some duplicate code around loading rollouts which can also be a helper

owenlin0 (commented):
overall seems fine, but would love to clean up the code a bit to make it more modular/readable

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2689,6 +2689,176 @@ impl CodexMessageProcessor {
...
+            && let Ok(existing_thread) = self.thread_manager.get_thread(existing_thread_id).await
+        {
```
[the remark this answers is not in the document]
maxj-oai: will do in a follow-up

codex-rs/app-server/src/codex_message_processor.rs
maxj-oai: done

owenlin0 (approved):

PR #11776, opened 2026-02-13
app-server: improve thread resume rejoin flow

maxj-oai:
thread/resume response includes latest turn with all items, in band so no events are stale or lost

Testing
- e2e tested using app-server-test-client using flow described in "Testing Thread Rejoin Behavior" in codex-rs/app-server-test-client/README.md
- e2e tested in codex desktop by reconnecting to a running turn


codex-rs/app-server/src/thread_state.rs
```diff
@@ -27,9 +38,11 @@ pub(crate) struct TurnSummary {
 pub(crate) struct ThreadState {
     pub(crate) pending_interrupts: PendingInterruptQueue,
     pub(crate) pending_rollbacks: Option<ConnectionRequestId>,
+    pub(crate) active_turn_state_rollout_items: Vec<RolloutItem>,
```
owenlin0: Can we store app-server ThreadItems instead? That's the public interface for clients using app-server API

codex-rs/rmcp-client/src/auth_status.rs
```diff
@@ -33,12 +34,15 @@ pub async fn determine_streamable_http_auth_status(
         return Ok(McpAuthStatus::BearerToken);
     }
 
+    let default_headers = build_default_headers(http_headers, env_http_headers)?;
+    if default_headers.contains_key(AUTHORIZATION) {
+        return Ok(McpAuthStatus::BearerToken);
+    }
```
owenlin0, line 40: what's this? there's a few auth / rmcp changes in this PR

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -5642,22 +5660,53 @@ impl CodexMessageProcessor {
...
             let mut thread_state = thread_state.lock().await;
             if thread_state.listener_matches(&conversation) {
                 return;
             }
-            thread_state.set_listener(cancel_tx, &conversation);
+            thread_state.set_listener(cancel_tx, listener_command_tx, &conversation);
         }
         let outgoing_for_task = self.outgoing.clone();
         let fallback_model_provider = self.config.model_provider_id.clone();
+        let record_active_turn_state_in_memory = !self.single_client_mode;
```
owenlin0: nice!

codex-rs/rmcp-client/src/auth_status.rs
maxj-oai: oops rolled those in by mistake - from https://github.com/openai/codex/pull/11692
will remove

codex-rs/app-server/src/thread_state.rs
maxj-oai: goal is to recap the event stream from current turn; ThreadItems are missing a lot of types (ex EventMsg::ItemStarted)
should we extend ThreadItems to support all event types (not just the persisted ones)? or reapply the transform to ServerNotification etc from apply_bespoke_event_handling?

owenlin0: hmm, it's the same as what the client would get in `thread.turns` on resume though

i think it's fine if we don't have 100% event coverage, as it's the items that the clients care about 99% of the time. there are some one-off ones like `turn/diff/updated`, `turn/plan/updated`, `account/rateLimits/updated` that the client won't have access to via ThreadItems, but once the next notification is emitted for each of those, they can catch up then

re: `EventMsg::ItemStarted`, the client doesn't really need the `item/started` notification since the actual item was returned

maxj-oai: OK - I'll try it, we can add more in the future if needed
For the sake of simplicity then I will also remove the new field and simply add these items to the thread.turns

owenlin0: sounds good!

maxj-oai: done

codex-rs/app-server-protocol/src/protocol/thread_history.rs
```diff
@@ -241,33 +343,20 @@ impl ThreadHistoryBuilder {
...
             duration_ms: Some(duration_ms),
         };
+        self.upsert_item_in_turn_id(&payload.turn_id, item);
+    }
 
-        // Command completions can arrive out of order. Unified exec may return
-        // while a PTY is still running, then emit ExecCommandEnd later from a
-        // background exit watcher when that process finally exits. By then, a
-        // newer user turn may already have started. Route by event turn_id so
-        // replay preserves the original turn association.
```
owenlin0, line 254: did we intentionally or unintentionally drop this? this sounded useful

codex-rs/app-server-protocol/src/protocol/thread_history.rs
```diff
@@ -241,33 +343,20 @@ impl ThreadHistoryBuilder {
...
-        if let Some(turn) = self.current_turn.as_mut()
-            && turn.id == payload.turn_id
-        {
-            turn.items.push(item);
-            return;
-        }
-
-        // If the originating turn is already finalized, append there instead
-        // of attaching to whichever turn is currently active during replay.
```
owenlin0, line 263: here too?

owenlin0 (commented):
small comments but hopefully @celia-oai can take a closer look

codex-rs/app-server-protocol/src/protocol/thread_history.rs
maxj-oai: no this is still the case, the logic has just moved. Upsert_item_in_turn_id will attach it in the appropriate turn when the turn_id field is set

codex-rs/app-server-protocol/src/protocol/thread_history.rs
maxj-oai: same

codex-rs/app-server-protocol/src/protocol/thread_history.rs
maxj-oai: I readded the comment for clarity

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -182,7 +180,7 @@ pub(crate) async fn apply_bespoke_event_handling(
                     // Until we migrate the core to be aware of a first class FileChangeItem
                     // and emit the corresponding EventMsg, we repurpose the call_id as the item_id.
                     let item_id = call_id.clone();
-                    let patch_changes = convert_patch_changes(&changes);
+                    let patch_changes = codex_app_server_protocol::convert_patch_changes(&changes);
```
celia-oai: can we import at the start of the file for this

codex-rs/app-server/src/lib.rs
```diff
@@ -238,7 +238,8 @@ pub async fn run_main_with_transport(
                 Some(start_websocket_acceptor(bind_address, transport_event_tx.clone()).await?);
         }
     }
-    let shutdown_when_no_connections = matches!(transport, AppServerTransport::Stdio);
+    let single_client_mode = matches!(transport, AppServerTransport::Stdio);
```
celia-oai, line 241: I wonder if this should be a feature flag, something like `single_client_mode = matches!(transport, AppServerTransport::Stdio) or single_client_mode_feature_flag_on`

codex-rs/app-server-protocol/src/protocol/thread_history.rs
```diff
@@ -639,6 +830,30 @@ impl ThreadHistoryBuilder {
...
+            upsert_turn_item(&mut turn.items, item);
+            return;
+        }
+
+        if let Some(turn) = self.turns.iter_mut().find(|turn| turn.id == turn_id) {
+            upsert_turn_item(&mut turn.items, item);
+            return;
+        }
+
+        eprintln!(
```
celia-oai: should this be a warning log instead of printing

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2707,6 +2707,26 @@ pub enum ThreadItem {
     ContextCompaction { id: String },
 }
 
+impl ThreadItem {
+    pub fn item_id(&self) -> &str {
```
celia-oai: nit: rename to `id` maybe? `item.item_id()` reads a little redundant

celia-oai (approved):

codex-rs/app-server/src/lib.rs
maxj-oai: I don't think so; this is already user-controlled via the `--listen` cli argument - using stdio:// (the default) necessarily implies single-client mode (and shut down when no connections) vs a connection-oriented transport which supports multiple clients and should survive disconnect

PR #11780, opened 2026-02-13
chore(tui): reduce noisy key logging

apanasenko-oai:
### Why
The previous `codex-tui` unhandled key logging path was too noisy during day-to-day development while still retaining little operational value.

### What changed
- Removed the `debug-logs` feature from `codex-rs/tui/Cargo.toml`.
- Changed unhandled key event logging in `codex-rs/tui/src/bottom_pane/textarea.rs` from `tracing::debug!` behind `#[cfg(feature = "debug-logs")]` to `tracing::trace!` behind `#[cfg(debug_assertions)]`.
- Kept the signal available for deep debugging without exposing it in normal workflows.

### Verification
- Confirmed no `debug-logs` references remain in `codex-rs/tui`.

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -223,6 +223,7 @@ use crate::tasks::SessionTask;
 use crate::tasks::SessionTaskContext;
 use crate::tools::ToolRouter;
 use crate::tools::context::SharedTurnDiffTracker;
+use crate::tools::handlers::SEARCH_TOOL_BM25_TOOL_NAME;
 use crate::tools::js_repl::JsReplHandle;
 use crate::tools::network_approval::NetworkApprovalService;
 use crate::tools::network_approval::build_blocked_request_observer;
@@ -1672,7 +1673,7 @@ impl Session {
             };
             match response_item {
                 ResponseItem::FunctionCall { name, call_id, .. } => {
-                    if name == "search_tool_bm25" {
+                    if name == SEARCH_TOOL_BM25_TOOL_NAME {
                         search_call_ids.insert(call_id.clone());
                     }
                 }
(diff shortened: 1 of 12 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -1595,6 +1631,54 @@ impl Session {
...
+        let mut search_call_ids = HashSet::new();
+        let mut active_selected_tools: Option<Vec<String>> = None;
+
+        for item in rollout_items {
+            let RolloutItem::ResponseItem(response_item) = item else {
+                continue;
+            };
+            match response_item {
+                ResponseItem::FunctionCall { name, call_id, .. } => {
+                    if name == "search_tool_bm25" {
```
mzeng-openai: Use a const here?

apanasenko-oai: good call, thanks!

codex-rs/core/templates/search_tool/tool_description.md
mzeng-openai, line 3: Feels like BM25 is unnecessary detail to the model.

mzeng-openai: We shouldn't match against connector_id since the query is just plain text right?

mzeng-openai: Should we bring this upfront so it doesn't get buried in the context.

mzeng-openai: I'd probably rephrase this to be something like this to make it clear to the model:
```md
MCP tools of the apps below are hidden until you search for them with this tool:
{{app_names}}
```

mzeng-openai: Also do not search for anything that's not related to the apps listed above.

mzeng-openai (approved):

codex-rs/core/templates/search_tool/tool_description.md
apanasenko-oai: it give the model a sense what type of search it should perform.

PR #11783, opened 2026-02-13
[api] add service-tier

kaih-oai:
* adds service tier parameter for api requests.
* This allows users to configure usage of tiers like "flex" for better behavior wrt rate limits


codex-rs/core/src/config/profile.rs
```diff
@@ -26,6 +27,7 @@ pub struct ConfigProfile {
     pub model_reasoning_effort: Option<ReasoningEffort>,
     pub model_reasoning_summary: Option<ReasoningSummary>,
     pub model_verbosity: Option<Verbosity>,
+    pub model_service_tier: Option<ServiceTier>,
```
pakrym-oai, line 30: why only part of profile and not a main level prop?

pakrym-oai (changes requested):
Looks good, needs at least one integration test - using test_codex with `.with_config` that asserts the final request.

PR #11786, opened 2026-02-13
app-server: expose loaded thread status via read/list and notifications

euroelessar:
Motivation
- Today, a newly connected client has no direct way to determine the current runtime status of threads from read/list responses alone.
- This forces clients to infer state from transient events, which can lead to stale or inconsistent UI when reconnecting or attaching late.

Changes
- Add `status` to `thread/read` responses.
- Add `statuses` to `thread/list` responses.
- Emit `thread/status/changed` notifications with `threadId` and the new status.
- Track runtime status for all loaded threads and default unknown threads to `idle`.
- Update protocol/docs/tests/schema fixtures for the revised API.

Testing
- Validated protocol API changes with automated protocol tests and regenerated schema/type fixtures.
- Validated app-server behavior with unit and integration test suites, including status transitions and notifications.


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -5808,6 +5891,52 @@ impl CodexMessageProcessor {
...
+                                &[connection_id],
+                                ServerNotification::ThreadWatchUpdated(update),
+                            )
+                            .await;
+                    }
+                    Err(tokio::sync::broadcast::error::RecvError::Lagged(_)) => {
+                        let version = thread_watch_manager.current_version().await;
+                        let update = ThreadWatchUpdatedNotification {
+                            version,
+                            update: ThreadWatchUpdate::ResyncRequired,
```
maxj-oai: should we just kill the connection in this case? then client will reconnect and rerun thread-watch naturally without having to add a special case for this

we should probably do this for thread notification broadcasts as well; rn it is synchronous and a slow connection could cause backups

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1836,6 +1837,51 @@ pub struct ThreadLoadedListResponse {
     pub next_cursor: Option<String>,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct LoadedThreadEntry {
+    pub thread: Thread,
+    pub status: LoadedThreadStatus,
```
owenlin0: Could we just add this `status` field to the Thread object? We can then have a `NotLoaded` state for the ones that have not been resumed.

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1836,6 +1837,51 @@ pub struct ThreadLoadedListResponse {
...
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(tag = "type", rename_all = "camelCase")]
+#[ts(tag = "type")]
+#[ts(export_to = "v2/")]
+pub enum LoadedThreadStatus {
```
owenlin0: And then we can call it just `ThreadStatus`

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1836,6 +1837,51 @@ pub struct ThreadLoadedListResponse {
...
+    },
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub enum ThreadActiveFlag {
+    Running,
+    WaitingPermission,
+    WaitingUserInput,
```
owenlin0: isn't status `Active` + `waiting user input` the same as `Idle`?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1836,6 +1837,51 @@ pub struct ThreadLoadedListResponse {
...
+        outcome: ThreadTerminalOutcome,
+    },
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub enum ThreadActiveFlag {
+    Running,
+    WaitingPermission,
```
owenlin0: nit: `WaitingOnApproval`?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1836,6 +1837,51 @@ pub struct ThreadLoadedListResponse {
...
+    Idle,
+    #[serde(rename_all = "camelCase")]
+    #[ts(rename_all = "camelCase")]
+    Active {
+        active_flags: Vec<ThreadActiveFlag>,
+    },
+    #[serde(rename_all = "camelCase")]
+    #[ts(rename_all = "camelCase")]
+    Terminal {
+        outcome: ThreadTerminalOutcome,
```
owenlin0: do we need `Terminal` + `outcome`? This info would be stored in the thread's last turn's status (either through the `turn/completed` event for a live thread, or when resuming a thread and accessing the thread's `turns[-1].status`?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1836,6 +1837,51 @@ pub struct ThreadLoadedListResponse {
...
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub enum ThreadTerminalOutcome {
+    Completed,
+    Failed,
+    Interrupted,
+    Shutdown,
```
owenlin0: `status: terminal` + `Shutdown` == `status: NotLoaded`, right?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1808,6 +1808,7 @@ pub enum ThreadSortKey {
 #[ts(export_to = "v2/")]
 pub struct ThreadListResponse {
     pub data: Vec<Thread>,
+    pub statuses: HashMap<String, LoadedThreadStatus>,
```
owenlin0: we can remove this field if we put `status` on the `Thread` object

we could also add `status` (or `statuses`?) as a filter field for `thread/list` so the client can choose to make two calls to render the UI:
- list all active threads
- list all non-active (aka not loaded) threads

thoughts? feels cleaner

codex-rs/app-server-protocol/src/protocol/v2.rs
euroelessar: We could, in that case we probably would need to add proper value for `status` field in all the existing APIs which currently already return Thread object. Like `thread/resume`, `thread/started`, and others. What do you think?

owenlin0: yeah, i'd be fine with that!

codex-rs/app-server-protocol/src/protocol/v2.rs
euroelessar: not exactly, in case of `Active` there is an active turn, it's just being blocked on a user input
e.g. in planning mode model may be actively requesting feedback from the user

codex-rs/app-server-protocol/src/protocol/v2.rs
euroelessar: we want to differentiate idle vs failed statuses in order to render the list of threads correctly (e.g. in codex app we have a separate indicator for this case)

I may push it into single `Idle` state, and tell the client about last turn being interrupted vs failed as flags (similarly to how it's done for `Active` state)

codex-rs/app-server-protocol/src/protocol/v2.rs
owenlin0: fair point

codex-rs/app-server-protocol/src/protocol/v2.rs
owenlin0: ok sounds good

codex-rs/app-server-protocol/src/protocol/v2.rs
euroelessar: done!

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1836,6 +1836,40 @@ pub struct ThreadLoadedListResponse {
...
+    NotLoaded,
+    #[serde(rename_all = "camelCase")]
+    #[ts(rename_all = "camelCase")]
+    Idle {
+        idle_flags: Vec<ThreadIdleFlag>,
+    },
+    #[serde(rename_all = "camelCase")]
+    #[ts(rename_all = "camelCase")]
+    Active {
+        active_flags: Vec<ThreadActiveFlag>,
```
owenlin0, line 1851: it's not possible to be both `WaitingOnApproval` AND `WaitingOnUserInput` is it? is it possible to flatten the status to be
```
NotLoaded
Idle
SystemError
Running
WaitingOnApproval
WaitingOnUserInput
```
?

also, do we need Idle + system_error? What's that scenario?

euroelessar: hm, I've read more, and I think it's indeed not possible, but only due to user input tool spec not being marked as parallel
do we want to provide the guarantee that "model is not allowed to continue executing commands to research further while user is thinking on the response" forever? or how do we make that change in backward-compatible way?

on the other note, I'm okay to decouple SystemError from Idle, it's pretty self-container and is auto-cleared on status change anyway

owenlin0 (approved):

maxj-oai (approved):

PR #11787, opened 2026-02-13
Adjust MCP tool approval handling for custom servers

colby-oai:
Summary
This PR expands MCP client-side approval behavior beyond codex_apps and tightens elicitation capability signaling.

- Removed the codex_apps-only gate in MCP tool approval checks, so local/custom MCP servers are now eligible for the same client-side approval prompt flow when tool annotations indicate side effects.
- Updated approval memory keying to support tools without a connector ID (connector_id: Option<String>), allowing “Approve this Session” to be remembered even when connector metadata is missing.
- Updated prompt text for non-codex_apps tools to identify origin as The <server> MCP server instead of This app.
- Added MCP initialization capability policy so only codex_apps advertises MCP elicitation capability; other servers advertise no elicitation support.
- Added regression tests for:
server-specific prompt copy behavior
codex-apps-only elicitation capability advertisement

Testing
- Not run (not requested)


codex-rs/core/src/mcp_connection_manager.rs
```diff
@@ -1006,6 +1006,21 @@ impl From<anyhow::Error> for StartupOutcomeError {
...
+        // https://modelcontextprotocol.io/specification/2025-06-18/client/elicitation#capabilities
+        // indicates this should be an empty object.
+        Some(ElicitationCapability {
+            form: Some(FormElicitationCapability {
+                schema_validation: None,
+            }),
+            url: None,
+        })
+    } else {
+        None
```
mzeng-openai, line 1020: Why dropping elicitation for custom MCPs? I do think we have support for custom MCPs that @nornagon-openai built.

colby-oai: @mzeng-openai we do have support for server-driven elicitation for custom MCPs, yes. However we essentially need to make a choice between:
- rely on custom/random mcp servers to elicit at proper times
- rely on client-side annotations (+ CAM and other controls in the near future) 

Otherwise, we risk the bad UX of double approvals. The choice we felt was better here was to not advertise that we support elicitation **for custom mcps**  and instead look at annotations.

mzeng-openai: Can we restrict this to custom MCPs only? I don't think we want to just use clientside approval and bypass serverside elicitation all at once for Apps, since we have CAM and all other serverside monitors.

colby-oai: @mzeng-openai 
Absolutely! The only changes in this PR should be related to custom MCPs. Nothing in this PR touches how codex apps work

colby-oai: My understanding is *currently* (before this PR), codex apps are also doing client-side annotation checks. I understand they won't/shouldn't in the very near future as we use connector gateway. 

My change in this pr though, is simply extending the annotation checks that exist for codex apps to apply to custom MCPs as well

mzeng-openai (approved):

PR #11793, opened 2026-02-13
fix: send unfiltered models over model/list

sayan-oai:
### What
to unblock filtering models in VSCE, change `model/list` app-server endpoint to send all models + visibility field `showInPicker` so filtering can be done in VSCE if desired.

### Tests
Updated tests.

### DIFF
--- a/codex-rs/app-server-protocol/src/protocol/v2.rs
+++ b/codex-rs/app-server-protocol/src/protocol/v2.rs
@@ -1124,7 +1124,7 @@ pub struct Model {
     pub upgrade: Option<String>,
     pub display_name: String,
     pub description: String,
-    pub show_in_picker: bool,
+    pub hidden: bool,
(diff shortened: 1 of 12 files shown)


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1121,6 +1124,7 @@ pub struct Model {
     pub upgrade: Option<String>,
     pub display_name: String,
     pub description: String,
+    pub show_in_picker: bool,
```
owenlin0: maybe `hidden: bool` just to match `include_hidden`?

owenlin0 (approved):

codex-rs/app-server-protocol/src/protocol/v2.rs
sayan-oai: sounds good

PR #11798, opened 2026-02-14
fix: Restricted Read: /System is too permissive for macOS platform de…

leoshimo-oai:
…fault

Update the list of platform defaults included for `ReadOnlyAccess`.

When `ReadOnlyAccess::Restricted::include_platform_defaults` is `true`, the policy defined in `codex-rs/core/src/seatbelt_platform_defaults.sbpl` is appended to enable macOS programs to function properly.


viyatb-oai (approved):

leoshimo-oai:
recheck

owenlin0 (approved):

PR #11811, opened 2026-02-14
fix update info for brew

xl-openai:
Check /opt/homebrew first.

### DIFF
--- a/codex-rs/tui/src/update_action.rs
+++ b/codex-rs/tui/src/update_action.rs
@@ -48,14 +48,14 @@ fn detect_update_action(
     managed_by_npm: bool,
     managed_by_bun: bool,
 ) -> Option<UpdateAction> {
-    if is_macos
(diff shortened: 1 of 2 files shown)


joshka-oai (approved):

codex-rs/tui/src/update_action.rs
```diff
@@ -48,13 +48,13 @@ fn detect_update_action(
     managed_by_npm: bool,
     managed_by_bun: bool,
 ) -> Option<UpdateAction> {
-    if managed_by_npm {
+    if is_macos && current_exe.starts_with("/opt/homebrew") {
```
bolinfest, line 51: If the user has installed `npm` via brew, then I believe `npm install -g @openai/codex` will put it in `/opt/homebrew/lib/node_modules`, so it is not correct to change this.

bolinfest (changes requested):
As written, this appears to introduce a regression.

PR #11825, opened 2026-02-14
fix: use dash-separated HTTP header name for session ID (#11732)

### LINKED ISSUE #11732: Use dash-separated header session-id instead of session_id
## What version of Codex is running?
openai/codex current main in v0.101.0.

## What issue are you seeing?
The Codex client sends requests to our internal Envoy-based model gateway. Our gateway rejects HTTP headers that contain underscores (_) `session_id`.

This behavior is intentional on our side and aligns with common HTTP best practices. Underscores in HTTP header names are discouraged due to interoperability issues with common web servers and proxies, which often ignore or drop such headers by default.

As a result, requests from the Codex client fail when underscore-containing headers are present.

## Why this matters

While we can relax the validation rule in our gateway, we prefer not to, since:

- It aligns with widely adopted HTTP conventions.
- It avoids subtle interoperability issues across proxies and infrastructure components.
- It maintains consistency with common production hardening practices.

## Request
Would it be possible for the Codex client to switch from underscore-separated header names to dash-separated header names (e.g., x-foo-bar instead of x_foo_bar)?

Using dash-separated headers would improve compatibility with common HTTP infrastructure without requiring downstream gateways to weaken validation rules.

echoVic:
## Summary

Replace underscore-separated `session_id` HTTP header with dash-separated `session-id` to comply with HTTP best practices.

## Problem

The Codex client sends a `session_id` HTTP header in API requests. Many proxies and gateways (e.g., Envoy, nginx) reject or silently drop headers containing underscores by default, as this aligns with [RFC 7230](https://tools.ietf.org/html/rfc7230) conventions and common production hardening practices.

This causes requests to fail when routed through infrastructure that enforces strict header name validation.

## Fix

- Rename `session_id` → `session-id` in `codex-rs/codex-api/src/requests/headers.rs`
- Update corresponding test assertions in `codex-rs/codex-api/tests/clients.rs` and `codex-rs/core/tests/suite/client.rs`

## Impact

This is a breaking change for any server-side code that reads the `session_id` header by name. The OpenAI backend should be updated to accept `session-id` (or both) before merging.

Fixes #11732

### DIFF
--- a/codex-rs/codex-api/src/requests/headers.rs
+++ b/codex-rs/codex-api/src/requests/headers.rs
@@ -5,7 +5,7 @@ use http::HeaderValue;
 pub fn build_conversation_headers(conversation_id: Option<String>) -> HeaderMap {
     let mut headers = HeaderMap::new();
     if let Some(id) = conversation_id {
-        insert_header(&mut headers, "session_id", &id);
+        insert_header(&mut headers, "session-id", &id);
     }
     headers
 }
--- a/codex-rs/codex-api/tests/clients.rs
+++ b/codex-rs/codex-api/tests/clients.rs
@@ -330,7 +330,7 @@ async fn azure_default_store_attaches_ids_and_headers() -> Result<()> {
     let req = &requests[0];
 
     assert_eq!(
-        req.headers.get("session_id").and_then(|v| v.to_str().ok()),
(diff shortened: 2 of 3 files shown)


etraut-openai:

PR #11841, opened 2026-02-15
fix(core): preserve non-app MCP tools in apps mode

karel-openai:
## Summary
This fixes a regression where enabling `Feature::Apps` could hide non-`codex_apps` MCP tools (e.g. Notion MCP write tools) from the model tool list.

In `built_tools`, the apps path rebuilt `mcp_tools` from:
- selected MCP tools (if any), plus
- filtered `codex_apps` tools.

When there was no MCP tool selection yet, this effectively dropped non-`codex_apps` servers from model-visible tools.

## Root Cause
`built_tools` applied apps filtering by constructing a reduced MCP map and assigning it back to `mcp_tools`.
That reduced map only guaranteed `codex_apps` tool inclusion, so tools from other MCP servers could disappear even though those servers were started and healthy.

## Fix
- Introduce `merge_apps_mode_mcp_tools(...)` to merge app-filtered tools with selected tools.
- Preserve non-`codex_apps` MCP tools when there is no explicit MCP tool selection.
- Preserve explicit selection behavior when selection is present.

## Tests
Added unit tests in `codex-rs/core/src/codex.rs`:
- `apps_mode_keeps_non_codex_apps_mcp_tools_without_selection`
- `apps_mode_respects_explicit_tool_selection`

Ran:
- `cargo test -p codex-core apps_mode_keeps_non_codex_apps_mcp_tools_without_selection -- --nocapture`
- `cargo test -p codex-core apps_mode_respects_explicit_tool_selection -- --nocapture`

## User Impact
With apps enabled, MCP tools from non-app servers (including Notion MCP write tools) remain callable unless explicitly filtered by selection semantics.

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -4722,6 +4722,25 @@ fn filter_codex_apps_mcp_tools(
         .collect()
 }
 
+fn merge_apps_mode_mcp_tools(
+    mcp_tools: &HashMap<String, crate::mcp_connection_manager::ToolInfo>,
+    mut selected_mcp_tools: HashMap<String, crate::mcp_connection_manager::ToolInfo>,
+    apps_mcp_tools: HashMap<String, crate::mcp_connection_manager::ToolInfo>,
+    preserve_non_codex_apps_tools: bool,
+) -> HashMap<String, crate::mcp_connection_manager::ToolInfo> {
+    if preserve_non_codex_apps_tools {
+        // Keep non-codex_apps MCP tools available when there is no explicit MCP tool selection.
+        for (name, tool) in mcp_tools {
+            if tool.server_name != CODEX_APPS_MCP_SERVER_NAME {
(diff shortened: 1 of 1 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -4902,18 +4921,21 @@ async fn built_tools(
...
         let apps_mcp_tools =
             filter_codex_apps_mcp_tools_only(&mcp_tools, explicitly_enabled.as_ref());
-        selected_mcp_tools.extend(apps_mcp_tools);
-
-        mcp_tools = selected_mcp_tools;
+        mcp_tools = merge_apps_mode_mcp_tools(
+            &mcp_tools,
+            selected_mcp_tools,
+            apps_mcp_tools,
+            selected_tool_names.is_none(),
```
mzeng-openai, line 4937: I think we should include non-apps MCP tools regardless of the current selection.

Issue #11856, opened 2026-02-15
Add toggle/config for CLI spinner (accessibility)


BlindMaster24:
### What variant of Codex are you using?

CLI

### What feature would you like to see?

Please add a setting to enable/disable or customize the spinner in Codex CLI.
I am blind / visually impaired, and constant spinner animation can make screen-reader use harder.
It would be great to have an accessibility option (for example: spinner on/off, reduced updates, or a simpler mode).

### Additional information

Default behavior can stay the same for sighted users, but an explicit accessibility toggle would help a lot.

PR #11860, opened 2026-02-15
Fixed screen reader regression in CLI

etraut-openai:
The `tui.animations` switch should gate all animations in the TUI, but a recent change introduced a regression that didn't include the gate. This makes it difficult to use the TUI with a screen reader.

This fix addresses #11856


edward-bayes (approved):
lgtm!

PR #11865, opened 2026-02-15
feat(tui) Vim mode

dylan-hurd-oai:
## Summary
Some of us really like Vim bindings when working with prompts! This doesn't support full capabilities Couple of notes on implementation:

* `/vim` to toggle
* No support for Visual mode _yet_, just Normal with navigation + Insert
* `Esc` in Insert mode switches to Normal mode, then `Esc`-`Esc` to edit previous message

## Testing
- [x] Updated existing snapshots
- [x] Added some functional tests


codex-rs/tui/src/bottom_pane/snapshots/codex_tui__bottom_pane__chat_composer__tests__footer_mode_ctrl_c_then_esc_hint.snap
```diff
@@ -10,4 +10,4 @@ expression: terminal.backend()
 "                                                                                                    "
 "                                                                                                    "
 "                                                                                                    "
-"  esc esc to edit previous message                                                                  "
+"  esc esc to edit previous message                                               100% context left  "
```
joshka-oai, line 13: Is this a `/statusline` thing showing up or an intentional unrelated change?

codex-rs/tui/src/bottom_pane/chat_composer.rs
```diff
@@ -3557,20 +3618,16 @@ impl ChatComposer {
                     )
                 };
                 let right_line = if status_line_active {
-                    let full =
-                        mode_indicator_line(self.collaboration_mode_indicator, show_cycle_hint);
-                    let compact = mode_indicator_line(self.collaboration_mode_indicator, false);
+                    let full = self.mode_indicator_line(show_cycle_hint);
+                    let compact = self.mode_indicator_line(false);
```
joshka-oai, line 3622: This could do with some comments / better names - looking at this diff and this code for the first time, I have no idea what these variables control and hence what the impact of this change is.

docs/tui-chat-composer.md
```diff
@@ -38,6 +38,12 @@ The solution is to detect paste-like _bursts_ and buffer them into a single expl
 2. **Paste burst**: transient detection state for non-bracketed paste.
    - implemented by `PasteBurst`
 
+`ChatComposer` also coordinates the textarea input mode:
+
+- **Vim mode**: modal editing (normal/insert). While in Vim normal mode, paste-burst detection
+  is disabled so rapid command keystrokes are never buffered as paste.
+- Slash command `/vim` enables vim mode in the TUI and the public composer widget.
```
joshka-oai, line 45: I'm curious, why a slash command for this rather than a feature flag?
Neither is wrong here, I'm mostly thinking that you might want people to enable this for all their sessions generally (or `--enable vim` for per session use)

codex-rs/tui/src/bottom_pane/textarea.rs
```diff
@@ -112,6 +130,38 @@ impl TextArea {
...
+
+    pub(crate) fn vim_mode_label(&self) -> Option<&'static str> {
+        if !self.vim_enabled {
+            return None;
+        }
+        Some(match self.vim_mode {
+            VimMode::Normal => "Normal",
+            VimMode::Insert => "Insert",
+        })
+    }
```
joshka-oai, line 163: This just gets called then translated to vim:normal / vim:insert- can it instead just pass through the enum to the calling code (vim_mode_indicator_span above) and convert there?

codex-rs/tui/src/bottom_pane/textarea.rs
joshka-oai, line 1: nit: I'd personally suggest moving most of the additive code for vim mode into a small module called from this to have it more self contained (easier to have tests that are close to the file). Codex is fairly good at fixing this up.

codex-rs/tui/src/bottom_pane/textarea.rs
```diff
@@ -1462,6 +1907,32 @@ mod tests {
         assert_eq!(t.cursor(), elem_start);
     }
 
+    #[test]
```
joshka-oai, line 1910: Lots more tests on this (probably) - guessing codex can generate those.

codex-rs/tui/src/slash_command.rs
```diff
@@ -84,8 +85,9 @@ impl SlashCommand {
             SlashCommand::Plan => "switch to Plan mode",
             SlashCommand::Collab => "change collaboration mode (experimental)",
             SlashCommand::Agent => "switch the active agent thread",
-            SlashCommand::Approvals => "choose what Codex is allowed to do",
+            SlashCommand::Approvals => "choose what Codex can do without approval",
```
joshka-oai, line 88: Probably out of scope for this change?

joshka-oai (approved):
General approval to the shape of this. The biggest push back is that there are places beyond the prompt area where we might want vim mode too. If we only handle that here, those other places might feel weird (e.g. additional info when answering a plan question).

- biggest question on design is /vim vs vim feature flag. I'd think a feature flag would be more useful on this unless there's a common need to enable / disable this at runtime.
- needs more tests on the actual vim movement code etc.
- refactor the additive parts of the new vim code into a small module (makes it easier to see the tests with the code, smaller files, smaller cognitive load, etc. - avoids the tendency towards god classes)
- running the doc pass skill I mentioned privately over this adds a few nice to have comments (didn't push this change from my local review copy, 
- I asked codex after the doc pass "How good / right is the vim logic in textarea.rs?" and it came up with a reasonable amount of findings - maybe worth running locally - just small things that would be papercuts for vim users. Worth taking a look at and seeing if there's things you'd fix.

Overall a good baseline that could be shipped and improved on, or held for a bit and improved before shipping.

Note - we've got configurable keybindings coming soon, so it might be worth holding this for a few days and landing this after that to make things a bit easier to coordinate.

PR #11866, opened 2026-02-15
chore(core) rm Feature::RequestRule

dylan-hurd-oai:
## Summary
This feature is now reasonably stable, let's remove it so we can simplify our upcoming iterations here.

## Testing 
- [x] Existing tests pass


codex-rs/core/src/features.rs
```diff
@@ -508,12 +506,6 @@ pub const FEATURES: &[FeatureSpec] = &[
         stage: Stage::UnderDevelopment,
         default_enabled: false,
     },
-    FeatureSpec {
```
jif-oai, line 511: this will break all the config that have the rule. Can you move it to `Stage::Removed` instead?

jif-oai (approved):

codex-rs/core/src/features.rs
dylan-hurd-oai: ah right - will do!

Issue #11298, opened 2026-02-10
Keeps asking for permission to run commands


BlueBlazin:
### What version of the Codex App are you using (From “About Codex” dialog)?

260208.1016 (571)

### What subscription do you have?

Pro

### What issue are you seeing?

The "Yes, and don't ask again for commands that start with" is broken.

### What steps can reproduce the bug?

Get Codex to run a command. Click "Yes, and don't ask again for commands that start with". Get Codex to run same command again.

### What is the expected behavior?

Subsequent commands should not require permission.

### Additional information

My recommendation is to rename Codex to 'Codex (Supervised)' until this is fixed.

etraut-openai:
Can you provide a more specific example of a series of commands where you saw this occur? It's likely that the commands differ slightly, so the approval engine categorized them as distinct commands.

BlueBlazin:
@etraut-openai I think they are the same commands (but I will verify and get back to you), however they all have some other stuff before them (`FOO=1 bar --baz`). Could it be that the permission UI is correctly parsing just the command but when that check happens the next time it tries matching the full command string from the beginning?

etraut-openai:
Yeah, that's likely the cause. Specific examples would be useful when you have time. Thanks!

BlueBlazin:
@etraut-openai Yep, just been collecting a few. All of these prompted permission requests (for `cargo flamegraph`):

```
mkdir -p perf && CARGO_PROFILE_RELEASE_DEBUG=true cargo flamegraph --bin foo --output perf/a.svg -- -S -c "input1"
mkdir -p perf && CARGO_PROFILE_RELEASE_DEBUG=true cargo flamegraph --bin foo --output perf/b.svg -- -S -c "input2"
mkdir -p perf && CARGO_PROFILE_RELEASE_DEBUG=true cargo flamegraph --bin foo --output perf/c.svg -- -c "input3"
mkdir -p perf && CARGO_PROFILE_RELEASE_DEBUG=true cargo flamegraph --bin foo --output perf/d.svg -- -S -c "input4"
```

etraut-openai:
Thanks for the example. The command parser currently doesn't handle commands with an environment variable assignment prefix. That explains why the calls in your example above are not exempted by the "Don't ask me again" logic. We'll need to think about this case more to figure out the best way to handle it. Thanks again for the examples — very helpful!

almuqrin:
Hello everyone, can you please check this https://github.com/openai/codex/pull/11868

[...]

PR #11868, opened 2026-02-15
fix(core): execpolicy matching for redirected shell scripts

### LINKED ISSUE #10321: Asks for permission to run command `xcodebuild`, even after permitting all `xcodebuild` commands

0.93

Pro

Codex 5.2 (medium reasoning)

MacOS

Warp

I granted permission to run all `xcodebuild` commands by selecting the "Yes, and don't ask again" option from the CLI. Then in future runs, even when running the same command with the same set of arguments, it still asks for sandboxing permissions. This persists even after quitting and restarting the codex CLI. 

I am not running with dangerous permissions. I'm only running `codex` for my conversations.

Image

1. Prompt the agent to run the command `xcodebuild -scheme Hubble -destination 'generic/platform=iOS'
  │ -quiet CODE_SIGNING_ALLOWED=NO build 2>&1`.
2. Notice the agent tries to run this command and receives a sandbox error (seems odd it doesn't ask me and intentionally needs to fail first)?
3. Select option 2, "Yes, and don't ask again"
4. On a future turn, prompt to run the same `xcodebuild ...` command again. The sandbox permissions error reoccurs unexpectedly.

I should not have to grant permissions again for the current session and all future CLI sessions. 

### Additional information

This is my Codex config:

```toml
[projects."/Users/benholmes/Projects/Hubble"]
trust_level = "trusted"

[projects."/Users/benholmes/Projects/simplestack-store"]
trust_level = "trusted"

[projects."/Users/benholmes/Projects/simple-stack"]
trust_level = "trusted"

[features]
unified_exec = true
```

almuqrin:
## Summary
Fix execpolicy matching when a wrapped shell script contains redirection (for example `> /tmp/log` or `2>&1`).

Fixes #10321.

Before this change, commands like `gh run view ... > /tmp/log` or `xcodebuild ... 2>&1` could repeatedly prompt for approval even after selecting "Yes, and don't ask again".

## Problem
Approval persistence relies on extracting inner command prefixes from wrapper invocations such as `zsh -lc "..."`.

For redirected scripts, `core/src/exec_policy.rs` called `parse_shell_lc_plain_commands`, which rejects redirections by design. That forced fallback to the outer wrapper argv (`zsh -lc ...`), so saved prefix rules such as `["gh", "run", "view"]` did not match future runs.

## Root Cause
- `commands_for_exec_policy` used a parser that only accepts "plain" word-only commands.
- Redirected shell scripts were treated as unparseable and matched against wrapper argv instead of the actual inner command.
- Result: approval rules appeared to be "not remembered" for many real commands that include redirection.

## Fix
1. Add execpolicy-focused shell parsing path in `shell-command/src/bash.rs`:
   - `parse_shell_lc_commands_for_exec_policy` accepts redirected statements and command attachments needed for matching (`file_redirect`, `variable_assignment`, comments).
   - It remains fail-closed for unsupported/unsafe constructs (substitutions/control-flow still reject).
2. Switch `core/src/exec_policy.rs` to use `parse_shell_lc_commands_for_exec_policy` in `commands_for_exec_policy`.
3. Add focused regression coverage in core + integration tests.

## Why This Is Safe
- The change only affects command extraction for execpolicy matching.
- If script structure is unsupported/ambiguous, behavior still falls back conservatively to existing prompt path.
- No broadening of execution permission outside existing prefix-rule semantics.

### Representative command
`/bin/zsh -lc "gh run view --help > /tmp/codex-gh-run-view-help.log && echo saved"`

### Before
- Select "Yes, and don't ask again for commands that start with `gh run view`".
- Same command shape triggers approval again because parser falls back to wrapper argv.

### After
- Same saved rule matches inner command extraction.
- No repeated approval prompt for the same command family.

## Tests Added
- `core/src/exec_policy.rs`
  - `evaluates_redirected_shell_script_against_prefix_rules`
  - `commands_for_exec_policy_parses_redirected_sequences`
- `shell-command/src/bash.rs`
  - `parse_shell_lc_commands_for_exec_policy_supports_redirects`
  - `parse_shell_lc_commands_for_exec_policy_rejects_substitutions`
- `core/tests/suite/approvals.rs`
  - `gh_run_view_prefix_rule_allows_redirected_command_without_prompt`

### Local bug reproduction
- Reproduced in this environment using redirected wrapped command + saved prefix rule.
- Verified fail-before/pass-after by temporarily restoring old parser call in `core/src/exec_policy.rs`:
  - without fix: integration test requests approval unexpectedly,
  - with fix restored: integration test passes.

### Local checks run
- `cargo test -p codex-core redirected -- --nocapture`
- `cargo test -p codex-core gh_run_view_prefix_rule_allows_redirected_command_without_prompt -- --nocapture`
- `just fix -p codex-core`

All passed.

## Related Reports
- Same root-cause pattern (redirection + repeated approvals): #10321
  - Issue example includes `xcodebuild ... 2>&1` and repeated approval despite prior allow.
- Potentially related but likely different root causes were also reviewed (for example env-var-prefix parsing reports such as #11298).

## Existing PRs Checked
- No open PR found that fixes this redirection-specific wrapper matching gap.
- Related prior work:
  - #10941 (wrapper canonicalization + heredoc prefix support)
  - #9565 (multiline quoted args parsing)

This PR complements those by covering non-heredoc redirection command extraction.

### DIFF
--- a/codex-rs/core/src/exec_policy.rs
+++ b/codex-rs/core/src/exec_policy.rs
@@ -24,7 +24,7 @@ use thiserror::Error;
 use tokio::fs;
 use tokio::task::spawn_blocking;
 
-use crate::bash::parse_shell_lc_plain_commands;
+use crate::bash::parse_shell_lc_commands_for_exec_policy;
 use crate::bash::parse_shell_lc_single_command_prefix;
 use crate::sandboxing::SandboxPermissions;
 use crate::tools::sandboxing::ExecApprovalRequirement;
@@ -469,7 +469,7 @@ fn default_policy_path(codex_home: &Path) -> PathBuf {
 }
 
 fn commands_for_exec_policy(command: &[String]) -> (Vec<Vec<String>>, bool) {
-    if let Some(commands) = parse_shell_lc_plain_commands(command)
+    if let Some(commands) = parse_shell_lc_commands_for_exec_policy(command)
         && !commands.is_empty()
     {
         return (commands, false);
@@ -994,6 +994,40 @@ prefix_rule(pattern=["rm"], decision="forbidden")
         );
     }
 
+    #[tokio::test]
+    async fn evaluates_redirected_shell_script_against_prefix_rules() {
+        let policy_src = r#"prefix_rule(pattern=["gh", "run", "view"], decision="allow")"#;
+        let mut parser = PolicyParser::new();
+        parser
+            .parse("test.rules", policy_src)
+            .expect("parse policy");
+        let policy = Arc::new(parser.build());
+
+        let command = vec![
+            "bash".to_string(),
+            "-lc".to_string(),
+            "gh run view 123 --log > /tmp/out.log && echo saved".to_string(),
+        ];
+
+        let requirement = ExecPolicyManager::new(policy)
+            .create_exec_approval_requirement_for_command(ExecApprovalRequest {
+                command: &command,
+                approval_policy: AskForApproval::OnRequest,
+                sandbox_policy: &SandboxPolicy::new_read_only_policy(),
+                sandbox_permissions: SandboxPermissions::UseDefault,
+                prefix_rule: None,
+            })
+            .await;
+
+        assert_eq!(
+            requirement,
+            ExecApprovalRequirement::Skip {
+                bypass_sandbox: true,
+                proposed_execpolicy_amendment: None,
+            }
+        );
+    }
+
     #[test]
     fn commands_for_exec_policy_falls_back_for_empty_shell_script() {
         let command = vec!["bash".to_string(), "-lc".to_string(), "".to_string()];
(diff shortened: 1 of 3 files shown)


almuqrin:

almuqrin:
@etraut-openai understandable, please check [the issue I linked](https://github.com/openai/codex/issues/10321), and, keep in mind that:

1. I've reproduced the issue locally.
2. I've verified fail-before/pass-after:
    1. Before my changes: integration test requests approval unexpectedly
    2. After my changes: integration test passes

erikeckhardt:
This change, broadening `execpolicy` parsing to accept redirected shell scripts, also appears to allow `variable_assignment` nodes and then ignore them when reconstructing the command prefix. That means a rule like `prefix_rule(pattern=["gh", "run", "view"], decision="allow")` will match environment-prefixed invocations such as:

```
LD_PRELOAD=/tmp/x gh run view
```

This is not fully safe, because `LD_PRELOAD` tells a Unix dynamic linker to load a shared library before the program’s normal libraries. This can be useful, but is capable of completely changing a given command's behavior, to potentially allow unsafe/unexpected execution behavior. LD_PRELOAD can be useful for debugging, profiling, or instrumentation, but makes `LD_PRELOAD=/tmp/x gh run view` meaningfully different from plain `gh run view`, which is likely outside a user’s expectation when approving ordinary use of that command.

`LD_PRELOAD` is only one example here. Mitigating that specific variable would not address the underlying issue: environment pre-assignment in general can materially change execution semantics while still matching the same reconstructed command prefix.

PR #11871, opened 2026-02-15
feat(core) Introduce Feature::RequestPermissions

dylan-hurd-oai:
## Summary
Introduces the initial implementation of Feature::RequestPermissions. RequestPermissions allows the model to request that a command be run inside the sandbox, with additional permissions, like writing to a specific folder. Eventually this will include other rules as well, and the ability to persist these permissions, but this PR is already quite large - let's get the core flow working and go from there!

Screenshot 2026-02-15 at 2 26 22 PM

## Testing
- [x] Added tests
- [x] Tested locally
- [x] Feature


codex-rs/core/src/tools/runtimes/unified_exec.rs
```diff
@@ -45,6 +47,7 @@ pub struct UnifiedExecRequest {
     pub network: Option<NetworkProxy>,
     pub tty: bool,
     pub sandbox_permissions: SandboxPermissions,
+    pub additional_permissions: Option<AdditionalPermissions>,
```
pakrym-oai, line 50: Should AdditionalPermissions be encapsulated in sandbox_permissions?

pakrym-oai: Oh, it's already the case. Why don't we store additional_permissions inside the SandboxPermissions::WithAdditionalPermissions?

codex-rs/tui/src/chatwidget/tests.rs
```diff
@@ -2527,6 +2528,7 @@ async fn exec_approval_decision_truncates_multiline_and_long_commands() {
         ),
         network_approval_context: None,
         proposed_execpolicy_amendment: None,
+        additional_permissions: None,
```
pakrym-oai, line 2827: Should network_approval_context and additional_permissions be somehow merged?

codex-rs/protocol/src/prompts/permissions/approval_policy/on_request_rule_request_permission.md
```diff
@@ -0,0 +1,28 @@
+# Permission Requests
+
+Commands may require user approval before execution. Prefer requesting sandboxed additional permissions instead of asking to run fully outside the sandbox.
+
+## Preferred request mode
+
+When you need extra filesystem access for one command, use:
+
+- `sandbox_permissions: "with_additional_permissions"`
```
pakrym-oai, line 9: do we need to make the model set sandbox_permissions?

codex-rs/core/src/tools/runtimes/unified_exec.rs
pakrym-oai: Got it, it's the parameter of the tool call.

codex-rs/core/src/sandboxing/mod.rs
```diff
@@ -239,10 +421,11 @@ impl SandboxManager {
 
 pub async fn execute_env(
     env: ExecRequest,
-    policy: &SandboxPolicy,
+    _policy: &SandboxPolicy,
```
pakrym-oai: this feels wrong. If there is another way to pass sandbox policy in should we delete `_policy: &SandboxPolicy` parameter from all the callsites?

codex-rs/core/src/tools/sandboxing.rs
```diff
@@ -183,12 +184,58 @@ pub(crate) fn default_exec_approval_requirement(
...
     NoOverride,
     BypassSandboxFirstAttempt,
 }
 
+pub(crate) fn sandbox_override_for_first_attempt(
+    sandbox_permissions: SandboxPermissions,
+    exec_approval_requirement: &ExecApprovalRequirement,
+) -> SandboxOverride {
+    if sandbox_permissions.requires_escalated_permissions()
+        || (!sandbox_permissions.uses_additional_permissions()
```
pakrym-oai: why do we have to check for sandbox_permissions.uses_additional_permissions? If requires escalation is true we should just escalate

codex-rs/core/src/codex.rs
```diff
@@ -8460,4 +8464,60 @@ mod tests {
 
         pretty_assertions::assert_eq!(output, expected);
     }
+
+    #[tokio::test]
+    async fn unified_exec_releases_process_id_when_additional_permissions_invalid() {
```
pakrym-oai: this is a very specific test. what is the connection between process  ids and additional permissions?

pakrym-oai:
I think this PR needs a ton more integration tests that both verify that additional permissions work, error responses an that things are escalated/not escalated accordingly.

codex-rs/core/src/sandboxing/mod.rs
```diff
@@ -82,11 +88,180 @@ pub enum SandboxPreference {
...
+                exclude_slash_tmp: *exclude_slash_tmp,
+            }
+        }
+        SandboxPolicy::ReadOnly { access } => {
+            if extra_writes.is_empty() {
+                SandboxPolicy::ReadOnly {
+                    access: merge_read_only_access_with_additional_reads(access, extra_reads),
+                }
+            } else {
+                SandboxPolicy::WorkspaceWrite {
```
jif-oai, line 250: I think here there is a small bug. `ReadOnly` basically means the user does not want the model to write anywhere (cwd included).

In this specific case, the model ask to be able to write in one specific directory (from `extra_writes`). But if we check how `WorkspaceWrite` is used, we see in `get_writable_roots_with_cwd` that by default it gives access to `writable_root=extra_writes` AND the `cwd` (and `/tmp` btw).

This means this permission request implicitly give access to the full CWD

codex-rs/core/src/sandboxing/mod.rs
```diff
@@ -82,11 +88,180 @@ pub enum SandboxPreference {
...
+            AbsolutePathBuf::resolve_path_against_base(&path, command_cwd).map_err(|err| {
+                format!(
+                    "{permission_kind} path `{}` cannot be resolved against cwd `{}`: {err}",
+                    path.display(),
+                    command_cwd.display()
+                )
+            })?
+        };
+
+        let canonicalized = resolved
```
jif-oai, line 138: I would not make this best effort. This means that if the path does not exist, we can still grant a permission to access it. This sounds like a potential TOCTOU race

codex-rs/core/src/tools/handlers/mod.rs
```diff
@@ -49,3 +54,51 @@ where
         FunctionCallError::RespondToModel(format!("failed to parse function arguments: {err}"))
     })
 }
+
+pub(super) fn normalize_and_validate_additional_permissions(
```
jif-oai, line 60: can we docstring this?

codex-rs/core/src/tools/handlers/mod.rs
```diff
@@ -49,3 +54,51 @@ where
...
+    if !request_permission_enabled
+        && (sandbox_permissions.uses_additional_permissions() || additional_permissions.is_some())
+    {
+        return Err(
+            "additional permissions are disabled; enable `features.request_permission` before using `with_additional_permissions`"
+                .to_string(),
+        );
+    }
+
+    if sandbox_permissions.uses_additional_permissions() {
```
jif-oai: this thing is a bit strange

Generally speaking we have 2 concepts here:
* Approval policy: turn-based policy. We require it to be `OnRequest` otherwise we reject
* Tool specs: thread-based definition. We only include `additional_permissions` if the feature flag is enabled

It means that with the current behaviour, the model has no way to know if the command will get rejected or not due to the approval policy. That might confuse it or even prevent it from asking for approval. I think we should find a way to notify the model when there is a change in here

codex-rs/core/src/tools/runtimes/unified_exec.rs
dylan-hurd-oai: I went through the same thought process, will see if I can clean this up a bit more

codex-rs/protocol/src/prompts/permissions/approval_policy/on_request_rule_request_permission.md
dylan-hurd-oai: Technically no, but I think this is helpful for consistency with `require_escalated`

codex-rs/core/src/sandboxing/mod.rs
dylan-hurd-oai: Should have added a comment here - I agree that this doesn't map perfectly. We could add a new `SandboxPolicy::SpecificWrite` which doesn't include the cwd

codex-rs/tui/src/chatwidget/tests.rs
dylan-hurd-oai: Soon! I think both are still actively being worked on - I'd like to merge once we've stabilized the APIs

codex-rs/core/src/sandboxing/mod.rs
dylan-hurd-oai: Nice catch!

codex-rs/core/src/tools/sandboxing.rs
dylan-hurd-oai: hm I think the idea here was that `uses_additional_permissions()` should always correspond to NoOverride but i think that's actually incorrect - fixed

codex-rs/core/src/sandboxing/mod.rs
dylan-hurd-oai: Agreed, good callout. Looking at our callsites, we're pulling from the ExecRequest anyway, so a detached policy argument feels less correct cc @jif-oai

codex-rs/core/src/sandboxing/mod.rs
dylan-hurd-oai: I've opened https://github.com/openai/codex/pull/12504 as an example of how we might support fine-grained controls here

codex-rs/core/src/tools/handlers/mod.rs
dylan-hurd-oai: Yeah tool definitions are tricky. I agree we need to address this before rollout, but I think this is something we should solve more generally via ContextManager. e.g. we could store tool definitions from turn to turn and send a developer message when they change cc @charley-oai

codex-rs/core/src/sandboxing/mod.rs
dylan-hurd-oai: Fixed

codex-rs/core/src/sandboxing/mod.rs
jif-oai: SG, must be a blocker to release the feature

codex-rs/core/src/tools/handlers/mod.rs
```diff
@@ -49,3 +54,58 @@ where
...
+    if !request_permission_enabled
+        && (uses_additional_permissions || additional_permissions.is_some())
+    {
+        return Err(
+            "additional permissions are disabled; enable `features.request_permission` before using `with_additional_permissions`"
+                .to_string(),
+        );
+    }
+
+    if uses_additional_permissions {
```
jif-oai, line 81: I guess this is intended that if we are in full access, the request is basically ignored?

codex-rs/core/src/sandboxing/mod.rs
```diff
@@ -81,11 +87,183 @@ pub enum SandboxPreference {
...
+                // UnderDevelopment, it's a useful approximation of the desired behavior.
+                SandboxPolicy::WorkspaceWrite {
+                    writable_roots: dedup_absolute_paths(extra_writes),
+                    read_only_access: merge_read_only_access_with_additional_reads(
+                        access,
+                        extra_reads,
+                    ),
+                    network_access: false,
+                    exclude_tmpdir_env_var: false,
+                    exclude_slash_tmp: false,
```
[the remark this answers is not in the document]
jif-oai: As discussed

jif-oai (commented):
Not ready to land because of the read escalation thing but good to iterate on

jif-oai (approved):

codex-rs/core/src/sandboxing/mod.rs
dylan-hurd-oai: 100% agreed!

codex-rs/core/src/tools/handlers/mod.rs
dylan-hurd-oai: I think the intention here is to remain similar to OnRequest behavior!

dylan-hurd-oai: This code might be overly defensive, since it's already handled [here](https://github.com/openai/codex/blob/90c26e7f92ad67988a6dbfc98c2b9e9f800b0073/codex-rs/core/src/tools/handlers/shell.rs#L305-L313). will think about how to clean up as we iterate on this feature

PR #11873, opened 2026-02-15
fix(tui) Consolidate Permissions logic

dylan-hurd-oai:
## Summary
This logic was split into 2 disparate and confusing places:
1. utils::approvals_presets
2. chatwidget.rs

The first is no longer used by the mcp-server or app-server, since behavior there has diverged quite a bit. We could eventually converge this logic, but right now it has one consumer and the code is quite mangled trying to layer tui-specific logic on top.

We instead consolidate this into the `PermissionsPreset` construct, which will replace ApprovalPreset (and is more correctly named).

## Testing
- [x] Tests pass, additional tests added


codex-rs/tui/src/permissions/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+            id: "read-only",
+            label: "Read Only",
+            description: "Codex can read files in the current workspace. Approval is required to edit files or access the internet.",
+            approval: AskForApproval::OnRequest,
+            sandbox: SandboxPolicy::new_read_only_policy(),
+        },
+        PermissionsPreset {
+            id: "auto",
+            label: "Default",
+            description: "Codex can read and edit files in the current workspace, and run commands. Approval is required to access the internet or edit other files.",
```
bolinfest, line 43: Remove comma?

```suggestion
            description: "Codex can read and edit files in the current workspace and run commands. Approval is required to access the internet or edit other files.",
```

codex-rs/tui/src/permissions/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        .into_iter()
+        .filter(PermissionsPreset::is_visible)
+        .map(|preset| preset.to_selection_item(config))
+        .collect()
+}
+
+impl PermissionsPreset {
+    fn is_visible(&self) -> bool {
+        matches!(self.id, "auto" | "full-access")
+            || (cfg!(target_os = "windows") && matches!(self.id, "read-only"))
```
bolinfest, line 68: read-only is only supported on Windows?

codex-rs/tui/src/permissions/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// Stable identifier for the preset.
+    pub id: &'static str,
+    /// Display label shown in UIs.
+    pub label: &'static str,
+    /// Short human description shown next to the label in UIs.
+    pub description: &'static str,
+    /// Approval policy to apply.
+    pub approval: AskForApproval,
+    /// Sandbox policy to apply.
+    pub sandbox: SandboxPolicy,
```
bolinfest, line 27: Could make this a closure that returns a `SandboxPolicy` instead?

codex-rs/tui/src/permissions/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+            is_current: self.is_current(config),
+            actions: self.actions(config),
+            dismiss_on_select: true,
+            disabled_reason: self.disabled_reason(config),
+            ..Default::default()
+        }
+    }
+
+    fn is_current(&self, config: &Config) -> bool {
+        self.approval == config.permissions.approval_policy.value()
```
bolinfest, line 90: Should there be a separate id that corresponds to "custom" and should equality be based on id rather than approval and sandbox values?

codex-rs/tui/src/permissions/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+use codex_core::protocol::Op;
+use codex_core::protocol::SandboxPolicy;
+#[cfg(target_os = "windows")]
+use codex_core::protocol_config_types::WindowsSandboxLevel;
+
+/// A simple preset pairing an approval policy with a sandbox policy.
+#[derive(Debug, Clone)]
+pub struct PermissionsPreset {
+    /// Stable identifier for the preset.
+    pub id: &'static str,
```
bolinfest, line 19: enum instead of static string?

codex-rs/tui/src/permissions/mod.rs
dylan-hurd-oai: I agree! Was planning to do that in a follow-up PR, but not opposed to including here

codex-rs/tui/src/permissions/mod.rs
dylan-hurd-oai: cc @iceweasel-oai we've left this in place as part of the windows sandbox path. Will defer to him for correct values here - we might want to make read-only `untrusted` as well.

codex-rs/tui/src/permissions/mod.rs
dylan-hurd-oai: I was split on this, since you might have many custom presets. Maybe the right setup is something like this:
```
enum PermissionsId {
  ReadOnly
  Default
  FullAccess
  Custom(custom_id: string)
}
```

PR #11899, opened 2026-02-16
feat: prefer `gpt-5.3-codex-spark` model for explorer when available

jif-oai:
Summary
- make the explorer profile pick the preferred `gpt-5.3-codex-spark` when it appears in the available model list and fall back to the mini model otherwise
- pass the cached available models into `apply_to_config` so spawned agents can choose their roles appropriately
- add unit coverage ensuring the explorer profile respects the available models list

Testing
- Not run (not requested)


Szpadel:
One thing that came to my mind: What will happen in case of weekly/5h limit running our for spark model?
As I understand this implementation this will still lead to failures, maybe more generic solution to trigger fallback in case primary model call fails?

Issue #11902, opened 2026-02-16
macOS app: voice transcription send fails with 'Error starting conversation' and creates zombie thread


twentyOne2x:
### What version of the Codex App are you using (From “About Codex” dialog)?

Latest as of 2026-02-16 (will add exact build # shortly)

### What subscription do you have?

N/A

### What issue are you seeing?

When using the macOS app voice transcription (mic), after transcribing and pressing Send on a *new thread*, the UI briefly flashes the transcribed text into the chat input, attempts to submit it, then errors immediately with the toast: `Error starting conversation`.

Despite the error, the app still creates a thread entry which then loads forever. These stuck threads also cannot be archived.

Model: `GPT-5.3-Codex`
Reasoning: `Extra High`

### What steps can reproduce the bug?

1. Open the Codex macOS app
2. Start a new thread
3. Use voice transcription (mic), speak a prompt
4. Press Send
5. Observe instant toast `Error starting conversation`
6. Observe a new thread is created but is stuck loading forever and cannot be archived

### What is the expected behavior?

- Transcribed text should send normally and start the conversation
- If the conversation fails to start, the app should not create a broken thread, or should allow it to be recovered/archived

### Additional information

- This is reproducible for me on a new thread (not intermittent)
- Happy to provide app version/build number, macOS version, and logs if you can point to the right log location

twentyOne2x:
Opened PR #11905 to make v1  work even when  is missing/null by resolving the rollout from . This should address the stuck/zombie threads being unarchivable (even if the underlying conversation-start error persists).

twentyOne2x:
PR up: #11910

This should address the instant “Error starting conversation” path when the v1 `newConversation` flow ends up starting an ephemeral session (no `rolloutPath`), which then errors while responding and leaves a zombie thread in the UI.

twentyOne2x:
New repro detail: the failure seems input-device-dependent.

- Voice transcription works for me w/ AirPods Pro selected as input, but fails when selecting an external mic.
- When it fails, the desktop logs show a local exception path (not a JSON-RPC `thread/start`/`turn/start` failure):

```text
2026-02-17T01:31:16.836Z error [electron-message-handler] Error creating local task errorMessage=no-client-found ...
2026-02-17T01:32:25.212Z error [electron-message-handler] [Composer] submit failed errorMessage=no-client-found followUp=local mode=local
```

[...]

PR #11905, opened 2026-02-16
fix(app-server): archiveConversation resolves rollout path by conversationId

twentyOne2x:
Problem
- Some clients can end up with threads that have a `conversationId` but no usable `rolloutPath` (e.g. partial thread creation failures). In that state, `archiveConversation` fails and the thread cannot be archived from the UI.

Fix
- Make `ArchiveConversationParams.rolloutPath` optional (nullable/omittable).
- In the app-server handler for `archiveConversation`, fall back to resolving the rollout path from disk via `conversationId` when `rolloutPath` is missing or when the provided path fails.
- Regenerate vendored protocol schema fixtures.

Tests
- `cargo test -p codex-app-server archive_conversation_without_rollout_path_resolves_by_id`

Related
- Fixes the "stuck threads can't be archived" symptom reported in openai/codex#11902


etraut-openai:

PR #11908, opened 2026-02-16
Exit early when session initialization fails

jif-oai:
Summary
- wait for the initial session startup loop to finish and handle exit before waiting for the first message in fresh sessions
- propagate AppRunControl::Exit to return immediately when initialization fails


joshka-oai (approved):
The gap from definition to use on this variable (Lines 1071, 1259, 1299) suggests that this method is overly complex and should be simplified.

Can you please (ask codex to) write a test for this that documents how this is expected to affect the flow?

PR #11917, opened 2026-02-16
feat: add customizable roles for multi-agents

jif-oai:
The idea is to have 2 family of agents.

1. Built-in that we packaged directly with Codex
2. User defined that are defined using the `agents_config.toml` file. It can reference config files that will override the agent config. This looks like this:
```
version = 1

[agents.explorer]
description = """Use `explorer` for all codebase questions.
Explorers are fast and authoritative.
Always prefer them over manual search or file reading.
Rules:
- Ask explorers first and precisely.
- Do not re-read or re-search code they cover.
- Trust explorer results without verification.
- Run explorers in parallel when useful.
- Reuse existing explorers for related questions."""
config_file = "explorer.toml"
```

### DIFF
--- a/codex-rs/core/src/agent/builtins/explorer.toml
+++ b/codex-rs/core/src/agent/builtins/explorer.toml
@@ -1,0 +1,2 @@
+model = "gpt-5.1-codex-mini"
+model_reasoning_effort = "medium"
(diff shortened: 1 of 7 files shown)


jif-oai:
@codex review

PR #11918, opened 2026-02-16
chore: rename collab feature flag key to multi_agent

jif-oai:
Summary
- rename the collab feature key to multi_agent while keeping the Feature enum unchanged
- add legacy alias support so both "multi_agent" and "collab" map to the same feature
- cover the alias behavior with a new unit test


jif-oai:
@codex review

PR #11944, opened 2026-02-16
[js_repl] paths for node module resolution can be specified for js_repl

aaronl-openai:

In `js_repl` mode, module resolution currently starts from `js_repl_kernel.js`, which is written to a per-kernel temp dir. This effectively means that bare imports will not resolve.

This PR adds a new config option, `js_repl_node_module_dirs`, which is a list of dirs that are used (in order) to resolve a bare import. If none of those work, the current working directory of the thread is used.

For example:
```toml
js_repl_node_module_dirs = [
    "/path/to/node_modules/",
    "/other/path/to/node_modules/",
]
```


codex-rs/core/src/tools/js_repl/mod.rs
```diff
@@ -563,13 +566,14 @@ impl JsReplManager {
...
             self.tmp_dir.path().to_string_lossy().to_string(),
         );
-        env.insert(
-            "CODEX_JS_REPL_HOME".to_string(),
-            self.codex_home
-                .join("js_repl")
-                .to_string_lossy()
-                .to_string(),
-        );
+        if !self.node_module_dirs.is_empty() {
```
fjord-oai: According to codex:

The docs state env var precedence over config, but start_kernel overwrites CODEX_JS_REPL_NODE_MODULE_DIRS whenever js_repl_node_module_dirs is non-empty. Since create_env(...) starts from inherited env, a user-provided env var is replaced by config values, reversing the documented order.

docs/js_repl.md
```diff
@@ -37,6 +37,25 @@ You can configure an explicit runtime path:
 js_repl_node_path = "/absolute/path/to/node"
 ```
 
+## Module resolution
```
fjord-oai, line 40: can we add some tests for this?

docs/js_repl.md
```diff
@@ -37,6 +37,25 @@ You can configure an explicit runtime path:
...
+
+`js_repl` resolves bare specifiers (for example `await import("atlas")`) using an ordered
+search path:
+
+1. `CODEX_JS_REPL_NODE_MODULE_DIRS` (PATH-delimited list)
+2. `js_repl_node_module_dirs` in config/profile (array of absolute paths)
+3. Thread working directory (cwd)
```
fjord-oai: we probably shouldn't default to including the cwd, right? since for atlas we only want the read-only system modules and a conversation-specific dir.

codex-rs/core/src/tools/js_repl/kernel.js
```diff
@@ -111,6 +111,31 @@ const tmpDir = process.env.CODEX_JS_TMP_DIR || process.cwd();
 // Explicit long-lived mutable store exposed as `codex.state`. This is useful
 // when callers want shared state without relying on lexical binding carry-over.
 const state = {};
+const nodeModuleDirEnv = process.env.CODEX_JS_REPL_NODE_MODULE_DIRS ?? "";
```
fjord-oai, line 120: codex complains that:

Right now `resolveSpecifier` still permits path-style specifiers (`./`, `../`, absolute, `file:`) and resolves bare imports via CJS-style `require.resolve`. That breaks strict control of module search order and can fail for ESM-only packages (`exports.import`). Can we tighten this so non-bare specifiers are rejected up front, and bare packages are resolved per base with `createRequire(path.join(base, "__codex_repl__.js")).resolve(specifier, { conditions: new Set(["node", "import"]) })`, then imported by file URL? Also please remove the `import(specifier)` fallback path so resolution never falls back to the temp kernel directory.


here's the code it suggests to fix (we should add unit tests around this too):
at the top of the file:
```js
const { builtinModules, createRequire } = require("node:module");
```

and then replacing the resolution code with:
```js
const nodeModuleDirEnv = process.env.CODEX_JS_REPL_NODE_MODULE_DIRS ?? "";
const moduleSearchBases = (() => {
  const bases = [];
  const seen = new Set();
  for (const entry of nodeModuleDirEnv.split(path.delimiter)) {
    const trimmed = entry.trim();
    if (!trimmed) {
      continue;
    }
    const resolved = path.isAbsolute(trimmed)
      ? trimmed
      : path.resolve(process.cwd(), trimmed);
    const base = path.basename(resolved) === "node_modules" ? path.dirname(resolved) : resolved;
    if (seen.has(base)) {
      continue;
    }
    seen.add(base);
    bases.push(base);
  }

  const cwd = process.cwd();
  if (!seen.has(cwd)) {
    bases.push(cwd);
  }
  return bases;
})();

const importResolveConditions = new Set(["node", "import"]);
const requireByBase = new Map();

function getRequireForBase(base) {
  let req = requireByBase.get(base);
  if (!req) {
    req = createRequire(path.join(base, "__codex_js_repl__.cjs"));
    requireByBase.set(base, req);
  }
  return req;
}

function isModuleNotFoundError(err) {
  return err?.code === "MODULE_NOT_FOUND" || err?.code === "ERR_MODULE_NOT_FOUND";
}

function isWithinBaseNodeModules(base, resolvedPath) {
  const nodeModulesRoot = path.resolve(base, "node_modules");
  const relative = path.relative(nodeModulesRoot, resolvedPath);
  return relative !== "" && !relative.startsWith("..") && !path.isAbsolute(relative);
}

function isBarePackageSpecifier(specifier) {
  if (typeof specifier !== "string" || !specifier || specifier.trim() !== specifier) {
    return false;
  }
  if (specifier.startsWith("./") || specifier.startsWith("../")) {
    return false;
  }
  if (specifier.startsWith("/") || specifier.startsWith("\\")) {
    return false;
  }
  if (path.isAbsolute(specifier)) {
    return false;
  }
  if (/^[a-zA-Z][a-zA-Z\d+.-]*:/.test(specifier)) {
    return false;
  }
  if (specifier.includes("\\")) {
    return false;
  }
  return true;
}

function resolveBareSpecifier(specifier) {
  let firstResolutionError = null;

  for (const base of moduleSearchBases) {
    try {
      const resolved = getRequireForBase(base).resolve(specifier, {
        conditions: importResolveConditions,
      });
      if (isWithinBaseNodeModules(base, resolved)) {
        return resolved;
      }
      // Ignore resolutions that escape this base via parent node_modules lookup.
    } catch (err) {
      if (isModuleNotFoundError(err)) {
        continue;
      }
      if (!firstResolutionError) {
        firstResolutionError = err;
      }
    }
  }

  if (firstResolutionError) {
    throw firstResolutionError;
  }
  return null;
}

function resolveSpecifier(specifier) {
  if (specifier.startsWith("node:") || builtinModuleSet.has(specifier)) {
    if (isDeniedBuiltin(specifier)) {
      throw new Error(`Importing module "${specifier}" is not allowed in js_repl`);
    }
    return { kind: "builtin", specifier: toNodeBuiltinSpecifier(specifier) };
  }

  if (!isBarePackageSpecifier(specifier)) {
    throw new Error(
      `Unsupported import specifier "${specifier}" in js_repl. Use a package name like "lodash" or "@scope/pkg".`,
    );
  }

  const resolvedBare = resolveBareSpecifier(specifier);
  if (!resolvedBare) {
    throw new Error(`Module not found: ${specifier}`);
  }

  return { kind: "path", path: resolvedBare };
}

function importResolved(resolved) {
  if (resolved.kind === "builtin") {
    return import(resolved.specifier);
  }
  if (resolved.kind === "path") {
    return import(pathToFileURL(resolved.path).href);
  }
  throw new Error(`Unsupported module resolution kind: ${resolved.kind}`);
}

```

docs/js_repl.md
aaronl-openai: so for use the current working drive is the read-write conversation specific dir. I'm not sure if that's true by convention though (e.g. if the current working drive could sometimes be something that you don't have special access to), in which case we could just drop this special handling and rely solely on the `CODEX_JS_REPL_NODE_MODULE_DIRS`. I do think it will be pretty common to use cwd for node modules though so it might be annoying to always have to specify that.

fjord-oai: yeah, that's fair. we should just make it really clear in the docs that cwd is always included then

fjord-oai (approved):

PR #11964, opened 2026-02-17
add(core): safety check downgrade warning

fouad-openai:
Add per-turn notice when a request is downgraded to a fallback model due to cyber safety checks.

**Changes**

  - codex-api: Emit a ServerModel event based on the openai-model response header and/or response payload (SSE + WebSocket), including when the model changes mid-stream.
  - core: When the server-reported model differs from the requested model, emit a single per-turn warning explaining the reroute to gpt-5.2 and directing users to Trusted
    Access verification and the cyber safety explainer.
  - app-server (v2): Surface these cyber model-routing warnings as synthetic userMessage items with text prefixed by Warning: (and document this behavior).

### DIFF
--- a/codex-rs/app-server/README.md
+++ b/codex-rs/app-server/README.md
@@ -557,7 +557,7 @@ Today both notifications carry an empty `items` array even when item events were
 
 `ThreadItem` is the tagged union carried in turn responses and `item/*` notifications. Currently we support events for the following items:
 
-- `userMessage` — `{id, content}` where `content` is a list of user inputs (`text`, `image`, or `localImage`).
+- `userMessage` — `{id, content}` where `content` is a list of user inputs (`text`, `image`, or `localImage`). Cyber model-routing warnings are surfaced as synthetic `userMessage` items with `text` prefixed by `Warning:`.
 - `agentMessage` — `{id, text}` containing the accumulated agent reply.
 - `plan` — `{id, text}` emitted for plan-mode turns; plan text can stream via `item/plan/delta` (experimental).
 - `reasoning` — `{id, summary, content}` where `summary` holds streamed reasoning summaries (applicable for most OpenAI models) and `content` holds raw reasoning blocks (applicable for e.g. open source models).
--- a/codex-rs/app-server/src/bespoke_event_handling.rs
+++ b/codex-rs/app-server/src/bespoke_event_handling.rs
@@ -68,6 +68,7 @@ use codex_app_server_protocol::TurnInterruptResponse;
 use codex_app_server_protocol::TurnPlanStep;
 use codex_app_server_protocol::TurnPlanUpdatedNotification;
 use codex_app_server_protocol::TurnStatus;
+use codex_app_server_protocol::UserInput as V2UserInput;
 use codex_app_server_protocol::build_turns_from_rollout_items;
 use codex_core::CodexThread;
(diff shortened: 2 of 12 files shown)


codex-rs/app-server/README.md
```diff
@@ -557,7 +557,7 @@ Today both notifications carry an empty `items` array even when item events were
 
 `ThreadItem` is the tagged union carried in turn responses and `item/*` notifications. Currently we support events for the following items:
 
-- `userMessage` — `{id, content}` where `content` is a list of user inputs (`text`, `image`, or `localImage`).
+- `userMessage` — `{id, content}` where `content` is a list of user inputs (`text`, `image`, or `localImage`). Cyber model-routing warnings are surfaced as synthetic `userMessage` items with `text` prefixed by `Warning:`.
```
apanasenko-oai, line 560: cc: @owenlin0 @celia-oai if you want to iterate on V2 API for warning.

apanasenko-oai (approved):

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -122,6 +125,35 @@ pub(crate) async fn apply_bespoke_event_handling(
...
+            if matches!(api_version, ApiVersion::V2)
+                && is_safety_check_downgrade_warning(&warning_event.message)
+            {
+                let item = ThreadItem::UserMessage {
+                    id: warning_item_id(&event_turn_id, &warning_event.message),
+                    content: vec![V2UserInput::Text {
+                        text: format!("Warning: {}", warning_event.message),
+                        text_elements: Vec::new(),
+                    }],
+                };
```
shijie-oai, line 138: Want to confirm this - user message usually contains user initiated content and not a server event. ~If we will address this as a fast follow that is fine but want to make sure that is clear here.~

The more I think about this, I don't think this is the right event for us to send for a warning. Would like to get more input from @owenlin0 and @celia-oai.

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -122,6 +125,35 @@ pub(crate) async fn apply_bespoke_event_handling(
...
             handle_turn_complete(conversation_id, event_turn_id, &outgoing, &thread_state).await;
         }
+        EventMsg::Warning(warning_event) => {
+            if matches!(api_version, ApiVersion::V2)
+                && is_safety_check_downgrade_warning(&warning_event.message)
+            {
+                let item = ThreadItem::UserMessage {
+                    id: warning_item_id(&event_turn_id, &warning_event.message),
+                    content: vec![V2UserInput::Text {
+                        text: format!("Warning: {}", warning_event.message),
```
shijie-oai, line 135: Just to confirm - is this ready for UI the display directly?

codex-rs/app-server/src/bespoke_event_handling.rs
shijie-oai: We can create a new `ServerNotification::modelDowngrade` that contains the initial and downgraded model and better to construct that warning message in extension and app to have better locale support.

owenlin0: Yes, this is kind of abusing the user message item. We should use a new server notification (`ModelDowngradedNotification` sounds good). @fouad-openai mind addressing?

PR #12001, opened 2026-02-17
Feat: add model reroute notification

shijie-oai:
### Summary
Builiding off https://github.com/openai/codex/pull/11964/files/5c75aa7b89a70bc2cc410a6fd238749306ec4c5e#diff-058ae8f109a8b84b4b79bbfa45f522c2233b9d9e139696044ae374d50b6196e0, we have created a `model/rerouted` notification that captures the event so that consumers can render as expected. Keep the `EventMsg::Warning` path in core so that this does not affect TUI rendering. 

`model/rerouted` is meant to be generic to account for future usage including capacity planning etc.


codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -125,32 +123,18 @@ pub(crate) async fn apply_bespoke_event_handling(
...
-                let started = ItemStartedNotification {
-                    thread_id: conversation_id.to_string(),
-                    turn_id: event_turn_id.clone(),
-                    item: item.clone(),
-                };
-                outgoing
-                    .send_server_notification(ServerNotification::ItemStarted(started))
-                    .await;
-                let completed = ItemCompletedNotification {
+        EventMsg::Warning(_warning_event) => {}
```
shijie-oai, line 126: Let me know if I should keep this.

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3305,6 +3312,17 @@ pub struct AccountLoginCompletedNotification {
...
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ModelReroutedNotification {
+    pub thread_id: String,
+    pub turn_id: String,
+    pub from_model: String,
+    pub to_model: String,
+    pub reason: ModelRerouteReason,
+}
```
shijie-oai, line 3324: We will construct the message, tooltip etc on the app side. This also allows up to be more dynamic in the future.

codex-rs/app-server/src/bespoke_event_handling.rs
shijie-oai: I think I should.

owenlin0 (approved):

codex-rs/core/src/codex.rs
```diff
@@ -2579,6 +2581,16 @@ impl Session {
             "Your account was flagged for potentially high-risk cyber activity and this request was routed to gpt-5.2 as a fallback. To regain access to gpt-5.3-codex, apply for trusted access: {CYBER_VERIFY_URL} or learn more: {CYBER_SAFETY_URL}"
         );
 
+        self.send_event(
+            turn_context,
+            EventMsg::ModelReroute(ModelRerouteEvent {
+                from_model: requested_model.clone(),
+                to_model: server_model.clone(),
+                reason: ModelRerouteReason::HighRiskCyberActivity,
```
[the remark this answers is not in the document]
owenlin0: seems legit?

PR #12024, opened 2026-02-17
Enforce errors on overriding built-in model providers

etraut-openai:
We receive bug reports from users who attempt to override one of the three built-in model providers (openai, ollama, or lmstuio). Currently, these overrides are silently ignored. This PR makes it an error to override them.

## Summary
- add validation for `model_providers` so `openai`, `ollama`, and `lmstudio` keys now produce clear configuration errors instead of being silently ignored

### DIFF
--- a/codex-rs/core/src/config/config_tests.rs
+++ b/codex-rs/core/src/config/config_tests.rs
@@ -160,35 +160,6 @@ consolidation_model = "gpt-5"
 }
 
 #[test]
-fn test_toml_parsing_rejects_reserved_model_provider_override_openai() {
-    let cfg = r#"
-[model_providers.openai]
-name = "OpenAI Custom"
-"#;
-
-    let err = toml::from_str::<ConfigToml>(cfg).expect_err("should reject reserved provider");
-    let message = err.to_string();
-    assert!(message.contains("reserved built-in provider IDs"));
-    assert!(message.contains("`openai`"));
-}
-
-#[test]
-fn test_toml_parsing_rejects_multiple_reserved_model_provider_overrides() {
-    let cfg = r#"
-[model_providers.ollama]
-name = "Ollama Override"
-
-[model_providers.openai]
-name = "OpenAI Override"
-"#;
-
-    let err = toml::from_str::<ConfigToml>(cfg).expect_err("should reject reserved providers");
-    let message = err.to_string();
-    assert!(message.contains("`openai`"));
-    assert!(message.contains("`ollama`"));
-}
-
-#[test]
 fn parses_bundled_skills_config() {
     let cfg: ConfigToml = toml::from_str(
         r#"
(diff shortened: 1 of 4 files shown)


codex-rs/core/src/config/mod.rs
```diff
@@ -139,6 +139,8 @@ pub(crate) const DEFAULT_AGENT_JOB_MAX_RUNTIME_SECONDS: Option<u64> = None;
 
 pub const CONFIG_TOML_FILE: &str = "config.toml";
 const OPENAI_BASE_URL_ENV_VAR: &str = "OPENAI_BASE_URL";
+const RESERVED_MODEL_PROVIDER_IDS: [&str; 3] =
+    ["openai", OLLAMA_OSS_PROVIDER_ID, LMSTUDIO_OSS_PROVIDER_ID];
```
etraut-openai: This line uses a hard-coded string for "openai". It should use the recently-added constant for the openai provider.

codex-rs/core/src/config/config_tests.rs
```diff
@@ -4929,6 +4972,29 @@ fn test_load_config_rejects_legacy_ollama_chat_provider_with_helpful_error() ->
     Ok(())
 }
 
+#[test]
+fn test_load_config_rejects_programmatic_reserved_model_provider_override() -> std::io::Result<()> {
```
etraut-openai: I don't think we need this test.

codex-rs/core/src/config/config_tests.rs
```diff
@@ -159,6 +159,49 @@ consolidation_model = "gpt-5"
...
+"#;
+
+    let err = toml::from_str::<ConfigToml>(cfg).expect_err("should reject reserved provider");
+    let message = err.to_string();
+    assert!(message.contains("reserved built-in provider IDs"));
+    assert!(message.contains("`openai`"));
+}
+
+#[test]
+fn test_toml_parsing_rejects_multiple_reserved_model_provider_overrides() {
```
etraut-openai: I don't think we need this test.

codex-rs/core/src/config/config_tests.rs
```diff
@@ -159,6 +159,49 @@ consolidation_model = "gpt-5"
     );
 }
 
+#[test]
+fn test_toml_parsing_rejects_reserved_model_provider_override_openai() {
```
etraut-openai: I don't think we need this test either.

codex-rs/core/src/config/config_tests.rs
```diff
@@ -159,6 +159,49 @@ consolidation_model = "gpt-5"
...
+    let err = toml::from_str::<ConfigToml>(cfg).expect_err("should reject reserved providers");
+    let message = err.to_string();
+    assert!(message.contains("`openai`"));
+    assert!(message.contains("`ollama`"));
+}
+
+#[test]
+fn test_toml_parsing_allows_non_reserved_model_provider() {
+    let cfg = r#"
+[model_providers.openai-custom]
```
etraut-openai: I don't think we need this test either.

bolinfest (approved):

PR #12025, opened 2026-02-17
app-server support for Windows sandbox setup.

iceweasel-oai:
app-server support for initiating Windows sandbox setup.
server responds quickly to setup request and makes a future RPC call back to client when the setup finishes.

The TUI implementation is unaffected but in a future PR I'll update the TUI to use the shared setup helper (`windows_sandbox.run_windows_sandbox_setup`)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -6042,6 +6052,53 @@ impl CodexMessageProcessor {
...
+            WindowsSandboxSetupMode::Elevated => CoreWindowsSandboxSetupMode::Elevated,
+            WindowsSandboxSetupMode::Unelevated => CoreWindowsSandboxSetupMode::Unelevated,
+        };
+        let config = Arc::clone(&self.config);
+        let outgoing = ThreadScopedOutgoingMessageSender::new(
+            Arc::clone(&self.outgoing),
+            vec![request_id.connection_id],
+        );
+
+        tokio::spawn(async move {
```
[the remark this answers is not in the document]
iceweasel-oai: this just won't happen in practice. There won't be multiple concurrent requests. Might fix later

.github/workflows/rust-ci.yml
```diff
@@ -513,7 +513,7 @@ jobs:
       - name: Install DotSlash
         uses: facebook/install-dotslash@v2
 
-      - uses: dtolnay/rust-toolchain@1.93.0
```
dylan-hurd-oai, line 516: let's rebase out .github changes

codex-rs/app-server/tests/suite/v2/mod.rs
```diff
@@ -29,3 +29,5 @@ mod thread_unarchive;
 mod turn_interrupt;
 mod turn_start;
 mod turn_steer;
+#[cfg(not(target_os = "windows"))]
+mod windows_sandbox_setup;
```
dylan-hurd-oai, line 32: should this be `#[cfg(target_os = "windows")]`?

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -657,6 +663,10 @@ impl CodexMessageProcessor {
                 self.list_mcp_server_status(to_connection_request_id(request_id), params)
                     .await;
             }
+            ClientRequest::WindowsSandboxSetupStart { request_id, params } => {
+                self.windows_sandbox_setup_start(to_connection_request_id(request_id), params)
+                    .await;
+            }
```
dylan-hurd-oai, line 671: [non-blocking] I think this api is a reasonable start but I wonder if we'll want either (a) progress notifications or (b) a polling api with progress results. Fine to leave as a follow-up!

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3151,6 +3151,35 @@ pub struct WindowsWorldWritableWarningNotification {
...
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct WindowsSandboxSetupStartParams {
+    pub mode: WindowsSandboxSetupMode,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct WindowsSandboxSetupStartResponse {}
```
dylan-hurd-oai: Should we just add a basic `started: true`? In case we hit an error before we spawn the tokio proccess.

codex-rs/core/src/windows_sandbox.rs
```diff
@@ -243,6 +247,185 @@ pub fn run_setup_refresh_with_extra_read_roots(
...
+    pub mode: WindowsSandboxSetupMode,
+    pub policy: SandboxPolicy,
+    pub policy_cwd: PathBuf,
+    pub command_cwd: PathBuf,
+    pub env_map: HashMap<String, String>,
+    pub codex_home: PathBuf,
+    pub active_profile: Option<String>,
+}
+
+pub async fn run_windows_sandbox_setup(request: WindowsSandboxSetupRequest) -> anyhow::Result<()> {
```
dylan-hurd-oai, line 267: Can we consolidate with `run_elevated_setup` api? Should tui be calling this instead? Fine to leave as a follow-up but trying to understand end state

.github/workflows/rust-ci.yml
iceweasel-oai: oh yeah, I removed these locally but I think I forgot to push.

codex-rs/app-server/src/codex_message_processor.rs
iceweasel-oai: my previous PR had a "heartbeat" call that we can make to the UI. Unfortunately it's hard to implement a good progress bar, since most of the setup is like a single very slow windows API call.

codex-rs/app-server/tests/suite/v2/mod.rs
iceweasel-oai: I'm cleaning this up a bit. It's confusing

codex-rs/core/src/windows_sandbox.rs
iceweasel-oai: yep - this is a wrapper around `run_elevated_setup` and `run_legacy_setup_preflight` and I am planning on having the TUI use this instead of `run_elevated_setup` directory. Just wanted to do it in a separate PR to keep this one tight.

codex-rs/app-server-protocol/src/protocol/v2.rs
iceweasel-oai: yeah good idea - added

dylan-hurd-oai (approved):

PR #12031, opened 2026-02-17
Add openai_base_url config override for built-in provider

etraut-openai:
We regularly get bug reports from users who mistakenly have the `OPENAI_BASE_URL` environment variable set. This PR deprecates this environment variable in favor of a top-level config key `openai_base_url` that is used for the same purpose. By making it a config key, it will be more visible to users. It will also participate in all of the infrastructure we've added for layered and managed configs.

Summary
- introduce the `openai_base_url` top-level config key, update schema/tests, and route the built-in openai provider through it while
- fall back to deprecated `OPENAI_BASE_URL` env var but warn user of deprecation when no `openai_base_url` config key is present
- update CLI, SDK, and TUI code to prefer the new config path (with a deprecated env-var fallback) and document the SDK behavior change

### DIFF
--- a/codex-rs/core/src/config/config_tests.rs
+++ b/codex-rs/core/src/config/config_tests.rs
@@ -4134,7 +4134,7 @@ model_verbosity = "high"
         supports_websockets: false,
     };
     let model_provider_map = {
-        let mut model_provider_map = built_in_model_providers(None);
+        let mut model_provider_map = built_in_model_providers(/* openai_base_url */ None);
         model_provider_map.insert("openai-custom".to_string(), openai_custom_provider.clone());
         model_provider_map
     };
--- a/codex-rs/core/src/config/config_tests.rs
+++ b/codex-rs/core/src/config/config_tests.rs
@@ -4134,7 +4134,7 @@ model_verbosity = "high"
         supports_websockets: false,
     };
     let model_provider_map = {
-        let mut model_provider_map = built_in_model_providers();
+        let mut model_provider_map = built_in_model_providers(None);
         model_provider_map.insert("openai-custom".to_string(), openai_custom_provider.clone());
         model_provider_map
     };
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -138,6 +138,7 @@ pub(crate) const DEFAULT_AGENT_MAX_DEPTH: i32 = 1;
 pub(crate) const DEFAULT_AGENT_JOB_MAX_RUNTIME_SECONDS: Option<u64> = None;
 
 pub const CONFIG_TOML_FILE: &str = "config.toml";
+const OPENAI_BASE_URL_ENV_VAR: &str = "OPENAI_BASE_URL";
 
 fn resolve_sqlite_home_env(resolved_cwd: &Path) -> Option<PathBuf> {
     let raw = std::env::var(codex_state::SQLITE_HOME_ENV).ok()?;
@@ -2260,30 +2261,31 @@ impl Config {
                 Some(trimmed.to_string())
             }
         });
-        let openai_base_url_from_env = std::env::var("OPENAI_BASE_URL").ok().and_then(|value| {
-            let trimmed = value.trim();
-            if trimmed.is_empty() {
-                None
-            } else {
-                Some(trimmed.to_string())
-            }
-        });
+        let openai_base_url_from_env = std::env::var(OPENAI_BASE_URL_ENV_VAR)
+            .ok()
+            .and_then(|value| {
+                let trimmed = value.trim();
+                if trimmed.is_empty() {
+                    None
+                } else {
+                    Some(trimmed.to_string())
+                }
+            });
         if openai_base_url_from_env.is_some() {
             let message = if openai_base_url.is_some() {
-                "`OPENAI_BASE_URL` is deprecated and ignored because `openai_base_url` is set in config.toml. Remove `OPENAI_BASE_URL` from your environment."
+                format!(
+                    "`{OPENAI_BASE_URL_ENV_VAR}` is deprecated and ignored because `openai_base_url` is set in config.toml. Remove `{OPENAI_BASE_URL_ENV_VAR}` from your environment."
+                )
             } else {
-                "`OPENAI_BASE_URL` is deprecated. Set `openai_base_url` in config.toml instead."
+                format!(
+                    "`{OPENAI_BASE_URL_ENV_VAR}` is deprecated. Set `openai_base_url` in config.toml instead."
+                )
             };
-            startup_warnings.push(message.to_string());
+            startup_warnings.push(message);
         }
         let effective_openai_base_url = openai_base_url.or(openai_base_url_from_env);
 
-        let mut model_providers = built_in_model_providers();
-        if let Some(base_url) = effective_openai_base_url
-            && let Some(openai_provider) = model_providers.get_mut("openai")
-        {
-            openai_provider.base_url = Some(base_url);
-        }
+        let mut model_providers = built_in_model_providers(effective_openai_base_url);
         // Merge user-defined providers into the built-in list.
         for (key, provider) in cfg.model_providers.into_iter() {
             model_providers.entry(key).or_insert(provider);
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -2272,16 +2272,16 @@ impl Config {
                 }
             });
         if openai_base_url_from_env.is_some() {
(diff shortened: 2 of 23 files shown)


codex-rs/core/tests/common/test_codex_exec.rs
```diff
@@ -23,7 +23,7 @@ impl TestCodexExecBuilder {
     pub fn cmd_with_server(&self, server: &MockServer) -> assert_cmd::Command {
         let mut cmd = self.cmd();
         let base = format!("{}/v1", server.uri());
-        cmd.env("OPENAI_BASE_URL", base);
+        cmd.arg("-c").arg(format!("openai_base_url=\"{base}\""));
```
bolinfest: `base` should be TOML-string-escaped or something?

etraut-openai:
@codex review

codex-rs/core/src/config/mod.rs
```diff
@@ -1687,7 +1690,38 @@ impl Config {
...
 
+        let openai_base_url = cfg.openai_base_url.as_ref().and_then(|value| {
+            let trimmed = value.trim();
+            if trimmed.is_empty() {
+                None
+            } else {
+                Some(trimmed.to_string())
+            }
+        });
+        let openai_base_url_from_env = std::env::var("OPENAI_BASE_URL").ok().and_then(|value| {
```
bolinfest: Arguably we should have a `const` for `"OPENAI_BASE_URL"`.

codex-rs/core/src/config/mod.rs
```diff
@@ -1687,7 +1690,38 @@ impl Config {
...
+                Some(trimmed.to_string())
+            }
+        });
+        if openai_base_url_from_env.is_some() {
+            let message = if openai_base_url.is_some() {
+                "`OPENAI_BASE_URL` is deprecated and ignored because `openai_base_url` is set in config.toml. Remove `OPENAI_BASE_URL` from your environment."
+            } else {
+                "`OPENAI_BASE_URL` is deprecated. Set `openai_base_url` in config.toml instead."
+            };
+            startup_warnings.push(message.to_string());
```
bolinfest: What do you think about having a config option to suppress this warning? I'm thinking about the person who has `OPENAI_BASE_URL` for legitimate reasons when doing other work and will now get nagged about this every time (unless they create a wrapper script that does `unset OPENAI_BASE_URL` before calling `codex`? which they can't even avoid when Codex App runs `codex` on their behalf?)

codex-rs/core/src/config/mod.rs
```diff
@@ -1687,7 +1690,38 @@ impl Config {
...
+        }
+        let effective_openai_base_url = openai_base_url.or(openai_base_url_from_env);
+
         let mut model_providers = built_in_model_providers();
```
bolinfest: Should we change `built_in_model_providers()` to take `effective_openai_base_url` so we don't have to mutate the result?

bolinfest: Or perhaps `built_in_model_providers()` should take `&Config` and much of the logic you have here to derive `effective_openai_base_url` would move into `codex-rs/core/src/model_provider_info.rs` or something closer to `ModelProviderInfo`?

codex-rs/core/src/models_manager/manager.rs
```diff
@@ -64,6 +64,21 @@ impl ModelsManager {
         codex_home: PathBuf,
         auth_manager: Arc<AuthManager>,
         model_catalog: Option<ModelsResponse>,
+    ) -> Self {
+        Self::new_with_provider(
+            codex_home,
+            auth_manager,
+            ModelProviderInfo::create_openai_provider(),
```
bolinfest: It feels like the no-arg version of `create_openai_provider()` should not exist if we want to be sure `openai_base_url` is honored consistently?

codex-rs/core/src/thread_manager.rs
```diff
@@ -681,4 +703,32 @@ mod tests {
...
+        let codex_home = tempdir().expect("create temp codex home");
+        let manager = ThreadManager::new_with_models_provider(
+            codex_home.path().to_path_buf(),
+            auth_manager,
+            SessionSource::Exec,
+            provider,
+            None,
+        );
+
+        let _ = manager.list_models(RefreshStrategy::OnlineIfUncached).await;
```
bolinfest: I guess you're initializing the mock with `models: vec![]` so that's why you're doing `let _` rather than asserting the response value?

codex-rs/core/tests/common/test_codex_exec.rs
```diff
@@ -35,6 +36,10 @@ impl TestCodexExecBuilder {
     }
 }
 
+fn toml_string_literal(value: &str) -> String {
+    serde_json::to_string(value).expect("serialize TOML string literal")
```
bolinfest, line 40: `toml::to_string()`?

codex-rs/core/src/thread_manager.rs
bolinfest: All of the existing call sites pass this as `config.codex_home`, meaning all the existing call sites have `Config` in scope. Perhaps instead of introducing `new_with_models_provider()`, `new()` should change to take `Config` instead of `codex_home: PathBuf` and then all this special-casing of the OpenAI model provider can be internal to `ThreadManager`? As it stands, all of the existing callsites seem to be preceded by:

```rust
        let openai_models_provider = config
            .model_providers
            .get(OPENAI_PROVIDER_ID)
            .cloned()
            .unwrap_or_else(codex_core::ModelProviderInfo::create_openai_provider);
```

so it would be nice to avoid this duplication.

codex-rs/core/src/config/mod.rs
etraut-openai: Yeah, I see your point. I'm not a fan of adding another config key to suppress this warning. Instead, let's assume that if someone sets the new `openai_base_url` config key that they know what they're doing here. No need to warn them about the existence of the env variable. I've changed this to a log rather than a user-visible error.

codex-rs/core/src/config/mod.rs
```diff
@@ -2132,7 +2136,40 @@ impl Config {
         let agent_roles =
             agent_roles::load_agent_roles(&cfg, &config_layer_stack, &mut startup_warnings)?;
 
-        let mut model_providers = built_in_model_providers();
+        let openai_base_url = cfg.openai_base_url.as_ref().and_then(|value| {
+            let trimmed = value.trim();
```
bolinfest: I don't know whether we should feel the need to trim this now that it's no longer an env var: we might as well be strict on inputs?

codex-rs/core/src/config/mod.rs
```diff
@@ -2132,7 +2136,40 @@ impl Config {
...
+                None
+            } else {
+                Some(trimmed.to_string())
+            }
+        });
+        let openai_base_url_from_env =
+            std::env::var(OPENAI_BASE_URL_ENV_VAR)
+                .ok()
+                .and_then(|value| {
+                    let trimmed = value.trim();
```
bolinfest: Could omit this `trim()` too?

codex-rs/core/src/model_provider_info.rs
```diff
@@ -225,17 +226,10 @@ impl ModelProviderInfo {
             .map(Duration::from_millis)
             .unwrap_or(Duration::from_millis(DEFAULT_STREAM_IDLE_TIMEOUT_MS))
     }
-    pub fn create_openai_provider() -> ModelProviderInfo {
+    pub fn create_openai_provider(base_url: Option<String>) -> ModelProviderInfo {
```
bolinfest, line 230: nit: newline before method while you're here?

codex-rs/core/src/config/config_tests.rs
```diff
@@ -4016,7 +4016,7 @@ model_verbosity = "high"
         supports_websockets: false,
     };
     let model_provider_map = {
-        let mut model_provider_map = built_in_model_providers();
+        let mut model_provider_map = built_in_model_providers(None);
```
bolinfest: Seeing this literal `None` in so many places is a little distracting because I have to jump-to-definition to know what the `None` is for. I feel like there are ~three ways to address this:

1. Do `let base_url = None` and then pass in `base_url`.
2. In the call site, do something like `built_in_model_providers(/*base_url=*/ None)` (I would be a big fan of this if we had a clippy check that could verify the comment matched the arg name.)
3. Create two versions of the method, one that takes `String` instead of `Some` and one that doesn't take an argument.

bolinfest (approved):
Nice job! All my comments are small stuff, so feel free to take it or leave it.

I think I'll ask Codex to create that Clippy rule, though...

PR #12040, opened 2026-02-17
tui: exit session on Ctrl+C in cwd change prompt

charley-oai:
## Summary
- change the cwd-change prompt (shown when resuming/forking across different directories) so `Ctrl+C`/`Ctrl+D` exits the session instead of implicitly selecting "Use session directory"
- introduce explicit prompt and resolver exit outcomes so this intent is propagated cleanly through both startup resume/fork and in-app `/resume` flows
- add a unit test that verifies `Ctrl+C` exits rather than selecting an option

## Why
Previously, pressing `Ctrl+C` on this prompt silently picked one of the options, which made it hard to abort. This aligns the prompt with the expected quit behavior.

## Codex author
`codex resume 019c6d39-bbfb-7dc3-8008-1388a054e86d`

### DIFF
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -1452,8 +1452,11 @@ impl App {
                         )
                         .await?
                         {
-                            Some(cwd) => cwd,
-                            None => current_cwd.clone(),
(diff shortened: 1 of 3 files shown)


charley-oai:
@codex review this

sayan-oai (approved):

PR #12044, opened 2026-02-17
don't fail if an npm publish attempt is for an existing version.

iceweasel-oai:


.github/workflows/rust-release.yml
```diff
@@ -611,7 +611,22 @@ jobs:
...
+            publish_output="$("${publish_cmd[@]}" 2>&1)"
+            publish_status=$?
+            set -e
+
+            echo "${publish_output}"
+            if [[ ${publish_status} -eq 0 ]]; then
+              continue
+            fi
+
+            if grep -qiE "previously published|cannot publish over|version already exists" <<< "${publish_output}"; then
```
iceweasel-oai, line 624: according to codex we could see any of these phrases for this type of issue. Most recently we saw the `cannot publish over` variant

dylan-hurd-oai (approved):

PR #12049, opened 2026-02-17
Support implicit skill invocation analytics events

alexsong-oai:
- use `skills_for_cwd` lookup to scope allowed skills and build invocation context for downstream processing
- add detection in `stream_events_utils` to classify tool calls as implicit skill invocations per the proposal (script runners, extensions, `scripts` dirs, and SKILL.md reads)
- deduplicate invocations per turn and emit analytics/OTEL events on the same background queue as explicit invokes


codex-rs/core/src/analytics_client.rs
```diff
@@ -208,7 +218,8 @@ struct CodexAppMetadata {
     turn_id: Option<String>,
     app_name: Option<String>,
     product_client_id: Option<String>,
-    invoke_type: Option<String>,
+    #[serde(rename = "invoke_type")]
```
xl-openai: why rename here?

codex-rs/core/src/stream_events_utils.rs
```diff
@@ -158,6 +256,193 @@ pub(crate) async fn handle_output_item_done(
     Ok(output)
 }
 
+async fn maybe_emit_implicit_skill_invocation(ctx: &mut HandleOutputCtx<'_>, item: &ResponseItem) {
```
xl-openai: can we move this logic to skills folder?

codex-rs/core/src/codex.rs
```diff
@@ -4324,6 +4327,13 @@ pub(crate) async fn run_turn(
         thread_id,
         turn_context.sub_id.clone(),
     );
+    let mut implicit_invocation_context = build_implicit_invocation_context(
```
xl-openai: Can we just build this once (as part of skill manager)?

codex-rs/core/src/stream_events_utils.rs
```diff
@@ -35,16 +47,100 @@ pub(crate) struct OutputItemResult {
...
+        tracking,
+    })
+}
+
+pub(crate) struct HandleOutputCtx<'a> {
     pub sess: Arc<Session>,
     pub turn_context: Arc<TurnContext>,
     pub tool_runtime: ToolCallRuntime,
     pub cancellation_token: CancellationToken,
+    pub implicit_invocation_context: Option<&'a mut ImplicitInvocationContext>,
```
pakrym-oai: can we put this field onto turn context and not have it be mut and force lifetime onto HandleOutputCtx?

codex-rs/core/src/stream_events_utils.rs
```diff
@@ -158,6 +256,193 @@ pub(crate) async fn handle_output_item_done(
...
+    }
+
+    if let Some(candidate) = detect_skill_doc_read(detector, tokens.as_slice(), workdir.as_path()) {
+        return Some(candidate);
+    }
+
+    None
+}
+
+fn parse_implicit_detection_command(
```
pakrym-oai: It's not great that we are reparsing params again. Can we call maybe detec skill invocation in the command handlers themselves?

codex-rs/core/src/stream_events_utils.rs
```diff
@@ -248,3 +533,139 @@ pub(crate) fn response_input_to_response_item(input: &ResponseInputItem) -> Opti
         _ => None,
     }
 }
+
+#[cfg(test)]
+mod tests {
```
pakrym-oai: let's move all of this skill detection logic somewhere close to other skill logic.

codex-rs/core/src/codex.rs
pakrym-oai: Would be nice to avoid having to carry two skill related params too.

alexsong-oai: added implicit_invocation_context as a field of SkillLoadOutcome

codex-rs/core/src/stream_events_utils.rs
alexsong-oai: moved to the handlers

alexsong-oai:
A known issue is that for non-script skills like pdf, the model reads SKILL.md only the first time, and in later turns it may only run generic commands like rg or pdftoppm. The current path/command-based implicit detection cannot such cases. A potential enhancement is to also detect assistant messages like `Use the 'pdf' skill....`

codex-rs/core/src/tools/handlers/unified_exec.rs
```diff
@@ -128,6 +129,13 @@ impl ToolHandler for UnifiedExecHandler {
         let response = match tool_name.as_str() {
             "exec_command" => {
                 let args: ExecCommandArgs = parse_arguments(&arguments)?;
+                maybe_emit_implicit_skill_invocation(
```
xl-openai, line 135: Why do you need to call maybe_emit_implicit_skill_invocation in two places?

codex-rs/core/src/skills/invocation_utils.rs
```diff
@@ -0,0 +1,380 @@
...
+}
+
+#[derive(Default, Debug)]
+pub(crate) struct ImplicitSkillDetector {
+    pub(crate) by_scripts_dir: HashMap<PathBuf, ImplicitSkillCandidate>,
+    pub(crate) by_skill_doc_path: HashMap<PathBuf, ImplicitSkillCandidate>,
+}
+
+#[derive(Debug)]
+pub(crate) struct ImplicitInvocationContext {
```
xl-openai, line 24: we could just use ImplicitSkillDetector directly?

alexsong-oai: Yep, but I want to keep the flexibility in case we want to add more fields to ImplicitInvocationContext.

codex-rs/core/src/tools/handlers/unified_exec.rs
alexsong-oai: they're two different command types, shell vs exec_command.

xl-openai (approved):
LGTM from skills side

pakrym-oai (approved):

PR #12054, opened 2026-02-17
feat(config): add permissions.network proxy config wiring

viyatb-oai:
## Summary

Implements the `ConfigToml.permissions.network` and uses it to populate `NetworkProxyConfig`. We now parse a new nested permissions/network config shape which is converted into the proxy’s runtime config.

When managed requirements exist, we still apply those constraints on top of user settings (so managed policy still wins).

* Cleaned up the old constructor path so it now accepts both user config + managed constraints directly.
* Updated the reload path so live proxy config reloads respect [permissions.network] too, while still supporting the existing top-level [network] format.

### Behavior
- User-defined `[permissions.network]` values are now honored.
- Managed constraints still take effect and are validated against the resulting policy.


codex-rs/core/src/config/mod.rs
```diff
@@ -904,6 +906,11 @@ pub struct ConfigToml {
     /// Sandbox configuration to apply if `sandbox` is `WorkspaceWrite`.
     pub sandbox_workspace_write: Option<SandboxWorkspaceWrite>,
 
+    /// Nested permissions settings.
+    #[serde(default)]
+    #[schemars(skip)]
```
bolinfest: What is the justification for the `schemars` stuff in here?

codex-rs/core/src/config/mod.rs
```diff
@@ -1129,6 +1136,85 @@ impl From<ConfigToml> for UserSavedConfig {
...
+#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq, Eq, JsonSchema)]
+#[schemars(deny_unknown_fields)]
+pub struct PermissionsToml {
+    /// Network proxy settings used by managed network mode.
+    pub network: Option<NetworkToml>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq, Eq, JsonSchema)]
+#[schemars(deny_unknown_fields)]
+pub struct NetworkToml {
```
bolinfest: Maybe this should be in its own file?

codex-rs/core/src/network_proxy_loader.rs
```diff
@@ -48,9 +51,12 @@ async fn build_config_state_with_mtimes() -> Result<(ConfigState, Vec<LayerMtime
     .context("failed to load Codex config")?;
 
     let merged_toml = config_layer_stack.effective_config();
-    let config: NetworkProxyConfig = merged_toml
+    let mut config: NetworkProxyConfig = merged_toml
```
bolinfest: Something about this makes me uncomfortable: I want it to be easy to reason about whether requirements are respected for network configuration.

codex-rs/core/src/config/mod.rs
```diff
@@ -1218,6 +1304,16 @@ pub struct GhostSnapshotToml {
...
 impl ConfigToml {
+    pub(crate) fn network_proxy_config(&self) -> NetworkProxyConfig {
+        let mut config = NetworkProxyConfig::default();
+        if let Some(permissions) = self.permissions.as_ref()
+            && let Some(network) = permissions.network.as_ref()
+        {
+            network.apply_to_network_proxy_config(&mut config);
+        }
+        config
+    }
```
bolinfest: Why should this be available on `ConfigToml`? Again, I want it to be easy to reason about how `requirements.toml` is applied. I feel better when I see "raw TOML" as input and a struct with a bunch of `Constrained<T>` fields on it as output.

Also suggestion to reduce `mut`:

```suggestion
    pub(crate) fn network_proxy_config(&self) -> NetworkProxyConfig {
        if let Some(permissions) = self.permissions.as_ref()
            && let Some(network) = permissions.network.as_ref()
        {
            // Perhaps this should be pushed down to apply_to_network_proxy_config()?
            let mut config = NetworkProxyConfig::default();
            network.apply_to_network_proxy_config(&mut config);
        } else {
            NetworkProxyConfig::default()
        }
    }
```

codex-rs/core/src/config/mod.rs
viyatb-oai: ah this was to not expose the config in the schema but I think we can just not add it to the docs until the full thing lands.

codex-rs/core/src/config/mod.rs
viyatb-oai: as in for the full permissions or just network toml?

codex-rs/core/src/network_proxy_loader.rs
viyatb-oai: fixed

bolinfest (approved):
I think we can start with this, but if we support changing the network between turns, then we will probably have to change things to look more like the other places we use `Constrained<T>` to ensure all modifications continue to honor the constraints.

PR #12061, opened 2026-02-18
Chore: remove response model check and rely on header model for downgrade

shijie-oai:
### Summary
Ensure that we use the model value from the response header only so that we are guaranteed with the correct slug name. We are no longer checking against the model value from response so that we are less likely to have false positive. 

There are two different treatments - for SSE we use the header from the response and for websocket we check top-level events.


codex-rs/core/src/codex.rs
```diff
@@ -2440,7 +2440,9 @@ impl Session {
         server_model: String,
     ) -> bool {
         let requested_model = turn_context.model_info.slug.clone();
-        if server_model == requested_model {
+        let server_model_normalized = server_model.to_ascii_lowercase();
+        let requested_model_normalized = requested_model.to_ascii_lowercase();
+        if server_model_normalized == requested_model_normalized {
```
shijie-oai, line 2445: We check lowercase slug value just for extra safety.

codex-rs/codex-api/src/sse/responses.rs
```diff
@@ -179,22 +180,28 @@ impl ResponsesStreamEvent {
...
+            .and_then(|response| {
+                response
+                    .get("headers")
+                    .and_then(header_openai_model_value_from_json)
+            })
+            .or_else(|| {
+                self.headers
+                    .as_ref()
+                    .and_then(header_openai_model_value_from_json)
+            })
```
shijie-oai: https://github.com/openai/openai/pull/698880/changes#diff-2922dfe194fe2d2985789cff43d60658a78cb43e54420a88641003a74393fdd0  fallback for websocket.

codex-rs/codex-api/src/sse/responses.rs
```diff
@@ -179,22 +180,28 @@ impl ResponsesStreamEvent {
...
-fn extract_server_model(value: &Value) -> Option<String> {
-    value
-        .get("model")
-        .and_then(json_value_as_string)
-        .or_else(|| {
-            value
-                .get("headers")
-                .and_then(header_openai_model_value_from_json)
-        })
-}
```
shijie-oai, line 196: internalize this function.

codex-rs/codex-api/src/sse/responses.rs
```diff
@@ -179,22 +180,28 @@ impl ResponsesStreamEvent {
...
+    /// 2. top-level `headers` for websocket metadata events (for example
+    ///    `codex.response.metadata`).
     pub fn response_model(&self) -> Option<String> {
-        self.response.as_ref().and_then(extract_server_model)
+        self.response
+            .as_ref()
+            .and_then(|response| {
+                response
+                    .get("headers")
+                    .and_then(header_openai_model_value_from_json)
```
dylan-hurd-oai: nit: can we unnest this?

shijie-oai: Updated!

codex-rs/codex-api/src/sse/responses.rs
```diff
@@ -179,22 +180,29 @@ impl ResponsesStreamEvent {
...
+            .and_then(|response| response.get("headers"))
+            .and_then(header_openai_model_value_from_json);
+
+        match response_headers_model {
+            Some(model) => Some(model),
+            None => self
+                .headers
+                .as_ref()
+                .and_then(header_openai_model_value_from_json),
+        }
```
shijie-oai, line 202: Not sure if this is what unested here really mean here but this should be cleaner.

dylan-hurd-oai (commented):
I don't have enough context on this problem to confidently stamp but the general approach LGTM - if we need to urgently block let me know and we can review together

codex-rs/README.md
```diff
@@ -1,3 +1,4 @@
+abcd
```
fouad-openai:
```suggestion
```

codex-rs/README.md
```diff
@@ -99,3 +100,4 @@ This folder is the root of a Cargo workspace. It contains quite a bit of experim
 - [`cli/`](./cli) CLI multitool that provides the aforementioned CLIs via subcommands.
 
 If you want to contribute or inspect behavior in detail, start by reading the module-level `README.md` files under each crate and run the project workspace from the top-level `codex-rs` directory so shared config, features, and build scripts stay aligned.
+avsdfasd
```
fouad-openai:
```suggestion
```

fouad-openai (approved):
approved

PR #10673, opened 2026-02-04
Add feature-gated js_repl polling flow

fjord-oai:
## Summary

Adds a feature-gated polling mode for `js_repl`.

With `js_repl_polling` enabled, `js_repl` can submit async JS executions and immediately return an `exec_id` / `session_id`, while `js_repl_poll` provides incremental logs and final results for that execution.

### Feature flag and tool exposure
- Added `js_repl_polling`
- Gated it behind `js_repl`
- Added `js_repl_poll` to the tool spec only when polling is enabled
- Allowed direct `js_repl_poll` calls in `js_repl_tools_only` mode when polling is enabled

### Polling API
- `js_repl` now accepts polling-related first-line pragma args:
  - `poll=true`
  - `session_id=<id>`
- Added `js_repl_poll`
  - takes `exec_id`
  - optionally takes `yield_time_ms`
  - returns JSON with `exec_id`, `session_id`, `status`, `logs`, `output`, and `error`
- Extended `js_repl_reset`
  - `{"session_id":"..."}` resets a single polling session
  - `{}` resets all `js_repl` kernels

### Polling runtime
- Added session-based polling support in `JsReplManager`
- Enforces one active exec per polling session
- Allows multiple polling sessions in parallel
- Buffers incremental logs for polling and retains completed execs for replay for a bounded window
- Emits output delta events while a polled exec is running

### Semantics and robustness
- Polling mode does not support `timeout_ms`; callers must use `js_repl_poll.yield_time_ms`
- Omitted or short poll waits clamp to the intended idle long-poll behavior
- Keeps poll semantics aligned with background-terminal expectations:
  - return early when logs or completion are available
  - keep polling the same `exec_id` until `status` becomes `completed` or `error`
- Releases a polling session for reuse before async completion-event emission
- Replaced non-polling timeout string matching with a typed local execute error
- Removed redundant kernel-side log buffering in streaming mode and used host-side buffered logs for final poll output fallback

### Parsing / guidance
- Updated freeform `js_repl` parsing to support polling pragma args
- Keeps comma-separated pragma args tolerant for compatibility, while preferring space-separated args in guidance
- Added feature-aware error/help text so polling-specific guidance only appears when polling is enabled
- Updated project-doc and `docs/js_repl.md` guidance for the polling flow

#### [git stack](https://github.com/magus/git-stack-cli)
- ✅ `01` https://github.com/openai/codex/pull/10674
- ✅ `02` https://github.com/openai/codex/pull/10672
[...]

PR #12069, opened 2026-02-18
Stop filtering model tools in js_repl_tools_only mode

fjord-oai:
## Summary
This change removes tool-list filtering in `js_repl_tools_only` mode and relies on the normal model tool descriptions, while still enforcing that tool execution must go through `js_repl` + `codex.tool(...)`.

## Motivation
The previous `js_repl_tools_only` filtering hid most tools from the model request, which diverged from standard tool-list behavior and made signatures less discoverable. I tested that this filtering is not needed, and the model can follow the prompt to only call tools via `js_repl`.

## What Changed
- `filter_tools_for_model(...)` in `core/src/tools/spec.rs` is now a pass-through (no filtering when `js_repl_tools_only` is enabled).
- Updated tests to assert that model tools are not filtered in `js_repl_tools_only` mode.
- Updated dynamic-tool test to assert dynamic tools remain visible in model tool specs.
- Removed obsolete test helper used only by the old filtering assertions.

## Safety / Behavior
- This commit does **not** relax execution policy.
- Direct model tool calls remain blocked in `js_repl_tools_only` mode (except internal `js_repl` tools), and callers are instructed to use `js_repl` + `codex.tool(...)`.

## Testing
- `cargo test -p codex-core js_repl_tools_only`
- Manual rollout validation showed the model can follow the `js_repl` routing instructions without needing filtered tool lists.

- 👉 `1` https://github.com/openai/codex/pull/12069
- ⏳ `2` https://github.com/openai/codex/pull/10673
- ⏳ `3` https://github.com/openai/codex/pull/10670

### DIFF
--- a/codex-rs/core/src/tools/spec.rs
+++ b/codex-rs/core/src/tools/spec.rs
@@ -114,17 +114,6 @@ impl ToolsConfig {
     }
 }
 
-pub(crate) fn filter_tools_for_model(tools: Vec<ToolSpec>, config: &ToolsConfig) -> Vec<ToolSpec> {
-    if !config.js_repl_tools_only {
-        return tools;
-    }
-
-    tools
-        .into_iter()
-        .filter(|spec| matches!(spec.name(), "js_repl" | "js_repl_reset"))
-        .collect()
-}
-
 /// Generic JSON‑Schema subset needed for our tool definitions
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 #[serde(tag = "type", rename_all = "lowercase")]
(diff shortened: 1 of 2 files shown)


codex-rs/core/src/tools/spec.rs
```diff
@@ -112,15 +112,8 @@ impl ToolsConfig {
     }
 }
 
-pub(crate) fn filter_tools_for_model(tools: Vec<ToolSpec>, config: &ToolsConfig) -> Vec<ToolSpec> {
-    if !config.js_repl_tools_only {
-        return tools;
-    }
-
+pub(crate) fn filter_tools_for_model(tools: Vec<ToolSpec>, _config: &ToolsConfig) -> Vec<ToolSpec> {
```
aaronl-openai: do we want to get rid of this entirely since it's a no-op or is it still useful to keep around?

aaronl-openai (approved):
nice!

codex-rs/core/src/tools/spec.rs
fjord-oai: thanks, fixed!

owenlin0 (approved):

PR #12081, opened 2026-02-18
fix(linux-sandbox): mount /dev in bwrap sandbox

### LINKED ISSUE #12056: Allow access of `/dev/urandom` from bubblewrap sandbox

built from `main` (a1b8e34938a26b3910de2abdb2e984aa12624354)

Plus

gpt-5.3-codex

Linux 4.18.0-553.77.1.el8_10.x86_64 x86_64 x86_64

_No response_

I'm trying the experimental feature `bubblewrap` as the sandbox option. This is because `landlock` is not available on my kernel. So far this feature looks good to me with some slight issues.

When a sandbox command needs to read from `/dev/urandom`, it fails.

Some commands (like git) requires secure random number from that device.

I'm building codex directly from `main`.

Set the following in config:
```
features.use_linux_sandbox_bwrap = true
sandbox_mode = "workspace-write"
approval_policy = "never"
```

Tell agent:
```
try to read from `/dev/urandom`
```

Agent ran:
```
head -c 16 /dev/urandom | od -An -tx1
```

Command failed with:
```
head: cannot open '/dev/urandom' for reading: Permission denied
```

Agent successfully reads from secure random.

### Additional information

The error is weird because root is bind read-only while it looks like reading `/dev/urandom` requires full permission.

I read through the current sandbox code (`linux-sandbox/src/bwrap.rs`). Currently it only adds `--dev-bind /dev/null /dev/null`. Consider to add other bindings (`/dev/urandom`, `/dev/zero`, etc.) or maybe use bubblewrap's `--dev /dev` option (not sure if it works with read-only root binding though).

viyatb-oai:
## Summary
- Updates the Linux bubblewrap sandbox args to mount a minimal `/dev` using `--dev /dev` instead of only binding `/dev/null`. tools needing entropy (git, crypto libs, etc.) can fail.

- Changed mount order so `--dev /dev` is added before writable-root `--bind` mounts, preserving writable `/dev/*` submounts like `/dev/shm`

## Why
Fixes sandboxed command failures when reading `/dev/urandom` (and similar standard device-node access).

Fixes https://github.com/openai/codex/issues/12056


codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -148,13 +148,22 @@ fn run_bwrap_with_proc_fallback(
...
     }
 
-    let network_mode = bwrap_network_mode(sandbox_policy, allow_network_for_proxy);
+    if network_mode != BwrapNetworkMode::FullAccess
+        && !preflight_unshare_net_support(sandbox_policy_cwd, sandbox_policy, mount_proc)
+    {
+        eprintln!(
+            "codex-linux-sandbox: bwrap could not configure loopback in a network namespace; retrying without --unshare-net"
+        );
+        network_mode = BwrapNetworkMode::FullAccess;
```
bolinfest: Is this meant to be part of this PR? I don't think we should be silently upgrading to `FullAccess` for `network_mode`: we should just fail?

codex-rs/linux-sandbox/src/bwrap.rs
bolinfest, line 196: Why is this inside this loop?

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -368,4 +369,80 @@ mod tests {
             ]
         );
     }
+
+    #[test]
```
bolinfest: What I would really like to see are integration tests rather than unit tests.

codex-rs/linux-sandbox/src/bwrap.rs
viyatb-oai: For each read-only path, this is adding its own `--ro-bind ...` protection.
```
--ro-bind /dev/null /repo/.git
--ro-bind /dev/null /repo/.codex
```

bolinfest (approved):

PR #12084, opened 2026-02-18
cli: strip trailing memory citation footer in TUI rendering

zuxin-oai:
## Summary
- Strip trailing memory citation footers (`Memory used: ...`) in CLI/TUI markdown rendering only because they are not rendering well. We still show citations in desktop app.
- Apply stripping in both streaming and final render paths.
- Add unit and integration tests for canonical footer and chunked streaming cases.

This keeps raw model/history payloads unchanged while improving CLI display quality.


codex-rs/tui/src/markdown_stream.rs
```diff
@@ -2,6 +2,61 @@ use ratatui::text::Line;
 
 use crate::markdown;
 
+const MEMORY_CITATION_PREFIX: &str = "Memory used:";
```
jif-oai, line 5: this is too common. We need a way stronger pattern

jif-oai (commented):
I'm not sure we should strip it.. we should instead force the model to use a nice rendering

PR #12087, opened 2026-02-18
feat: add Reject approval policy with granular prompt rejection controls

bolinfest:
## Why

We need a way to auto-reject specific approval prompt categories without switching all approvals off.

The goal is to let users independently control:
- sandbox escalation approvals,
- execpolicy `prompt` rule approvals,
- MCP elicitation prompts.

## What changed

- Added a new primary approval mode in `protocol/src/protocol.rs`:

```rust
pub enum AskForApproval {
    // ...
    Reject(RejectConfig),
    // ...
}

pub struct RejectConfig {
    pub sandbox_approval: bool,
    pub rules: bool,
    pub mcp_elicitations: bool,
}
```

- Wired `RejectConfig` semantics through approval paths in `core`:
  - `core/src/exec_policy.rs`
    - rejects rule-driven prompts when `rules = true`
    - rejects sandbox/escalation prompts when `sandbox_approval = true`
    - preserves rule priority when both rule and sandbox prompt conditions are present
  - `core/src/tools/sandboxing.rs`
    - applies `sandbox_approval` to default exec approval decisions and sandbox-failure retry gating
  - `core/src/safety.rs`
    - keeps `Reject { all false }` behavior aligned with `OnRequest` for patch safety
    - rejects out-of-root patch approvals when `sandbox_approval = true`
  - `core/src/mcp_connection_manager.rs`
    - auto-declines MCP elicitations when `mcp_elicitations = true`

- Ensured approval policy used by MCP elicitation flow stays in sync with constrained session policy updates.

- Updated app-server v2 conversions and generated schema/TypeScript artifacts for the new `Reject` shape.

## Verification

Added focused unit coverage for the new behavior in:
- `core/src/exec_policy.rs`
- `core/src/tools/sandboxing.rs`
- `core/src/mcp_connection_manager.rs`
- `core/src/safety.rs`
- `core/src/tools/runtimes/apply_patch.rs`

Key cases covered include rule-vs-sandbox prompt precedence, MCP auto-decline behavior, and patch/sandbox retry behavior under `RejectConfig`.


codex-rs/core/src/exec_policy.rs
```diff
@@ -465,6 +495,20 @@ pub fn render_decision_for_unmatched_command(
...
+                Decision::Allow
+            }
+            SandboxPolicy::ReadOnly { .. } | SandboxPolicy::WorkspaceWrite { .. } => {
+                if sandbox_permissions.requires_escalated_permissions() {
+                    Decision::Prompt
+                } else {
+                    Decision::Allow
+                }
+            }
+        },
```
dylan-hurd-oai, line 518: would be great to have tests for this piece in particular

codex-rs/protocol/src/protocol.rs
```diff
@@ -377,11 +377,60 @@ pub enum AskForApproval {
...
+                ..
+            }
+        )
+    }
+
+    pub const fn rejects_rules_approval(self) -> bool {
+        matches!(self, Self::Reject { rules: true, .. })
+    }
+
+    pub const fn rejects_mcp_elicitations(self) -> bool {
```
dylan-hurd-oai, line 411: I think we'd have some product questions to sort out before actually enforcing this, but should this also match on `Never`?

codex-rs/protocol/src/protocol.rs
```diff
@@ -377,11 +377,60 @@ pub enum AskForApproval {
...
+
     /// Never ask the user to approve commands. Failures are immediately returned
     /// to the model, and never escalated to the user for approval.
     Never,
 }
 
+impl AskForApproval {
+    /// Backward-compatible defaults for `approval_policy = "reject"` in
+    /// `config.toml`.
+    pub const fn reject_defaults() -> Self {
```
dylan-hurd-oai: nit: we use `new_workspace_write_policy()` etc. in SandboxPolicy, should we use a similar naming convention here?

dylan-hurd-oai (approved):
couple of non-blocking comments!

codex-rs/protocol/src/protocol.rs
bolinfest: Done.

codex-rs/protocol/src/protocol.rs
bolinfest: Thinking about this, these methods are specific to the Reject variant, so I introduced the RejectConfig struct and moved these methods to RejectConfig.

codex-rs/core/src/exec_policy.rs
bolinfest: added!

PR #12101, opened 2026-02-18
client side modelinfo overrides

sayan-oai:
TL;DR
Add top-level `model_catalog_json` config support so users can supply a local model catalog override from a JSON file path (including adding new models) without backend changes.

### Problem
Codex previously had no clean client-side way to replace/overlay model catalog data for local testing of model metadata and new model entries.

### Fix
- Add top-level `model_catalog_json` config field (JSON file path).
- Apply catalog entries when resolving `ModelInfo`:
  1. Base resolved model metadata (remote/fallback)
  2. Catalog overlay from `model_catalog_json`
  3. Existing global top-level overrides (`model_context_window`, `model_supports_reasoning_summaries`, etc.)

### Note
Will revisit per-field overrides in a follow-up

### Tests
Added tests


codex-rs/core/src/config/mod.rs
```diff
@@ -1227,6 +1269,51 @@ pub struct GhostSnapshotToml {
...
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq, JsonSchema)]
+#[schemars(deny_unknown_fields)]
+pub struct ModelInfoConfigToml {
+    /// Optional full catalog overlay applied on top of bundled and fetched model metadata.
+    /// Accepts:
+    /// - a path to a JSON file containing `ModelsResponse`,
+    /// - a JSON string containing a serialized `ModelsResponse`, or
+    /// - a TOML inline `ModelsResponse` object.
```
sayan-oai: we could drop support for JSON string and get rid of the custom deserialize, but that would make it harder for codex web to supply a full `ModelInfo` object if files cant be put into containers.

codex-rs/protocol/src/openai_models.rs
```diff
@@ -213,8 +214,14 @@ const fn default_effective_context_window_percent() -> i64 {
 }
 
 /// Model metadata returned by the Codex backend `/models` endpoint.
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, TS, JsonSchema)]
+///
+/// `#[derive(ModelPatch)]` auto-generates `ModelInfoPatch`, which is used by `model_info.overrides`
+/// in config to apply per-field updates. All fields are patchable unless marked `#[model_patch(skip)]`.
+/// Patch updates are set-only.
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, TS, JsonSchema, ModelPatch)]
```
sayan-oai: use a macro to generate the partial `ModelInfoPatch` type from `ModelInfo`. a bit complex, but avoids statically defining a mirrored type in code and having ugly apply logic like was drafted [here](https://github.com/openai/codex/pull/11404/changes#diff-6b7cffb6cfba0fa10a78348a1947a1f579f090cb2fd86c0da709c77f2d416c27R263).

to avoid either approach, we could support passing overrides only via a `catalog_json` file path or a predetermined struct with a subset of `ModelInfo` fields that could be overridden. this would force codex web to put a file with the updated models catalog into the container though.

codex-rs/core/src/models_manager/manager.rs
```diff
@@ -77,6 +77,9 @@ impl ModelsManager {
         if let Err(err) = self.refresh_available_models(refresh_strategy).await {
             error!("failed to refresh available models: {err}");
         }
+        // todo(sayan): Make preset/list/default-model flows config-aware so
+        // `model_info.catalog_json` and `model_info.overrides` are reflected in
+        // picker/app-server model metadata as well (today only get_model_info uses them).
```
sayan-oai: can do this as part of this pr or in a follow up, but it doesnt block the eval/codex web use cases because only UX currently misses the overrides.

codex-rs/core/src/models_manager/manager.rs
aibrahim-oai: Should we do the loading inside `load_remote_models_from_file`? we decide which file to load from inside it.

sayan-oai: this would make the model catalog a replacement for the existing models.json rather than an overlay, but i think thats fair.

pakrym-oai (approved):

PR #12140, opened 2026-02-18
Refactor network approvals to host/protocol/port scope

viyatb-oai:
## Summary
Simplify network approvals by removing per-attempt proxy correlation and moving to session-level approval dedupe keyed by (host, protocol, port). Instead of encoding attempt IDs into proxy credentials/URLs, we now treat approvals as a destination policy decision.

- Concurrent calls to the same destination share one approval prompt.
- Different destinations (or same host on different ports) get separate prompts.
- Allow once approves the current queued request group only.
- Allow for session caches that (host, protocol, port) and auto-allows future matching requests.
- Never policy continues to deny without prompting.

Example:
- 3 calls: 
  - a.com (line 443)
  - b.com (line 443)
  - a.com (line 443)
=> 2 prompts total (a, b), second a waits on the first decision.
- a.com:80 is treated separately from a.com line 443

## Testing
- `just fmt` (in `codex-rs`)
- `cargo test -p codex-core tools::network_approval::tests`
- `cargo test -p codex-core` (unit tests pass; existing integration-suite failures remain in this environment)


codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -104,6 +105,20 @@ use tracing::error;
 
 type JsonValue = serde_json::Value;
 
+struct CommandExecutionApprovalPresentation {
+    network_approval_context: Option<V2NetworkApprovalContext>,
+    command: Option<String>,
+    cwd: Option<PathBuf>,
+    command_actions: Option<Vec<V2ParsedCommand>>,
+    completion_item: Option<CommandExecutionCompletionItem>,
+}
```
dylan-hurd-oai, line 114: I'm not sure about this structure - shouldn't it be an enum?

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -257,7 +273,31 @@ pub(crate) async fn apply_bespoke_event_handling(
...
+                            command_actions: command_actions.clone(),
+                        };
+                        CommandExecutionApprovalPresentation {
+                            network_approval_context: None,
+                            command: Some(command_string),
+                            cwd: Some(cwd.clone()),
+                            command_actions: Some(command_actions.clone()),
+                            completion_item: Some(completion_item),
+                        }
+                    };
```
dylan-hurd-oai, line 309: We're likely going to need client updates to handle this properly!

codex-rs/core/src/unified_exec/mod.rs
```diff
@@ -145,7 +145,7 @@ struct ProcessEntry {
     process_id: String,
     command: Vec<String>,
     tty: bool,
-    network_attempt_id: Option<String>,
+    network_approval_registration_id: Option<String>,
```
dylan-hurd-oai: Can you explain why we're renaming to registration id?

dylan-hurd-oai (commented):
Looks directionally reasonable - couple of questions!

codex-rs/core/src/unified_exec/mod.rs
viyatb-oai: ah just no better name came to mind - i think `network_approval_id` is much cleaner and easier to understand

codex-rs/app-server/src/bespoke_event_handling.rs
viyatb-oai: done

dylan-hurd-oai: I think what we want here is tuple-style:

```
enum CommandExecutionApprovalPresentation {
    Network(V2NetworkApprovalContext),
    Command(CommandExecutionCompletionItem),
}
```

codex-rs/core/src/unified_exec/async_watcher.rs
```diff
@@ -139,43 +139,6 @@ pub(crate) fn spawn_exit_watcher(
...
-                            .unified_exec_manager
-                            .release_process_id(&process_id)
-                            .await;
-                        break;
-                    }
-                }
-            }
-        }
-    });
-}
```
dylan-hurd-oai, line 177: why don't we need this anymore?

dylan-hurd-oai (approved):
some non-blocking comments

codex-rs/core/src/unified_exec/async_watcher.rs
viyatb-oai: no need for the watcher anymore because denial is enforced in-band at the proxy decision point now, not via a polled deferred outcome in unified-exec. The process manager only keeps the registration ID for lifecycle cleanup.

PR #12162, opened 2026-02-18
Fix stale config docs link in deprecation notices

senkushal:
Updates deprecation notice docs URL to the current config reference and updates the related test.
This fixes issue: https://github.com/openai/codex/issues/12123


etraut-openai:
Thanks for the fix. This was on my to-do list for this morning. :)

etraut-openai:
Actually, this isn't the correct link. It should be changed to `https://developers.openai.com/codex/config-basic#feature-flags`. I'll take care of it.

PR #12195, opened 2026-02-18
Fix skills lookup for symlinked cwd

shepting:
## Summary
- canonicalize cwd paths when matching skills entries so symlinked repo paths are recognized
- keep fast path for exact path matches
- add unix-only test for symlinked cwd matching

## Background
Our company started with Claude Code (hence the `.claude/skills` directory) and later added Codex support by symlinking `.agents/skills -> .claude/skills`. We run Codex from a symlinked repo path (e.g. `~/workspace/apps`), so skills load and can run, but the `/skills` list makes it look like skills are missing because the cwd equality check compared the symlink path to the real path. Canonicalizing both sides fixes the mismatch.

## Testing
- `just fmt`
- `cargo test -p codex-tui`

## Screenshot
Before (no skills showing here):
Screenshot 2026-02-18 at 3 34 37 PM

After (all skills showing):
Screenshot 2026-02-18 at 3 33 24 PM

## Please Review
@jif-oai @JaviSoto


etraut-openai:
@codex review

PR #12217, opened 2026-02-19
fix(absolute-path): avoid panic in parent() under EMFILE

### LINKED ISSUE #12216: Panic in AbsolutePathBuf::parent() under EMFILE (Too many open files)
## Summary
When file descriptors are exhausted (`EMFILE` / "Too many open files"), Codex can panic in `AbsolutePathBuf::parent()`.

Observed panic:

- `parent of AbsolutePathBuf must be absolute: Os { code: 24, ... "Too many open files" }`
- Followed by: `in-flight tool future failed during drain: ... JoinError::Panic(...)`

## Why this happens
`AbsolutePathBuf::parent()` currently does this:

- calls `self.0.parent()`
- then re-validates via `Self::from_absolute_path(p).expect(...)`

`from_absolute_path()` uses `path_absolutize::absolutize()`, and `path-absolutize` calls `std::env::current_dir()` even for already-absolute paths.

Under `EMFILE`, `current_dir()` fails, so the `expect(...)` in `parent()` panics.

## Impact
- An otherwise recoverable resource-exhaustion condition is escalated into a process panic.
- This is easiest to hit during high parallelism (many sub-agents/tool tasks).

## Repro notes
This was observed with low soft `ulimit -n` (`256`) and heavy parallel tool/sub-agent activity.

## Proposed fix
In `AbsolutePathBuf::parent()`, avoid re-absolutizing and avoid `expect`:

- construct directly from known parent path: `Self(p.to_path_buf())`

This preserves the invariant (parent of an absolute path is absolute) and removes the unnecessary `current_dir()` dependency on the hot path.

I have an isolated PR ready with this change plus tests.

BleedingDev:
## Summary
Fixes a panic path when file descriptors are exhausted (`EMFILE` / "Too many open files").

`AbsolutePathBuf::parent()` was re-absolutizing the parent path with `from_absolute_path(...).expect(...)`. That call path goes through `path-absolutize`, which calls `current_dir()`. Under `EMFILE`, `current_dir()` fails and the `expect` panics.

This change makes `parent()` construct directly from the known parent path:

- from: `Self::from_absolute_path(p).expect(...)`
- to: `Self(p.to_path_buf())`

## Why this is safe
`self` is already an `AbsolutePathBuf`, so `self.0.parent()` is already absolute when present. Re-normalizing is unnecessary for this operation.

## Tests
- Added unit test that parent of an absolute child remains absolute.
- Added unix-only test that root has no parent.

Run locally:
- `just fmt` (in `codex-rs`)
- `cargo test -p codex-utils-absolute-path`

## Related
Closes #12216.

### DIFF
--- a/codex-rs/utils/absolute-path/src/lib.rs
+++ b/codex-rs/utils/absolute-path/src/lib.rs
@@ -68,10 +68,7 @@ impl AbsolutePathBuf {
     }
 
     pub fn parent(&self) -> Option<Self> {
-        self.0.parent().map(|p| {
-            #[expect(clippy::expect_used)]
-            Self::from_absolute_path(p).expect("parent of AbsolutePathBuf must be absolute")
-        })
+        self.0.parent().map(|p| Self(p.to_path_buf()))
     }
 
     pub fn as_path(&self) -> &Path {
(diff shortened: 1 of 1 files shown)


BleedingDev:
recheck

etraut-openai:

PR #12218, opened 2026-02-19
app-server: harden disconnect cleanup paths

maxj-oai:
Hardens codex-rs/app-server connection lifecycle and outbound routing for websocket clients. Fixes some FUD I was having

  - Added per-connection disconnect signaling (CancellationToken) for websocket transports.
  - Split websocket handling into independent inbound/outbound tasks coordinated by cancellation.
  - Changed outbound routing so websocket connections use non-blocking try_send; slow/full websocket writers are disconnected instead of stalling broadcast delivery.
  - Kept stdio behavior blocking-on-send (no forced disconnect) so local stdio clients are not dropped when queues are temporarily full.
  - Simplified outbound router flow by removing deferred pending_closed_connections handling.
  - Added guards to drop incoming response/notification/error messages from unknown connections.
  - Fixed listener teardown race in thread listener tasks using a listener_generation check so stale tasks do not clear newer listeners.

Fixes https://linear.app/openai/issue/CODEX-4966/multiclient-handle-slow-notification-consumers

  ## Tests

  Added/updated transport tests covering:

  - broadcast does not block on a slow/full websocket connection
  - stdio connection waits instead of disconnecting on full queue

I (maxj) have tested manually and will retest before landing


codex-rs/app-server/src/lib.rs
```diff
@@ -370,61 +371,43 @@ pub async fn run_main_with_transport(
...
                     let Some(envelope) = envelope else {
                         break;
                     };
-                    let disconnected_connections =
-                        route_outgoing_envelope(&mut outbound_connections, envelope).await;
-                    pending_closed_connections.extend(disconnected_connections);
-                }
-            }
-
-            while let Some(connection_id) = pending_closed_connections.front().copied() {
```
celia-oai, line 448: is it fine that we no long have this for stdio code path? seems like a behavior change?

codex-rs/app-server/src/transport.rs
```diff
@@ -341,7 +431,11 @@ async fn run_websocket_connection(
                         }
                     }
                     Some(Ok(WebSocketMessage::Ping(payload))) => {
-                        if websocket_writer.send(WebSocketMessage::Pong(payload)).await.is_err() {
+                        if writer_control_tx
+                            .send(WebSocketMessage::Pong(payload))
+                            .await
```
celia-oai: we could still get stuck here right if the channel is full?

celia-oai (approved):

codex-rs/app-server/src/lib.rs
maxj-oai: yeah this is intended - we replace this vector based system with an explicit cancel signal. but it only ever applied to the connection-oriented transports. the stdio transport is always 1-1 with a client, so any disconnect will close the app server.

codex-rs/app-server/src/transport.rs
maxj-oai: ah yes - this should be try send - will fix

PR #12220, opened 2026-02-19
feat(windows-sandbox): add network proxy support

viyatb-oai:
## Summary

This PR makes Windows sandbox proxying enforceable by routing proxy-only runs through the existing `offline` sandbox user and reserving direct network access for the existing `online` sandbox user.

In brief:

- if a Windows sandbox run should be proxy-enforced, we run it as the `offline` user
- the `offline` user gets firewall rules that block direct outbound traffic and only permit the configured localhost proxy path
- if a Windows sandbox run should have true direct network access, we run it as the `online` user
- no new sandbox identity is introduced

This brings Windows in line with the intended model: proxy use is not just env-based, it is backed by OS-level egress controls. Windows already has two sandbox identities:

- `offline`: intended to have no direct network egress
- `online`: intended to have full network access

This PR makes proxy-enforced runs use that model directly.

### Proxy-enforced runs

When proxy enforcement is active:

- the run is assigned to the `offline` identity
- setup extracts the loopback proxy ports from the sandbox env
- Windows setup programs firewall rules for the `offline` user that:
  - block all non-loopback outbound traffic
  - block loopback UDP
  - block loopback TCP except for the configured proxy ports
  - optionally allow broader localhost access when `allow_local_binding=1`

So the sandboxed process can only talk to the local proxy. It cannot open direct outbound sockets or do local UDP-based DNS on its own.The proxy then performs the real outbound network access outside that restricted sandbox identity.

### Direct-network runs

When proxy enforcement is not active and full network access is allowed:

- the run is assigned to the `online` identity
- no proxy-only firewall restrictions are applied
- the process gets normal direct network access

### Unelevated vs elevated

The restricted-token / unelevated path cannot enforce per-identity firewall policy by itself.

So for Windows proxy-enforced runs, we transparently use the logon-user sandbox path under the hood, even if the caller started from the unelevated mode. That keeps enforcement real instead of best-effort.


codex-rs/windows-sandbox-rs/src/identity.rs
```diff
@@ -153,6 +161,23 @@ pub fn require_logon_sandbox_creds(
             None
         }
     };
+    if network_identity.uses_offline_identity() {
+        if let (Some(marker), Some(_)) = (&existing_marker, &identity) {
+            if marker.proxy_ports != desired_offline_proxy_settings.proxy_ports
+                || marker.allow_local_binding
```
iceweasel-oai: I don't love checking specific fields in the marker to determine if the setup needs to be re-run. That's what the SETUP_VERSION is supposed to be for.

That said, given that these proxy_settings could change post set-up, it might be unavoidable.

iceweasel-oai (commented):
overall looks good.

One downside is that this will cause an unexpected elevated setup UAC prompt for all users. But that is an existing limitation where we don't have a UI treatment letting people know that they need to re-setup their sandbox.

iceweasel-oai (approved):

codex-rs/windows-sandbox-rs/src/identity.rs
iceweasel-oai: I kinda want to get rid of the marker at some point anyway, and have the setup basically inspect all the things it cares about and determine if anything is stale. One day.

viyatb-oai: Folded this into `SetupMarker` so `identity.rs` no longer inspects marker fields directly when deciding whether setup is stale. We still need the runtime comparison because `SETUP_VERSION` covers schema/behavior revisions, but the offline firewall allowlist comes from the current proxy env and can change without a setup binary version bump. I also added focused coverage for the online-ignore/offline-drift cases.

PR #12228, opened 2026-02-19
feat: add configurable write_stdin timeout

jif-oai:
Add max timeout as config for `write_stdin`. This is only used for empty `write_stdin`. 

Also increased the default value from 30s to 5mins.

### DIFF
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -51,6 +51,8 @@ use crate::protocol::ReadOnlyAccess;
 use crate::protocol::SandboxPolicy;
 #[cfg(target_os = "macos")]
 use crate::seatbelt_permissions::MacOsSeatbeltProfileExtensions;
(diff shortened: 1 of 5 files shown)


codex-rs/core/src/config/mod.rs
```diff
@@ -380,6 +382,10 @@ pub struct Config {
     /// If set to `true`, used only the experimental unified exec tool.
     pub use_experimental_unified_exec_tool: bool,
 
+    /// Maximum poll window for background terminal output (`write_stdin`), in milliseconds.
+    /// Default: `300000` (5 minutes).
+    pub background_terminal_timeout: u64,
```
pakrym-oai: We might want a different name. Right now it reads as the timeout for the entire terminal.

pakrym-oai: maximum_terminal_wait_time ?

pakrym-oai (approved):

PR #12252, opened 2026-02-19
Fix compaction context reinjection and model baselines

charley-oai:
## Summary
- move regular-turn context diff/full-context persistence into `run_turn` so pre-turn compaction runs before incoming context updates are recorded
- after successful pre-turn compaction, rely on a cleared `reference_context_item` to trigger full context reinjection on the follow-up regular turn (manual `/compact` keeps replacement history summary-only and also clears the baseline)
- preserve `<model_switch>` when full context is reinjected, and inject it *before* the rest of the full-context items
- scope `reference_context_item` and `previous_model` to regular user turns only so standalone tasks (`/compact`, shell, review, undo) cannot suppress future reinjection or `<model_switch>` behavior
- make context-diff persistence + `reference_context_item` updates explicit in the regular-turn path, with clearer docs/comments around the invariant
- stop persisting local `/compact` `RolloutItem::TurnContext` snapshots (only regular turns persist `TurnContextItem` now)
- simplify resume/fork previous-model/reference-baseline hydration by looking up the last surviving turn context from rollout lifecycle events, including rollback and compaction-crossing handling
- remove the legacy fallback that guessed from bare `TurnContext` rollouts without lifecycle events
- update compaction/remote-compaction/model-visible snapshots and compact test assertions (including remote compaction mock response shape)

## Why
We were persisting incoming context items before spawning the regular turn task, which let pre-turn compaction requests accidentally include incoming context diffs without the new user message. Fixing that exposed follow-on baseline issues around `/compact`, resume/fork, and standalone tasks that could cause duplicate context injection or suppress `<model_switch>` instructions.

This PR re-centers the invariants around regular turns:
- regular turns persist model-visible context diffs/full reinjection and update the `reference_context_item`
- standalone tasks do not advance those regular-turn baselines
- compaction clears the baseline when replacement history may have stripped the referenced context diffs

## Follow-ups (TODOs left in code)
- `TODO(ccunningham)`: fix rollback/backtracking baseline handling more comprehensively
- `TODO(ccunningham)`: include pending incoming context items in pre-turn compaction threshold estimation
- `TODO(ccunningham)`: inject updated personality spec alongside `<model_switch>` so some model-switch paths can avoid forced full reinjection
- `TODO(ccunningham)`: review task turn lifecycle (`TurnStarted`/`TurnComplete`) behavior and emit task-start context diffs for task types that should have them (excluding `/compact`)

## Validation
- `just fmt`
- CI should cover the updated compaction/resume/model-visible snapshot expectations and rollout-hydration behavior
- I did **not** rerun the full local test suite after the latest resume-lookup / rollout-persistence simplifications

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -18,7 +18,7 @@ use crate::analytics_client::build_track_events_context;
 use crate::apps::render_apps_section;
 use crate::commit_attribution::commit_message_trailer_instruction;
 use crate::compact;
-use crate::compact::CompactCallsite;
+use crate::compact::InjectTurnContextBeforeLastUserMessage;
 use crate::compact::run_inline_auto_compact_task;
 use crate::compact::should_use_remote_compact_task;
 use crate::compact_remote::run_inline_remote_auto_compact_task;
@@ -4628,9 +4628,13 @@ pub(crate) async fn run_turn(
 
                 // as long as compaction works well in getting us way below the token limit, we shouldn't worry about being in an infinite loop.
                 if token_limit_reached && needs_follow_up {
-                    if run_auto_compact(&sess, &turn_context, CompactCallsite::MidTurn)
-                        .await
-                        .is_err()
+                    if run_auto_compact(
+                        &sess,
+                        &turn_context,
+                        InjectTurnContextBeforeLastUserMessage::TurnContext,
+                    )
+                    .await
+                    .is_err()
                     {
                         return None;
                     }
@@ -4757,7 +4761,12 @@ async fn run_pre_sampling_compact(
         .unwrap_or(i64::MAX);
     // Compact if the total usage tokens are greater than the auto compact limit
     if total_usage_tokens >= auto_compact_limit {
-        run_auto_compact(sess, turn_context, CompactCallsite::PreTurn).await?;
+        run_auto_compact(
+            sess,
+            turn_context,
+            InjectTurnContextBeforeLastUserMessage::DontInjectTurnContext,
+        )
+        .await?;
         compacted = true;
     }
     Ok(compacted)
@@ -4800,7 +4809,7 @@ async fn maybe_run_previous_model_inline_compact(
         run_auto_compact(
             sess,
             &previous_model_turn_context,
-            CompactCallsite::PreTurnModelSwitch,
+            InjectTurnContextBeforeLastUserMessage::DontInjectTurnContext,
         )
         .await?;
         return Ok(true);
@@ -4811,13 +4820,22 @@ async fn maybe_run_previous_model_inline_compact(
 async fn run_auto_compact(
     sess: &Arc<Session>,
     turn_context: &Arc<TurnContext>,
-    callsite: CompactCallsite,
+    inject_turn_context_before_last_user_message: InjectTurnContextBeforeLastUserMessage,
 ) -> CodexResult<()> {
     if should_use_remote_compact_task(&turn_context.provider) {
-        run_inline_remote_auto_compact_task(Arc::clone(sess), Arc::clone(turn_context), callsite)
-            .await?;
+        run_inline_remote_auto_compact_task(
+            Arc::clone(sess),
+            Arc::clone(turn_context),
+            inject_turn_context_before_last_user_message,
+        )
+        .await?;
     } else {
-        run_inline_auto_compact_task(Arc::clone(sess), Arc::clone(turn_context), callsite).await?;
+        run_inline_auto_compact_task(
+            Arc::clone(sess),
+            Arc::clone(turn_context),
+            inject_turn_context_before_last_user_message,
+        )
+        .await?;
     }
     Ok(())
 }
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -5646,6 +5646,9 @@ async fn try_run_sampling_request(
     prompt: &Prompt,
     cancellation_token: CancellationToken,
 ) -> CodexResult<SamplingRequestResult> {
+    // Persist one TurnContext marker per sampling request (not just per user turn) so rollout
+    // analysis can reconstruct API-turn boundaries. `run_turn` persists model-visible context
+    // diffs/full reinjection earlier in the same regular turn before reaching this path.
     let rollout_item = RolloutItem::TurnContext(turn_context.to_turn_context_item());
 
     feedback_tags!(
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -3946,10 +3946,10 @@ mod handlers {
         let turn_context = sess.new_default_turn_with_sub_id(sub_id).await;
 
         let mut history = sess.clone_history().await;
-        // TODO: Follow up on rollback/backtracking baseline handling. Truncating
-        // history should also invalidate/recompute `reference_context_item` and
-        // `previous_model` so the next regular turn replays any dropped context
-        // diffs / model-switch instructions.
+        // TODO(ccunningham): Fix rollback/backtracking baseline handling.
+        // Truncating history should also invalidate/recompute
+        // `reference_context_item` and `previous_model` so the next regular
+        // turn replays any dropped context diffs / model-switch instructions.
         history.drop_last_n_user_turns(num_turns);
 
         // Replace with the raw items. We don't want to replace with a normalized
@@ -4325,6 +4325,10 @@ pub(crate) async fn run_turn(
         collaboration_mode_kind: turn_context.collaboration_mode.mode,
     });
     sess.send_event(&turn_context, event).await;
+    // TODO(ccunningham): Pre-turn compaction runs before context updates and the
+    // new user message are recorded. Estimate pending incoming items (context
+    // diffs/full reinjection + user input) and trigger compaction preemptively
+    // when they would push the thread over the compaction threshold.
     let pre_sampling_compacted = match run_pre_sampling_compact(&sess, &turn_context).await {
         Ok(compacted) => compacted,
         Err(_) => {
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -18,7 +18,7 @@ use crate::analytics_client::build_track_events_context;
 use crate::apps::render_apps_section;
 use crate::commit_attribution::commit_message_trailer_instruction;
 use crate::compact;
-use crate::compact::InjectTurnContextBeforeLastUserMessage;
+use crate::compact::InitialContextInjection;
 use crate::compact::run_inline_auto_compact_task;
 use crate::compact::should_use_remote_compact_task;
 use crate::compact_remote::run_inline_remote_auto_compact_task;
@@ -4631,7 +4631,7 @@ pub(crate) async fn run_turn(
                     if run_auto_compact(
                         &sess,
                         &turn_context,
-                        InjectTurnContextBeforeLastUserMessage::TurnContext,
+                        InitialContextInjection::BeforeLastUserMessage,
                     )
                     .await
                     .is_err()
@@ -4761,12 +4761,7 @@ async fn run_pre_sampling_compact(
         .unwrap_or(i64::MAX);
     // Compact if the total usage tokens are greater than the auto compact limit
     if total_usage_tokens >= auto_compact_limit {
-        run_auto_compact(
-            sess,
-            turn_context,
-            InjectTurnContextBeforeLastUserMessage::DontInjectTurnContext,
-        )
-        .await?;
+        run_auto_compact(sess, turn_context, InitialContextInjection::DoNotInject).await?;
         compacted = true;
     }
     Ok(compacted)
@@ -4809,7 +4804,7 @@ async fn maybe_run_previous_model_inline_compact(
         run_auto_compact(
             sess,
             &previous_model_turn_context,
-            InjectTurnContextBeforeLastUserMessage::DontInjectTurnContext,
+            InitialContextInjection::DoNotInject,
         )
         .await?;
         return Ok(true);
@@ -4820,20 +4815,20 @@ async fn maybe_run_previous_model_inline_compact(
 async fn run_auto_compact(
     sess: &Arc<Session>,
     turn_context: &Arc<TurnContext>,
-    inject_turn_context_before_last_user_message: InjectTurnContextBeforeLastUserMessage,
+    initial_context_injection: InitialContextInjection,
 ) -> CodexResult<()> {
     if should_use_remote_compact_task(&turn_context.provider) {
         run_inline_remote_auto_compact_task(
             Arc::clone(sess),
             Arc::clone(turn_context),
-            inject_turn_context_before_last_user_message,
+            initial_context_injection,
         )
         .await?;
     } else {
         run_inline_auto_compact_task(
             Arc::clone(sess),
             Arc::clone(turn_context),
-            inject_turn_context_before_last_user_message,
+            initial_context_injection,
         )
         .await?;
     }
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -6678,8 +6678,10 @@ mod tests {
             .await;
 
         assert_eq!(
-            session.reference_context_item().await,
-            Some(previous_context_item)
+            serde_json::to_value(session.reference_context_item().await)
+                .expect("serialize seeded reference context item"),
+            serde_json::to_value(Some(previous_context_item))
+                .expect("serialize expected reference context item")
         );
     }
 
@@ -6729,7 +6731,7 @@ mod tests {
             }))
             .await;
 
-        assert_eq!(session.reference_context_item().await, None);
+        assert!(session.reference_context_item().await.is_none());
     }
 
     #[tokio::test]
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -1605,15 +1605,11 @@ impl Session {
                 let previous_model =
                     previous_regular_turn_context_item.map(|ctx| ctx.model.clone());
                 let curr = turn_context.model_info.slug.as_str();
-                let can_seed_reference_context_item = !crossed_compaction_after_turn
-                    && previous_regular_turn_context_item.is_some_and(|ctx| ctx.model == curr);
-                let reference_context_item = if can_seed_reference_context_item {
+                let reference_context_item = if !crossed_compaction_after_turn {
                     previous_regular_turn_context_item.cloned()
                 } else {
                     // Keep the baseline empty when compaction may have stripped the referenced
-                    // context diffs, or when resuming on a different model. The first resumed
-                    // turn will then do full reinjection (which also preserves personality spec
-                    // when the model changes).
+                    // context diffs so the first resumed regular turn fully reinjects context.
                     None
                 };
                 {
@@ -2798,6 +2794,9 @@ impl Session {
                     turn_context,
                 )
             {
+                // TODO(ccunningham): When a model switch changes the effective personality
+                // instructions, inject the updated personality spec alongside <model_switch>
+                // here so resume/model-switch paths can avoid forcing full reinjection.
                 initial_context.insert(0, model_switch_item);
             }
             initial_context
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -1715,6 +1715,9 @@ impl Session {
     ///
     /// Returns `(None, false)` when no completed turn with a persisted
     /// `TurnContextItem` is present in the lifecycle-aware rollout view.
+    // TODO(ccunningham): Simplify this lookup by sharing rollout traversal/rollback application
+    // with `reconstruct_history_from_rollout` so resume/fork baseline hydration does not need a
+    // second bespoke rollout scan.
     fn last_rollout_regular_turn_context_lookup(
         rollout_items: &[RolloutItem],
     ) -> (Option<&TurnContextItem>, bool) {
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -2749,7 +2749,6 @@ impl Session {
         &self,
         turn_context: &TurnContext,
         previous_user_turn_model: Option<&str>,
-        emit_raw_events: bool,
     ) {
         let reference_context_item = self.reference_context_item().await;
         let settings_update_items = self.build_settings_update_items(
@@ -2776,13 +2775,8 @@ impl Session {
             settings_update_items
         };
         if !context_items.is_empty() {
-            if emit_raw_events {
-                self.record_conversation_items(turn_context, &context_items)
-                    .await;
-            } else {
-                self.record_into_history(&context_items, turn_context).await;
-                self.persist_rollout_response_items(&context_items).await;
-            }
+            self.record_conversation_items(turn_context, &context_items)
+                .await;
         }
 
         let mut state = self.state.lock().await;
@@ -4406,7 +4400,6 @@ pub(crate) async fn run_turn(
     sess.record_context_updates_and_set_reference_context_item(
         turn_context.as_ref(),
         previous_model.as_deref(),
-        true,
     )
     .await;
 
@@ -6602,14 +6595,14 @@ mod tests {
         assert_eq!(expected, history_before_seed.raw_items());
 
         session
-            .record_context_updates_and_set_reference_context_item(&turn_context, None, false)
+            .record_context_updates_and_set_reference_context_item(&turn_context, None)
             .await;
         expected.extend(session.build_initial_context(&turn_context).await);
         let history_after_seed = session.clone_history().await;
         assert_eq!(expected, history_after_seed.raw_items());
 
         session
-            .record_context_updates_and_set_reference_context_item(&turn_context, None, false)
+            .record_context_updates_and_set_reference_context_item(&turn_context, None)
             .await;
         let history_after_second_seed = session.clone_history().await;
         assert_eq!(expected, history_after_second_seed.raw_items());
@@ -7926,7 +7919,7 @@ mod tests {
      {
         let (session, turn_context) = make_session_and_context().await;
         session
-            .record_context_updates_and_set_reference_context_item(&turn_context, None, false)
+            .record_context_updates_and_set_reference_context_item(&turn_context, None)
             .await;
         let history = session.clone_history().await;
         let initial_context = session.build_initial_context(&turn_context).await;
@@ -7957,7 +7950,7 @@ mod tests {
             .record_into_history(std::slice::from_ref(&compacted_summary), &turn_context)
             .await;
         session
-            .record_context_updates_and_set_reference_context_item(&turn_context, None, false)
+            .record_context_updates_and_set_reference_context_item(&turn_context, None)
             .await;
         session.clear_reference_context_item().await;
         session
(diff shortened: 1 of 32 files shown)


charley-oai:
@codex review this

codex-rs/core/src/codex.rs
```diff
@@ -4277,13 +4292,22 @@ pub(crate) async fn run_turn(
...
+        Ok(compacted) => compacted,
+        Err(_) => {
+            error!("Failed to run pre-sampling compact");
+            return None;
+        }
+    };
+
+    let previous_model = sess.previous_model().await;
+    sess.record_context_updates_and_set_reference_context_item(
+        turn_context.as_ref(),
```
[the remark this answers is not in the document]
charley-oai: Will do in followup PR

codex-rs/core/src/codex.rs
```diff
@@ -3898,8 +3915,6 @@ mod handlers {
         }
 
         let turn_context = sess.new_default_turn_with_sub_id(sub_id).await;
-        sess.set_previous_model(Some(turn_context.model_info.slug.clone()))
-            .await;
 
         let mut history = sess.clone_history().await;
         history.drop_last_n_user_turns(num_turns);
```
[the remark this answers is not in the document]
charley-oai: Will do in followup PR

charley-oai:
@codex review this

codex-rs/core/src/state/session.rs
```diff
@@ -81,12 +77,16 @@ impl SessionState {
         self.history.set_token_info(info);
     }
 
-    pub(crate) fn set_previous_context_item(&mut self, item: Option<TurnContextItem>) {
-        self.history.set_previous_context_item(item);
+    /// Storage-only setter for the reference context snapshot.
```
pakrym-oai: nit: Is this a useful comment?

codex-rs/core/src/tasks/compact.rs
```diff
@@ -25,20 +25,23 @@ impl SessionTask for CompactTask {
...
             let _ = session.services.otel_manager.counter(
                 "codex.task.compact",
                 1,
                 &[("type", "local")],
             );
-            let _ = crate::compact::run_compact_task(session, ctx, input).await;
+            crate::compact::run_compact_task(session.clone(), ctx, input).await
+        };
+        if compact_result.is_ok() {
+            session.clear_reference_context_item().await;
```
pakrym-oai: why isn't it cleared automatically on `replace_history` ?

codex-rs/core/src/tasks/mod.rs
```diff
@@ -121,8 +121,6 @@ impl Session {
     ) {
         self.abort_all_tasks(TurnAbortReason::Replaced).await;
         self.clear_connector_selection().await;
-        self.seed_initial_context_if_needed(turn_context.as_ref())
```
pakrym-oai, line 124: Yes! 🥳

codex-rs/core/src/codex.rs
```diff
@@ -1599,18 +1600,18 @@ impl Session {
...
                     Self::extract_mcp_tool_selection_from_rollout(&rollout_items);
-                let previous_model = Self::last_rollout_model_name(&rollout_items)
+                let previous_model = Self::last_rollout_regular_turn_model_name(&rollout_items)
                     .map(std::string::ToString::to_string);
                 {
                     let mut state = self.state.lock().await;
-                    state.initial_context_seeded = false;
+                    state.set_reference_context_item(None);
                 }
                 self.set_previous_model(previous_model).await;
```
pakrym-oai: why are some set_previous_model calls gone and others not?

codex-rs/core/src/codex.rs
```diff
@@ -1599,18 +1600,18 @@ impl Session {
...
 
                 // If resuming, warn when the last recorded model differs from the current one.
                 let curr = turn_context.model_info.slug.as_str();
-                if let Some(prev) =
-                    Self::last_rollout_model_name(&rollout_items).filter(|p| *p != curr)
+                if let Some(prev) = Self::last_rollout_regular_turn_model_name(&rollout_items)
```
pakrym-oai: at this point isn't it better to pull the last rollout context item?

codex-rs/core/tests/suite/snapshots/all__suite__compact__pre_turn_compaction_including_incoming_shapes.snap
```diff
@@ -12,14 +12,13 @@ Scenario: Pre-turn auto-compaction with a context override emits the context dif
...
+07:message/user:<SUMMARIZATION_PROMPT>
 
 ## Local Post-Compaction History Layout
-00:message/developer:<PERMISSIONS_INSTRUCTIONS>
-01:message/user:<AGENTS_MD>
-02:message/user:<ENVIRONMENT_CONTEXT:cwd=PRETURN_CONTEXT_DIFF_CWD>
-03:message/user:USER_ONE
-04:message/user:USER_TWO
-05:message/user:<COMPACTION_SUMMARY>\nPRE_TURN_SUMMARY
+00:message/user:USER_ONE
```
pakrym-oai, line 18: very nice

codex-rs/core/tests/suite/snapshots/all__suite__compact_remote__remote_manual_compact_with_history_shapes.snap
```diff
@@ -13,9 +12,9 @@ Scenario: Remote manual /compact where remote compact output is summary-only: fo
 04:message/assistant:FIRST_REMOTE_REPLY
 
 ## Remote Post-Compaction History Layout
-00:message/developer:<PERMISSIONS_INSTRUCTIONS>
-01:message/user:<AGENTS_MD>
-02:message/user:<ENVIRONMENT_CONTEXT:cwd=<CWD>>
-03:message/user:REMOTE_COMPACTED_SUMMARY
-04:compaction:encrypted=true
+00:message/user:REMOTE_COMPACTED_SUMMARY
```
pakrym-oai: Why both REMOTE_COMPACTED_SUMMARY and compaction:encrypted?

codex-rs/core/tests/suite/snapshots/all__suite__compact_remote__remote_mid_turn_compaction_multi_summary_reinjects_above_last_summary_shapes.snap
```diff
@@ -1,21 +1,18 @@
...
 
-## Remote Compaction Request
+## Second Turn Request (After Compaction)
 00:message/user:USER_ONE
-01:message/developer:<PERMISSIONS_INSTRUCTIONS>
-02:message/user:<AGENTS_MD>
-03:message/user:<ENVIRONMENT_CONTEXT:cwd=<CWD>>
-04:message/user:<COMPACTION_SUMMARY>\nREMOTE_OLDER_SUMMARY
+01:message/user:<COMPACTION_SUMMARY>\nREMOTE_OLDER_SUMMARY
+02:message/user:<COMPACTION_SUMMARY>\nREMOTE_LATEST_SUMMARY
```
pakrym-oai: why two COMPACTION_SUMMARies?

codex-rs/core/tests/suite/snapshots/all__suite__compact_remote__remote_mid_turn_compaction_shapes.snap
```diff
@@ -13,8 +13,8 @@ Scenario: Remote mid-turn continuation compaction after tool output: compact req
 05:function_call_output:unsupported call: test_tool
 
 ## Remote Post-Compaction History Layout
-00:message/user:USER_ONE
-01:message/developer:<PERMISSIONS_INSTRUCTIONS>
-02:message/user:<AGENTS_MD>
-03:message/user:<ENVIRONMENT_CONTEXT:cwd=<CWD>>
+00:message/developer:<PERMISSIONS_INSTRUCTIONS>
```
pakrym-oai, line 16: this doesn't look correct, why is the user turn all the way down?

codex-rs/core/tests/suite/compact.rs
```diff
@@ -2421,19 +2421,28 @@ async fn manual_compact_twice_preserves_latest_user_messages() {
...
-            .as_slice()
-            .starts_with(seeded_user_prefix),
-        "final request should start with seeded user prefix from first request: {seeded_user_prefix:?}"
-    );
-    let final_output = &final_request_user_texts[seeded_user_prefix.len()..];
-    let expected = vec![
+        !seeded_user_prefix.is_empty(),
+        "first turn should include seeded user prefix before the submitted user message"
+    );
+    let seeded_user_prefix_index = final_request_user_texts
```
pakrym-oai: this test was hard to understand but now it's even harder

codex-rs/core/src/compact.rs
```diff
@@ -272,24 +289,64 @@ pub(crate) fn is_summary_message(message: &str) -> bool {
...
+    mut compacted_history: Vec<ResponseItem>,
+    initial_context: Vec<ResponseItem>,
+) -> Vec<ResponseItem> {
+    let last_real_user_index = compacted_history
+        .iter()
+        .enumerate()
+        .rev()
+        .find_map(
+            |(i, item)| match crate::event_mapping::parse_turn_item(item) {
+                Some(TurnItem::UserMessage(user)) if !is_summary_message(&user.message()) => {
```
pakrym-oai: `!is_summary_message`, do we parse summary as a user message?

codex-rs/core/src/compact.rs
```diff
@@ -272,24 +289,64 @@ pub(crate) fn is_summary_message(message: &str) -> bool {
...
+                    Some(i)
+                }
+                _ => None,
+            },
+        );
+    let insertion_index = last_real_user_index.or_else(|| {
```
pakrym-oai: Why do we have two very similar loops?

codex-rs/core/src/compact.rs
```diff
@@ -272,24 +289,64 @@ pub(crate) fn is_summary_message(message: &str) -> bool {
...
+    sess: &Session,
+    turn_context: &TurnContext,
     mut compacted_history: Vec<ResponseItem>,
-    initial_context: &[ResponseItem],
+    callsite: CompactCallsite,
 ) -> Vec<ResponseItem> {
+    // Mid-turn compaction is the only callsite that must inject initial context above the last user message
+    // in the replacement history (preturn callsites simply inject initial context after the compaction item,
+    // but for mid-turn compaction the model has been trained to expect the compaction item last).
+    let initial_context = if matches!(callsite, CompactCallsite::MidTurn) {
```
pakrym-oai: Instead of the callsite can we make what we're doing explicit? `inject_full_context` ? or make turn_context parameter optional and only inject when it's provided?

pakrym-oai: Or just take the initial context as the parameter

codex-rs/core/src/compact.rs
```diff
@@ -1083,7 +1149,10 @@ keep me updated
             phase: None,
         }];
 
-        let refreshed = process_compacted_history(compacted_history, &initial_context);
+        let mut refreshed = compacted_history;
+        refreshed.retain(should_keep_compacted_history_item);
```
pakrym-oai: It's so strange that these tests are  copying calculations instead testing the real logic.

pakrym-oai: They should be calling process compacted history.

codex-rs/core/src/compact.rs
```diff
@@ -32,6 +32,14 @@ pub const SUMMARIZATION_PROMPT: &str = include_str!("../templates/compact/prompt
 pub const SUMMARY_PREFIX: &str = include_str!("../templates/compact/summary_prefix.md");
 const COMPACT_USER_MESSAGE_MAX_TOKENS: usize = 20_000;
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum CompactCallsite {
```
pakrym-oai: I'd like to get rid of this if possible. If it's used to decide behaviors we should replace with exact behavior being enabled. If it's used for logging, I'd start without and see if we really need it. I think we can tell by surrounding logs.

codex-rs/core/src/codex.rs
```diff
@@ -1697,13 +1698,82 @@ impl Session {
...
-            } else {
-                None
+    /// Returns the model from the latest completed user turn that persisted a
+    /// `RolloutItem::TurnContext` marker.
+    ///
+    /// This derives turn membership from persisted turn lifecycle events so
+    /// compact/shell-only turns (which have no user message event) do not
+    /// overwrite the previous-model baseline. `ThreadRolledBack` markers are
+    /// applied so resume/fork uses the post-rollback history view.
+    fn last_rollout_regular_turn_model_name(rollout_items: &[RolloutItem]) -> Option<&str> {
```
pakrym-oai: Why don't we pick up the last TurnContext without scanning for user turn? Do we need all the detection complexity?

codex-rs/core/src/codex.rs
```diff
@@ -2033,18 +2103,18 @@ impl Session {
     fn build_settings_update_items(
         &self,
         previous_context: Option<&TurnContextItem>,
-        resumed_model: Option<&str>,
+        previous_user_turn_model: Option<&str>,
```
pakrym-oai, line 2168: why do we need this parameter when we have TurnContextItem?

codex-rs/core/src/codex.rs
```diff
@@ -2678,14 +2724,70 @@ impl Session {
...
+    /// developer message so model-specific instructions are not lost.
+    ///
+    /// Invariant: this is the only runtime path that writes a non-`None`
+    /// `reference_context_item`. Non-regular tasks intentionally do not update that
+    /// baseline; `reference_context_item` tracks the latest regular model turn.
+    pub(crate) async fn record_context_updates_and_set_reference_context_item(
+        &self,
+        turn_context: &TurnContext,
+        previous_user_turn_model: Option<&str>,
+        force_full_context_injection: bool,
```
pakrym-oai: force_full_context_injection isn't this parameter === previous_user_turn_model: None? 
Can we remove the extra parameter?

codex-rs/core/src/codex.rs
```diff
@@ -2678,14 +2724,70 @@ impl Session {
...
+            {
+                initial_context.insert(0, model_switch_item);
+            }
+            initial_context
+        } else {
+            // Steady-state path: append only context diffs to minimize token overhead.
+            settings_update_items
+        };
+        if !context_items.is_empty() {
+            self.record_into_history(&context_items, turn_context).await;
```
pakrym-oai: record_conversation_items?

codex-rs/core/src/codex.rs
```diff
@@ -2678,14 +2724,70 @@ impl Session {
...
+    ) {
+        let reference_context_item = self.reference_context_item().await;
+        let settings_update_items = self.build_settings_update_items(
+            reference_context_item.as_ref(),
+            previous_user_turn_model,
+            turn_context,
+        );
+        let should_inject_full_context =
+            force_full_context_injection || reference_context_item.is_none();
+        let context_items = if should_inject_full_context {
```
pakrym-oai, line 2816: This smells like a hack. Should model switch message be added outside of `record_context_updates_and_set_reference_context_item` ? by the model switch routine itself?

codex-rs/core/src/codex.rs
charley-oai: The new invariant of set_previous_model is
```
   /// Model used by the latest regular user turn, used for model-switch handling
    /// on subsequent regular turns (including full-context reinjection after
    /// resume or `/compact`).
```
it was set in lots of places that were not regular user turns, violating this new definition.

The reason we need this definition is because a review turn or shell command after model switch could prevent the <model_switch> message from being injected now that we only emit context diffs in regular tasks (inside `run_turn`)

<model_switch> is like the only context item which can't be seeded from initial context so previous_model has to be tracked separately

codex-rs/core/src/codex.rs
charley-oai: previous_context (reference_context_item) is the diff baseline for environment/permissions/personality, and it can be intentionally None (for example after /compact to force full context reinjection)

This comment explains more why we still need previous_model https://github.com/openai/codex/pull/12252/changes/b66f6fb0ad4097ca70c82367842fb2cdb17e4cf1#r2834317238

I'll rename `previous_context` to `reference_context_item` for clarity there

codex-rs/core/src/codex.rs
```diff
@@ -4425,6 +4520,11 @@ pub(crate) async fn run_turn(
     let response_item: ResponseItem = initial_input_for_turn.clone().into();
     sess.record_user_prompt_and_emit_turn_item(turn_context.as_ref(), &input, response_item)
         .await;
+    // Track the previous-model baseline from the regular user-turn path only so
```
pakrym-oai, line 4584: We need to figure out a systematic solution here. For example, force model switch compact for other tasks (doesn't have to be a part of this PR)

pakrym-oai (changes requested):
I'd love to (if possible):
1. get rid of CompactCallsite
2. get rid of last_model and replace it with last context item

codex-rs/core/src/codex.rs
```diff
@@ -3898,10 +3915,12 @@ mod handlers {
...
         let turn_context = sess.new_default_turn_with_sub_id(sub_id).await;
-        sess.set_previous_model(Some(turn_context.model_info.slug.clone()))
-            .await;
 
         let mut history = sess.clone_history().await;
+        // TODO(ccunningham): Fix rollback/backtracking baseline handling.
+        // Truncating history should also invalidate/recompute
+        // `reference_context_item` and `previous_model` so the next regular
+        // turn replays any dropped context diffs / model-switch instructions.
         history.drop_last_n_user_turns(num_turns);
```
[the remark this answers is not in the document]
charley-oai: Will do in followup PR

codex-rs/core/src/compact.rs
charley-oai: This was for the "inject before last user message, or before last summary if there are no user messages in compacted_history" but I think we can dd because we will  never have 0 user messages but multiple summaries in compacted_history

codex-rs/core/src/compact.rs
charley-oai: Yes for local compaction

pakrym-oai: We maybe should fix the `parse_turn_item` to skip those.

codex-rs/core/tests/suite/snapshots/all__suite__compact_remote__remote_mid_turn_compaction_shapes.snap
charley-oai: Because these turn contexts apply to the user message; mid turn injects turn context above the last user message

codex-rs/core/tests/suite/snapshots/all__suite__compact_remote__remote_manual_compact_with_history_shapes.snap
charley-oai: Will fix in this PR

codex-rs/core/src/compact.rs
charley-oai: Will try in a followup PR

codex-rs/core/src/codex.rs
charley-oai: I guess because /compact writes a TurnContext too, and our definition of `reference_turn_context` is that it's the `TurnContextItem` from the last regular user turn (so not /compact)

> Runtime TurnContext writes I found:
> 
>   - Regular sampling turn path: `codex-rs/core/src/codex.rs:5656`
>       - inside try_run_sampling_request(...)
>       - this is called from the regular turn sampling loop (run_sampling_request / run_turn) (`codex-rs/core/src/codex.rs:5076`)
>   - Local compaction task (/compact): `codex-rs/core/src/compact.rs:143`
>       - it persists RolloutItem::TurnContext(...) before issuing the compaction sampling request

charley-oai: Discussed, we are going to make /compact stop writing TurnContextItem

codex-rs/core/src/codex.rs
```diff
@@ -4425,6 +4543,11 @@ pub(crate) async fn run_turn(
     let response_item: ResponseItem = initial_input_for_turn.clone().into();
     sess.record_user_prompt_and_emit_turn_item(turn_context.as_ref(), &input, response_item)
         .await;
+    // Track the previous-model baseline from the regular user-turn path only so
+    // standalone tasks (compact/shell/review/undo) cannot suppress future
+    // `<model_switch>` injections.
+    sess.set_previous_model(Some(turn_context.model_info.slug.clone()))
```
charley-oai, line 4587: Why are we doing set_previous_model here?

codex-rs/core/src/codex.rs
```diff
@@ -1697,14 +1707,112 @@ impl Session {
...
+                    } else {
+                        saw_surviving_compaction_after_candidate = true;
+                    }
+                }
+                _ => {}
             }
-        })
+        }
+
+        (None, false)
```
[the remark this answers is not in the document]
charley-oai: @pakrym is this kind of thing (legacy rollouts) fine to ignore for the sake of code simplicity?

codex-rs/core/src/codex.rs
```diff
@@ -3927,10 +4029,12 @@ mod handlers {
...
         let turn_context = sess.new_default_turn_with_sub_id(sub_id).await;
-        sess.set_previous_model(Some(turn_context.model_info.slug.clone()))
-            .await;
 
         let mut history = sess.clone_history().await;
+        // TODO(ccunningham): Fix rollback/backtracking baseline handling.
+        // Truncating history should also invalidate/recompute
+        // `reference_context_item` and `previous_model` so the next regular
+        // turn replays any dropped context diffs / model-switch instructions.
         history.drop_last_n_user_turns(num_turns);
```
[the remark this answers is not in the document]
charley-oai: Followup PR

codex-rs/core/src/codex.rs
charley-oai: > Why this exact spot in run_turn:
> 
>   - It only runs on the regular user-turn path (not /compact, shell, review, undo), so standalone tasks cannot corrupt that baseline.
>   - It runs after record_user_prompt_and_emit_turn_item(...), so the turn has actually become a real user turn in history before we advance the
>     baseline.
>   - It updates the baseline for the next turn even if this turn later aborts/fails mid-sampling (which is usually what we want for model-switch
>     + compaction heuristics).
> 
>   So the short version: it’s there to preserve the invariant “previous_model tracks the last regular user turn only,” and to do it early enough
>   that failures don’t lose that baseline.

codex-rs/core/src/codex.rs
```diff
@@ -2678,14 +2750,66 @@ impl Session {
...
+        } else {
+            // Steady-state path: append only context diffs to minimize token overhead.
+            self.build_settings_update_items(
+                reference_context_item.as_ref(),
+                previous_user_turn_model,
+                turn_context,
+            )
+        };
+        if !context_items.is_empty() {
+            self.record_conversation_items(turn_context, &context_items)
```
pakrym-oai, line 2842: do we want to record turn context here as well?

codex-rs/core/src/compact.rs
```diff
@@ -32,10 +32,37 @@ pub const SUMMARIZATION_PROMPT: &str = include_str!("../templates/compact/prompt
...
+///
+/// Pre-turn/manual compaction variants use `DoNotInject`: they replace history with a summary and
+/// clear `reference_context_item`, so the next regular turn will fully reinject initial context
+/// after compaction.
+///
+/// Mid-turn compaction must use `BeforeLastUserMessage` because the model is trained to see the
+/// compaction summary as the last item in history after mid-turn compaction; we therefore inject
+/// initial context into the replacement history just above the last real user message.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum InitialContextInjection {
```
pakrym-oai, line 45: nit: bool inject_context would've done it

codex-rs/core/src/codex.rs
```diff
@@ -4630,12 +4760,18 @@ pub(crate) async fn run_turn(
     last_agent_message
 }
 
+/// Runs pre-sampling compaction checks and returns whether compaction actually ran.
+///
+/// Returns:
+/// - `Ok(true)` when at least one pre-sampling compact path ran successfully.
+/// - `Ok(false)` when no pre-sampling compaction was needed.
+/// - `Err(_)` only when a pre-sampling compact path was attempted and failed.
 async fn run_pre_sampling_compact(
```
pakrym-oai, line 4801: we don't use the return value, do we? can we revert?

codex-rs/core/tests/suite/snapshots/all__suite__compact_remote__remote_mid_turn_compaction_multi_summary_reinjects_above_last_summary_shapes.snap
```diff
@@ -1,21 +1,18 @@
...
+01:compaction:encrypted=true
 
-## Remote Compaction Request
+## Second Turn Request (After Compaction)
 00:message/user:USER_ONE
-01:message/developer:<PERMISSIONS_INSTRUCTIONS>
-02:message/user:<AGENTS_MD>
-03:message/user:<ENVIRONMENT_CONTEXT:cwd=<CWD>>
-04:message/user:<COMPACTION_SUMMARY>\nREMOTE_OLDER_SUMMARY
+01:compaction:encrypted=true
```
pakrym-oai, line 14: We might want to fix the compaction helper to drop compaction items.

codex-rs/core/src/compact.rs
```diff
@@ -272,24 +320,63 @@ pub(crate) fn is_summary_message(message: &str) -> bool {
     message.starts_with(format!("{SUMMARY_PREFIX}\n").as_str())
 }
 
-pub(crate) fn process_compacted_history(
+pub(crate) async fn process_compacted_history(
```
pakrym-oai: nit: this should move to remote, it's not used for local 

not blocking

pakrym-oai (approved):
last_rollout_regular_turn_context_lookup  and reconstruct_history_from_rollout need unification but can be done separately.

codex-rs/core/src/codex.rs
charley-oai: Discussed: larger impact (per turn `TurnContextItem` rather than per sampling request), will leave out of this PR

codex-rs/core/src/compact.rs
charley-oai: I kind of feel like variants are more readable at the callsite, and also this makes clear where the initial context is being injected so maybe a bit more readable and friendly to future vibe coders?

codex-rs/core/tests/suite/snapshots/all__suite__compact_remote__remote_mid_turn_compaction_multi_summary_reinjects_above_last_summary_shapes.snap
charley-oai: Added a TODO

charley-oai:
@codex review this

codex-rs/core/src/codex.rs
```diff
@@ -4648,55 +4783,75 @@ async fn run_pre_sampling_compact(
         .unwrap_or(i64::MAX);
     // Compact if the total usage tokens are greater than the auto compact limit
     if total_usage_tokens >= auto_compact_limit {
-        run_auto_compact(sess, turn_context).await?;
+        run_auto_compact(sess, turn_context, InitialContextInjection::DoNotInject).await?;
```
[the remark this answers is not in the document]
charley-oai: We have a TODO for this bruh

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

PR #12267, opened 2026-02-19
Add MCP server context to otel tool_result logs

colby-oai:
Summary
- capture the origin for each configured MCP server and expose it via the connection manager
- plumb MCP server name/origin into tool logging and emit codex.tool_result events with those fields
- add unit coverage for origin parsing and extend OTEL tests to assert empty MCP fields for non-MCP tools
- currently not logging full urls or url paths to prevent logging potentially sensitive data

Testing
- Not run (not requested)


codex-rs/core/src/tools/registry.rs
```diff
@@ -294,6 +305,20 @@ fn sandbox_policy_tag(policy: &SandboxPolicy) -> &'static str {
     }
 }
 
+async fn resolve_mcp_tool_result_fields(
```
apanasenko-oai: can you inline this if don't use it more than once?

apanasenko-oai (approved):

apanasenko-oai (approved):

PR #12279, opened 2026-02-19
Fix skills lookup across symlinked worktrees

etraut-openai:
Summary
- make `skills_for_cwd` canonicalize paths so entries matching real directory targets are found even when the caller is in a symlinked directory
- add a regression test that exercises symlink coverage and avoids the earlier lint issue by handling `dunce::canonicalize` failures cleanly

This is a fix for https://github.com/openai/codex/pull/12279.

### DIFF
--- a/codex-rs/tui/src/chatwidget/skills.rs
+++ b/codex-rs/tui/src/chatwidget/skills.rs
@@ -144,9 +144,22 @@ impl ChatWidget {
 }
 
 fn skills_for_cwd(cwd: &Path, skills_entries: &[SkillsListEntry]) -> Vec<ProtocolSkillMetadata> {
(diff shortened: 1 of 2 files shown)


codex-rs/tui/src/chatwidget/skills.rs
```diff
@@ -144,9 +144,22 @@ impl ChatWidget {
 }
 
 fn skills_for_cwd(cwd: &Path, skills_entries: &[SkillsListEntry]) -> Vec<ProtocolSkillMetadata> {
+    let cwd_canonical = dunce::canonicalize(cwd).ok();
```
xl-openai, line 147: Why this could be an issue? Core and TUI should just use the same cwd, regardless if it is a symlinked one or not?

PR #12289, opened 2026-02-20
CODEX-4927: Surface local login entitlement denials in browser

daniel-oai:
## Problem
Users without Codex access can hit a confusing local login loop. In the denial case, the callback could fall through to generic behavior (including a plain "Missing authorization code" page) instead of clearly explaining that access was denied.

Screenshot 2026-02-19 at 11 43 45 PM
Screenshot 2026-02-19 at 11 44 53 PM

## Scope
This PR improves local login error clarity only. It does not change entitlement policy, RBAC rules, or who is allowed to use Codex.

## What Changed
- The local OAuth callback handler now parses `error` and `error_description` on `/auth/callback` and exits the callback loop with a real failure.
- Callback failures render a branded local Codex error page instead of a generic/plain page.
- `access_denied` + `missing_codex_entitlement` is now mapped to an explicit user-facing message telling the user Codex is not enabled for their workspace and to contact their workspace administrator for access.
- Unknown OAuth callback errors continue to use a generic error page while preserving the OAuth error code/details for debugging.
- Added the login error page template to Bazel assets so the local binary can render it in Bazel builds.

## Non-goals
- No TUI onboarding/toast changes in this PR.
- No backend entitlement or policy changes.

## Tests
- Added an end-to-end `codex-login` test for `access_denied` + `missing_codex_entitlement` and verified the page shows the actionable admin guidance.
- Added an end-to-end `codex-login` test for a generic `access_denied` reason to verify we keep a generic fallback page/message.


codex-rs/login/src/server.rs
```diff
@@ -670,23 +715,109 @@ pub(crate) fn ensure_workspace_allowed(
...
+/// Converts OAuth callback errors into a user-facing message.
+fn oauth_callback_error_message(error_code: &str, error_description: Option<&str>) -> String {
+    if error_code == "access_denied"
+        && error_description.is_some_and(|description| {
+            description
+                .to_ascii_lowercase()
+                .contains("missing_codex_entitlement")
+        })
+    {
+        return "Sign-in was denied because this account is missing Codex entitlement.".to_string();
```
xl-openai: No so actionable? maybe "Codex is not enabled for your workspace, please contact your workspace admin?"

codex-rs/login/src/server.rs
```diff
@@ -670,23 +715,109 @@ pub(crate) fn ensure_workspace_allowed(
...
+
+/// Renders the branded error page used by callback failures.
+fn render_login_error_page(
+    message: &str,
+    error_code: Option<&str>,
+    error_description: Option<&str>,
+) -> Vec<u8> {
+    let template = include_str!("assets/error.html");
+    let code = error_code.unwrap_or("unknown_error");
+    let description = error_description.unwrap_or(message);
```
xl-openai: We might want to map this to something more meaningful/actionable for user.

xl-openai:
I think a failed page is good enough for all produce surfaces. But thanks for adding this!!

codex-rs/login/src/server.rs
daniel-oai: Yep, agreed. I added a known-error mapping for the entitlement denial (admin-actionable copy) and kept a generic fallback for unknown callback errors. There is now a dedicated e2e test for both the entitlement-specific page and the generic fallback path.

codex-rs/login/src/server.rs
daniel-oai: Updated this path to special-case `access_denied` + `missing_codex_entitlement` with explicit copy: Codex is not enabled for the workspace and the user should contact their workspace administrator for access. I also removed the TUI error surfacing from this PR so the browser page is the primary UX.

xl-openai (approved):

PR #12301, opened 2026-02-20
Add field to Thread object for the latest rename set for a given thread

natea-oai:
Exposes through the app server updated names set for a thread. This enables other surfaces to use the core as the source of truth for thread naming. `threadName` is gathered using the helper functions used to interact with `session_index.jsonl`, and is hydrated in:
- `thread/list`
- `thread/read`
- `thread/resume`
- `thread/unarchive`
- `thread/rollback`

We don't do this for `thread/start` and `thread/fork`.


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2545,18 +2548,36 @@ impl CodexMessageProcessor {
...
+        for summary in summaries {
+            let conversation_id = summary.conversation_id;
+            thread_ids.insert(conversation_id);
+
+            let thread = summary_to_thread(summary);
+            status_ids.push(thread.id.clone());
+            threads.push((conversation_id, thread));
+        }
+
+        let names = match find_thread_names_by_ids(&self.config.codex_home, &thread_ids).await {
```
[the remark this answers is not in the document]
natea-oai: known constraint

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2212,6 +2212,8 @@ pub struct Thread {
     pub agent_role: Option<String>,
     /// Optional Git metadata captured when the thread was created.
     pub git_info: Option<GitInfo>,
+    /// Optional user-facing thread title.
+    pub thread_name: Option<String>,
```
pakrym-oai: this property is already on the thread. `name`?

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -1187,6 +1191,16 @@ pub(crate) async fn apply_bespoke_event_handling(
                                 thread.status = thread_watch_manager
                                     .loaded_status_for_thread(&thread.id)
                                     .await;
+                                match find_thread_name_by_id(codex_home, &conversation_id).await {
```
pakrym-oai, line 1194: is this method call slow. Are we getting N+1 problem?

codex-rs/app-server-protocol/src/protocol/v2.rs
natea-oai: agree

codex-rs/app-server/src/bespoke_event_handling.rs
natea-oai: it's slow in being linear in number of total renames, where a thread without a rename results in scanning the full file. but this also is highly dependent on the amount of renames for the user - one remedy prior to sqlite could be to limit the number of searches to some reasonable upper limit.

natea-oai: for batched calls like thread/list we have a batched helper for gathering names. for the single thread calls, I'm unsure if there is a more efficient way to do this

pakrym-oai (changes requested):
Need tests for changed endpoints.

natea-oai:
tested this a bit manually by sticking 50k rows in my session_index.jsonl file and performance in app for loading seemed little changed

natea-oai:
tested this a bit manually by sticking 50k rows in my session_index.jsonl file and performance in app for loading seemed little changed

codex-rs/app-server/tests/suite/v2/thread_fork.rs
```diff
@@ -70,8 +71,19 @@ async fn thread_fork_creates_new_thread_and_emits_started() -> Result<()> {
...
+    let fork_result = fork_resp.result.clone();
     let ThreadForkResponse { thread, .. } = to_response::<ThreadForkResponse>(fork_resp)?;
 
+    // Wire contract: thread title field is `name`.
+    let thread_json = fork_result
+        .get("thread")
+        .and_then(Value::as_object)
+        .expect("thread/fork result.thread must be an object");
+    assert!(
+        thread_json.contains_key("name"),
```
pakrym-oai: assert the actual value?

codex-rs/app-server/tests/suite/v2/thread_fork.rs
```diff
@@ -115,6 +127,15 @@ async fn thread_fork_creates_new_thread_and_emits_started() -> Result<()> {
         mcp.read_stream_until_notification_message("thread/started"),
     )
     .await??;
+    let started_params = notif.params.clone().expect("params must be present");
+    let started_thread_json = started_params
+        .get("thread")
+        .and_then(Value::as_object)
+        .expect("thread/started params.thread must be an object");
+    assert!(
```
pakrym-oai: assert the actual value?

pakrym-oai (approved):
In places where we assert that the key exist, we should assert the value (even if it's null)

PR #12312, opened 2026-02-20
feat: add config `allow_login_shell`

jif-oai:


codex-rs/core/src/tools/spec.rs
```diff
@@ -234,14 +241,6 @@ fn create_exec_command_tool() -> ToolSpec {
...
             },
         ),
-        (
-            "login".to_string(),
-            JsonSchema::Boolean {
-                description: Some(
-                    "Whether to run the shell with -l/-i semantics. Defaults to true.".to_string(),
-                ),
-            },
-        ),
```
joshka-oai, line 244: What's the impact of dropping this part?

codex-rs/core/src/config/mod.rs
```diff
@@ -139,6 +139,8 @@ pub struct Permissions {
     pub sandbox_policy: Constrained<SandboxPolicy>,
     /// Effective network configuration applied to all spawned processes.
     pub network: Option<NetworkProxySpec>,
+    /// Whether shell-based tools may request login shell semantics.
+    pub allow_login_shell: bool,
```
joshka-oai: It may be worth going a little deeper in these doc (summary then body) and describing the impact of this setting on how codex works (i.e. it adds -l to bash commands, loads different shell startup scripts, may mean that your path is changed. This will flow into developers.openai.com docs so it's useful to have enough docs to help prime that with things users would search for when solving problems, and not just the what does this turn on / off here). Same comment for the below.

It also flows to the config schema - put yourself in the mindset of someone asking themselves why this setting is false when reading a config.toml. If they can hover over this and get that context, that's a strong win.

Codex prompt something like "add user facing docs to the config settings that help users understand why they'd change the setting, what problems it helps, how it affects the various places that we run code on behalf of the user (unified / shell etc.)

joshka-oai (approved):
I wasn't sure about whether we currently tell the model before hand that login shells are disabled by this. Should we?

codex-rs/core/src/tools/spec.rs
jif-oai: Model will not be able to use "login" anymore

PR #12313, opened 2026-02-20
Handle orphan exec ends without clobbering active exploring cell

jif-oai:
Summary
- distinguish exec end handling targets (active tracking, active orphan history, new cell) so unified exec responses don’t clobber unrelated exploring cells
- ensure orphan ends flush existing exploring history when complete, insert standalone history entries, and keep active cells correct
- add regression tests plus a snapshot covering the new behavior and expose the ExecCell completion result for verification

Fix for https://github.com/openai/codex/issues/12278

### DIFF
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -2315,6 +2315,12 @@ impl ChatWidget {
     }
 
     pub(crate) fn handle_exec_end_now(&mut self, ev: ExecCommandEndEvent) {
+        enum ExecEndTarget {
+            ActiveTracked,
+            OrphanHistoryWhileActiveExec,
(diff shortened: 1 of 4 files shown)


joshka-oai:
Ran a documentation-pass skill with Codex to generate docs/comments to help understand this a bit better.

Pushed a docs-only follow-up commit on top of this PR that:
- documents the exec-end routing contract in chatwidget (tracked call vs orphan end vs new cell)
- documents that ExecCell::complete_call returning false is a routing-mismatch signal, not a benign no-op

No runtime behavior changes in this follow-up; it is only rustdoc/comments. I also ran cargo test -p codex-tui exec_end_without_begin after the doc pass.

joshka-oai (approved):

PR #12314, opened 2026-02-20
fix(core) Filter non-matching prefix rules

dylan-hurd-oai:
## Summary
`gpt-5.3-codex` really likes to write complicated shell scripts, and suggest a partial prefix_rule that wouldn't actually approve the command. We should only show the `prefix_rule` suggestion from the model if it would actually fully approve the command the user is seeing.

This will technically cause more instances of overly-specific suggestions when we fallback, but I think the UX is clearer, particularly when the model doesn't necessarily understand the current limitations of execpolicy parsing.

## Testing
 - [x] Add unit tests
 - [x] Add integration tests


jif-oai:
@codex review

pakrym-oai (approved):

PR #12319, opened 2026-02-20
feat: add service name to app-server

jif-oai:
Add service name to the app-server so that the app can use it's own service name

This is on thread level because later we might plan the app-server to become a singleton on the computer


codex-rs/core/tests/suite/mod.rs
```diff
@@ -84,7 +84,6 @@ mod live_reload;
 mod model_info_overrides;
 mod model_overrides;
 mod model_switching;
-mod model_tools;
```
jif-oai, line 87: flaky and pointless

owenlin0 (approved):

Issue #36, opened 2025-04-16
NODE_OPTIONS=--no-deprecation: not found


okey00:
Running via WSL2 on Windows 11 and receiving the following error... 

codex: 15: exec: NODE_OPTIONS=--no-deprecation: not found

mfrancis107:
Mac also

LucasArmand:
1. cd to /mnt/c/Users/yourname/AppData/Roaming/npm
2. open the codex file with editor of your choice (i.e. vim, nano)
3. Find this line (around line 15):
`exec NODE_OPTIONS=--no-deprecation node ...`
4. Replace it with two lines instead:
```
export NODE_OPTIONS=--no-deprecation
exec node ...
```
... is the rest of the code after the --no-deprecation flag in the original line

jonchurch:
This ultimately is coming from the shebang here https://github.com/openai/codex/blob/e9f84eab012bf5d59a02bc79e4d6fb1d44572718/codex-cli/src/cli.tsx#L1

looks like a portability issue? I dont see the error on mac os 15.3.2

okey00:
Thanks Lucas for the quick response. That worked

jonchurch:
@okey00 That's a workaround, not a fix. You should keep this issue open until root cause is determined and a fix merged

dahifi:
I did a patch in my bin folder so that I wouldn't have to edit the npm package. I've put in a PR that might fix this, hopefully. https://github.com/openai/codex/issues/42

okey00:
Reopened as suggested by Jon as it's a workaround in it's current state.

tibo-openai:
This was fixed by https://github.com/openai/codex/pull/96

PR #12334, opened 2026-02-20
feat(tui): add /title terminal title configuration

yvolovich-cyber:
## Problem

When multiple Codex sessions are open at once, terminal tabs and windows are hard to distinguish from each other. The existing status line only helps once the TUI is already focused, so it does not solve the "which tab is this?" problem.

This PR adds a first-class `/title` command so the terminal window or tab title can carry a short, configurable summary of the current session.

## Screenshot

image

## Mental model

`/statusline` and `/title` are separate status surfaces with different constraints. The status line is an in-app footer that can be denser and more detailed. The terminal title is external terminal metadata, so it needs short, stable segments that still make multiple sessions easy to tell apart.

The `/title` configuration is an ordered list of compact items. By default it renders `spinner,project`, so active sessions show lightweight progress first while idle sessions still stay easy to disambiguate. Each configured item is omitted when its value is not currently available rather than forcing a placeholder.

## Non-goals

This does not merge `/title` into `/statusline`, and it does not add an arbitrary free-form title string. The feature is intentionally limited to a small set of structured items so the title stays short and reviewable.

This also does not attempt to restore whatever title the terminal or shell had before Codex started. When Codex clears the title, it clears the title Codex last wrote.

## Tradeoffs

A separate `/title` command adds some conceptual overlap with `/statusline`, but it keeps title-specific constraints explicit instead of forcing the status line model to cover two different surfaces.

Title refresh can happen frequently, so the implementation now shares parsing and git-branch orchestration between the status line and title paths, and caches the derived project-root name by cwd. That keeps the hot path cheap without introducing background polling.

## Architecture

The TUI gets a new `/title` slash command and a dedicated picker UI for selecting and ordering terminal-title items. The chosen ids are persisted in `tui.terminal_title`, with `spinner` and `project` as the default when the config is unset. `status` remains available as a separate text item, so configurations like `spinner,status` render compact progress like `⠋ Working`.

`ChatWidget` now refreshes both status surfaces through a shared `refresh_status_surfaces()` path. That shared path parses configured items once, warns on invalid ids once, synchronizes shared cached state such as git-branch lookup, then renders the footer status line and terminal title from the same snapshot.

Low-level OSC title writes live in `codex-rs/tui/src/terminal_title.rs`, which owns the terminal write path and last-mile sanitization before emitting OSC 0.

## Security

Terminal-title text is treated as untrusted display content before Codex emits it. The write path strips control characters, removes invisible and bidi formatting characters that can make the title visually misleading, normalizes whitespace, and caps the emitted length.

References used while implementing this:

- [xterm control sequences](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html)
- [WezTerm escape sequences](https://wezterm.org/escape-sequences.html)
- [CWE-150: Improper Neutralization of Escape, Meta, or Control Sequences](https://cwe.mitre.org/data/definitions/150.html)
- [CERT VU#999008 (Trojan Source)](https://kb.cert.org/vuls/id/999008)
- [Trojan Source disclosure site](https://trojansource.codes/)
- [Unicode Bidirectional Algorithm (UAX #9)](https://www.unicode.org/reports/tr9/)
- [Unicode Security Considerations (UTR #36)](https://www.unicode.org/reports/tr36/)

## Observability

Unknown configured title item ids are warned about once instead of repeatedly spamming the transcript. Live preview applies immediately while the `/title` picker is open, and cancel rolls the in-memory title selection back to the pre-picker value.

If terminal title writes fail, the TUI emits debug logs around set and clear attempts. The rendered status label intentionally collapses richer internal states into compact title text such as `Starting...`, `Ready`, `Thinking...`, `Working...`, `Waiting...`, and `Undoing...` when `status` is configured.

## Tests

Ran:

- `just fmt`
- `cargo test -p codex-tui`

At the moment, the red Windows `rust-ci` failures are due to existing `codex-core` `apply_patch_cli` stack-overflow tests that also reproduce on `main`. The `/title`-specific `codex-tui` suite is green.


codex-rs/tui/src/chatwidget.rs
```diff
@@ -964,6 +991,114 @@ impl ChatWidget {
...
+        let had_previous_title = self.last_terminal_title.is_some();
+        self.last_terminal_title = title.clone();
+
+        if let Some(title) = title {
+            if let Err(err) = set_terminal_title(&title) {
+                tracing::debug!(error = %err, "failed to set terminal title");
+            }
+        } else if had_previous_title && let Err(err) = clear_terminal_title() {
+            tracing::debug!(error = %err, "failed to clear terminal title");
+        }
```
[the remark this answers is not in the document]
yvolovich-cyber: Agreed. I updated the cache semantics so `last_terminal_title` only changes after successful `set_terminal_title()` / `clear_terminal_title()`. On write failure we now preserve the prior cached value, which avoids suppressing retries. Fixed in 4082887f4.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -964,6 +991,114 @@ impl ChatWidget {
...
+                proper_join(invalid_items.as_slice())
+            );
+            self.on_warning(message);
+        }
+
+        if items.is_empty() {
+            if self.last_terminal_title.take().is_some()
+                && let Err(err) = clear_terminal_title()
+            {
+                tracing::debug!(error = %err, "failed to clear terminal title");
```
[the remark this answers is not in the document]
yvolovich-cyber: Good catch. I agree. I changed `refresh_terminal_title()` so `last_terminal_title` is only updated after a successful terminal clear/write. If `clear_terminal_title()` fails, we keep the previous cached value so a later refresh will retry. Fixed in 4082887f4.

codex-rs/tui/src/terminal_title.rs
```diff
@@ -0,0 +1,150 @@
...
+const MAX_TERMINAL_TITLE_CHARS: usize = 240;
+
+pub(crate) fn set_terminal_title(title: &str) -> io::Result<()> {
+    if !stdout().is_terminal() {
+        return Ok(());
+    }
+
+    let title = sanitize_terminal_title(title);
+    if title.is_empty() {
+        return Ok(());
```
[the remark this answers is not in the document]
yvolovich-cyber: Good catch. I agree. `set_terminal_title()` previously returned success when sanitization stripped everything, which could leave a stale title cached by the caller. I changed it to clear the terminal title instead (`clear_terminal_title()`) in that case. Fixed in bf56a46c3.

yvolovich-cyber (commented):
> ### 💡 Codex Review
> https://github.com/openai/codex/blob/4082887f47a770270b80cf96bb0286f239ad0f95/codex-rs/tui/src/chatwidget.rs#L1135-L1137
> 
> **P2 Badge Refresh git-branch title when only /title enables it**
> Update `request_status_line_branch_refresh` to consider terminal-title config too. It currently returns unless the **status line** includes `git-branch`, but terminal titles now also use the same cached branch value. If a user enables `git-branch` in `/title` only, branch lookup is never re-triggered after turn completion, so the title can show a stale branch after `git checkout`/branch-changing commands.

Addressed it in `2cf7abd05`.

What changed:
- `request_status_line_branch_refresh()` now refreshes when `git-branch` is configured in either `/statusline` or `/title`
- added a regression test for the title-only `git-branch` case after `TurnComplete`

yvolovich-cyber:
@codex review

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -6261,9 +6261,22 @@ async fn handle_pending_thread_resume_request(
...
     };
-    thread.status = thread_watch_manager
+    let mut status = thread_watch_manager
         .loaded_status_for_thread(&thread.id)
         .await;
+    let has_in_progress_turn = thread
+        .turns
+        .iter()
+        .any(|turn| matches!(turn.status, TurnStatus::InProgress));
+    if matches!(status, ThreadStatus::Idle) && (active_turn.is_some() || has_in_progress_turn) {
```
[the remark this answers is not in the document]
yvolovich-cyber: Agreed. This now checks only the latest stored turn (`latest_turn_is_in_progress(&thread.turns)`) instead of `any()` historical `InProgress`, which avoids marking resumed idle threads active because of stale history. I also added a unit test for the stale historical in-progress case. Fixed in `9f62b3929`.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -4516,6 +4757,81 @@ impl ChatWidget {
...
+        let mut chars = value.chars();
+        let head: String = chars.by_ref().take(max_chars).collect();
+        if chars.next().is_none() || max_chars <= 3 {
+            return head;
+        }
+
+        let mut truncated = head.chars().take(max_chars - 3).collect::<String>();
+        truncated.push_str("...");
+        truncated
+    }
```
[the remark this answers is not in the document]
yvolovich-cyber: Good point. I changed `truncate_terminal_title_part()` to truncate on grapheme clusters using `UnicodeSegmentation`, so emoji and combining sequences are not split. I also added `terminal_title_part_truncation_preserves_grapheme_clusters` regression coverage. Fixed in `9f62b3929`.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -1008,6 +1036,131 @@ impl ChatWidget {
         self.refresh_status_line();
     }
 
+    /// Recomputes and emits the terminal title from config and runtime state.
+    pub(crate) fn refresh_terminal_title(&mut self) {
```
jif-oai: Lots of duplicates with `refresh_status_line`
We should be able to merge/reuse stuff

codex-rs/tui/src/chatwidget/tests.rs
```diff
@@ -1652,6 +1653,7 @@ async fn make_chatwidget_manual(
         had_work_activity: false,
         saw_plan_update_this_turn: false,
         saw_plan_item_this_turn: false,
+        last_plan_progress: None,
```
jif-oai, line 1924: A `on_task_started` should also reset this

codex-rs/tui/src/chatwidget.rs
```diff
@@ -4680,6 +4922,81 @@ impl ChatWidget {
...
+    fn terminal_title_status_text(&self) -> String {
+        if self.mcp_startup_status.is_some() {
+            return "Starting...".to_string();
+        }
+
+        if !self.bottom_pane.is_task_running() {
+            return "Ready".to_string();
+        }
+
+        if self.current_status_header == "Working" {
```
jif-oai: this kind of thing is brittle.... we should have something stronger matching on the EventMsg

codex-rs/tui/src/chatwidget.rs
jif-oai: this function calls some file walker at each call to retrieve the git status for example. And it is called VERY frequently. We should cache most of those things instead of recomputing each time

jif-oai (commented):
I don't see anything that reset the title after a `CMD + C`. Can you make sure the title always get reset when we exit Codex?

codex-rs/tui/src/chatwidget/tests.rs
yvolovich-cyber: Agreed. `on_task_started()` now resets `last_plan_progress`, so stale `task-progress` values do not leak into the next terminal title. I also added `on_task_started_resets_terminal_title_task_progress` coverage. Fixed in `be6d8f3bc`.

codex-rs/tui/src/chatwidget.rs
yvolovich-cyber: Agreed. I removed the string matching on `current_status_header` for terminal-title status and replaced it with an explicit `TerminalTitleStatusKind` state updated by the relevant event handlers (`TerminalInteraction`, `UndoStarted`, reasoning/background updates, and task start). I also updated tests to drive this via events. Fixed in `be6d8f3bc`.

codex-rs/tui/src/chatwidget.rs
yvolovich-cyber: Agreed. I refactored the status and terminal-title refresh path to share a single `StatusSurfaceSelections` parse plus shared invalid-item warnings and git-branch state orchestration. `refresh_status_line()` now computes selections once and reuses them for both renders, which removes the duplicate logic in this path. Fixed in `be6d8f3bc`.

yvolovich-cyber: Good call. `refresh_terminal_title()` is called very frequently, and the expensive repeated filesystem walk here was the project-root/git-root resolution used by the project label. I added a cwd-keyed `status_line_project_root_name_cache` to avoid repeated parent-directory walks, and the shared refresh refactor also avoids duplicate parsing and git-branch orchestration in the `refresh_status_line()` -> `refresh_terminal_title()` path. Branch lookup remains async and cached by cwd. Fixed in `be6d8f3bc`.

codex-rs/tui/src/terminal_title.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    execute!(stdout(), SetWindowTitle(title))
+}
+
+pub(crate) fn clear_terminal_title() -> io::Result<()> {
+    if !stdout().is_terminal() {
+        return Ok(());
+    }
+
+    execute!(stdout(), SetWindowTitle(String::new()))
```
jif-oai, line 73: this is not the same thing as restoring the previous one

codex-rs/tui/src/chatwidget.rs
```diff
@@ -914,6 +972,7 @@ impl ChatWidget {
     fn set_status(&mut self, header: String, details: Option<String>) {
         self.current_status_header = header.clone();
         self.bottom_pane.update_status(header, details);
+        self.refresh_terminal_title();
```
jif-oai: I think in the end this should become async. Can be done as follow-up

codex-rs/tui/src/bottom_pane/title_setup.rs
```diff
@@ -0,0 +1,229 @@
...
+                    .filter_map(|item| item.id.parse::<TerminalTitleItem>().ok())
+                    .collect::<Vec<_>>();
+                app_event.send(AppEvent::TerminalTitleSetupPreview { items });
+            })
+            .on_confirm(|ids, app_event| {
+                let items = ids
+                    .iter()
+                    .map(|id| id.parse::<TerminalTitleItem>())
+                    .collect::<Result<Vec<_>, _>>()
+                    .unwrap_or_default();
```
jif-oai: We shouldn't call `app_event.send()` if the parsing fail

codex-rs/tui/src/chatwidget.rs
```diff
@@ -991,6 +1092,64 @@ impl ChatWidget {
...
+        }
+    }
+
+    pub(crate) fn refresh_status_line(&mut self) {
+        let selections = self.status_surface_selections();
+        self.warn_invalid_status_line_items_once(&selections.invalid_status_line_items);
+        self.warn_invalid_terminal_title_items_once(&selections.invalid_terminal_title_items);
+        self.sync_status_surface_shared_state(&selections);
+        self.refresh_status_line_from_selections(&selections);
+        self.refresh_terminal_title_from_selections(&selections);
```
jif-oai: strange to update the title in a function named "..._status_line"

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -6333,12 +6333,7 @@ async fn handle_pending_thread_resume_request(
             return;
```
jif-oai: Why did you touch this file? This look like Codex doing stuff without reading through it after...

yvolovich-cyber: ah, sorry for the messup, this is me managing git branches incorrectly :) reverted

codex-rs/tui/src/chatwidget.rs
yvolovich-cyber: Renamed the shared entrypoint to `refresh_status_surfaces()` so the name matches what it does. `refresh_status_line_from_selections()` stays as the footer-only renderer.

codex-rs/tui/src/bottom_pane/title_setup.rs
yvolovich-cyber: Agreed. `on_change()` and `on_confirm()` now return early on parse failure instead of sending an app event. I also factored the parsing into `parse_terminal_title_items()` so preview, change, and confirm all share the same validation path.

codex-rs/tui/src/terminal_title.rs
yvolovich-cyber: There is no portable and reliable way to restore (specifically read) terminal title.
We only clear the title Codex set, not to restore whatever title was there before, which keeps the behavior portable across terminals.
I also changed `set_terminal_title()` so "sanitized to no visible content" is no longer treated as an implicit `clear_terminal_title()`.

codex-rs/tui/src/chatwidget.rs
yvolovich-cyber: Agreed. I kept this synchronous for now, but I narrowed the hot path so `set_status()` only refreshes the terminal title when `/title` actually includes `status`, and the shared status-surface path now reuses cached/shared state instead of recomputing it. I think making the remaining title refresh path async can be a follow-up.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -7965,8 +8402,11 @@ fn has_websocket_timing_metrics(summary: RuntimeMetricsSummary) -> bool {
 
 impl Drop for ChatWidget {
     fn drop(&mut self) {
-        self.reset_realtime_conversation_state();
+        if let Err(err) = self.clear_managed_terminal_title() {
+            tracing::debug!(error = %err, "failed to clear terminal title on drop");
```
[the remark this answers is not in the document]
yvolovich-cyber: Agreed. I changed the replacement path to go through `App::replace_chat_widget()`, which first calls `skip_terminal_title_cleanup_on_drop()` on the old widget and only then assigns the replacement. That keeps drop-time cleanup for real app exit, but avoids clearing the new widget title during thread and session switches. I also added regression coverage in `app.rs` and `chatwidget/tests.rs`. Fixed in `781b749b8`.

codex-rs/core/src/config/edit.rs
```diff
@@ -76,6 +80,22 @@ pub fn status_line_items_edit(items: &[String]) -> ConfigEdit {
...
 
+/// Produces a config edit that sets `[tui].terminal_title` to an explicit ordered list.
+///
+/// The array is written even when it is empty so "disabled title updates" stays
+/// distinct from "unset, so use defaults".
+pub fn terminal_title_items_edit(items: &[String]) -> ConfigEdit {
+    let mut array = toml_edit::Array::new();
+    for item in items {
+        array.push(item.clone());
+    }
```
joshka-oai: nit: in idiomatic rust this should be something like:

```suggestion
    let array = items.cloned().collect<toml_edit::Array>();
```

I'm unsure if Array has a `FromIter` impl that would make this just work, worth a small follow up PR if it doesn't.

codex-rs/tui/src/bottom_pane/title_setup.rs
```diff
@@ -0,0 +1,273 @@
...
+    /// Compact runtime status (Ready, Working, Thinking, ...).
+    Status,
+    /// Current thread title (if available).
+    Thread,
+    /// Current git branch (if available).
+    GitBranch,
+    /// Current model name.
+    Model,
+    /// Latest checklist task progress from `update_plan` (if available).
+    #[strum(to_string = "task-progress")]
```
joshka-oai: Any idea why this is necessary? Seems it should be picked up by the kebab_case serialization above.

codex-rs/tui/src/bottom_pane/title_setup.rs
```diff
@@ -0,0 +1,273 @@
...
+            TerminalTitleItem::Thread => "Current thread title (omitted until available)",
+            TerminalTitleItem::GitBranch => "Current Git branch (omitted when unavailable)",
+            TerminalTitleItem::Model => "Current model name",
+            TerminalTitleItem::TaskProgress => {
+                "Latest task progress from update_plan (omitted until available)"
+            }
+        }
+    }
+
+    pub(crate) fn render(&self) -> &'static str {
```
joshka-oai: This is oddly named for something that returns static strings that are examples of the values that will be used. Reading this I was initially confused about why this always returns e.g. `my-project`. Add a comment about what this is for.

codex-rs/tui/src/bottom_pane/title_setup.rs
```diff
@@ -0,0 +1,273 @@
...
+            }
+        }
+
+        for item in TerminalTitleItem::iter() {
+            let item_id = item.to_string();
+            if used_ids.contains(&item_id) {
+                continue;
+            }
+            items.push(Self::title_select_item(item, false));
+        }
```
joshka-oai: I suspect this could be much simpler and more readable with itertools.
Something like the following (ask codex to simplify with itertools)

```suggestion
        let ticked = title_items.iter().unique().filter_map(|id| id.parse::<TerminalTitleItem>).collect_vec();
        let unticked = TerminalTitleItem::iter().filter(|id| ticked.contains(id));
        let items = ticked.chain(unticked);
```

codex-rs/tui/src/terminal_title.rs
```diff
@@ -0,0 +1,184 @@
+//! Terminal-title output helpers for the TUI.
+//!
+//! This module owns the low-level OSC title write path and the sanitization
+//! that happens immediately before we emit it. It is intentionally narrow:
+//! callers decide when the title should change and whether an empty title means
+//! "leave the old title alone" or "clear the title Codex last wrote".
+//! This module does not attempt to read or restore the terminal's previous
+//! title because that is not portable across terminals.
```
joshka-oai, line 8: It's worth noting, probably here, why sanitization is necessary and what it means.

codex-rs/tui/src/terminal_title.rs
```diff
@@ -0,0 +1,184 @@
...
+        ))
+    }
+
+    #[cfg(windows)]
+    fn is_ansi_code_supported(&self) -> bool {
+        true
+    }
+}
+
+fn sanitize_terminal_title(title: &str) -> String {
```
joshka-oai, line 104: Add a doc here to tell the devs what the intention is of sanitization (mechanically)

codex-rs/tui/src/terminal_title.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    sanitized
+}
+
+fn is_disallowed_terminal_title_char(ch: char) -> bool {
+    if ch.is_control() {
+        return true;
+    }
+
+    // Strip Trojan-Source-related bidi controls plus common non-rendering
```
joshka-oai, line 147: Explain trojan-source-related here?

codex-rs/tui/src/chatwidget.rs
```diff
@@ -670,6 +684,18 @@ pub(crate) struct ChatWidget {
     session_network_proxy: Option<codex_protocol::protocol::SessionNetworkProxyRuntime>,
     // Shared latch so we only warn once about invalid status-line item IDs.
     status_line_invalid_items_warned: Arc<AtomicBool>,
+    // Shared latch so we only warn once about invalid terminal-title item IDs.
+    terminal_title_invalid_items_warned: Arc<AtomicBool>,
+    // Last terminal title emitted, to avoid writing duplicate OSC updates.
+    last_terminal_title: Option<String>,
+    // Clears the terminal title on drop unless the widget is being intentionally replaced.
+    clear_terminal_title_on_drop: bool,
```
joshka-oai: Would it simplify this code to instead clear when codex finishes?

codex-rs/tui/src/chatwidget.rs
joshka-oai, line 1: chatwidget.rs is a very large module and it tends to collect random implementations of things. Prefer to pull new additive code modules out into a small modules where possible (ask code to do that). This makes the tests and helper functions tend to be closely located and makes it easier for humans to orient on the new code.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -1030,6 +1157,80 @@ impl ChatWidget {
...
+    pub(crate) fn skip_terminal_title_cleanup_on_drop(&mut self) {
+        self.clear_terminal_title_on_drop = false;
+    }
+
+    #[cfg(test)]
+    pub(crate) fn clears_terminal_title_on_drop(&self) -> bool {
+        self.clear_terminal_title_on_drop
+    }
+
+    fn refresh_terminal_title_from_selections(&mut self, selections: &StatusSurfaceSelections) {
```
joshka-oai: Complex enough to require a doc comment

joshka-oai (commented):
Incomplete review - I still need to review chatwidget.rs in more detail, but if you agree to extract the new code to another module, then it's possible that would be worth reviewing after that change.

yvolovich-cyber:
@codex review

codex-rs/core/src/config/edit.rs
yvolovich-cyber: Switched both config array builders to `items.iter().cloned().collect::<toml_edit::Array>()` so the status-line and terminal-title edits both use the idiomatic path.

codex-rs/tui/src/bottom_pane/title_setup.rs
yvolovich-cyber: Removed the explicit `#[strum(to_string = "task-progress")]`; `serialize_all = "kebab_case"` already produces the right wire value.

codex-rs/tui/src/bottom_pane/title_setup.rs
yvolovich-cyber: Renamed this to `preview_example()` and added a doc comment clarifying that it returns static sample text for the picker preview, not live session values.

codex-rs/tui/src/bottom_pane/title_setup.rs
yvolovich-cyber: Simplified the picker assembly with `itertools::Itertools::unique()`, preserving the configured valid selection order first and then appending the remaining items.

codex-rs/tui/src/terminal_title.rs
yvolovich-cyber: Expanded the module docs to explain why title sanitization is needed before writing OSC, specifically because the title can include untrusted model output, thread names, paths, and config text.

codex-rs/tui/src/terminal_title.rs
yvolovich-cyber: Added docs on `set_terminal_title()` and `sanitize_terminal_title()` describing the mechanics: collapse whitespace, drop disallowed codepoints, and bound the visible length before emitting OSC 0.

codex-rs/tui/src/terminal_title.rs
yvolovich-cyber: Added an explicit note in the sanitization docs/comments that bidi and invisible formatting codepoints are stripped to avoid Trojan-Source-style misleading rendering.

codex-rs/tui/src/chatwidget.rs
yvolovich-cyber: Moved terminal-title cleanup to `App::drop`, so cleanup now happens once when Codex exits instead of when a `ChatWidget` instance is replaced during thread and session switches.

codex-rs/tui/src/chatwidget.rs
yvolovich-cyber: Pulled the new title and status-surface logic into `tui/src/chatwidget/status_surfaces.rs` so the additive code, helper logic, and related docs stay together instead of expanding `chatwidget.rs`.

codex-rs/tui/src/chatwidget.rs
yvolovich-cyber: Added doc comments on the shared status-surface refresh path and the terminal-title render and cleanup path in the extracted module so the control flow is easier to follow.

codex-rs/tui/src/chatwidget/status_surfaces.rs
```diff
@@ -0,0 +1,623 @@
+//! Status-line and terminal-title rendering helpers for `ChatWidget`.
+//!
+//! Keeping this logic in a focused submodule makes the additive title/status
+//! behavior easier to review without paging through the rest of `chatwidget.rs`.
+
+use super::*;
+
+pub(super) const DEFAULT_TERMINAL_TITLE_ITEMS: [&str; 2] = ["status", "thread"];
```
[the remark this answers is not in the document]
yvolovich-cyber: Default terminal title items now match the current documented `project,spinner` behavior. `DEFAULT_TERMINAL_TITLE_ITEMS` uses `project` and `spinner`, so unset `tui.terminal_title` keeps project disambiguation while leaving status text opt-in.

codex-rs/tui/src/chatwidget/status_surfaces.rs
```diff
@@ -0,0 +1,623 @@
...
+    /// Clears the terminal title Codex most recently wrote, if any.
+    ///
+    /// This does not attempt to restore the shell or terminal's previous title;
+    /// it only clears the managed title and updates the cache after a successful
+    /// OSC write.
+    pub(crate) fn clear_managed_terminal_title(&mut self) -> std::io::Result<()> {
+        if self.last_terminal_title.is_some() {
+            clear_terminal_title()?;
+            self.last_terminal_title = None;
+        }
```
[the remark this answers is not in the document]
yvolovich-cyber: Handled in `replace_chat_widget()`: the previous cache is only carried forward when the replacement widget has no cached title, so empty or disabled replacement titles still clear stale title state instead of losing the cleanup path. Added regression coverage for that empty-replacement case in `app.rs`.

codex-rs/tui/src/app.rs
```diff
@@ -1525,6 +1528,11 @@ impl App {
         self.chat_widget.set_pending_thread_approvals(Vec::new());
     }
 
+    fn replace_chat_widget(&mut self, mut chat_widget: ChatWidget) {
+        chat_widget.last_terminal_title = self.chat_widget.last_terminal_title.take();
```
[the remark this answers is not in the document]
yvolovich-cyber: Handled in `replace_chat_widget()`: a replacement widget keeps its own `last_terminal_title` when it already refreshed and cached a new title. The old cache is only carried forward when the replacement cache is `None`, and `app.rs` now covers both cases.

codex-rs/tui/src/app.rs
```diff
@@ -1557,8 +1568,9 @@ impl App {
             startup_tooltip_override: None,
             status_line_invalid_items_warned: self.status_line_invalid_items_warned.clone(),
             session_telemetry: self.session_telemetry.clone(),
+            terminal_title_invalid_items_warned: self.terminal_title_invalid_items_warned.clone(),
         };
-        self.chat_widget = ChatWidget::new(init, self.server.clone());
+        self.replace_chat_widget(ChatWidget::new(init, self.server.clone()));
```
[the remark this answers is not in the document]
yvolovich-cyber: Fixed in `c3d3e54`. `replace_chat_widget()` now refreshes status surfaces immediately after the swap, so start-fresh, resume, fork, and replay paths all recompute or clear the terminal title from the replacement widget's config instead of carrying a stale cached value. Added regression coverage for both clearing an inherited title when the replacement disables titles and preserving the replacement's own configured title.

codex-rs/tui/src/chatwidget/status_surfaces.rs
```diff
@@ -0,0 +1,631 @@
...
+    /// Startup takes precedence over normal task states, and idle state renders
+    /// as `Ready` regardless of the last active status bucket.
+    pub(super) fn terminal_title_status_text(&self) -> String {
+        if self.mcp_startup_status.is_some() {
+            return "Starting".to_string();
+        }
+
+        if !self.bottom_pane.is_task_running() {
+            return "Ready".to_string();
+        }
```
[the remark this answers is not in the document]
yvolovich-cyber: Handled in `42a73c1`. Undo now counts as active terminal-title progress even though the bottom-pane task-running flag stays false, so `/title` configs with `status` or `spinner` render `Undoing` and the spinner during undo. Added regression coverage for the no-task-running undo path.

codex-rs/tui/src/chatwidget/status_surfaces.rs
```diff
@@ -0,0 +1,644 @@
...
+            .fold(String::new(), |mut title, (item, value)| {
+                title.push_str(item.separator_from_previous(previous));
+                title.push_str(&value);
+                previous = Some(item);
+                title
+            });
+        let title = (!title.is_empty()).then_some(title);
+        if self.last_terminal_title == title {
+            return;
+        }
```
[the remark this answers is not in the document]
yvolovich-cyber: Handled in 5173c9465. Spinner refresh scheduling now happens even when the rendered title text is unchanged, and the regression test asserts that the next frame request is still enqueued on that no-op refresh path.

codex-rs/tui/src/chatwidget/status_surfaces.rs
```diff
@@ -0,0 +1,650 @@
+//! Status-line and terminal-title rendering helpers for `ChatWidget`.
+//!
+//! Keeping this logic in a focused submodule makes the additive title/status
+//! behavior easier to review without paging through the rest of `chatwidget.rs`.
+
+use super::*;
+
+pub(super) const DEFAULT_TERMINAL_TITLE_ITEMS: [&str; 2] = ["project", "spinner"];
```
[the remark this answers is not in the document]
yvolovich-cyber: Updated the PR description to match the current behavior. The implementation, config docs, and picker intentionally use `project,spinner` as the default now; `status` is a separate text item, so `spinner,status` gives `⠋ Working` when that denser title is wanted.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -1066,6 +1102,14 @@ impl ChatWidget {
         self.current_status_header = header.clone();
         self.bottom_pane
             .update_status(header, details, details_capitalization, details_max_lines);
+        if self
+            .config
+            .tui_terminal_title
+            .as_ref()
+            .is_none_or(|items| items.iter().any(|item| item == "status"))
+        {
```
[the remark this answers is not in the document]
yvolovich-cyber: Handled in `f6b501db9`. `set_status()` no longer treats the default title config as if `status` were selected. It now refreshes title text for explicit `status` usage, while still refreshing the spinner-only default for the undo path where active progress is driven by `Undoing` rather than the task-running flag. Added regression coverage for that default `spinner,project` undo case.

joshka-oai (approved):

codex-rs/core/src/memories/prompts.rs
joshka-oai, line 1: This whole commit seems like it's unrelated to this PR (unless I'm missing something here).
Sometimes codex gets a little too eager to fix things which shouldn't be part of a PR - guessing that happened here?

codex-rs/core/Cargo.toml
```diff
@@ -21,7 +21,6 @@ anyhow = { workspace = true }
 arc-swap = "1.8.2"
 async-channel = { workspace = true }
 async-trait = { workspace = true }
-askama = { workspace = true }
```
joshka-oai, line 24: Same here (seems not really part of this PR)

codex-rs/core/src/memories/prompts.rs
yvolovich-cyber: yes, sorry, I'll review manually the things

codex-rs/core/Cargo.toml
yvolovich-cyber: sorry, dropped that commit

joshka-oai (approved):

PR #12357, opened 2026-02-20
feat(core): persist network approvals in execpolicy

viyatb-oai:
## Summary
Persist network approval allow/deny decisions as `network_rule(...)` entries in execpolicy (not proxy config)

It adds `network_rule` parsing + append support in `codex-execpolicy`, including `decision="prompt"` (parse-only; not compiled into proxy allow/deny lists)
- compile execpolicy network rules into proxy allow/deny lists and update the live proxy state on approval
- preserve requirements execpolicy `network_rule(...)` entries when merging with file-based execpolicy
- reject broad wildcard hosts (for example `*`) for persisted `network_rule(...)`


codex-rs/core/src/codex.rs
```diff
@@ -2216,6 +2222,118 @@ impl Session {
...
+            NetworkPolicyRuleAction::Allow => (ExecPolicyDecision::Allow, "Allow"),
+            NetworkPolicyRuleAction::Deny => (ExecPolicyDecision::Forbidden, "Deny"),
+        };
+        let protocol_label = match network_approval_context.protocol {
+            NetworkApprovalProtocol::Http => "http",
+            NetworkApprovalProtocol::Https => "https_connect",
+            NetworkApprovalProtocol::Socks5Tcp => "socks5_tcp",
+            NetworkApprovalProtocol::Socks5Udp => "socks5_udp",
+        };
+        let justification = format!("{action_verb} {protocol_label} access to {host}");
```
dylan-hurd-oai: let's pull this out of codex.rs

viyatb-oai: done!

codex-rs/execpolicy/src/amend.rs
```diff
@@ -90,6 +97,66 @@ pub fn blocking_append_allow_prefix_rule(
...
+    match std::fs::create_dir(dir) {
+        Ok(()) => {}
+        Err(ref source) if source.kind() == std::io::ErrorKind::AlreadyExists => {}
+        Err(source) => {
+            return Err(AmendError::CreatePolicyDir {
+                dir: dir.to_path_buf(),
+                source,
+            });
+        }
+    }
```
dylan-hurd-oai, line 143: can we deduplicate this with `blocking_append_allow_prefix_rule`

codex-rs/execpolicy/src/amend.rs
```diff
@@ -215,4 +282,44 @@ prefix_rule(pattern=["echo", "Hello, world!"], decision="allow")
...
+        )
+        .expect("append network rule");
+
+        let contents = std::fs::read_to_string(&policy_path).expect("read policy");
+        assert_eq!(
+            contents,
+            r#"network_rule(host="api.github.com", protocol="https", decision="allow", justification="Allow https_connect access to api.github.com")
+"#
+        );
+    }
```
dylan-hurd-oai, line 294: let's add another test that adds both network and prefix rules

codex-rs/network-proxy/src/runtime.rs
```diff
@@ -527,6 +590,41 @@ impl NetworkProxyState {
...
+            Self::Allow => (
+                &mut config.network.allowed_domains,
+                &mut config.network.denied_domains,
+            ),
+            Self::Deny => (
+                &mut config.network.denied_domains,
+                &mut config.network.allowed_domains,
+            ),
+        }
+    }
```
dylan-hurd-oai, line 630: This function feels weird - shouldn't this be a method on `NetworkProxyConfig`?

dylan-hurd-oai (commented):
Couple smaller comments - then would like to add some integration tests for this! Similar to [approvals.rs](https://github.com/openai/codex/blob/1af2a37ada16b8227ecd1b8e8411d2449d97ec21/codex-rs/core/tests/suite/approvals.rs#L199)

Once these are addressed, looks good

codex-rs/network-proxy/src/runtime.rs
viyatb-oai: added

viyatb-oai:
adding a allow positive test needed harness changes in how the cloud requirements are setup in the test, so skipped that.

codex-rs/network-proxy/src/runtime.rs
```diff
@@ -526,7 +526,7 @@ impl NetworkProxyState {
             };
 
             let mut candidate = previous_cfg.clone();
-            let (target_entries, opposite_entries) = target.split_lists(&mut candidate);
+            let (target_entries, opposite_entries) = candidate.split_domain_lists_mut(target);
```
dylan-hurd-oai, line 529: This is not blocking, but I'm curious about this abstraction and if we can simplify this to avoid passing around mutable references - can fix in a follow-up PR.

dylan-hurd-oai (approved):

PR #12370, opened 2026-02-20
Add ability to attach extra files to feedback

pakrym-oai:
Allow clients to provide extra files.


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1506,6 +1471,7 @@ pub struct FeedbackUploadParams {
     #[ts(optional = nullable)]
     pub thread_id: Option<String>,
     pub include_logs: bool,
+    pub extra_log_files: Vec<PathBuf>,
```
sayan-oai: required? im assuming this is fine for us because extension/app update client + server together, but for 3p this could be breaking?

sayan-oai (approved):
one question/nit but lgtm

codex-rs/app-server-protocol/src/protocol/v2.rs
sayan-oai: hmm i guess that would be super unlikely for feedback upload though. nvm

pakrym-oai: no, it should be optional, you are right

PR #12383, opened 2026-02-20
add new scopes to login

adaley-openai:
Validated login + refresh flows. Removing scopes from the refresh request until we have upgrade flow in place. Confirmed that tokens refresh with existing scopes.

### DIFF
--- a/codex-rs/core/src/auth.rs
+++ b/codex-rs/core/src/auth.rs
@@ -626,7 +626,9 @@ async fn request_chatgpt_token_refresh(
         client_id: CLIENT_ID,
         grant_type: "refresh_token",
         refresh_token,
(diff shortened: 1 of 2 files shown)


codex-rs/core/src/auth.rs
```diff
@@ -717,7 +719,7 @@ struct RefreshRequest {
     client_id: &'static str,
     grant_type: &'static str,
     refresh_token: String,
-    scope: &'static str,
+    //scope: &'static str,
```
hanyu-openai: nit: do  we want to keep commented code or remove it?

hanyu-openai (approved):

caseychow-oai (approved):

PR #12407, opened 2026-02-21
tests(js_repl): stabilize CI runtime test execution

fjord-oai:
## Summary

Stabilize `js_repl` runtime test setup in CI and move tool-facing `js_repl` behavior coverage into integration tests.

This is a test/CI change only. No production `js_repl` behavior change is intended.

## Why

- Bazel test sandboxes (especially on macOS) could resolve a different `node` than the one installed by `actions/setup-node`, which caused `js_repl` runtime/version failures.
- `js_repl` runtime tests depend on platform-specific sandbox/test-harness behavior, so they need explicit gating in a base-stability commit.
- Several tests in the `js_repl` unit test module were actually black-box/tool-level behavior tests and fit better in the integration suite.

## Changes

- Add `actions/setup-node` to the Bazel and Rust `Tests` workflows, using the exact version pinned in the repo’s Node version file.
- In Bazel (non-Windows), pass `CODEX_JS_REPL_NODE_PATH=$(which node)` into test env so `js_repl` uses the `actions/setup-node` runtime inside Bazel tests.
- Add a new integration test suite for `js_repl` tool behavior and register it in the core integration test suite module.
- Move black-box `js_repl` behavior tests into the integration suite (persistence/TLA, builtin tool invocation, recursive self-call rejection, `process` isolation, blocked builtin imports).
- Keep white-box manager/kernel tests in the `js_repl` unit test module.
- Gate `js_repl` runtime tests to run only on macOS and only when a usable Node runtime is available (skip on other platforms / missing Node in this commit).

## Impact

- Reduces `js_repl` CI failures caused by Node resolution drift in Bazel.
- Improves test organization by separating tool-facing behavior tests from white-box manager/kernel tests.
- Keeps the base commit stable while expanding `js_repl` runtime coverage.

- ✅ `1` https://github.com/openai/codex/pull/12372
- 👉 `2` https://github.com/openai/codex/pull/12407
- ⏳ `3` https://github.com/openai/codex/pull/12185
- ⏳ `4` https://github.com/openai/codex/pull/10673

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -47,6 +47,11 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      - name: Set up Node.js for js_repl tests
+        uses: actions/setup-node@v6
+        with:
+          node-version-file: codex-rs/node-version.txt
+
       # Some integration tests rely on DotSlash being installed.
       # See https://github.com/openai/codex/pull/7617.
       - name: Install DotSlash
@@ -155,6 +160,13 @@ jobs:
             --build_metadata=ROLE=CI
             --build_metadata=VISIBILITY=PUBLIC
           )
+
+          if [[ "${RUNNER_OS:-}" != "Windows" ]]; then
+            # Bazel test sandboxes on macOS may resolve an older Homebrew `node`
+            # before the `actions/setup-node` runtime on PATH.
+            node_bin="$(which node)"
+            bazel_args+=("--test_env=CODEX_JS_REPL_NODE_PATH=${node_bin}")
+          fi
 
           if [[ -n "${BUILDBUDDY_API_KEY:-}" ]]; then
             echo "BuildBuddy API key is available; using remote Bazel configuration."
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -499,6 +499,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+      - name: Set up Node.js for js_repl tests
+        uses: actions/setup-node@v6
+        with:
+          node-version-file: codex-rs/node-version.txt
       - name: Install Linux build dependencies
         if: ${{ runner.os == 'Linux' }}
         shell: bash
--- a/codex-rs/core/src/tools/js_repl/mod.rs
+++ b/codex-rs/core/src/tools/js_repl/mod.rs
@@ -1636,6 +1636,12 @@ mod tests {
     }
 
     async fn can_run_js_repl_runtime_tests() -> bool {
+        // These white-box runtime tests rely on the unit-test harness and are
+        // only required on macOS. Linux uses the codex-linux-sandbox arg0
+        // dispatch path, which is exercised in integration tests instead.
+        if !cfg!(target_os = "macos") {
+            return false;
+        }
         if std::env::var_os("CODEX_SANDBOX").is_some() {
             return false;
         }
@@ -1666,47 +1672,6 @@ mod tests {
             pkg_dir.join("index.js"),
             format!("export const value = \"{value}\";\n"),
         )?;
-        Ok(())
-    }
-
-    #[tokio::test]
-    async fn js_repl_persists_top_level_bindings_and_supports_tla() -> anyhow::Result<()> {
-        if !can_run_js_repl_runtime_tests().await {
-            return Ok(());
-        }
-
-        let (session, turn) = make_session_and_context().await;
-        let session = Arc::new(session);
-        let turn = Arc::new(turn);
-        let tracker = Arc::new(tokio::sync::Mutex::new(TurnDiffTracker::default()));
-        let manager = turn.js_repl.manager().await?;
-
-        let first = manager
-            .execute(
-                Arc::clone(&session),
-                Arc::clone(&turn),
-                Arc::clone(&tracker),
-                JsReplArgs {
-                    code: "let x = await Promise.resolve(41); console.log(x);".to_string(),
-                    timeout_ms: Some(10_000),
-                },
-            )
-            .await?;
-        assert!(first.output.contains("41"));
-
-        let second = manager
-            .execute(
-                Arc::clone(&session),
-                Arc::clone(&turn),
-                Arc::clone(&tracker),
-                JsReplArgs {
-                    code: "console.log(x + 1);".to_string(),
-                    timeout_ms: Some(10_000),
-                },
-            )
-            .await?;
-
-        assert!(second.output.contains("42"));
         Ok(())
     }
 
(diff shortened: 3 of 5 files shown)


codex-rs/.config/nextest.toml
```diff
@@ -11,3 +11,12 @@ slow-timeout = { period = "1m", terminate-after = 4 }
 [[profile.default.overrides]]
 filter = 'test(approval_matrix_covers_all_modes)'
 slow-timeout = { period = "30s", terminate-after = 2 }
+
+[[profile.default.overrides]]
```
bolinfest: I did not know about this config file! Clearly I missed https://github.com/openai/codex/pull/6612.

codex-rs/core/src/tools/js_repl/mod.rs
```diff
@@ -1287,18 +1287,51 @@ pub(crate) fn resolve_node(config_path: Option<&Path>) -> Option<PathBuf> {
...
+    use ctor::ctor;
     use pretty_assertions::assert_eq;
     use std::fs;
     use std::path::Path;
+    #[cfg(target_os = "linux")]
+    use std::path::PathBuf;
     use tempfile::tempdir;
 
+    #[cfg(target_os = "linux")]
+    #[ctor]
```
bolinfest: Using `#[ctor]` in a unit test seems....suboptimal. Perhaps these should become integration tests that leverage `codex-rs/core/tests/suite/mod.rs`?

fjord-oai: fixed by moving tests that were more integratino-like to tests/suite and then switching the remaining tests here to yolo mode

.github/workflows/bazel.yml
```diff
@@ -47,6 +47,16 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      - name: Set up Node.js for js_repl tests
+        uses: actions/setup-node@v6
+        with:
+          node-version-file: codex-rs/node-version.txt
+          check-latest: true
```
bolinfest: I would prefer not to do this: this means behavior changes can be introduced that are not tied to commits.

.github/workflows/bazel.yml
```diff
@@ -47,6 +47,16 @@ jobs:
...
+
+      - name: Make Node.js available in PATH (Unix)
+        if: runner.os != 'Windows'
+        run: cp "$(which node)" /usr/local/bin
```
bolinfest: What uses `/usr/local/bin/node` instead of the first thing returned by `which node`? Is a test running with a more restricted `$PATH` than whatever is used at this moment for `which`? Or does it just need to be at a fixed location for test purposes?

.github/workflows/rust-ci.yml
```diff
@@ -499,6 +499,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+      - name: Set up Node.js for js_repl tests
+        uses: actions/setup-node@v6
+        with:
+          node-version-file: codex-rs/node-version.txt
+          check-latest: true
```
bolinfest: Same here.

codex-rs/core/src/tools/js_repl/mod.rs
```diff
@@ -1287,18 +1287,25 @@ pub(crate) fn resolve_node(config_path: Option<&Path>) -> Option<PathBuf> {
...
-    use codex_protocol::models::ResponseInputItem;
-    use codex_protocol::openai_models::InputModality;
     use pretty_assertions::assert_eq;
     use std::fs;
     use std::path::Path;
     use tempfile::tempdir;
 
+    fn configure_js_repl_test_sandbox(turn: &mut crate::codex::TurnContext) {
+        // Manager-level runtime tests don't need to exercise Linux arg0 sandbox dispatch.
+        turn.sandbox_policy = SandboxPolicy::DangerFullAccess;
```
bolinfest: This should not be possible...let me send a PR.

Specifically, these fields should be `Constrained<T>` and it should not be possible to modify them in ways that violate `requirements.toml`.

bolinfest: See https://github.com/openai/codex/pull/12473.

.github/workflows/bazel.yml
fjord-oai: Good catch, i removed the `check-latest: true`, so it should be pinned to exactly `codex-rs/node-version.txt`, which is version controlled.

.github/workflows/rust-ci.yml
fjord-oai: also fixed

.github/workflows/bazel.yml
fjord-oai: The `/usr/local/bin/node` path was just a stable way to pass the `actions/setup-node` binary into Bazel tests, since the original issue was that sandboxed tests (especially on macOS) could resolve a different `PATH` and pick the wrong `node`.

I’ve switched it to pass `$(which node)` instead, so we still force Bazel tests to use the setup-node runtime (via `CODEX_JS_REPL_NODE_PATH`), but no longer rely on copying it to a fixed path. Verified this passes CI.

codex-rs/core/src/tools/js_repl/mod.rs
fjord-oai: Switched to the new pattern of `.set(SandboxPolicy::DangerFullAccess)`

codex-rs/.config/nextest.toml
fjord-oai: no longer needed since we're skipping windows tests

codex-rs/core/src/tools/js_repl/mod.rs
```diff
@@ -2081,7 +1813,7 @@ console.log(out.output?.body?.text ?? "");
         let config_base = tempdir()?;
         let cwd_dir = tempdir()?;
 
-        let (session, mut turn) = make_session_and_context().await;
+        let (session, mut turn) = crate::codex::make_session_and_context().await;
```
bolinfest: I don't want you to have to wait for CI again, but in general, relying on the import is cleaner (also, this change seems to add unnecessary noise to this PR?).

codex-rs/core/tests/suite/js_repl.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    test.submit_turn(prompt).await?;
+    Ok(second_mock)
+}
+
+fn can_run_js_repl_runtime_tests() -> bool {
+    // These js_repl runtime coverage tests are required on macOS. On other
+    // platforms we skip here and rely on the integration paths that are wired
+    // for those sandbox launch flows.
+    if !cfg!(target_os = "macos") {
```
bolinfest: Don't these integration tests have the `ctor` behavior that you need?

bolinfest (approved):

codex-rs/core/tests/suite/js_repl.rs
fjord-oai: thanks for catching this, removed!

codex-rs/core/src/tools/js_repl/mod.rs
fjord-oai: fixed!

PR #12441, opened 2026-02-21
Delete AggregatedStream

pakrym-oai:
Used only in test


bolinfest (approved):
D . E . L . E . T . E . D

PR #12444, opened 2026-02-21
feat(tui) /clear

won-openai:
# /clear feature! 

/clear will clear your terminal while preserving the context/state of the thread.


aibrahim-oai (commented):
add snapshots

Issue #12453, opened 2026-02-21
TUI: Alt-d should delete next word in chat prompt


dougEfresh:
## Summary
In the TUI chat prompt, Alt-d should delete the next word (readline-style). Currently it does nothing.

## Expected behavior
Alt-d deletes the word ahead of the cursor without affecting earlier text.

## Actual behavior
No deletion occurs when pressing Alt-d in the chat prompt.

## Rationale
Readline-compatible word editing shortcuts are standard for terminal users, and Alt-b/Alt-f already move by word in the prompt. Adding Alt-d completes the common word-edit trio.

## Environment
- OS: Arch Linux
- Terminal: Ghostty
- Shell: zsh

## Additional context
I have a small patch + test ready locally and will wait for an invitation before opening a PR if that’s preferred.

PR #12455, opened 2026-02-21
feat(tui): support Alt-d delete-forward-word

dougEfresh:
Alt-d should delete the next word. It didn’t. Now it does. Added a small test so it stays that way.

Details:
File updated: [codex-rs/tui/src/bottom_pane/textarea.rs](./codex-rs/tui/src/bottom_pane/textarea.rs) 
Test added: delete_forward_word_alt_d — verifies Alt-d deletes the next word and keeps the cursor position correct.

Solves  Issue #12453

### DIFF
--- a/codex-rs/tui/src/bottom_pane/textarea.rs
+++ b/codex-rs/tui/src/bottom_pane/textarea.rs
@@ -322,7 +322,12 @@ impl TextArea {
                 code: KeyCode::Delete,
                 modifiers: KeyModifiers::ALT,
                 ..
-            }  => self.delete_forward_word(),
+            }
+            | KeyEvent {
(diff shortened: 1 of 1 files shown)


dougEfresh:

dougEfresh:
@etraut-openai  Totally understood. 

Just saying this is a 4 line change (excluding test) which addresses a mismatch between standard shell key bindings (i.e. readline)  and codex TUI.

etraut-openai:
@codex review

etraut-openai:
Even simple keybinding changes like this can be trickier that one might think because of the wide variety of platforms and terminal emulators that we need to support, but this one looks fine to me.

PR #12517, opened 2026-02-22
codex-rs/app-server: graceful websocket restart on Ctrl-C

maxj-oai:
## Summary
- add graceful websocket app-server restart on Ctrl-C by draining until no assistant turns are running
- stop the websocket acceptor and disconnect existing connections once the drain condition is met
- add a websocket integration test that verifies Ctrl-C waits for an in-flight turn before exit

## Verification
- `cargo check -p codex-app-server --quiet`
- `cargo test -p codex-app-server --test all suite::v2::connection_handling_websocket`
- I (maxj) tested remote and local Codex.app


codex-rs/app-server/src/lib.rs
```diff
@@ -255,17 +257,27 @@ pub async fn run_main_with_transport(
...
+                start_websocket_acceptor(
+                    bind_address,
+                    transport_event_tx.clone(),
+                    shutdown_token.clone(),
+                )
+                .await?,
+            );
+            websocket_accept_shutdown = Some(shutdown_token);
         }
     }
```
bolinfest: Use an `enum` as a better way to represent this state. Something like this:

```suggestion
    enum TransportState {
        Stdio,
        WebSocket {
            accept_handle: JoinHandle<()>,
            shutdown_token: CancellationToken,
        }
    }
    let transport_state = match transport {
AppServerTransport::Stdio => {
start_stdio_connection(transport_event_tx.clone(), &mut stdio_handles).await?;
TransportState::Stdio
}
AppServerTransport::WebSocket { bind_address } => {
            let shutdown_token = CancellationToken::new();
            let accept_handle = Some(
                start_websocket_acceptor(
                    bind_address,
                    transport_event_tx.clone(),
                    shutdown_token.clone(),
                )
                .await?,
            );
            TransportState { accept_handle, shutdown_token }
};
}
```

codex-rs/app-server/tests/suite/v2/connection_handling_websocket.rs
```diff
@@ -1,19 +1,35 @@
 use anyhow::Context;
 use anyhow::Result;
 use anyhow::bail;
+#[cfg(unix)]
```
bolinfest: Is websocket UNIX-specific, in general? If so, it feels like this would be cleaner if the UNIX-specific tests lived in their own file.

codex-rs/app-server/src/lib.rs
```diff
@@ -464,11 +486,69 @@ pub async fn run_main_with_transport(
...
+        let mut running_turn_count_rx = processor.subscribe_running_assistant_turn_count();
         let mut connections = HashMap::<ConnectionId, ConnectionState>::new();
+        let websocket_accept_shutdown = websocket_accept_shutdown.clone();
         async move {
             let mut listen_for_threads = true;
+            let mut restart_requested = false;
+            let mut restart_forced = false;
+            let mut last_logged_running_turn_count = None;
             loop {
+                if restart_requested {
```
bolinfest: How hard is it to move the contents of this `if` to a helper function? Or can we just break out of the loop when `restart_requested` is set to true?

codex-rs/app-server/src/lib.rs
```diff
@@ -464,11 +486,69 @@ pub async fn run_main_with_transport(
...
+                    let running_turn_count = *running_turn_count_rx.borrow();
+                    if restart_forced || running_turn_count == 0 {
+                        if restart_forced {
+                            info!(
+                                "received second Ctrl-C; forcing restart with {} running assistant turn(s) and {} connection(s)",
+                                running_turn_count,
+                                connections.len()
```
bolinfest: Note that simple values can be inlined into the format string:

```suggestion
                                "received second Ctrl-C; forcing restart with {running_turn_count} running assistant turn(s) and {} connection(s)",
                                connections.len()
```

codex-rs/app-server/src/lib.rs
```diff
@@ -619,8 +699,11 @@ pub async fn run_main_with_transport(
     let _ = processor_handle.await;
     let _ = outbound_handle.await;
 
+    if let Some(shutdown_token) = websocket_accept_shutdown {
+        shutdown_token.cancel();
+    }
     if let Some(handle) = websocket_accept_handle {
-        handle.abort();
+        let _ = handle.await;
```
bolinfest: So we `await` instead of `abort()` now because we're confident it will exit cleanly?

bolinfest (commented):
I have to come back to this later, but here are my comments thus far.

codex-rs/app-server/tests/suite/v2/connection_handling_websocket.rs
maxj-oai: no - just this specific test is unix only due to how it simulates ctrl-c

codex-rs/app-server/src/lib.rs
maxj-oai: yep! we now expect the websocket acceptor task to exit cleanly after shutdown_token.cancel()

codex-rs/app-server/src/lib.rs
maxj-oai: done

codex-rs/app-server/src/lib.rs
maxj-oai: done

codex-rs/app-server/src/lib.rs
maxj-oai: done

codex-rs/app-server/tests/suite/v2/connection_handling_websocket.rs
owenlin0: It does seem like it'd be cleaner if we move all the ctrl-c testing to a unix-specific file so we don't litter `#[cfg(unix)]` in a bunch of places in this file

owenlin0 (commented):
Once we receive a ctrl-c I believe we'd want to prevent 1) existing clients from triggering new `turn/start` and 2) new clients from connecting?

ideally if they do they get a JSON-RPC error indicating that the server is shutting down.

from codex: "After the first Ctrl-C, this loop keeps processing IncomingMessage and ConnectionOpened, and the websocket acceptor is only canceled once running_turn_count reaches zero. In practice that means existing clients can still start another turn/start, and brand new websocket clients can still connect, which can bump the running-turn count back up and postpone the restart indefinitely. That defeats the core guarantee of a graceful drain: once shutdown is requested, the server should stop taking new work and wait only for already-running turns to finish."

owenlin0 (approved):

PR #12535, opened 2026-02-23
core: customize collab config deprecation warning

edward-bayes:
### Overview
- Change legacy collab=true config warning to a clearer config-error message

### Tests
- Passes tests
- Add unit test coverage for the collab alias deprecation notice strings

### DIFF
--- a/codex-rs/core/src/features.rs
+++ b/codex-rs/core/src/features.rs
@@ -368,13 +368,25 @@ fn legacy_usage_notice(alias: &str, feature: Feature) -> (String, Option<String>
             (summary, Some(web_search_details().to_string()))
         }
         _ => {
-            let summary = format!("`{alias}` is deprecated. Use `[features].{canonical}` instead.");
-            let details = if alias == canonical {
-                None
+            let (summary, details) = if alias == "collab" && feature == Feature::Collab {
+                (
+                    "Your configuration file has an error.".to_string(),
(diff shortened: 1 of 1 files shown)


codex-rs/core/src/features.rs
```diff
@@ -368,13 +368,25 @@ fn legacy_usage_notice(alias: &str, feature: Feature) -> (String, Option<String>
             (summary, Some(web_search_details().to_string()))
         }
         _ => {
-            let summary = format!("`{alias}` is deprecated. Use `[features].{canonical}` instead.");
-            let details = if alias == canonical {
-                None
+            let (summary, details) = if alias == "collab" && feature == Feature::Collab {
```
etraut-openai, line 371: I don't think this should be specific to "collab". I think your wording is an improvement over the existing wording, but let's use it in the general case for all feature flag deprecations rather than adding a code path that's specific to "collab".

Also, this error message is used in the TUI as well as the app server, and the "Your configuration file has an error" will result in a redundant message in the TUI if this change is made. If you want to add a general "Your configuration file has an error" for the app server code path, it should go in a different place.

etraut-openai (changes requested):

codex-rs/core/src/features.rs
jif-oai: Agree with this comment

PR #12536, opened 2026-02-23
core: Update repository root discovery to honor configured project markers

peterdelevoryas:
I don't think repository root discovery has been considering
`project_root_markers` in `$CODEX_HOME/config.toml`. That's normally
fine, but I work in a Sapling repository, so I would like to be able
to configure the project root markers to include `.hg` and `.sl`,
if-possible.

I don't want to change any behavior in this commit though, I *just* want
to fix the consideration of this configuration option. We could expand
the default configuration to other source control systems later.

I tested the whole thing end-to-end before/after with a local Sapling repo:

Note the AGENTS.md discovery from `/status` while starting Codex in a subdirectory of the repo. I added `project_root_markers = [".git", ".sl"]` to my `~/.codex/config.toml` beforehand.

## Before

Screenshot From 2026-02-22 20-00-22

## After

Screenshot From 2026-02-22 20-12-12

### DIFF
--- a/codex-rs/core/src/config_loader/tests.rs
+++ b/codex-rs/core/src/config_loader/tests.rs
@@ -1248,6 +1248,14 @@ async fn project_root_markers_supports_alternate_markers() -> std::io::Result<()
     Ok(())
 }
 
+#[test]
+fn default_project_root_markers_include_git_only() {
+    assert_eq!(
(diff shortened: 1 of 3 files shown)


etraut-openai:

Please open a bug report, and let's discuss there.

PR #12540, opened 2026-02-23
fix: add ellipsis for truncated status indicator

sayan-oai:
#### What

- Add ellipsis truncation of the status indicator, similar to equivalent truncation done in the footer.
- Extract truncation helpers into separate file

https://github.com/user-attachments/assets/a2d5f22f-8adc-456e-8059-97359194c25c

#### Tests
Updated relevant snapshot tests


codex-rs/tui/src/line_truncation.rs
joshka-oai, line 1: This probably doesn't have performance concerns with the amount of data we're likely to throw through it initially, but it may do for larger amounts. It's worth noting that if we do expect this to be called in loops that it should be tuned for perf as it's got a couple of duplicated O(N) things where you might want to speed things up. Not a big deal, but the sort of thing that would be good to leave a note about in the method docs.

joshka-oai (approved):

PR #12549, opened 2026-02-23
fix: show command running in background terminal in details under status indicator

sayan-oai:
#### What
Display in-progress background terminal command in `status.details` (right under header) rather than inline, as it gets cut off currently.

###### Before
image

###### After
image

#### Tests
Added/updated tests


codex-rs/tui/src/bottom_pane/mod.rs
```diff
@@ -1275,6 +1282,8 @@ mod tests {
         pane.update_status(
             "Working".to_string(),
             Some("First detail line\nSecond detail line".to_string()),
+            true,
```
joshka-oai: bool params are a pain because someone unfamiliar with the code always has to read the method to work out what they mean. Two easy approaches to fix:
- two methods with intention revealing name (update_status, update_status_with_capitalize)
- use a small enum that is intention revealing (https://rust-lang.github.io/api-guidelines/type-safety.html#c-custom-type)

Small nit - likely doesn't matter for codex reading this, does for humans.

codex-rs/tui/src/chatwidget/snapshots/codex_tui__chatwidget__tests__unified_exec_wait_status_renders_command_in_single_details_row.snap
```diff
@@ -0,0 +1,11 @@
+---
+source: tui/src/chatwidget/tests.rs
+expression: rendered
+---
+• Waiting for background terminal (0s • esc to …
+  └ cargo test -p codex-core -- --exact…
+
+
```
joshka-oai, line 8: What's the extra line here?

joshka-oai (approved):

codex-rs/tui/src/chatwidget/snapshots/codex_tui__chatwidget__tests__unified_exec_wait_status_renders_command_in_single_details_row.snap
sayan-oai: that's already there in existing snapshot tests (see [here](https://github.com/openai/codex/blob/main/codex-rs/tui/src/bottom_pane/snapshots/codex_tui__bottom_pane__tests__status_only_snapshot.snap)).

codex says it's an explicit spacer inserted by BottomPane::as_renderable when a status/footer is visible and there are no queued messages.

PR #12557, opened 2026-02-23
Hot-reload thread config after batch config writes.

mzeng-openai:
- [x] Hot-reload thread config after batch config writes.


codex-rs/app-server/src/message_processor.rs
```diff
@@ -515,7 +520,12 @@ impl MessageProcessor {
         params: ConfigBatchWriteParams,
     ) {
         match self.config_api.batch_write(params).await {
-            Ok(response) => self.outgoing.send_response(request_id, response).await,
+            Ok(response) => {
+                self.codex_message_processor
+                    .reload_user_config_for_threads()
+                    .await;
+                self.outgoing.send_response(request_id, response).await;
```
shijie-oai, line 527: I am not sure if this is expensive - but for mcp we only let the thread now that it needs to refresh but it only does that when a new turn is submitted.

shijie-oai: Also this would really depends on consumer? A better pattern is probably expose `ReloadUserConfig` as a method via app server and client can call on demand?

mzeng-openai: I think both have pros and cons. My concern is that if we rely on the consumer to make the call, every consumer has to take the hit that their config changes might not be picked up by config writes.
Also it looks like it won't be too expensive to hot-reload user config. Per Codex:


> The write path would only fan out one Op::ReloadUserConfig per loaded thread in [thread_manager.rs (line 279)](app://-/index.html?hostId=local#), and submit() is just an async channel send in [codex_thread.rs (line 59)](app://-/index.html?hostId=local#) and [codex.rs (line 505)](app://-/index.html?hostId=local#). So the config-write RPC would not wait for each thread to finish reloading.
The actual per-thread work is fairly small in [codex.rs (line 2178)](app://-/index.html?hostId=local#): synchronous read_to_string, TOML parse, clone the existing Config, replace the user layer, store it back. No MCP reconnect, no network, no process spawning.

mzeng-openai: And the most important things is we don't write configs that often.

PR #12560, opened 2026-02-23
app-server: Replay pending item requests on `thread/resume`

euroelessar:
Replay pending client requests after `thread/resume` and emit resolved notifications when those requests clear so approval/input UI state stays in sync after reconnects and across subscribed clients.

Affected RPCs:
- `item/commandExecution/requestApproval`
- `item/fileChange/requestApproval`
- `item/tool/requestUserInput`

Motivation:
- Resumed clients need to see pending approval/input requests that were already outstanding before the reconnect.
- Clients also need an explicit signal when a pending request resolves or is cleared so stale UI can be removed on turn start, completion, or interruption.

Implementation notes:
- Use pending client requests from `OutgoingMessageSender` in order to replay them after `thread/resume` attaches the connection, using original request ids.
- Emit `item/commandExecution/approvalResolved`, `item/fileChange/approvalResolved`, and `item/tool/requestUserInputResolved` when pending requests are answered or cleared by lifecycle cleanup.
- Update the app-server protocol schema, generated TypeScript bindings, and README docs for the replay/resolution flow.

High-level test plan:
- Added automated coverage for replaying pending command execution and file change approval requests on `thread/resume`.
- Added automated coverage for resolved notifications in command approval, file change approval, request_user_input, turn start, and turn interrupt flows.
- Verified schema/docs updates in the relevant protocol and app-server tests.

Manual testing:
- Tested reconnect/resume with multiple connections.
- Confirmed state stayed in sync between connections.


codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -837,7 +837,10 @@ server_notification_definitions! {
     PlanDelta => "item/plan/delta" (v2::PlanDeltaNotification),
     CommandExecutionOutputDelta => "item/commandExecution/outputDelta" (v2::CommandExecutionOutputDeltaNotification),
     TerminalInteraction => "item/commandExecution/terminalInteraction" (v2::TerminalInteractionNotification),
+    CommandExecutionApprovalResolved => "item/commandExecution/approvalResolved" (v2::CommandExecutionApprovalResolvedNotification),
     FileChangeOutputDelta => "item/fileChange/outputDelta" (v2::FileChangeOutputDeltaNotification),
+    FileChangeApprovalResolved => "item/fileChange/approvalResolved" (v2::FileChangeApprovalResolvedNotification),
+    ToolRequestUserInputResolved => "item/tool/requestUserInputResolved" (v2::ToolRequestUserInputResolvedNotification),
```
owenlin0: ooc, why do we need these? can't clients resolve things based on receiving `item/completed`?

owenlin0: i.e. server will send:
- `item/started` notification for file change / command execution / etc.
- `item/commandExecution/requestApproval` request
- a client responds
- `item/completed` notification

euroelessar: 1. we don't have `item/started`/`item/completed` events for user input at all, as there is no corresponding item, iiuc
2. in case of file changes, they are synthesized, but not always? e.g. I don't think we never emit `item/completed` in cases like turn interruption and such (which could be a bug); also harness is allowed to send multiple requests for the same item, so there is no one-to-one mapping between requests & items for file changes approvals in particular, it's many-to-one

due to points above it's just easier (from protocol perspective & client-side consumption) and more consistent to have dedicated events

as a side note, it should be fine to merge these events into just single one if you prefer, I've only split them to mirror having multiple independent rpcs

codex-rs/app-server/src/outgoing_message.rs
```diff
@@ -64,6 +64,18 @@ struct PendingCallbackEntry {
     thread_id: Option<ThreadId>,
 }
 
+fn server_request_id(request: &ServerRequest) -> &RequestId {
```
maxj-oai: nit can we move this to ServerRequest::id(&self)

codex-rs/app-server/src/outgoing_message.rs
```diff
@@ -175,14 +213,59 @@ impl OutgoingMessageSender {
...
+    }
+
+    pub(crate) async fn replay_request_to_connections_if_pending(
+        &self,
+        connection_ids: &[ConnectionId],
+        request: ServerRequest,
+        notifications_before_request: &[ServerNotification],
+    ) -> bool {
+        // Hold the callback map lock until the replay is queued so an already-resolved
+        // request cannot be replayed to a resumed client.
```
maxj-oai: stresses me out - prefer to make this not needed for correctness, or use try_send
I don't fully understand what race this prevents - even if we send an already-resolved request event that should be ok right? since the client will observe the resolution as well right after (due to ordering guarantee provided by the thread listener task)

codex-rs/app-server/src/thread_state.rs
```diff
@@ -29,13 +32,17 @@ pub(crate) struct PendingThreadResumeRequest {
 }
 
 pub(crate) enum ThreadListenerCommand {
-    SendThreadResumeResponse(PendingThreadResumeRequest),
+    SendThreadResumeResponse(Box<PendingThreadResumeRequest>),
+    RemoveRequest {
```
maxj-oai: nit: RemoveServerRequest to disambiguate server->client reqs vs client->server

codex-rs/app-server/src/outgoing_message.rs
euroelessar: yeah, I agree, we already serialize resolution properly so should be fine, simplified

codex-rs/app-server-protocol/src/protocol/common.rs
owenlin0: ah ok, we might want to add an `item` for representing the user input tool call. I had to do that recently for dynamic tool calls: https://github.com/openai/codex/pull/12732

but you're right, we're going to move to a world soon where there are multiple approvals for a CommandExecution so we can't rely on `item/completed` for that either. proceed :)

owenlin0: naming nit: since our requests are like `item/commandExecution/requestApproval` should we name these notifications `item/commandExecution/requestApproval/resolved`?

owenlin0: basically just appending `/resolved` to the corresponding server method name

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -151,11 +208,13 @@ pub(crate) async fn apply_bespoke_event_handling(
     } = event;
     match msg {
         EventMsg::TurnStarted(_) => {
+            abort_pending_client_requests(conversation_id, &thread_state, &outgoing).await;
```
owenlin0: interesting, is this just out of abundance of caution that we're doing this? why is this necessary here?

maxj-oai (approved):
awesome, ty!

codex-rs/app-server/src/bespoke_event_handling.rs
owenlin0: btw there's also a `cancel_requests_for_thread` method on `OutgoingMessageSender`, can we reuse that?

codex-rs/app-server/src/thread_state.rs
```diff
@@ -44,6 +51,7 @@ pub(crate) struct TurnSummary {
 pub(crate) struct ThreadState {
     pub(crate) pending_interrupts: PendingInterruptQueue,
     pub(crate) pending_rollbacks: Option<ConnectionRequestId>,
+    pending_client_requests: Vec<ServerRequest>,
```
owenlin0: Do we need to track this here? we already have tracking on OutgoingMessageSender `request_id_to_callback`

codex-rs/app-server/src/bespoke_event_handling.rs
euroelessar: done! had to update it a bit to preserve behavior on turn ends to avoid additional unintended side effects

codex-rs/app-server/src/thread_state.rs
euroelessar: reused!

codex-rs/app-server/src/outgoing_message.rs
```diff
@@ -155,12 +186,15 @@ impl OutgoingMessageSender {
 
     async fn send_request_with_id_to_connections(
         &self,
-        connection_ids: &[ConnectionId],
+        connection_ids: Option<&[ConnectionId]>,
         request: ServerRequestPayload,
         thread_id: Option<ThreadId>,
+        tracked_request: bool,
```
owenlin0: what makes a request "tracked" or not?

codex-rs/app-server/src/outgoing_message.rs
```diff
@@ -90,6 +93,23 @@ impl ThreadScopedOutgoingMessageSender {
             .await
     }
 
+    pub(crate) async fn send_request_with_server_request(
```
owenlin0: what's the difference between this and `send_request`?

codex-rs/app-server/src/outgoing_message.rs
```diff
@@ -248,23 +300,83 @@ impl OutgoingMessageSender {
...
+    pub(crate) async fn cancel_requests_for_thread(
+        &self,
+        thread_id: ThreadId,
+        error: Option<JSONRPCErrorError>,
+    ) {
+        self.cancel_requests_for_thread_matching(thread_id, error, false)
+            .await
+    }
+
+    pub(crate) async fn cancel_tracked_requests_for_thread(
```
owenlin0: i don't quite understand why we need to make a tracked vs. non-tracked distinction, can we unify?

codex-rs/app-server/src/outgoing_message.rs
```diff
@@ -399,6 +511,19 @@ impl OutgoingMessageSender {
...
+fn compare_request_ids(left: &RequestId, right: &RequestId) -> CmpOrdering {
+    match (left, right) {
+        (RequestId::Integer(left), RequestId::Integer(right)) => left.cmp(right),
+        (RequestId::String(left), RequestId::String(right)) => left.cmp(right),
+        (RequestId::Integer(_), RequestId::String(_)) => CmpOrdering::Less,
+        (RequestId::String(_), RequestId::Integer(_)) => CmpOrdering::Greater,
+    }
+}
+
+fn sort_requests_by_id(requests: &mut [ServerRequest]) {
```
owenlin0: this is only used once, let's just inline this where it's used

codex-rs/app-server/src/outgoing_message.rs
```diff
@@ -399,6 +511,19 @@ impl OutgoingMessageSender {
     }
 }
 
+fn compare_request_ids(left: &RequestId, right: &RequestId) -> CmpOrdering {
```
owenlin0: i think you can do `impl Ord for RequestId` instead of a top-level function

codex-rs/app-server/tests/suite/v2/request_user_input.rs
```diff
@@ -96,17 +99,46 @@ async fn request_user_input_round_trip() -> Result<()> {
...
+                    notification
+                        .params
+                        .clone()
+                        .expect("item/tool/requestUserInputResolved params"),
+                )?;
+                assert_eq!(resolved.thread_id, thread.id);
+                assert_eq!(resolved.request_id, resolved_request_id);
+                saw_resolved = true;
+            }
+            "codex/event/task_complete" => {
```
owenlin0: nit: can remove this event, we plan to deprecate all `codex/event/*` notifications

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -6827,7 +6953,27 @@ async fn handle_pending_thread_resume_request(
         reasoning_effort,
     };
     outgoing.send_response(request_id, response).await;
-    thread_state.lock().await.add_connection(connection_id);
+
+    let pending_requests = outgoing.pending_requests_for_thread(conversation_id).await;
+    let file_change_started = {
```
owenlin0: interesting, why is there special handling for file changes here?

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -6827,7 +6953,27 @@ async fn handle_pending_thread_resume_request(
...
+        let mut state = thread_state.lock().await;
+        state.add_connection(connection_id);
+        state.turn_summary.file_change_started.clone()
+    };
+
+    for request in pending_requests {
+        let notifications_before_request =
+            pending_client_request_resume_notifications(&file_change_started, &request);
```
owenlin0: what's the goal of this step? is this needed?

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -6739,6 +6747,124 @@ impl CodexMessageProcessor {
...
+                    thread_id,
+                    request_id,
+                },
+            )
+        }
+    };
+    outgoing.send_server_notification(notification).await;
+}
+
+pub(crate) async fn abort_pending_client_requests(outgoing: &ThreadScopedOutgoingMessageSender) {
```
owenlin0: can this be a method on `ThreadScopedOutgoingMessageSender`?

owenlin0 (commented):
trying my best to review but it's getting quite hard to follow :(

codex-rs/app-server/src/codex_message_processor.rs
owenlin0: I guess it's to handle this part from the PR description:
> resend prerequisite notifications like item/started for pending file change approvals.

why do we need to resend `item/started`? 🤔 

the `thread/resume` response should already have a FileChange item in `pending` on `thread.turns[-1].items[-1]`, right?

codex-rs/app-server/src/outgoing_message.rs
euroelessar: it's essentially a workaround to avoid closing rpcs for dynamic tool & v1, as I wanted to maintain status quo in their behavior
alternatively could check for the type (but that feels more hackier), or spend more type drilling into them to make sure they are not broken by the change

codex-rs/app-server/src/codex_message_processor.rs
euroelessar: `item/started` for file changes is synthesized on `EventMsg::ApplyPatchApprovalRequest` event, among others
though thinking more about it, we probably should move this particular piece of logic into `ThreadHistoryBuilder`, so that item is properly emitted by other piece of thread resumption logic (which currently doesn't handle this event)
will get back to you if that works

owenlin0: cool cool, would be great to handle in ThreadHistoryBuilder if we aren't

codex-rs/app-server/src/outgoing_message.rs
euroelessar: bit the bullet and gone through all the remaining server->client requests to add proper handling of abortion into them, so no need for tracked requests anymore

codex-rs/app-server/src/outgoing_message.rs
euroelessar: this one was returning a request id as well, merged them into one & updated all callsites for simplicity

codex-rs/app-server/src/codex_message_processor.rs
euroelessar: yep, updated

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -151,11 +189,13 @@ pub(crate) async fn apply_bespoke_event_handling(
     } = event;
     match msg {
         EventMsg::TurnStarted(_) => {
+            outgoing.abort_pending_client_requests().await;
```
owenlin0: nit: maybe worth a comment on why we need this, still not super clear to me 🤷

codex-rs/app-server/src/thread_state.rs
```diff
@@ -29,13 +31,24 @@ pub(crate) struct PendingThreadResumeRequest {
 }
 
 pub(crate) enum ThreadListenerCommand {
```
owenlin0, line 33: prob worth a comment on what this (and the variants) are used for

codex-rs/app-server/src/thread_state.rs
```diff
@@ -29,13 +31,24 @@ pub(crate) struct PendingThreadResumeRequest {
...
+    CommandExecutionRequestApproval,
+    FileChangeRequestApproval,
+    ToolRequestUserInput,
 }
 
 /// Per-conversation accumulation of the latest states e.g. error message while a turn runs.
 #[derive(Default, Clone)]
 pub(crate) struct TurnSummary {
-    pub(crate) file_change_started: HashSet<String>,
+    pub(crate) file_change_started: HashMap<String, Vec<FileUpdateChange>>,
```
owenlin0: curious, why did we need to change it to a hashmap here?

codex-rs/app-server/src/outgoing_message.rs
```diff
@@ -99,6 +100,20 @@ impl ThreadScopedOutgoingMessageSender {
...
 
+    pub(crate) async fn abort_pending_client_requests(&self) {
+        self.outgoing
+            .cancel_requests_for_thread(
+                self.thread_id,
+                Some(JSONRPCErrorError {
+                    code: INTERNAL_ERROR_CODE,
+                    message: "client request resolved because the turn state was changed"
+                        .to_string(),
+                    data: Some(serde_json::json!({ "reason": "turnTransition" })),
```
owenlin0: can import the const you defined: `TURN_TRANSITION_PENDING_REQUEST_ERROR_REASON`

owenlin0 (approved):
nice, much cleaner!!

Issue #11008, opened 2026-02-07
Codex locks-up / deadlocks if you steer just as it's printing final message


guidedways:
### What version of Codex CLI is running?

0.99.0-alpha.9

### What subscription do you have?

Pro

### Which model were you using?

gpt-5.3-codex-xhigh

### What platform is your computer?

macOS

### What terminal emulator and version are you using (if applicable)?

iTerm2

### What issue are you seeing?

If codex is in the process of printing out it's final `summary` and you at that very moment hit `return` on a long prompt, it just sits there in limbo - you cannot quit, it won't respond, it allows you to type more and enter more prompts but it does nothing - forever lost. You need to then `/quit` and `codex resume ...` to get it back to where it was.

### What steps can reproduce the bug?

* Make sure Steer is on
* Give it some work
* When it's giving you it's final work `summary`... make sure you've entered a prompt and then immediately hit `return` while it's in the middle of printing it (and hasn't finished its final summary, still streaming
* Bingo - locked up. Codex can no longer recover.

Reproducible 100% of the times.

etraut-openai:
@guidedways, we haven't had a chance to look at this yet. You're the only one who has reported it so far. Since you seem to be able to repro the problem, if you'd like to dig into it and try to find the root cause, that would be helpful. I'm not asking for a PR, but a concise description of the problem — a prompt that we can use to ask Codex to fix it.

PR #12569, opened 2026-02-23
fix(tui): queue steer Enter while final answer is still streaming to prevent dead state

guidedways:
## Summary
This fixes a TUI race (https://github.com/openai/codex/issues/11008) where pressing Enter with Steer enabled while the assistant is still streaming the final answer could put Codex into a non-recoverable “running” state (no further prompts handled until exiting and resuming).

## Root Cause
In steer mode, `InputResult::Submitted` could submit immediately even while a final-answer stream was active. That immediate submission races with turn completion and can strand turn state.

## Fix
When handling `InputResult::Submitted`, we now queue instead of immediate-submit if a final-answer stream is active (`stream_controller.is_some()`).

This keeps behavior deterministic:
- Prompt is preserved in the queue.
- `on_task_complete()` drains queued input through `maybe_send_next_queued_input()`.
- Follow-up prompts continue in FIFO order after completion.

## Why this resolves the “dead mode”
The problematic timing window is now converted into queueing, so prompts entered during final streaming are not lost and are processed after the current output ends. The model continues handling prompts normally without requiring `/quit` + `resume`.

## Tests
Added regression coverage in `tui/src/chatwidget/tests.rs`:

- `steer_enter_queues_while_final_answer_stream_is_active`
- `steer_enter_during_final_stream_preserves_follow_up_prompts_in_order`

Both fail on old behavior and pass with this fix.


guidedways:
@codex review

charley-oai (approved):

guidedways:
Some other issues may also be related to this PR: https://github.com/openai/codex/issues/11047

Issue #12450, opened 2026-02-21
Support for Tree-based Conversation Management (Chat Tree / Branch)


Reekin:
### What variant of Codex are you using?

App,CLI

### What feature would you like to see?

During conversations with the agent, it is very common to want to roll back the context to a specific previous node.

For example: You might go through several rounds of chatting to finalize the overall development direction for a feature. Then, while developing a specific part, you encounter unexpected behavior and spend multiple rounds providing detailed feedback and fixing bugs. Once that feature is done, you want the chat history to revert to the point before diving into all those bug-fixing details, so you can start working on another related requirement (which still requires the high-level context from the initial conversation).

Without a convenient way to manage context, we are currently forced into one of two bad options:

1. Continue developing in the original session, which pollutes the context with a massive amount of irrelevant details that subsequent development doesn't care about.
2. Ask the agent to summarize the first half of the conversation into a file, and then open a new session for the agent to read it. This is highly tedious and very prone to context omission/loss.
Furthermore, considering that in the Codex app, a single session might be bound to a specific worktree (which I personally think is a debatable design choice—I actually believe worktrees should be bound to sub-agents instead), switching sessions introduces a lot of friction. On the other hand, staying in the original session with accumulated "garbage context" leads to frequent context compactions, which severely degrades the agent's execution quality.

To address this, I have designed a solution for convenient context management via a `ChatTree` and developed a PoC demo.

Below are the PRD and a demonstration video of how it operates:

[PRD](https://github.com/Reekin/codex/blob/codex/chat-tree/chat-tree-PRD.md)
[Repo](https://github.com/Reekin/codex/tree/codex/chat-tree)

https://github.com/user-attachments/assets/fdc78c1a-b95a-4405-8158-56aa8e139dbd

### Additional information

I'd love to hear your thoughts on this pain point and the proposed solution.

Reekin:
> Potential duplicates detected. Please review them and close your issue if it is a duplicate.

While this proposal shares some conceptual overlap with standard `rewind` or `edit history` features, there is a fundamental distinction:

My goal is not simply to "modify" past prompts or reset the conversation. Instead, I am proposing a non-linear, multi-branching tree structure.

In a standard "edit" workflow, you usually lose the subsequent messages or create a linear replacement. In contrast, with a Conversation Tree:
1. Node Switching: You can jump to any historical node at any time.
[...]

PR #12610, opened 2026-02-23
tui: add /fork turn picker UI

yvolovich-cyber:
## Problem

The backend can already fork a conversation from an arbitrary user turn, but the TUI still treated `/fork` as "fork from the latest turn only". That left the feature inaccessible from the main CLI flow even though the protocol support already existed.

This PR makes the branch point explicit in the TUI before the current thread is forked.

image

## Mental model

Running `/fork` now opens an alternate-screen picker built from the current conversation's rollout log. The picker reconstructs the active branch by replaying rollout items, pairing each surviving user turn with the assistant response that follows it, and dropping turns that were later removed by rollback events.

The list shows a compact preview of both sides of each turn: the user request on the first line and the model response on the second line. The lower preview pane shows the selected turn in more detail, and selection can be moved with arrows, `j`/`k`, or digit shortcuts.

When the user confirms a selection, the picker converts that visible row into the `nth_user_message` value expected by the existing fork RPC. Selecting the newest surviving turn intentionally preserves the existing "fork from head" behavior by using the same sentinel value the RPC already understands.

## Non-goals

This PR does not change server-side fork semantics, add new fork RPC surface area, or expose historical rolled-back branches as selectable targets. It also does not add filtering, search, or per-turn timestamps to the picker.

## Tradeoffs

The picker renders normalized previews instead of a lossless transcript. That keeps the UI compact and readable, but it means the preview is intentionally an approximation of the original turn content.

This PR also keeps the existing `usize::MAX` sentinel contract for "fork from the latest turn". That avoids protocol churn, but it means the TUI still needs to understand a low-level RPC convention instead of calling a more explicit API.

## Architecture

The TUI now has a dedicated `/fork` picker module that does two jobs:

1. It derives a presentation-only list of forkable turns from the current rollout log.
2. It renders an alternate-screen selection UI and returns the exact fork parameter expected by the existing RPC.

The derivation step is rollback-aware. User messages create new entries, assistant messages attach preview text to the newest entry, and rollback events truncate the derived list so the UI matches the active branch of the conversation rather than the full event log.

At the app layer, the `/fork` handler now loads those entries, handles empty or unreadable rollout state, launches the picker, and passes the returned `nth_user_message` value into the existing `fork_thread` flow.

## Observability

The existing `codex.thread.fork` counter remains the primary fork metric. Failure cases continue to surface as inline TUI error messages when the rollout file cannot be read or when the fork RPC fails.

Cancellation remains intentionally quiet: backing out of the picker redraws the main UI and does not create a new fork or emit additional user-visible history.

## Tests

This PR includes `codex-tui` coverage for:

- rollback-aware extraction of effective forkable turns
- the `nth_user_message` mapping used to preserve fork semantics
- snapshot coverage for the `/fork` picker UI

Validated locally with `cargo test -p codex-tui`.

### DIFF
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -1840,6 +1840,9 @@ impl App {
                         );
                         self.chat_widget
                             .add_plain_history_lines(vec!["/fork".magenta().into()]);
+                        // The picker returns the exact `nth_user_message` value expected by the
+                        // fork RPC, including the existing `usize::MAX` sentinel for "fork from
+                        // the latest surviving turn".
                         match self
                             .server
                             .fork_thread(nth_user_message, self.config.clone(), path.clone(), false)
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -1800,20 +1800,49 @@ impl App {
             AppEvent::ForkCurrentSession => {
                 self.otel_manager
                     .counter("codex.thread.fork", 1, &[("source", "slash_command")]);
-                let summary = session_summary(
-                    self.chat_widget.token_usage(),
-                    self.chat_widget.thread_id(),
-                    self.chat_widget.thread_name(),
-                );
-                self.chat_widget
-                    .add_plain_history_lines(vec!["/fork".magenta().into()]);
                 if let Some(path) = self.chat_widget.rollout_path() {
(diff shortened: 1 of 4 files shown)


ignatremizov:
Wouldn't a fast branch/tree-like UI be better? See https://github.com/openai/codex/issues/12450

PR #12612, opened 2026-02-23
Unify rollout reconstruction with resume/fork TurnContext hydration

charley-oai:
## Summary

This PR unifies rollout history reconstruction and resume/fork metadata hydration under a single `Session::reconstruct_history_from_rollout` implementation.

The key change from main is that replay metadata now comes from the same reconstruction pass that rebuilds model-visible history, instead of doing a second bespoke rollout scan to recover `previous_model` / `reference_context_item`.

### Unified reconstruction output

`reconstruct_history_from_rollout` now returns a single `RolloutReconstruction` bundle containing:

- rebuilt `history`
- `previous_model`
- `reference_context_item`

Resume and fork both consume that shared output directly.

### Reverse replay core

The reconstruction logic moved into `codex-rs/core/src/codex/rollout_reconstruction.rs` and now scans rollout items newest-to-oldest.

That reverse pass:

- derives `previous_model`
- derives whether `reference_context_item` is preserved or cleared
- stops early once it has both resume metadata and a surviving `replacement_history` checkpoint

History materialization is still bridged eagerly for now by replaying only the surviving suffix forward, which keeps the history result stable while moving the control flow toward the future lazy reverse loader design.

### Removed bespoke context lookup

This deletes `last_rollout_regular_turn_context_lookup` and its separate compaction-aware scan.

The previous model / baseline metadata is now computed from the same replay state that rebuilds history, so resume/fork cannot drift from the reconstructed transcript view.

### `TurnContextItem` persistence contract

`TurnContextItem` is now treated as the replay source of truth for durable model-visible baselines.

This PR keeps the following contract explicit:

- persist `TurnContextItem` for the first real user turn so resume can recover `previous_model`
- persist it for later turns that emit model-visible context updates
- if mid-turn compaction reinjects full initial context into replacement history, persist a fresh `TurnContextItem` after `Compacted` so resume/fork can re-establish the baseline from the rewritten history
- do not treat manual compaction or pre-sampling compaction as creating a new durable baseline on their own

## Behavior Preserved

- rollback replay stays aligned with `drop_last_n_user_turns`
- rollback skips only user turns
- incomplete active user turns are dropped before older finalized turns when rollback applies
- unmatched aborts do not consume the current active turn
- missing abort IDs still conservatively clear stale compaction state
- compaction clears `reference_context_item` until a later `TurnContextItem` re-establishes it
- `previous_model` still comes from the newest surviving user turn that established one

## Tests

Targeted validation run for the current branch shape:

- `cd codex-rs && cargo test -p codex-core --lib codex::rollout_reconstruction_tests -- --nocapture`
- `cd codex-rs && just fmt`

The branch also extracts the rollout reconstruction tests into `codex-rs/core/src/codex/rollout_reconstruction_tests.rs` so this logic has a dedicated home instead of living inline in `codex.rs`.


charley-oai:
@codex review this

codex-rs/core/src/codex.rs
```diff
@@ -2635,6 +2635,15 @@ impl Session {
             }
         }
 
+        if let Some(active_turn) = active_turn.take()
```
charley-oai: Is this trailing incomplete turn case even worth defending against? I guess this can happen if the user kills the process and it doesn't have a chance to complete the turn?

charley-oai:
@codex review this

charley-oai:
@codex review this

codex-rs/core/src/codex.rs
```diff
@@ -2594,8 +2446,86 @@ impl Session {
         &self,
         turn_context: &TurnContext,
         rollout_items: &[RolloutItem],
-    ) -> Vec<ResponseItem> {
+    ) -> RolloutReconstruction {
```
pakrym-oai: any chance we can move this out of codex.rs. Adding 1k more lines to it is rough.

charley-oai: Mostly tests but true

codex-rs/core/src/codex.rs
```diff
@@ -2616,14 +2546,199 @@ impl Session {
...
+                        } else {
+                            active_turn.has_midturn_compaction = true;
+                        }
+                    } else {
+                        replayed_segments.push(RolloutReplayMetaSegment::CompactionOutsideTurn);
+                    }
                 }
                 RolloutItem::EventMsg(EventMsg::ThreadRolledBack(rollback)) => {
                     history.drop_last_n_user_turns(rollback.num_turns);
+                    let mut turns_to_drop =
```
pakrym-oai: a bit strange to run two version of this logic in parallel, one to drop user messages, another to drop turns

codex-rs/core/src/codex/rollout_reconstruction.rs
```diff
@@ -0,0 +1,309 @@
...
+        }
+
+        let mut history = ContextManager::new();
+        let mut saw_turn_lifecycle_event = false;
+        let mut active_turn: Option<ActiveRolloutTurn> = None;
+        let mut replayed_segments = Vec::new();
+        let push_replayed_turn = |replayed_segments: &mut Vec<RolloutReplayMetaSegment>,
+                                  active_turn: ActiveRolloutTurn| {
+            replayed_segments.push(RolloutReplayMetaSegment::Turn(Box::new(
+                ReplayedRolloutTurn {
```
pakrym-oai: we are copying the struct into almost the same struct `ActiveRolloutTurn` -- >`ReplayedRolloutTurn`.

codex-rs/core/src/codex.rs
charley-oai: I think we can't unify them because history only contains model-visible data whereas our stack of user turns contains metadata needed to fetch `previous_model` and `reference_context_item`

codex-rs/core/src/codex/rollout_reconstruction.rs
charley-oai: The conversion serves to eliminate dead/unused fields
- dropping turn_id, which is only needed while matching lifecycle events
- dropping saw_user_message, which is only needed to decide whether the span counts as a user turn at all

charley-oai:
@codex review this

charley-oai:
@codex review this

codex-rs/core/src/codex/rollout_reconstruction.rs
```diff
@@ -0,0 +1,278 @@
...
+                        active_segment.reference_context_item,
+                        TurnReferenceContextItem::NeverSet
+                    ) {
+                        active_segment.reference_context_item = TurnReferenceContextItem::Cleared;
+                    }
+                    if active_segment.base_replacement_history.is_none()
+                        && let Some(replacement_history) = &compacted.replacement_history
+                    {
+                        active_segment.base_replacement_history = Some(replacement_history);
+                        rollout_suffix = &rollout_items[index + 1..];
```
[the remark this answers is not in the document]
charley-oai: Unnecessary complexity for a case that could never happen in practice

codex-rs/core/src/compact.rs
```diff
@@ -224,15 +224,23 @@ async fn run_compact_task_inner(
...
     let rollout_item = RolloutItem::Compacted(CompactedItem {
         message: summary_text.clone(),
         replacement_history: Some(new_history),
     });
-    sess.persist_rollout_items(&[rollout_item]).await;
+    let rollout_items = if let Some(turn_context_item) = reference_context_item {
+        // Mid-turn compaction re-injected initial context into the replacement history, so
+        // persist a fresh `TurnContextItem` after `Compacted` to re-establish the baseline for
+        // resume/fork replay.
+        vec![rollout_item, RolloutItem::TurnContext(turn_context_item)]
```
pakrym-oai: super nit: can you  have a separate additional sess.persist_rollout_items call? 

Might make logic a bit easier to read

    sess.persist_rollout_items(&rollout_items).await;
 if if let Some(turn_context_item) = reference_context_item
    sess.persist_rollout_items([reference_context_item]).await;

codex-rs/core/src/compact_remote.rs
```diff
@@ -159,16 +159,26 @@ async fn run_remote_compact_task_inner_impl(
...
         .await;
     sess.recompute_token_usage(turn_context).await;
 
     let compacted_item = CompactedItem {
         message: String::new(),
         replacement_history: Some(new_history),
     };
-    sess.persist_rollout_items(&[RolloutItem::Compacted(compacted_item)])
-        .await;
+    let rollout_items = if let Some(turn_context_item) = reference_context_item {
```
pakrym-oai: I almost wish this was part of replace history somehow

codex-rs/core/src/codex/rollout_reconstruction.rs
```diff
@@ -0,0 +1,277 @@
...
+                RolloutItem::ResponseItem(response_item) => {
+                    history.record_items(
+                        std::iter::once(response_item),
+                        turn_context.truncation_policy,
+                    );
+                }
+                RolloutItem::Compacted(compacted) => {
+                    if let Some(replacement_history) = &compacted.replacement_history {
+                        history.replace(replacement_history.clone());
+                    } else {
```
pakrym-oai, line 248: I think we never produce Compacted without replacement_history anymore. Is it still worth keeping this condition?

pakrym-oai: I'd love to drop initial_context and disconnect this method from `Session.`

codex-rs/core/src/codex/rollout_reconstruction.rs
```diff
@@ -0,0 +1,277 @@
...
+
+        let initial_context = self.build_initial_context(turn_context, None).await;
+        let mut history = ContextManager::new();
+        if let Some(base_replacement_history) = base_replacement_history {
+            history.replace(base_replacement_history.to_vec());
+        }
+        // Materialize exact history semantics from the replay-derived suffix. The eventual lazy
+        // design should keep this same replay shape, but drive it from a resumable reverse source
+        // instead of an eagerly loaded `&[RolloutItem]`.
+        for item in rollout_suffix {
```
pakrym-oai, line 235: Considering we have a separate loop anyway (ideally we won't), can we drop rollout_suffix and add `break` to RolloutItem::Compacted branch?

pakrym-oai: Ah, is it because the history is append to the end only?

codex-rs/core/src/codex/rollout_reconstruction.rs
```diff
@@ -0,0 +1,277 @@
...
+                &mut base_replacement_history,
+                &mut previous_model,
+                &mut reference_context_item,
+                &mut pending_rollback_turns,
+            );
+        }
+
+        let initial_context = self.build_initial_context(turn_context, None).await;
+        let mut history = ContextManager::new();
+        if let Some(base_replacement_history) = base_replacement_history {
```
pakrym-oai, line 229: Why do we have `base_replacement_history`? Wouldn't the loop below replace what's needed?

codex-rs/core/src/codex/rollout_reconstruction.rs
```diff
@@ -0,0 +1,277 @@
...
+    active_segment: ActiveReplaySegment<'a>,
+    base_replacement_history: &mut Option<&'a [ResponseItem]>,
+    previous_model: &mut Option<String>,
+    reference_context_item: &mut TurnReferenceContextItem,
+    pending_rollback_turns: &mut usize,
+) {
+    // Thread rollback always targets the newest surviving user turns, so consume that
+    // skip budget before letting this segment contribute metadata or a compaction base.
+    if *pending_rollback_turns > 0 {
+        if active_segment.counts_as_user_turn {
```
pakrym-oai, line 53: the drop_last_n_user_turns is base on user messages, this logic is based on turns. Do they align?

pakrym-oai (approved):
Is rollback behavior aligned?
Can we simplify by removing support for non-replacement_history compaction? 
If we are keeping two loops, can we make them independent and bail compact item when loading ResponseItems - that will make me sleep a bit better at night.

codex-rs/core/src/codex/rollout_reconstruction.rs
charley-oai: That's needed to keep track of whether we can stop consuming backwards, and then at the beginning of the forward pass we use it to start off the `history` we build

codex-rs/core/src/codex/rollout_reconstruction.rs
charley-oai: I think this logic is based on user messages too, `counts_as_user_turn` is carefully designed to align with `drop_last_n_user_turns` logic (and I added a bunch of tests to ensure they align)

codex-rs/core/src/codex/rollout_reconstruction.rs
charley-oai: I think it's hard to do the second as part of the first because of None replacement_history compacts which want the whole history so far (when going left to right)

codex-rs/core/src/codex/rollout_reconstruction.rs
charley-oai: Kept it for backwards compat (maybe later we can fully remove), but I simplified a bit to not pass turn_context into the None replacement_history compact anymore

codex-rs/core/src/compact_remote.rs
charley-oai: Consolidated into `replace_compacted_history`

codex-rs/core/src/codex/rollout_reconstruction.rs
```diff
@@ -0,0 +1,294 @@
...
+                        active_segment.reference_context_item,
+                        TurnReferenceContextItem::NeverSet
+                    ) {
+                        active_segment.reference_context_item = TurnReferenceContextItem::Cleared;
+                    }
+                    if active_segment.base_replacement_history.is_none()
+                        && let Some(replacement_history) = &compacted.replacement_history
+                    {
+                        active_segment.base_replacement_history = Some(replacement_history);
+                        rollout_suffix = &rollout_items[index + 1..];
```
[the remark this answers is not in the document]
charley-oai: Can't rollback in same turn as compaction

PR #12623, opened 2026-02-23
ignore PRESETS models when custom model_catalog_json specified

sayan-oai:
Prevent extra models (defined in `PRESETS`) from appearing in available models when the user explicitly overrides built in models list with a custom one.

These presets are clobbered by remote models in the existing merge process, so they are commonly overridden anyways.

Added a unit test.


aibrahim-oai (approved):
add e2e integration test

PR #12630, opened 2026-02-24
fix(core): use runtime config for review reasoning effort

hdcodedev:
### What
Fix `/review` so it uses the active runtime turn context when selecting model config, especially reasoning effort, instead of relying on potentially stale startup config.

### Why
Fixes #12495. After switching model/effort in-session (for example to `gpt-5.1-codex-mini` + `high`), `/review` could still use startup effort (for example `xhigh`) and fail with unsupported-effort errors.

### How
Updated review flow in `core/src/codex.rs`:
* Review now derives config from `parent_turn_context.config` instead of startup config.
* Added a shared helper, `select_reasoning_effort_for_model(...)`, and reused it in both model switching and review.
* Review now sets `per_turn_config.model_reasoning_effort` from that helper, preserving current runtime effort when supported and falling back when not.

### Testing
Added/updated coverage in `core/tests/suite/review.rs`:
* `review_uses_current_reasoning_effort_after_model_switch`
* `review_falls_back_reasoning_effort_when_current_effort_is_unsupported`

Validation run:
* `cargo clippy -p codex-core --tests`
* `cargo test -p codex-core`

### Before fix
https://github.com/user-attachments/assets/80b19bbb-ab4f-4b41-bb97-55224ca2205a

### After fix
TODO

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -575,6 +575,30 @@ pub(crate) struct TurnContext {
     pub(crate) dynamic_tools: Vec<DynamicToolSpec>,
     pub(crate) turn_metadata_state: Arc<TurnMetadataState>,
 }
+
+fn select_reasoning_effort_for_model(
+    current_reasoning_effort: Option<ReasoningEffortConfig>,
+    model_info: &ModelInfo,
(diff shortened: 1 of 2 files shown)


etraut-openai:

PR #12633, opened 2026-02-24
Add skill approval event/response

pakrym-oai:
Set the stage for  skill-level permission approval in addition to command-level.

Behind a feature flag.


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3431,6 +3431,29 @@ pub struct FileChangeRequestApprovalResponse {
     pub decision: FileChangeApprovalDecision,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
```
bolinfest: I would favor annotating with `experimental` since we are probably not ready for partners to treat this as a stable API? Note this should reduce the number of changes to app-server-protocol/scheme.

codex-rs/app-server/tests/suite/v2/skill_approval.rs
```diff
@@ -0,0 +1,115 @@
...
+    assert_eq!(params.skill_name, "test-skill");
+
+    mcp.send_response(request_id, serde_json::json!({ "decision": "approve" }))
+        .await?;
+
+    timeout(
+        DEFAULT_READ_TIMEOUT,
+        mcp.read_stream_until_notification_message("codex/event/task_complete"),
+    )
+    .await??;
```
bolinfest: Is it necessary to wait for both of these notifications?

Hmm, I see a `wait_for_turn_completed()` function, but it's not a general utility?

codex-rs/app-server/tests/suite/v2/skill_approval.rs
```diff
@@ -0,0 +1,115 @@
...
+
+model_provider = "mock_provider"
+
+[features]
+skill_approval = true
+
+[model_providers.mock_provider]
+name = "Mock provider for test"
+base_url = "{server_uri}/v1"
+wire_api = "responses"
```
bolinfest: We can stop doing this now, right?

bolinfest (approved):

codex-rs/app-server/tests/suite/v2/skill_approval.rs
pakrym-oai: Yeah, but we also need to get rid of having to copy this snippet around.

PR #12638, opened 2026-02-24
refactor: decouple shell-escalation from codex-core

bolinfest:
## Why

After removing `exec-server`, the next step is to wire a new shell tool to `codex-rs/shell-escalation` directly.

That is blocked while `codex-shell-escalation` depends on `codex-core`, because the new integration would require `codex-core` to depend on `codex-shell-escalation` and create a dependency cycle.

This change ports the reusable pieces from the earlier prep work, but drops the old compatibility shim because `exec-server`/MCP support is already gone.

### Decouple `shell-escalation` from `codex-core`

- Introduce a crate-local `SandboxState` in `shell-escalation`
- Introduce a `ShellCommandExecutor` trait so callers provide process execution/sandbox integration
- Update `EscalateServer::exec(...)` and `run_escalate_server(...)` to use the injected executor
- Remove the direct `codex_core::exec::process_exec_tool_call(...)` call from `shell-escalation`
- Remove the `codex-core` dependency from `codex-shell-escalation`

### Restore reusable policy adapter exports

- Re-enable `unix::core_shell_escalation`
- Export `ShellActionProvider` and `ShellPolicyFactory` from `shell-escalation`
- Keep the crate root API simple (no `legacy_api` compatibility layer)

### Port socket fixes from the earlier prep commit

- Use `socket2::Socket::pair_raw(...)` for AF_UNIX socketpairs and restore `CLOEXEC` explicitly on both endpoints
- Keep `CLOEXEC` cleared only on the single datagram client FD that is intentionally passed across `exec`
- Clean up `tokio::AsyncFd::try_io(...)` error handling in the socket helpers

## Verification

- `cargo shear`
- `cargo clippy -p codex-shell-escalation --tests`


codex-rs/shell-escalation/src/unix/escalate_server.rs
```diff
@@ -142,6 +148,7 @@ pub trait EscalationPolicyFactory {
     fn create_policy(&self, policy: Arc<RwLock<Policy>>, stopwatch: Stopwatch) -> Self::Policy;
 }
 
+#[allow(clippy::too_many_arguments)]
 pub async fn run_escalate_server(
     exec_params: ExecParams,
     sandbox_state: &SandboxState,
```
pakrym-oai: can executor "capture" the sandbox state?

codex-rs/shell-escalation/src/unix/socket.rs
```diff
@@ -314,11 +321,11 @@ async fn send_stream_frame(
...
         let mut guard = socket.writable().await?;
-        let result = guard.try_io(|inner| {
-            send_stream_chunk(inner.get_ref(), &frame[written..], fds, include_fds)
-        });
-        let bytes_written = match result {
-            Ok(bytes_written) => bytes_written?,
+        let bytes_written = match guard
+            .try_io(|inner| send_stream_chunk(inner.get_ref(), &frame[written..], fds, include_fds))
+        {
+            Ok(Ok(bytes_written)) => bytes_written,
```
pakrym-oai: nit:

```
 let read = match guard.try_io(|inner| inner.get_ref().recv(&mut payload[filled..])) {
    Ok(result) => result?,
    Err(_) => continue,
};

```

pakrym-oai (approved):

codex-rs/shell-escalation/src/unix/core_shell_escalation.rs
```diff
@@ -40,7 +40,7 @@ impl ShellPolicyFactory {
     }
 }
 
-struct ShellEscalationPolicy {
+pub struct ShellEscalationPolicy {
```
pakrym-oai, line 46: who uses this pub?

PR #12641, opened 2026-02-24
feat(core) add source for thread-derived rules

dylan-hurd-oai:
## Summary
When adding a new rule from an ExecApprovalRequest, we should add a justification for it.

## Testing
- [x] Added unit tests

### DIFF
--- a/codex-rs/core/src/exec_policy.rs
+++ b/codex-rs/core/src/exec_policy.rs
@@ -18,7 +18,7 @@ use codex_execpolicy::NetworkRuleProtocol;
 use codex_execpolicy::Policy;
 use codex_execpolicy::PolicyParser;
 use codex_execpolicy::RuleMatch;
-use codex_execpolicy::blocking_append_allow_prefix_rule;
+use codex_execpolicy::blocking_append_allow_prefix_rule_with_justification;
 use codex_execpolicy::blocking_append_network_rule;
 use codex_protocol::approvals::ExecPolicyAmendment;
 use codex_protocol::protocol::AskForApproval;
@@ -42,6 +42,7 @@ const REJECT_RULES_APPROVAL_REASON: &str =
 const RULES_DIR_NAME: &str = "rules";
 const RULE_EXTENSION: &str = "rules";
 const DEFAULT_POLICY_FILE: &str = "default.rules";
+const THREAD_PERSISTED_JUSTIFICATION: &str = "persisted during thread";
 static BANNED_PREFIX_SUGGESTIONS: &[&[&str]] = &[
     &["python3"],
     &["python3", "-"],
@@ -290,7 +291,13 @@ impl ExecPolicyManager {
         spawn_blocking({
             let policy_path = policy_path.clone();
             let prefix = prefix.clone();
-            move || blocking_append_allow_prefix_rule(&policy_path, &prefix)
+            move || {
+                blocking_append_allow_prefix_rule_with_justification(
+                    &policy_path,
+                    &prefix,
+                    Some(THREAD_PERSISTED_JUSTIFICATION),
+                )
+            }
         })
         .await
         .map_err(|source| ExecPolicyUpdateError::JoinBlockingTask { source })?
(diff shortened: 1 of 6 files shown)


dylan-hurd-oai:
@codex review

codex-rs/core/src/exec_policy.rs
```diff
@@ -42,6 +42,7 @@ const REJECT_RULES_APPROVAL_REASON: &str =
 const RULES_DIR_NAME: &str = "rules";
 const RULE_EXTENSION: &str = "rules";
 const DEFAULT_POLICY_FILE: &str = "default.rules";
+const THREAD_PERSISTED_JUSTIFICATION: &str = "persisted during thread";
```
bolinfest, line 45: I don't think this is clear. I'm also not sure that our users realize that when we say "thread," we aren't talking about an OS-thread.

Possible additional context that could be useful:

- The full command that this was a part of when it was approved (e.g., `cd foo/bar && echo hello world`) assuming the full command is longer than the prefix?
- Maybe a date or datetime when it was approved? (I know this will make the test a bit more annoying to write.)
- Your own username? e.g. "Always allow `echo hello` was added by dhurd on 2026-03-05T11:16:01"?

codex-rs/core/src/exec_policy.rs
```diff
@@ -1861,23 +1888,75 @@ prefix_rule(pattern=["git"], decision="prompt")
...
         assert_eq!(
             contents,
-            r#"prefix_rule(pattern=["echo", "hello"], decision="allow")
+            r#"prefix_rule(pattern=["echo", "hello"], decision="allow", justification="persisted during thread")
 "#
         );
     }
 
+    #[tokio::test]
+    async fn append_execpolicy_amendment_preserves_existing_justification_when_deduped() {
```
bolinfest, line 1914: More like preserves the absence of a `justification` if it exists?

codex-rs/execpolicy/src/amend.rs
```diff
@@ -122,10 +142,14 @@ pub fn blocking_append_network_rule(
         args.push(format!("justification={justification}"));
     }
     let rule = format!("network_rule({})", args.join(", "));
-    append_rule_line(policy_path, &rule)
+    append_rule_line(policy_path, &rule, None)
 }
 
-fn append_rule_line(policy_path: &Path, rule: &str) -> Result<(), AmendError> {
+fn append_rule_line(
```
bolinfest, line 148: What is the `base_rule`?

codex-rs/execpolicy/src/amend.rs
```diff
@@ -172,7 +199,18 @@ fn append_locked_line(policy_path: &Path, line: &str) -> Result<(), AmendError>
             source,
         })?;
 
-    if contents.lines().any(|existing| existing == line) {
+    if contents.lines().any(|existing| {
```
bolinfest, line 202: There is nothing to guarantee that the user has not reformatted the file to use line breaks. I think we should be using the Starlark parser, though if this is urgent, then I suppose it can be done in a follow-up?

Though admittedly, because Starlark supports macros, it is possible the rule could have been generated from a function and is not defined as a top-level declaration...

codex-rs/execpolicy/src/amend.rs
```diff
@@ -122,10 +142,14 @@ pub fn blocking_append_network_rule(
...
     let rule = format!("network_rule({})", args.join(", "));
-    append_rule_line(policy_path, &rule)
+    append_rule_line(policy_path, &rule, None)
 }
 
-fn append_rule_line(policy_path: &Path, rule: &str) -> Result<(), AmendError> {
+fn append_rule_line(
+    policy_path: &Path,
+    rule: &str,
+    base_rule: Option<&str>,
```
bolinfest, line 151: `prefix_match` is maybe more accurate?

bolinfest (approved):
As noted, I think this could be a bit tighter, though I don't know if it's worth litigating?

PR #12643, opened 2026-02-24
Fix @mention token parsing in chat composer

etraut-openai:
Fixes #12175

If a user types an npm package name with multiple `@` symbols like `npx -y @foo/bar@latest`, the TUI currently treats this as though it's attempting to invoke the file picker.

### What changed

- **Generalized `@` token parsing**
  - `current_prefixed_token(...)` now treats `@` as a token start **only at a whitespace boundary** (or start-of-line).
  - If the cursor is on a nested `@` inside an existing whitespace-delimited token (for example `@scope/pkg@latest`), it keeps the surrounding token active instead of starting a new token at the second `@`.
  - It also avoids misclassifying mid-word usages like `foo@bar` as an `@` file token.

- **Enter behavior with file popup**
  - If the file-search popup is open but has **no selected match**, pressing `Enter` now closes the popup and falls through to normal submit behavior.
  - This prevents pasted strings containing `@...` from blocking submission just because file-search was active with no actionable selection.

### Testing
I manually built and tested the scenarios involved with the bug report and related use of `@` mentions to verify no regressions

### DIFF
--- a/codex-rs/tui/src/bottom_pane/chat_composer.rs
+++ b/codex-rs/tui/src/bottom_pane/chat_composer.rs
@@ -1641,7 +1641,11 @@ impl ChatComposer {
             } => {
                 let Some(sel) = popup.selected_match() else {
                     self.active_popup = ActivePopup::None;
-                    return (InputResult::None, true);
+                    return if key_event.code == KeyCode::Enter {
+                        self.handle_key_event_without_popup(key_event)
+                    } else {
+                        (InputResult::None, true)
+                    };
                 };
 
                 let sel_path = sel.to_string_lossy().to_string();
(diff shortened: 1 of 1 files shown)


codex-rs/tui/src/bottom_pane/chat_composer.rs
```diff
@@ -2014,7 +2014,71 @@ impl ChatComposer {
     ///
     /// The returned string **does not** include the leading `@`.
     fn current_at_token(textarea: &TextArea) -> Option<String> {
```
xl-openai, line 2031: Can we update the comments for the newly added logic?

etraut-openai:
@codex review

etraut-openai:
@codex review

xl-openai (approved):

PR #12648, opened 2026-02-24
fix(exec) Patch resume test race condition

dylan-hurd-oai:
## Summary
The test exec_resume_last_respects_cwd_filter_and_all_flag makes one session “newest” by resuming it, but rollout updated_at is stored/sorted at second precision. On fast CI (especially Windows), the touch could land in the same second as initial session creation, making ordering nondeterministic.

This change adds a short sleep before the recency-touch step so the resumed session is guaranteed to have a later updated_at, preserving the intended assertion without changing product behavior.


etraut-openai (approved):
Adding a sleep to a test is usually sets off big red flags, but I think it's reasonable in this case.

dylan-hurd-oai:
Agreed - ideally we'd have millisecond granularity on the timestamps, but will defer to @jif-oai since I suspect that's a larger change

PR #12052, opened 2026-02-17
feat(core): zsh exec bridge

owenlin0:
zsh fork PR stack:
- https://github.com/openai/codex/pull/12051 
- https://github.com/openai/codex/pull/12052 👈 

### Summary
This PR introduces a feature-gated native shell runtime path that routes shell execution through a patched zsh exec bridge, removing MCP-specific behavior from the shell hot path while preserving existing CommandExecution lifecycle semantics.

When shell_zsh_fork is enabled, shell commands run via patched zsh with per-`execve` interception through EXEC_WRAPPER. Core receives wrapper IPC requests over a Unix socket, applies existing approval policy, and returns allow/deny before the subcommand executes.

### What’s included
**1) New zsh exec bridge runtime in core**
- Wrapper-mode entrypoint (maybe_run_zsh_exec_wrapper_mode) for EXEC_WRAPPER invocations.
- Per-execution Unix-socket IPC handling for wrapper requests/responses.
- Approval callback integration using existing core approval orchestration.
- Streaming stdout/stderr deltas to existing command output event pipeline.
- Error handling for malformed IPC, denial/abort, and execution failures.

**2) Session lifecycle integration**
SessionServices now owns a `ZshExecBridge`.
Session startup initializes bridge state; shutdown tears it down cleanly.

**3) Shell runtime routing (feature-gated)**
When `shell_zsh_fork` is enabled:
- Build execution env/spec as usual.
- Add wrapper socket env wiring.
- Execute via `zsh_exec_bridge.execute_shell_request(...)` instead of the regular shell path.
- Non-zsh-fork behavior remains unchanged.

**4) Config + feature wiring**
- Added `Feature::ShellZshFork` (under development).
- Added config support for `zsh_path` (optional absolute path to patched zsh):
- `Config`, `ConfigToml`, `ConfigProfile`, overrides, and schema.
- Session startup validates that `zsh_path` exists/usable when zsh-fork is enabled.
- Added startup test for missing `zsh_path` failure mode.

**5) Seatbelt/sandbox updates for wrapper IPC**
- Extended seatbelt policy generation to optionally allow outbound connection to explicitly permitted Unix sockets.
- Wired sandboxing path to pass wrapper socket path through to seatbelt policy generation.
- Added/updated seatbelt tests for explicit socket allow rule and argument emission.

**6) Runtime entrypoint hooks**
- This allows the same binary to act as the zsh wrapper subprocess when invoked via `EXEC_WRAPPER`.

**7) Tool selection behavior**
[...]

PR #12649, opened 2026-02-24
feat: run zsh fork shell tool via shell-escalation

bolinfest:
## Why

This PR switches the `shell_command` zsh-fork path over to `codex-shell-escalation` so the new shell tool can use the shared exec-wrapper/escalation protocol instead of the `zsh_exec_bridge` implementation that was introduced in https://github.com/openai/codex/pull/12052. `zsh_exec_bridge` relied on UNIX domain sockets, which is not as tamper-proof as the FD-based approach in `codex-shell-escalation`.

## What Changed

- Added a Unix zsh-fork runtime adapter in `core` (`core/src/tools/runtimes/shell/unix_escalation.rs`) that:
  - runs zsh-fork commands through `codex_shell_escalation::run_escalate_server`
  - bridges exec-policy / approval decisions into `ShellActionProvider`
  - executes escalated commands via a `ShellCommandExecutor` that calls `process_exec_tool_call`
- Updated `ShellRuntime` / `ShellCommandHandler` / tool spec wiring to select a `shell_command` backend (`classic` vs `zsh-fork`) while leaving the generic `shell` tool path unchanged.
- Removed the `zsh_exec_bridge`-based session service and deleted `core/src/zsh_exec_bridge/mod.rs`.
- Moved exec-wrapper entrypoint dispatch to `arg0` by handling the `codex-execve-wrapper` arg0 alias there, and removed the old `codex_core::maybe_run_zsh_exec_wrapper_mode()` hooks from `cli` and `app-server` mains.
- Added the needed `codex-shell-escalation` dependencies for `core` and `arg0`.

## Tests

- `cargo test -p codex-core shell_zsh_fork_prefers_shell_command_over_unified_exec`
- `cargo test -p codex-app-server turn_start_shell_zsh_fork -- --nocapture`
  - verifies zsh-fork command execution and approval flows through the new backend
  - includes subcommand approve/decline coverage using the shared zsh DotSlash fixture in `app-server/tests/suite/zsh`
- To test manually, I added the following to `~/.codex/config.toml`:

```toml
zsh_path = "/Users/mbolin/code/codex3/codex-rs/app-server/tests/suite/zsh"

[features]
shell_zsh_fork = true
```

Then I ran `just c` to run the dev build of Codex with these changes and sent it the message:

```
run `echo $0`
```

And it replied with:

```
  echo $0 printed:

  /Users/mbolin/code/codex3/codex-rs/app-server/tests/suite/zsh

  In this tool context, $0 reflects the script path used to invoke the shell, not just zsh.
```

so the tool appears to be wired up correctly.

## Notes

- The zsh subcommand-decline integration test now uses `rm` under a `WorkspaceWrite` sandbox. The previous `/usr/bin/true` scenario is auto-allowed by the new `shell-escalation` policy path, which no longer produces subcommand approval prompts.


codex-rs/core/src/tools/runtimes/shell.rs
```diff
@@ -26,27 +29,38 @@ use crate::tools::sandboxing::ToolCtx;
...
     pub network: Option<NetworkProxy>,
     pub sandbox_permissions: SandboxPermissions,
     pub justification: Option<String>,
     pub exec_approval_requirement: ExecApprovalRequirement,
 }
 
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
+pub(crate) enum ShellRuntimeBackend {
+    #[default]
+    Generic,
```
pakrym-oai, line 67: Generic and classic?

codex-rs/core/src/tools/runtimes/shell.rs
```diff
@@ -26,27 +29,38 @@ use crate::tools::sandboxing::ToolCtx;
...
+    ShellCommandClassic,
+    ShellCommandZshFork,
+}
+
 #[derive(Default)]
-pub struct ShellRuntime;
+pub struct ShellRuntime {
+    #[cfg_attr(not(unix), allow(dead_code))]
```
pakrym-oai, line 83: Do we need the viral cross compilation?

codex-rs/core/src/tools/runtimes/shell.rs
```diff
@@ -181,34 +200,16 @@ impl ToolRuntime<ShellRequest, ExecToolCallOutput> for ShellRuntime {
...
-                .env_for(spec, req.network.as_ref())
-                .map_err(|err| ToolError::Codex(err.into()))?;
-            return ctx
-                .session
-                .services
-                .zsh_exec_bridge
-                .execute_shell_request(&env, &ctx.session, &ctx.turn, &ctx.call_id)
-                .await;
+        #[cfg(unix)]
+        if self.backend == ShellRuntimeBackend::ShellCommandZshFork {
```
pakrym-oai, line 217: there are 3 items in ShellRuntimeBackend, should we use match?

codex-rs/core/src/tools/handlers/shell.rs
```diff
@@ -63,6 +73,25 @@ impl ShellHandler {
 }
 
 impl ShellCommandHandler {
+    pub(crate) fn classic() -> Self {
+        Self {
+            backend: ShellCommandBackend::Classic,
```
pakrym-oai: can this be a param in initializer?

codex-rs/core/src/sandboxing/mod.rs
```diff
@@ -163,19 +163,13 @@ impl SandboxManager {
...
-                    .as_ref()
-                    .map_or_else(Vec::new, |path| vec![path.clone()]);
                 let mut args = create_seatbelt_command_args(
                     command.clone(),
                     policy,
                     sandbox_policy_cwd,
                     enforce_managed_network,
                     network,
-                    &zsh_exec_bridge_allowed_unix_sockets,
+                    &[],
```
pakrym-oai, line 356: is this param used anymore?

codex-rs/core/src/state/service.rs
```diff
@@ -27,7 +27,8 @@ pub(crate) struct SessionServices {
     pub(crate) mcp_connection_manager: Arc<RwLock<McpConnectionManager>>,
     pub(crate) mcp_startup_cancellation_token: Mutex<CancellationToken>,
     pub(crate) unified_exec_manager: UnifiedExecProcessManager,
-    pub(crate) zsh_exec_bridge: ZshExecBridge,
+    #[cfg_attr(not(unix), allow(dead_code))]
```
pakrym-oai, line 30: can we keep the prop and keep it None?

codex-rs/core/src/sandboxing/mod.rs
bolinfest: no; will address in a follow-up

codex-rs/core/src/tools/runtimes/shell.rs
bolinfest: done! added a wildcard import to try to make it less verbose

codex-rs/core/src/tools/handlers/shell.rs
bolinfest: introduced `impl From<ShellCommandBackendConfig> for ShellCommandHandler`

codex-rs/shell-escalation/src/unix/escalate_server.rs
```diff
@@ -58,6 +58,9 @@ pub struct ExecParams {
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
 pub struct ExecResult {
     pub exit_code: i32,
+    pub stdout: String,
```
pakrym-oai, line 61: I don't know whether it's worth splitting from core, this is an exact copy of ExecToolCallOutput

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
```diff
@@ -0,0 +1,455 @@
...
+            script: script.clone(),
+            login: true,
+        }),
+        _ => Err(ToolError::Rejected(
+            "unexpected shell command format for zsh-fork execution".to_string(),
+        )),
+    }
+}
+
+fn map_exec_result(
```
pakrym-oai, line 387: Should we move this code into core and avoid these extra transformations between exactly the same structs?

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
```diff
@@ -0,0 +1,455 @@
...
+
+#[derive(Debug, Eq, PartialEq)]
+struct ParsedShellCommand {
+    script: String,
+    login: bool,
+}
+
+fn extract_shell_script(command: &[String]) -> Result<ParsedShellCommand, ToolError> {
+    match command {
+        [_, flag, script, ..] if flag == "-c" => Ok(ParsedShellCommand {
```
pakrym-oai, line 373: don't we already have this function in shell_command?

codex-rs/core/src/tools/runtimes/shell.rs
bolinfest: I'll have Codex document it.

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
pakrym-oai: maybe extend extract_bash_command ?

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
```diff
@@ -0,0 +1,455 @@
...
+            Decision::Forbidden => EscalateAction::Deny {
+                reason: Some("Execution forbidden by policy".to_string()),
+            },
+            Decision::Prompt => {
+                if self.approval_policy == AskForApproval::Never {
+                    EscalateAction::Deny {
+                        reason: Some("Execution forbidden by policy".to_string()),
+                    }
+                } else if decision_driven_by_policy {
+                    EscalateAction::Escalate
```
pakrym-oai: why do we immidiately escalate if the policy asks us to prompt/

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
```diff
@@ -0,0 +1,455 @@
...
+                        ReviewDecision::Denied => EscalateAction::Deny {
+                            reason: Some("User denied execution".to_string()),
+                        },
+                        ReviewDecision::Abort => EscalateAction::Deny {
+                            reason: Some("User cancelled execution".to_string()),
+                        },
+                    }
+                }
+            }
+            Decision::Allow => EscalateAction::Run,
```
pakrym-oai: Shouldn't this be escalate?

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
bolinfest: sort of, but the purpose is different here: if it doesn't match it's an error

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
bolinfest: OK, fair point: I'll take a deeper look tomorrow.

pakrym-oai (approved):

PR #12650, opened 2026-02-24
chore: rm hardcoded PRESETS list

sayan-oai:
rm `PRESETS` list harcoded in `model_presets` as we now have bundled `models.json` with equivalent info.

update logic to rely on bundled models instead, update tests.


aibrahim-oai (approved):

sayan-oai:
@codex review this

PR #12697, opened 2026-02-24
Ensure shell command skills trigger approval

pakrym-oai:
Summary
- detect skill-invoking shell commands based on the original command string, request approvals when needed, and cache positive decisions per session
- keep implicit skill invocation emitted after approval and keep skill approval decline messaging centralized to the shell handler
- expand and adjust skill approval tests to cover shell-based skill scripts while matching the new detection expectations

Testing
- Not run (not requested)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -1758,8 +1758,10 @@ impl CodexMessageProcessor {
             None => None,
         };
         let windows_sandbox_level = WindowsSandboxLevel::from_config(&self.config);
+        let command = params.command;
         let exec_params = ExecParams {
-            command: params.command,
+            original_command: command.join(" "),
```
bolinfest, line 1763: why not shlex.join?

pakrym-oai: this codepath doesn't matter anymore this is v1 run_one_off_command

codex-rs/core/src/tools/handlers/shell.rs
```diff
@@ -55,6 +53,8 @@ impl ShellHandler {
     ) -> ExecParams {
         ExecParams {
             command: params.command.clone(),
+            original_command: shlex::try_join(params.command.iter().map(String::as_str))
```
bolinfest, line 60: I thought we finally created a utility for this, but `rg shlex codex-rs/utils` turns up nothing...

codex-rs/core/src/skills/invocation_utils.rs
```diff
@@ -53,6 +69,82 @@ fn detect_implicit_skill_invocation_for_command(
...
+
+pub(crate) async fn ensure_skill_approval_for_command(
+    sess: &Session,
+    turn_context: &TurnContext,
+    item_id: &str,
+    command: &str,
+    workdir: &Path,
+) -> bool {
+    if !turn_context.features.enabled(Feature::SkillApproval) {
+        return true;
```
bolinfest, line 107: Should this be `false`? If the feature is disabled, it feels like we haven't demonstrated the skill is approved?

codex-rs/core/src/skills/invocation_utils.rs
```diff
@@ -53,6 +69,82 @@ fn detect_implicit_skill_invocation_for_command(
     None
 }
 
+pub(crate) fn detect_implicit_skill_script_invocation_for_command(
+    outcome: &SkillLoadOutcome,
+    command: &str,
+    workdir: &Path,
+) -> Option<SkillMetadata> {
```
bolinfest, line 76: Should it be possible to match multiple skills from one `command`?

codex-rs/core/src/skills/invocation_utils.rs
bolinfest: If this should be `true`, I'm unclear on the contract of this function: can you add a docstring?

bolinfest (approved):

owenlin0 (approved):

codex-rs/core/src/skills/invocation_utils.rs
pakrym-oai: Sorry, forgot to respond. I expect this to change to tri-state we start inferring permissions from skills.

True - inject permissions
False - reject invocation 
None - do nothing.

PR #12702, opened 2026-02-24
Improve macOS Seatbelt network and unix socket handling

viyatb-oai:
This improves macOS Seatbelt handling for sandboxed tool processes.

## Changes
- Allow dual-stack local binding in proxy-managed sessions, while still keeping traffic limited to loopback and configured proxy endpoints.
- Replace the old generic unix-socket path rule with explicit AF_UNIX permissions for socket creation, bind, and outbound connect.
- Keep explicitly approved wrapper sockets connect-only.

Local helper servers are less likely to fail when binding on macOS.
Tools using local unix-socket IPC should work more reliably under the sandbox.
Full-network sessions, proxy fail-closed behavior, and proxy lifecycle are unchanged.


codex-rs/core/src/seatbelt.rs
```diff
@@ -218,29 +218,57 @@ fn unix_socket_dir_params(proxy: &ProxyPolicyInputs) -> Vec<(String, PathBuf)> {
...
+
+    let mut policy = String::new();
+    policy.push_str("(allow system-socket (socket-domain AF_UNIX))\n");
     if matches!(
         proxy.unix_domain_socket_policy,
         UnixDomainSocketPolicy::AllowAll
     ) {
-        return "(allow network* (subpath \"/\"))\n".to_string();
+        // Keep "allow all" scoped to absolute filesystem-backed socket paths instead
+        // of a completely unqualified unix-socket allowance.
```
nornagon-openai: What's the reasoning for permitting FS sockets but disallowing abstract sockets?

codex-rs/core/src/seatbelt.rs
```diff
@@ -218,29 +218,57 @@ fn unix_socket_dir_params(proxy: &ProxyPolicyInputs) -> Vec<(String, PathBuf)> {
...
-    unix_socket_path_params(proxy)
-        .iter()
-        .map(|param| format!("(allow network* (subpath (param \"{}\")))\n", param.key))
-        .collect()
+    for param in socket_params {
+        // Use subpath so allowlists cover sockets created beneath approved directories.
+        policy.push_str(&format!(
+            "(allow network-bind (local unix-socket (subpath (param \"{}\"))))\n",
+            param.key
+        ));
```
nornagon-openai, line 254: This isn't fully escaped. Is param.key always controlled by us, or can it be influenced by other inputs?

codex-rs/core/src/seatbelt.rs
```diff
@@ -218,29 +218,57 @@ fn unix_socket_dir_params(proxy: &ProxyPolicyInputs) -> Vec<(String, PathBuf)> {
...
-        let mut policy =
-            String::from("; allow outbound access only to configured loopback proxy endpoints\n");
+    let should_use_restricted_network_policy =
+        !proxy.ports.is_empty() || proxy.has_proxy_config || enforce_managed_network;
+    if should_use_restricted_network_policy {
+        let mut policy = if proxy.ports.is_empty() {
+            String::from("; restrict network access without inferred proxy endpoints\n")
+        } else {
+            String::from("; allow outbound access only to configured loopback proxy endpoints\n")
+        };
```
nornagon-openai: it seems odd to branch here only to add a comment. This can safely be dropped.

codex-rs/core/src/seatbelt.rs
viyatb-oai: afaik Seatbelt can meaningfully constrain only path-backed AF_UNIX sockets. Abstract namespace unix sockets are a Linux-specific mechanism rather than a macOS thing.

codex-rs/core/src/seatbelt.rs
viyatb-oai: controlled by us, but I made this more clear!

codex-rs/core/src/seatbelt.rs
viyatb-oai: dropped

nornagon-openai (approved):

PR #12712, opened 2026-02-24
Promote js_repl to experimental with Node requirement

fjord-oai:
## Summary

- Promote `js_repl` to an experimental feature that users can enable from `/experimental`.
- Add `js_repl` experimental metadata, including the Node prerequisite and activation guidance.
- Add regression coverage for the feature metadata and the `/experimental` popup.

## What Changed

- Changed `Feature::JsRepl` from `Stage::UnderDevelopment` to `Stage::Experimental`.
- Added experimental metadata for `js_repl` in `core/src/features.rs`:
  - name: `JavaScript REPL`
  - description: calls out interactive website debugging, inline JavaScript execution, and the required Node version (`>= v24.13.1`)
  - announcement: tells users to enable it, then start a new chat or restart Codex
- Added a core unit test that verifies:
  - `js_repl` is experimental
  - `js_repl` is disabled by default
  - the hardcoded Node version in the description matches `node-version.txt`
- Added a TUI test that opens the `/experimental` popup and verifies the rendered `js_repl` entry includes the Node requirement text.

## Testing

- `just fmt`
- `cargo test -p codex-tui`
- `cargo test -p codex-core` (unit-test phase passed; stopped during the long `tests/all.rs` integration suite)

### DIFF
--- a/codex-rs/core/src/features.rs
+++ b/codex-rs/core/src/features.rs
@@ -460,7 +460,11 @@ pub const FEATURES: &[FeatureSpec] = &[
     FeatureSpec {
         id: Feature::JsRepl,
         key: "js_repl",
-        stage: Stage::UnderDevelopment,
+        stage: Stage::Experimental {
+            name: "JavaScript REPL",
+            menu_description: "Enable a persistent Node-backed JavaScript REPL for interactive website debugging and other inline JavaScript execution capabilities. Requires Node >= v24.13.1 installed.",
+            announcement: "NEW: JavaScript REPL is now available in /experimental. Enable it, then start a new chat or restart Codex to use it.",
+        },
         default_enabled: false,
     },
     FeatureSpec {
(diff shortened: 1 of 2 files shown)


codex-rs/core/src/features.rs
```diff
@@ -425,6 +429,29 @@ pub struct FeatureSpec {
...
+                .stage
+                .experimental_menu_description()
+                .map(str::to_owned),
+        }
+    }
+
+    pub fn experimental_announcement(&self) -> Option<&'static str> {
+        self.stage.experimental_announcement()
+    }
+}
```
dylan-hurd-oai: This feels a tad overkill for the node version - do we expect the minimum version to update frequently?

dylan-hurd-oai (approved):

codex-rs/core/src/features.rs
fjord-oai: Nah, just thought it would be helpful. If you'd like I can remove the extra wiring. Users will still get a reasonable error message if they try to actually use the tool with a node that's too old.

dylan-hurd-oai: I don't feel super strongly, it seems like we already have this hooked up to app-server anyway.

fjord-oai: Realized that the simpler approach is to hard code the string in the feature description to keep things simple, and then check against node-version.txt in the unit test.

PR #12731, opened 2026-02-25
chore: migrate additional permissions to PermissionProfile

celia-oai:
This PR replaces the old `additional_permissions.fs_read/fs_write` shape with a shared `PermissionProfile`
  model and wires it through the command approval, sandboxing, protocol, and TUI layers. The schema is adopted from the `SkillManifestPermissions`, which is also refactored to use this unified struct. This helps us easily expose permission profiles in app server/core as a follow-up.


codex-rs/core/src/skills/permissions.rs
```diff
@@ -3,10 +3,13 @@ use std::path::Component;
 use std::path::Path;
 use std::path::PathBuf;
 
+use codex_protocol::models::MacOsAutomationValue;
+use codex_protocol::models::MacOsPermissions;
+use codex_protocol::models::MacOsPreferencesValue;
+use codex_protocol::models::PermissionProfile;
```
dylan-hurd-oai, line 11: I think these are getting hit by the macos target - we could macro the imports or qualify the use inside the function?

dylan-hurd-oai (approved):
Couple build errors to fix but this looks directionally reasonable to me

PR #12732, opened 2026-02-25
feat(app-server): add ThreadItem::DynamicToolCall

owenlin0:
Previously, clients would call `thread/start` with dynamic_tools set, and when a model invokes a dynamic tool, it would just make the server->client `item/tool/call` request and wait for the client's response to complete the tool call. This works, but it doesn't have an `item/started` or `item/completed` event.

Now we are doing this:
- [new] emit `item/started` with `DynamicToolCall` populated with the call arguments
- send an `item/tool/call` server request
- [new] once the client responds, emit `item/completed` with `DynamicToolCall` populated with the response.

Also, with `persistExtendedHistory: true`, dynamic tool calls are now reconstructable in `thread/read` and `thread/resume` as `ThreadItem::DynamicToolCall`.


owenlin0:
@codex review

celia-oai (approved):

PR #12735, opened 2026-02-25
Enable request_user_input in Default mode

charley-oai:
## Summary
- allow `request_user_input` in Default collaboration mode as well as Plan
- update the Default-mode instructions to prefer assumptions first and use `request_user_input` only when a question is unavoidable
- update request_user_input and app-server tests to match the new Default-mode behavior
- refactor collaboration-mode availability plumbing into `CollaborationModesConfig` for future mode-related flags

## Codex author
`codex resume 019c9124-ed28-7c13-96c6-b916b1c97d49`


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1250,6 +1251,38 @@ pub struct ModelListResponse {
...
+pub struct CollaborationModeMask {
+    pub name: String,
+    pub mode: Option<ModeKind>,
+    pub model: Option<String>,
+    #[serde(rename = "reasoning_effort")]
+    #[ts(rename = "reasoning_effort")]
+    pub reasoning_effort: Option<Option<ReasoningEffort>>,
+    #[serde(rename = "developer_instructions")]
+    #[ts(rename = "developer_instructions")]
+    #[deprecated(
```
aibrahim-oai: is it deprecated or just use it to override instructions?

charley-oai: I'll just remove the field. This type is only used for the return type of the list endpoint, not params of any other endpoints

aibrahim-oai (approved):

PR #12740, opened 2026-02-25
Add macOS and Linux direct install script

efrazer-oai:
## Summary
- add a direct install script for macOS and Linux at `scripts/install/install.sh`
- stage `install.sh` into `dist/` during release so it is published as a GitHub release asset
- reuse the existing platform npm payload so the installer includes both `codex` and `rg`

## Testing
- `bash -n scripts/install/install.sh`
- local macOS `curl | sh` smoke test against a locally served copy of the script


EFRAZER-oai:

EFRAZER-oai:
recheck

mzeng-openai (approved):

PR #12753, opened 2026-02-25
Surface skill permission profiles in zsh-fork exec approvals

celia-oai:
## Summary

  - Preserve each skill’s raw permissions block as a permission_profile on SkillMetadata during skill loading.
  - Keep compiling that same metadata into the existing runtime Permissions object, so current enforcement
    behavior stays intact.
  - When zsh-fork intercepts execution of a script that belongs to a skill, include the skill’s
    permission_profile in the exec approval request.
  - This lets approval UIs show the extra filesystem access the skill declared when prompting for approval.


codex-rs/core/src/skills/loader.rs
```diff
@@ -539,17 +541,18 @@ fn load_skill_metadata(
     Option<SkillInterface>,
     Option<SkillDependencies>,
     Option<SkillPolicy>,
+    Option<PermissionProfile>,
```
bolinfest: In a follow-up, we should figure out the story between `PermissionProfile` and `Permissions`.

Also, we should return a struct, this tuple is hard to work with...

codex-rs/core/src/skills/loader.rs
```diff
@@ -1356,6 +1366,17 @@ permissions:
             outcome.errors
         );
         assert_eq!(outcome.skills.len(), 1);
+        assert_eq!(
+            outcome.skills[0].permission_profile,
+            Some(PermissionProfile {
+                network: Some(true),
+                file_system: Some(FileSystemPermissions {
+                    read: Some(vec![PathBuf::from("./data")]),
```
bolinfest, line 1379: We should move from `PathBuf` to `AbsolutePathBuf` in `FileSystemPermissions`, but then we have to be sure to deserialize the YAML using `AbsolutePathBufGuard`.

bolinfest (approved):

codex-rs/core/src/skills/loader.rs
celia-oai: yeah linter complained as well so I changed

PR #12758, opened 2026-02-25
feat: include available decisions in command approval requests

bolinfest:
Command-approval clients currently infer which choices to show from side-channel fields like `networkApprovalContext`, `proposedExecpolicyAmendment`, and `additionalPermissions`. That makes the request shape harder to evolve, and it forces each client to replicate the server's heuristics instead of receiving the exact decision list for the prompt.

This PR introduces a mapping between `CommandExecutionApprovalDecision` and `codex_protocol::protocol::ReviewDecision`:

```rust
impl From<CoreReviewDecision> for CommandExecutionApprovalDecision {
    fn from(value: CoreReviewDecision) -> Self {
        match value {
            CoreReviewDecision::Approved => Self::Accept,
            CoreReviewDecision::ApprovedExecpolicyAmendment {
                proposed_execpolicy_amendment,
            } => Self::AcceptWithExecpolicyAmendment {
                execpolicy_amendment: proposed_execpolicy_amendment.into(),
            },
            CoreReviewDecision::ApprovedForSession => Self::AcceptForSession,
            CoreReviewDecision::NetworkPolicyAmendment {
                network_policy_amendment,
            } => Self::ApplyNetworkPolicyAmendment {
                network_policy_amendment: network_policy_amendment.into(),
            },
            CoreReviewDecision::Abort => Self::Cancel,
            CoreReviewDecision::Denied => Self::Decline,
        }
    }
}
```

And updates `CommandExecutionRequestApprovalParams` to have a new field:

```rust
available_decisions: Option<Vec<CommandExecutionApprovalDecision>>
```

when, if specified, should make it easier for clients to display an appropriate list of options in the UI.

This makes it possible for `CoreShellActionProvider::prompt()` in `unix_escalation.rs` to specify the `Vec<ReviewDecision>` directly, adding support for `ApprovedForSession` when approving a skill script, which was previously missing in the TUI.

Note this results in a significant change to `exec_options()` in `approval_overlay.rs`, as the displayed options are now derived from `available_decisions: &[ReviewDecision]`.

## What Changed

- Add `available_decisions` to [`ExecApprovalRequestEvent`](https://github.com/openai/codex/blob/de00e932dd9801de0a4faac0519162099753f331/codex-rs/protocol/src/approvals.rs#L111-L175), including helpers to derive the legacy default choices when older senders omit the field.
- Map `codex_protocol::protocol::ReviewDecision` to app-server `CommandExecutionApprovalDecision` and expose the ordered list as experimental `availableDecisions` in [`CommandExecutionRequestApprovalParams`](https://github.com/openai/codex/blob/de00e932dd9801de0a4faac0519162099753f331/codex-rs/app-server-protocol/src/protocol/v2.rs#L3798-L3807).
- Thread optional `available_decisions` through the core approval path so Unix shell escalation can explicitly request `ApprovedForSession` for session-scoped approvals instead of relying on client heuristics. [`unix_escalation.rs`](https://github.com/openai/codex/blob/de00e932dd9801de0a4faac0519162099753f331/codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs#L194-L214)
- Update the TUI approval overlay to build its buttons from the ordered decision list, while preserving the legacy fallback when `available_decisions` is missing.
- Update the app-server README, test client output, and generated schema artifacts to document and surface the new field.

## Testing

- Add `approval_overlay.rs` coverage for explicit decision lists, including the generic `ApprovedForSession` path and network approval options.
- Update `chatwidget/tests.rs` and app-server protocol tests to populate the new optional field and keep older event shapes working.

## Developers Docs

- If we document `item/commandExecution/requestApproval` on [developers.openai.com/codex](https://developers.openai.com/codex), add experimental `availableDecisions` as the preferred source of approval choices and note that older servers may omit it.


codex-rs/app-server-protocol/schema/typescript/v2/CommandExecutionRequestApprovalParams.ts
```diff
@@ -49,4 +50,8 @@ proposedExecpolicyAmendment?: ExecPolicyAmendment | null,
 /**
  * Optional proposed network policy amendments (allow/deny host) for future requests.
  */
-proposedNetworkPolicyAmendments?: Array<NetworkPolicyAmendment> | null, };
+proposedNetworkPolicyAmendments?: Array<NetworkPolicyAmendment> | null,
```
pakrym-oai, line 53: should this be removed in the new design?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -757,6 +759,27 @@ pub enum CommandExecutionApprovalDecision {
...
+            } => Self::AcceptWithExecpolicyAmendment {
+                execpolicy_amendment: proposed_execpolicy_amendment.into(),
+            },
+            CoreReviewDecision::ApprovedForSession => Self::AcceptForSession,
+            CoreReviewDecision::NetworkPolicyAmendment {
+                network_policy_amendment,
+            } => Self::ApplyNetworkPolicyAmendment {
+                network_policy_amendment: network_policy_amendment.into(),
+            },
+            CoreReviewDecision::Denied => Self::Decline,
```
pakrym-oai: We should maybe not list this one.

codex-rs/app-server-protocol/src/protocol/v2.rs
pakrym-oai, line 755: I know that it's not your change but a little strange that these are separate.

codex-rs/protocol/src/approvals.rs
```diff
@@ -116,6 +124,52 @@ impl ExecApprovalRequestEvent {
...
     }
+
+    pub fn default_available_decisions(
+        network_approval_context: Option<&NetworkApprovalContext>,
+        proposed_execpolicy_amendment: Option<&ExecPolicyAmendment>,
+        proposed_network_policy_amendments: Option<&[NetworkPolicyAmendment]>,
+        additional_permissions: Option<&PermissionProfile>,
+    ) -> Vec<ReviewDecision> {
+        if network_approval_context.is_some() {
+            let mut decisions = vec![ReviewDecision::Approved, ReviewDecision::ApprovedForSession];
```
pakrym-oai, line 149: This logic feels extra convoluted. Why don't we define a single list and build it up based on available options?

Instead we have 3 different branches where we rebuild the list from scrathc.

pakrym-oai:
No app-server integration test?

pakrym-oai (approved):

codex-rs/app-server-protocol/schema/typescript/v2/CommandExecutionRequestApprovalParams.ts
bolinfest: It's not marked "experimental," so it's not officially safe to remove...

codex-rs/app-server-protocol/src/protocol/v2.rs
bolinfest: Removing...

codex-rs/app-server-protocol/src/protocol/v2.rs
bolinfest: Yes, I think this is WIP.

codex-rs/protocol/src/approvals.rs
bolinfest: I agree this is a bit gross.

Ideally, we would consolidate this logic in ExecApprovalRequestEvent such that we always went from `ExecApprovalRequestEvent` to `Vec<ReviewDecision>`.

The problem is that not every callsite into `request_command_approval()`:

https://github.com/openai/codex/blob/c1851be1ed8f6ad4c8786f335e9e65b9b9d910fb/codex-rs/core/src/codex.rs#L2594-L2606

Has a full `ExecApprovalRequestEvent` object. For example, `unified_exec.rs` grabs stuff from seemingly all over the place:

https://github.com/openai/codex/blob/c1851be1ed8f6ad4c8786f335e9e65b9b9d910fb/codex-rs/core/src/tools/runtimes/unified_exec.rs#L114-L127

PR #12763, opened 2026-02-25
feat: support OAuth resource parameter for MCP login

### LINKED ISSUE #12762: MCP OAuth login cannot pass RFC 8707 resource parameter
## Summary
OAuth login for MCP servers does not provide a way to include the RFC 8707 `resource` parameter in the authorization request. Some OAuth providers require `resource`, so the login flow fails even though the user completes authorization in the browser.

## Repro
1. Configure a streamable HTTP MCP server that uses OAuth and requires `resource` on the authorization request.
2. Run `codex mcp login <server-name>`.
3. The browser flow completes, but the token exchange fails (no `access_token`), because the authorization URL never included `resource`.

## Expected
Codex should allow specifying an OAuth resource value per MCP server and append it to the authorization URL when starting the login flow.

## Notes
The current flow only passes `scope`, `redirect_uri`, and `client_name`. There is no config/CLI way to pass an OAuth `resource` value.

xrmzju:
## Summary
Add optional per-server `oauth_resource` configuration and append it to the OAuth authorization URL during MCP login. This supports OAuth providers that require RFC 8707 `resource` in the authorization request.

## Changes
- Add `oauth_resource` to MCP server config (validated only for streamable HTTP).
- Propagate `oauth_resource` through CLI/app-server login paths.
- Append `resource` to the authorization URL when present.

## Testing
- Not run (config-only + URL construction change).

Fixes #12762

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -4065,6 +4065,7 @@ impl CodexMessageProcessor {
             http_headers,
             env_http_headers,
             scopes.as_deref().unwrap_or_default(),
+            server.oauth_resource.as_deref(),
             timeout_secs,
             config.mcp_oauth_callback_port,
             config.mcp_oauth_callback_url.as_deref(),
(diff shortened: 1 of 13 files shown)


xrmzju:

xrmzju:
Addressed Codex review feedback: removed invalid oauth_resource tokens from config structs and removed duplicate oauth_resource field in run_add. Pushed fix in 8247638; CI should re-run (Bazel failures can be ignored per prior note).

PR #12774, opened 2026-02-25
feat: add post-compaction sub-agent infos

jif-oai:


codex-rs/core/src/environment_context.rs
```diff
@@ -16,6 +16,7 @@ pub(crate) struct EnvironmentContext {
     pub cwd: Option<PathBuf>,
     pub shell: Shell,
     pub network: Option<NetworkContext>,
+    pub subagents: Option<String>,
```
pakrym-oai, line 17: We should consider adding subagent info as a seprate message. @charley-oai has a PR where all separate messages are getting merged.

jif-oai: Happy to do it once the PR is there

charley-oai (approved):

PR #12795, opened 2026-02-25
Revert "fix(bazel): replace askama templates with include_str! in memories"

jif-oai:
Reverts openai/codex#11778


codex-rs/core/BUILD.bazel
```diff
@@ -14,6 +14,12 @@ codex_rust_crate(
     ) + [
         "//codex-rs:node-version.txt",
     ],
+    rustc_env = {
+        # Askama resolves template paths relative to CARGO_MANIFEST_DIR. In
```
zbarsky-openai, line 18: interseting, I've seen some weirdness/confusion in crate around when CARGO_MANIFEST_DIR is resolved (build time of dep vs build time of consuming crate) so I wonder if it's related? But anyway if this compiles I guess that works as well :) Just a bit worried about the blast radius here since `core` is so big...

zbarsky-openai (approved):

Issue #12755, opened 2026-02-25
Your access token could not be refreshed because your refresh token was already used. Please log out and sign in again.


andreaslhv:
### What version of Codex CLI is running?

codex-cli 0.104.0

### What subscription do you have?

Enterprise

### Which model were you using?

gpt-5.3-codex high

### What platform is your computer?

Darwin 25.3.0 arm64 arm

### What terminal emulator and version are you using (if applicable)?

Terminal.app

### What issue are you seeing?

Problem with Atlassian MCP - "Your access token could not be refreshed because your refresh token was already used. Please log out and sign in again."

Usually `codex mcp remove atlassian` followed by `codex mcp add atlassian --url https://mcp.atlassian.com/v1/mcp` resolved the refresh issue, but not this time.

### What steps can reproduce the bug?

1. `codex mcp add atlassian --url https://mcp.atlassian.com/v1/mcp`
2. `codex`
3. `Can you see Jira?`

### What is the expected behavior?

It should see Jira (worked last week)

PR #12815, opened 2026-02-25
Fix MCP OAuth refresh-token races across Codex processes

etraut-openai:
Addresses https://github.com/openai/codex/issues/12755

Token refresh for MCP OAuth suffered from the same problem that our main Codex ChatGPT auth had until recently. This PR updates the MCP logic to mirror the solution in place for ChatGPT auth token refresh.

MCP OAuth credentials are shared across processes via keyring or
`CODEX_HOME/.credentials.json`, but each process kept its own in-memory
snapshot and refreshed from that local copy. That meant one process could
rotate the refresh token while another still tried to use the stale token,
causing `refresh_token_reused` failures.

This changes the MCP refresh path to use a guarded reload flow modeled on
the main ChatGPT auth fix:
- reload shared MCP OAuth credentials before any refresh attempt
- compare persisted credentials to the in-memory cached snapshot with
  refresh-specific equality
- if persisted credentials changed, adopt them into the live RMCP auth runtime
  and skip the local refresh
- if persisted credentials are unchanged, perform the refresh and persist the
  rotated credentials
- if persisted credentials are missing, treat that as logged out and clear the
  live runtime auth instead of using a stale refresh token

To keep the running RMCP auth stack in sync without reconnecting, the client
now owns the `AuthorizationManager` credential store and can update the live
runtime in place when reloaded credentials win.

Also adds:
- unit tests for guarded reload decision logic in `codex-rs/rmcp-client/src/oauth.rs`
- an end-to-end two-client integration test in
  `codex-rs/core/tests/suite/rmcp_client.rs`
- refresh-token support in the streamable HTTP MCP test server fixture

### DIFF
--- a/codex-rs/rmcp-client/src/oauth.rs
+++ b/codex-rs/rmcp-client/src/oauth.rs
@@ -376,6 +376,8 @@ impl OAuthPersistor {
     /// MCP OAuth credentials live in shared storage, but each Codex process also keeps an
     /// in-memory snapshot. Before refreshing, reload the shared credentials and compare them to
     /// the cached copy:
+    /// - if the local cache was cleared, reload shared storage first so this process can recover
+    ///   when another process logs in and persists fresh credentials;
     /// - if shared storage changed, another process already refreshed, so adopt those credentials
     ///   in the live runtime and skip the local refresh;
     /// - if shared storage is unchanged, this process still owns the refresh and can rotate the
@@ -383,10 +385,22 @@ impl OAuthPersistor {
     /// - if shared storage no longer has credentials, treat that as logged out and clear the live
     ///   runtime instead of sending a stale refresh token.
     pub(crate) async fn refresh_if_needed(&self) -> Result<()> {
-        let cached_credentials = {
+        let mut cached_credentials = {
             let guard = self.inner.last_credentials.lock().await;
             guard.clone()
         };
+
+        if cached_credentials.is_none()
+            && let Some(credentials) = load_oauth_tokens_when_cache_missing(
+                &self.inner.server_name,
+                &self.inner.url,
+                self.inner.store_mode,
+            )
+        {
+            self.apply_runtime_credentials(Some(credentials.clone()))
+                .await?;
+            cached_credentials = Some(credentials);
+        }
 
         match self.guarded_refresh_outcome(cached_credentials.as_ref()) {
             GuardedRefreshOutcome::NoAction => Ok(()),
@@ -499,6 +513,18 @@ fn load_oauth_tokens_for_guarded_refresh(
                 server_name,
             )
         }
+    }
+}
+
+fn load_oauth_tokens_when_cache_missing(
+    server_name: &str,
+    url: &str,
+    store_mode: OAuthCredentialsStoreMode,
+) -> Option<StoredOAuthTokens> {
+    match load_oauth_tokens_for_guarded_refresh(server_name, url, store_mode) {
+        GuardedRefreshPersistedCredentials::Loaded(Some(credentials)) => Some(credentials),
+        GuardedRefreshPersistedCredentials::Loaded(None)
+        | GuardedRefreshPersistedCredentials::ReloadFailed => None,
     }
 }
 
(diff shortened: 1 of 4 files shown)


etraut-openai:
@codex review

btpod:
Will we get a fix out soon? Codex is basically unusable currently

```Your access token could not be refreshed because your refresh token was already use```

File: codex-rs/rmcp-client/src/rmcp_client.rs [hunk 1]
```diff
@@ -595,22 +604,20 @@ async fn create_oauth_transport_and_runtime(
 )> {
     let http_client =
```
pakrym-oai, line 605: would've been very nice to reuse default_client we use elsewhere, it sets all the right headers.

pakrym-oai: unrelated, sorry

codex-rs/rmcp-client/src/rmcp_client.rs
```diff
@@ -358,6 +361,12 @@ impl RmcpClient {
             }
         };
 
+        if let Some(runtime) = &oauth_persistor
```
pakrym-oai, line 364: nit: oauth_persistor is a bad name for this type at this point. McpAuthManager maybe

codex-rs/rmcp-client/src/oauth.rs
```diff
@@ -350,28 +371,220 @@ impl OAuthPersistor {
...
+            match credentials.as_ref() {
+                Some(credentials) => {
+                    self.inner
+                        .runtime_credentials
+                        .save(StoredCredentials {
+                            client_id: credentials.client_id.clone(),
+                            token_response: Some(credentials.token_response.0.clone()),
+                        })
+                        .await?;
+                    guard
```
pakrym-oai, line 470: Does client id ever change on refresh?

codex-rs/rmcp-client/src/oauth.rs
```diff
@@ -350,28 +371,220 @@ impl OAuthPersistor {
...
+    store_mode: OAuthCredentialsStoreMode,
+) -> Option<StoredOAuthTokens> {
+    match load_oauth_tokens_for_guarded_refresh(server_name, url, store_mode) {
+        GuardedRefreshPersistedCredentials::Loaded(Some(credentials)) => Some(credentials),
+        GuardedRefreshPersistedCredentials::Loaded(None)
+        | GuardedRefreshPersistedCredentials::ReloadFailed => None,
+    }
+}
+
+fn load_oauth_tokens_for_guarded_refresh_with_keyring_fallback<K: KeyringStore>(
```
pakrym-oai, line 531: can we avoid a separate set of reading method for guarded refresh and reuse existing read methods?

codex-rs/rmcp-client/src/oauth.rs
```diff
@@ -521,6 +734,61 @@ fn token_needs_refresh(expires_at: Option<u64>) -> bool {
...
+            GuardedRefreshOutcome::ReloadedNoChange
+        }
+        Some(persisted_credentials) => {
+            GuardedRefreshOutcome::ReloadedChanged(persisted_credentials)
+        }
+        None => GuardedRefreshOutcome::MissingOrInvalid,
+    }
+}
+
+fn oauth_tokens_equal_for_refresh(
```
pakrym-oai, line 757: do we need this much logic? can we just compare the access token?

codex-rs/rmcp-client/src/oauth.rs
```diff
@@ -273,15 +276,32 @@ struct OAuthPersistorInner {
...
+enum GuardedRefreshOutcome {
+    NoAction,
+    ReloadedChanged(StoredOAuthTokens),
+    ReloadedNoChange,
+    MissingOrInvalid,
+    ReloadFailed,
+}
+
+#[derive(Debug, PartialEq)]
+enum GuardedRefreshPersistedCredentials {
```
pakrym-oai, line 294: can this be `Option<StoredOAuthTokens>` ?

codex-rs/rmcp-client/src/oauth.rs
```diff
@@ -273,15 +276,32 @@ struct OAuthPersistorInner {
     server_name: String,
     url: String,
     authorization_manager: Arc<Mutex<AuthorizationManager>>,
+    runtime_credentials: InMemoryCredentialStore,
```
pakrym-oai, line 279: unfortunate that we have multiple creadential storages to keep in sync. Any chance we can reuse runtime_credentials as last_credentials?

codex-rs/rmcp-client/src/oauth.rs
```diff
@@ -350,28 +371,194 @@ impl OAuthPersistor {
...
+                            "failed to refresh OAuth tokens for server {}",
+                            self.inner.server_name
+                        )
+                    })?;
+                }
+
+                self.persist_if_needed().await
+            }
+            GuardedRefreshOutcome::MissingOrInvalid => self.apply_runtime_credentials(None).await,
+            GuardedRefreshOutcome::ReloadFailed => Ok(()),
```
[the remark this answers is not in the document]
pakrym-oai: Maybe at least log?

pakrym-oai (commented):
The core logic looks good but I wonder if we can collapse reader methods.

PR #12823, opened 2026-02-25
Enforce user input length cap

etraut-openai:
Currently there is no bound on the length of a user message submitted in the TUI or through the app server interface. That means users can paste many megabytes of text, which can lead to bad performance, hangs, and crashes. In extreme cases, it can lead to a [kernel panic](https://github.com/openai/codex/issues/12323).

This PR limits the length of a user input to 2**20 (about 1M) characters. This value was chosen because it fills the entire context window on the latest models, so accepting longer inputs wouldn't make sense anyway.

Summary
- add a shared `MAX_USER_INPUT_TEXT_CHARS` constant in codex-protocol and surface it in TUI and app server code
- block oversized submissions in the TUI submit flow and emit error history cells when validation fails
- reject heavy app-server requests with JSON-RPC `-32602` and structured `input_too_large` data, plus document the behavior

Testing
- ran the IDE extension with this change and verified that when I attempt to paste a user message that's several MB long, it correctly reports an error instead of crashing or making my computer hot.


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -300,6 +287,7 @@ use crate::thread_state::ThreadStateManager;
 
 const THREAD_LIST_DEFAULT_LIMIT: usize = 25;
 const THREAD_LIST_MAX_LIMIT: usize = 100;
+const INPUT_TOO_LARGE_ERROR_CODE: &str = "input_too_large";
```
owenlin0: feels like `app-server/src/error_code.rs` is a better place to define this

codex-rs/app-server/tests/suite/v2/turn_start.rs
```diff
@@ -59,6 +61,7 @@ const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs
 const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
 const TEST_ORIGINATOR: &str = "codex_vscode";
 const LOCAL_PRAGMATIC_TEMPLATE: &str = "You are a deeply pragmatic, effective software engineer.";
+const INVALID_PARAMS_ERROR_CODE: i64 = -32602;
```
owenlin0: should this just import from the error_code file?

codex-rs/app-server/tests/suite/v2/turn_steer.rs
```diff
@@ -17,10 +17,12 @@ use codex_app_server_protocol::TurnStartResponse;
 use codex_app_server_protocol::TurnSteerParams;
 use codex_app_server_protocol::TurnSteerResponse;
 use codex_app_server_protocol::UserInput as V2UserInput;
+use codex_protocol::user_input::MAX_USER_INPUT_TEXT_CHARS;
 use tempfile::TempDir;
 use tokio::time::timeout;
 
 const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+const INVALID_PARAMS_ERROR_CODE: i64 = -32602;
```
owenlin0: import

codex-rs/app-server/tests/suite/output_schema.rs
```diff
@@ -21,6 +23,7 @@ use tempfile::TempDir;
 use tokio::time::timeout;
 
 const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+const INVALID_PARAMS_ERROR_CODE: i64 = -32602;
```
owenlin0: import

codex-rs/protocol/src/user_input.rs
```diff
@@ -3,6 +3,8 @@ use serde::Deserialize;
 use serde::Serialize;
 use ts_rs::TS;
 
+pub const MAX_USER_INPUT_TEXT_CHARS: usize = 1 << 20;
```
owenlin0, line 7: maybe worth documenting why we chose this number? apparently we approximate 4 bytes == 1 token here: https://github.com/openai/codex/blob/main/codex-rs/core/src/truncate.rs#L10

so 1M chars is actually 25% of a 1M token context window.

I think being conservative is fine since we want enough context window for the agent to do work too.

codex-rs/app-server/README.md
```diff
@@ -127,19 +127,14 @@ Example with notification opt-out:
...
 - `thread/rollback` — drop the last N turns from the agent’s in-memory context and persist a rollback marker in the rollout so future resumes see the pruned history; returns the updated `thread` (with `turns` populated) on success.
-- `turn/start` — add user input to a thread and begin Codex generation; responds with the initial `turn` object and streams `turn/started`, `item/*`, and `turn/completed` notifications. For `collaborationMode`, `settings.developer_instructions: null` means "use built-in instructions for the selected mode".
-- `turn/steer` — add user input to an already in-flight turn without starting a new turn; returns the active `turnId` that accepted the input.
+- `turn/start` — add user input to a thread and begin Codex generation; responds with the initial `turn` object and streams `turn/started`, `item/*`, and `turn/completed` notifications. For `collaborationMode`, `settings.developer_instructions: null` means "use built-in instructions for the selected mode". Requests whose combined text input exceeds `1,048,576` characters are rejected before a turn starts with JSON-RPC `invalid params` (`-32602`) and structured `error.data`.
```
owenlin0: Codex has a tendency to add every little thing to the README, but since this section is an API overview and not an exhaustive API reference I think it's fine to omit IMO. Not a strong opinion if you'd still rather keep.

owenlin0 (approved):
small comments but lgtm

codex-rs/protocol/src/user_input.rs
owenlin0: maybe we should double this actually (i.e. target ~50% of a 1M token limit)? what do you think?

etraut-openai: I think we should start with 1M. Our biggest models today have much smaller context windows than 1M tokens. We can increase it in the future if/when the context windows grow.

PR #12828, opened 2026-02-25
fix(exec): avoid duplicate final output in interactive terminals

pdonizete:
## Summary
Fixes duplicate output in `codex exec` interactive runs.

## Problem
In non-interactive exec mode, the final assistant message was shown twice in a terminal:
- once on stderr during event processing (`AgentMessage`)
- once again on stdout in `print_final_output`

## Change
- Added a guard so final message is printed to stdout **only when stdout is not a terminal** (pipes/CI use-cases).
- Kept interactive terminal behavior single-print (no duplicate visible output).
- Added unit tests for the new output decision helper.

## Why this is safe
- Scripts that consume stdout still receive the final message.
- Interactive users no longer see duplicated final output.

Fixes #12566.


etraut-openai:

PR #12849, opened 2026-02-26
Add realtime audio device config

aibrahim-oai:
## Summary
- add top-level realtime audio config for microphone and speaker selection
- apply configured devices when starting realtime capture and playback
- keep missing-device behavior on the system default fallback path

## Validation
- just write-config-schema
- cargo test -p codex-core realtime_audio
- cargo test -p codex-tui
- just fix -p codex-core
- just fix -p codex-tui
- just fmt

### DIFF
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -2177,13 +2177,13 @@ impl Config {
                 .chatgpt_base_url
                 .or(cfg.chatgpt_base_url)
                 .unwrap_or("https://chatgpt.com/backend-api/".to_string()),
-            realtime_audio: cfg.realtime.and_then(|realtime| realtime.audio).map_or_else(
-                RealtimeAudioConfig::default,
-                |audio| RealtimeAudioConfig {
+            realtime_audio: cfg
+                .realtime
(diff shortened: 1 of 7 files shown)


codex-rs/core/src/config/mod.rs
```diff
@@ -5971,6 +6009,40 @@ experimental_realtime_ws_backend_prompt = "prompt from config"
         );
         Ok(())
     }
+
+    #[test]
+    fn realtime_audio_loads_from_config_toml() -> std::io::Result<()> {
+        let cfg: ConfigToml = toml::from_str(
+            r#"
+[realtime.audio]
```
pakrym-oai: just `audio`?

pakrym-oai (approved):

codex-rs/core/src/config/mod.rs
aibrahim-oai: makes sense

sayan-oai (approved):

codex-rs/core/src/config/mod.rs
aibrahim-oai: @codex do this change

PR #12864, opened 2026-02-26
feat: load from plugins

xl-openai:
Support loading plugins.

Plugins can now be enabled via [plugins.<name>] in config.toml. They are loaded as first-class entities through PluginsManager, and their default skills/ and .mcp.json contributions are integrated into the existing skills and MCP flows.


codex-rs/core/src/codex.rs
```diff
@@ -335,6 +338,19 @@ impl Codex {
                 err.message
             );
         }
+        for plugin in loaded_plugins
```
pakrym-oai: can this be logged in plugins_manager.plugins_for_config ?

codex-rs/core/src/skills/loader.rs
```diff
@@ -145,20 +150,6 @@ impl fmt::Display for SkillParseError {
 
 impl Error for SkillParseError {}
 
-pub fn load_skills(config: &Config) -> SkillLoadOutcome {
```
xl-openai, line 149: Cleanup of functions not longer in use.

codex-rs/core/src/plugins.rs
```diff
@@ -0,0 +1,675 @@
...
+        if !plugins_feature_enabled_from_stack(&config_layer_stack) {
+            let mut cache = match self.cache.write() {
+                Ok(cache) => cache,
+                Err(err) => err.into_inner(),
+            };
+            *cache = Some(PluginLoadOutcome::default());
+            return PluginLoadOutcome::default();
+        }
+
+        if !force_reload && let Some(outcome) = self.cached_outcome() {
```
sayan-oai: `plugins()` is reading the global config stack reloaded from disk, but `plugins_for_config()` above reads from the session's resolved `config_layer_stack`.

yet both write to/read from the same cache slot.

so `skills_for_cwd()` can reuse plugin roots from a different config stack

codex-rs/core/src/plugins.rs
```diff
@@ -0,0 +1,675 @@
...
+        Ok(plugins) => plugins,
+        Err(err) => {
+            warn!("invalid plugins config: {err}");
+            HashMap::new()
+        }
+    }
+}
+
+fn load_plugin(config_name: String, plugin: &PluginConfig) -> LoadedPlugin {
+    let plugin_root = plugin.path.to_path_buf();
```
sayan-oai: nit/small question: why make the lossy conversion from `AbsolutePathBuf` to `PathBuf`?

sayan-oai (commented):
one main question about different config sources for loading plugins

codex-rs/core/src/plugins.rs
xl-openai: configured_plugins_from_stack currently reads only from the user layer, so the behavior is consistent for now. We should decide whether we actually need per-cwd plugin loading at all.

xl-openai: Added a comment to the function.

sayan-oai: hmm, i dont see a need for per-cwd plugin loading right now

xl-openai: Actually maybe it would be useful. Let me add it back.

sayan-oai (approved):
moving back to per-cwd cache, other than that lgtm

codex-rs/core/src/plugins.rs
```diff
@@ -0,0 +1,627 @@
...
+            let mut cache = match self.cache_by_cwd.write() {
+                Ok(cache) => cache,
+                Err(err) => err.into_inner(),
+            };
+            cache.insert(cwd.to_path_buf(), PluginLoadOutcome::default());
+            return PluginLoadOutcome::default();
+        }
+
+        if !force_reload && let Some(outcome) = self.cached_outcome_for_cwd(cwd) {
+            return outcome;
```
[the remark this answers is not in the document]
xl-openai: Yes but there will be a force_reload endpoint later.

codex-rs/core/src/plugins.rs
```diff
@@ -0,0 +1,627 @@
...
+    let Ok(config_toml) = effective_config.try_into::<ConfigToml>() else {
+        warn!("failed to deserialize config when checking plugin feature flag");
+        return false;
+    };
+    let config_profile = config_toml
+        .get_config_profile(config_toml.profile.clone())
+        .unwrap_or_else(|_| ConfigProfile::default());
+    let features =
+        Features::from_config(&config_toml, &config_profile, FeatureOverrides::default());
+    features.enabled(Feature::Plugins)
```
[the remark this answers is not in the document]
xl-openai: it is ok.

PR #12868, opened 2026-02-26
tui: align pending steers with core acceptance

charley-oai:
## Summary
- submit `Enter` steers immediately while a turn is already running instead of routing them through `queued_user_messages`
- keep those submitted steers visible in the footer as `pending_steers` until core records them as a user message or aborts the turn
- reconcile pending steers on `ItemCompleted(UserMessage)`, not `RawResponseItem`
- emit user-message item lifecycle for leftover pending input at task finish, then remove the TUI `TurnComplete` fallback
- keep `queued_user_messages` for actual queued drafts, rendered below pending steers

## Problem
While the assistant was generating, pressing `Enter` could send the input into `queued_user_messages`. That queue only drains after the turn ends, so ordinary steers behaved like queued drafts instead of landing at the next core sampling boundary.

The first version of this fix also used `RawResponseItem` to decide when a steer had landed. Review feedback was that this is the wrong abstraction for client behavior.

There was also a late edge case in core: if pending steer input was accepted after the final sampling decision but before `TurnComplete`, core would record that user message into history at task finish without emitting `ItemStarted(UserMessage)` / `ItemCompleted(UserMessage)`. TUI had a fallback to paper over that gap locally.

## Approach
- `Enter` during an active turn now submits a normal `Op::UserTurn` immediately
- TUI keeps a local pending-steer preview instead of rendering that user message into history immediately
- when core records the steer as `ItemCompleted(UserMessage)`, TUI matches and removes the corresponding pending preview, then renders the committed user message
- core now emits the same user-message lifecycle when `on_task_finished(...)` drains leftover pending user input, before `TurnComplete`
- with that lifecycle gap closed in core, TUI no longer needs to flush pending steers into history on `TurnComplete`
- if the turn is interrupted, pending steers and queued drafts are both restored into the composer, with pending steers first

## Notes
- `Tab` still uses the real queued-message path
- `queued_user_messages` and `pending_steers` are separate state with separate semantics
- the pending-steer matching key is built directly from `UserInput`
- this removes the new TUI dependency on `RawResponseItem`

## Codex author
`codex resume 019c9259-8b56-7b02-9528-57c81c76c67d`


charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

codex-rs/tui/src/chatwidget/snapshots/codex_tui__chatwidget__tests__deltas_then_same_final_message_are_rendered_snapshot.snap
```diff
@@ -1,5 +0,0 @@
----
-source: tui/src/chatwidget/tests.rs
-expression: combined
----
-• Here is the result.
```
charley-oai, line 5: - This test was specifically exercising AgentMessageDelta* followed by a duplicate legacy EventMsg::AgentMessage
  - That is not the live path we want to preserve anymore
  - Keeping it would just pin stale compatibility behavior into the PR

charley-oai:
@codex review this

charley-oai:
@codex review this

codex-rs/tui/src/chatwidget.rs
```diff
@@ -4298,9 +4340,10 @@ impl ChatWidget {
         match msg {
             EventMsg::SessionConfigured(e) => self.on_session_configured(e),
             EventMsg::ThreadNameUpdated(e) => self.on_thread_name_updated(e),
-            EventMsg::AgentMessage(AgentMessageEvent { message, .. }) => {
+            EventMsg::AgentMessage(AgentMessageEvent { message, .. }) if from_replay => {
                 self.on_agent_message(message)
             }
+            EventMsg::AgentMessage(AgentMessageEvent { .. }) => {}
```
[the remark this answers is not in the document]
charley-oai: Not important

codex-rs/core/src/tasks/mod.rs
```diff
@@ -213,8 +215,22 @@ impl Session {
...
+                    )
+                    .await;
+                } else {
+                    self.record_conversation_items(
+                        turn_context.as_ref(),
+                        std::slice::from_ref(&response_item),
+                    )
+                    .await;
+                }
+            }
```
charley-oai, line 236: Small core bugfix to emit `ItemStarted(UserMessage)`/`ItemCompleted(UserMessage)` for leftover pending user input at task finish (so we can rely on `ItemCompleted(UserMessage)` for core user message acceptance in all cases.

charley-oai:
@codex review this

charley-oai:
@codex review this

codex-rs/tui/src/chatwidget.rs
```diff
@@ -4388,9 +4467,18 @@ impl ChatWidget {
         match msg {
             EventMsg::SessionConfigured(e) => self.on_session_configured(e),
             EventMsg::ThreadNameUpdated(e) => self.on_thread_name_updated(e),
-            EventMsg::AgentMessage(AgentMessageEvent { message, .. }) => {
+            EventMsg::AgentMessage(AgentMessageEvent { .. })
+                if matches!(replay_kind, Some(ReplayKind::ThreadSnapshot))
+                    && !self.is_review_mode => {}
+            EventMsg::AgentMessage(AgentMessageEvent { message, .. })
```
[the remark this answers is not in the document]
charley-oai: 1. ReplayKind::ResumeInitialMessages

  - This is rollout/session resume replay.
  - Legacy AgentMessage is still rendered here because from_replay is true in chatwidget.rs:4473.
  - So older rollout files that only have legacy assistant events are not broken by the thread-snapshot guard.

  2. ReplayKind::ThreadSnapshot

  - This is thread-switch replay from TUI’s in-memory per-thread event buffer in app.rs:933, app.rs:1469, and app.rs:287.
  - That buffer is populated from live Events via app.rs:1119, not from old rollout replay.
  - So “older rollouts” do not go through this path.

charley-oai:
@codex review this

charley-oai:
@codex review this

codex-rs/core/src/tasks/mod.rs
```diff
@@ -213,8 +215,22 @@ impl Session {
...
+                    self.record_user_prompt_and_emit_turn_item(
+                        turn_context.as_ref(),
+                        &user_message.content,
+                        response_item,
+                    )
+                    .await;
+                } else {
+                    self.record_conversation_items(
+                        turn_context.as_ref(),
+                        std::slice::from_ref(&response_item),
```
aibrahim-oai, line 232: will be recording this twice?

charley-oai: Hmm no it's just recorded once?

aibrahim-oai: ah it's if/else

aibrahim-oai: I don't like how the code is looking but I see why. Can we leave a comment here?

aibrahim-oai (approved):

PR #12870, opened 2026-02-26
tui: restore visible line numbers for hidden file links

pash-openai:
we recently changed file linking so the model uses markdown links when it wants something to be clickable.

This works well across the GUI surfaces because they can render markdown cleanly and use the full absolute path in the anchor target.

A previous pass hid the absolute path in the TUI (and only showed the label), but that also meant we could lose useful location info when the model put the line number or range in the anchor target instead of the label.

This follow-up keeps the TUI behavior simple while making local file links feel closer to the old TUI file reference style.

key changes:
- Local markdown file links in the TUI keep the old file-ref feel: code styling, no underline, no visible absolute path.
- If the hidden local anchor target includes a location suffix and the label does not already include one, we append that suffix to the visible label.
- This works for single lines, line/column references, and ranges.
- If the label already includes the location, we leave it alone.
- normal web links keep the old TUI markdown-link behavior

some examples:
- `[foo.rs](/abs/path/foo.rs)` renders as `foo.rs`
- `[foo.rs](/abs/path/foo.rs:45)` renders as `foo.rs:45`
- `[foo.rs](/abs/path/foo.rs:45:3-48:9)` renders as `foo.rs:45:3-48:9`
- `[foo.rs:45](/abs/path/foo.rs:45)` stays `foo.rs:45`
- `[docs](https://example.com/docs)` still renders like a normal web link

how it looks:
Screenshot 2026-02-26 at 9 27 55 AM

### DIFF
--- a/codex-rs/tui/src/markdown_render.rs
+++ b/codex-rs/tui/src/markdown_render.rs
@@ -92,6 +92,7 @@ struct LinkState {
     destination: String,
     show_destination: bool,
     hidden_location_suffix: Option<String>,
+    label_start_span_idx: usize,
     label_styled: bool,
 }
 
@@ -107,16 +108,6 @@ static HASH_LOCATION_SUFFIX_RE: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(r"^L\d+(?:C\d+)?(?:-L\d+(?:C\d+)?)?$").expect("valid hash location regex")
 });
 
-fn hidden_local_link_location_suffix(dest_url: &str) -> Option<String> {
-    if !is_local_path_like_link(dest_url) {
-        return None;
-    }
-
-    parse_hash_location_suffix(dest_url)
-        .or_else(|| parse_colon_location_suffix(dest_url))
-        .map(std::string::ToString::to_string)
-}
-
 fn is_local_path_like_link(dest_url: &str) -> bool {
     dest_url.starts_with("file://")
         || dest_url.starts_with('/')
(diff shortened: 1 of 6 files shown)


codex-rs/tui/src/markdown_render.rs
```diff
@@ -491,20 +533,53 @@ where
...
+                }
                 self.push_span(" (".into());
                 self.push_span(Span::styled(link.destination, self.styles.link));
                 self.push_span(")".into());
+            } else if let Some(location_suffix) = link.hidden_location_suffix.as_deref() {
+                let label_text = self
+                    .current_line_content
+                    .as_ref()
+                    .map(|line| {
+                        line.spans[link.label_start_span_idx..]
```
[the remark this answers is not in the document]
pash-openai: fixed

codex-rs/tui/src/markdown_render.rs
```diff
@@ -89,12 +91,29 @@ pub(crate) fn render_markdown_text_with_width(input: &str, width: Option<usize>)
...
+    hidden_location_suffix: Option<String>,
+    label_start_span_idx: usize,
+    label_styled: bool,
 }
 
 fn should_render_link_destination(dest_url: &str) -> bool {
     !is_local_path_like_link(dest_url)
 }
 
+static COLON_LOCATION_SUFFIX_RE: LazyLock<Regex> =
```
jif-oai, line 104: A good practice for `LazyLock` that can panic is to have a test that just load them and reference the test next to the `LazyLock`.

Simply something like
```
#[test]
fn load_regex() {
  let _colon = COLON_LOCATION_SUFFIX_RE;
  let _hash = HASH_LOCATION_SUFFIX_RE;
}
```

codex-rs/tui/src/markdown_render.rs
```diff
@@ -109,6 +128,64 @@ fn is_local_path_like_link(dest_url: &str) -> bool {
...
+    if !is_local_path_like_link(dest_url) {
+        return None;
+    }
+
+    parse_hash_location_suffix(dest_url)
+        .and_then(normalize_hash_location_suffix)
+        .or_else(|| parse_colon_location_suffix(dest_url).map(std::string::ToString::to_string))
+}
+
+fn extract_location_suffix(text: &str) -> Option<&str> {
```
jif-oai: Codex love to creates tons of very small function. But if this is not for a testing reason, this does not make any sense. Can you inline all of them?

codex-rs/tui/src/markdown_render.rs
```diff
@@ -109,6 +128,64 @@ fn is_local_path_like_link(dest_url: &str) -> bool {
...
+    HASH_LOCATION_SUFFIX_RE
+        .is_match(fragment)
+        .then_some(text.get(text.len() - fragment.len() - 1..)?)
+}
+
+fn parse_colon_location_suffix(text: &str) -> Option<&str> {
+    COLON_LOCATION_SUFFIX_RE.find(text).map(|m| m.as_str())
+}
+
+fn normalize_hash_location_suffix(suffix: &str) -> Option<String> {
```
jif-oai: I think this code might be useful for others so this would be relevant to put it somewhere in `codex-utils-$`

jif-oai (approved):

codex-rs/tui/src/markdown_render.rs
```diff
@@ -491,20 +568,55 @@ where
...
+        let show_destination = should_render_link_destination(&dest_url);
+        let label_styled = !show_destination;
+        let label_start_span_idx = self
+            .current_line_content
+            .as_ref()
+            .map(|line| line.spans.len())
+            .unwrap_or(0);
+        if label_styled {
+            self.push_inline_style(self.styles.code);
+        }
```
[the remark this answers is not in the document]
pash-openai: web link labels were already plain text and only the appended destination was styled

PR #12918, opened 2026-02-26
Skip history metadata scan for subagents

daveaitel-openai:
Summary
- Skip `history_metadata` scanning when spawning subagents to avoid expensive per-spawn history scans.
- Keeps behavior unchanged for normal sessions.

Testing
  - `cd codex-rs && cargo test -p codex-core` 
  -  Failing in this environment (pre-existing and I don't think something I did?):   
  -  `suite::cli_stream::responses_mode_stream_cli` (SIGKILL + OTEL export error to http://localhost:14318/v1/logs)
  - `suite::grep_files::grep_files_tool_collects_matches` (unsupported call: grep_files)   
  -  `suite::grep_files::grep_files_tool_reports_empty_results` (unsupported call: grep_files)

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -1124,7 +1124,7 @@ impl Session {
         //
         // - initialize RolloutRecorder with new or resumed session info
         // - perform default shell discovery
-        // - load history metadata
+        // - load history metadata (skipped for subagents)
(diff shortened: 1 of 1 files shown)


jif-oai (approved):
lgtm
We should in the end drop this file or limit its size...

PR #12936, opened 2026-02-26
Add a background job to refresh the requirements local cache

alexsong-oai:
- Update the cloud requirements cache TTL to 30 minutes.
- Add a background job to refresh the cache every 5 minutes.
- Ensure there is only one refresh job per process.

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -186,7 +186,6 @@ use codex_arg0::Arg0DispatchPaths;
 use codex_backend_client::Client as BackendClient;
 use codex_chatgpt::connectors;
 use codex_cloud_requirements::cloud_requirements_loader;
-use codex_cloud_requirements::stop_cloud_requirements_refresher;
 use codex_core::AuthManager;
 use codex_core::CodexAuth;
 use codex_core::CodexThread;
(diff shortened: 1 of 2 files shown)


codex-rs/cloud-requirements/src/lib.rs
```diff
@@ -325,6 +345,51 @@ impl CloudRequirementsService {
...
+                .await
+                .inspect_err(|_| {
+                    tracing::warn!(
+                        "Timed out refreshing cloud requirements cache from remote; keeping existing cache"
+                    );
+                });
+        }
+    }
+
+    async fn refresh_cache(&self) {
```
xl-openai: can we just use fetch_with_timeout or somehow dry the logic a little bit?

alexsong-oai: this fn already reuses fetch_with_retries for the core fetch logic. The rest is just deriving the required params.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -1451,6 +1452,7 @@ impl CodexMessageProcessor {
                 data: None,
             });
         }
+        stop_cloud_requirements_refresher();
```
xl-openai: Can the refresh job detect this and exit on its own (i.e., be more self-maintaining)?

xl-openai: maybe just quit the refresh_cache_in_background loop once `self.auth_manager.auth()` failed?

xl-openai (approved):

PR #12943, opened 2026-02-26
Codex/winget auto update

iceweasel-oai:
Publish CLI releases to winget.

Uses https://github.com/vedantmgoyal9/winget-releaser to greatly reduce boilerplate needed to create winget-pkgs manifets


.github/workflows/rust-release.yml
```diff
@@ -639,6 +635,25 @@ jobs:
             exit "${publish_status}"
           done
 
+  winget:
+    name: winget
+    needs: release
+    if: ${{ !contains(needs.release.outputs.version, '-') }}
```
bolinfest, line 643: I assume this is meant to exclude alpha releases? (Or more specifically: it should only allow mainline releases?) Maybe add a comment to that effect?

.github/workflows/rust-release.yml
```diff
@@ -639,6 +635,25 @@ jobs:
...
+    runs-on: ubuntu-latest
```
bolinfest, line 646: Oh nice: we don't need a Windows machine for this?

bolinfest (approved):

.github/workflows/rust-release.yml
iceweasel-oai: correct. I added a comment.

.github/workflows/rust-release.yml
iceweasel-oai: nope - that community action basically mimics the previous winget version (which I created manually) and updates the version. Also added a comment explaining this

PR #12947, opened 2026-02-26
feat: add local date/timezone to turn environment context

celia-oai:
## Summary

  This PR includes the session's local date and timezone in the model-visible environment context and persists that data in `TurnContextItem`.

  ## What changed
  - captures the current local date and IANA timezone when building a turn context, with a UTC fallback if the timezone lookup fails
  - includes current_date and timezone in the serialized <environment_context> payload
  - stores those fields on TurnContextItem so they survive rollout/history handling, subagent review threads, and resume flows
  - treats date/timezone changes as environment updates, so prompt caching and context refresh logic do not silently reuse stale time context
  - updates tests to validate the new environment fields without depending on a single hardcoded environment-context string

## test

built a local build and saw it in the rollout file:
```
{"timestamp":"2026-02-26T21:39:50.737Z","type":"response_item","payload":{"type":"message","role":"user","content":[{"type":"input_text","text":"<environment_context>\n  <shell>zsh</shell>\n  <current_date>2026-02-26</current_date>\n  <timezone>America/Los_Angeles</timezone>\n</environment_context>"}]}}
```

### DIFF
--- a/codex-rs/Cargo.toml
+++ b/codex-rs/Cargo.toml
@@ -186,6 +186,7 @@ icu_locale_core = "2.1"
 icu_provider = { version = "2.1", features = ["sync"] }
 ignore = "0.4.23"
 image = { version = "^0.25.9", default-features = false }
+iana-time-zone = "0.1.64"
 include_dir = "0.7.4"
 indexmap = "2.12.0"
 insta = "1.46.3"
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -53,6 +53,8 @@ use crate::util::error_or_panic;
 use crate::ws_version_from_features;
 use async_channel::Receiver;
 use async_channel::Sender;
+use chrono::Local;
+use chrono::Utc;
 use codex_hooks::HookEvent;
 use codex_hooks::HookEventAfterAgent;
 use codex_hooks::HookPayload;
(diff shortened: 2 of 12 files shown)


codex-rs/Cargo.toml
```diff
@@ -186,6 +186,7 @@ icu_locale_core = "2.1"
 icu_provider = { version = "2.1", features = ["sync"] }
 ignore = "0.4.23"
 image = { version = "^0.25.9", default-features = false }
+iana-time-zone = "0.1.64"
```
pakrym-oai, line 189: is this a huge dep?

celia-oai: codex says:

> • No. In this workspace it is not a huge dependency.
> 
>   iana-time-zone = "0.1.64" is a small platform abstraction crate for getting the local IANA time zone. Also, that manifest entry is a semver range, so the lockfile currently resolves it to 0.1.65, not exactly 0.1.64.
> 
>   What matters here:
> 
>   - On this macOS workspace, cargo tree -p iana-time-zone shows it resolves to just core-foundation-sys.
>   - The lockfile lists Android/wasm/Windows helpers too, but those are target-specific, not all compiled on this platform.
>   - It is already in the graph via chrono, so in this repo the incremental cost is effectively negligible.

codex-rs/core/src/codex.rs
```diff
@@ -7489,6 +7516,12 @@ mod tests {
         session
             .record_context_updates_and_set_reference_context_item(&turn_context, None)
             .await;
+        let subsequent_update_items = session.build_settings_update_items(
+            Some(&turn_context.to_turn_context_item()),
+            None,
+            &turn_context,
+        );
+        expected.extend(subsequent_update_items);
```
charley-oai: We should not be building settings update items outside of `record_context_updates_and_set_reference_context_item`, the whole point of that function was to consolidate all the various context state update injections. (Or maybe I'm misunderstanding the reason for this?)

celia-oai: updated

charley-oai (approved):

PR #12948, opened 2026-02-26
Support multimodal custom tool outputs

fjord-oai:
## Summary

This changes `custom_tool_call_output` to use the same output payload shape as `function_call_output`, so freeform tools can return either plain text or structured content items.

The main goal is to let `js_repl` return image content from nested `view_image` calls in its own `custom_tool_call_output`, instead of relying on a separate injected message.

## What changed

- Changed `custom_tool_call_output.output` from `string` to `FunctionCallOutputPayload`
- Updated freeform tool plumbing to preserve structured output bodies
- Updated `js_repl` to aggregate nested tool content items and attach them to the outer `js_repl` result
- Removed the old `js_repl` special case that injected `view_image` results as a separate pending user image message
- Updated normalization/history/truncation paths to handle multimodal `custom_tool_call_output`
- Regenerated app-server protocol schema artifacts

## Behavior

Direct `view_image` calls still return a `function_call_output` with image content.

When `view_image` is called inside `js_repl`, the outer `js_repl` `custom_tool_call_output` now carries:
- an `input_text` item if the JS produced text output
- one or more `input_image` items from nested tool results

So the nested image result now stays inside the `js_repl` tool output instead of being injected as a separate message.

## Compatibility

This is intended to be backward-compatible for resumed conversations.

Older histories that stored `custom_tool_call_output.output` as a plain string still deserialize correctly, and older histories that used the previous injected-image-message flow also continue to resume.

Added regression coverage for resuming a pre-change rollout containing:
- string-valued `custom_tool_call_output`
- legacy injected image message history

- 👉 `1` https://github.com/openai/codex/pull/12948


codex-rs/core/src/context_manager/history.rs
```diff
@@ -366,10 +366,25 @@ impl ContextManager {
                 }
             }
             ResponseItem::CustomToolCallOutput { call_id, output } => {
-                let truncated = truncate_text(output, policy_with_serialization_budget);
+                let body = match &output.body {
```
pakrym-oai: nit: maybe a method to share the logic with FunctionCallOutput?

codex-rs/core/src/context_manager/normalize.rs
```diff
@@ -59,7 +59,10 @@ pub(crate) fn ensure_call_outputs_present(items: &mut Vec<ResponseItem>) {
                         idx,
                         ResponseItem::CustomToolCallOutput {
                             call_id: call_id.clone(),
-                            output: "aborted".to_string(),
+                            output: FunctionCallOutputPayload {
+                                body: FunctionCallOutputBody::Text("aborted".to_string()),
+                                ..Default::default()
```
pakrym-oai: nit: may as well provide the second property value

codex-rs/core/src/context_manager/normalize.rs
```diff
@@ -263,6 +266,24 @@ pub(crate) fn strip_images_when_unsupported(
                     *content_items = normalized_content_items;
                 }
             }
+            ResponseItem::CustomToolCallOutput { output, .. } => {
```
pakrym-oai: can we merge both branches into one, the type of output is the same

codex-rs/core/src/tools/js_repl/mod.rs
```diff
@@ -1460,8 +1501,13 @@ struct RunToolResult {
 
 #[derive(Debug)]
 enum ExecResultMessage {
-    Ok { output: String },
-    Err { message: String },
+    Ok {
+        output: String,
```
pakrym-oai: consider reusing FunctionCallOutputContentItem directly instead of having both

codex-rs/core/src/client_common.rs
```diff
@@ -86,9 +86,11 @@ fn reserialize_shell_outputs(items: &mut [ResponseItem]) {
         }
         ResponseItem::CustomToolCallOutput { call_id, output } => {
```
pakrym-oai: merge branches?

codex-rs/core/tests/common/responses.rs
```diff
@@ -264,6 +264,19 @@ impl ResponsesRequest {
             .unwrap_or(Value::Null);
         match output {
             Value::String(text) => Some((Some(text), None)),
+            Value::Array(items) => {
+                let text = items
```
pakrym-oai: let's return none if >1 content item or the item isn't text

codex-rs/core/tests/common/test_codex.rs
```diff
@@ -397,9 +397,8 @@ impl TestCodexHarness {
         let bodies = self.request_bodies().await;
         custom_tool_call_output(&bodies, call_id)
             .get("output")
-            .and_then(Value::as_str)
-            .expect("output string")
-            .to_string()
+            .map(output_value_to_text)
+            .unwrap_or_default()
```
pakrym-oai: can you make this panic for none?

codex-rs/core/tests/common/test_codex.rs
```diff
@@ -449,6 +448,23 @@ fn function_call_output<'a>(bodies: &'a [Value], call_id: &str) -> &'a Value {
     panic!("function_call_output {call_id} not found");
 }
 
+fn output_value_to_text(value: &Value) -> String {
```
pakrym-oai: kind of similar to the other helper in responses.

codex-rs/core/tests/suite/client.rs
```diff
@@ -344,6 +344,198 @@ async fn resume_includes_initial_messages_and_sends_prior_items() {
     assert!(pos_environment < pos_new_user);
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn resume_replays_legacy_custom_tool_output_and_image_message() {
```
pakrym-oai: are we testing anything new here? I think we have coverage for messages with images.

pakrym-oai (approved):

codex-rs/core/tests/suite/client.rs
fjord-oai: I simplified the test and added a comment. It’s not generic image-message coverage, it's specifically exercising resume compatibility for the old `js_repl` image rollout shape, calls that used `view_image` were persisted as a string `custom_tool_call_output` plus a separate injected `input_image` message.

codex-rs/core/src/context_manager/history.rs
```diff
@@ -508,7 +516,8 @@ fn image_data_url_estimate_adjustment(item: &ResponseItem) -> (i64, i64) {
                 }
             }
         }
-        ResponseItem::FunctionCallOutput { output, .. } => {
+        ResponseItem::FunctionCallOutput { output, .. }
```
fjord-oai, line 519: @pakrym-oai can you verify that i've updated this history accounting correctly?

fjord-oai: chatted offline, looks good

PR #12968, opened 2026-02-27
notify: include client in legacy hook payload

bolinfest:
## Why

The `notify` hook payload did not identify which Codex client started the turn. That meant downstream notification hooks could not distinguish between completions coming from the TUI and completions coming from app-server clients such as VS Code or Xcode. Now that the Codex App provides its own desktop notifications, it would be nice to be able to filter those out.

This change adds that context without changing the existing payload shape for callers that do not know the client name, and keeps the new end-to-end test cross-platform.

## What changed

- added an optional top-level `client` field to the legacy `notify` JSON payload
- threaded that value through `core` and `hooks`; the internal session and turn state now carries it as `app_server_client_name`
- set the field to `codex-tui` for TUI turns
- captured `initialize.clientInfo.name` in the app server and applied it to subsequent turns before dispatching hooks
- replaced the notify integration test hook with a `python3` script so the test does not rely on Unix shell permissions or `bash`
- documented the new field in `docs/config.md`

## Testing

- `cargo test -p codex-hooks`
- `cargo test -p codex-tui`
- `cargo test -p codex-app-server suite::v2::initialize::turn_start_notify_payload_includes_initialize_client_name -- --exact --nocapture`
- `cargo test -p codex-core` (`src/lib.rs` passed; `core/tests/all.rs` still has unrelated existing failures in this environment)

## Docs

The public config reference on `developers.openai.com/codex` should mention that the legacy `notify` payload may include a top-level `client` field. The TUI reports `codex-tui`, and the app server reports `initialize.clientInfo.name` when it is available.

### DIFF
--- a/codex-rs/tui/src/chatwidget/agent.rs
+++ b/codex-rs/tui/src/chatwidget/agent.rs
@@ -13,6 +13,17 @@ use tokio::sync::mpsc::unbounded_channel;
 use crate::app_event::AppEvent;
 use crate::app_event_sender::AppEventSender;
 
+const TUI_NOTIFY_CLIENT: &str = "codex-tui";
+
+async fn initialize_app_server_client_name(thread: &CodexThread) {
+    if let Err(err) = thread
+        .set_app_server_client_name(Some(TUI_NOTIFY_CLIENT.to_string()))
+        .await
+    {
+        tracing::error!("failed to set app server client name: {err}");
+    }
+}
+
 /// Spawn the agent bootstrapper and op forwarding loop, returning the
 /// `UnboundedSender<Op>` used by the UI to submit operations.
 pub(crate) fn spawn_agent(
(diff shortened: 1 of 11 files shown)


codex-rs/tui/src/chatwidget/agent.rs
```diff
@@ -13,6 +13,17 @@ use tokio::sync::mpsc::unbounded_channel;
 use crate::app_event::AppEvent;
 use crate::app_event_sender::AppEventSender;
 
+const TUI_NOTIFY_CLIENT: &str = "codex-tui";
+
+async fn initialize_app_server_client_name(thread: &CodexThread) {
```
celia-oai, line 18: nit: this name makes it sound like the TUI is built on top of app server. prob worth adding an inline comment to explain that we are special casing TUI surface

celia-oai (approved):

PR #12969, opened 2026-02-27
app-server: source /feedback logs from sqlite at trace level

charley-oai:
## Summary
- write app-server SQLite logs at TRACE level when SQLite is enabled
- source app-server `/feedback` log attachments from SQLite for the requested thread when available
- flush buffered SQLite log writes before `/feedback` queries them so newly emitted events are not lost behind the async inserter
- include same-process threadless SQLite rows in those `/feedback` logs so the attachment matches the process-wide feedback buffer more closely
- keep the existing in-memory ring buffer fallback unchanged, including when the SQLite query returns no rows

## Details
- add a byte-bounded `query_feedback_logs` helper in `codex-state` so `/feedback` does not fetch all rows before truncating
- scope SQLite feedback logs to the requested thread plus threadless rows from the same `process_uuid`
- format exported SQLite feedback lines with the log level prefix to better match the in-memory feedback formatter
- add an explicit `LogDbLayer::flush()` control path and await it in app-server before querying SQLite for feedback logs
- pass optional SQLite log bytes through `codex-feedback` as the `codex-logs.log` attachment override
- leave TUI behavior unchanged apart from the updated `upload_feedback` call signature
- add regression coverage for:
  - newest-within-budget ordering
  - excluding oversized newest rows
  - including same-process threadless rows
  - keeping the newest suffix across mixed thread and threadless rows
  - matching the feedback formatter shape aside from span prefixes
  - falling back to the in-memory snapshot when SQLite returns no logs
  - flushing buffered SQLite rows before querying

## Follow-up
- SQLite feedback exports still do not reproduce span prefixes like `feedback-thread{thread_id=...}:`; there is a `TODO(ccunningham)` in `codex-rs/state/src/log_db.rs` for that follow-up.

## Testing
- `cd codex-rs && cargo test -p codex-state`
- `cd codex-rs && cargo test -p codex-app-server`
- `cd codex-rs && just fmt`

### DIFF
--- a/codex-rs/app-server/Cargo.toml
+++ b/codex-rs/app-server/Cargo.toml
@@ -31,6 +31,7 @@ codex-protocol = { workspace = true }
 codex-app-server-protocol = { workspace = true }
 codex-feedback = { workspace = true }
 codex-rmcp-client = { workspace = true }
+codex-state = { workspace = true }
 codex-utils-absolute-path = { workspace = true }
 codex-utils-json-to-toml = { workspace = true }
 chrono = { workspace = true }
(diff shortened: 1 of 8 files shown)


charley-oai:
@codex review this

charley-oai:
@codex review this

jif-oai (approved):

PR #13021, opened 2026-02-27
Add model availability NUX tooltips

aibrahim-oai:
- override startup tooltips with model availability NUX and persist per-model show counts in config
- stop showing each model after four exposures and fall back to normal tooltips


sayan-oai (approved):
integration tests?

aibrahim-oai:
@codex can you add e2e integration test?

PR #13033, opened 2026-02-27
fix(app-server): make thread/start non-blocking

owenlin0:
Stop `thread/start` from blocking other app-server requests.

Before this change, `thread/start ran` inline on the request loop, so slow startup paths like MCP auth checks could hold up unrelated requests on the same connection, including `thread/loaded/list`. This moves `thread/start` into a background task.

While doing so, it revealed an issue where we were doing nested locking (and there were some race conditions possible that could introduce a "phantom listener"). This PR also refactors the listener/subscription bookkeeping - listener/subscription state is now centralized in `ThreadStateManager` instead of being split across multiple lock domains. That makes late auto-attach on `thread/start` race-safe and avoids reintroducing disconnected clients as phantom subscribers.

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -393,6 +393,23 @@ pub(crate) enum ApiVersion {
     V2,
 }
 
+#[derive(Clone)]
+struct ListenerTaskContext {
+    thread_manager: Arc<ThreadManager>,
+    thread_state_manager: ThreadStateManager,
+    outgoing: Arc<OutgoingMessageSender>,
+    thread_watch_manager: ThreadWatchManager,
+    fallback_model_provider: String,
+    codex_home: PathBuf,
+    single_client_mode: bool,
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+enum EnsureConversationListenerResult {
+    Attached,
+    ConnectionClosed,
+}
+
 pub(crate) struct CodexMessageProcessorArgs {
     pub(crate) auth_manager: Arc<AuthManager>,
     pub(crate) thread_manager: Arc<ThreadManager>,
@@ -2025,7 +2042,7 @@ impl CodexMessageProcessor {
         }
     }
 
-    async fn thread_start(&mut self, request_id: ConnectionRequestId, params: ThreadStartParams) {
+    async fn thread_start(&self, request_id: ConnectionRequestId, params: ThreadStartParams) {
         let ThreadStartParams {
             model,
             model_provider,
@@ -2054,11 +2071,51 @@ impl CodexMessageProcessor {
             personality,
         );
         typesafe_overrides.ephemeral = ephemeral;
-
+        let cli_overrides = self.cli_overrides.clone();
         let cloud_requirements = self.current_cloud_requirements();
+        let listener_task_context = ListenerTaskContext {
+            thread_manager: Arc::clone(&self.thread_manager),
+            thread_state_manager: self.thread_state_manager.clone(),
+            outgoing: Arc::clone(&self.outgoing),
+            thread_watch_manager: self.thread_watch_manager.clone(),
+            fallback_model_provider: self.config.model_provider_id.clone(),
+            codex_home: self.config.codex_home.clone(),
+            single_client_mode: self.single_client_mode,
+        };
+
+        tokio::spawn(async move {
+            Self::thread_start_task(
+                listener_task_context,
+                cli_overrides,
+                cloud_requirements,
+                request_id,
+                config,
+                typesafe_overrides,
+                dynamic_tools,
+                persist_extended_history,
+                service_name,
+                experimental_raw_events,
+            )
+            .await;
+        });
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    async fn thread_start_task(
+        listener_task_context: ListenerTaskContext,
+        cli_overrides: Vec<(String, TomlValue)>,
+        cloud_requirements: CloudRequirementsLoader,
+        request_id: ConnectionRequestId,
+        config_overrides: Option<HashMap<String, serde_json::Value>>,
+        typesafe_overrides: ConfigOverrides,
+        dynamic_tools: Option<Vec<ApiDynamicToolSpec>>,
+        persist_extended_history: bool,
+        service_name: Option<String>,
+        experimental_raw_events: bool,
+    ) {
         let config = match derive_config_from_params(
-            &self.cli_overrides,
-            config,
+            &cli_overrides,
+            config_overrides,
             typesafe_overrides,
             &cloud_requirements,
         )
@@ -2071,7 +2128,10 @@ impl CodexMessageProcessor {
                     message: format!("error deriving config: {err}"),
                     data: None,
                 };
-                self.outgoing.send_error(request_id, error).await;
+                listener_task_context
+                    .outgoing
+                    .send_error(request_id, error)
+                    .await;
                 return;
             }
         };
@@ -2086,7 +2146,10 @@ impl CodexMessageProcessor {
                     message,
                     data: None,
                 };
-                self.outgoing.send_error(request_id, error).await;
+                listener_task_context
+                    .outgoing
+                    .send_error(request_id, error)
+                    .await;
                 return;
             }
             dynamic_tools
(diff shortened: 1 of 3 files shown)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2048,11 +2031,61 @@ impl CodexMessageProcessor {
...
         let cloud_requirements = self.current_cloud_requirements();
+        let outgoing = Arc::clone(&self.outgoing);
+        let thread_manager = Arc::clone(&self.thread_manager);
+        let thread_state_manager = self.thread_state_manager.clone();
+        let thread_watch_manager = self.thread_watch_manager.clone();
+        let fallback_model_provider = self.config.model_provider_id.clone();
+        let codex_home = self.config.codex_home.clone();
+        let single_client_mode = self.single_client_mode;
+
+        tokio::spawn(async move {
```
pakrym-oai, line 2086: drive by: should the dispatcher be spawning these? we are slowly ending up with spawn per method

owenlin0: @pakrym-oai totally, i think we should generalize this. planning to follow up with a more holistic approach

bolinfest: I avoided doing `tokio::spawn()` by default for everything so that if we had an incoming message on the stream where we _knew_ it had to be processed in full before processing new messages in order to enforce certain invariants, then we could do that. It's possible that's not necessary or maybe even realistic (are we confident in such invariants), but that was the original motivation.

codex-rs/app-server/src/thread_state.rs
```diff
@@ -127,49 +127,62 @@ struct SubscriptionState {
...
-                self.thread_ids_by_connection
+                state
+                    .thread_ids_by_connection
                     .remove(&subscription_state.connection_id);
             }
         }
 
-        if let Some(thread_state) = self.thread_states.get(&thread_id) {
+        if let Some(thread_state) = state.thread_states.get(&thread_id) {
             let mut thread_state = thread_state.lock().await;
```
bolinfest: We're taking a new lock while holding `self.state.lock()` which feels like we might put ourselves at risk of deadlock...

Do we have other options here?

owenlin0:
@codex review

codex-rs/app-server/src/thread_state.rs
owenlin0: good catch, looks like there are actually a few places where we do this. will ask codex to take a pass at refactoring them all

owenlin0:
@codex review

owenlin0:
@codex review

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -393,6 +393,23 @@ pub(crate) enum ApiVersion {
     V2,
 }
 
+#[derive(Clone)]
+struct ListenerTaskContext {
+    thread_manager: Arc<ThreadManager>,
+    thread_state_manager: ThreadStateManager,
```
bolinfest, line 399: I'm surprised this is not `Arc<ThreadStateManager>`.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -393,6 +393,23 @@ pub(crate) enum ApiVersion {
...
+    outgoing: Arc<OutgoingMessageSender>,
+    thread_watch_manager: ThreadWatchManager,
```
bolinfest, line 401: This uses `Arc` internally, at least.

codex-rs/app-server/src/thread_state.rs
```diff
@@ -208,141 +263,209 @@ impl ThreadStateManager {
...
-        !thread_state
-            .lock()
-            .await
-            .subscribed_connection_ids()
-            .is_empty()
+        let state = self.state.lock().await;
+        state
+            .connection_ids_by_thread
+            .get(&thread_id)
+            .is_some_and(|connection_ids| !connection_ids.is_empty())
```
bolinfest: FYI, if you make this a one-liner, then the lock for `state` automatically gets dropped at the end of the statement.

codex-rs/app-server/src/thread_state.rs
```diff
@@ -208,141 +263,209 @@ impl ThreadStateManager {
...
+                continue;
+            };
+            let thread_state = thread_state.lock().await;
+            if no_subscribers {
+                tracing::debug!(
+                    thread_id = %thread_id,
+                    connection_id = ?connection_id,
+                    listener_generation = thread_state.listener_generation,
+                    "retaining thread listener after connection disconnect left zero subscribers"
+                );
```
bolinfest: I would try to avoid doing I/O while holding a lock and it appears you don't even need to take the lock when `!no_subscribers` is `true`, so consider something like:

```suggestion
for (thread_id, no_subscribers, thread_state) in thread_states {
    if !no_subscribers { continue; }
    let Some(thread_state) = thread_state else { continue; };
    let listener_generation = thread_state.lock().await.listener_generation;
    tracing::debug!(/* ... */);
}
```

codex-rs/app-server/src/thread_state.rs
```diff
@@ -136,54 +123,108 @@ struct SubscriptionState {
 }
 
 #[derive(Default)]
-pub(crate) struct ThreadStateManager {
+struct ThreadStateManagerInner {
+    live_connections: HashSet<ConnectionId>,
     thread_states: HashMap<ThreadId, Arc<Mutex<ThreadState>>>,
```
bolinfest: Should `thread_states` and `connection_ids_by_thread` share one `HashMap` where the key is `ThreadId` and the value is a struct containing `Arc<Mutex<ThreadState>>` and `HashSet<ConnectionId>`? My gut feeling is that it would make some things more atomic, but you tell me.

codex-rs/app-server/src/codex_message_processor.rs
bolinfest: Oh, I see the `Arc` is internal as a result of this PR.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2123,30 +2186,28 @@ impl CodexMessageProcessor {
                     session_configured.rollout_path.clone(),
                 );
 
-                // Auto-attach a thread listener when starting a thread.
```
bolinfest, line 2126: Should we keep the first line of this comment?

codex-rs/app-server/src/codex_message_processor.rs
bolinfest, line 2560: Currently, `CodexMessageProcessor::process_request()` processes all of these `thread_` messages sequentially.

Are any of these operations potentially long-running such that they should be dispatched onto a tokio task? Or if we did that, do we need processing queues keyed by `thread_id`?

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -6417,12 +6579,13 @@ impl CodexMessageProcessor {
...
-        let fallback_model_provider = self.config.model_provider_id.clone();
-        let single_client_mode = self.single_client_mode;
-        let codex_home = self.config.codex_home.clone();
+        let outgoing_for_task = Arc::clone(&listener_task_context.outgoing);
+        let thread_manager = Arc::clone(&listener_task_context.thread_manager);
+        let thread_state_manager = listener_task_context.thread_state_manager.clone();
+        let thread_watch_manager = listener_task_context.thread_watch_manager.clone();
+        let fallback_model_provider = listener_task_context.fallback_model_provider.clone();
+        let codex_home = listener_task_context.codex_home.clone();
+        let single_client_mode = listener_task_context.single_client_mode;
```
bolinfest: It might be easier to follow with destructuring:

```rust
let ListenerTaskContext {
  outgoing,
  // ...
} = listener_task_context;
```

bolinfest (approved):
Approved with comments!

codex-rs/app-server/src/thread_state.rs
owenlin0: good call, agreed! updating

codex-rs/app-server/src/codex_message_processor.rs
owenlin0: Good question, I think this needs a closer look. We probably do want a processing queue for each thread for maximum app-server throughput 🤔

PR #13050, opened 2026-02-27
Add under-development original-resolution view_image support

fjord-oai:
## Summary

Add original-resolution support for `view_image` behind the under-development `view_image_original_resolution` feature flag.

When the flag is enabled and the target model is `gpt-5.3-codex` or newer, `view_image` now preserves original PNG/JPEG/WebP bytes and sends `detail: "original"` to the Responses API instead of using the legacy resize/compress path.

## What changed

- Added `view_image_original_resolution` as an under-development feature flag.
- Added `ImageDetail` to the protocol models and support for serializing `detail: "original"` on tool-returned images.
- Added `PromptImageMode::Original` to `codex-utils-image`.
  - Preserves original PNG/JPEG/WebP bytes.
  - Keeps legacy behavior for the resize path.
- Updated `view_image` to:
  - use the shared `local_image_content_items_with_label_number(...)` helper in both code paths
  - select original-resolution mode only when:
    - the feature flag is enabled, and
    - the model slug parses as `gpt-5.3-codex` or newer
- Kept local user image attachments on the existing resize path; this change is specific to `view_image`.
- Updated history/image accounting so only `detail: "original"` images use the docs-based GPT-5 image cost calculation; legacy images still use the old fixed estimate.
- Added JS REPL guidance, gated on the same feature flag, to prefer JPEG at 85% quality unless lossless is required, while still allowing other formats when explicitly requested.
- Updated tests and helper code that construct `FunctionCallOutputContentItem::InputImage` to carry the new `detail` field.

### Feature off
- `view_image` keeps the existing resize/re-encode behavior.
- History estimation keeps the existing fixed-cost heuristic.

### Feature on + `gpt-5.3-codex+`
- `view_image` sends original-resolution images with `detail: "original"`.
- PNG/JPEG/WebP source bytes are preserved when possible.
- History estimation uses the GPT-5 docs-based image-cost calculation for those `detail: "original"` images.

- 👉 `1` https://github.com/openai/codex/pull/13050
- ⏳ `2` https://github.com/openai/codex/pull/13331
- ⏳ `3` https://github.com/openai/codex/pull/13049


codex-rs/core/Cargo.toml
```diff
@@ -62,6 +63,7 @@ eventsource-stream = { workspace = true }
 futures = { workspace = true }
 http = { workspace = true }
 iana-time-zone = { workspace = true }
+image = { workspace = true, features = ["jpeg", "png"] }
```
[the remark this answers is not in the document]
fjord-oai: fixed

codex-rs/core/src/tools/handlers/view_image.rs
```diff
@@ -14,19 +19,44 @@ use crate::tools::context::ToolPayload;
...
 #[derive(Deserialize)]
 struct ViewImageArgs {
     path: String,
 }
 
+fn supports_original_resolution_model(slug: &str) -> bool {
+    // Match `gpt-X.Y...` model slugs and enable original-resolution images for
+    // GPT models at version 5.3-codex or newer.
+    let Some(version_suffix) = slug.strip_prefix("gpt-") else {
+        return false;
```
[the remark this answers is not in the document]
fjord-oai: Fixed. @codex review again

codex-rs/core/src/context_manager/history.rs
```diff
@@ -489,47 +497,72 @@ fn base64_data_url_payload_len(url: &str) -> Option<usize> {
     if !has_base64_marker {
         return None;
     }
-    Some(payload.len())
+    Some(payload)
+}
+
+fn base64_data_url_payload_len(url: &str) -> Option<usize> {
+    parse_base64_image_data_url(url).map(str::len)
```
pakrym-oai: nit: might as well inline

codex-rs/core/src/context_manager/history.rs
```diff
@@ -489,47 +497,72 @@ fn base64_data_url_payload_len(url: &str) -> Option<usize> {
...
+}
+
+fn estimate_original_image_bytes(image_url: &str) -> Option<i64> {
```
pakrym-oai, line 515: I don't know how often we call the estimator method but assume that it's callers don't expect it to be this expensive.

codex-rs/core/src/context_manager/history_tests.rs
```diff
@@ -1465,7 +1484,73 @@ fn multiple_inline_images_apply_multiple_fixed_costs() {
...
+        call_id: "call-original".to_string(),
+        output: FunctionCallOutputPayload::from_content_items(vec![
+            FunctionCallOutputContentItem::InputImage {
+                image_url,
+                detail: Some(ImageDetail::Original),
+            },
+        ]),
+    };
+
+    let raw_len = serde_json::to_string(&item).unwrap().len() as i64;
```
pakrym-oai, line 1517: I'm not sure it's a useful tests it it contains a copy of the main logic. We are comparing the logic to itself.

codex-rs/core/src/tools/handlers/view_image.rs
```diff
@@ -14,19 +19,46 @@ use crate::tools::context::ToolPayload;
...
 const VIEW_IMAGE_UNSUPPORTED_MESSAGE: &str =
     "view_image is not allowed because you do not support image inputs";
+const MIN_ORIGINAL_RESOLUTION_MODEL_VERSION: (u32, u32) = (5, 3);
 
 #[derive(Deserialize)]
 struct ViewImageArgs {
     path: String,
 }
 
+fn supports_original_resolution_model(slug: &str) -> bool {
```
pakrym-oai: this value needs to go onto model_info and come from backend as part of model definition

codex-rs/utils/image/src/lib.rs
```diff
@@ -99,6 +125,20 @@ pub fn load_and_resize_to_fit(path: &Path) -> Result<EncodedImage, ImageProcessi
...
 
+fn can_preserve_source_bytes(format: ImageFormat, mode: PromptImageMode) -> bool {
+    match mode {
+        PromptImageMode::ResizeToFit => matches!(format, ImageFormat::Png | ImageFormat::Jpeg),
+        PromptImageMode::Original => {
+            // Public API docs explicitly call out non-animated GIF support only.
+            // Preserve byte-for-byte only for formats we can safely pass through.
+            matches!(
+                format,
+                ImageFormat::Png | ImageFormat::Jpeg | ImageFormat::WebP
```
pakrym-oai: Why is this logic different based on PromptImageMode ?

fjord-oai: I wasn't sure if we wanted to change the previous behavior to support webp also, so kept everything under the feature flag. Should we just add webp support for the general case?

codex-rs/core/src/context_manager/history_tests.rs
fjord-oai: yeah, that's fair. reimplemented with a constant

codex-rs/core/src/context_manager/history.rs
fjord-oai: as per offline convo, added a caching layer

codex-rs/core/src/tools/handlers/view_image.rs
fjord-oai: fixed!

codex-rs/core/src/context_manager/history.rs
```diff
@@ -489,47 +509,71 @@ fn base64_data_url_payload_len(url: &str) -> Option<usize> {
...
     }
-    Some(payload.len())
+    Some(payload)
+}
+
+fn estimate_original_image_bytes(image_url: &str) -> Option<i64> {
+    let key = sha1_digest(image_url.as_bytes());
+    ORIGINAL_IMAGE_ESTIMATE_CACHE.get_or_insert_with(key, || {
+        let payload = parse_base64_image_data_url(image_url)?;
+        let bytes = BASE64_STANDARD.decode(payload).ok()?;
```
pakrym-oai: some of these can use logs

pakrym-oai (approved):

codex-rs/utils/image/src/lib.rs
fjord-oai: chatted offline. will just add webp generally

codex-rs/core/src/context_manager/history.rs
fjord-oai: added logs

PR #13052, opened 2026-02-27
core: reuse parent shell snapshot for thread-spawn subagents

daveaitel-openai:
## Summary
- reuse the parent shell snapshot when spawning/forking/resuming `SessionSource::SubAgent(SubAgentSource::ThreadSpawn { .. })` sessions
- plumb inherited snapshot through `AgentControl -> ThreadManager -> Codex::spawn -> SessionConfiguration`
- skip shell snapshot refresh on cwd updates for thread-spawn subagents so inherited snapshots are not replaced

## Why
- avoids per-subagent shell snapshot creation and cleanup work
- keeps thread-spawn subagents on the parent snapshot path, matching the intended parent/child snapshot model

## Validation
- `just fmt` (in `codex-rs`)
- `cargo test -p codex-core --no-run`
- `cargo test -p codex-core spawn_agent -- --nocapture`
- `cargo test -p codex-core --test all suite::agent_jobs::spawn_agents_on_csv_runs_and_exports`

## Notes
- full `cargo test -p codex-core --test all` was left running separately for broader verification

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -324,6 +324,7 @@ impl Codex {
         dynamic_tools: Vec<DynamicToolSpec>,
         persist_extended_history: bool,
         metrics_service_name: Option<String>,
+        inherited_shell_snapshot: Option<Arc<ShellSnapshot>>,
     ) -> CodexResult<CodexSpawnOk> {
         let (tx_sub, rx_sub) = async_channel::bounded(SUBMISSION_CHANNEL_CAPACITY);
         let (tx_event, rx_event) = async_channel::unbounded();
@@ -447,6 +448,7 @@ impl Codex {
             session_source,
             dynamic_tools,
             persist_extended_history,
+            inherited_shell_snapshot,
         };
 
         // Generate a unique ID for the lifetime of this Codex session.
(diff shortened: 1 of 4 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -2071,6 +2081,13 @@ impl Session {
             return;
         }
 
+        if matches!(
```
jif-oai, line 2084: This means we never refresh the shell snapshot. But what if a sub-agent change cwd for example? In this case we need to be able to refresh it

codex-rs/core/src/thread_manager.rs
```diff
@@ -506,6 +512,7 @@ impl ThreadManagerState {
         agent_control: AgentControl,
         session_source: SessionSource,
         persist_extended_history: bool,
+        inherited_shell_snapshot: Option<Arc<ShellSnapshot>>,
```
jif-oai, line 515: In the end we will have to pack all those things in a struct instead of having to wire those agrs everywhere but we can do as a follow-up

jif-oai (approved):

codex-rs/core/src/thread_manager.rs
daveaitel-openai: 100% agree.

codex-rs/core/src/codex.rs
daveaitel-openai: This is a really good point and I'm not sure how to address it yet.

PR #13053, opened 2026-02-27
fix: merge skill permissions with the current sandbox by default

bolinfest:
## Why

Skill script approvals currently replace the active sandbox with only the permissions declared in the skill's `openai.yaml`. That means attaching a script to a skill can make it _less_ capable than the same script would be without the skill, which is the opposite of what additive skill permissions are supposed to do.

This change makes skill permissions additive by default so a skill can request extra access without discarding the current turn sandbox. For cases where a skill author really does want a fully specified sandbox, it also adds an explicit opt-in for exact replacement semantics.

## What changed

- added `permissions.mode`, with `merge` as the default and `exact` as an opt-in override
- changed skill loading so merge-mode skills keep a normalized `PermissionProfile`, while exact-mode skills compile a standalone `Permissions` object
- changed `zsh-fork` skill escalation to merge the skill's declared permission profile into the current turn sandbox by default
- preserved replacement behavior for `permissions.mode: exact`, including explicit approval copy that calls out the exact-sandbox behavior
- documented the new `permissions` shape in the sample `openai.yaml` reference
- added loader and skill approval coverage for both the default merged behavior and the exact override

## Testing

- added loader tests covering default merge behavior and `permissions.mode: exact`
- added `skill_approval` coverage for merged session approvals and exact-mode replacement
- ran `cargo test -p codex-core`

### DIFF
--- a/codex-rs/app-server-protocol/schema/json/ClientRequest.json
+++ b/codex-rs/app-server-protocol/schema/json/ClientRequest.json
@@ -1618,6 +1618,10 @@
                 "type": "fullAccess"
               }
             },
+            "networkAccess": {
+              "default": false,
+              "type": "boolean"
+            },
             "type": {
               "enum": [
                 "readOnly"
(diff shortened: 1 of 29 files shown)


bolinfest:
https://github.com/openai/codex/pull/13061 fixes this instead.

PR #13063, opened 2026-02-27
Make cloud_requirements fail close

alexsong-oai:
Make it fail-close only for CLI for now
Will extend this for app-server later

### DIFF
--- a/codex-rs/cloud-requirements/src/lib.rs
+++ b/codex-rs/cloud-requirements/src/lib.rs
@@ -304,7 +304,7 @@ impl CloudRequirementsService {
         self.fetch_with_retries(&auth, chatgpt_user_id, account_id)
             .await
             .ok_or_else(|| {
-                let message = "Failed to load your workspace-managed config";
+                let message = "failed to load your workspace-managed config";
                 tracing::error!(
                     path = %self.cache_path.display(),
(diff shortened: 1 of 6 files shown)


codex-rs/cloud-requirements/src/lib.rs
```diff
@@ -4,9 +4,9 @@
 //! from the local filesystem. It only applies to Business (aka Enterprise CBP) or Enterprise ChatGPT
 //! customers.
 //!
-//! Today, fetching is best-effort: on error or timeout, Codex continues without cloud requirements.
-//! We expect to tighten this so that Enterprise ChatGPT customers must successfully fetch these
-//! requirements before Codex will run.
+//! Fetching fails closed for eligible ChatGPT Business and Enterprise accounts. When cloud
```
xl-openai, line 7: Today do we check if it is a ChatGPT Business and Enterprise accounts? If not we should, if so we could change the error message to something more speicfic/meaningful, like:

"Failed to load your workspace-managed config".

alexsong-oai: Yes, we only fetch for ChatGPT Business and Enterprise accounts

xl-openai (approved):

PR #9393, opened 2026-01-16
Persist text elements through TUI input and history

charley-oai:
Continuation of breaking up this PR https://github.com/openai/codex/pull/9116

## Summary
- Thread user text element ranges through TUI/TUI2 input, submission, queueing, and history so placeholders survive resume/edit flows.
- Preserve local image attachments alongside text elements and rehydrate placeholders when restoring drafts.
- Keep model-facing content shapes clean by attaching UI metadata only to user input/events (no API content changes).

## Key Changes
- TUI/TUI2 composer now captures text element ranges, trims them with text edits, and restores them when submission is suppressed.
- User history cells render styled spans for text elements and keep local image paths for future rehydration.
- Initial chat widget bootstraps accept empty `initial_text_elements` to keep initialization uniform.
- Protocol/core helpers updated to tolerate the new InputText field shape without changing payloads sent to the API.

PR #13070, opened 2026-02-27
Keep large-paste placeholders intact during file completion

etraut-openai:
Addresses https://github.com/openai/codex/issues/13040

Fixes a regression in 0.106.0 introduced in https://github.com/openai/codex/pull/9393

Summary
- replace only the active completion range so unrelated text elements (e.g., large-paste placeholders) stay atomic and can still expand
- add a regression test verifying large paste placeholders persist through completions and submit
- could not fetch issue details via GitHub API because network access is disabled in this sandboxed environment


charley-oai (approved):
Thanks

Issue #12170, opened 2026-02-18
VS Code Codex extension repeatedly requires sign-in with ChatGPT auth, but works with API key auth


TheMrAnderson:
### What version of the IDE extension are you using?

0.4.76

### What subscription do you have?

Plus

### Which IDE are you using?

VSCode

### What platform is your computer?

Linux x64 (Ubuntu kernel 6.17.0-14-generic)

### What issue are you seeing?

Problem:
Codex requires ChatGPT sign-in every time VS Code restarts.

Auth cache (~/.codex/auth.json) exists, refreshes correctly, and last_refresh updates.

However, extension logs show:

Request failed error={"code":-32600,"message":"email and plan type are required for chatgpt authentication"} method=account/read

This causes Codex to treat the session as unauthenticated and prompt sign-in again.

Switching to API key auth via OPENAI_API_KEY resolves the issue completely.

Actual behavior:
Codex fails account/read and forces re-authentication every launch.

Workaround:
Using OPENAI_API_KEY works correctly.

### What steps can reproduce the bug?

Steps to reproduce:
1. Sign into Codex via ChatGPT OAuth
2. Restart VS Code
3. Codex prompts for sign-in again
4. Logs show account/read failure

### What is the expected behavior?

Codex should reuse valid OAuth session from ~/.codex/auth.json

TheMrAnderson:
Request failed error={"code":-32600,"message":"email and plan type are required for chatgpt authentication"} method=account/read

Also `~/.codex/auth.json` exists and `last_refresh` updates with each login

etraut-openai:
I think this is a duplicate of #13007, and I have a PR that should fix it: https://github.com/openai/codex/pull/13072.

PR #13072, opened 2026-02-27
Handle missing plan info for ChatGPT accounts

etraut-openai:
Addresses https://github.com/openai/codex/issues/13007 and https://github.com/openai/codex/issues/12170

There are situations where the ChatGPT auth backend might return a JWT that contains no plan information. Most code paths already handle this case well, but the internal implementation of the "account/read" app server call was failing in this case (returning an error rather than properly returning None for the plan).

This resulted in a situation where users needed to log in every time the extension or app started even if they successfully logged in the last time.

Summary
- allow ChatGPT-authenticated accounts to fall back to `AccountPlanType::Unknown` when the token omits the plan claim
- add regression coverage in `app-server/tests/suite/v2/account.rs` to confirm `account/read` returns `plan_type: Unknown` when the claim is absent
- ensure the Rust auth helpers and fixtures treat missing plan claims as Optional and default to `Unknown`

### DIFF
--- a/codex-rs/app-server/tests/suite/v2/account.rs
+++ b/codex-rs/app-server/tests/suite/v2/account.rs
@@ -1227,3 +1227,45 @@ async fn get_account_with_chatgpt() -> Result<()> {
     assert_eq!(received, expected);
     Ok(())
 }
+
+#[tokio::test]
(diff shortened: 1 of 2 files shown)


etraut-openai:
@codex review

owenlin0 (approved):

PR #13079, opened 2026-02-28
app-server: Silence thread status changes caused by thread being created

euroelessar:
Currently we emit `thread/status/changed` with `Idle` status right before sending `thread/started` event (which also has `Idle` status in it).
It feels that there is no point in that as client has no way to know prior state of the thread as it didn't exist yet, so silence these kinds of notifications.

### DIFF
--- a/codex-rs/app-server/src/thread_status.rs
+++ b/codex-rs/app-server/src/thread_status.rs
@@ -91,8 +91,11 @@ impl ThreadWatchManager {
     }
 
     pub(crate) async fn upsert_thread(&self, thread: Thread) {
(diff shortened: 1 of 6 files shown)


codex-rs/app-server/src/thread_status.rs
```diff
@@ -256,6 +259,11 @@ impl ThreadWatchManager {
             .await;
     }
 
+    async fn upsert_thread_with_notification(&self, thread: Thread, emit_notification: bool) {
```
maxj-oai: nit: suggest inlining this into the 2 callers as it's not adding much

maxj-oai (approved):

PR #13082, opened 2026-02-28
Suppress duplicate assistant output on stdout in interactive sessions

etraut-openai:
Addresses #12566

Summary
- stop printing the final assistant message on stdout when the process is running in a terminal so interactive users only see it once
- add a helper that gates the stdout emission and cover it with unit tests


etraut-openai:
@codex review

aibrahim-oai (approved):

PR #13092, opened 2026-02-28
Add request permissions tool

mousseau-oai:
Adds a built-in `request_permissions` tool and wires it through the Codex core, protocol, and app-server layers so a running turn can ask the client for additional permissions instead of relying on a static session policy. 

The new flow emits a `RequestPermissions` event from core, tracks the pending request by call ID, forwards it through app-server v2 as an `item/permissions/requestApproval` request, and resumes the tool call once the client returns an approved subset of the requested permission profile.


codex-rs/core/src/tools/spec.rs
```diff
@@ -968,6 +969,37 @@ fn create_request_user_input_tool(
...
+        },
+    );
+    properties.insert(
+        "permissions".to_string(),
+        JsonSchema::Object {
+            properties: BTreeMap::new(),
+            required: None,
+            additional_properties: Some(true.into()),
+        },
+    );
```
dylan-hurd-oai: How does the model know what to put in this struct?

mousseau-oai: Good point, expanded the tool definition.

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -2121,6 +2174,174 @@ fn mcp_server_elicitation_response_from_client_result(
...
+            id: call_id,
+            response,
+        })
+        .await
+    {
+        error!("failed to submit RequestPermissionsResponse: {err}");
+    }
+}
+
+fn sanitize_granted_permissions(
```
dylan-hurd-oai: [non-blocking] Should this be in `core`? Somewhat similar to `merge_macos_seatbelt_profile_extensions`, but I _think_ this is functionally `intersect_permissions_profile?

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -2121,6 +2174,174 @@ fn mcp_server_elicitation_response_from_client_result(
...
+            Some(CoreNetworkPermissions {
+                enabled: Some(true),
+            }),
+        ) => Some(CoreNetworkPermissions {
+            enabled: Some(true),
+        }),
+        _ => None,
+    };
+
+    let macos = match (requested.macos, granted.macos) {
```
dylan-hurd-oai: [non-blocking] this feels like it should be a helper in codex-rs/core/src/sandboxing/macos_permissions.rs

codex-rs/core/src/sandboxing/mod.rs
```diff
@@ -135,19 +135,29 @@ impl EffectiveSandboxPermissions {
...
+                .map(|paths| normalize_permission_paths(paths, "file_system.read"));
+            let write = file_system
+                .write
+                .map(|paths| normalize_permission_paths(paths, "file_system.write"));
+            FileSystemPermissions { read, write }
+        })
+        .filter(|file_system| !file_system.is_empty());
+    let macos = additional_permissions
+        .macos
+        .filter(|macos| macos != &MacOsSeatbeltProfileExtensions::default());
```
dylan-hurd-oai: [blocking] I don't think is strictly correct - what if the current permissions profile has zero macos permissions, and the user wants to request the default permissions?

For example: 

```
pub enum MacOsPreferencesPermission {
    None,
    // IMPORTANT: ReadOnly needs to be the default because it's the
    // security-sensitive default and keeps cf prefs working.
    #[default]
    ReadOnly,
    ReadWrite,
}
```

A user might mistakenly set `MacOsPreferencesPermission::None` without realizing it - it's important that the model be able to request this permission

codex-rs/core/src/tools/spec.rs
```diff
@@ -294,94 +373,25 @@ fn create_approval_parameters(request_permission_enabled: bool) -> BTreeMap<Stri
...
-                                            .to_string(),
-                                    ),
-                                },
-                            )]),
-                            required: None,
-                            additional_properties: Some(false.into()),
-                        },
+                        create_network_permissions_schema(
+                            "Set to true to enable network access for this command.",
+                        ),
```
dylan-hurd-oai: [non-blocking] I see there are subtle differences between the commands but I actually think we can simplify this to one declaration for now. We can try tweaking these later if we think we need to adjust them to help with model behavior.

dylan-hurd-oai (approved):
Some non-blocking comments, would be great to clean these up but excited to get this in!

codex-rs/app-server/src/bespoke_event_handling.rs
mousseau-oai: Migrated to helper.

codex-rs/app-server/src/bespoke_event_handling.rs
mousseau-oai: Also migrated into helper.

codex-rs/core/src/tools/spec.rs
mousseau-oai: Simplified to single definition.

codex-rs/core/src/sandboxing/mod.rs
mousseau-oai: No longer filtering out non-default profiles.

dylan-hurd-oai (approved):

Issue #6649, opened 2025-11-14
MCP Playwright & chrome-devtools doesn't work in Codex 5.1, GPT 5.1 Bug


ArtAndrew:
### What version of Codex is running?

0.58

### What subscription do you have?

Plus Team

### Which model were you using?

gpt 5, codex 5, codex 5.1, gpt 5.1

### What platform is your computer?

Macos

### What issue are you seeing?

• Subject: Issues using MCP Playwright & Chrome DevTools in Codex CLI (“browser already in use”, transport closed)

  Hi Codex team,

  I’m running Codex CLI with MCPs chrome-devtools and playwright enabled, and both browser tools are effectively unusable in this
  environment due to how the browser instances are managed.

  Environment (inside Codex CLI sandbox)

  - OS: Linux (container) as root
  - Sandbox: danger-full-access (FS), approval_policy=never
  - Tools: chrome-devtools-mcp and playwright-mcp (Chromium)
  - Cache dirs: ~/.cache/chrome-devtools-mcp/chrome-profile, ~/.cache/ms-playwright/mcp-chrome

  Chrome DevTools MCP behavior

  - First call: chrome-devtools/new_page with URL http://localhost:3000 returns:

    The browser is already running for /root/.cache/chrome-devtools-mcp/chrome-profile. Use --isolated to run multiple browser instances.
  - Subsequent calls (list_pages, take_snapshot) either return the same “already running” message or:

    `tool call error: tool call failed for 'chrome-devtools/take_snapshot'
  - From inside the agent I can see Chrome processes:
      - npm exec chrome-devtools-mcp@latest ...
      - /opt/google/chrome/chrome --user-data-dir=/root/.cache/chrome-devtools-mcp/chrome-profile ... --remote-debugging-pipe
  - Attempts to clean up:
      - pkill -f chrome-devtools-mcp / pkill -f '/opt/google/chrome/chrome' results in execution error: Sandbox(Signal(15)) from the Codex
        harness; i.e. the agent cannot reliably stop the stale browser/MCP processes.

  So the MCP server/browser can get into a state where it thinks a browser is already bound to that profile, but the agent has no way to
  request a fresh --isolated instance or to reset the profile/process, and eventually the MCP transport closes.

  Playwright MCP behavior

  - browser_install works and reports no tabs.
  - First navigation:
      - browser_navigate to http://localhost:3000 returns:

[...]

PR #13098, opened 2026-02-28
Recover MCP tools after transport closes

swordfish444:
## Summary
This PR makes MCP tool calls self-heal when the underlying MCP transport has closed, instead of requiring a Codex app restart.

It addresses the user-facing failure pattern reported in #6649 (Playwright / Chrome DevTools MCP ending in `Transport closed` and staying wedged).

## What changed
- Added a restartable MCP client wrapper in `McpConnectionManager`.
- On tool-call failure containing `transport closed`, Codex now:
  1. Rebuilds the MCP client for that server.
  2. Re-sends latest sandbox state to the restarted server.
  3. Retries the same tool call once.
- Keeps generation-based stale guards so concurrent failures do not trigger duplicate restarts.
- Added integration regression coverage:
  - New test where a stdio MCP server exits after the first tool call; second call must recover successfully.
- Extended `test_stdio_server` with a test-only `MCP_TEST_EXIT_AFTER_CALL` mode used by the regression test.

## Why this helps browser MCPs
Playwright and Chrome DevTools MCP failures in long-lived sessions often surface as `Transport closed` and remain unrecoverable until app restart. This change converts that into in-session recovery for subsequent calls.

## QA checklist
- [x] `just fmt`
- [x] `just fix -p codex-core`
- [x] `just fix -p codex-rmcp-client`
- [x] `cargo build -p codex-rmcp-client --bin test_stdio_server`
- [x] `cargo test -p codex-core mcp_tool_call_recovers_from_transport_closed`
- [x] `cargo test -p codex-rmcp-client`
- [ ] `cargo test --all-features` (not run yet; can run before merge if requested)

cc @gpeal for visibility, since this overlaps the browser MCP transport thread history.

### DIFF
--- a/codex-rs/core/src/mcp_connection_manager.rs
+++ b/codex-rs/core/src/mcp_connection_manager.rs
@@ -489,6 +489,168 @@ impl AsyncManagedClient {
     }
 }
 
+#[derive(Clone)]
+struct RestartableManagedClient {
+    server_name: String,
+    config: McpServerConfig,
+    store_mode: OAuthCredentialsStoreMode,
+    cancel_token: CancellationToken,
+    tx_event: Sender<Event>,
+    elicitation_requests: ElicitationRequestManager,
+    codex_apps_tools_cache_context: Option<CodexAppsToolsCacheContext>,
+    sandbox_state: Arc<Mutex<SandboxState>>,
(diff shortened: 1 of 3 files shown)


etraut-openai:

PR #13201, opened 2026-03-02
Slash copy osc52 wsl support

won-openai:
This PR is a followup to the /copy feature to support WSL and SSH!


codex-rs/tui/src/clipboard_text.rs
```diff
@@ -1,11 +1,208 @@
...
+#[cfg(all(not(target_os = "android"), unix))]
+use std::fs::OpenOptions;
+#[cfg(not(target_os = "android"))]
+use std::io::IsTerminal;
+#[cfg(not(target_os = "android"))]
+use std::io::Write;
+#[cfg(not(target_os = "android"))]
+use std::io::stdout;
+#[cfg(all(not(target_os = "android"), target_os = "linux"))]
+use std::process::Stdio;
```
joshka-oai, line 31: nit: If there's a lot of per target code, it's worth using the approach of putting that in a module and flagging on the module rather than the individual statements. This is the idiomatic approach used in a bunch of libraries including the rust standard lib. (Just a verbosity thing mostly).

codex-rs/tui/src/clipboard_text.rs
```diff
@@ -1,11 +1,208 @@
...
+
+#[cfg(all(not(target_os = "android"), target_os = "linux"))]
+use crate::clipboard_paste::is_probably_wsl;
+
 #[cfg(not(target_os = "android"))]
 pub fn copy_text_to_clipboard(text: &str) -> Result<(), String> {
```
joshka-oai, line 56: nit: I'd suggest adding some docs here - the fallback order is encoded into the code here and it likely should be documented in a way that means that someone doesn't have to read the code to derive that order (a small list would work).

codex-rs/tui/src/clipboard_text.rs
```diff
@@ -1,11 +1,208 @@
...
+            match copy_via_wsl_clipboard(text) {
+                Ok(()) => return Ok(()),
+                Err(wsl_err) => format!("{native_err}; WSL fallback failed: {wsl_err}"),
+            }
+        } else {
+            native_err
+        };
+
+        if should_try_osc52_fallback(
+            stdout().is_terminal(),
```
joshka-oai: nit: generally avoid bool params to methods as it often leads to worse looking code (it's not so bad here). It might be reasonable to call `stdout().is_terminal()` from within the function itself rather than passing in the bool.

codex-rs/tui/src/clipboard_text.rs
```diff
@@ -1,11 +1,208 @@
...
+            std::env::var_os("DISPLAY").as_deref(),
+            std::env::var_os("WAYLAND_DISPLAY").as_deref(),
+            std::env::var_os("TERM_PROGRAM").as_deref(),
```
joshka-oai: We should probably fold some of this into the terminal detection logic so that we don't have to look at the env variables each time. This could be a followup PR to improve / refactor this.

codex-rs/tui/src/clipboard_text.rs
```diff
@@ -1,11 +1,208 @@
...
+            copy_via_osc52(text)
+                .map_err(|osc_err| format!("{native_err}; OSC 52 fallback failed: {osc_err}"))
+        } else {
+            Err(native_err)
+        }
+    })
+}
+
+#[cfg(not(target_os = "android"))]
+fn copy_via_osc52(text: &str) -> Result<(), String> {
```
joshka-oai, line 90: It's worth documenting this a bit to help anyone looking at what the expectation is - include such things as how it fails, that it opens the tty device. Codex is good at understanding these issues automatically when coding, but someone coming along reading this in the future may not have the same context.

codex-rs/tui/src/clipboard_text.rs
```diff
@@ -1,11 +1,208 @@
...
+    })?;
+    #[cfg(windows)]
+    stdout().write_all(sequence.as_bytes()).map_err(|e| {
+        format!("clipboard unavailable: failed to write OSC 52 escape sequence: {e}")
+    })?;
+    #[cfg(windows)]
+    stdout().flush().map_err(|e| {
+        format!("clipboard unavailable: failed to flush OSC 52 escape sequence: {e}")
+    })?;
+    Ok(())
```
joshka-oai, line 115: This is a good example of where a small module per target makes sense.

```
fn copy_via_osc52(...) {
    #[cfg(unix)]
    unix::copy_via_osc52(...)
    #[cfg(windows)]
    win::copy_via_osc52(...)
}

#[cfg(unix)]
mod unix { 
    use ...;
    fn copy_via_osc52(...) { ... }
}

#[cfg(windows)]
mod win {
    use ...;
    fn copy_via_osc52(...) { ... }
}
```

codex-rs/tui/src/clipboard_text.rs
```diff
@@ -1,11 +1,208 @@
...
+        ])
+        .spawn()
+        .map_err(|e| format!("clipboard unavailable: failed to spawn powershell.exe: {e}"))?;
+
+    child
+        .stdin
+        .take()
+        .ok_or_else(|| "clipboard unavailable: failed to open powershell.exe stdin".to_string())?
+        .write_all(text.as_bytes())
+        .map_err(|e| format!("clipboard unavailable: failed to write to powershell.exe: {e}"))?;
```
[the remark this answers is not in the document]
joshka-oai: This is likely worth investigating more.

joshka-oai (approved):
General approval. I'd suggest throwing codex at the nits / refactoring ideas I put here.
I'd also want to make sure the powershell issue is covered.

PR #13212, opened 2026-03-02
add fast mode toggle

pash-openai:
- add a local Fast mode setting in codex-core (similar to how model id is currently stored on disk locally)
- send `service_tier=priority` on requests when Fast is enabled
- add `/fast` in the TUI and persist it locally
- feature flag


codex-rs/tui/src/bottom_pane/chat_composer.rs
```diff
@@ -2257,16 +2276,9 @@ impl ChatComposer {
         {
             let treat_as_plain_text = input_starts_with_space || name.contains('/');
             if !treat_as_plain_text {
-                let is_builtin = slash_commands::find_builtin_command(
-                    name,
-                    self.collaboration_modes_enabled,
-                    self.connectors_enabled,
-                    self.personality_command_enabled,
-                    self.realtime_conversation_enabled,
```
pash-openai, line 2270: had too many arguments, so had to refactor to adhere to lint rules

codex-rs/core/src/client.rs
```diff
@@ -495,6 +502,23 @@ impl ModelClient {
...
+    /// tier, but invalidate request-level reuse state so the next request becomes a fresh
+    /// `response.create`.
+    ///
+    /// This keeps the handshake win from prewarm without leaking stale warmup request state into
+    /// the first user turn.
+    pub fn reset_prewarm_for_service_tier(&mut self, service_tier: ServiceTier) {
+        if self.service_tier == service_tier {
+            return;
+        }
+        self.service_tier = service_tier;
```
pash-openai: We reuse the prewarmed auth, but don't reuse the prewarmed responses.create request

codex-rs/core/src/features.rs
```diff
@@ -660,6 +662,16 @@ pub const FEATURES: &[FeatureSpec] = &[
         stage: Stage::Stable,
         default_enabled: true,
     },
+    FeatureSpec {
```
pash-openai, line 665: I set this flag to just block the UI so users are not able to turn it on if the flag is off. This is a simple enough gate, and I did not add a second runtime guard to suppress service_tier on the wire, because that would expand the scope of what is meant to be a small, temporary rollout safeguard. 

The tradeoff is that unsupported paths, like manual config edits, are not blocked by this PR - i am okay with this tradeoff

codex-rs/core/src/features.rs
```diff
@@ -660,6 +662,16 @@ pub const FEATURES: &[FeatureSpec] = &[
...
+        id: Feature::FastMode,
+        key: "fast_mode",
+        stage: Stage::Experimental {
```
aibrahim-oai: Let's have it the least value under development to not show to users before we are ready

codex-rs/protocol/src/protocol.rs
```diff
@@ -244,6 +244,10 @@ pub enum Op {
         #[serde(skip_serializing_if = "Option::is_none")]
         summary: Option<ReasoningSummaryConfig>,
 
+        /// Updated service tier preference for future turns.
+        #[serde(skip_serializing_if = "Option::is_none")]
+        service_tier: Option<ServiceTier>,
```
aibrahim-oai: I think we should add it to Op::UserTurn also

codex-rs/core/tests/responses_headers.rs
```diff
@@ -135,6 +136,109 @@ async fn responses_stream_includes_subagent_header_on_review() {
...
+        conversation_id,
+        provider.clone(),
+        session_source,
+        config.model_verbosity,
+        None::<ResponsesWebsocketVersion>,
+        false,
+        false,
+        None,
+    );
+    let mut client_session = client.new_session_with_service_tier(ServiceTier::Fast);
```
aibrahim-oai: I want to see some integration tests w/ and w/o websockets where we have op with Fast/standard also changing on the next turn and assert the expected result.

codex-rs/core/src/client.rs
```diff
@@ -495,6 +502,23 @@ impl ModelClient {
...
 }
 
+impl ModelClientSession {
+    /// Reuse the startup-prewarmed websocket connection when the first real turn changes service
+    /// tier, but invalidate request-level reuse state so the next request becomes a fresh
+    /// `response.create`.
+    ///
+    /// This keeps the handshake win from prewarm without leaking stale warmup request state into
+    /// the first user turn.
+    pub fn reset_prewarm_for_service_tier(&mut self, service_tier: ServiceTier) {
```
aibrahim-oai: @pakrym need your eye on this. I think ideally we would send update item but maybe that's too much hustle.

codex-rs/core/src/client.rs
```diff
@@ -190,6 +191,7 @@ pub struct ModelClient {
 pub struct ModelClientSession {
     client: ModelClient,
     websocket_session: WebsocketSession,
+    service_tier: ServiceTier,
```
pakrym-oai: why is the value here and not a parameter?

codex-rs/core/src/client.rs
pakrym-oai: I don't think this code should exist at all. We compare request bodies to know whether previous request can be reused. Service tier is part of the body.  Things should "just work".

codex-rs/core/src/features.rs
pash-openai: Updated this to `Stage::UnderDevelopment` so it stays off `/experimental` and other user-facing surfaces until we’re ready to expose it.

codex-rs/protocol/src/config_types.rs
```diff
@@ -113,6 +113,30 @@ pub enum WebSearchMode {
...
+pub enum ServiceTier {
+    #[default]
+    Standard,
+    Fast,
+}
+
+impl ServiceTier {
+    pub const fn as_api_service_tier(self) -> Option<&'static str> {
+        match self {
+            Self::Standard => None,
```
pakrym-oai: I'm not a huge fan of this logic. It looks like we have and enum that matches the api but then it has a magical meaning. 

Wonder if we should have Option<ServiceTier> in the config to represent not set.

codex-rs/protocol/src/config_types.rs
```diff
@@ -113,6 +113,30 @@ pub enum WebSearchMode {
...
 }
 
+#[derive(
+    Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq, Display, JsonSchema, TS, Default,
+)]
+#[serde(rename_all = "lowercase")]
+#[strum(serialize_all = "lowercase")]
+pub enum ServiceTier {
+    #[default]
+    Standard,
```
pakrym-oai, line 121: We should have all variants here. Users have been asking for api service tiers since the beginning of time.

pash-openai: Agreed - will add this in as a separate PR when we're ready to support this for API users

codex-rs/tui/src/chatwidget.rs
```diff
@@ -6845,6 +6894,26 @@ impl ChatWidget {
...
+                sandbox_policy: None,
+                windows_sandbox_level: None,
+                model: None,
+                effort: None,
+                summary: None,
+                service_tier: Some(service_tier),
+                collaboration_mode: None,
+                personality: None,
+            }));
+        self.app_event_tx
```
pakrym-oai: three separate events?

codex-rs/tui/src/slash_command.rs
```diff
@@ -88,6 +89,7 @@ impl SlashCommand {
             SlashCommand::MemoryDrop => "DO NOT USE",
             SlashCommand::MemoryUpdate => "DO NOT USE",
             SlashCommand::Model => "choose what model and reasoning effort to use",
+            SlashCommand::Fast => "toggle Fast mode for supported models",
```
pakrym-oai, line 93: should this be conditional?

codex-rs/tui/src/app_event.rs
```diff
@@ -197,6 +201,11 @@ pub(crate) enum AppEvent {
         personality: Personality,
     },
 
+    /// Persist the selected service tier to the appropriate config.
+    PersistServiceTierSelection {
```
pakrym-oai, line 202: is there a reason why update and persist are separate?

codex-rs/core/src/config/mod.rs
```diff
@@ -185,6 +186,9 @@ pub struct Config {
     /// Optional override of model selection.
     pub model: Option<String>,
 
+    /// Service tier preference for new turns.
+    pub service_tier: ServiceTier,
```
pakrym-oai: Consider Optional to signal no value set.

codex-rs/core/tests/suite/agent_websocket.rs
```diff
@@ -244,3 +246,128 @@ async fn websocket_v2_test_codex_shell_chain() -> Result<()> {
...
+            .and_then(Value::as_array)
+            .is_some_and(|items| !items.is_empty())
+    );
+
+    server.shutdown().await;
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn websocket_v2_first_turn_drops_fast_tier_after_startup_prewarm() -> Result<()> {
```
pakrym-oai, line 299: I don't think we need this many tests for the "settings changes" case but don't feel strongly.

pakrym-oai (approved):

codex-rs/core/tests/suite/agent_websocket.rs
pash-openai: Just wanted the confidence after dropping reset_prewarm_for_service_tier

codex-rs/tui/src/chatwidget.rs
pash-openai: now it's 2. matches the way updating model works (though model has 3 events)

codex-rs/tui/src/app_event.rs
pash-openai: immediate local responsiveness vs async/potentially failing persistence -- we do the same thing with updating model.

PR #13218, opened 2026-03-02
Add role-specific subagent nickname overrides

gabec-openai:
## Summary
- add `nickname_candidates` to agent role config
- use role-specific nickname pools for spawned and resumed subagents
- validate and schema-generate the new config surface

## Testing
- `just fmt`
- `just write-config-schema`
- `just fix -p codex-core`
- `cargo test -p codex-core`
- `cargo test`

### DIFF
--- a/codex-rs/core/src/agent/control.rs
+++ b/codex-rs/core/src/agent/control.rs
@@ -1,5 +1,7 @@
 use crate::agent::AgentStatus;
 use crate::agent::guards::Guards;
+use crate::agent::role::DEFAULT_ROLE_NAME;
+use crate::agent::role::resolve_role_config;
 use crate::agent::status::is_final;
 use crate::error::CodexErr;
 use crate::error::Result as CodexResult;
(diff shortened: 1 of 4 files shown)


codex-rs/core/src/config/mod.rs
```diff
@@ -2329,6 +2335,33 @@ impl Config {
         }
     }
 
+    fn validate_agent_role_nickname_candidates(
```
jif-oai: Should we also check for duplicates?

codex-rs/core/src/agent/role.rs
```diff
@@ -100,6 +94,23 @@ pub(crate) async fn apply_role_to_config(
     Ok(())
 }
 
+pub(crate) fn nickname_candidates_for_role<'a>(
+    config: &'a Config,
```
jif-oai, line 137: why playing with lifetime specifiers if you're not async?

codex-rs/core/src/agent/control.rs
```diff
@@ -31,14 +32,36 @@ pub(crate) struct SpawnAgentOptions {
...
+fn agent_nickname_candidates(
+    config: &crate::config::Config,
+    role_name: Option<&str>,
+) -> Vec<String> {
+    nickname_candidates_for_role(config, role_name)
+        .map(|candidates| {
+            candidates
+                .iter()
+                .map(|name| name.trim())
+                .filter(|name| !name.is_empty())
```
jif-oai: this can't happen because of the validation

codex-rs/core/src/agent/control.rs
```diff
@@ -31,14 +32,36 @@ pub(crate) struct SpawnAgentOptions {
...
         .map(str::trim)
         .filter(|name| !name.is_empty())
         .collect()
 }
 
+fn agent_nickname_candidates(
+    config: &crate::config::Config,
+    role_name: Option<&str>,
+) -> Vec<String> {
+    nickname_candidates_for_role(config, role_name)
```
jif-oai: Could we avoid the extra nesting here? would be smoother to read

codex-rs/core/src/config/mod.rs
```diff
@@ -2329,6 +2336,46 @@ impl Config {
         }
     }
 
+    fn normalize_agent_role_nickname_candidates(
```
jif-oai, line 2373: nit: can we do stronger normalization? I would like prevent user from inserting `<`, `&`, ... to prevent XML collision or bad interpretation from the model

Same thing for non-UTF-8

Basically we could accept only alphanum and ` `, `-`, `_` or something like this

jif-oai (approved):
LGTM once my last comment is processed

PR #13275, opened 2026-03-02
fix(network-proxy): reject mismatched host headers

viyatb-oai:
## Summary
- reject plain HTTP absolute-form requests whose Host header does not match the request target authority
- add host/port-aware Host header validation for non-default ports
- add regression coverage for mismatched Host forwarding and validator edge cases


itsmandatory (approved):
LGTM

PR #13276, opened 2026-03-02
start of hooks engine

eternal-openai:
(Experimental)

This PR adds a first MVP for hooks, with SessionStart and Stop

The core design is:

- hooks live in a dedicated engine under codex-rs/hooks
- each hook type has its own event-specific file
- hook execution is synchronous and blocks normal turn progression while running
- matching hooks run in parallel, then their results are aggregated into a normalized HookRunSummary

On the AppServer side, hooks are exposed as operational metadata rather than transcript-native items:

- new live notifications: hook/started, hook/completed
- persisted/replayed hook results live on Turn.hookRuns
- we intentionally did not add hook-specific ThreadItem variants

Hooks messages are not persisted, they remain ephemeral. The context changes they add are (they get appended to the user's prompt)


eternal-openai:

codex-rs/hooks/src/engine/dispatcher.rs
```diff
@@ -0,0 +1,125 @@
...
+
+pub(crate) async fn execute_handlers<T>(
+    shell: &CommandShell,
+    handlers: Vec<ConfiguredHandler>,
+    input_json: String,
+    cwd: &Path,
+    turn_id: Option<String>,
+    parse: fn(&ConfiguredHandler, CommandRunResult, Option<String>) -> ParsedHandler<T>,
+) -> Vec<ParsedHandler<T>> {
+    let results = join_all(
```
jif-oai, line 72: this means you run everything in parallel. So the ordering of the hooks is not respected, even for "Sync" hooks

codex-rs/exec/src/event_processor_with_human_output.rs
```diff
@@ -856,6 +840,8 @@ impl EventProcessor for EventProcessorWithHumanOutput {
             | EventMsg::AgentReasoningRawContentDelta(_)
             | EventMsg::ItemStarted(_)
             | EventMsg::ItemCompleted(_)
+            | EventMsg::HookStarted(_)
+            | EventMsg::HookCompleted(_)
```
jif-oai: I think this should be forwarded. Otherwise a user can't see what happen during failure

codex-rs/exec/src/event_processor_with_human_output.rs
```diff
@@ -937,6 +923,8 @@ impl EventProcessorWithHumanOutput {
                 | EventMsg::AgentReasoningRawContentDelta(_)
                 | EventMsg::ItemStarted(_)
                 | EventMsg::ItemCompleted(_)
+                | EventMsg::HookStarted(_)
+                | EventMsg::HookCompleted(_)
```
jif-oai: same

codex-rs/core/src/codex.rs
```diff
@@ -5177,6 +5204,57 @@ pub(crate) async fn run_turn(
...
+                        if stop_hook_active {
+                            sess.send_event(
+                                &turn_context,
+                                EventMsg::Warning(WarningEvent {
+                                    message: "Stop hook blocked twice in the same turn; ignoring the second block to avoid an infinite loop.".to_string(),
+                                }),
+                            )
+                            .await;
+                        } else {
+                            stop_hook_active = true;
```
jif-oai, line 5708: this thing is sticky. It means that, for example:

1. Turn starts. `stop_hook_active = false`
2. The model produces a final answer candidate. Stop hook runs with `stop_hook_active = false` and do something blocking
3. We sets `stop_hook_active = true` and injects the hook feedback back to the model  
4. The model follows that feedback, runs tools, do his life, and then produces another final answer candidate. 
5. Suppose that answer is still not acceptable for a different reason, and the stop hook wants to block again, for example “tests passed, but now update the changelog”.
6. That second stop-hook run is invoked with `stop_hook_active = true`. So if it blocks, core goes into the “blocked twice in the same turn; ignore the second block” path and game over

codex-rs/app-server-protocol/src/protocol/thread_history.rs
```diff
@@ -166,6 +162,7 @@ impl ThreadHistoryBuilder {
             EventMsg::ExitedReviewMode(payload) => self.handle_exited_review_mode(payload),
             EventMsg::ItemStarted(payload) => self.handle_item_started(payload),
             EventMsg::ItemCompleted(payload) => self.handle_item_completed(payload),
+            EventMsg::HookStarted(_) | EventMsg::HookCompleted(_) => {}
```
jif-oai, line 169: I think we should keep them. This is useful information of the UX at least

codex-rs/hooks/src/engine/dispatcher.rs
```diff
@@ -0,0 +1,125 @@
...
+                (Some(matcher), Some(source)) => regex::Regex::new(matcher)
+                    .map(|regex| regex.is_match(source))
+                    .unwrap_or(false),
+                (None, _) => true,
+                _ => false,
+            },
+            HookEventName::Stop => true,
+        })
+        .filter(|handler| {
+            seen.insert((
```
jif-oai: Here this means you dedup but ignores "matcher". But can't you have 2 times that same hook with different matcher in which case you want to keep both?

codex-rs/hooks/src/events/session_start.rs
```diff
@@ -0,0 +1,322 @@
...
+    pub session_id: ThreadId,
+    pub cwd: PathBuf,
+    pub transcript_path: Option<PathBuf>,
+    pub model: String,
+    pub permission_mode: String,
+    pub source: SessionStartSource,
+}
+
+#[derive(Debug)]
+pub struct SessionStartOutcome {
```
jif-oai, line 44: Where is `suppressOutput`?

codex-rs/hooks/src/events/session_start.rs
```diff
@@ -0,0 +1,322 @@
...
+
+use crate::engine::CommandShell;
+use crate::engine::ConfiguredHandler;
+use crate::engine::command_runner::CommandRunResult;
+use crate::engine::dispatcher;
+use crate::engine::output_parser;
+use crate::schema::SessionStartCommandInput;
+
+#[derive(Debug, Clone, Copy)]
+pub enum SessionStartSource {
```
jif-oai, line 19: Where is clear?

codex-rs/hooks/src/engine/dispatcher.rs
eternal-openai: That's on purpose based on my understanding of the existing spec. I'm unclear what sync hooks are

codex-rs/app-server-protocol/src/protocol/thread_history.rs
eternal-openai: I used to have a mechanism for storing hooks and reloading them in the UIs (via HookRunSummary) but @pakrym-oai recommended that I remove that in favor of potentially doing a unified metadata persistence approach later

codex-rs/hooks/src/events/session_start.rs
eternal-openai: Deferred for a future PR

codex-rs/hooks/src/events/session_start.rs
eternal-openai: Also deferred

codex-rs/core/src/codex.rs
eternal-openai: Yeah, totally agree basically deferred also, but I have already begin work on the followup PR that implements this Stop -> new prompt flow & the stop_hook_active mitigation for infinite loops

codex-rs/hooks/src/engine/dispatcher.rs
eternal-openai: Thank you! Implemented

codex-rs/exec/src/event_processor_with_human_output.rs
eternal-openai: Implemented

codex-rs/exec/src/event_processor_with_human_output.rs
eternal-openai: Implemented

codex-rs/hooks/src/engine/dispatcher.rs
jif-oai: Would be good to make sure we don't miss something here

codex-rs/hooks/src/events/stop.rs
```diff
@@ -0,0 +1,361 @@
...
+                                kind: HookOutputEntryKind::Stop,
+                                text: stop_reason_text,
+                            });
+                        }
+                    } else if parsed.should_block {
+                        status = HookRunStatus::Blocked;
+                        should_block = true;
+                        block_reason = parsed.reason.clone();
+                        block_message_for_model = parsed.reason.clone();
+                        if let Some(reason) = parsed.reason {
```
jif-oai: How does it work if there is no reason? How is the model supposed to know what to do? What do you restart the turn on?

codex-rs/hooks/src/engine/command_runner.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    let mut command = build_command(shell, handler);
+    command
+        .current_dir(cwd)
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .kill_on_drop(true);
+
+    let mut child = match command.spawn() {
```
jif-oai, line 41: Discussion point on sandboxing here:
Hooks run outside of our sandboxing which, I think, come from the fact that a hook is user defined. But `Stop` hooks for example take the assistant message as input. So it means that in practice, a hook can execute something coming from the model. That sounds a bit dangerous to me. Any thoughts on this @bolinfest ?

codex-rs/core/src/codex.rs
jif-oai: Make sure this fix is logged somewhere then :)

codex-rs/hooks/src/events/stop.rs
```diff
@@ -0,0 +1,361 @@
...
+        Some(error) => {
+            status = HookRunStatus::Failed;
+            entries.push(HookOutputEntry {
+                kind: HookOutputEntryKind::Error,
+                text: error.to_string(),
+            });
+        }
+        None => match run_result.exit_code {
+            Some(0) => {
+                if let Some(parsed) = output_parser::parse_stop(&run_result.stdout) {
```
jif-oai: We shouldn't silently ignore if the JSON parsing fails

codex-rs/hooks/src/events/session_start.rs
jif-oai: Log it then please. Linear or whatever :) to make sure we don't loose track

Otherwise, drop it from the schema for now

jif-oai (commented):
Looks better to me. I think we should discuss the sandboxing point

codex-rs/hooks/src/events/stop.rs
sayan-oai: +1

codex-rs/hooks/src/engine/dispatcher.rs
sayan-oai: i have the same confusion. is sync vs. async whether they block the turn, rather than whether they're serialized? this should be clarified in docs/code, otherwise I'd think I could serialize hook execution (hook A formats, hook B runs tests given formatting finished).

codex-rs/hooks/src/events/session_start.rs
```diff
@@ -0,0 +1,361 @@
...
+                        should_stop = true;
+                        stop_reason = parsed.universal.stop_reason.clone();
+                        if let Some(stop_reason_text) = parsed.universal.stop_reason {
+                            entries.push(HookOutputEntry {
+                                kind: HookOutputEntryKind::Stop,
+                                text: stop_reason_text,
+                            });
+                        }
+                    }
+                } else if let Some(additional_context) = trimmed_non_empty(&run_result.stdout) {
```
sayan-oai: do we want to inject stdout into model context if parsing fails? this also seems different than the behavior for stop hooks, which doesn't inject on parse failure (it seems to silently no-op).

codex-rs/hooks/src/events/session_start.rs
```diff
@@ -0,0 +1,361 @@
...
+                    if let Some(additional_context) = parsed.additional_context {
+                        entries.push(HookOutputEntry {
+                            kind: HookOutputEntryKind::Context,
+                            text: additional_context.clone(),
+                        });
+                        if parsed.universal.continue_processing {
+                            additional_context_for_model = Some(additional_context);
+                        }
+                    }
+                    let _ = parsed.universal.suppress_output;
```
sayan-oai, line 177: should we include suppress_output in the schema if we ignore it here and on `Stop`?

sayan-oai: im assuming this is deferred, like `clear`. we should document this in code.

codex-rs/core/src/codex.rs
```diff
@@ -5394,6 +5490,57 @@ pub(crate) async fn run_turn(
...
+                        )
+                        .await;
+                    }
+                    let stop_outcome = sess.hooks().run_stop(stop_request).await;
+                    for completed in stop_outcome.hook_events {
+                        sess.send_event(&turn_context, EventMsg::HookCompleted(completed))
+                            .await;
+                    }
+                    if stop_outcome.should_block {
+                        if stop_hook_active {
```
sayan-oai, line 5699: maybe this is aspirational/deferred, but is it possible to have a less coarse way of defending against infinite loops? instead of a bool, could we filter by unique `reason`s, or have a >1 limit? it seems plausible to want to have >1 stop hook run per turn. feel free to push back, not blocking.

codex-rs/hooks/src/engine/dispatcher.rs
```diff
@@ -0,0 +1,190 @@
...
+            command: command.to_string(),
+            timeout_sec: 5,
+            status_message: None,
+            source_path: PathBuf::from("/tmp/hooks.json"),
+            display_order,
+        }
+    }
+
+    #[test]
+    fn select_handlers_keeps_duplicate_stop_handlers() {
```
sayan-oai, line 143: do we want to keep duplicated handlers? why not dedupe?

sayan-oai (commented):
mostly behavioral questions. the overall structure makes sense to me.

codex-rs/hooks/src/events/stop.rs
sayan-oai: +1, restarting the turn with no additional context seems bad.

codex-rs/hooks/src/engine/dispatcher.rs
eternal-openai: Yeah, I had it an earlier version of the PR, but jif pointed out that it's probably a mistake. If the user's config has duplicates defined, it could be for a valid usecase (the script does something different the second time) and generally I think respecting that is probably right. Though if you disagree, happy to put it back

codex-rs/core/src/codex.rs
eternal-openai: Yes, that's coming in the next PR using the `stop_hook_active` mechanism

codex-rs/hooks/src/events/session_start.rs
eternal-openai: Yep, exactly

codex-rs/hooks/src/engine/dispatcher.rs
eternal-openai: As far as I'm aware, the mechanics are correct here:

 - Hooks execute in parallel to maximize overall processing speed
 - The agent loop waits for the hook to complete in the typical case
 - The spec supports doing async (i.e. hooks run in the background - we'll do that later)

codex-rs/hooks/src/engine/command_runner.rs
eternal-openai: Would love to get Bolin's thoughts, and yeah, hooks are inherently unsafe -- though I don't think it's on us to provide their security

codex-rs/hooks/src/events/stop.rs
eternal-openai: Totally agree, great catch, making a change to enforce that reason has to be there if stop-block is true (which is also part of the spec)

eternal-openai:
```
Stop hook (failed)
  warning: Wizard Tower Stop hook reviewed the completed reply (129 chars).
  error: hook returned decision "block" without a non-empty reason
```

Implemented. Also handling the exit code 2 with black stderr variant of the same issue

codex-rs/hooks/src/events/stop.rs
eternal-openai: Great point, thank you, implemented

codex-rs/hooks/src/events/session_start.rs
eternal-openai: Thank you, now treating malformed JSON-like output as an error, while preserving intentional plain text output

eternal-openai:
Appreciate the hardening pass, feels like we're close! Changes are made, please review once more 🙏

sayan-oai (approved):

PR #13283, opened 2026-03-02
Codex docs: update repo install docs for native installers

efrazer-oai:
## Summary

We're migrating to native curl and powershell install commands. This PR updates our documentation to prioritize native CLI install commands over package manager install commands. 

We keep npm and Homebrew documented as fallback install methods instead of the primary path.


codex-cli/README.md
```diff
@@ -1,10 +1,11 @@
 <h1 align="center">OpenAI Codex CLI</h1>
 <p align="center">Lightweight coding agent that runs in your terminal</p>
 
-<p align="center"><code>npm i -g @openai/codex</code></p>
```
etraut-openai: This README is for the original typescript implementation of codex, which was replaced by the rust implementation. No need to update it.

efrazer-oai:
@codex review

PR #13285, opened 2026-03-02
feat(app-server): add tracing to all app-server APIs

owenlin0:
### Overview
This PR adds the first piece of tracing for app-server JSON-RPC requests.

There are two main changes:
- JSON-RPC requests can now take an optional W3C trace context at the top level via a `trace` field (`traceparent` / `tracestate`).
- app-server now creates a dedicated request span for every inbound JSON-RPC request in `MessageProcessor`, and uses the request-level trace context as the parent when present.

For compatibility with existing flows, app-server still falls back to the TRACEPARENT env var when there is no request-level traceparent.

This PR is intentionally scoped to the app-server boundary. In a followup, we'll actually propagate trace context through the async handoff into core execution spans like run_turn, which will make app-server traces much more useful.

### Spans
A few details on the app-server span shape:
- each inbound request gets its own server span
- span/resource names are based on the JSON-RPC method (`initialize`, `thread/start`, `turn/start`, etc.)
- spans record transport (stdio vs websocket), request id, connection id, and client name/version when available
- `initialize` stores client metadata in session state so later requests on the same connection can reuse it

### DIFF
--- a/codex-rs/Cargo.lock
+++ b/codex-rs/Cargo.lock
@@ -1428,6 +1428,7 @@ dependencies = [
  "codex-feedback",
  "codex-file-search",
  "codex-login",
+ "codex-otel",
  "codex-protocol",
  "codex-rmcp-client",
  "codex-shell-command",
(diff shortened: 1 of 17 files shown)


owenlin0:
@codex review

owenlin0:
@codex review

pakrym-oai (approved):

PR #13292, opened 2026-03-03
[feedback] diagnostics

rhan-oai:
- added header logic to display diagnostics on cli
- added logic for collecting env vars

Screenshot 2026-03-03 at 3 49 31 PM

Screenshot 2026-03-02 at 6 47 54 PM


codex-rs/feedback/src/feedback_diagnostics.rs
```diff
@@ -0,0 +1,292 @@
+use std::collections::HashMap;
```
pakrym-oai, line 1: move into feedback crate?

codex-rs/tui/src/chatwidget.rs
```diff
@@ -1257,6 +1257,7 @@ impl ChatWidget {
     ) {
         // Build a fresh snapshot at the time of opening the note overlay.
         let snapshot = self.feedback.snapshot(self.thread_id);
+        let diagnostics = codex_core::feedback_diagnostics::FeedbackDiagnostics::collect_from_env();
```
pakrym-oai: should this be collected in feedback.snapshot ?

rhan-oai: feedback.snapshot returns a CodexLogSnapshot, do we want to include diagnostics in that? i guess the diagnostics are kind of a log snapshot but since they exist separately as env variables not sure if it makes sense if its not surely consistent? lmk

codex-rs/tui/src/bottom_pane/feedback_view.rs
```diff
@@ -58,11 +64,17 @@ pub(crate) struct FeedbackNoteView {
     complete: bool,
 }
 
+struct PreparedFeedbackAttachments {
+    attachment_paths: Vec<PathBuf>,
+    _diagnostics_attachment: Option<FeedbackDiagnosticsAttachment>,
```
pakrym-oai: why is this `_***`?

codex-rs/tui/src/chatwidget.rs
```diff
@@ -1254,6 +1254,17 @@ impl ChatWidget {
...
+        self.show_feedback_note(category, include_logs, diagnostics);
+    }
+
+    fn show_feedback_note(
+        &mut self,
+        category: crate::app_event::FeedbackCategory,
+        include_logs: bool,
+        diagnostics: codex_feedback::feedback_diagnostics::FeedbackDiagnostics,
     ) {
         // Build a fresh snapshot at the time of opening the note overlay.
```
pakrym-oai: codex_feedback::feedback_diagnostics::FeedbackDiagnostics::collect_from_env(); should run as part of feedback.snapshot

codex-rs/feedback/src/feedback_diagnostics.rs
```diff
@@ -0,0 +1,293 @@
...
+        &self.path
+    }
+}
+
+impl FeedbackDiagnostics {
+    pub fn collect_from_env() -> Self {
+        Self::collect_from_pairs(std::env::vars())
+    }
+
+    pub fn collect_from_pairs<I, K, V>(pairs: I) -> Self
```
pakrym-oai: this shouldn't be public

pakrym-oai (approved):

codex-rs/tui/src/chatwidget.rs
pakrym-oai: you can rename the thing we return a FeedbackSnapshot and then it'll fit.

My goal is to have `upload_feedback` "just work" and upload everything including diagnostics.

PR #13329, opened 2026-03-03
[js_repl] Hard-stop active js_repl execs on explicit user interrupts

aaronl-openai:
## Summary
- hard-stop `js_repl` only for `TurnAbortReason::Interrupted`, preserving the persistent REPL across replaced turns
- track the current top-level exec by turn and only reset when the interrupted turn owns submitted work or a freshly started kernel for the current exec attempt
- close both interrupt races: the write-window race by marking the exec as submitted before async pipe writes begin, and the startup-window race by tracking fresh-kernel ownership until submission
- add regression coverage for interrupted in-flight execs and the pending-kernel-start window

## Why
Stopping a turn previously surfaced `aborted by user after Xs` even though the underlying `js_repl` kernel could continue executing. Earlier fixes also risked resetting the session-scoped REPL too broadly or missing already-dispatched work. This change keeps cleanup scoped to explicit stop semantics and makes the interrupt path line up with both submitted execs and newly started kernels.

## Testing
- `just fmt`
- `cargo test -p codex-core`
- `just fix -p codex-core`

`cargo test -p codex-core` passes the updated `js_repl` coverage, including the new startup-window regression test, but still has unrelated integration failures in this environment outside `js_repl`.


codex-rs/core/src/tasks/mod.rs
```diff
@@ -277,6 +277,13 @@ impl Session {
             .abort(session_ctx, Arc::clone(&task.turn_context))
             .await;
 
+        if reason == TurnAbortReason::Interrupted
```
fjord-oai: Can we avoid making `Session::handle_task_abort()` know about `js_repl` specifically? I agree we need explicit interrupt cleanup for long-lived runtimes, but this feels like orchestration code accumulating tool-specific teardown. Since we already have a separate interrupt-cleanup path for `unified_exec`, could we push this behind something like a more generic turn/session-level `cleanup_after_interrupt()` hook instead, so `tasks/mod.rs` stays responsible for abort flow rather than individual tool/runtime details?

codex-rs/core/src/tools/js_repl/mod.rs
```diff
@@ -658,9 +725,25 @@ impl JsReplManager {
         })?;
         self.reset_kernel().await;
         Self::clear_all_exec_tool_calls_map(&self.exec_tool_calls).await;
+        self.clear_current_exec().await;
         Ok(())
     }
 
+    pub async fn interrupt_turn_exec(&self, turn_id: &str) -> Result<bool, FunctionCallError> {
```
fjord-oai, line 807: Do we need `interrupt_turn_exec()` to be selective here, or could interrupt cleanup just call `manager.reset()` whenever this turn has initialized `js_repl`? The manager is already turn-scoped, and `js_repl_reset` is just a thin model-facing wrapper around `reset()`, so the new `current_exec` / `pending_owner_turn` bookkeeping feels like extra state whose only job is deciding whether reset should be a no-op. I think we could drop most of the extra bookkeeping and do something like:
```
if let Some(manager) = task.turn_context.js_repl.manager_if_initialized() { manager.reset().await?; }
```

aaronl-openai: ah my understanding is that manager is session scoped not turn scoped so if the js_repl was used in a previous turn, and then an interrupt happens in the current turn but before js_repl is actually used, we'd end up unnecessarily resetting it.

fjord-oai: yup, that's a good point! i'm good with it as is then

fjord-oai (approved):
approved, but get codex team input on the new interrupt cleanup mechanism

codex-rs/core/src/codex/interrupt_cleanup.rs
```diff
@@ -0,0 +1,22 @@
...
+
+impl Session {
+    pub(crate) async fn close_unified_exec_processes(&self) {
+        self.services
+            .unified_exec_manager
+            .terminate_all_processes()
+            .await;
+    }
+
+    pub(crate) async fn cleanup_after_interrupt(&self, turn_context: &Arc<TurnContext>) {
```
owenlin0: interesting, why break this out into a separate file? don't see a lot of value in doing so. could this just be methods in `tasks/mod.rs`?

owenlin0 (approved):

PR #13333, opened 2026-03-03
Refactor plugin config and cache path

xl-openai:
Update config.toml plugin entries to use <plugin_name>@<marketplace_name> as the key.
Plugin now stays in [plugins/cache/marketplace-name/plugin-name/$version/]
Clean up the plugin code structure.
Add plugin install functionality (not used yet).

### DIFF
--- a/codex-rs/core/src/plugins/manager.rs
+++ b/codex-rs/core/src/plugins/manager.rs
@@ -1,6 +1,7 @@
 use super::load_plugin_manifest;
 use super::plugin_manifest_name;
 use super::store::DEFAULT_PLUGIN_VERSION;
+use super::store::PluginId;
 use super::store::PluginInstallRequest;
 use super::store::PluginInstallResult;
 use super::store::PluginStore;
@@ -155,7 +156,7 @@ impl PluginsManager {
 
         ConfigService::new_with_defaults(self.codex_home.clone())
             .write_value(ConfigValueWriteParams {
-                key_path: format!("plugins.{}", result.plugin_key),
+                key_path: format!("plugins.{}", result.plugin_id.as_key()),
                 value: json!({
                     "enabled": true,
                 }),
@@ -285,7 +286,8 @@ fn configured_plugins_from_stack(
 
 fn load_plugin(config_name: String, plugin: &PluginConfig, store: &PluginStore) -> LoadedPlugin {
     let plugin_version = DEFAULT_PLUGIN_VERSION.to_string();
-    let plugin_root = store.plugin_root(&config_name, &plugin_version);
+    let plugin_root = PluginId::parse(&config_name)
+        .map(|plugin_id| store.plugin_root(&plugin_id, &plugin_version));
     let root = match &plugin_root {
         Ok(plugin_root) => plugin_root.clone(),
         Err(_) => store.root().clone(),
@@ -731,8 +733,7 @@ mod tests {
         assert_eq!(
             result,
             PluginInstallResult {
-                plugin_key: "sample-plugin@debug".to_string(),
-                plugin_name: "sample-plugin".to_string(),
+                plugin_id: PluginId::new("sample-plugin".to_string(), "debug".to_string()).unwrap(),
                 plugin_version: "local".to_string(),
                 installed_path,
             }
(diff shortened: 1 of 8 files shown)


codex-rs/core/src/plugins/store.rs
```diff
@@ -0,0 +1,356 @@
...
+        return Err(PluginStoreError::InvalidPluginKey(format!(
+            "invalid plugin key `{plugin_key}`; expected <plugin>@<marketplace>"
+        )));
+    }
+    validate_plugin_key_segment(plugin_name, "plugin name", plugin_key)?;
+    validate_plugin_key_segment(marketplace_name, "marketplace name", plugin_key)?;
+    Ok((plugin_name, marketplace_name))
+}
+
+fn validate_plugin_key_segment(
```
sayan-oai: optional nit: maybe just inline the two calls in `parse_plugin_key`? there's already a lot of standalone functions in this file

codex-rs/core/src/plugins/store.rs
```diff
@@ -0,0 +1,356 @@
...
+            return Err(PluginStoreError::InvalidPlugin(format!(
+                "plugin source path is not a directory: {}",
+                source_path.display()
+            )));
+        }
+
+        let plugin_name = plugin_name_for_source(&source_path)?;
+        let marketplace_name = request
+            .marketplace_name
+            .filter(|name| !name.trim().is_empty())
```
sayan-oai, line 113: we dont do any validation on `marketplace_name` and then call rm/copy on it later. can we validate the field (maybe just call `plugin_root` which already validates)?

codex-rs/core/src/plugins/manager.rs
```diff
@@ -128,6 +143,50 @@ impl PluginsManager {
...
+        request: PluginInstallRequest,
+    ) -> Result<PluginInstallResult, PluginInstallError> {
+        let store = self.store.clone();
+        let result = tokio::task::spawn_blocking(move || store.install(request))
+            .await
+            .map_err(PluginInstallError::join)??;
+
+        ConfigService::new_with_defaults(self.codex_home.clone())
+            .write_value(ConfigValueWriteParams {
+                key_path: format!("plugins.{}", result.plugin_key),
```
sayan-oai: we separate by `.` here but `ConfigService::parse_key_path` splits on `.`, so this would cause weird key splitting I think.

codex-rs/core/src/plugins/store.rs
```diff
@@ -0,0 +1,356 @@
...
+use super::manifest::PLUGIN_MANIFEST_PATH;
+use super::plugin_manifest_name;
+use codex_utils_absolute_path::AbsolutePathBuf;
+use std::fs;
+use std::io;
+use std::path::Path;
+use std::path::PathBuf;
+
+const DEFAULT_MARKETPLACE_NAME: &str = "debug";
+pub(crate) const DEFAULT_PLUGIN_VERSION: &str = "local";
```
sayan-oai, line 11: question: does version matter enough to put it in `PluginConfig`? or do we just want to abstract from users for now?

codex-rs/core/src/plugins/store.rs
```diff
@@ -0,0 +1,356 @@
...
+        ))
+    })?;
+
+    let plugin_name = plugin_manifest_name(&manifest, source_path);
+    validate_plugin_segment(&plugin_name, "plugin name")
+        .map_err(PluginStoreError::InvalidPlugin)
+        .map(|_| plugin_name)
+}
+
+fn parse_plugin_key(plugin_key: &str) -> Result<(&str, &str), PluginStoreError> {
```
sayan-oai: should we just make this a type rather than a specifically-formatted string? like `PluginId {name, marketplace}`, or something?

codex-rs/core/src/plugins/store.rs
xl-openai: Good call. Switched to a character whitelist.

codex-rs/core/src/plugins/store.rs
xl-openai: Version will likely come directly from the marketplace file. We can decide later whether we want to also persist it to config, but LOCAL will be the default anyhow. We might need to support auto update as well.

sayan-oai (approved):

PR #13334, opened 2026-03-03
app-server service tier plumbing (plus some cleanup)

pash-openai:
followup to https://github.com/openai/codex/pull/13212 to expose fast tier controls to app server
(majority of this PR is generated schema jsons - actual code is +69 / -35 and +24 tests )

- add service tier fields to the app-server protocol surfaces used by thread lifecycle, turn start, config, and session configured events
- thread service tier through the app-server message processor and core thread config snapshots
- allow runtime config overrides to carry service tier for app-server callers

cleanup:
- Removing useless "legacy" code supporting "standard" - we moved to None | "fast", so "standard" is not needed.


codex-rs/core/src/config/mod.rs
```diff
@@ -5655,33 +5659,6 @@ trust_level = "untrusted"
         Ok(())
     }
 
-    #[test]
-    fn legacy_standard_service_tier_loads_as_default_none() -> anyhow::Result<()> {
```
pash-openai, line 5659: Not needed - this was an artifact from moving to None | "fast" in the previous PR

codex-rs/core/config.schema.json
```diff
@@ -1328,21 +1328,10 @@
...
-        {
-          "enum": [
-            "fast"
-          ],
-          "type": "string"
-        },
-        {
-          "description": "Legacy compatibility value for older local config files.",
-          "enum": [
-            "standard"
```
pash-openai, line 1341: cleaned up

jif-oai (approved):
lgtm

codex-rs/protocol/src/config_types.rs
```diff
@@ -117,8 +117,6 @@ pub enum WebSearchMode {
 #[serde(rename_all = "lowercase")]
 #[strum(serialize_all = "lowercase")]
 pub enum ServiceTier {
-    /// Legacy compatibility value for older local config files.
-    Standard,
     Fast,
```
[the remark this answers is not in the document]
pash-openai: older configs with standard never existed

PR #13366, opened 2026-03-03
[bazel] Bump rules_rs and llvm

zbarsky-openai:
With upstream fixes we can simplify some things:
- More careful treatment of deps in macro allows to drop integration_deps_extra hack
- previous `libclang.so` was not hermetic, specify more properly following rules_rs recommendation for the override
- rework `all_crate_deps` handling to account for rules-rs no longer wrapping the dep lists in `rust_deps` rule (which broke CcInfo propagation).
- No need for special proc-macro handling as rules_rs transparently handles it now
- windows import annotations are handled transparently now
- switch to official rules_rs release instead of side branch


bolinfest (approved):
Thanks: this certainly feels simpler and I'm excited to try this out!

PR #13382, opened 2026-03-03
fix(tui): decode ANSI alpha-channel encoding in syntax themes

fcoury:
## Problem

The `ansi`, `base16`, and `base16-256` syntax themes are designed to emit ANSI palette colors so that highlighted code respects the user's terminal color scheme. Syntect encodes this intent in the alpha channel of its `Color` struct — a convention shared with `bat` — but `convert_style` was ignoring it entirely, treating every foreground color as raw RGB. This caused ANSI-family themes to produce hard-coded RGB values (e.g. `Rgb(0x02, 0, 0)` instead of `Green`), defeating their purpose and rendering them as near-invisible dark colors on most terminals.

Reported in #12890.

## Mental model

Syntect themes use a compact encoding in their `Color` struct:

| `alpha` | Meaning of `r` | Mapped to |
|---------|----------------|-----------|
| `0x00` | ANSI palette index (0–255) | `RtColor::Black`…`Gray` for 0–7, `Indexed(n)` for 8–255 |
| `0x01` | Unused (sentinel) | `None` — inherit terminal default fg/bg |
| `0xFF` | True RGB red channel | `RtColor::Rgb(r, g, b)` |
| other | Unexpected | `RtColor::Rgb(r, g, b)` (silent fallback) |

This encoding is a bat convention that three bundled themes rely on. The new `convert_syntect_color` function decodes it; `ansi_palette_color` maps indices 0–7 to ratatui's named ANSI variants.

| macOS - Dark | macOS - Light | Windows - ansi | Windows - base16 |
|---|---|---|---|
| macos-dark | macos-light | windows-ansi | windows-base16 |

## Non-goals

- Background color decoding — we intentionally skip backgrounds to preserve the terminal's own background. The decoder supports it, but `convert_style` does not apply it.
- Italic/underline changes — those remain suppressed as before.
- Custom `.tmTheme` support for ANSI encoding — only the bundled themes use this convention.

## Tradeoffs

- The alpha-channel encoding is an undocumented bat/syntect convention, not a formal spec. We match bat's behavior exactly, trading formality for ecosystem compatibility.
- Indices 0–7 are mapped to ratatui's named variants (`Black`, `Red`, …, `Gray`) rather than `Indexed(0)`…`Indexed(7)`. This lets terminals apply bold/bright semantics to named colors, which is the expected behavior for ANSI themes, but means the two representations are not perfectly round-trippable.

## Architecture

All changes are in `codex-rs/tui/src/render/highlight.rs`, within the style-conversion layer between syntect and ratatui:

```
syntect::highlighting::Color
  └─ convert_syntect_color(color)  [NEW — alpha-dispatch]
       ├─ a=0x00 → ansi_palette_color()  [NEW — index→named/indexed]
       ├─ a=0x01 → None (terminal default)
       ├─ a=0xFF → Rgb(r,g,b) (standard opaque path)
       └─ other  → Rgb(r,g,b) (silent fallback)
```

`convert_style` delegates foreground mapping to `convert_syntect_color` instead of inlining the `Rgb(r,g,b)` conversion. The core highlighter is refactored into `highlight_to_line_spans_with_theme` (accepts an explicit theme reference) so tests can highlight against specific themes without mutating process-global state.

### ANSI-family theme contract

The ANSI-family themes (`ansi`, `base16`, `base16-256`) rely on upstream alpha-channel encoding from two_face/syntect. We intentionally do **not** validate this contract at runtime — if the upstream format changes, the `ansi_themes_use_only_ansi_palette_colors` test catches it at build time, long before it reaches users. A runtime warning would be unactionable noise.

### Warning copy cleanup

User-facing warning messages were rewritten for clarity:
- Removed internal jargon ("alpha-encoded ANSI color markers", "RGB fallback semantics", "persisted override config")
- Dropped "syntax" prefix from "syntax theme" — users just think "theme"
- Downgraded developer-only diagnostics (duplicate override, resolve fallback) from `warn` to `debug`

## Observability

- The `ansi_themes_use_only_ansi_palette_colors` test enforces the ANSI-family contract at build time.
- The snapshot test provides a regression tripwire for palette color output.
- User-facing warnings are limited to actionable issues: unknown theme names and invalid custom `.tmTheme` files.

## Tests

- **Unit tests for each alpha branch:** `alpha=0x00` with low index (named color), `alpha=0x00` with high index (`Indexed`), `alpha=0x01` (terminal default), unexpected alpha (falls back to RGB), ANSI white → Gray mapping.
- **Integration test:** `ansi_family_themes_use_terminal_palette_colors_not_rgb` — highlights a Rust snippet with each ANSI-family theme and asserts zero `Rgb` foreground colors appear.
- **Snapshot test:** `ansi_family_foreground_palette_snapshot` — records the exact set of unique foreground colors each ANSI-family theme produces, guarding against regressions.
- **Warning validation tests:** verify user-facing warnings for missing custom themes, invalid `.tmTheme` files, and bundled theme resolution.

## Test plan

- [ ] `cargo test -p codex-tui` passes all new and existing tests
- [ ] Select `ansi`, `base16`, or `base16-256` theme and verify code blocks render with terminal palette colors (not near-black RGB)
- [ ] Select a standard RGB theme (e.g. `dracula`) and verify no regression in color output

### DIFF
--- a/codex-rs/tui/src/render/highlight.rs
+++ b/codex-rs/tui/src/render/highlight.rs
@@ -1020,7 +1020,7 @@ mod tests {
             font_style: FontStyle::empty(),
         };
         let rt = convert_style(syn);
-        assert_eq!(rt.fg, Some(RtColor::Indexed(0x9a)));
+        assert!(matches!(rt.fg, Some(RtColor::Indexed(0x9a))));
     }
 
     #[test]
@@ -1062,7 +1062,7 @@ mod tests {
             font_style: FontStyle::empty(),
         };
         let rt = convert_style(syn);
-        assert_eq!(rt.fg, Some(RtColor::Rgb(10, 20, 30)));
+        assert!(matches!(rt.fg, Some(RtColor::Rgb(10, 20, 30))));
     }
 
     #[test]
--- a/codex-rs/tui/src/render/highlight.rs
+++ b/codex-rs/tui/src/render/highlight.rs
@@ -31,6 +31,7 @@ use std::path::PathBuf;
 use std::sync::OnceLock;
 use std::sync::RwLock;
 use syntect::easy::HighlightLines;
+use syntect::highlighting::Color as SyntectColor;
 use syntect::highlighting::FontStyle;
 use syntect::highlighting::Highlighter;
 use syntect::highlighting::Style as SyntectStyle;
@@ -49,8 +50,60 @@ static THEME: OnceLock<RwLock<Theme>> = OnceLock::new();
 static THEME_OVERRIDE: OnceLock<Option<String>> = OnceLock::new();
 static CODEX_HOME: OnceLock<Option<PathBuf>> = OnceLock::new();
 
+const ANSI_ALPHA_INDEX: u8 = 0x00;
+const ANSI_ALPHA_DEFAULT: u8 = 0x01;
+const OPAQUE_ALPHA: u8 = 0xFF;
+
 fn syntax_set() -> &'static SyntaxSet {
     SYNTAX_SET.get_or_init(two_face::syntax::extra_newlines)
+}
+
+fn append_warning(warning: &mut Option<String>, msg: String) {
+    if let Some(existing) = warning {
+        if !existing.contains(&msg) {
+            existing.push('\n');
+            existing.push_str(&msg);
+        }
+    } else {
+        *warning = Some(msg);
+    }
+}
+
+fn is_ansi_family_theme_name(name: &str) -> bool {
+    matches!(name, "ansi" | "base16" | "base16-256")
+}
+
+fn is_ansi_alpha_signal(color: Option<SyntectColor>) -> bool {
+    color.is_some_and(|c| matches!(c.a, ANSI_ALPHA_INDEX | ANSI_ALPHA_DEFAULT))
+}
+
+fn theme_uses_ansi_alpha_encoding(theme: &Theme) -> bool {
+    is_ansi_alpha_signal(theme.settings.foreground)
+        || is_ansi_alpha_signal(theme.settings.background)
+        || theme.scopes.iter().any(|item| {
+            is_ansi_alpha_signal(item.style.foreground)
+                || is_ansi_alpha_signal(item.style.background)
+        })
+}
+
+fn ansi_family_theme_contract_warning_for_theme(name: &str, theme: &Theme) -> Option<String> {
+    if !is_ansi_family_theme_name(name) || theme_uses_ansi_alpha_encoding(theme) {
+        return None;
+    }
+    Some(format!(
+        "Syntax theme \"{name}\" resolved, but no alpha-encoded ANSI color \
+         markers were detected. This may indicate an upstream theme-format \
+         change. Codex will continue with RGB fallback semantics."
+    ))
+}
+
+fn validate_ansi_family_theme_contract(
+    name: Option<&str>,
+    codex_home: Option<&Path>,
+) -> Option<String> {
+    let name = name?;
+    let theme = resolve_theme_by_name(name, codex_home)?;
+    ansi_family_theme_contract_warning_for_theme(name, &theme)
 }
 
 /// Set the user-configured syntax theme override and codex home path.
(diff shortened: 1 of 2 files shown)


codex-rs/tui/src/render/highlight.rs
```diff
@@ -49,10 +50,62 @@ static THEME: OnceLock<RwLock<Theme>> = OnceLock::new();
...
 
+const ANSI_ALPHA_INDEX: u8 = 0x00;
+const ANSI_ALPHA_DEFAULT: u8 = 0x01;
+const OPAQUE_ALPHA: u8 = 0xFF;
+
 fn syntax_set() -> &'static SyntaxSet {
     SYNTAX_SET.get_or_init(two_face::syntax::extra_newlines)
 }
 
+fn append_warning(warning: &mut Option<String>, msg: String) {
```
joshka-oai: nit: when functions are defined before they're used, reviewing the code requires you to read things out of order, and poses questions somewhat like "why is this method useful" for this and the following methods below. I have a strong preference to write code top-to-bottom outside-in when possible as it eases this sort of review. Not a huge deal and often a very subjective opinion.

codex-rs/tui/src/render/highlight.rs
```diff
@@ -49,10 +50,62 @@ static THEME: OnceLock<RwLock<Theme>> = OnceLock::new();
...
+                || is_ansi_alpha_signal(item.style.background)
+        })
+}
+
+fn ansi_family_theme_contract_warning_for_theme(name: &str, theme: &Theme) -> Option<String> {
+    if !is_ansi_family_theme_name(name) || theme_uses_ansi_alpha_encoding(theme) {
+        return None;
+    }
+    Some(format!(
+        "Syntax theme \"{name}\" resolved, but no alpha-encoded ANSI color \
```
etraut-openai: This warning text is pretty hard to understand. I wonder if a warning is really needed here. I think it's unlikely to be common enough to merit an in-app warning. It's probably sufficient to just log it and fall back. If we do keep the warning, I think it should be rewritten to make it clearer.

codex-rs/tui/src/render/highlight.rs
```diff
@@ -49,10 +50,62 @@ static THEME: OnceLock<RwLock<Theme>> = OnceLock::new();
...
+            existing.push_str(&msg);
+        }
+    } else {
+        *warning = Some(msg);
+    }
+}
+
+fn is_ansi_family_theme_name(name: &str) -> bool {
+    matches!(name, "ansi" | "base16" | "base16-256")
+}
```
joshka-oai: Can we hang this off a newtype perhaps? So this becomes:
```suggestion
/// description of why this is useful / used
fn is_ansi_family(self) -> bool {
    matches!(self, ThemeName::Ansi | ThemeName::Base16 | ThemeName::Base16_256)
}
```

codex-rs/tui/src/render/highlight.rs
```diff
@@ -49,10 +50,62 @@ static THEME: OnceLock<RwLock<Theme>> = OnceLock::new();
...
+        return None;
+    }
+    Some(format!(
+        "Syntax theme \"{name}\" resolved, but no alpha-encoded ANSI color \
+         markers were detected. This may indicate an upstream theme-format \
+         change. Codex will continue with RGB fallback semantics."
+    ))
+}
+
+fn validate_ansi_family_theme_contract(
```
joshka-oai: docs - what is the expected validation?

etraut-openai (approved):
Looks pretty good. I left one comment — not sure the warning is needed here. Seems to be adding a lot of code for little benefit.

codex-rs/tui/src/render/highlight.rs
fcoury: Good catch, I agree with you. Let me fix it.

joshka-oai (approved):
When a module grows a large percentage of code like this, it's worth considering whether the code can be split into a new more focused module (e.g. `highlight/syntect_theme.rs` (or whatever makes sense - `theme.rs`?). This helps keep the tests, and docs about why and how this code works pretty well self-contained / local / collocated and makes the additive and changed parts of a change like this more obvious. A bunch of the mental model docs in the PR probably belong in the code too.

Stamped - mainly because this improves how this looks significantly, so code nits are more about making it read nicely.

codex-rs/tui/src/render/highlight.rs
fcoury: Never thought of that, good insight. I will keep that in mind.

codex-rs/tui/src/render/highlight.rs
fcoury: I removed the check and warning completely. This is a development time concern, not runtime. The test `ansi_themes_use_only_ansi_palette_colors` already covers the case where upstream changes behavior.

codex-rs/tui/src/render/highlight.rs
fcoury: Most of the functions you flagged here (`append_warning`, `is_ansi_family_theme_name`, etc.) have been removed since they were not useful at runtime. We already catch drifts from upstream with an existing test.

codex-rs/tui/src/render/highlight.rs
fcoury: Same applies here, I just removed this function as well.

codex-rs/tui/src/render/highlight.rs
fcoury: Function also removed.

PR #13392, opened 2026-03-03
refactor: prepare unified exec for zsh-fork backend

bolinfest:
## Why

`shell_zsh_fork` already provides stronger guarantees around which executables receive elevated permissions. To reuse that machinery from unified exec without pushing Unix-specific escalation details through generic runtime code, the escalation bootstrap and session lifetime handling need a cleaner boundary.

That boundary also needs to be safe for long-lived sessions: when an intercepted shell session is closed or pruned, any in-flight approval workers and any already-approved escalated child they spawned must be torn down with the session, and the inherited escalation socket must not leak into unrelated subprocesses.

## What Changed

- Extracted a reusable `EscalationSession` and `EscalateServer::start_session(...)` in `shell-escalation` so callers can get the wrapper/socket env overlay and keep the escalation server alive without immediately running a one-shot command.
- Documented that `EscalationSession::env()` and `ShellCommandExecutor::run(...)` exchange only that env overlay, which callers must merge into their own base shell environment.
- Clarified the prepared-exec helper boundary in `core` by naming the new helper APIs around `ExecRequest`, while keeping the legacy `execute_env(...)` entrypoints as thin compatibility wrappers for existing callers that still use the older naming.
- Added a small post-spawn hook on the prepared execution path so the parent copy of the inheritable escalation socket is closed immediately after both the existing one-shot shell-command spawn and the unified-exec spawn.
- Made session teardown explicit with session-scoped cancellation: dropping an `EscalationSession` or canceling its parent request now stops intercept workers, and the server-spawned escalated child uses `kill_on_drop(true)` so teardown cannot orphan an already-approved child.
- Added `UnifiedExecBackendConfig` plumbing through `ToolsConfig`, a `shell::zsh_fork_backend` facade, and an opaque unified-exec spawn-lifecycle hook so unified exec can prepare a wrapped `zsh -c/-lc` request without storing `EscalationSession` directly in generic process/runtime code.
- Kept the existing `shell_command` zsh-fork behavior intact on top of the new bootstrap path. Tool selection is unchanged in this PR: when `shell_zsh_fork` is enabled, `ShellCommand` still wins over `exec_command`.

## Verification

  - includes coverage for `start_session_exposes_wrapper_env_overlay`
  - includes coverage for `exec_closes_parent_socket_after_shell_spawn`
  - includes coverage for `dropping_session_aborts_intercept_workers_and_kills_spawned_child`
- `cargo test -p codex-core shell_zsh_fork_prefers_shell_command_over_unified_exec`
- `cargo test -p codex-core --test all shell_zsh_fork_prompts_for_skill_script_execution`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/13392).
* #13432
* __->__ #13392

### DIFF
--- a/codex-rs/shell-escalation/src/unix/escalate_server.rs
+++ b/codex-rs/shell-escalation/src/unix/escalate_server.rs
@@ -3,11 +3,14 @@ use std::os::fd::AsRawFd;
 use std::path::PathBuf;
 use std::process::Stdio;
 use std::sync::Arc;
+use std::sync::Mutex;
 use std::time::Duration;
 
 use anyhow::Context as _;
 use codex_utils_absolute_path::AbsolutePathBuf;
+use socket2::Socket;
 use tokio::process::Command;
+use tokio::task::JoinHandle;
 use tokio_util::sync::CancellationToken;
 
 use crate::unix::escalate_protocol::ESCALATE_SOCKET_ENV_VAR;
@@ -32,12 +35,20 @@ use crate::unix::socket::AsyncSocket;
 #[async_trait::async_trait]
 pub trait ShellCommandExecutor: Send + Sync {
     /// Runs the requested shell command and returns the captured result.
+    ///
+    /// `env_overlay` contains only the wrapper/socket variables exported by
+    /// `EscalationSession::env()`, not a complete child environment.
+    /// Implementations should merge it into whatever base environment they use
+    /// for the shell process. `after_spawn` should be invoked immediately after
+    /// the shell process has been spawned so the parent copy of the inherited
+    /// escalation socket can be closed.
     async fn run(
         &self,
         command: Vec<String>,
         cwd: PathBuf,
-        env: HashMap<String, String>,
+        env_overlay: HashMap<String, String>,
         cancel_rx: CancellationToken,
+        after_spawn: Option<Box<dyn FnOnce() + Send>>,
     ) -> anyhow::Result<ExecResult>;
 
     /// Prepares an escalated subcommand for execution on the server side.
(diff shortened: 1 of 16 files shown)


jif-oai (commented):
I might be wrong but I think we have a bug in here.

For example:
1. an `exec_command` open a shell
2. A wrapped child exec is intercepted by your code
3. the new detached worker handle it (somewhere in `handle_escalate_session_with_policy`)
4. We approve
5. It reaches `command.spawn()` and then the `child.wait().await`
6. The terminal session get closed or pruned
7. Approval is granted, so it reaches command.spawn() and then child.wait().await: escalate_server.rs:282-303.
8. Now the terminal session is closed or pruned.
9. `EscalationSession` is dropped whchi does `self.task.abort();`
10. this does not closed the per-intercept worker

As a result, we have an "orphan" process running that can't be cancelled as not attached to a unified_exec session

What do you think?

codex-rs/shell-escalation/src/unix/escalate_server.rs
```diff
@@ -128,65 +196,75 @@ impl EscalateServer {
...
             continue;
         }
         let stream_socket = AsyncSocket::from_fd(fds.remove(0))?;
         let policy = Arc::clone(&policy);
         let command_executor = Arc::clone(&command_executor);
-        tokio::spawn(async move {
-            if let Err(err) =
-                handle_escalate_session_with_policy(stream_socket, policy, command_executor).await
+        let cancellation_token = cancellation_token.clone();
+        let worker = tokio::spawn(async move {
```
jif-oai: You spawn the worker before recording `AbortHandle`.. I think this leaves a race: 
* if `EscalationSession` is dropped in the good window window, `impl Drop` can miss this worker.
* so the shared cancellation token is never cancelled (on drop)
* the worker and any escalated child become orphan and outlive

codex-rs/shell-escalation/src/unix/escalate_server.rs
```diff
@@ -128,65 +196,75 @@ impl EscalateServer {
...
+                command_executor,
+                cancellation_token,
+            )
+            .await
             {
                 tracing::error!("escalate session failed: {err:?}");
             }
         });
+        if let Ok(mut handles) = worker_abort_handles.lock() {
+            handles.push(worker.abort_handle());
```
jif-oai: Would be cool here to drop all the finished handle from `handles`. Otherwise this can grow infinitely

I would do a simple
```
handles.retain(|handle| !handle.is_finished());
```

jif-oai (approved):
lgtm once those 2 are fixed or discussed

PR #13401, opened 2026-03-03
feat: load plugin apps

sayan-oai:
load plugin-apps from `.app.json`.

make apps runtime-mentionable iff `codex_apps` MCP actually exposes tools for that `connector_id`.

if the app isn't available, it's filtered out of runtime connector set, so no tools are added and no app-mentions resolve.

right now we don't have a clean cli-side error for an app not being installed. can look at this after.

### Tests
Added tests, tested locally that using a plugin that bundles an app picks up the app.

### DIFF
--- a/codex-rs/core/src/connectors.rs
+++ b/codex-rs/core/src/connectors.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::env;
 use std::path::PathBuf;
 use std::sync::LazyLock;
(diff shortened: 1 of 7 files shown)


codex-rs/core/src/connectors.rs
```diff
@@ -575,6 +601,30 @@ where
     accessible
 }
 
+fn plugin_app_to_app_info(plugin_app: PluginApp) -> AppInfo {
+    let connector_id = plugin_app.connector_id;
+    let name = if plugin_app.alias.trim().is_empty() {
```
xl-openai: Does app / connector have a formal name already or do we still need an alias?

codex-rs/core/src/connectors.rs
```diff
@@ -575,6 +601,30 @@ where
...
+        connector_id.clone()
+    } else {
+        plugin_app.alias
+    };
+
+    AppInfo {
```
xl-openai, line 628: Wondering if we should fetch the real info from server rather than leave them blank?

codex-rs/core/src/connectors.rs
sayan-oai: good point. we can use the formal name by setting the initial name as just `connector_id`.

xl-openai (approved):

PR #13408, opened 2026-03-03
[Hooks] Schema changes only - AppServer protocol additions & Hooks I/O schema

eternal-openai:
This is just the schema + in development feature flag set up from https://github.com/openai/codex/pull/13276 to make review a bit easier

## Testing

- no functionality changes
- the codex apps should work as expected

## Details

- `codex_hooks` feature flag: introduces an under-development feature gate, off by default.
- Rust-owned hook schemas: adds Rust-generated JSON Schema fixtures for hook command I/O (`SessionStart` and `Stop`) so hook contracts live in `codex-rs`
- Core protocol hook events:
  - `EventMsg::HookStarted`: emitted when a hook run begins.
  - `EventMsg::HookCompleted`: emitted when a hook run finishes.
- Core hook schema objects:
  - `HookEventName`: identifies which hook event ran.
  - `HookHandlerType`: identifies what kind of handler ran.
  - `HookExecutionMode`: describes how the handler was executed.
  - `HookScope`: describes where the hook applies.
  - `HookRunStatus`: captures the final outcome of the run.
  - `HookOutputEntry`: represents one normalized output row from a hook.
  - `HookRunSummary`: represents the full normalized, displayable, and persistable result of one hook run.
- AppServer v2 protocol:
  - adds `hook/started`
  - adds `hook/completed`
  - adds `Turn.hookRuns` so completed hook runs can be attached to a turn
- Thread history / persistence scaffolding: completed hook runs can now be threaded through rollout history reconstruction onto their owning turn.
- Compile-only plumbing: adds minimal no-op handling in app-server, exec, mcp-server, core, and TUI so the new protocol surface builds cleanly on `main`.


codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -1175,6 +1176,8 @@ pub(crate) async fn apply_bespoke_event_handling(
                 .send_server_notification(ServerNotification::ItemCompleted(notification))
                 .await;
         }
+        EventMsg::HookStarted(_) => {}
```
jif-oai, line 1179: why having them if we don't emit them? 

This should be mapped to your `hook/completed`, ...

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -1607,6 +1610,7 @@ async fn emit_turn_completed_with_status(
         turn: Turn {
             id: event_turn_id,
             items: vec![],
+            hook_runs: vec![],
```
jif-oai, line 1613: huh? This sounds strange to never populate this no?

codex-rs/core/src/rollout/policy.rs
```diff
@@ -161,6 +161,8 @@ fn event_msg_persistence_mode(ev: &EventMsg) -> Option<EventPersistenceMode> {
         | EventMsg::ShutdownComplete
         | EventMsg::DeprecationNotice(_)
         | EventMsg::ItemStarted(_)
+        | EventMsg::HookStarted(_)
```
jif-oai, line 164: Any reason not to persist them? My guess would be that this is quite relevant

codex-rs/app-server-protocol/src/protocol/thread_history.rs
```diff
@@ -868,9 +871,35 @@ impl ThreadHistoryBuilder {
...
+            if let Some(turn) = self.turns.iter_mut().find(|turn| turn.id == turn_id) {
+                turn.hook_runs.push(run);
+                return;
+            }
+
+            warn!("dropping hook run for unknown turn id `{turn_id}`");
+            return;
+        }
+
+        self.ensure_turn().hook_runs.push(run);
```
jif-oai, line 893: This thing will check the current turn if there is one, otherwise create a mock turn in the history in order to push the run

But you don't seem to be filtering out non-turn based hooks such as "session start" for example. So why would we attach this one to a turn?

codex-rs/hooks/src/schema.rs
```diff
@@ -0,0 +1,289 @@
...
+    Ok(serde_json::to_vec_pretty(&value)?)
+}
+
+fn schema_for_type<T>() -> RootSchema
+where
+    T: JsonSchema,
+{
+    SchemaSettings::draft07()
+        .with(|settings| {
+            settings.option_add_null_type = false;
```
jif-oai, line 183: I don't understand why do we have this but you also redefine your own `NullableString`? Someting feels a bit wrong here...

As a result you can see that your JSON schema can put `null` when an `Option` is `None` but `schemars` will actually not include them...

codex-rs/protocol/src/protocol.rs
```diff
@@ -1192,6 +1194,97 @@ pub enum EventMsg {
...
+#[serde(rename_all = "snake_case")]
+pub enum HookScope {
+    Thread,
+    Turn,
+}
+
+#[derive(Debug, Clone, Copy, Deserialize, Serialize, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "snake_case")]
+pub enum HookRunStatus {
+    Running,
```
jif-oai, line 1229: The `Running` state is never used and I don't find any gates for race on subscription/execution + this is never reported to the user

jif-oai (commented):
OOC, why fully dropping the previous hooks implementation (not drop here but mostly incompatible with what you're building here)

IMO the implementation direction was quite good

codex-rs/core/src/rollout/policy.rs
eternal-openai: Persistence is handled by HookRunSummary in the larger PR, so that only the summaries of what the hook returned are persisted

codex-rs/app-server-protocol/src/protocol/thread_history.rs
eternal-openai: Start session imo does have a turn attached, it's the first prompt that's sent when starting or resuming a session. I think the mechanics are simpler doing it this way with context additions

codex-rs/hooks/src/schema.rs
eternal-openai: Thanks, I had removed NullableString for now, since it would throw warnings in this version of the PR

PR #13413, opened 2026-03-04
copy command-runner to CODEX_HOME so sandbox users can always execute it

iceweasel-oai:
• Keep Windows sandbox runner launches working from packaged installs by running the helper from a user-owned runtime location.

  On some Windows installs, the packaged helper location is difficult to use reliably for sandboxed runner launches even though the binaries are present. This change works around that by copying codex-
  command-runner.exe into CODEX_HOME/.sandbox-bin/, reusing that copy across launches, and falling back to the existing packaged-path lookup if anything goes wrong.

  The runtime copy lives in a dedicated directory with tighter ACLs than .sandbox: sandbox users can read and execute the runner there, but they cannot modify it. This keeps the workaround focused on the
  command runner, leaves the setup helper on its trusted packaged path, and adds logging so it is clear which runner path was selected at launch.

### DIFF
--- a/codex-rs/windows-sandbox-rs/src/helper_materialization.rs
+++ b/codex-rs/windows-sandbox-rs/src/helper_materialization.rs
@@ -1,0 +1,350 @@
+use anyhow::anyhow;
+use anyhow::Context;
+use anyhow::Result;
+use std::collections::HashMap;
+use std::fs;
+use std::io::Write;
+use std::path::Path;
+use std::path::PathBuf;
+use std::sync::Mutex;
+use std::sync::OnceLock;
+use tempfile::NamedTempFile;
+
+use crate::logging::log_note;
+use crate::sandbox_bin_dir;
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub(crate) enum HelperExecutable {
+    CommandRunner,
+}
+
+impl HelperExecutable {
+    fn file_name(self) -> &'static str {
+        match self {
+            Self::CommandRunner => "codex-command-runner.exe",
+        }
+    }
+
+    fn label(self) -> &'static str {
+        match self {
+            Self::CommandRunner => "command-runner",
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+enum CopyOutcome {
+    Reused,
+    ReCopied,
+}
+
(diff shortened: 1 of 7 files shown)


iceweasel-oai:
@codex review

codex-rs/windows-sandbox-rs/src/helper_materialization.rs
```diff
@@ -0,0 +1,350 @@
...
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => return Ok(false),
+        Err(err) => {
+            return Err(err)
+                .with_context(|| format!("read helper destination metadata {}", destination.display()));
+        }
+    };
+
+    if source_meta.len() != destination_meta.len() {
+        return Ok(false);
+    }
```
dylan-hurd-oai, line 253: Should we validate on a hash of this executable as well?

dylan-hurd-oai, line 253: Should we validate on a hash of this executable as well?

codex-rs/windows-sandbox-rs/src/helper_materialization.rs
```diff
@@ -0,0 +1,350 @@
...
+    PathBuf::from(kind.file_name())
+}
+
+pub(crate) fn resolve_helper_for_launch(
+    kind: HelperExecutable,
+    codex_home: &Path,
+    log_dir: Option<&Path>,
+) -> PathBuf {
+    match copy_helper_if_needed(kind, codex_home, log_dir) {
```
dylan-hurd-oai, line 64: If I'm understanding correctly, we're doing this for all windows installs, correct? Is this desired?

dylan-hurd-oai, line 64: If I'm understanding correctly, we're doing this for all windows installs, correct? Is this desired?

dylan-hurd-oai (approved):
2 non-blocking comments, though I would strongly suggest we fast-follow on executable hashing

codex-rs/windows-sandbox-rs/src/helper_materialization.rs
iceweasel-oai: yeah I'll add the hashing in soon.

codex-rs/windows-sandbox-rs/src/helper_materialization.rs
iceweasel-oai: yeah, I tested it in the TUI and it works fine there too

PR #13414, opened 2026-03-04
feat(app-server): add a skills/changed v2 notification

owenlin0:
This adds a first-class app-server v2 `skills/changed` notification for the existing skills live-reload signal.

Before this change, clients only had the legacy raw `codex/event/skills_update_available` event. With this PR, v2 clients can listen for a typed JSON-RPC notification instead of depending on the legacy `codex/event/*` stream, which we want to remove soon.


owenlin0:
@codex review

codex-rs/app-server-protocol/schema/json/codex_app_server_protocol.schemas.json
```diff
@@ -6039,6 +6039,26 @@
           "title": "Thread/closedNotification",
           "type": "object"
         },
+        {
+          "properties": {
+            "method": {
+              "enum": [
+                "skills/changed"
```
pakrym-oai, line 6046: why changed vs updated?

owenlin0: I had `updated` in mind as well but codex pushed me away with this justification:

> `skills/changed` is a better fit for what this notification actually means.
> 
> updated implies the server is telling the client “here is the new authoritative state” or at least “the resource has been refreshed.” That’s not what we have here. This notification is just an invalidation signal coming from the watcher: something on disk changed, and the client should re-run skills/list if it cares.
> 
> changed is weaker and more accurate:
> - it doesn’t imply the server is pushing the updated skill list
> - it matches the underlying watcher concept (`SkillsChanged`)
> - it leaves room for changes that are additive, deletions, edits, enable/disable effects, or anything else that should cause a refresh

owenlin0: if you feel strongly happy to update! i'm ok either way I think

pakrym-oai (approved):

codex-rs/app-server-protocol/schema/json/codex_app_server_protocol.schemas.json
pakrym-oai: don't feel strongly

PR #13418, opened 2026-03-04
Add code_mode experimental feature

pakrym-oai:
A much narrower and more isolated (no node features) version of js_repl


docs/code_mode.md
```diff
@@ -0,0 +1,36 @@
+# Code Mode (`code_mode`)
+
+`code_mode` runs JavaScript in a Node-backed `node:vm` context.
```
viyatb-oai: does this still run under the codex platform specific sandbox?

pakrym-oai: No, we'll eventually run in v8 isolate.

cconger (approved):

PR #13424, opened 2026-03-04
feat(core, tracing): add a span representing a turn

owenlin0:
This is PR 3 of the app-server tracing rollout.

PRs https://github.com/openai/codex/pull/13285 and https://github.com/openai/codex/pull/13368 gave us inbound request spans in app-server and propagated trace context through Submission. This change finishes the next piece in core: when a request actually starts a turn, we now create a core-owned long-lived span that stays open for the real lifetime of the turn.

What changed:
- `Session::spawn_task` can now optionally create a long-lived turn span and run the spawned task inside it
- `turn/start` uses that path, so normal turn execution stays under a single core-owned span after the async handoff
- `review/start` uses the same pattern
- added a unit test that verifies the spawned turn task inherits the submission dispatch trace ancestry

**Why**
The app-server request span is intentionally short-lived. Once work crosses into core, we still want one span that covers the actual execution window until completion or interruption. This keeps that ownership where it belongs: in the layer that owns the runtime lifecycle.

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -4066,8 +4066,13 @@ mod handlers {
             sess.refresh_mcp_servers_if_requested(&current_context)
                 .await;
             let regular_task = sess.take_startup_regular_task().await.unwrap_or_default();
-            sess.spawn_task(Arc::clone(&current_context), items, regular_task)
-                .await;
+            sess.spawn_task(
+                Arc::clone(&current_context),
+                items,
+                regular_task,
+                Some("turn/start"),
+            )
+            .await;
         }
     }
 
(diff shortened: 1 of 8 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -4825,7 +4832,8 @@ async fn spawn_review_thread(
     // TODO(ccunningham): Review turns currently rely on `spawn_task` for TurnComplete but do not
     // emit a parent TurnStarted. Consider giving review a full parent turn lifecycle
     // (TurnStarted + TurnComplete) for consistency with other standalone tasks.
-    sess.spawn_task(tc.clone(), input, ReviewTask::new()).await;
+    sess.spawn_task(tc.clone(), input, ReviewTask::new(), Some("review/start"))
```
pakrym-oai: can this value be on ReviewTask ?

pakrym-oai (approved):
Let's move task name to the SessionTask trait

PR #13425, opened 2026-03-04
feat(app-server): support mcp elicitations in v2 api

owenlin0:
This adds a first-class server request for MCP server elicitations: `mcpServer/elicitation/request`.

Until now, MCP elicitation requests only showed up as a raw `codex/event/elicitation_request` event from core. That made it hard for v2 clients to handle elicitations using the same request/response flow as other server-driven interactions (like shell and `apply_patch` tools).

This also updates the underlying MCP elicitation request handling in core to pass through the full MCP request (including URL and form data) so we can expose it properly in app-server.

### Why not `item/mcpToolCall/elicitationRequest`?
This is because MCP elicitations are related to MCP servers first, and only optionally to a specific MCP tool call.

In the MCP protocol, elicitation is a server-to-client capability: the server sends `elicitation/create`, and the client replies with an elicitation result. RMCP models it that way as well.

In practice an elicitation is often triggered by an MCP tool call, but not always.

### What changed
- add `mcpServer/elicitation/request` to the v2 app-server API
- translate core `codex/event/elicitation_request` events into the new v2 server request
- map client responses back into `Op::ResolveElicitation` so the MCP server can continue
- update app-server docs and generated protocol schema
- add an end-to-end app-server test that covers the full round trip through a real RMCP elicitation flow
  - The new test exercises a realistic case where an MCP tool call triggers an elicitation, the app-server emits mcpServer/elicitation/request, the client accepts it, and the tool call resumes and completes successfully.

### app-server API flow
- Client starts a thread with `thread/start`.
- Client starts a turn with `turn/start`.
- App-server sends `item/started` for the `mcpToolCall`.
- While that tool call is in progress, app-server sends `mcpServer/elicitation/request`.
- Client responds to that request with `{ action: "accept" | "decline" | "cancel" }`.
- App-server sends `serverRequest/resolved`.
- App-server sends `item/completed` for the mcpToolCall.
- App-server sends `turn/completed`.
- If the turn is interrupted while the elicitation is pending, app-server still sends `serverRequest/resolved` before the turn finishes.


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -4031,6 +4016,49 @@ pub struct FileChangeRequestApprovalResponse {
     pub decision: FileChangeApprovalDecision,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "lowercase")]
```
bolinfest: I know it ends up the same, but should these be `camelCase`?

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -1989,6 +2014,60 @@ async fn on_request_user_input_response(
...
+    request_id: codex_protocol::mcp::RequestId,
+    pending_request_id: RequestId,
+    receiver: oneshot::Receiver<ClientRequestResult>,
+    conversation: Arc<CodexThread>,
+    thread_state: Arc<Mutex<ThreadState>>,
+    permission_guard: ThreadWatchActiveGuard,
+) {
+    let response = receiver.await;
+    resolve_server_request_on_thread_listener(&thread_state, pending_request_id).await;
+    drop(permission_guard);
```
bolinfest, line 2038: What's with this `drop()` across an async boundary? What is prevented until this is dropped?

codex-rs/app-server-protocol/src/protocol/v2.rs
owenlin0: yes, for consistency's sake let's update

codex-rs/app-server/src/bespoke_event_handling.rs
owenlin0: From codex:
> `permission_guard` is just a status-tracking RAII guard from `ThreadWatchManager`. Creating it increments the thread’s pending-approval count, which marks the thread status as `WaitingOnApproval`; dropping it decrements that count.
> 
> So the explicit `drop(permission_guard)` here is just ending that status as soon as the client response has been received and the pending server request has been resolved. It is not protecting anything async-sensitive; it prevents the thread from still appearing to be waiting on approval while we process the response afterward

owenlin0: looks like we have this pattern in the various handlers for server requests

owenlin0: the guard is specific to each server request, so multiple MCP elicitations should be ok

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -4031,6 +4016,49 @@ pub struct FileChangeRequestApprovalResponse {
...
+#[ts(export_to = "v2/")]
+pub struct McpServerElicitationRequestParams {
+    pub thread_id: String,
+    /// Active Codex turn when the elicitation was observed, if available.
+    ///
+    /// This is nullable because MCP defines elicitation as a general
+    /// server-to-client request, not a turn- or tool-call-scoped primitive.
+    pub turn_id: Option<String>,
+    pub server_name: String,
+    pub message: String,
```
mzeng-openai: Can we use the standard MCP elicit request payload here? I think that's the rich metadata that the group was talking about. It already comes with rmcp: `rmcp::Model::CreateElicitationRequestParams`

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -4031,6 +4016,49 @@ pub struct FileChangeRequestApprovalResponse {
...
+    // TODO: If the MCP elicitation is triggered by an MCP tool call, we should expose the McpToolCall
+    // item ID as well. However, we need to update core to plumb this through first.
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct McpServerElicitationRequestResponse {
+    pub action: McpServerElicitationAction,
```
mzeng-openai, line 4192: Same here, can we use the standard MCP elicit response?

codex-rs/app-server-protocol/src/protocol/v2.rs
owenlin0: had to define a copy of it in v2.rs to be able to export it properly to typescript and json schema, but it's the same shape

codex-rs/app-server-protocol/src/protocol/v2.rs
owenlin0: same as above - had to define a copy of it in v2.rs to be able to export it properly to typescript and json schema, but it's the same shape

mzeng-openai (approved):

PR #13432, opened 2026-03-04
refactor: route zsh-fork through unified exec

bolinfest:
## Why

When `unified_exec` and `shell_zsh_fork` are enabled together, unified-exec sessions need to stay on the configured zsh-fork binary so exec interception remains reliable for approvals, execpolicy, and escalation behavior.

In practice, those sessions could drift away from the configured `zsh_path` in a few ways:

- routing could fall back to the wrong tool surface when both features were enabled
- explicit `exec_command.shell` overrides could request a different shell
- nested `exec zsh` calls could hop onto the host `zsh`
- privileged login-shell sessions could surface startup helpers like `/usr/libexec/path_helper` as separate escalated execs
- privileged `exec_command` sessions could let later `write_stdin` subcommands inherit that broader privilege instead of dropping back to turn-default policy
- zsh-fork escalation-session resources could stay resident until the model polled again

That weakens the guarantees the zsh-fork path is supposed to provide and makes privileged interactive sessions much noisier than intended.

## What Changed

- Kept zsh-fork sessions on `exec_command` / unified exec when both features are enabled, and hid the `shell` parameter from the `exec_command` schema when the session is actively using zsh-fork (`core/src/tools/spec.rs`, `core/src/tools/handlers/unified_exec.rs`).
- In active zsh-fork unified-exec sessions, ignore model-provided shell overrides and always launch the configured `zsh_path` (`core/src/tools/handlers/unified_exec.rs`).
- Replaced permissive shell-shape scanning with deterministic parsing of the original shell argv (`program`, `-c` / `-lc`, `script`) before preparing the zsh-fork unified-exec path (`core/src/tools/runtimes/shell/unix_escalation.rs`, `core/src/tools/runtimes/shell/zsh_fork_backend.rs`, `core/src/tools/runtimes/unified_exec.rs`).
- For privileged unified-exec zsh-fork launches, build an exact intercepted inner `zsh -c ...` target instead of heuristically matching later startup execs. When a matching shell snapshot exists for a login-shell request, that inner command now sources the snapshot, reapplies explicit env overrides, and then runs the original script. This preserves login-derived shell state without rerunning startup helpers under zsh-fork interception (`core/src/tools/runtimes/mod.rs`, `core/src/tools/runtimes/shell/unix_escalation.rs`).
- Kept nested intercepted `zsh` execs on zsh-fork by:
  - recognizing approved host `zsh` installations
  - forcing escalation when an intercepted exec targets host `zsh` instead of the configured `zsh_path`
  - rewriting that intercepted program path back to the configured `zsh_path`
  (`core/src/tools/runtimes/shell/unix_escalation.rs`).
- Kept skill lookup pinned to the turn `cwd` and re-read execpolicy for intercepted execs so approved amendments continue to apply to later matching subcommands in the same session (`core/src/tools/runtimes/shell/unix_escalation.rs`).
- Ensured `write_stdin` subcommands in a privileged zsh-fork unified-exec TTY fall back to turn-default sandbox policy unless separately allowed, while still honoring matching `prefix_rule()` approvals (`core/src/tools/runtimes/shell/unix_escalation.rs`, `core/tests/suite/unified_exec.rs`).
- Released zsh-fork escalation-session resources when unified-exec processes exit, even if the model never polls the session again (`core/src/unified_exec/process.rs`, `core/src/tools/runtimes/shell/zsh_fork_backend.rs`).
- Expanded coverage for the routed unified-exec zsh-fork path:
  - app-server v2 turn-start coverage, including interrupt cleanup and the login-startup-helper regression (`app-server/tests/suite/v2/turn_start_zsh_fork.rs`)
  - approval / amendment behavior (`core/tests/suite/approvals.rs`)
  - skill prompt and turn-`cwd` behavior (`core/tests/suite/skill_approval.rs`)
  - interactive unified-exec behavior, including executable-path assertions, nested `zsh` rewrite coverage, and the privileged-`write_stdin` sandbox regression (`core/tests/suite/unified_exec.rs`)
  - unit tests for shell parsing, snapshot-backed exact matching, shell-override handling, and `zsh` rewrite predicates (`core/src/tools/handlers/unified_exec_tests.rs`, `core/src/tools/runtimes/shell/unix_escalation_tests.rs`)
- Added Windows-safe `#[cfg(unix)]` usage around zsh-fork-only integration coverage.

## Verification

- `cargo test -p codex-core build_exact_match_zsh_fork_reexec --lib`
- `cargo test -p codex-core maybe_wrap_shell_lc_with_snapshot --lib`
- `cargo test -p codex-core --test all unified_exec_zsh_fork_write_stdin_reverts_to_default_sandbox_and_honors_prefix_rule`
- `cargo test -p codex-app-server --test all turn_start_shell_zsh_fork_login_startup_helper_does_not_prompt_separately_v2`

## Manual Testing

Run Codex built from source with both features enabled as follows:

```shell
just codex --enable unified_exec --enable shell_zsh_fork --config zsh_path=$(realpath codex-rs/app-server/tests/suite/zsh)
```

Then ask Codex to do something like:

```text
Start a new instance of a shell using exec_command where it just runs `ls` but keep it around.
```

Then tell it:

```text
Run python3 in the shell. Once the REPL starts, run `import os; print(os.getpid())` and tell me what you get.
```

In my case, it replied with:

```text
python3 started in the existing shell session, and import os; print(os.getpid()) printed:

55692
```

Then over in a plain shell, I checked `pstree` and confirmed that the Python process was still running under the configured zsh-fork shell rather than the host shell.

### DIFF
--- a/codex-rs/core/src/tools/handlers/unified_exec.rs
+++ b/codex-rs/core/src/tools/handlers/unified_exec.rs
@@ -349,11 +349,19 @@ pub(crate) fn get_command(
             let shell = model_shell.as_ref().unwrap_or(session_shell.as_ref());
             Ok(shell.derive_exec_args(&args.cmd, use_login_shell))
         }
-        UnifiedExecShellMode::ZshFork(zsh_fork_config) => Ok(vec![
-            zsh_fork_config.shell_zsh_path.to_string_lossy().to_string(),
-            if use_login_shell { "-lc" } else { "-c" }.to_string(),
-            args.cmd.clone(),
-        ]),
+        UnifiedExecShellMode::ZshFork(zsh_fork_config) => {
+            if args.shell.is_some() {
+                return Err(
+                    "shell override is not supported when the zsh-fork backend is enabled."
+                        .to_string(),
+                );
+            }
+            Ok(vec![
+                zsh_fork_config.shell_zsh_path.to_string_lossy().to_string(),
+                if use_login_shell { "-lc" } else { "-c" }.to_string(),
+                args.cmd.clone(),
+            ])
+        }
     }
 }
 
--- a/codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
+++ b/codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
@@ -15,6 +15,7 @@ use crate::shell::ShellType;
 use crate::skills::SkillMetadata;
 use crate::tools::runtimes::ExecveSessionApproval;
 use crate::tools::runtimes::build_command_spec;
+use crate::tools::runtimes::maybe_build_snapshot_restore_preamble;
 use crate::tools::sandboxing::SandboxAttempt;
 use crate::tools::sandboxing::SandboxablePreference;
 use crate::tools::sandboxing::ToolCtx;
@@ -36,6 +37,7 @@ use codex_protocol::protocol::ReviewDecision;
 use codex_protocol::protocol::SandboxPolicy;
 use codex_shell_command::bash::parse_shell_lc_plain_commands;
 use codex_shell_command::bash::parse_shell_lc_single_command_prefix;
+use codex_shell_command::parse_command::shlex_join;
 use codex_shell_escalation::EscalateServer;
 use codex_shell_escalation::EscalationDecision;
 use codex_shell_escalation::EscalationExecution;
@@ -50,16 +52,26 @@ use codex_shell_escalation::ShellCommandExecutor;
 use codex_shell_escalation::Stopwatch;
 use codex_utils_absolute_path::AbsolutePathBuf;
 use std::collections::HashMap;
+use std::ffi::OsStr;
+use std::path::Path;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
-use tokio::sync::RwLock;
 use tokio_util::sync::CancellationToken;
 use uuid::Uuid;
 
 pub(crate) struct PreparedUnifiedExecZshFork {
     pub(crate) exec_request: ExecRequest,
     pub(crate) escalation_session: EscalationSession,
+}
+
+/// Rewrites a unified-exec shell launch so login-shell startup runs in an
+/// outer shell, while the original script is executed by an exact matched
+/// inner zsh-fork subprocess.
+#[derive(Clone, Debug, Eq, PartialEq)]
+struct ExactMatchZshForkReexec {
+    outer_command: Vec<String>,
+    exact_intercepted_command: Vec<String>,
 }
 
 const PROMPT_CONFLICT_REASON: &str =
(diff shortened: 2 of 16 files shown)


codex-rs/core/src/tools/spec.rs
```diff
@@ -110,15 +110,15 @@ impl ToolsConfig {
...
-        } else if features.enabled(Feature::ShellZshFork) {
-            ConfigShellToolType::ShellCommand
         } else if features.enabled(Feature::UnifiedExec) {
             // If ConPTY not supported (for old Windows versions), fallback on ShellCommand.
             if codex_utils_pty::conpty_supported() {
                 ConfigShellToolType::UnifiedExec
             } else {
                 ConfigShellToolType::ShellCommand
             }
+        } else if features.enabled(Feature::ShellZshFork) {
```
jif-oai: this will send out of distribution the sessions that had `Feature::ShellZshFork == true` because `unified_exec` will step in first

jif-oai (approved):

codex-rs/core/src/tools/spec.rs
bolinfest: Above, `unified_exec_backend` is set to `UnifiedExecBackendConfig::ZshFork` when `features.enabled(Feature::ShellTool) && features.enabled(Feature::ShellZshFork)` is set, so I think it works out?

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
jif-oai: `proposed_execpolicy_amendment` is set to `None` here. This means we now never have amendment. Is this expected (maybe unrelated to this PR but catching this now)

codex-rs/core/src/tools/handlers/unified_exec.rs
```diff
@@ -286,12 +289,17 @@ pub(crate) fn get_command(
...
     session_shell: Arc<Shell>,
     allow_login_shell: bool,
+    unified_exec_backend: UnifiedExecBackendConfig,
 ) -> Result<Vec<String>, String> {
-    let model_shell = args.shell.as_ref().map(|shell_str| {
-        let mut shell = get_shell_by_model_provided_path(&PathBuf::from(shell_str));
-        shell.shell_snapshot = crate::shell::empty_shell_snapshot_receiver();
-        shell
-    });
+    let model_shell = if unified_exec_backend == UnifiedExecBackendConfig::ZshFork {
```
jif-oai: We should drop the `args.shell` from the tool spec then. Otherwise, the model can ask for bash and still end up in zsh

codex-rs/core/src/tools/runtimes/shell/zsh_fork_backend.rs
jif-oai: We need a weak somewhere and a dropper

This structure is held in `UnifiedExecProcess` as `_spawn_lifecycle: SpawnLifecycleHandle,` and this structure holds the server-side socket 

BUT, `unified_exec` processes don't get drop when they exit. They get dropped only when they exit AND the model polled them (either directly if during an `exec_command` or after the last `write_stdin`)
But if the model "forget" about the process or do not poll it, we keep everything and so we keep the `socket` which is, I think, a bad residency practice

wdyt?

codex-rs/core/src/tools/spec.rs
bolinfest: The routing is still gated by backend selection. `unified_exec` does take precedence when enabled, but in this config the backend is explicitly `UnifiedExecBackendConfig::ZshFork`, so sessions still use the zsh-fork path. I also removed `shell` from the `exec_command` tool schema when that backend is active to avoid behavioral mismatch.

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
```diff
@@ -87,11 +99,55 @@ fn approval_sandbox_permissions(
...
+) -> ExactMatchZshForkReexec {
+    let snapshot_restore_preamble = shell_command
+        .login
+        .then(|| maybe_build_snapshot_restore_preamble(session_shell, cwd, explicit_env_overrides))
+        .flatten();
+    let exact_intercepted_script = snapshot_restore_preamble
+        .as_ref()
+        .map(|preamble| format!("{preamble}\n\n{}", shell_command.script))
+        .unwrap_or_else(|| shell_command.script.clone());
+    let exact_intercepted_command = vec![
```
jif-oai, line 126: I think the exact match will create some issues... what if you don't have a shell snapshot available for example? Then we can still have a `-lc`

This will basically break the pre-approved path

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
```diff
@@ -223,44 +283,92 @@ pub(super) async fn try_run_zsh_fork(
...
-        ctx.session.services.exec_policy.current().as_ref().clone(),
-    ));
+    // The outer shell may be the snapshot bootstrap shell instead of the
+    // configured zsh binary, so only validate the `[program, -c/-lc, script]`
+    // argv shape here.
+    let exact_match_reexec = if matches!(
+        req.sandbox_permissions,
+        SandboxPermissions::RequireEscalated
+    ) {
+        let mut explicit_env_overrides = req.env.clone();
```
jif-oai, line 307: this changes the contract of the shell snapshot. It means you will restore something coming from before the shell_snapshot on top of the shell_snapshot

PR #13437, opened 2026-03-04
[js_repl] Support local ESM file imports

aaronl-openai:
## Summary
- add `js_repl` support for dynamic imports of relative and absolute local ESM `.js` / `.mjs` files
- keep bare package imports on the native Node path and resolved from REPL-global search roots (`CODEX_JS_REPL_NODE_MODULE_DIRS`, then `cwd`), even when they originate from imported local files
- restrict static imports inside imported local files to other local relative/absolute `.js` / `.mjs` files, and surface a clear error for unsupported top-level static imports in the REPL cell
- run imported local files inside the REPL VM context so they can access `codex.tmpDir`, `codex.tool`, captured `console`, and Node-like `import.meta` helpers
- reload local files between execs so later `await import("./file.js")` calls pick up edits and fixed failures, while preserving package/builtin caching and persistent top-level REPL bindings
- make `import.meta.resolve()` self-consistent by allowing the returned `file://...` URLs to round-trip through `await import(...)`
- update both public and injected `js_repl` docs to clarify the narrowed contract, including global bare-import resolution behavior for local absolute files

## Testing
- `cargo test -p codex-core js_repl_`
- built codex binary and verified behavior


codex-rs/core/src/tools/js_repl/kernel.js
fjord-oai, line 1: from my local codex conversation:

[P3] Top-level static local imports still fail with a low-level VM error. In [kernel.js (line 1004)](app://-/index.html?hostId=local#) the cell linker returns importResolved(...); for file specifiers, [that path (line 432)](app://-/index.html?hostId=local#) returns module.namespace, not a vm.Module. So import "./foo.js" still trips ERR_VM_MODULE_NOT_MODULE rather than a purposeful js_repl message.

fjord-oai, line 1: codex also suggested:

I’d add one runtime test proving an imported local file can see REPL globals such as codex.tmpDir or captured console. That’s the behavior a future “simplify to native import” refactor is most likely to break.

fjord-oai (approved):
lgtm after comments

PR #13446, opened 2026-03-04
[tui] Show speed in session header

pash-openai:
- add a speed row to the startup/session header under the model row
- render the speed row with the same styling pattern as the model row, using /fast to change
- show only Fast or Standard to users and update the affected snapshots


jif-oai (approved):
Is there anything that says `fast` is more expensive? I think we should tell the user in some way

pakrym-oai (approved):

PR #13448, opened 2026-03-04
seatbelt: honor split filesystem sandbox policies

bolinfest:
## Why

After `#13440` and `#13445`, macOS Seatbelt policy generation was still deriving filesystem and network behavior from the legacy `SandboxPolicy` projection.

That projection loses explicit unreadable carveouts and conflates split network decisions, so the generated Seatbelt policy could still be wider than the split policy that Codex had already computed.

## What changed

- added Seatbelt entrypoints that accept `FileSystemSandboxPolicy` and `NetworkSandboxPolicy` directly
- built read and write policy stanzas from access roots plus excluded subpaths so explicit unreadable carveouts survive into the generated Seatbelt policy
- switched network policy generation to consult `NetworkSandboxPolicy` directly
- failed closed when managed-network or proxy-constrained sessions do not yield usable loopback proxy endpoints
- updated the macOS callers and test helpers that now need to carry the split policies explicitly

## Verification

- added regression coverage in `core/src/seatbelt.rs` for unreadable carveouts under both full-disk and scoped-readable policies
- verified the current PR state with `just clippy`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/13448).
* #13453
* #13452
* #13451
* #13449
* __->__ #13448
* #13445
* #13440
* #13439


codex-rs/core/src/seatbelt.rs
```diff
@@ -305,16 +331,75 @@ pub(crate) fn create_seatbelt_command_args(
...
         None,
     )
 }
 
+fn root_absolute_path() -> AbsolutePathBuf {
+    match AbsolutePathBuf::from_absolute_path(Path::new("/")) {
+        Ok(path) => path,
+        Err(err) => panic!("root path must be absolute: {err}"),
+    }
+}
```
dylan-hurd-oai, line 350: This should probably be a method on `AbsolutePathBuf`?

dylan-hurd-oai (approved):
1 non-blocking comment from me, but also let's see what codex says about the current version 😄

codex-rs/core/src/seatbelt.rs
viyatb-oai: probably but we only use it here in seatbelt. trying to reduce number of edits outside this file :(

PR #13454, opened 2026-03-04
add metric for per-turn token usage

jif-oai:


pash-openai (approved):
terminal bench will enjoy this

PR #13465, opened 2026-03-04
Standardize context fragment handling

charley-oai:
## Summary
- Standardizes model-visible context injection around shared typed fragments and two envelopes:
  - `ModelVisibleContextFragment`
  - `ModelVisibleContextRole` (`DeveloperContextRole` / `ContextualUserContextRole`)
  - `DeveloperEnvelopeBuilder` / `ContextualUserEnvelopeBuilder`
- Unifies turn-state context assembly for both `build_initial_context` and `build_settings_update_items` under a single ordered registry in `core/src/context_manager/updates.rs` (`REGISTERED_TURN_STATE_FRAGMENT_BUILDERS`).
- Simplifies `TurnContextDiffFragment` to one entrypoint:
  - `build(turn_context, reference_context_item, params)`
  This removes pass-based boilerplate and keeps fragment extraction/diff/render logic together on each fragment type.
- Removes old dual-registry scaffolding (`contextual_user_update_fragments.rs`, pass enums/adapter plumbing) and keeps one obvious place to register new turn-state fragments.
- Preserves explicit text boundaries between content items (`\n\n`) because adjacent text content items are concatenated model-side.

## Standardized Rules
- All model-visible injected context is represented as a typed fragment.
- There are exactly two model-visible envelopes: one developer message and one contextual-user message.
- Contextual-user fragments use stable markers so history parsing/diffing can recognize them as contextual state.
- Fragments derived from turn-state that need reinjection/diff behavior implement `TurnContextDiffFragment`.
- New model-visible context should not hand-build raw `ResponseItem::Message` payloads outside envelope builders.

## Additional Fixes Included
- Split custom override semantics from system developer context via `CustomDeveloperInstructions` (with compatibility aliasing retained where needed).
- Kept `developerInstructions` API docs aligned in app-server protocol schema/readme.
- Fixed permission-policy edge cases:
  - prompt on unmatched `with_additional_permissions` requests in restricted/on-request flows
  - include `request_permissions` in reject-policy instructions
- Aligned naming/docs for contextual-user fragments (`AgentsMdInstructions`, plugin-instructions naming, model-visible docs).

## Docs
- `docs/model-visible-context.md` now documents the single registry + one-method diff trait path.
- `AGENTS.md` now points contributors to the same blessed registration path.
- App-server docs/schema text was updated to match current `developerInstructions` behavior.

## Validation
- `just fmt`
- Attempted targeted `codex-core` tests after rebase; current branch state still has compile errors in `codex-core` tests (for example missing `DeveloperInstructions` import in `core/src/codex_tests.rs` and `RejectConfig` field update in `core/src/exec_policy.rs`) that need follow-up in this PR series.


codex-rs/core/src/session_prefix.rs
```diff
@@ -2,15 +2,62 @@ use codex_protocol::protocol::AgentStatus;
...
+        SUBAGENTS_FRAGMENT.wrap_body(lines)
+    }
+}
+
+struct SubagentNotification<'a> {
+    agent_id: &'a str,
+    status: &'a AgentStatus,
+}
+
+impl ContextualUserFragment for SubagentNotification<'_> {
```
pakrym-oai: why is this user?

codex-rs/core/src/contextual_user_message.rs
```diff
@@ -63,6 +71,30 @@ impl ContextualUserFragmentDefinition {
...
+    fn into_content_item(self) -> ContentItem
+    where
+        Self: Sized,
+    {
+        ContentItem::InputText {
+            text: self.serialize_to_text(),
+        }
+    }
+
+    fn into_response_item(self) -> ResponseItem
```
pakrym-oai: why do we create response item just to then discard the item? Especially considering the builder uses into_content_item

codex-rs/core/src/commit_attribution.rs
```diff
@@ -7,11 +9,11 @@ fn build_commit_message_trailer(config_attribution: Option<&str>) -> Option<Stri
 
 pub(crate) fn commit_message_trailer_instruction(
     config_attribution: Option<&str>,
-) -> Option<String> {
+) -> Option<DeveloperInstructions> {
```
pakrym-oai: Why do we need both ContextualUserFragment and DeveloperInstructions? The only difference is the `role` on the message. No point to have two parallel trees of types

pakrym-oai: We can also encode role in the fragment definition.

codex-rs/core/src/contextual_user_message.rs
charley-oai: Yeah good point. Deleted `into_response_item()` from the trait. The shared builders keep
using `into_content_item()`, and the few places that genuinely need a full message now construct it explicitly through `fragment.spec().into_message(fragment.render_text())`.

codex-rs/core/src/commit_attribution.rs
charley-oai: Created a shared `ModelVisibleFragment` abstraction to unify these

codex-rs/core/src/session_prefix.rs
charley-oai: I think it should be developer, will change

charley-oai:
@codex review this

codex-rs/core/src/model_visible_context.rs
```diff
@@ -0,0 +1,265 @@
...
+    ModelVisibleContextFragmentSpec::contextual_user(PLUGINS_OPEN_TAG, PLUGINS_CLOSE_TAG);
+
+const CONTEXTUAL_USER_FRAGMENT_SPECS: &[ModelVisibleContextFragmentSpec] = &[
+    AGENTS_MD_FRAGMENT_SPEC,
+    ENVIRONMENT_CONTEXT_FRAGMENT_SPEC,
+    SKILL_FRAGMENT_SPEC,
+    USER_SHELL_COMMAND_FRAGMENT_SPEC,
+    TURN_ABORTED_FRAGMENT_SPEC,
+    PLUGINS_FRAGMENT_SPEC,
+];
```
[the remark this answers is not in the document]
charley-oai: Not important

codex-rs/core/src/agent/control.rs
```diff
@@ -587,7 +588,7 @@ mod tests {
             let ResponseItem::Message { role, content, .. } = item else {
                 return false;
             };
-            if role != "user" {
+            if role != "developer" {
```
charley-oai: changing subagent message to dev

codex-rs/core/tests/suite/plugins.rs
```diff
@@ -196,8 +205,15 @@ async fn plugin_skills_append_to_instructions() -> Result<()> {
         "expected path {expected_path_str} in instructions"
     );
     assert!(
-        instructions_text.find("## Plugins") < instructions_text.find("## Skills"),
```
charley-oai: after the refactor, ## Plugins is no longer in instructions_text at all.

That assertion used to check ordering within one string. Now we intentionally split into two user fragments:

- AGENTS fragment (instructions_text)
- plugins fragment (plugin_text)

So we replaced it with stronger checks in plugins.rs:

- instructions_text does not contain ## Plugins
- plugin_text does contain ## Plugins
- plugin fragment appears after AGENTS fragment in request input order

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

PR #13494, opened 2026-03-04
Align SQLite feedback logs with feedback formatter

charley-oai:
## Summary
- store a pre-rendered `feedback_log_body` in SQLite so `/feedback` exports keep span prefixes and structured event fields
- render SQLite feedback exports with timestamps and level prefixes to match the old in-memory feedback formatter, while preserving existing trailing newlines
- count `feedback_log_body` in the SQLite retention budget so structured or span-prefixed rows still prune correctly
- bound `/feedback` row loading in SQL with the retention estimate, then apply exact whole-line truncation in Rust so uploads stay capped without splitting lines

## Details
- add a `feedback_log_body` column to `logs` and backfill it from `message` for existing rows
- capture span names plus formatted span and event fields at write time, since SQLite does not retain enough structure to reconstruct the old formatter later
- keep SQLite feedback queries scoped to the requested thread plus same-process threadless rows
- restore a SQL-side cumulative `estimated_bytes` cap for feedback export queries so over-retained partitions do not load every matching row before truncation
- add focused formatting coverage for exported feedback lines and parity coverage against `tracing_subscriber`

## Testing
- cargo test -p codex-state
- just fix -p codex-state
- just fmt

codex author: `codex resume 019ca1b0-0ecc-78b1-85eb-6befdd7e4f1f`


codex-rs/state/src/log_db.rs
```diff
@@ -228,6 +239,69 @@ where
     thread_id
 }
 
+fn format_feedback_log_body<S>(
```
jif-oai, line 242: Can we try to have only one formatter then?

codex-rs/state/src/runtime/logs.rs
```diff
@@ -13,10 +13,20 @@ impl StateRuntime {
...
+            "INSERT INTO logs (ts, ts_nanos, level, target, message, feedback_log_body, thread_id, process_uuid, module_path, file, line, estimated_bytes) ",
         );
         builder.push_values(entries, |mut row, entry| {
-            let estimated_bytes = entry.message.as_ref().map_or(0, String::len) as i64
+            let feedback_log_body = entry.feedback_log_body.as_ref().or(entry.message.as_ref());
+            // Keep about 10 MiB of reader-visible log content per partition.
+            // `query_logs` uses `message`, while `/feedback` exports
+            // `feedback_log_body`, so charge the larger payload without
+            // double-counting the same event.
+            let estimated_bytes = entry
```
jif-oai: Shouldn't we consider both payload in the size compute?

charley-oai: We drop `message` now

codex-rs/state/logs_migrations/0002_logs_feedback_log_body.sql
```diff
@@ -0,0 +1,54 @@
...
+    file,
+    line,
+    thread_id,
+    process_uuid,
+    estimated_bytes
+FROM logs_old;
+
+DROP TABLE logs_old;
+
+CREATE INDEX idx_logs_ts ON logs(ts DESC, ts_nanos DESC, id DESC);
```
jif-oai, line 49: This is a lot of index. Might be ok but be careful that this will drastically slow down insertion time

jif-oai (approved):

codex-rs/state/logs_migrations/0002_logs_feedback_log_body.sql
charley-oai: Will trim `idx_logs_process_uuid`, the rest are useful I think

PR #13496, opened 2026-03-04
feat: merge skill permission profiles into the turn sandbox for zsh-fork execs

celia-oai:
## Summary

This changes the Unix shell escalation path for skill-matched executables to apply a skill's `PermissionProfile` as additive permissions on top of the existing turn/request sandbox policy.

Previously, skill-matched executables compiled the skill permission profile into a standalone sandbox policy and executed against that replacement policy. Now they go through the same `additional_permissions` merge path used elsewhere in shell sandbox preparation.

## What Changed

- Changed `skill_escalation_execution()` to return `EscalationPermissions::PermissionProfile(...)` for non-empty skill permission profiles.
- Kept empty or missing skill permission profiles on the `TurnDefault` path.
- Added tests covering the new additive skill-permission behavior.
- Added inline comments in `prepare_escalated_exec()` clarifying the difference between additive permission merging and fully specified replacement sandbox policies.
- Removed the now-unused skill permission compiler module after switching this path away from standalone compiled skill sandbox policies.

## Testing

- Ran `just fmt` in `codex-rs`
- Ran `cargo test -p codex-core`

`cargo test -p codex-core` still hits an unrelated existing failure: `shell_snapshot::tests::snapshot_shell_does_not_inherit_stdin`

## Follow-up

This change intentionally does not merge skill-specific macOS seatbelt profile extensions through the `additional_permissions` path yet. Filesystem and network permissions now follow the additive merge path, but seatbelt extension permissions still need separate handling in a follow-up PR.


codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
```diff
@@ -228,15 +227,13 @@ impl CoreShellActionProvider {
...
+        skill
+            .permission_profile
+            .clone()
+            .filter(|permission_profile| !permission_profile.is_empty())
+            .map(|permission_profile| {
+                EscalationExecution::Permissions(EscalationPermissions::PermissionProfile(
+                    permission_profile,
                 ))
             })
             .unwrap_or(EscalationExecution::TurnDefault)
```
bolinfest: The clone() is only necessary when non-empty, right? So might this be clearer?

```suggestion
        if skill.permission_profile.is_empty() {
            EscalationExecution::TurnDefault
        } else {
            EscalationExecution::Permissions(EscalationPermissions::PermissionProfile(
                    permission_profile.clone()
           ))
       }
```

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
```diff
@@ -718,15 +715,19 @@ impl ShellCommandExecutor for CoreShellCommandExecutor {
...
-                command,
-                workdir,
-                env,
-                &self.sandbox_policy,
-                Some(permission_profile),
-                None,
-            )?,
+            )) => {
+                // Merge additive permissions into the existing turn/request sandbox policy.
+                self.prepare_sandboxed_exec(
```
bolinfest: I think what would help improve clarity is if `prepare_sandboxed_exec()` were updated to take a struct because then it would be clear that `Some(permission_profile)` is passed as `additional_permissions` and that `macos_seatbelt_profile_extensions` is `None`.

dylan-hurd-oai (approved):

bolinfest (approved):

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
celia-oai: actually, i just realized we have two redundant structs?
```
#[derive(Debug, Clone, Default, Eq, Hash, PartialEq, Serialize, Deserialize, JsonSchema, TS)]
pub struct MacOsPermissions {
    pub preferences: Option<MacOsPreferencesValue>,
    pub automations: Option<MacOsAutomationValue>,
    pub accessibility: Option<bool>,
    pub calendar: Option<bool>,
}
```

and

```

#[derive(Debug, Clone, PartialEq, Eq, Default)]
pub struct MacOsSeatbeltProfileExtensions {
    pub macos_preferences: MacOsPreferencesPermission,
    pub macos_automation: MacOsAutomationPermission,
    pub macos_accessibility: bool,
    pub macos_calendar: bool,
}
```
should I just keep the second one?

celia-oai: will do that in the second pr

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
celia-oai: agreed this is cleaner

PR #13499, opened 2026-03-04
core/protocol: add structured macOS additional permissions and merge them into sandbox execution

celia-oai:
## Summary
- Introduce strongly-typed macOS additional permissions across protocol/core/app-server boundaries.
- Merge additional permissions into effective sandbox execution, including macOS seatbelt profile extensions.
- Expand docs, schema/tool definitions, UI rendering, and tests for `network`, `file_system`, and `macos` additional permissions.


codex-rs/core/src/sandboxing/macos_permissions.rs
```diff
@@ -0,0 +1,246 @@
...
+            macos_calendar: base.macos_calendar || requested.macos_calendar,
+        }),
+        None => Some(requested),
+    }
+}
+
+#[cfg(target_os = "macos")]
+pub(crate) fn build_macos_seatbelt_profile_extensions(
+    permissions: &MacOsPermissions,
+) -> Option<MacOsSeatbeltProfileExtensions> {
```
dylan-hurd-oai: I think this should just be a `From` impl? Since it always returns Some(MacOsSeatbeltProfileExtensions)

codex-rs/core/src/sandboxing/macos_permissions.rs
```diff
@@ -0,0 +1,246 @@
...
+use codex_protocol::models::MacOsAutomationPermission;
+#[cfg(target_os = "macos")]
+use codex_protocol::models::MacOsAutomationValue;
+use codex_protocol::models::MacOsPermissions;
+use codex_protocol::models::MacOsPreferencesPermission;
+#[cfg(target_os = "macos")]
+use codex_protocol::models::MacOsPreferencesValue;
+use codex_protocol::models::MacOsSeatbeltProfileExtensions;
+#[cfg(target_os = "macos")]
+use tracing::warn;
```
bolinfest: Note: I would just use the fully-qualified `tracing::warn()` to avoid the conditional compilation here.

The other `#[cfg(target_os = "macos")]` are somewhat understandable.

codex-rs/core/src/sandboxing/macos_permissions.rs
```diff
@@ -0,0 +1,246 @@
...
+
+pub(crate) fn merge_macos_seatbelt_profile_extensions(
+    base: Option<&MacOsSeatbeltProfileExtensions>,
+    permissions: Option<&MacOsPermissions>,
+) -> Option<MacOsSeatbeltProfileExtensions> {
+    let Some(permissions) = permissions else {
+        return base.cloned();
+    };
```
dylan-hurd-oai, line 15: this feels a bit messy to accept `base` as an `Option` here, but i guess if base is None and permissions is Some it's reasonable? Feels like we might want a helper and a core logic function for the merge?

codex-rs/core/src/sandboxing/macos_permissions.rs
```diff
@@ -0,0 +1,246 @@
...
+    _: &MacOsPermissions,
+) -> Option<MacOsSeatbeltProfileExtensions> {
+    None
+}
+
+/// Merges two preferences permissions by keeping the more permissive one.
+///
+/// The larger rank wins: `None < ReadOnly < ReadWrite`. When both sides have
+/// the same rank, this keeps `base`.
+fn merge_macos_preferences_permission(
```
dylan-hurd-oai: Instead of merge, can we call this union or something to indicate we're taking the more permissive? I'm assuming we we'll eventually need something similar in a restrictive sense as well

codex-rs/core/src/sandboxing/macos_permissions.rs
```diff
@@ -0,0 +1,246 @@
...
+                MacOsAutomationPermission::None
+            } else {
+                MacOsAutomationPermission::BundleIds(bundle_ids)
+            }
+        }
+        None => default,
+    }
+}
+
+#[cfg(test)]
```
bolinfest: I think you want this?

```suggestion
#[cfg(all(test, target_os = "macos"))]
```

codex-rs/core/src/sandboxing/macos_permissions.rs
```diff
@@ -0,0 +1,246 @@
...
+    base: &MacOsPreferencesPermission,
+    requested: &MacOsPreferencesPermission,
+) -> MacOsPreferencesPermission {
+    match preferences_permission_rank(base).cmp(&preferences_permission_rank(requested)) {
+        Ordering::Less => requested.clone(),
+        Ordering::Equal | Ordering::Greater => base.clone(),
+    }
+}
+
+fn preferences_permission_rank(permission: &MacOsPreferencesPermission) -> u8 {
```
dylan-hurd-oai: Couldn't we just use `#[derive(PartialOrd, Ord)]` macros for this?

codex-rs/core/src/sandboxing/macos_permissions.rs
```diff
@@ -0,0 +1,246 @@
...
+        MacOsPreferencesPermission::ReadOnly => 1,
+        MacOsPreferencesPermission::ReadWrite => 2,
+    }
+}
+
+/// Merges two automation permissions by keeping the more permissive result.
+///
+/// `All` wins over everything, `None` yields to the other side, and two bundle
+/// ID allowlists are unioned together.
+fn merge_macos_automation_permission(
```
dylan-hurd-oai: Ditto `union` comment above, merge feels ambiguous as to whether we're doing `&` or `|`

codex-rs/core/src/sandboxing/macos_permissions.rs
```diff
@@ -0,0 +1,246 @@
...
+            MacOsSeatbeltProfileExtensions {
+                macos_preferences: MacOsPreferencesPermission::ReadOnly,
+                ..Default::default()
+            }
+        );
+    }
+
+    #[cfg(target_os = "macos")]
+    #[test]
+    fn merge_extensions_widens_permissions() {
```
dylan-hurd-oai, line 89: Could we add 1-2 more test cases here?

codex-rs/core/src/sandboxing/mod.rs
```diff
@@ -502,6 +552,35 @@ mod tests {
...
+                calendar: Some(true),
+            }),
+            ..Default::default()
+        })
+        .expect("permissions");
+
+        assert_eq!(
+            permissions.macos,
+            Some(MacOsPermissions {
+                preferences: Some(MacOsPreferencesValue::Mode("readwrite".to_string())),
```
bolinfest: Why is the value associated with `Mode` a `String` instead of an enum?

codex-rs/core/src/sandboxing/mod.rs
```diff
@@ -260,9 +268,31 @@ fn sandbox_policy_with_additional_permissions(
...
             }
         }
+    }
+}
+
+pub(crate) fn effective_sandbox_permissions(
+    sandbox_policy: &SandboxPolicy,
+    macos_seatbelt_profile_extensions: Option<&MacOsSeatbeltProfileExtensions>,
+    additional_permissions: Option<&PermissionProfile>,
+) -> EffectiveSandboxPermissions {
```
dylan-hurd-oai: This seems like it should be a constructor for EffectiveSandboxPermissions?

bolinfest (approved):

codex-rs/core/src/sandboxing/mod.rs
```diff
@@ -326,14 +356,18 @@ impl SandboxManager {
...
             windows_sandbox_level,
         } = request;
-        let effective_policy =
-            if let Some(additional_permissions) = spec.additional_permissions.take() {
-                sandbox_policy_with_additional_permissions(policy, &additional_permissions)?
-            } else {
-                policy.clone()
-            };
+        #[cfg(not(target_os = "macos"))]
+        let macos_seatbelt_profile_extensions = None;
```
dylan-hurd-oai, line 362: Likely out of scope of this PR, but it feels like a really leaky abstraction to have to use target_os macros everywhere we declare this variable, particularly when it's part of the `effective_sandbox_permissions` constructor. I wonder if we could make this easier to work with somehow, e.g. passing around the outer struct more consistently

codex-rs/core/src/tools/spec.rs
```diff
@@ -234,7 +234,7 @@ fn create_approval_parameters(request_permission_enabled: bool) -> BTreeMap<Stri
             JsonSchema::String {
                 description: Some(
                     if request_permission_enabled {
-                        "Sandbox permissions for the command. Use \"with_additional_permissions\" to request additional sandboxed filesystem access (preferred), or \"require_escalated\" to request running without sandbox restrictions; defaults to \"use_default\"."
+                        "Sandbox permissions for the command. Use \"with_additional_permissions\" to request additional sandboxed filesystem, network, or macOS permissions (preferred), or \"require_escalated\" to request running without sandbox restrictions; defaults to \"use_default\"."
```
dylan-hurd-oai, line 256: non-blocking but we likely need an approach for restricting platform-specific permissions from the model

codex-rs/core/src/tools/spec.rs
```diff
@@ -272,36 +272,100 @@ fn create_approval_parameters(request_permission_enabled: bool) -> BTreeMap<Stri
...
+                properties: BTreeMap::from([
+                    (
+                        "network".to_string(),
+                        JsonSchema::Object {
+                            properties: BTreeMap::from([(
+                                "enabled".to_string(),
+                                JsonSchema::Boolean {
                                     description: Some(
-                                        "Additional filesystem paths to grant read access for this command."
+                                        "Set to true to enable network access for this command."
```
dylan-hurd-oai, line 302: Also non-blocking, but we'll need to continue to iterate on this, right?

dylan-hurd-oai (commented):
Overall aligned with this direction! Left some comments on code structure

codex-rs/core/src/tools/spec.rs
celia-oai: Yup!

codex-rs/core/src/sandboxing/mod.rs
celia-oai: I refactored the code to use the same enum as you defined, so there's no two separate models

codex-rs/core/src/sandboxing/macos_permissions.rs
celia-oai: I feel it makes sense to have the None handling here? lmk if you prefer just keeping the core merging logic, but I feel this is cleaner?

dylan-hurd-oai: reasonable! i'm peaceful with this approach

dylan-hurd-oai (approved):

PR #13504, opened 2026-03-04
Preserve persisted thread git info in resume

joeytrasatti-openai:
## Summary
- ensure `thread.resume` reuses the stored `gitInfo` instead of rebuilding it from the live working tree
- persist and apply thread git metadata through the resume flow and add a regression test covering branch mismatch cases

## Testing
- Not run (not requested)


owenlin0 (commented):
this code currently feels very special-cased for `gitinfo`. I think there's actually a structural problem with `thread/resume` - codex suggests this, which I think is a great idea:
> The cleaner approach is to make resume build Thread the same way `thread/read` already does: prefer DB-backed summary for mutable metadata, then layer rollout-derived turns on top, and merge the active turn for the running-thread case. That removes the git special case entirely.

Want to take a stab at this?

owenlin0:
It seems like we now have a few methods that look similar:
- load_thread_for_running_resume_response
- load_thread_for_cold_resume_response
- load_thread_summary_for_rollout

I think we can make it cleaner. Codex suggests this: 

> There are really only two concerns here:
> - Build the base Thread metadata.
> - Populate/merge turns for the resume response.
> 
> A cleaner shape would be:
> - Keep one metadata loader: `load_thread_summary_for_rollout(...) -> Thread`
> - Add one turn hydrator: `populate_resume_turns(thread: &mut Thread, rollout_path, active_turn_override)` or similar
> - Handle the Forked special case inline or behind a tiny helper like `build_thread_summary_for_forked_history(...)`

owenlin0 (approved):

PR #13433, opened 2026-03-04
feat: track plugins mcps/apps and add plugin info to user_instructions

sayan-oai:
### first half of changes, followed by #13510

Track plugin capabilities as derived summaries on `PluginLoadOutcome` for enabled plugins with at least one skill/app/mcp.

Also add `Plugins` section to `user_instructions` injected on session start. These introduce the plugins concept and list enabled plugins, but do NOT currently include paths to enabled plugins or details on what apps/mcps the plugins contain (current plan is to inject this on @-mention). that can be adjusted in a follow up and based on evals.

### tests
Added/updated tests, confirmed locally that new `Plugins` section + currently enabled plugins show up in `user_instructions`.

PR #13510, opened 2026-03-04
add @plugin mentions

sayan-oai:
## Note-- added plugin mentions via @, but that conflicts with file mentions

depends and builds upon #13433.

- introduces explicit `@plugin` mentions. this injects the plugin's mcp servers, app names, and skill name format into turn context as a dev message.
  - we do not yet have UI for these mentions, so we currently parse raw text (as opposed to skills and apps which have UI chips, autocomplete, etc.) this depends on a `plugins/list` app-server endpoint we can feed the UI with, which is upcoming
- also annotate mcp and app tool descriptions with the plugin(s) they come from. this gives the model a first class way of understanding what tools come from which plugins, which will help implicit invocation.

### Tests
Added and updated tests, unit and integration. Also confirmed locally a raw `@plugin` injects the dev message, and the model knows about its apps, mcps, and skills.


codex-rs/core/src/mcp/mod.rs
```diff
@@ -217,7 +217,12 @@ pub async fn collect_mcp_snapshot(config: &Config) -> McpListToolsResponseEvent
         config.cli_auth_credentials_store_mode,
     );
     let auth = auth_manager.auth().await;
-    let mcp_manager = McpManager::new(Arc::new(PluginsManager::new(config.codex_home.clone())));
+    let plugins_manager = Arc::new(PluginsManager::new(config.codex_home.clone()));
+    let tool_plugin_provenance = plugins_manager
+        .plugins_for_config(config)
+        .tool_plugin_provenance()
+        .clone();
+    let mcp_manager = McpManager::new(plugins_manager);
```
pakrym-oai: why isn't tool_plugin_provenance stuff not handled inside McpManager?

codex-rs/core/src/codex.rs
```diff
@@ -5722,6 +5773,55 @@ async fn built_tools(
     )))
 }
 
+fn build_plugin_injections(
```
pakrym-oai: can we move this out please?

codex-rs/core/src/mentions.rs
```diff
@@ -48,6 +49,103 @@ pub(crate) fn collect_explicit_app_ids(input: &[UserInput]) -> HashSet<String> {
...
+///
+/// It is hand-rolled because core only has a `$...` / `[$...](...)` mention
+/// parser today, and the existing TUI `@...` logic is file-autocomplete, not
+/// turn-time parsing.
+///
+/// Long term, explicit plugin picks should come through structured
+/// `plugin://...` mentions, likely via `UserInput::Mention`, once clients can list
+/// plugins and the UI has plugin-mention support (likely a plugins/list app-server
+/// endpoint). Even then, this may stay as a text fallback, similar to skills/apps.
+pub(crate) fn collect_explicit_plugin_mentions(
```
pakrym-oai, line 66: can we have one parser that handles all mentions at least for core parsing logic?

codex-rs/core/src/codex.rs
```diff
@@ -5722,6 +5773,55 @@ async fn built_tools(
...
+            )
+        })
+        .collect::<HashMap<&str, String>>();
+
+    // Turn each explicit @plugin mention into a developer hint that points the
+    // model at the plugin's visible MCP servers, enabled apps, and skill prefix.
+    mentioned_plugins
+        .iter()
+        .filter_map(|plugin| {
+            let available_mcp_servers = plugin
```
pakrym-oai: can we stick plugin info onto ToolInfo and simplify this logic. same with apps?

codex-rs/core/src/codex.rs
```diff
@@ -5000,12 +5028,29 @@ pub(crate) async fn run_turn(
...
+
     let mut explicitly_enabled_connectors = collect_explicit_app_ids(&input);
     explicitly_enabled_connectors.extend(collect_explicit_app_ids_from_skill_items(
         &skill_items,
         &available_connectors,
         &skill_name_counts_lower,
     ));
+    // Explicit @plugin mentions can make a plugin's enabled apps callable for
+    // this turn without persisting those connectors as sticky user selections.
+    let mut turn_enabled_connectors = explicitly_enabled_connectors.clone();
```
xl-openai, line 5036: One limitation is that when the user continues the conversation, the new turn won’t have this enabled.

Maybe we should just enable it for the entire conversation/thread instead.

codex-rs/core/src/codex.rs
xl-openai: For skills, we already have inject.rs. But if we want a dedicated mention + injector, we could combine them into a new file.

codex-rs/core/src/mentions.rs
sayan-oai: this parsing logic is temporary while Xin's `plugin/list` [pr](https://github.com/openai/codex/pull/13540) hasn't landed. once that lands, I think we can generalize the existing `$app` mention parsing logic to `@plugin`.

This implementation is just a holdover so we can test with things like `@figma` and `@google drive` in the meantime.

given that, is it ok to leave as-is with a todo for deletion?

codex-rs/core/src/codex.rs
sayan-oai: we could, but i think we'd run into problems because both ToolInfo and AppInfo are cached.

right now plugins are structured to support per-cwd plugins in the future, rather than just user-scoped plugins. ToolInfo and AppInfo are cached by user, so that could cause issues.

even if we move to user-scoped plugins only, we'd also have to invalidate cache on plugin enablement/disablement, or recompute at runtime anyways.

codex-rs/core/src/codex.rs
sayan-oai: then all the tools for all the apps in that plugin would be injected into Prompt.tools on every subsequent turn, which would likely flood the context window as @mzeng-openai pointed out.

codex-rs/core/src/mentions.rs
pakrym-oai: Works!

codex-rs/core/src/mcp_connection_manager.rs
```diff
@@ -1135,6 +1158,56 @@ pub(crate) fn filter_mcp_tools_by_name(
         .collect()
 }
 
+fn annotate_tools_with_plugin_sources(
```
pakrym-oai: can this be done as part of building ToolInfos so we don't have to remember to call annotate_tools_with_plugin_sources ?

xl-openai (approved):

pakrym-oai (approved):

PR #13540, opened 2026-03-05
support plugin/list.

xl-openai:
Introduce a plugin/list which reads from local marketplace.json.
Also update the signature for plugin/install.


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -4994,16 +5079,14 @@ impl CodexMessageProcessor {
...
             plugin_name,
-            cwd,
         } = params;
 
         let plugins_manager = self.thread_manager.plugins_manager();
         let request = PluginInstallRequest {
             plugin_name,
-            marketplace_name,
-            cwd: cwd.unwrap_or_else(|| self.config.cwd.clone()),
+            marketplace_path,
```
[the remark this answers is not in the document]
xl-openai: I think this is OK.

codex-rs/core/src/plugins/manager.rs
```diff
@@ -267,6 +216,40 @@ impl PluginsManager {
...
+                    path: marketplace.path,
+                    plugins: marketplace
+                        .plugins
+                        .into_iter()
+                        .map(|plugin| ConfiguredMarketplacePluginSummary {
+                            // Enabled state is keyed by `<plugin>@<marketplace>`, so duplicate
+                            // marketplace files with the same declared name intentionally share
+                            // the same config toggle.
+                            enabled: configured_plugins
+                                .get(&format!("{}@{marketplace_name}", plugin.name))
```
sayan-oai: like you say here, our downstream logic keys by plugin@marketplace, yet we return duplicate plugin@marketplace entries from `plugin/list`. Why? if we are effectively treating them as a single resolved plugin, shouldn't we enforce at most one resolved plugin per plugin@markeplace, and return that from plugin/list?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2531,6 +2548,34 @@ pub struct SkillsListEntry {
...
+    pub plugins: Vec<PluginSummary>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct PluginSummary {
+    pub name: String,
+    pub source: PluginSource,
+    pub enabled: bool,
```
sayan-oai, line 2569: like you mentioned, we should probably include installed status in an enum here (something like: uninstalled, disabled, enabled).

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -4924,6 +4895,80 @@ impl CodexMessageProcessor {
...
+                            request_id,
+                            format!("failed to list marketplace plugins: {err}"),
+                        )
+                        .await;
+                    }
+                    MarketplaceError::InvalidMarketplaceFile { .. }
+                    | MarketplaceError::PluginNotFound { .. }
+                    | MarketplaceError::DuplicatePlugin { .. }
+                    | MarketplaceError::InvalidPlugin(_) => {
+                        self.send_invalid_request_error(request_id, err.to_string())
```
sayan-oai: i think in plugin/install we treat an invalid/missing marketplace file as an internal error and not an invalid request. we should unify.

codex-rs/app-server-protocol/src/protocol/v2.rs
xl-openai: Will do it as a follow up PR.

codex-rs/core/src/plugins/manager.rs
xl-openai: Update to keep the first one.

codex-rs/app-server/src/codex_message_processor.rs
xl-openai: Fixed.

codex-rs/core/src/plugins/manager.rs
```diff
@@ -267,6 +279,56 @@ impl PluginsManager {
...
+        Ok(marketplaces
+            .into_iter()
+            .filter_map(|marketplace| {
+                let marketplace_name = marketplace.name.clone();
+                let plugins = marketplace
+                    .plugins
+                    .into_iter()
+                    .filter_map(|plugin| {
+                        let plugin_key = format!("{}@{marketplace_name}", plugin.name);
+                        if !seen_plugin_keys.insert(plugin_key.clone()) {
```
sayan-oai, line 301: we silently drop here based on first seen, but in plugin/install we error on duplicates. so i think if there are duplicate plugins in a single marketplace file, plugin/list will return one, but plugin/install will fail on the same marketplace file.

is there a reason to keep the behavior different or should we unify on one (silent drop or error)?

sayan-oai (approved):
one piece of feedback that should be addressed, but approving to unblock

codex-rs/core/src/plugins/manager.rs
xl-openai: This is a pretty annoying edge case, but I agree with you — I’ll handle both scenarios and just pick one.
I just don’t want to block plugin listing because of a minor issue on a marketplace.json file.

sayan-oai: yea good to merge without this, but we should track + address soon.

PR #13579, opened 2026-03-05
[MCP] Add support for cimd oauth authorization

gpeal:
Authored by @ethriel3695 but he wasn't able to open it due to our current external contributor policy.

This PR adds support for CIMD as a fallback MCP oauth method if dynamic client registration fails. Validated against the storybook MCP.

### DIFF
--- a/codex-rs/rmcp-client/src/perform_oauth_login.rs
+++ b/codex-rs/rmcp-client/src/perform_oauth_login.rs
@@ -364,27 +364,32 @@ fn callback_path_from_redirect_uri(redirect_uri: &str) -> Result<String> {
     Ok(parsed.path().to_string())
 }
 
+fn uses_default_cimd_redirect_uri(redirect_uri: &str) -> bool {
+    matches!(
+        redirect_uri,
+        DEFAULT_CIMD_REDIRECT_URI_ROOT | DEFAULT_CIMD_REDIRECT_URI_CALLBACK
+    )
+}
+
 fn validate_callback_listener_settings(
     callback_port: Option<u16>,
     callback_url: Option<&str>,
 ) -> Result<()> {
-    let Some(callback_port) = callback_port else {
-        return Ok(());
-    };
     let Some(callback_url) = callback_url else {
         return Ok(());
     };
-    let parsed = Url::parse(callback_url)
-        .with_context(|| format!("invalid MCP OAuth callback URL `{callback_url}`"))?;
-    let Some(callback_url_port) = parsed.port() else {
+    if !uses_default_cimd_redirect_uri(callback_url) {
+        return Ok(());
+    };
+    let Some(callback_port) = callback_port else {
         return Ok(());
     };
 
-    if callback_url_port == callback_port {
+    if callback_port == DEFAULT_CIMD_CALLBACK_PORT {
         Ok(())
     } else {
         Err(anyhow!(
-            "MCP OAuth callback URL `{callback_url}` uses port `{callback_url_port}` but `mcp_oauth_callback_port` is set to `{callback_port}`"
+            "MCP OAuth callback URL `{callback_url}` is a built-in Codex client metadata redirect URI, so `mcp_oauth_callback_port` must be `{DEFAULT_CIMD_CALLBACK_PORT}` or unset"
         ))
     }
 }
@@ -473,12 +478,16 @@ impl OauthLoginFlow {
         let supports_default_cimd_metadata = client_id_metadata_document_supported(&metadata);
         let callback_port_is_explicitly_set = callback_port.is_some();
         let callback_url_is_explicitly_set = callback_url.is_some();
+        let callback_url_uses_default_cimd_redirect = callback_url
+            .map(uses_default_cimd_redirect_uri)
+            .unwrap_or(false);
         let should_rebind_callback_for_cimd_fallback =
             !callback_port_is_explicitly_set && !callback_url_is_explicitly_set;
+        let should_bind_to_default_cimd_port = should_start_with_default_cimd_metadata
+            || (!callback_port_is_explicitly_set && callback_url_uses_default_cimd_redirect);
 
         let bind_host = callback_bind_host(callback_url);
-        let callback_port =
-            resolve_callback_port(callback_port, should_start_with_default_cimd_metadata)?;
+        let callback_port = resolve_callback_port(callback_port, should_bind_to_default_cimd_port)?;
         let mut callback_listener =
             start_callback_listener(&bind_host, callback_port, callback_url)?;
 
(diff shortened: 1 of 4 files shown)


ethriel3695:
@codex review

codex-rs/rmcp-client/src/perform_oauth_login.rs
```diff
@@ -24,6 +26,13 @@ use crate::save_oauth_tokens;
 use crate::utils::apply_default_headers;
 use crate::utils::build_default_headers;
 
+const DEFAULT_CIMD_CLIENT_METADATA_URL: &str =
+    "https://raw.githubusercontent.com/openai/codex/main/codex-rs/client-metadata.json";
+const CLIENT_ID_METADATA_DOCUMENT_SUPPORTED_FIELD: &str = "client_id_metadata_document_supported";
+const DEFAULT_CIMD_CALLBACK_PORT: u16 = 33418;
```
gpeal, line 39: How was this port chosen? Ideally, it's auto-selected unless the user [overrides it in their config](https://github.com/openai/codex/blob/main/codex-rs/core/src/config/mod.rs#L337)

ethriel3695: @gpeal 
I based it on the format that VSCode has for their `client-metadata.json`
https://vscode.dev/oauth/client-metadata.json

For CIMD specifically, there’s a hard constraint:

If `client_id` is a metadata URL, the server can validate `redirect_uri` against the `redirect_uris` listed in that JSON.
So true: "any ephemeral port" isn't possible with a static metadata file unless the server supports wildcard/relaxed loopback matching, which I don't believe we do on the Chromatic side.

We can implement auto-selection for a limited number of ports without exposing additional configuration to users.
Something similar to the following:

* Keep the user override
* Ship a bounded port pool in client-metadata.json (example: 33418..33427, both for  `/` and `/callback`)
* Auto-pick the first available port from that pool when the user did not override
* If the user override is outside the pool, return a clear error

What are your thoughts here?

codex-rs/rmcp-client/src/perform_oauth_login.rs
```diff
@@ -239,17 +248,12 @@ struct OauthLoginFlow {
...
-fn resolve_callback_port(callback_port: Option<u16>) -> Result<Option<u16>> {
-    if let Some(config_port) = callback_port {
-        if config_port == 0 {
-            bail!(
-                "invalid MCP OAuth callback port `{config_port}`: port must be between 1 and 65535"
-            );
-        }
-        return Ok(Some(config_port));
+fn resolve_callback_port(callback_port: Option<u16>) -> Result<u16> {
+    let port = callback_port.unwrap_or(DEFAULT_CIMD_CALLBACK_PORT);
```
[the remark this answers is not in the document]
gpeal: This feedback is valid

gpeal (commented):
I would run `/review` in the Codex app through this. It caught the same thing flagged by the review bot plus this one

```
callback_bind_host() coerces a configured callback URL on ::1 to 127.0.0.1, but resolve_redirect_uri() still sends the browser to the original IPv6 redirect URI. With mcp_oauth_callback_url = "http://[::1]:...", Codex listens only on IPv4 while the OAuth provider redirects to IPv6, so the callback never reaches the local server and the login times out. The listener needs to preserve the IPv6 loopback address instead of collapsing it to IPv4.
```

codex-rs/rmcp-client/src/perform_oauth_login.rs
ethriel3695: I ran `/review` in Codex and made the necessary changes .

ethriel3695:
@codex review

ethriel3695:
@codex review

codex-rs/rmcp-client/src/perform_oauth_login.rs
```diff
@@ -24,6 +27,13 @@ use crate::save_oauth_tokens;
 use crate::utils::apply_default_headers;
 use crate::utils::build_default_headers;
 
+const DEFAULT_CIMD_CLIENT_METADATA_URL: &str =
```
bolinfest, line 35: Should this point to `main` instead of a specific commit hash?

I guess if we always want it to grab the "latest," then `main` makes sense, but we should add a docstring to make it clear this is a conscious choice.

codex-rs/rmcp-client/src/perform_oauth_login.rs
bolinfest: This should be memorialized in a docstring on `DEFAULT_CIMD_CALLBACK_PORT` if we stick with this approach.

codex-rs/rmcp-client/src/perform_oauth_login.rs
```diff
@@ -239,17 +249,28 @@ struct OauthLoginFlow {
...
-                "invalid MCP OAuth callback port `{config_port}`: port must be between 1 and 65535"
-            );
+struct CallbackListener {
+    guard: CallbackServerGuard,
+    redirect_uri: String,
+    rx: oneshot::Receiver<(String, String)>,
+}
+
+fn resolve_callback_port(
+    callback_port: Option<u16>,
```
bolinfest, line 318: Consider: https://doc.rust-lang.org/nightly/core/num/type.NonZeroU16.html

codex-rs/rmcp-client/src/perform_oauth_login.rs
```diff
@@ -440,6 +557,84 @@ impl OauthLoginFlow {
...
+fn validate_redirect_uri_for_default_cimd_metadata(redirect_uri: &str) -> Result<()> {
+    if matches!(
+        redirect_uri,
+        DEFAULT_CIMD_REDIRECT_URI_ROOT | DEFAULT_CIMD_REDIRECT_URI_CALLBACK
+    ) {
+        return Ok(());
+    }
+
+    bail!(
+        "MCP OAuth callback URL `{redirect_uri}` is incompatible with built-in Codex client metadata; use `{DEFAULT_CIMD_REDIRECT_URI_ROOT}` or `{DEFAULT_CIMD_REDIRECT_URI_CALLBACK}`"
```
bolinfest: Slightly more canonical:

```suggestion
    ) {
        Ok(())
    } else {
        Err(anyhow!(
        "MCP OAuth callback URL `{redirect_uri}` is incompatible with built-in Codex client metadata; use `{DEFAULT_CIMD_REDIRECT_URI_ROOT}` or `{DEFAULT_CIMD_REDIRECT_URI_CALLBACK}`")
    }
```

codex-rs/rmcp-client/src/perform_oauth_login.rs
```diff
@@ -459,12 +654,49 @@ fn append_query_param(url: &str, key: &str, value: Option<&str>) -> String {
 
 #[cfg(test)]
 mod tests {
```
bolinfest: FYI, I haven't codemodded the entire repo yet, but since this is relatively new code, I would follow the pattern in https://github.com/openai/codex/pull/13780 and introduce a `_tests.rs` file and do:

```
#[cfg(test)]
#[path = "perform_oauth_login.rs_tests.rs"]
mod tests;
```

bolinfest (approved):
Small comments on business logic: I'm trusting that between @gpeal and @ethriel3695 you feel good about the test coverage?

codex-rs/rmcp-client/src/perform_oauth_login.rs
ethriel3695: @bolinfest I added comments to address this note.

codex-rs/rmcp-client/src/perform_oauth_login.rs
ethriel3695: @bolinfest I added comments to address this note and I made the change to utilize a `NonZeroU16` type.

codex-rs/rmcp-client/src/perform_oauth_login.rs
ethriel3695: @bolinfest I made the change to factor out the tests into a `_tests.rs` per the new pattern.

ethriel3695: @bolinfest here is a snapshot as well of it working.
Screenshot 2026-03-09 at 11 47 54

Screenshot 2026-03-09 at 11 54 55

pakrym-oai (approved):

PR #13592, opened 2026-03-05
Prefer websockets when providers support them

pakrym-oai:
Remove all flags and model settings.


sayan-oai (changes requested):
Blocking issue:

The new test `sdk/typescript/tests/run.test.ts` now hard-fails when `CODEX_HOME` is unset:

```ts
const codexHome = process.env.CODEX_HOME;
if (!codexHome) {
  throw new Error("CODEX_HOME must be set when running SDK tests");
}
```

In CI (`sdk` workflow), `CODEX_HOME` is not set, so this test fails deterministically. The previous version of this test did not require `CODEX_HOME`.

Please make the test self-contained (for example, create a temp home directory in-test and pass it through `env`) instead of requiring global environment setup.

codex-rs/core/tests/suite/client_websockets.rs
```diff
@@ -1459,22 +1480,21 @@ async fn websocket_harness_with_runtime_metrics(
...
 ) -> WebsocketTestHarness {
-    websocket_harness_with_options(server, false, true, websocket_v2_enabled, false).await
+    websocket_harness_with_options(server, false, true, websocket_v2_enabled).await
 }
 
 async fn websocket_harness_with_options(
     server: &WebSocketTestServer,
     runtime_metrics_enabled: bool,
     websocket_enabled: bool,
     websocket_v2_enabled: bool,
```
sayan-oai: should these be rm'd?

sayan-oai (approved):
small comment about dead args in test helper

PR #13607, opened 2026-03-05
Enabling CWD Saving for Image-Gen

won-openai:
Codex now saves the generated image on to your current working directory.


codex-rs/core/src/stream_events_utils.rs
```diff
@@ -157,7 +181,17 @@ pub(crate) async fn handle_output_item_done(
         }
         // No tool call: convert messages/reasoning into turn items and mark them as complete.
         Ok(None) => {
-            if let Some(turn_item) = handle_non_tool_response_item(&item, plan_mode) {
+            if let Some(mut turn_item) = handle_non_tool_response_item(&item, plan_mode) {
```
aibrahim-oai: we shouldn't change the item and put the link in the result. You can add a path item on it and do that work when constructing the item.

codex-rs/core/src/stream_events_utils.rs
```diff
@@ -158,12 +190,38 @@ pub(crate) async fn handle_output_item_done(
         // No tool call: convert messages/reasoning into turn items and mark them as complete.
         Ok(None) => {
             if let Some(turn_item) = handle_non_tool_response_item(&item, plan_mode) {
+                let turn_item = match turn_item {
+                    TurnItem::ImageGeneration(mut image_item) => {
+                        match save_image_generation_result_to_cwd(
```
aibrahim-oai: let's do this logic in `handle_non_tool_response_item`

aibrahim-oai (approved):

PR #13611, opened 2026-03-05
[telemetry] item level metadata

rhan-oai:
changes
- distinguish prompt queuing from prompt steering in core
- introduce metadata construct for user messages items
- includes is_plan_mode, is_tool_call_escalated, escalation_status, sandbox_policy, user_message_type, personality

considerations
- backwards compatible (metadata optional)
- ide_context and agents detection still somewhat brittle


codex-rs/protocol/src/items.rs
```diff
@@ -29,10 +29,73 @@ pub enum TurnItem {
...
+
+#[derive(Debug, Clone, Deserialize, Serialize, TS, JsonSchema, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+#[ts(rename_all = "snake_case")]
+pub enum UserMessagePersonality {
+    Friendly,
+    Pragmatic,
+}
+
+#[derive(Debug, Clone, Default, Deserialize, Serialize, TS, JsonSchema, PartialEq, Eq)]
```
dylan-hurd-oai, line 70: We shouldn't be generating typescript for these, they're internal

codex-rs/protocol/src/items.rs
```diff
@@ -29,10 +29,73 @@ pub enum TurnItem {
...
+    pub personality: Option<UserMessagePersonality>,
+}
+
 #[derive(Debug, Clone, Deserialize, Serialize, TS, JsonSchema)]
 pub struct UserMessageItem {
     pub id: String,
     pub content: Vec<UserInput>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub metadata: Option<UserMessageMetadata>,
```
dylan-hurd-oai, line 98: I don't think this is the right place to add this - tool calls and escalation outcomes aren't tied to a user message, right?

PR #13620, opened 2026-03-05
chore: improve DB flushing

jif-oai:
This branch:
* Avoid flushing DB when not necessary
* Filter events for which we perfom an `upsert` into the DB
* Add a dedicated update function of the `thread:updated_at` that is lighter

This should significantly reduce the DB lock contention. If it is not sufficient, we can de-sync the flush of the DB for `updated_at`


jif-oai:
@codex review

charley-oai (approved):

PR #13630, opened 2026-03-05
feat(otel, core): record turn TTFT and TTFM metrics in codex-core

owenlin0:
### Summary
This adds turn-level latency metrics for the first model output and the first completed agent message.
- `codex.turn.ttft.duration_ms` starts at turn start and records on the first output signal we see from the model. That includes normal assistant text, reasoning deltas, and non-text outputs like tool-call items.
- `codex.turn.ttfm.duration_ms` also starts at turn start, but it records when the first agent message finishes streaming rather than when its first delta arrives.

### Implementation notes
The timing is tracked in codex-core, not app-server, so the definition stays consistent across CLI, TUI, and app-server clients.

I reused the existing turn lifecycle boundary that already drives `codex.turn.e2e_duration_ms`, stored the turn start timestamp in turn state, and record each metric once per turn. 

I also wired the new metric names into the OTEL runtime metrics summary so they show up in the same in-memory/debug snapshot path as the existing timing metrics.

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -265,6 +265,7 @@ use crate::skills::resolve_skill_dependencies_for_turn;
 use crate::state::ActiveTurn;
 use crate::state::SessionServices;
 use crate::state::SessionState;
+use crate::state::TurnTimingState;
 use crate::state_db;
 use crate::tasks::GhostSnapshotTask;
 use crate::tasks::RegularTask;
@@ -331,6 +332,8 @@ pub(crate) const INITIAL_SUBMIT_ID: &str = "";
 pub(crate) const SUBMISSION_CHANNEL_CAPACITY: usize = 512;
 const CYBER_VERIFY_URL: &str = "https://chatgpt.com/cyber";
 const CYBER_SAFETY_URL: &str = "https://developers.openai.com/codex/concepts/cyber-safety";
+const TURN_TTFT_DURATION_METRIC: &str = "codex.turn.ttft.duration_ms";
+const TURN_TTFM_DURATION_METRIC: &str = "codex.turn.ttfm.duration_ms";
 
 impl Codex {
     /// Spawn a new [`Codex`] and initialize the session.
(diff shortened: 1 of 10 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -691,6 +692,7 @@ pub(crate) struct TurnContext {
     pub(crate) dynamic_tools: Vec<DynamicToolSpec>,
     pub(crate) turn_metadata_state: Arc<TurnMetadataState>,
     pub(crate) turn_skills: TurnSkillsContext,
+    pub(crate) turn_timing_state: Arc<Mutex<TurnTimingState>>,
```
pakrym-oai: let's put mutex inside `turn_timing_state`. This way consumers don't need to lock all the time.

codex-rs/core/src/codex.rs
```diff
@@ -6374,6 +6434,9 @@ async fn try_run_sampling_request(
         match event {
             ResponseEvent::Created => {}
             ResponseEvent::OutputItemDone(item) => {
+                if response_item_records_turn_ttft(&item) {
```
pakrym-oai: what if we encapsulate this logic in the timing thingy? pass an item to it and have it decide what metrics to aggregate/send.

codex-rs/core/src/state/turn.rs
```diff
@@ -51,6 +53,41 @@ pub(crate) struct RunningTask {
     pub(crate) _timer: Option<codex_otel::Timer>,
 }
 
+#[derive(Debug, Default)]
+pub(crate) struct TurnTimingState {
```
pakrym-oai: consider a separate file

codex-rs/core/src/codex.rs
```diff
@@ -5920,6 +5946,40 @@ fn agent_message_text(item: &codex_protocol::items::AgentMessageItem) -> String
         .collect()
 }
 
+fn response_item_records_turn_ttft(item: &ResponseItem) -> bool {
```
pakrym-oai: this should ideally go somewhere next to TurnTimingState

pakrym-oai (approved):
Can we offload more logic from codex.rs?

PR #13637, opened 2026-03-05
feat(rollout): preserve fork references across replay

friel-openai:
## Summary
- add fork reference rollout items so forked threads can reuse parent history instead of duplicating it
- preserve fork references during replay, resume, and thread reads
- fix nested fork replay and truncation behavior so archived or replayed parents continue to resolve correctly

## Notes
- this keeps fork-heavy sessions from growing rollout files superlinearly
- replay and summary paths now materialize fork references when needed instead of assuming fully inlined history


codex-rs/core/src/agent/control.rs
```diff
@@ -489,3 +505,1213 @@ impl AgentControl {
 #[cfg(test)]
 #[path = "control_tests.rs"]
 mod tests;
+#[cfg(test)]
+mod fork_reference_tests {
```
jif-oai: please check what is done with the other tests and make sure to do the same

codex-rs/core/src/agent/control.rs
```diff
@@ -169,6 +170,21 @@ impl AgentControl {
...
+                        forked_rollout_items =
+                            crate::rollout::truncation::materialize_rollout_items_for_replay(
+                                config.codex_home.as_path(),
+                                &forked_rollout_items,
+                            )
+                            .await;
+                    }
+                    forked_rollout_items.push(RolloutItem::ForkReference(ForkReferenceItem {
+                        rollout_path: rollout_path.clone(),
+                        nth_user_message: usize::MAX,
```
jif-oai: Using \usize::MAX` here makes the child point at the parent's live tail, so any parent turns added after spawn get pulled into the child on resume... this is not the correct behaviour

codex-rs/protocol/src/protocol.rs
```diff
@@ -2194,7 +2202,11 @@ impl InitialHistory {
                     .iter()
                     .filter_map(|ri| match ri {
                         RolloutItem::EventMsg(ev) => Some(ev.clone()),
-                        _ => None,
+                        RolloutItem::SessionMeta(_)
```
jif-oai: why? (same for every edit in this file)

codex-rs/core/src/thread_rollout_truncation.rs
```diff
@@ -68,6 +72,77 @@ pub(crate) fn truncate_rollout_before_nth_user_message_from_start(
     items[..cut_idx].to_vec()
 }
 
+pub(crate) async fn materialize_rollout_items_for_replay(
```
jif-oai: Can you document this public function please?

codex-rs/core/src/thread_rollout_truncation.rs
```diff
@@ -68,6 +72,77 @@ pub(crate) fn truncate_rollout_before_nth_user_message_from_start(
...
+                    }
+
+                    let resolved_rollout_path = match resolve_fork_reference_rollout_path(
+                        codex_home,
+                        &reference.rollout_path,
+                    )
+                    .await
+                    {
+                        Ok(path) => path,
+                        Err(err) => {
```
jif-oai, line 204: this will corrupt the replay

codex-rs/protocol/src/protocol.rs
```diff
@@ -2401,10 +2437,17 @@ pub struct SessionMetaLine {
     pub git: Option<GitInfo>,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema, TS)]
+pub struct ForkReferenceItem {
+    pub rollout_path: PathBuf,
+    pub nth_user_message: usize,
```
jif-oai: Let's not use `usize` here. Mainly if we sometimes use `usize::MAX`. This means the encoding become platform specific and we head toward problems with this. Just use `i32` or `i64` (no unsigned please)

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -8048,6 +8053,137 @@ fn preview_from_rollout_items(items: &[RolloutItem]) -> String {
...
+        let name = ancestor.file_name().and_then(OsStr::to_str)?;
+        if name == codex_core::SESSIONS_SUBDIR || name == codex_core::ARCHIVED_SESSIONS_SUBDIR {
+            ancestor.parent()
+        } else {
+            None
+        }
+    })
+}
+
+async fn materialize_rollout_items_for_replay(
```
jif-oai: Don't re-implement... re-use what is done in core/

jif-oai (changes requested):

PR #13644, opened 2026-03-06
fix: preserve zsh-fork escalation fds across unified-exec spawn paths

bolinfest:
## Why

`zsh-fork` sessions launched through unified-exec need the escalation socket to survive the wrapper -> server -> child handoff so later intercepted `exec()` calls can still reach the escalation server.

The inherited-fd spawn path also needs to avoid closing Rust's internal exec-error pipe, and the shell-escalation handoff needs to tolerate the receive-side case where a transferred fd is installed into the same stdio slot it will be mapped onto.

## What Changed

- Added `SpawnLifecycle::inherited_fds()` in `codex-rs/core/src/unified_exec/process.rs` and threaded inherited fds through `codex-rs/core/src/unified_exec/process_manager.rs` so unified-exec can preserve required descriptors across both PTY and no-stdin pipe spawn paths.
- Updated `codex-rs/core/src/tools/runtimes/shell/zsh_fork_backend.rs` to expose the escalation socket fd through the spawn lifecycle.
- Added inherited-fd-aware spawn helpers in `codex-rs/utils/pty/src/pty.rs` and `codex-rs/utils/pty/src/pipe.rs`, including Unix pre-exec fd pruning that preserves requested inherited fds while leaving `FD_CLOEXEC` descriptors alone. The pruning helper is now named `close_inherited_fds_except()` to better describe that behavior.
- Updated `codex-rs/shell-escalation/src/unix/escalate_client.rs` to duplicate local stdio before transfer and send destination stdio numbers in `SuperExecMessage`, so the wrapper keeps using its own `stdin`/`stdout`/`stderr` until the escalated child takes over.
- Updated `codex-rs/shell-escalation/src/unix/escalate_server.rs` so the server accepts the overlap case where a received fd reuses the same stdio descriptor number that the child setup will target with `dup2`.
- Added comments around the PTY stdio wiring and the overlap regression helper to make the fd handoff and controlling-terminal setup easier to follow.

## Verification

- `cargo test -p codex-utils-pty`
  - covers preserved-fd PTY spawn behavior, PTY resize, Python REPL continuity, exec-failure reporting, and the no-stdin pipe path
  - covers duplicated-fd transfer on the client side and verifies the overlap case by passing a pipe-backed stdin payload through the server-side `dup2` path

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/13644).
* #14624
* __->__ #13644

### DIFF
--- a/codex-rs/core/src/unified_exec/process_manager.rs
+++ b/codex-rs/core/src/unified_exec/process_manager.rs
@@ -537,24 +537,27 @@ impl UnifiedExecProcessManager {
             .command
             .split_first()
             .ok_or(UnifiedExecError::MissingCommandLine)?;
+        let inherited_fds = spawn_lifecycle.inherited_fds();
 
         let spawn_result = if tty {
-            codex_utils_pty::pty::spawn_process(
+            codex_utils_pty::pty::spawn_process_with_inherited_fds(
                 program,
                 args,
                 env.cwd.as_path(),
                 &env.env,
                 &env.arg0,
                 codex_utils_pty::TerminalSize::default(),
+                &inherited_fds,
             )
             .await
         } else {
-            codex_utils_pty::pipe::spawn_process_no_stdin(
+            codex_utils_pty::pipe::spawn_process_no_stdin_with_inherited_fds(
                 program,
                 args,
                 env.cwd.as_path(),
                 &env.env,
                 &env.arg0,
+                &inherited_fds,
             )
             .await
         };
--- a/codex-rs/utils/pty/src/process.rs
+++ b/codex-rs/utils/pty/src/process.rs
@@ -1,5 +1,7 @@
 use core::fmt;
 use std::io;
+#[cfg(unix)]
+use std::os::fd::RawFd;
 use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 use std::sync::Mutex as StdMutex;
@@ -41,9 +43,24 @@ impl From<TerminalSize> for PtySize {
     }
 }
 
+#[cfg(unix)]
+pub(crate) trait PtyHandleKeepAlive: Send {}
+
+#[cfg(unix)]
+impl<T: Send + ?Sized> PtyHandleKeepAlive for T {}
+
+pub(crate) enum PtyMasterHandle {
+    Resizable(Box<dyn MasterPty + Send>),
+    #[cfg(unix)]
+    Opaque {
+        raw_fd: RawFd,
+        _handle: Box<dyn PtyHandleKeepAlive>,
+    },
+}
+
 pub struct PtyHandles {
     pub _slave: Option<Box<dyn SlavePty + Send>>,
-    pub _master: Box<dyn MasterPty + Send>,
+    pub(crate) _master: PtyMasterHandle,
 }
 
 impl fmt::Debug for PtyHandles {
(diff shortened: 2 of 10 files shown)


codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -535,14 +535,16 @@ impl UnifiedExecProcessManager {
             .command
             .split_first()
             .ok_or(UnifiedExecError::MissingCommandLine)?;
+        let inherited_fds = spawn_lifecycle.inherited_fds();
 
         let spawn_result = if tty {
-            codex_utils_pty::pty::spawn_process(
+            codex_utils_pty::pty::spawn_process_with_inherited_fds(
```
jif-oai, line 543: is it normal we don't consider the file descriptors for the non-pty version? 
If yes, could you add a small comment explaining why. I'm not sure I follow this part

codex-rs/utils/pty/src/pty.rs
```diff
@@ -198,3 +253,221 @@ pub async fn spawn_process(
...
+async fn spawn_process_preserving_fds(
+    program: &str,
+    args: &[String],
+    cwd: &Path,
+    env: &HashMap<String, String>,
+    arg0: &Option<String>,
+    inherited_fds: &[RawFd],
+) -> Result<SpawnedProcess> {
+    let (master, slave) = open_unix_pty()?;
+    let mut command = StdCommand::new(arg0.as_ref().unwrap_or(&program.to_string()));
```
jif-oai: What if you have 
```
let program = "/bin/echo";
let args = ["hello"]
let arg0 = Some("my-echo".to_string())
```
I think your code will try to execute `my-echo` but in practice we would like to execute `/bin/echo` with the my-echo as arg0

I think this should never happen in practice but it looks a bit odd to me

codex-rs/utils/pty/src/process.rs
```diff
@@ -17,8 +16,8 @@ pub(crate) trait ChildTerminator: Send + Sync {
 }
 
 pub struct PtyHandles {
-    pub _slave: Option<Box<dyn SlavePty + Send>>,
-    pub _master: Box<dyn MasterPty + Send>,
+    pub _slave: Option<Box<dyn Any + Send>>,
```
jif-oai: ooc why putting `any`? Seems harder to reason about IMO

jif-oai (approved):
Will require some heavy testing in the end but I guess it's in the follow up

Approving to unblock but please make sure the point raised (mainly for non-pty) are not real issues

codex-rs/utils/pty/src/pty.rs
```diff
@@ -198,3 +250,214 @@ pub async fn spawn_process(
...
+            let num = entry
+                .ok()
+                .map(|entry| entry.file_name())
+                .and_then(|name| name.into_string().ok())
+                .and_then(|name| name.parse::<RawFd>().ok());
+            if let Some(num) = num {
+                if num <= 2 || preserved_fds.contains(&num) {
+                    continue;
+                }
+                fds.push(num);
```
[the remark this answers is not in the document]
bolinfest: Good catch. I addressed this by teaching fd pruning to skip descriptors that already have `FD_CLOEXEC` set. That preserves Rust’s internal exec-error pipe through `pre_exec`, so `spawn()` still reports exec failures correctly. See `close_random_fds_except()` in `utils/pty/src/pty.rs`.

codex-rs/utils/pty/src/pty.rs
bolinfest: Yep, agreed. I changed this to always execute `program` and apply `arg0` via `command.arg0(...)` when present, so we no longer accidentally treat `arg0` as the executable path.

PR #13645, opened 2026-03-06
Add timestamped SQLite /feedback logs without schema changes

charley-oai:
## Summary
- keep the SQLite schema unchanged (no migrations)
- add timestamps to SQLite-backed `/feedback` log exports
- keep the existing SQL-side byte cap behavior and newline handling
- document the remaining fidelity gap (span prefixes + structured fields) with TODOs

## Details
- update `query_feedback_logs` to format each exported line as:
  - `YYYY-MM-DDTHH:MM:SS.ffffffZ {level} {message}`
- continue scoping rows to requested-thread + same-process threadless logs
- continue capping in SQL before returning rows
- keep the existing fallback behavior unchanged when SQLite returns no rows
- update parity tests to normalize away the new timestamp prefix while we still only store `message`

## Follow-up
- TODO already in code: persist enough span/event metadata in SQLite to reproduce span prefixes and structured fields in `/feedback` exports

## Testing
- `cargo test -p codex-state`
- `just fmt`

### DIFF
--- a/codex-rs/state/src/log_db.rs
+++ b/codex-rs/state/src/log_db.rs
@@ -420,19 +420,16 @@ mod tests {
 
         drop(guard);
 
-        // TODO(ccunningham): Store enough span/event metadata in SQLite to reproduce
-        // span prefixes and structured fields in feedback exports.
+        // SQLite exports now include timestamps, while this test writer has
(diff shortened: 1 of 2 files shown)


charley-oai:
@codex review this

pakrym-oai (approved):

PR #13669, opened 2026-03-06
copy current exe to CODEX_HOME/.sandbox-bin for apply_patch

iceweasel-oai:
We do this for codex-command-runner.exe as well for the same reason.
Windows sandbox users cannot execute binaries in the WindowsApp/ installed directory for the Codex App. This causes apply-patch to fail because it tries to execute codex.exe as the sandbox user.

### DIFF
--- a/codex-rs/windows-sandbox-rs/src/helper_materialization.rs
+++ b/codex-rs/windows-sandbox-rs/src/helper_materialization.rs
@@ -88,6 +88,34 @@ pub(crate) fn resolve_helper_for_launch(
     }
 }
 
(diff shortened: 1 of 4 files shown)


codex-rs/windows-sandbox-rs/src/helper_materialization.rs
```diff
@@ -349,3 +377,4 @@ mod tests {
         );
     }
 }
+
```
dylan-hurd-oai, line 380: 👀

dylan-hurd-oai (approved):

Issue #13660, opened 2026-03-06
Codex CLI: Add fast mode status to the statusline


miraclebakelaser:
### What variant of Codex are you using?

CLI

### What feature would you like to see?

Add a statusline option that displays the current fast_mode status. 

`/statusline
  ```
  [x] model-with-reasoning  Current model name with reasoning level
  [x] fast-mode            Current fast mode status (omitted when off)
  [x] current-dir           Current working directory
  [ ] context-remaining     Percentage of context window remaining (omitted when unknown)
  [ ] context-window-size   Total context window size in tokens (omitted when unknown)
  [ ] model-name            Current model name
  [ ] project-root          Project root directory (omitted when unavailable)
  [ ] git-branch            Current Git branch (omitted when unavailable)
  [ ] context-used          Percentage of context window used (omitted when unknown)

 gpt-5.4 medium · fast · ~/project/path
```

in `config.toml`
```
status_line = [
   "model-with-reasoning",
   "fast-mode",
   "current-dir",
]
```

etraut-openai:
This is included in the latest version. See [this PR](https://github.com/openai/codex/pull/13670).

PR #13670, opened 2026-03-06
Add Fast mode status-line indicator

etraut-openai:
Addresses feature request #13660

Adds new option to `/statusline` so the status line can display "fast on" or "fast off"

Summary
- introduce a `FastMode` status-line item so `/statusline` can render explicit `Fast on`/`Fast off` text for the service tier
- wire the item into the picker metadata and resolve its string from `ChatWidget` without adding any unrelated `thread-name` logic or storage changes
- ensure the refresh paths keep the cached footer in sync when the service tier (fast mode) changes

Testing
- Manually tested

Here's what it looks like when enabled:

image


mrkn:
I personally like the "⚡️GPT-5.4" label shown in Codex.app. I’d be happy if Codex CLI also had a mode that could display it in the same way.

joshka-oai (approved):
This needs an insta snapshot test that asserts how this looks rather than asserting on values as it makes it easy to see visual drift better. Also +1 to the lightning bolt idea.

pash-openai (approved):

PR #13675, opened 2026-03-06
Allow full web search tool config

rm-openai:
Previously, we could only configure whether web search was on/off.

This PR enables sending along a web search config, which includes all the stuff responsesapi supports: filters, location, etc.


codex-rs/core/src/client_common.rs
```diff
@@ -195,6 +212,66 @@ pub(crate) mod tools {
         }
     }
 
+    fn serialize_web_search_filters<S>(
```
pakrym-oai, line 215: why does this need a custom serializer?

codex-rs/core/src/config/mod.rs
```diff
@@ -1637,6 +1643,59 @@ fn resolve_web_search_mode(
...
+                context_size: profile.context_size.or(base.context_size),
+                allowed_domains: profile
+                    .allowed_domains
+                    .clone()
+                    .or_else(|| base.allowed_domains.clone()),
+                location: match (base.location.as_ref(), profile.location.as_ref()) {
+                    (None, None) => None,
+                    (Some(base_location), None) => Some(base_location.clone()),
+                    (None, Some(profile_location)) => Some(profile_location.clone()),
+                    (Some(base_location), Some(profile_location)) => Some(WebSearchLocation {
```
pakrym-oai: can we maybe add `.merge()` helpers to these. types?

pakrym-oai (approved):
let's make merging less verbose

codex-rs/core/src/client_common.rs
apanasenko-oai: do we export this type as app_server protocol? 

cc: @owenlin0

owenlin0: took a look - luckily no, this is only used for codex harness -> responsesapi calls. opened a PR to clean this up while I am here: https://github.com/openai/codex/pull/14136

PR #13685, opened 2026-03-06
check app auth in plugin/install

sayan-oai:
#### What
on `plugin/install`, check if installed apps are already authed on chatgpt, and return list of all apps that are not. clients can use this list to trigger auth workflows as needed.

checks are best effort based on `codex_apps` loading, much like `app/list`.

#### Tests
Added integration tests, tested locally.


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1713,6 +1713,40 @@ pub struct AppInfo {
     pub plugin_display_names: Vec<String>,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+/// EXPERIMENTAL - app metadata summary for plugin-install responses.
+pub struct AppSummary {
```
sayan-oai, line 1720: not sure if we need all these fields, can reduce to `id`, `name`, and `url` if not needed.

mzeng-openai (approved):

PR #13692, opened 2026-03-06
Add guardian approval MVP

charley-oai:
## Summary
- add the guardian reviewer flow for `on-request` approvals in command, patch, sandbox-retry, and managed-network approval paths
- keep guardian behind `features.guardian_approval` instead of exposing a public `approval_policy = guardian` mode
- route ordinary `OnRequest` approvals to the guardian subagent when the feature is enabled, without changing the public approval-mode surface

## Public model
- public approval modes stay unchanged
- guardian is enabled via `features.guardian_approval`
- when that feature is on, `approval_policy = on-request` keeps the same approval boundaries but sends those approval requests to the guardian reviewer instead of the user
- `/experimental` only persists the feature flag; it does not rewrite `approval_policy`
- CLI and app-server no longer expose a separate `guardian` approval mode in this PR

## Guardian reviewer
- the reviewer runs as a normal subagent and reuses the existing subagent/thread machinery
- it is locked to a read-only sandbox and `approval_policy = never`
- it does not inherit user/project exec-policy rules
- it prefers `gpt-5.4` when the current provider exposes it, otherwise falls back to the parent turn's active model
- it fail-closes on timeout, startup failure, malformed output, or any other review error
- it currently auto-approves only when `risk_score < 80`

## Review context and policy
- guardian mirrors `OnRequest` approval semantics rather than introducing a separate approval policy
- explicit `require_escalated` requests follow the same approval surface as `OnRequest`; the difference is only who reviews them
- managed-network allowlist misses that enter the approval flow are also reviewed by guardian
- the review prompt includes bounded recent transcript history plus recent tool call/result evidence
- transcript entries and planned-action strings are truncated with explicit `<guardian_truncated ... />` markers so large payloads stay bounded
- apply-patch reviews include the full patch content (without duplicating the structured `changes` payload)
- the guardian request layout is snapshot-tested using the same model-visible Responses request formatter used elsewhere in core

## Guardian network behavior
- the guardian subagent inherits the parent session's managed-network allowlist when one exists, so it can use the same approved network surface while reviewing
- exact session-scoped network approvals are copied into the guardian session with protocol/port scope preserved
- those copied approvals are now seeded before the guardian's first turn is submitted, so inherited approvals are available during any immediate review-time checks

## Out of scope / follow-ups
- the sandbox-permission validation split was pulled into a separate PR and is not part of this diff
- a future follow-up can enable `serde_json` preserve-order in `codex-core` and then simplify the guardian action rendering further


charley-oai:
@codex review this

codex-rs/core/src/tools/handlers/mod.rs
```diff
@@ -102,6 +102,17 @@ pub(super) fn normalize_and_validate_additional_permissions(
...
+    if sandbox_permissions.requires_escalated_permissions()
+        && !matches!(
+            approval_policy,
+            AskForApproval::OnRequest | AskForApproval::Guardian
+        )
+    {
+        return Err(format!(
+            "approval policy is {approval_policy:?}; reject command — you should not ask for escalated permissions if the approval policy is {approval_policy:?}"
+        ));
+    }
```
dylan-hurd-oai: We should probably get rid of this function entirely - i'm fairly certain we validate this specific condition elsewhere

codex-rs/core/src/tools/handlers/shell.rs
```diff
@@ -341,16 +341,19 @@ impl ShellHandler {
         )
         .map_err(FunctionCallError::RespondToModel)?;
 
-        // Approval policy guard for explicit escalation in non-OnRequest modes.
-        if exec_params.sandbox_permissions.requests_sandbox_override()
```
dylan-hurd-oai, line 345: are we going to hit a merge conflict with the rename here?

codex-rs/core/src/tools/handlers/shell.rs
```diff
@@ -410,6 +414,18 @@ impl ShellHandler {
             match shell_runtime_backend {
                 Generic => ShellRuntime::new(),
                 backend @ (ShellCommandClassic | ShellCommandZshFork) => {
+                    // Guardian review is only implemented in the standard
+                    // ShellRuntime approval path. The zsh-fork backend has its
+                    // own escalation/approval stack, so keep guardian on the
+                    // classic shell_command backend for the MVP.
```
dylan-hurd-oai: is the zsh-fork path conflicting? why wouldn't we still allow the model to attempt to escalate?

codex-rs/core/src/tools/sandboxing.rs
```diff
@@ -109,8 +109,8 @@ where
 
 #[derive(Clone)]
 pub(crate) struct ApprovalCtx<'a> {
-    pub session: &'a Session,
-    pub turn: &'a TurnContext,
+    pub session: &'a Arc<Session>,
+    pub turn: &'a Arc<TurnContext>,
```
dylan-hurd-oai, line 115: @codex is this a safe change?

codex-rs/core/src/codex_tests.rs
```diff
@@ -3892,11 +3935,117 @@ async fn rejects_escalated_permissions_when_policy_not_on_request() {
     pretty_assertions::assert_eq!(exec_output.metadata, ResponseExecMetadata { exit_code: 0 });
     assert!(exec_output.output.contains("hi"));
 }
+
 #[tokio::test]
-async fn unified_exec_rejects_escalated_permissions_when_policy_not_on_request() {
+async fn guardian_allows_shell_additional_permissions_requests_past_policy_validation() {
```
dylan-hurd-oai: Can we keep the existing test in place here? and move the new guardian-specific tests to their own file?

codex-rs/core/src/guardian.rs
```diff
@@ -0,0 +1,760 @@
...
+    } else {
+        ReviewDecision::Denied
+    }
+}
+
+/// Builds the guardian prompt from three pieces:
+/// - the policy prompt
+/// - a compact transcript for authorization and local context
+/// - the exact action JSON being proposed for escalation
+async fn build_guardian_prompt(
```
dylan-hurd-oai: I think we'll likely want to iterate on this, but this seems like a solid starting point

codex-rs/core/src/guardian.rs
```diff
@@ -0,0 +1,760 @@
...
+        },
+        "required": ["risk_level", "risk_score", "rationale", "evidence"]
+    })
+}
+
+/// Embedded policy prompt for the guardian MVP.
+///
+/// Keeping the policy text local to this file makes the review contract easy to
+/// audit alongside the parsing and approval threshold logic.
+fn guardian_policy_prompt() -> &'static str {
```
dylan-hurd-oai: [non-blocking] nit: we have this and also guardian_prompt.md. should we dedupe?

codex-rs/core/src/guardian.rs
```diff
@@ -0,0 +1,760 @@
...
+        Some(network_proxy) => Some(network_proxy.proxy().current_cfg().await?),
+        None => None,
+    };
+    let guardian_config =
+        build_guardian_subagent_config(turn.config.as_ref(), live_network_config)?;
+
+    // `run_codex_thread_one_shot` is already the subagent runner used elsewhere
+    // in core. Reusing it here keeps the MVP aligned with the existing review
+    // subagent model instead of introducing a guardian-specific execution path.
+    let codex = run_codex_thread_one_shot(
```
dylan-hurd-oai: @codex will this config use the use the user's exec policy rules? if so, can we disable it?

dylan-hurd-oai (approved):
Some non-blocking comments - overall this feels pretty reasonable. Excited to try it out!

codex-rs/core/src/guardian.rs
```diff
@@ -0,0 +1,760 @@
...
+    approx_token_count(&render_transcript(
+        entries,
+        included_numbers,
+        per_entry_token_cap,
+    ))
+}
+
+/// Retains only human-readable user/assistant message content for guardian
+/// review and skips synthetic contextual scaffolding that would just add noise.
+fn collect_guardian_transcript_entries(items: &[ResponseItem]) -> Vec<GuardianTranscriptEntry> {
```
pakrym-oai, line 400: don't we want previous tool calls here?

codex-rs/core/src/tools/handlers/shell.rs
charley-oai: Not conflicting, it's just extra complexity that might be worth doing in a followup PR. Is that fine? I don't have good intuition for whether zsh-fork is commonly used or needed

There is also a design wrinkle: zsh-fork works at the intercepted execve level, so guardian would need a separate review shape there. The
classic path reviews the original shell action JSON; zsh-fork sees resolved executable invocations and can prompt at a different granularity.

codex-rs/core/src/tools/sandboxing.rs
charley-oai: Why this is necessary:

- Existing subagent flows like review in review.rs are called from places that already own Arc<Session> / Arc<TurnContext> (or can clone them
  directly), so they can pass those into run_codex_thread_one_shot(...) without changing any shared abstraction.
- The guardian subagent is launched from inside the tool approval hook, via start_approval_async(...) on the runtime in shell.rs and similar
  paths.
- That hook gets an ApprovalCtx from sandboxing.rs. Historically, that context only exposed borrowed session/turn references because all
  approval flows ended in request_command_approval(...), which only needs borrowed access.

So the answer is not “normal subagents don’t need Arcs.”
They do.
It’s that the normal subagent callsites already had them, while the approval-hook abstraction had erased them.

codex-rs/core/src/guardian.rs
charley-oai: Good point I think we need those

codex-rs/core/src/guardian.rs
charley-oai: As written before the fix, yes: the guardian subagent would have inherited the user/project exec-policy rules, because
`run_codex_thread_one_shot(...)` eventually goes through `Codex::spawn(...)`, and that loaded ExecPolicyManager from the cloned config stack for
every session.

I changed that locally so guardian does not inherit those rules. The behavior now is:

- guardian still reuses the parent’s managed-network allowlist
- guardian still runs read-only with approval_policy = never
- guardian gets an empty ExecPolicyManager instead of the parent’s rules/*.rules

Why this is the right behavior:

- exec-policy rules are user/project shaping, and could silently auto-approve or forbid commands inside the reviewer
- the guardian reviewer should rely on built-in safety checks and its prompt/policy, not caller-supplied exec-policy config
- otherwise the thing being reviewed could influence the reviewer’s own command behavior

codex-rs/core/src/tools/handlers/shell.rs
charley-oai: I actually cut the rename PR out of this one so no

codex-rs/core/src/guardian.rs
dylan-hurd-oai: thank you!

PR #13695, opened 2026-03-06
utils/pty: add streaming spawn and terminal sizing primitives

euroelessar:
Split stdout and stderr routing into reusable output sinks, expose explicit streaming spawn helpers, and thread terminal sizing through the shared PTY layer.

This keeps the spawn helpers intact so existing callers only need mechanical call-site updates while app-server gains the lower-level primitives it needs for interactive command execution.


codex-rs/utils/pty/src/pipe.rs
```diff
@@ -82,28 +84,29 @@ where
...
-async fn spawn_process_with_stdin_mode(
+pub async fn spawn_streaming_process(
     program: &str,
     args: &[String],
     cwd: &Path,
     env: &HashMap<String, String>,
     arg0: &Option<String>,
     stdin_mode: PipeStdinMode,
-) -> Result<SpawnedProcess> {
+    output_sink: OutputSink,
```
pakrym-oai: why does the sink need to be passed in? Why not keep the same design as before where we return the output (even if we split output in two separate streams)

codex-rs/utils/pty/src/process.rs
```diff
@@ -138,7 +202,102 @@ impl Drop for ProcessHandle {
...
 
-/// Return value from spawn helpers (PTY or pipe).
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum OutputStream {
+    Stdout,
+    Stderr,
+}
+
+#[derive(Clone, Debug)]
+pub enum OutputSink {
```
pakrym-oai: This type feels complicated. Can we have the output always in a separate mode and a helper that creates a merged consumer?

codex-rs/utils/pty/src/process.rs
```diff
@@ -29,11 +54,14 @@ impl fmt::Debug for PtyHandles {
 
 /// Handle for driving an interactive process (PTY or pipe).
 pub struct ProcessHandle {
-    writer_tx: mpsc::Sender<Vec<u8>>,
-    output_tx: broadcast::Sender<Vec<u8>>,
+    writer_tx: StdMutex<Option<mpsc::Sender<Vec<u8>>>>,
+    output_tx: Option<broadcast::Sender<Vec<u8>>>,
     killer: StdMutex<Option<Box<dyn ChildTerminator>>>,
+    #[allow(dead_code)]
```
pakrym-oai: why did this become dead code?

codex-rs/utils/pty/README.md
```diff
@@ -4,22 +4,29 @@ Lightweight helpers for spawning interactive processes either under a PTY (pseud
...
 - `spawn_pipe_process_no_stdin(program, args, cwd, env, arg0)` → `SpawnedProcess`
+- `pty::spawn_streaming_process(program, args, cwd, env, arg0, size, output_sink)` → `SpawnedStreamingProcess`
+- `pipe::spawn_streaming_process(program, args, cwd, env, arg0, stdin_mode, output_sink)` → `SpawnedStreamingProcess`
 - `conpty_supported()` → `bool` (Windows only; always true elsewhere)
+- `TerminalSize { rows, cols }` selects PTY dimensions in character cells.
 - `ProcessHandle` exposes:
   - `writer_sender()` → `mpsc::Sender<Vec<u8>>` (stdin)
   - `output_receiver()` → `broadcast::Receiver<Vec<u8>>` (stdout/stderr merged)
+  - `resize(TerminalSize)`
+  - `close_stdin()`
```
pakrym-oai, line 16: Why do we need an explicit close_stdin?

codex-rs/utils/pty/src/process.rs
euroelessar: so, always return two mpsc receivers & have a helper which would spawn a task to merge them & return one broadcast receiver (to maintain behavior of existing call sites)?

pakrym-oai (changes requested):
Looks like we've split how pty process work based on whether we want merged or separate output. I don't think we should push this distinction so deep into the process implementation. Instead the process should always operate in split mode and when needed we can combine streams into a merged one.

codex-rs/utils/pty/README.md
euroelessar: codex app runs some commands which require an ability to explicitly close stdin  (e.g. while running `git apply`)

codex-rs/utils/pty/src/process.rs
euroelessar: oops, bad leftover from prior iteration, thx for catching

codex-rs/utils/pty/src/pipe.rs
euroelessar: updated!

pakrym-oai (approved):

PR #13703, opened 2026-03-06
Clarify sandbox permission override helper semantics

charley-oai:
## Summary
Today `SandboxPermissions::requires_additional_permissions()` does not actually mean "is `WithAdditionalPermissions`". It returns `true` for any non-default sandbox override, including `RequireEscalated`. That broad behavior is relied on in multiple `main` callsites.

The naming is security-sensitive because `SandboxPermissions` is used on shell-like tool calls to tell the executor how a single command should relate to the turn sandbox:
- `UseDefault`: run with the turn sandbox unchanged
- `RequireEscalated`: request execution outside the sandbox
- `WithAdditionalPermissions`: stay sandboxed but widen permissions for that command only

## Problem
The old helper name reads as if it only applies to the `WithAdditionalPermissions` variant. In practice it means "this command requested any explicit sandbox override."

That ambiguity made it easy to read production checks incorrectly and made the guardian change look like a standalone `main` fix when it is not.

On `main` today:
- `shell` and `unified_exec` intentionally reject any explicit `sandbox_permissions` request unless approval policy is `OnRequest`
- `exec_policy` intentionally treats any explicit sandbox override as prompt-worthy in restricted sandboxes
- tests intentionally serialize both `RequireEscalated` and `WithAdditionalPermissions` as explicit sandbox override requests

So changing those callsites from the broad helper to a narrow `WithAdditionalPermissions` check would be a behavior change, not a pure cleanup.

## What This PR Does
- documents `SandboxPermissions` as a per-command sandbox override, not a generic permissions bag
- adds `requests_sandbox_override()` for the broad meaning: anything except `UseDefault`
- adds `uses_additional_permissions()` for the narrow meaning: only `WithAdditionalPermissions`
- keeps `requires_additional_permissions()` as a compatibility alias to the broad meaning for now
- updates the current broad callsites to use the accurately named broad helper
- adds unit coverage that locks in the semantics of all three helpers

## What This PR Does Not Do
This PR does not change runtime behavior. That is intentional.

### DIFF
--- a/codex-rs/core/src/exec_policy.rs
+++ b/codex-rs/core/src/exec_policy.rs
@@ -538,7 +538,7 @@ pub fn render_decision_for_unmatched_command(
                     // In restricted sandboxes (ReadOnly/WorkspaceWrite), do not prompt for
                     // non‑escalated, non‑dangerous commands — let the sandbox enforce
                     // restrictions (e.g., block network/write) without a user prompt.
-                    if sandbox_permissions.requires_additional_permissions() {
+                    if sandbox_permissions.requests_sandbox_override() {
                         Decision::Prompt
                     } else {
                         Decision::Allow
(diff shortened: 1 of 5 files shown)


charley-oai:
@codex review this

dylan-hurd-oai (approved):
naming is hard - good suggestion!

PR #13707, opened 2026-03-06
memories: make prompts branch-aware

andi-oai:
###### Why/Context/Summary
- Make memory extraction and consolidation branch-aware by passing rollout git-branch hints into stage-1 prompts, carrying branch metadata into phase-2 selection text, and persisting `git_branch` in raw memories and rollout summaries.
- Tighten the read/consolidation prompts so checkout-sensitive memories are treated as scoped to the right cwd/branch
- add a silent-maintenance rule so stale-memory cleanup does not get surfaced to users unless they explicitly asked about memory changes.

###### Test plan
- `just fmt`
- `cargo test -p codex-state`
- `cargo test -p codex-core memories`

### DIFF
--- a/codex-rs/core/src/memories/phase1.rs
+++ b/codex-rs/core/src/memories/phase1.rs
@@ -241,6 +241,7 @@ mod job {
             session,
             &thread.rollout_path,
             &thread.cwd,
+            thread.git_branch.as_deref(),
             stage_one_context,
         )
         .await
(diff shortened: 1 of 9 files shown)


jif-oai (changes requested):
his can't really work. Phase 1 only emit one single raw memory entry. You can't ask the model to emit multiple of them
+ the git info can be updated over time and we do not track this for now

PR #13711, opened 2026-03-06
feat: structured plugin parsing

sayan-oai:
#### What

Add structured `@plugin` parsing and TUI support for plugin mentions.

- Core: switch from plain-text `@display_name` parsing to structured `plugin://...` mentions via `UserInput::Mention` and `[$...](plugin://...)` links in text, same pattern as apps/skills.
- TUI: add plugin mention popup, autocomplete, and chips when typing `$`. Load plugin capability summaries and feed them into the composer; plugin mentions appear alongside skills and apps.
- Generalize mention parsing to a sigil parameter, still defaults to `$`

image

Builds on #13510. Currently clients have to build their own `id` via `plugin@marketplace` and filter plugins to show by `enabled`, but we will add `id` and `available` as fields returned from `plugin/list` soon.

####Tests

Added tests, verified locally.

### DIFF
--- a/codex-rs/app-server/README.md
+++ b/codex-rs/app-server/README.md
@@ -476,6 +476,26 @@ Invoke an app by including `$<app-slug>` in the text input and adding a `mention
 } } }
 ```
 
+### Example: Start a turn (invoke a plugin)
+
+Invoke a plugin by including a UI mention token such as `@sample` in the text input and adding a `mention` input item with the exact `plugin://<plugin-name>@<marketplace-name>` path returned by `plugin/list`.
+
+```json
+{ "method": "turn/start", "id": 35, "params": {
+    "threadId": "thr_123",
+    "input": [
+        { "type": "text", "text": "@sample Summarize the latest updates." },
+        { "type": "mention", "name": "Sample Plugin", "path": "plugin://sample@test" }
+    ]
+} }
+{ "id": 35, "result": { "turn": {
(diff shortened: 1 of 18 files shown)


codex-rs/app-server/README.md
```diff
@@ -476,6 +476,26 @@ Invoke an app by including `$<app-slug>` in the text input and adding a `mention
 } } }
 ```
 
+### Example: Start a turn (invoke a plugin)
+
+Invoke a plugin by including a UI mention token such as `@sample` in the text input and adding a `mention` input item with the exact `plugin://<plugin-name>@<marketplace-name>` path returned by `plugin/list`.
```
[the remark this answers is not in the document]
xl-openai: We will add it soon.it will be called id.

codex-rs/core/src/skills/injection.rs
```diff
@@ -178,12 +178,14 @@ impl<'a> ToolMentions<'a> {
 pub(crate) enum ToolMentionKind {
```
xl-openai, line 178: maybe move this file out of the skills folder as it is used for more than just skills? no preference.

xl-openai (approved):

codex-rs/core/src/skills/injection.rs
sayan-oai: fair, will split up that file in a follow up so this pr stays small.

PR #13712, opened 2026-03-06
feat: Add curated plugin marketplace + Metadata Cleanup.

xl-openai:
1. Add a synced curated plugin marketplace and include it in marketplace discovery.
2. Expose optional plugin.json interface metadata in plugin/list
3. Tighten plugin and marketplace path handling using validated absolute paths.
4. Let manifests override skill, MCP, and app config paths.
5. Restrict plugin enablement/config loading to the user config layer so plugin enablement is at global level


codex-rs/app-server-protocol/src/protocol/v2.rs
sayan-oai: nit: we should update this to say that the curated root is also considered when present

codex-rs/core/src/plugins/manifest.rs
```diff
@@ -1,12 +1,90 @@
...
+#[serde(rename_all = "camelCase")]
 pub(crate) struct PluginManifest {
+    #[serde(default)]
     name: String,
+    // Keep manifest paths as raw strings so we can validate the required `./...` syntax before
+    // resolving them under the plugin root.
+    #[serde(default)]
+    skills: Option<String>,
+    #[serde(default)]
+    hooks: Option<String>,
```
sayan-oai: should we include `hooks` before its officially added? or can we just add it once it's been implemented, and maybe leave a comment for now? right now the field isn't consumed.

codex-rs/core/src/plugins/manager.rs
```diff
@@ -217,6 +229,8 @@ impl PluginsManager {
             return PluginLoadOutcome::default();
         }
 
+        self.start_curated_repo_sync();
```
sayan-oai: `plugin/list` reaches this indirectly, which starts sync in the background, but marketplace listing still just checks whether the curated checkout already exists on disk. so on the first call after enabling plugins, `plugin/list` can omit curated entries just because the async clone/update has not finished yet.

should we move curated-catalog init to a marketplace/catalog lifecycle (i.e., first `plugin/list` after opt-in, or when enabling plugins feature)? and either block until sync is done or return an explicit syncing/not-ready state? or is best-effort enough? dont have a strong opinion, but want to clarify.

codex-rs/core/src/plugins/manager.rs
```diff
@@ -280,13 +294,16 @@ impl PluginsManager {
         config: &Config,
         additional_roots: &[AbsolutePathBuf],
     ) -> Result<Vec<ConfiguredMarketplaceSummary>, MarketplaceError> {
+        let installed_plugins = configured_plugins_from_stack(&config.config_layer_stack)
```
sayan-oai, line 298: right now `installed` just means there's an entry for the plugin in config, should we at least check if the plugin dir exists in cache? the user could have old plugins listed in config so `installed` seems like its too strong.

codex-rs/core/src/plugins/manager.rs
sayan-oai: basically it feels a bit unexpected to have this network/git side effect run in what seems to be a read helper: `plugins_for_layer_stack()`. im wondering if this should be a more first-class member or lifecycle in `PluginsManager`.

sayan-oai (approved):
also approving to unblock as the logic all seems sound, most of my comments are about organization/nits

codex-rs/core/src/plugins/manager.rs
xl-openai: Good call move to PluginsManager.

PR #13777, opened 2026-03-06
Fix subagent notifications after interruptive resend

gabec-openai:
## Summary
- share a single `AgentControl` across `ThreadManager` clones so interrupt-follow-up state is visible to every parent/child coordination path
- suppress transient `Interrupted` statuses during interruptive resend using a generation-based follow-up handle in the parent-visible status path
- apply the same reportable-status logic to `send_input` and `wait`, and cover the original and P1 repro cases with targeted tests

## Testing
- `cargo test -p codex-core completion_watcher -- --nocapture`
- `cargo test -p codex-core interruptive_resend -- --nocapture`

### DIFF
--- a/codex-rs/core/src/agent/control.rs
+++ b/codex-rs/core/src/agent/control.rs
@@ -22,12 +22,75 @@ use codex_protocol::protocol::SessionSource;
 use codex_protocol::protocol::SubAgentSource;
 use codex_protocol::protocol::TokenUsage;
 use codex_protocol::user_input::UserInput;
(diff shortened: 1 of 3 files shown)


jif-oai (commented):
> share a single AgentControl across ThreadManager clones so interrupt-follow-up state is visible to every parent/child coordination path

Actually we can't do this. The AgentControl is made to be shared only within a cohort of parent -> child agents... but the PR shows the problem quite well. Let me think on this :)

PR #13789, opened 2026-03-06
fix: reject global wildcard network proxy domains

viyatb-oai:
## Summary
- reject the global `*` domain pattern in proxy allow/deny lists and managed constraints introduced for testing earlier
- keep exact hosts plus scoped wildcards like `*.example.com` and `**.example.com`
- update docs and regression tests for the new invalid-config behavior


rreichel3-oai (approved):
Good catch

PR #13791, opened 2026-03-06
fix(core): skip exec approval for permissionless skill scripts

celia-oai:
## Summary

- Treat skill scripts with no permission profile, or an explicitly empty one, as permissionless and run them with the turn's existing sandbox instead of forcing an exec approval prompt.
- Keep the approval flow unchanged for skills that do declare additional permissions.
- Update the skill approval tests to assert that permissionless skill scripts do not prompt on either the initial run or a rerun.

## Why

Permissionless skills should inherit the current turn sandbox directly. Prompting for exec approval in that case adds friction without granting any additional capability.

### DIFF
--- a/codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
+++ b/codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
@@ -570,9 +570,22 @@ impl EscalationPolicy for CoreShellActionProvider {
         // In the usual case, the execve wrapper reports the command being
         // executed in `program`, so a direct skill lookup is sufficient.
         if let Some(skill) = self.find_skill(program).await {
-            // For now, we always prompt for scripts that look like they belong
-            // to skills, which means we ignore exec policy rules for those
-            // scripts.
+            // For now, scripts that look like they belong to skills bypass
+            // general exec policy evaluation. Permissionless skills inherit the
+            // turn sandbox directly; skills with declared permissions still
+            // prompt here before applying their permission profile.
+            let prompt_permissions = skill.permission_profile.clone();
+            if prompt_permissions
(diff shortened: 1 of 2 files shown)


codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
```diff
@@ -570,9 +570,22 @@ impl EscalationPolicy for CoreShellActionProvider {
...
+            if prompt_permissions
+                .as_ref()
+                .is_none_or(PermissionProfile::is_empty)
+            {
+                tracing::debug!(
+                    "Matched {program:?} to permissionless skill {skill:?}, inheriting turn sandbox"
+                );
+                return Ok(EscalationDecision::escalate(
+                    EscalationExecution::TurnDefault,
+                ));
```
leoshimo-oai, line 587: Does this need to switch based on `approval_policy` is `RejectAskForApproval`, or is that enforced at some other layer?

leoshimo-oai (approved):

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
leoshimo-oai: i.e. if `RejectAskForApproval`'s `sandboxApproval` is `false`, I actually do expect an approval prompt in that case.

dylan-hurd-oai (approved):

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
celia-oai: this will be a follow-up PR!

PR #13807, opened 2026-03-06
[elicitations] Support always allow option for mcp tool calls.

mzeng-openai:
- [x] Support always allow option for mcp tool calls, writes to config.toml.
- [x] Fix config hot-reload after starting a new thread for TUI.


codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -736,15 +730,20 @@ fn parse_mcp_tool_approval_elicitation_response(
...
+            match response
                 .meta
                 .as_ref()
                 .and_then(serde_json::Value::as_object)
                 .and_then(|meta| meta.get(MCP_TOOL_APPROVAL_PERSIST_KEY))
                 .and_then(serde_json::Value::as_str)
-                == Some(MCP_TOOL_APPROVAL_PERSIST_SESSION)
             {
-                return McpToolApprovalDecision::AcceptAndRemember;
+                Some(MCP_TOOL_APPROVAL_PERSIST_SESSION) => {
```
mzeng-openai, line 964: Client passes _meta["persist"] = "session" | "always" to indicate that we should persist this choice.

canvrno-oai (approved):

PR #13825, opened 2026-03-07
feat(core): support custom model aliases in config.toml

friel-openai:
## Summary
- add `config.toml`-defined custom model aliases that map to provider model slugs
- apply alias-specific `model_context_window` and `model_auto_compact_token_limit` overrides
- list configured custom aliases before bundled and remote models in the picker
- add config/schema/docs coverage plus a regression test for the `[[custom_models]]` array form

## Testing
- `cargo test -p codex-core models_manager::manager::tests::get_model_info_uses_custom_alias_metadata_and_request_model -- --exact`
- `cargo test -p codex-core models_manager::manager::tests::build_available_models_lists_custom_aliases_before_remote_models -- --exact`
- `cargo test -p codex-core config::tests::config_toml_deserializes_custom_models -- --exact`
- `just fix -p codex-core`


codex-rs/core/src/models_manager/manager.rs
```diff
@@ -202,10 +208,41 @@ impl ModelsManager {
...
+        custom_model: Option<&CustomModelConfig>,
     ) -> ModelInfo {
+        if let Some(custom_model) = custom_model {
+            let mut config = config.clone();
+            config.model_context_window = config
+                .model_context_window
+                .or(custom_model.model_context_window);
+            config.model_auto_compact_token_limit = config
+                .model_auto_compact_token_limit
+                .or(custom_model.model_auto_compact_token_limit);
```
nornagon-openai: is it intentional that the config-wide values take precedence here?

friel-openai: From Codex: Yes, that precedence was backwards. I flipped it so the alias-specific `model_context_window` / `model_auto_compact_token_limit` win when the alias is selected, and added a regression test covering global-vs-alias precedence.

PR #13829, opened 2026-03-07
add js_repl polling and reuse unified_exec managed process lifecycle

fjord-oai:
## Summary

This PR adds polling-mode `js_repl` execution and reuses the shared sandbox / managed-process infrastructure for `js_repl` kernel startup.

User-visible behavior:
- `js_repl` can now run in polling mode with reusable session ids
- `js_repl_poll` returns incremental `logs`, terminal `final_output`, and terminal `error`
- polling supports multimodal output, including explicit `codex.emitImage(...)`
- `js_repl_reset` clears polling session state, including completed exec state

Infrastructure changes:
- `js_repl` kernel launch now uses the shared sandbox-attempt path
- `js_repl` kernel lifecycle now uses the shared managed-process substrate
- `unified_exec` / `codex_utils_pty` now support the split stdout/stderr process shape that `js_repl` needs
- app-server command exec was updated to the current process-spawn API

### Polling semantics
- adds polling sessions for `js_repl`
- adds `js_repl_poll` and `js_repl_reset`
- poll results expose:
  - `logs`: incremental text since the previous poll
  - `final_output`: terminal `ExecResult.output`
  - `error`: terminal error, if any
- completed poll responses can include multimodal `content_items`

### Multimodal behavior
- `codex.emitImage(...)` can attach images to outer `js_repl` output in both poll and non-poll mode
- nested tool results such as `view_image` are not surfaced automatically; they must be explicitly emitted
- includes positive and negative coverage for explicit image emission

### Process / sandbox reuse
- `js_repl` now uses the same initial sandbox setup path as the main tool flow
- `js_repl` kernel startup now goes through the shared managed child-process layer
- split stdout/stderr process support was added where needed so `js_repl` can keep parsing stdout as protocol output while retaining stderr separately for diagnostics

### Other correctness changes
- `js_repl` itself remains non-parallel-safe, since non-poll mode still uses the singleton kernel
- poll sessions are only made reusable after nested `codex.tool(...)` results have been written back to the kernel
- reset clears completed polling state instead of leaving old `exec_id`s replayable
- idle-session pruning no longer removes unrelated state before a replacement kernel is successfully installed

## Why This Does Not Reuse All Of `unified_exec`

This PR reuses the parts of `unified_exec` that match `js_repl` well:
- sandbox setup
- managed child lifecycle
- spawn plumbing
- split-process support

It does not try to make `js_repl` a thin wrapper over generic `unified_exec` polling/completion.

That is because the models are different:
- `unified_exec` is process-centric
- `js_repl` is per-`exec_id` inside a long-lived kernel

In practice:
- `unified_exec` polling drains raw process output
- `js_repl` polling returns structured per-exec state
- `unified_exec` completion is process/session-oriented
- `js_repl` completion is a kernel `ExecResult` while the kernel remains alive
- `js_repl` also supports async kernel->host tool calls during execution

So the reuse boundary here is intentional: share the process/sandbox substrate, keep the `js_repl` protocol/session state machine separate.


codex-rs/app-server/src/command_exec.rs
```diff
@@ -462,36 +479,58 @@ async fn run_command(params: RunCommandParams) {
...
+    let (session, exit_rx, stdout_handle, stderr_handle) = match spawned {
+        SpawnedCommand::Merged(SpawnedProcess {
+            session,
+            output_rx,
+            exit_rx,
+        }) => {
+            let stdout_handle = spawn_process_output(SpawnProcessOutputParams {
+                connection_id: request_id.connection_id,
+                process_id: process_id.clone(),
+                output_rx: CommandOutputReceiver::Broadcast(output_rx),
```
jif-oai: Why do we drop the Mpsc? This means we will drop the lagged items and might not honor the `HeadTailBuffer` correctly

codex-rs/core/src/tools/js_repl/mod.rs
```diff
@@ -1,14 +1,13 @@
 use std::collections::HashMap;
```
jif-oai, line 1: Can we split down this 4k lines file and document the public functions?

codex-rs/core/src/tools/js_repl/mod.rs
```diff
@@ -59,6 +72,24 @@ const JS_REPL_EXEC_ID_LOG_LIMIT: usize = 8;
 const JS_REPL_MODEL_DIAG_STDERR_MAX_BYTES: usize = 1_024;
 const JS_REPL_MODEL_DIAG_ERROR_MAX_BYTES: usize = 256;
 const JS_REPL_TOOL_RESPONSE_TEXT_PREVIEW_MAX_BYTES: usize = 512;
+const JS_REPL_POLL_MIN_MS: u64 = 50;
+const JS_REPL_POLL_MAX_MS: u64 = crate::unified_exec::DEFAULT_MAX_BACKGROUND_TERMINAL_TIMEOUT_MS;
+const JS_REPL_POLL_DEFAULT_MS: u64 = crate::unified_exec::MIN_EMPTY_YIELD_TIME_MS;
+const JS_REPL_POLL_MAX_SESSIONS: usize = 16;
```
jif-oai, line 88: We also need a cap on completed execs in `exec_store`. Otherwise, the 300s graceful period can explode the number of closed `ExecBuffer` that holds a lot of things including a `Notify`. Those are limited on most OS so residency will become and issue

codex-rs/utils/pty/src/process.rs
```diff
@@ -171,9 +179,7 @@ impl ProcessHandle {
             }
         }
         if let Ok(mut h) = self.wait_handle.lock() {
```
jif-oai, line 180: Why? This means you detach the child process so if the cancel fail, the orphan process will stay up forever

codex-rs/app-server/src/command_exec.rs
fjord-oai: Thanks for the catch!

I fixed this at the shared `codex_utils_pty` layer instead of just in `command_exec.rs`:
- merged `SpawnedProcess.output_rx` is back to a lossless `mpsc::Receiver<Vec<u8>>`
- `app-server` now consumes that directly again
- `unified_exec` rebuilds its internal broadcast fanout from the lossless primary receiver, so it keeps the subscription model it needs without making the spawn API lossy

So the merged-output semantics are back in line with the pre-PR behavior.

codex-rs/core/src/tools/js_repl/mod.rs
fjord-oai: Agreed. I added a hard cap on retained completed poll execs in `exec_store` on top of the existing 300s replay window.

Completed entries now record completion order, and `complete_exec_in_store(...)` immediately evicts the oldest completed entries once the cap is exceeded, while preserving active execs. I also added a regression test covering that prune path.

codex-rs/utils/pty/src/process.rs
fjord-oai: I don’t think we want to switch this back.

`wait_handle` is only the background waiter/reaper. Dropping its `JoinHandle` detaches that task, but does not stop it, so it can still observe `child.wait()`, update exit state, and reap the child if it exits later.

Aborting that handle would not kill the process immediately. It would only remove the path that observes eventual exit and performs cleanup/bookkeeping. So if the initial kill fails, aborting the waiter actually makes things worse: the process can still keep running, but now we have also lost the only task that can notice a later exit.

The real limitation here is that termination is currently single-shot: if the first `kill()` fails, we do not have a later escalation/retry path. I think that is the right follow-up to discuss, rather than aborting `wait_handle`.

codex-rs/core/src/tools/js_repl/mod.rs
fjord-oai: Done:
- protocol types into `protocol.rs`
- node/version helpers into `node.rs`
- output/result helpers into `output.rs`
- unit tests into `tests.rs`

Also added rustdoc on the public `js_repl` surface.

PR #13830, opened 2026-03-07
split sandbox permission validation modes

charley-oai:
## Summary
- split `RequireEscalated` validation from `WithAdditionalPermissions` validation
- stop treating `with_additional_permissions` as if it were an unsandboxed escalation request
- add focused core regressions for both shell and unified exec

## Why
`SandboxPermissions` has two distinct non-default modes:
- `RequireEscalated`: request to run outside the sandbox
- `WithAdditionalPermissions`: stay sandboxed, but widen permissions for one command

On `main`, the handler-local checks in `shell` and `unified_exec` use `requests_sandbox_override()`, which conflates those two modes. That means a `with_additional_permissions` request in a non-`OnRequest` policy gets rejected as if it were an explicit unsandboxed execution request.

This patch makes the validation match the enum contract:
- shared validation rejects `RequireEscalated` unless the policy is `OnRequest`
- handler-local guards only handle the `WithAdditionalPermissions` flow

## Testing
- `just fmt`
- local `cargo test` was started for the targeted `codex-core` cases, but Cargo build-lock contention made it a poor signal; relying on CI for the authoritative test result

### DIFF
--- a/codex-rs/core/src/tools/handlers/mod.rs
+++ b/codex-rs/core/src/tools/handlers/mod.rs
@@ -102,6 +102,14 @@ pub(super) fn normalize_and_validate_additional_permissions(
         SandboxPermissions::WithAdditionalPermissions
     );
 
+    if sandbox_permissions.requires_escalated_permissions()
+        && approval_policy != AskForApproval::OnRequest
+    {
+        return Err(format!(
+            "approval policy is {approval_policy:?}; reject command — you should not ask for escalated permissions if the approval policy is {approval_policy:?}"
+        ));
+    }
+
     if !request_permission_enabled
         && (uses_additional_permissions || additional_permissions.is_some())
     {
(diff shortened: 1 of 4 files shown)


codex-rs/core/src/tools/handlers/mod.rs
```diff
@@ -102,6 +102,14 @@ pub(super) fn normalize_and_validate_additional_permissions(
         SandboxPermissions::WithAdditionalPermissions
     );
 
+    if sandbox_permissions.requires_escalated_permissions()
+        && approval_policy != AskForApproval::OnRequest
+    {
+        return Err(format!(
+            "approval policy is {approval_policy:?}; reject command — you should not ask for escalated permissions if the approval policy is {approval_policy:?}"
+        ));
+    }
```
dylan-hurd-oai, line 111: we shouldn't be doing this check in here, that's what the handler-local checks are for

dylan-hurd-oai (changes requested):
we can separate these errors if we want but the handler-local tools should not stop enforcing the validity of RequireEscalated when not in OnRequest.

PR #13850, opened 2026-03-07
feat: make interrupt state not final for multi-agents

jif-oai:
Make `interrupted` an agent state and make it not final. As a result, a `wait` won't return on an interrupted agent and no notification will be send to the parent agent.

The rationals are:
* If a user interrupt a sub-agent for any reason, you don't want the parent agent to instantaneously ask the sub-agent to restart
* If a parent agent interrupt a sub-agent, no need to add a noisy notification in the parent agen


gabec-openai (approved):
locally patched - seems to work much better!

aibrahim-oai (approved):

PR #13878, opened 2026-03-07
Serialize shell snapshot stdin test

aibrahim-oai:
## What changed
- `snapshot_shell_does_not_inherit_stdin` now runs under its own serial key.
- The change isolates it from other Unix shell-snapshot tests that also interact with stdin.

## Why this fixes the flake
- The failure was not a shell-snapshot logic bug. It was shared-stdin interference between concurrently executing tests.
- When multiple tests compete for inherited stdin at the same time, one test can observe EOF or consumed input that actually belongs to a different test.
- Running this specific test in a dedicated serial bucket guarantees exclusive ownership of stdin, which makes the assertion deterministic without weakening coverage.

## Scope
- Test-only change.

### DIFF
--- a/codex-rs/core/src/shell_snapshot.rs
+++ b/codex-rs/core/src/shell_snapshot.rs
@@ -740,7 +740,11 @@ mod tests {
 
         let dir = tempdir()?;
         let home = dir.path();
-        fs::write(home.join(".bashrc"), "read -r ignored\n").await?;
+        let read_status_path = home.join("stdin-read-status");
+        let read_status_display = read_status_path.display();
+        let bashrc =
+            format!("read -t 1 -r ignored\nprintf '%s' \"$?\" > \"{read_status_display}\"\n");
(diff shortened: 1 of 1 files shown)


codex-rs/core/src/shell_snapshot.rs
```diff
@@ -740,7 +740,11 @@ mod tests {
 
         let dir = tempdir()?;
         let home = dir.path();
-        fs::write(home.join(".bashrc"), "read -r ignored\n").await?;
+        let read_status_path = home.join("stdin-read-status");
+        let read_status_display = read_status_path.display();
+        let bashrc =
+            format!("read -t 1 -r ignored\nprintf '%s' \"$?\" > \"{read_status_display}\"\n");
```
jif-oai, line 748: Can you add a comment in the code? I don't understand what the print here is for?

jif-oai (approved):

PR #13880, opened 2026-03-07
Stabilize RMCP streamable HTTP readiness tests

aibrahim-oai:
## What changed
- The RMCP streamable HTTP tests now wait for metadata and tool readiness before issuing tool calls.
- OAuth state is isolated per test home.
- The helper server startup path now uses bounded bind retries so transient `AddrInUse` collisions do not fail the test immediately.

## Why this fixes the flake
- The old tests could begin issuing tool requests before the helper server had finished advertising its metadata and tools, so the first request sometimes raced the server startup sequence.
- On top of that, shared OAuth state and occasional bind collisions on CI runners introduced cross-test environmental noise unrelated to the functionality under test.
- Readiness polling makes the client wait for an observable “server is ready” signal, while isolated state and bounded bind retries remove external contention that was causing intermittent failures.

## Scope
- Test-only change.

### DIFF
--- a/codex-rs/core/tests/suite/rmcp_client.rs
+++ b/codex-rs/core/tests/suite/rmcp_client.rs
@@ -1106,8 +1106,6 @@ async fn wait_for_streamable_http_server(
     let deadline = Instant::now() + timeout;
     let metadata_url = format!("http://{address}/.well-known/oauth-authorization-server/mcp");
     let client = Client::builder().no_proxy().build()?;
-    let mut attempts = 0u32;
-
     loop {
         if let Some(status) = server_child.try_wait()? {
             return Err(anyhow::anyhow!(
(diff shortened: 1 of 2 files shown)


codex-rs/core/tests/suite/rmcp_client.rs
```diff
@@ -851,10 +854,8 @@ async fn streamable_http_tool_call_round_trip() -> anyhow::Result<()> {
     Ok(())
 }
 
-/// This test writes to a fallback credentials file in CODEX_HOME.
-/// Ideally, we wouldn't need to serialize the test but it's much more cumbersome to wire CODEX_HOME through the code.
-#[serial(codex_home)]
```
sayan-oai, line 856: is it ok to drop this given the test still changes process-global `CODEX_HOME` to make the fallback `.credentials.json` path work?

sayan-oai (approved):

PR #13882, opened 2026-03-07
Fix stdio-to-uds peer-close flake

aibrahim-oai:
## What changed
- `codex-stdio-to-uds` now tolerates `NotConnected` when `shutdown(Write)` happens after the peer has already closed.
- The socket test was rewritten to send stdin from a fixture file and to read an exact request payload length instead of waiting on EOF timing.

## Why this fixes the flake
- This one exposed a real cross-platform runtime edge case: on macOS, the peer can close first after a successful exchange, and `shutdown(Write)` can report `NotConnected` even though the interaction already succeeded.
- Treating that specific ordering as a harmless shutdown condition removes the production-level false failure.
- The old test compounded the problem by depending on EOF timing, which varies by platform and scheduler. Exact-length IO makes the test deterministic and focused on the actual data exchange.

## Scope
- Production logic change with matching test rewrite.


aibrahim-oai:
@codex review this

codex-rs/stdio-to-uds/src/lib.rs
```diff
@@ -39,9 +39,13 @@ pub fn run(socket_path: &Path) -> anyhow::Result<()> {
         io::copy(&mut handle, &mut stream).context("failed to copy data from stdin to socket")?;
     }
 
-    stream
-        .shutdown(Shutdown::Write)
-        .context("failed to shutdown socket writer")?;
+    // The peer can close immediately after sending its response. On macOS that
```
bolinfest: Are we certain that this can only happen on macOS?

codex-rs/stdio-to-uds/tests/stdio_to_uds.rs
```diff
@@ -31,37 +37,103 @@ fn pipes_stdin_and_stdout_through_socket() -> anyhow::Result<()> {
...
+            .map_err(|_| anyhow!("failed to send received bytes to test thread"))?;
         connection
             .write_all(b"response")
             .context("failed to write response to client")?;
+        let _ = event_tx.send("wrote response".to_string());
         Ok(())
     });
 
-    Command::new(codex_utils_cargo_bin::cargo_bin("codex-stdio-to-uds")?)
+    let stdin = std::fs::File::open(&request_path).context("failed to open child stdin fixture")?;
```
bolinfest, line 68: I feel like either here or in a docstring on `pipes_stdin_and_stdout_through_socket()`, some of the PR body should be memorialized as a comment to explain why this test code is complicated.

If I understand correctly, a key change is using `read_exact()` in `server_thread` instead of `read_to_end()` to eliminate the dependency on EOF timing.

I'm unclear whether the move from `assert_cmd::Command` to `std::process::Command` is necessary to ensure the test is no longer flaky. Is it mainly there to build up `server_events` so that if the test _does_ fail, the error includes the `server_events` received thus far, making it easier to debug?

bolinfest (approved):

PR #13883, opened 2026-03-07
Stabilize PTY Python REPL test

aibrahim-oai:
## What changed
- The PTY Python REPL test now starts Python with a startup marker already embedded in argv.
- The test waits for that marker in PTY output before making assertions.

## Why this fixes the flake
- The old version tried to probe the live REPL almost immediately after spawn.
- That races PTY initialization, Python startup, and prompt buffering, all of which vary across platforms and CI load.
- By having the child process emit a known marker as part of its own startup path, the test gets a deterministic synchronization point that comes from the process under test rather than from guessed timing.

## Scope
- Test-only change.

### DIFF
--- a/codex-rs/utils/pty/src/tests.rs
+++ b/codex-rs/utils/pty/src/tests.rs
@@ -130,52 +130,36 @@ async fn collect_output_until_exit(
 }
 
 async fn wait_for_python_repl_ready(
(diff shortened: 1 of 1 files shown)


euroelessar (approved):

jif-oai (approved):
I'm not 100% convinced this will be more stable but let's try

PR #13885, opened 2026-03-07
Stabilize thread resume replay tests

aibrahim-oai:
## What changed
- The thread-resume replay tests now use unchecked mock sequencing so the replay flow can complete before the test asserts.
- They also poll outbound `/responses` request counts and fail immediately if replay emits an unexpected extra request.

## Why this fixes the flake
- The previous version asserted while the replay machinery was still mid-flight, so the test was sometimes checking an intermediate state instead of the completed behavior.
- Strict mock sequencing made that problem worse by forcing the test to care about exact sub-step timing rather than about the end result.
- Letting replay settle and then asserting on stabilized request counts makes the test validate the real contract: the replay path finishes and does not send extra model requests.

## Scope
- Test-only change.

### DIFF
--- a/codex-rs/app-server/tests/suite/v2/thread_resume.rs
+++ b/codex-rs/app-server/tests/suite/v2/thread_resume.rs
@@ -4,7 +4,6 @@ use app_test_support::create_apply_patch_sse_response;
 use app_test_support::create_fake_rollout_with_text_elements;
 use app_test_support::create_final_assistant_message_sse_response;
 use app_test_support::create_mock_responses_server_repeating_assistant;
-use app_test_support::create_mock_responses_server_sequence;
 use app_test_support::create_mock_responses_server_sequence_unchecked;
 use app_test_support::create_shell_command_sse_response;
 use app_test_support::rollout_path;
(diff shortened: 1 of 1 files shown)


codex-rs/app-server/tests/suite/v2/thread_resume.rs
```diff
@@ -1146,6 +1177,7 @@ async fn thread_resume_replays_pending_file_change_request_approval() -> Result<
         primary.read_stream_until_notification_message("turn/completed"),
     )
     .await??;
+    wait_for_responses_request_count(&server, 3).await?;
```
owenlin0: hmm, do we care about how many requests that we sent to responses? seems brittle. can we keep just the `create_mock_responses_server_sequence_unchecked` fix and remove this one?

owenlin0 (approved):

codex-rs/app-server/tests/suite/v2/thread_resume.rs
aibrahim-oai: I think the wait is what removes the flakiness

owenlin0: interesting, i don't understand why if it's the last thing we await on in this test 🤷

Issue #13623, opened 2026-03-05
wrong display result of context window after adjusting gpt5.4 to 1M context window size


winoros:
### What version of Codex CLI is running?

v0.111.0

### What subscription do you have?

Pro

### Which model were you using?

gpt-5.4

### What platform is your computer?

Linux 6.17.0-14-generic x86_64 x86_64

### What terminal emulator and version are you using (if applicable)?

ssh

### What issue are you seeing?

I viewed the release article https://openai.com/index/introducing-gpt-5-4/.
So I changed the context window by editing the `config.toml`:
```
model_context_window = 1000000
model_auto_compact_token_limit = 900000
```

I tested some sessions already. None of them estimated the correct context window.
Some of them showed like this after a medium run:
<img width="906" height="51" alt="Image" src="https://github.com/user-attachments/assets/dd4491f7-427e-49a7-9366-5b227fa62fa6" />
It output the used token upon exit. But I think the result is wrong.
`Token usage: total=12,840 input=11,897 (+ 3,456 cached) output=943 (reasoning 516)`

Some of them showed like this after a short run

<img width="807" height="38" alt="Image" src="https://github.com/user-attachments/assets/5a1db73d-c91b-4065-b988-192d5673a733" />
And no used token info after exiting the cli.

### What steps can reproduce the bug?

Change the `config.toml`, then start a session and run any command.

### What is the expected behavior?

I'm setting the correct window size for gpt5.4. I think all things should work as usual.

PR #13896, opened 2026-03-07
Fix TUI context window display before first TokenCount

etraut-openai:
The TUI was showing the raw configured `model_context_window` until the first
`TokenCount` event arrived, even though core had already emitted the effective
runtime window on `TurnStarted`. This made the footer, status-line context
window, and `/status` output briefly inconsistent for models/configs where the
effective window differs from the configured value, such as the `gpt-5.4`
1,000,000-token override reported in #13623.

Update the TUI to cache `TurnStarted.model_context_window` immediately so
pre-token-count displays use the runtime effective window, and add regression
coverage for the startup path.


codex-rs/tui/src/app.rs
```diff
@@ -3316,7 +3316,7 @@ impl App {
     fn handle_codex_event_now(&mut self, event: Event) {
         let needs_refresh = matches!(
             event.msg,
-            EventMsg::SessionConfigured(_) | EventMsg::TokenCount(_)
+            EventMsg::SessionConfigured(_) | EventMsg::TurnStarted(_) | EventMsg::TokenCount(_)
```
charley-oai, line 3319: TurnStarted is needed here because it now changes status-line source data. Before the first TokenCount, the TUI caches TurnStarted.model_context_window so the displayed context window uses the effective runtime value rather than the raw configured value. Without refreshing on TurnStarted, the live status line stays stale until a later TokenCount, which preserves the original bug. SessionConfigured and TokenCount were already refresh points for the same reason: they each change inputs consumed by refresh_status_line().

charley-oai (approved):

PR #13897, opened 2026-03-07
guardian initial feedback / tweaks

charley-oai:
## Summary
- remove the remaining model-visible guardian-specific `on-request` prompt additions so enabling the feature does not change the main approval-policy instructions
- neutralize user-facing guardian wording to talk about automatic approval review / approval requests rather than a second reviewer or only sandbox escalations
- tighten guardian retry-context handling so agent-authored `justification` stays in the structured action JSON and is not also injected as raw retry context
- simplify guardian review plumbing in core by deleting dead prompt-append paths and trimming some request/transcript setup code

## Notable Changes
- delete the dead `permissions/approval_policy/guardian.md` append path and stop threading `guardian_approval_enabled` through model-facing developer-instruction builders
- rename the experimental feature copy to `Automatic approval review` and update the `/experimental` snapshot text accordingly
- make approval-review status strings generic across shell, patch, network, and MCP review types
- forward real sandbox/network retry reasons for shell and unified-exec guardian review, but do not pass agent-authored justification as raw retry context
- simplify `guardian.rs` by removing the one-field request wrapper, deduping reasoning-effort selection, and cleaning up transcript entry collection

## Testing
- `just fmt`
- full validation left to CI


codex-rs/core/src/guardian.rs
```diff
@@ -89,12 +89,6 @@ pub(crate) fn is_guardian_subagent_source(
     )
 }
 
-/// Canonical description of the action the guardian is being asked to review.
-#[derive(Debug, Clone, PartialEq)]
-pub(crate) struct GuardianReviewRequest {
-    pub(crate) action: Value,
-}
```
dylan-hurd-oai, line 96: I'm not super clear on why we have to downgrade the contract here to accept an unstructured json object. shouldn't we be making this _more_ structured?

charley-oai: This is not necessary - I felt it was just a simplification / eliminating a useless wrapper around the Value. I tried out making it structured and the code was way less simple / less readable, and the typefulness didn't seem to buy us anything really (since it's ultimately just destined to be converted to JSON and injected into the conversation, wouldn't ever want to use those values in code I think)

If you feel the type wrapper is useful though I can add it back

dylan-hurd-oai: I guess my point is that we're converting other structured requests, why not use those, keep the data fully-typed, and then only convert to json right where we need it?

We could have something like:

```
pub(crate) enum GuardianApprovalRequest {
  UnifiedExec(UnifiedExecRequest),
  Shell(ShellRequest),
  ApplyPatch(ApplyPatchApprovalRequestEvent)
}
```

charley-oai: Fair enough, I like that better

charley-oai:
@codex review this

dylan-hurd-oai (approved):

PR #13900, opened 2026-03-07
Fix Linux tmux segfault in user shell lookup

etraut-openai:
Replace the Unix shell lookup path in `codex-rs/core/src/shell.rs` to use
`libc::getpwuid_r()` instead of `libc::getpwuid()` when resolving the current
user's shell.

Why:
- `getpwuid()` can return pointers into libc-managed shared storage
- on the musl static Linux build, concurrent callers can race on that storage
- this matches the crash pattern reported in tmux/Linux sessions with parallel
  shell activity

Refs:
- Fixes #13842


qtnx:
Thank you for this. Please consider adding it to the next release. I implemented the same approach in my local fork, and Codex has been running for over two days without any issues in the same environment, setup, and prompt.

aibrahim-oai (approved):

PR #13931, opened 2026-03-08
JS Repl: Add config to experimentally swap js_repl

cconger:
This allows me to experiment with a different js_repl kernel provider in an external process.

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -283,7 +283,7 @@ use crate::tools::ToolRouter;
 use crate::tools::context::SharedTurnDiffTracker;
 use crate::tools::handlers::SEARCH_TOOL_BM25_TOOL_NAME;
 use crate::tools::js_repl::JsReplHandle;
-use crate::tools::js_repl::resolve_compatible_node;
+use crate::tools::js_repl::validate_js_repl_runtime;
 use crate::tools::network_approval::NetworkApprovalService;
 use crate::tools::network_approval::build_blocked_request_observer;
 use crate::tools::network_approval::build_network_policy_decider;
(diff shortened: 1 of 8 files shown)


codex-rs/core/src/tools/js_repl/mod.rs
```diff
@@ -900,7 +926,7 @@ impl JsReplManager {
 
         let mut child = cmd
             .spawn()
-            .map_err(|err| format!("failed to start Node runtime: {err}"))?;
+            .map_err(|err| format!("failed to start js_repl runtime: {err}"))?;
```
fjord-oai, line 929: can we include the exact command that failed in error messages like these, in case it's something unexpected

codex-rs/core/src/tools/js_repl/mod.rs
```diff
@@ -831,11 +860,8 @@ impl JsReplManager {
         }
 
         let spec = CommandSpec {
-            program: node_path.to_string_lossy().to_string(),
-            args: vec![
-                "--experimental-vm-modules".to_string(),
-                kernel_path.to_string_lossy().to_string(),
-            ],
+            program,
```
fjord-oai, line 863: probably worth adding an info! log for the final program/args that gets used

codex-rs/core/src/codex.rs
```diff
@@ -383,17 +383,21 @@ impl Codex {
...
+                config.js_repl_node_path.as_deref(),
+            )
+            .await
         {
             let _ = config.features.disable(Feature::JsRepl);
             let _ = config.features.disable(Feature::JsReplToolsOnly);
             let message = if config.features.enabled(Feature::JsRepl) {
                 format!(
-                    "`js_repl` remains enabled because enterprise requirements pin it on, but the configured Node runtime is unavailable or incompatible. {err}"
+                    "`js_repl` remains enabled because enterprise requirements pin it on, but the configured experimental js_repl runtime is unavailable or invalid. {err}"
```
fjord-oai, line 396: should probably include the program that was attempted here as well

fjord-oai (approved):

PR #13944, opened 2026-03-08
Stabilize subagent completion watcher ordering

aibrahim-oai:
## What changed
- The runtime now attaches the child-completion watcher before sending input to the subagent or resuming an existing child.
- The tests also wait for a final child status before asserting that the parent history was notified.

## Why this fixes the flake
- The bug was a real ordering race: if the child completed before the watcher subscription was attached, the parent could miss the completion signal entirely.
- The old assertions then amplified the problem by checking too early, while the system was still between “child done” and “parent observed that child is done.”
- Attaching the watcher first closes the race window in production, and waiting for the final child status makes the tests observe the stable state rather than an intermediate one.

## Scope
- Production logic change with matching test coverage.


codex-rs/core/src/agent/control.rs
```diff
@@ -429,23 +432,41 @@ impl AgentControl {
...
         };
+
+        let status_rx = self.subscribe_status(child_thread_id).await.ok();
         let control = self.clone();
         tokio::spawn(async move {
-            let status = match control.subscribe_status(child_thread_id).await {
-                Ok(mut status_rx) => {
+            let status = match status_rx {
+                Some(mut status_rx) => {
+                    debug!(%child_thread_id, "subagent completion watcher attached");
```
jif-oai: We should not log this. All logs are being stored now and we don't care
Same for all the logs below

codex-rs/core/src/agent/control.rs
```diff
@@ -212,8 +213,9 @@ impl AgentControl {
         // TODO(jif) add helper for drain
         state.notify_thread_created(new_thread.thread_id);
 
+        self.maybe_start_completion_watcher(new_thread.thread_id, notification_source)
```
jif-oai: You can't start the watcher before being sure the send_input is successulf
Or you need to make sure to kill it if `send_input` fails

PR #13953, opened 2026-03-08
python-sdk: generated type foundation (all v2 schemas)

shaqayeq-oai:
## Summary
Foundation PR only (base for PR #3).

This PR contains the SDK runtime foundation and generated artifacts:

- pinned runtime binary in `sdk/python/bin/` (`codex` or `codex.exe` by platform)
- single maintenance script: `sdk/python/scripts/update_sdk_artifacts.py`
- generated protocol/types artifacts under:
  - `sdk/python/src/codex_app_server/generated/protocol_types.py`
  - `sdk/python/src/codex_app_server/generated/schema_types.py`
  - `sdk/python/src/codex_app_server/generated/v2_all/*`
- generation-contract test wiring (`tests/test_contract_generation.py`)

## Release asset behavior
`update_sdk_artifacts.py` now:
- selects latest release by channel (`--channel stable|alpha`)
- resolves the correct asset for current OS/arch
- extracts platform binary (`codex` on macOS/Linux, `codex.exe` on Windows)
- keeps runtime on single pinned binary source in `sdk/python/bin/`

## Scope boundary
- ✅ PR #2 = binary + generation pipeline + generated types foundation
- ❌ PR #2 does **not** include examples/integration logic polish (that is PR #3)

## Validation
- Ran: `python scripts/update_sdk_artifacts.py --channel stable`
- Regenerated and committed resulting generated artifacts
- Local tests pass on branch


sdk/python/scripts/update_sdk_artifacts.py
```diff
@@ -0,0 +1,981 @@
...
+    ("codex_app_server.generated.v2_all.ThreadResumeParams", "AskForApproval"): "ResumeAskForApproval",
+    ("codex_app_server.generated.v2_all.ThreadResumeParams", "Personality"): "ResumePersonality",
+    ("codex_app_server.generated.v2_all.ThreadResumeParams", "SandboxMode"): "ResumeSandboxMode",
+    ("codex_app_server.generated.v2_all.ThreadForkParams", "AskForApproval"): "ForkAskForApproval",
+    ("codex_app_server.generated.v2_all.ThreadForkParams", "SandboxMode"): "ForkSandboxMode",
+    ("codex_app_server.generated.v2_all.TurnStartParams", "AskForApproval"): "TurnAskForApproval",
+    ("codex_app_server.generated.v2_all.TurnStartParams", "Personality"): "TurnPersonality",
+    ("codex_app_server.generated.v2_all.TurnStartParams", "ReasoningEffort"): "TurnReasoningEffort",
+    ("codex_app_server.generated.v2_all.TurnStartParams", "SandboxPolicy"): "TurnSandboxPolicy",
+    ("codex_app_server.generated.v2_all.TurnStartParams", "ReasoningSummary"): "TurnReasoningSummary",
```
owenlin0: curious why we need these type aliases, since things like `ReasoningSummary` is meant to be a shared type across a couple RPCs (same with the others here like `Personality` etc.)

sdk/python/src/codex_app_server/generated/v2_all/ContextCompactedNotification.py
```diff
@@ -0,0 +1,12 @@
...
+#   timestamp: <normalized>
+
+from __future__ import annotations
+
+from pydantic import BaseModel
+
+
+class ContextCompactedNotification(BaseModel):
+    threadId: str
+    turnId: str
```
owenlin0: to make things more pythonic, wonder if it's possible to produce these generated fields as snake_case, but send it over the wire to app-server as camelCase?

sdk/python/README.md
```diff
@@ -0,0 +1,77 @@
+# Codex App Server Python SDK
```
owenlin0: probably worth adding `(experimental)` for now

sdk/python/src/codex_app_server/generated/protocol_types.py
```diff
@@ -0,0 +1,52 @@
...
+    gitInfo: NotRequired[dict[str, Any] | None]
+    id: str
+    modelProvider: str
+    name: NotRequired[dict[str, Any]]
+    path: NotRequired[dict[str, Any]]
+    preview: str
+    source: dict[str, Any]
+    status: dict[str, Any]
+    turns: list[TurnObject]
+    updatedAt: int
```
owenlin0: snake_case would be so nice here

sdk/python/src/codex_app_server/generated/protocol_types.py
```diff
@@ -0,0 +1,52 @@
+from __future__ import annotations
+
+from typing import Any, NotRequired, TypedDict
+
+# Generated by scripts/update_sdk_artifacts.py
+
+class ThreadObject(TypedDict):
```
owenlin0: curious, why do we need a `*Object` suffix?

sdk/python/src/codex_app_server/generated/schema_types.py
```diff
@@ -0,0 +1,188 @@
...
+    name: Any
+    path: Any
+    preview: str
+    source: Any
+    status: Any
+    turns: list[Turn]
+    updatedAt: int
+
+@dataclass(slots=True, kw_only=True)
+class Thread:
```
owenlin0: What's the difference between `ThreadDict` vs. `Thread` (defined here) vs. `ThreadObject` (defined in protocol_types.py)? Do we need all those types? Can we simplify?

sdk/python/src/codex_app_server/generated/v2_all/AccountLoginCompletedNotification.py
```diff
@@ -0,0 +1,15 @@
...
+from __future__ import annotations
+
+from typing import Optional
+
+from pydantic import BaseModel
+
+
+class AccountLoginCompletedNotification(BaseModel):
+    error: Optional[str] = None
+    loginId: Optional[str] = None
```
owenlin0: nit: I think `str | None = None` is more idiomatic these days for optional types

sdk/python/src/codex_app_server/generated/v2_all/AccountUpdatedNotification.py
```diff
@@ -0,0 +1,45 @@
...
+
+class AuthMode2(Enum):
+    chatgpt = "chatgpt"
+
+
+class AuthMode3(Enum):
+    chatgptAuthTokens = "chatgptAuthTokens"
+
+
+class AuthMode(RootModel[Union[AuthMode1, AuthMode2, AuthMode3]]):
```
owenlin0: Curious why `AuthMode` is generated this way vs. `PlanType` below (which seems better and is less verbose)

sdk/python/src/codex_app_server/generated/v2_all/AccountRateLimitsUpdatedNotification.py
```diff
@@ -0,0 +1,47 @@
...
+from pydantic import BaseModel
+
+
+class CreditsSnapshot(BaseModel):
+    balance: Optional[str] = None
+    hasCredits: bool
+    unlimited: bool
+
+
+class PlanType(Enum):
```
owenlin0: Seems like shared enums are not generated once and imported (i.e. we define `PlanType` multiple times across various files) - possible to fix?

owenlin0 (commented):
Can we run `datamodel_code_generator` against the json schema bundle at `codex-rs/app-server-protocol/schema/json/codex_app_server_protocol.v2.schemas.json` instead of over the `codex-rs/app-server-protocol/schema/json/v2/` directory? I think that should fix duplicate definitions (e.g. `PlanType`).

Also, I asked Codex to walk me through the app-server schema -> python pipeline, and it seems like we have these stages:
1. **Source To Raw Models**: app-server json schema -> pydantic types
2. **Raw Models To Curated Python Shapes**: a subset of manually curated types which outputs both TypedDict and Pydantic types
3. **Raw Models To Public API Signatures**: generated Pydantic model -> ergonomic Python method signature

Can we simplify? Codex suggests this:

> The simplest functional shape is:
> 
> - one canonical generated wire-model layer
> - Pydantic only
> - no handwritten “curated schema shapes”
> - a thin handwritten client/public API on top
> That is the direction I’d recommend.
> 
> Use the bundled schema `codex_app_server_protocol.v2.schemas.json` as the canonical input, and generate a single Pydantic model namespace from it.
>
> The Python-facing models should use snake_case field names, but serialize and validate against camelCase on the wire.

I think that should allow us to delete `protocol_types.py` and `schema_types.py`

sdk/python/src/codex_app_server/client.py
```diff
@@ -0,0 +1,605 @@
...
+        params: V2TurnStartParams | JsonObject | None = None,
+    ) -> TurnStartResponse:
+        payload = {
+            **_params_dict(params),
+            "threadId": thread_id,
+            "input": self._normalize_input_items(input_items),
+        }
+        return self.request("turn/start", payload, response_model=TurnStartResponse)
+
+    def turn_text(
```
owenlin0: nit: `turn_text` seems unnecessary?

sdk/python/src/codex_app_server/client.py
```diff
@@ -0,0 +1,605 @@
...
+        thread_id: str | None = None,
+    ) -> TextTurnResult:
+        active_thread_id = thread_id
+        if active_thread_id is None:
+            start_params = V2ThreadStartParams(model=model) if model else None
+            started = self.thread_start(start_params)
+            active_thread_id = started.thread.id
+        return self.run_text_turn(active_thread_id, text)
+
+    def ask(
```
owenlin0: do we need `ask` and `ask_result`? These seem a bit weird

sdk/python/src/codex_app_server/client.py
```diff
@@ -0,0 +1,605 @@
...
+    def stream_until_methods(self, methods: Iterable[str] | str) -> list[Notification]:
+        target_methods = {methods} if isinstance(methods, str) else set(methods)
+        out: list[Notification] = []
+        while True:
+            notification = self.next_notification()
+            out.append(notification)
+            if notification.method in target_methods:
+                return out
+
+    def run_text_turn(
```
owenlin0: what's the use case of `run_text_turn`? do we need it?

sdk/python/src/codex_app_server/models.py
```diff
@@ -0,0 +1,107 @@
...
+class InitializeResponse(BaseModel):
+    serverInfo: ServerInfo | None = None
+    userAgent: str | None = None
+
+
+@dataclass(slots=True)
+class TextTurnResult:
+    thread_id: str
+    turn_id: str
+    deltas: list[AgentMessageDeltaNotification]
```
owenlin0: similarly, do we need `TextTurnResult`? wondering if we can remove.

but if you really do want it, instead of storing `deltas` could we store a list of agent messages instead? it'd feel a little less weird. (but still unclear of the use case of this)

sdk/python/tests/test_client_rpc_methods.py
```diff
@@ -0,0 +1,112 @@
...
+
+    assert event.method == "thread/tokenUsage/updated"
+    assert isinstance(event.payload, ThreadTokenUsageUpdatedNotification)
+    assert event.payload.turn_id == "turn-1"
+
+
+def test_codex_event_notifications_are_typed() -> None:
+    client = AppServerClient()
+    event = client._coerce_notification(
+        "codex/event/turn_aborted",
```
owenlin0: FYI we are dropping `codex/event/*` from app-server API very soon. No need to invest in any support for these

owenlin0 (approved):
much cleaner, thanks! just some non-blocking comments

PR #13978, opened 2026-03-08
enh: add feature-flagged server-side compaction

cooper-oai:
## summary
- add a default-off `server_side_compaction` flag for openai responses auto-compaction
- send `context_management` on eligible `/responses` requests instead of auto-calling the legacy `/compact` path
- apply streamed compaction checkpoints locally so follow-up turns reuse the inline summary without losing current-turn state
- keep manual compaction and non-openai providers on the existing client-side flow

## implementation notes
- this only changes auto-compaction on openai-compatible responses requests
- `compact_prompt` is ignored for inline auto-compaction. this is not a regression: codex already ignores it for openai auto-compaction today, and manual `/compact` still uses the existing prompt-based endpoint
- codex buffers streamed compaction items and rewrites history on `response.completed` so later streamed items keep their order and retries do not commit partial checkpoints
- when inline server-side compaction is available, we skip previous-model preflight compaction so openai auto-compaction uses a single path

## rollout
- land behind `server_side_compaction`, default off
- validate first against internal openai responses upstreams and app-server flows
- watch `codex.compaction` and `codex.compaction_checkpoint_applied`
- if stable, enable for a small internal cohort; manual compaction and non-openai providers continue on the existing flow

## testing
- ran unit tests
- tested with the local binary pointed at an openai responses api upstream


cooper-oai:
Known fast-follow: while testing this branch against a Responses API upstream configured via `env_key`, I hit a pre-existing request-compression/auth-gating bug that already exists on `main`.

Today, request compression is enabled based on cached ChatGPT auth state plus `provider.is_openai()`, while the actual outgoing bearer token can come from `provider.env_key`. On some Responses API upstreams that reject compressed `/v1/responses` bodies, that leads to `invalid_json` unless `--disable enable_request_compression` is set.

This is not introduced by the server-side compaction change; this branch just makes the issue easier to hit when validating against non-default OpenAI-shaped upstreams. I’m treating compression-capability/auth-source alignment as a fast-follow rather than part of this PR.

codex-rs/core/src/client.rs
```diff
@@ -533,6 +534,17 @@ impl ModelClientSession {
         };
         let text = create_text_param_for_request(verbosity, &prompt.output_schema);
         let prompt_cache_key = Some(self.client.state.conversation_id.to_string());
+        let context_management = prompt
+            .inline_compaction_threshold
+            .map(ApiContextManagement::compaction)
+            .map(|entry| vec![entry]);
+        if let Some(threshold) = prompt.inline_compaction_threshold {
+            trace!(
```
pakrym-oai: nit: the entire request is traced.

codex-rs/core/src/stream_events_utils.rs
```diff
@@ -163,6 +173,32 @@ pub(crate) async fn handle_output_item_done(
     let mut output = OutputItemResult::default();
     let plan_mode = ctx.turn_context.collaboration_mode.mode == ModeKind::Plan;
 
+    if matches!(item, ResponseItem::Compaction { .. }) {
```
pakrym-oai: Do we get no replacement history?

cooper-oai: We do, just not at `output_item.done`.

At this point I’m only buffering the streamed compaction item plus history_at_checkpoint. On `response.completed` we call `apply_server_side_compaction_checkpoint`, which rebuilds replacement_history from that checkpoint snapshot plus any later same-turn items that arrived after the compaction event, and then commits it via `replace_compacted_history`.

So replacement history is preserved; it’s just derived later so the rewrite stays atomic and includes the full turn.

codex-rs/core/tests/suite/compact_remote.rs
```diff
@@ -362,6 +379,304 @@ async fn remote_compact_runs_automatically() -> Result<()> {
     Ok(())
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn auto_server_side_compaction_uses_inline_context_management() -> Result<()> {
```
pakrym-oai, line 442: can we add a snapshot test that illustrates the context after a server-side compaction?

codex-rs/core/src/codex.rs
```diff
@@ -5519,10 +5618,159 @@ pub(crate) async fn run_turn(
...
+    turn_start_context_items: &[ResponseItem],
+    history_before_turn: &[ResponseItem],
+    history_at_checkpoint: &[ResponseItem],
+    current_history: &[ResponseItem],
+) -> Vec<ResponseItem> {
+    // Rebuild the active turn around the compaction checkpoint:
+    // 1. keep the turn-local items that existed when compaction fired
+    // 2. replace any prior same-turn compaction summary with the newest one
+    // 3. re-append items that arrived later in the same streamed response
+    // 4. reattach ghost snapshots at the end so undo state survives the rewrite
```
pakrym-oai: I'd fully ignore ghost snapshots.

codex-rs/core/src/codex.rs
```diff
@@ -5553,6 +5821,13 @@ async fn maybe_run_previous_model_inline_compact(
     turn_context: &Arc<TurnContext>,
     total_usage_tokens: i64,
 ) -> CodexResult<bool> {
+    // Keep OpenAI auto-compaction on one path. If inline server-side compaction is eligible for
+    // the current turn, let the normal pre-turn inline request handle it instead of running the
+    // older previous-model client-side preflight flow first.
+    if inline_server_side_compaction_threshold(sess, turn_context).is_some() {
```
pakrym-oai: Are you sure this is correct?

codex-rs/core/src/codex.rs
```diff
@@ -5519,10 +5618,159 @@ pub(crate) async fn run_turn(
...
+    }
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+struct PendingServerSideCompaction {
+    threshold: i64,
+    trigger: AutoCompactTrigger,
+}
+
+fn build_server_side_compaction_replacement_history(
```
pakrym-oai: This method looks very complicated. In theory all we need to do on compaction is to replace history with compaction item.

codex-rs/core/src/codex.rs
```diff
@@ -5372,13 +5423,18 @@ pub(crate) async fn run_turn(
             .map(|user_message| user_message.message())
             .collect::<Vec<String>>();
         let turn_metadata_header = turn_context.turn_metadata_state.current_header_value();
+        let inline_compaction_for_request = pending_server_side_compaction;
```
pakrym-oai: Why do we need to determine pending? Can't we always send a value?

codex-rs/core/src/codex.rs
cooper-oai: discussed offline - we will keep this as this is the current behavior for auto-compaction https://github.com/openai/codex/blob/9386182e6fe8734d32952f392e1f0803313fb470/codex-rs/core/src/compact_remote.rs#L87-L93

codex-rs/core/src/stream_events_utils.rs
cooper-oai: discussed offline - we will handle the compaction item as soon as we receive it, acknowledging that this means retries on failed streams may operate on mutated history if the failed stream sent a compaction item before it failed

codex-rs/core/src/codex.rs
cooper-oai: no, we should fall back to manual compaction here. thanks for catching this

codex-rs/core/src/codex.rs
cooper-oai: simplified

codex-rs/core/src/codex.rs
cooper-oai: good point. since the server will only compact as-needed, it's fine to send this on every request to OAI upstreams

codex-rs/core/src/codex.rs
```diff
@@ -5661,27 +5729,139 @@ pub(crate) async fn run_turn(
...
+    fn as_str(self) -> &'static str {
+        match self {
+            AutoCompactTrigger::AutoPreTurn => "auto_preturn",
+            AutoCompactTrigger::AutoFollowUp => "auto_followup",
+            AutoCompactTrigger::PreviousModelPreflight => "previous_model_preflight",
+        }
+    }
+}
+
+fn build_server_side_compaction_replacement_history(
```
pakrym-oai: can we inline?

codex-rs/core/src/codex.rs
```diff
@@ -3452,7 +3487,7 @@ impl Session {
     pub(crate) async fn record_context_updates_and_set_reference_context_item(
         &self,
         turn_context: &TurnContext,
-    ) {
+    ) -> Vec<ResponseItem> {
```
pakrym-oai: nit: revert?

codex-rs/core/src/codex.rs
```diff
@@ -5434,8 +5467,12 @@ pub(crate) async fn run_turn(
 
     let initial_input_for_turn: ResponseInputItem = ResponseInputItem::from(input.clone());
     let response_item: ResponseItem = initial_input_for_turn.clone().into();
-    sess.record_user_prompt_and_emit_turn_item(turn_context.as_ref(), &input, response_item)
-        .await;
+    sess.record_user_prompt_and_emit_turn_item(
+        turn_context.as_ref(),
+        &input,
+        response_item.clone(),
```
pakrym-oai: nit: revert?

codex-rs/core/src/codex.rs
```diff
@@ -5529,9 +5567,22 @@ pub(crate) async fn run_turn(
         .await
         {
             Ok(sampling_request_output) => {
+                if inline_server_side_compaction_threshold(&sess, &turn_context).is_some() {
+                    let result = if sampling_request_output.observed_server_side_compaction {
+                        "applied"
+                    } else {
+                        "skipped"
+                    };
+                    sess.services.session_telemetry.counter(
```
pakrym-oai: I think we should keep codex.inline_compaction to when it actually happens.

codex-rs/core/src/codex.rs
```diff
@@ -5551,7 +5602,8 @@ pub(crate) async fn run_turn(
 
                 // as long as compaction works well in getting us way below the token limit, we shouldn't worry about being in an infinite loop.
                 if token_limit_reached && needs_follow_up {
-                    if run_auto_compact(
+                    if inline_server_side_compaction_threshold(&sess, &turn_context).is_some() {
```
pakrym-oai: hm?

codex-rs/core/src/codex.rs
```diff
@@ -5661,24 +5713,45 @@ pub(crate) async fn run_turn(
...
+    turn_context: &TurnContext,
+) -> Option<i64> {
+    if !sess.enabled(Feature::ServerSideCompaction) {
+        return None;
+    }
+    if !should_use_remote_compact_task(&turn_context.provider) {
+        return None;
+    }
+    // OpenAI inline auto-compaction uses Responses `context_management`, which has no
+    // compaction-prompt field. Auto-compaction therefore ignores `compact_prompt`, while manual
```
pakrym-oai: Is this a useful comment?

codex-rs/core/src/codex.rs
```diff
@@ -5661,24 +5713,45 @@ pub(crate) async fn run_turn(
...
 ) -> CodexResult<()> {
     let total_usage_tokens_before_compaction = sess.get_total_token_usage().await;
-    maybe_run_previous_model_inline_compact(
+    if maybe_run_previous_model_inline_compact(
         sess,
         turn_context,
         total_usage_tokens_before_compaction,
     )
-    .await?;
+    .await?
```
pakrym-oai: did this need to change?

codex-rs/core/src/codex.rs
```diff
@@ -5529,9 +5567,22 @@ pub(crate) async fn run_turn(
         .await
         {
             Ok(sampling_request_output) => {
+                if inline_server_side_compaction_threshold(&sess, &turn_context).is_some() {
+                    let result = if sampling_request_output.observed_server_side_compaction {
```
pakrym-oai: observed_server_side_compaction is not worth carrying around just for this

codex-rs/core/src/codex.rs
```diff
@@ -6809,6 +6889,16 @@ async fn try_run_sampling_request(
                 let output_result = handle_output_item_done(&mut ctx, item, previously_active_item)
                     .instrument(handle_responses)
                     .await?;
+                observed_server_side_compaction |= saw_server_side_compaction;
+                if let Some(ServerSideCompaction { item, turn_item }) =
+                    output_result.server_side_compaction
+                {
+                    sess.apply_server_side_compaction(turn_context.as_ref(), item)
+                        .await;
+                    sess.emit_turn_item_started(&turn_context, &turn_item).await;
```
pakrym-oai: what is the turn_item in this case?

codex-rs/core/src/stream_events_utils.rs
```diff
@@ -163,6 +172,27 @@ pub(crate) async fn handle_output_item_done(
...
+        let turn_item = TurnItem::ContextCompaction(match previously_active_item {
```
pakrym-oai: why don't we create the item right where we emit it?

codex-rs/core/src/stream_events_utils.rs
```diff
@@ -163,6 +172,27 @@ pub(crate) async fn handle_output_item_done(
     let mut output = OutputItemResult::default();
     let plan_mode = ctx.turn_context.collaboration_mode.mode == ModeKind::Plan;
 
+    if matches!(item, ResponseItem::Compaction { .. }) {
```
pakrym-oai: can we handle compaction even before dispatching into handle_output_item_done ?

codex-rs/core/src/stream_events_utils.rs
```diff
@@ -320,6 +350,9 @@ pub(crate) async fn handle_non_tool_response_item(
             }
             Some(turn_item)
         }
+        ResponseItem::Compaction { .. } => {
+            Some(TurnItem::ContextCompaction(ContextCompactionItem::new()))
```
pakrym-oai: surprised we also create the item here.

pakrym-oai (changes requested):
Few more comments on the core compaction part.

codex-rs/core/src/codex.rs
```diff
@@ -3177,6 +3179,40 @@ impl Session {
...
+                .filter(|history_item| matches!(history_item, ResponseItem::GhostSnapshot { .. }))
+                .cloned(),
+        );
+
+        let reference_context_item = Some(turn_context.to_turn_context_item());
+        let compacted_item = CompactedItem {
+            message: String::new(),
+            replacement_history: Some(replacement_history.clone()),
+        };
+        self.replace_compacted_history(replacement_history, reference_context_item, compacted_item)
```
charley-oai, line 3258: Important: `process_compacted_history` reinjects turn context above the last user message before the compaction item (for mid-turn compaction), and filters the replacement history by `should_keep_compacted_history_item`; I don't see us doing that here?

How would we handle that logic for server side compaction?

Can server side compaction run pre-turn (after user message but before the first assistant message), or does it always run mid-turn?

codex-rs/core/src/codex.rs
cooper-oai: updated this to explicitly check for enablement and compute the compaction threshold at serialization time

codex-rs/core/src/stream_events_utils.rs
cooper-oai: updated to do that

codex-rs/core/src/codex.rs
cooper-oai: no not strictly needed, the downstream guard against the re-check is redundant i think

codex-rs/core/src/codex.rs
cooper-oai: removed

codex-rs/core/src/codex.rs
cooper-oai: removed

PR #14009, opened 2026-03-08
feat(core) Persist request_permission data across turns

dylan-hurd-oai:
## Summary
request_permissions flows should support persisting results for the session.

Open Question: Still deciding if we need within-turn approvals - this adds complexity but I could see it being useful

## Testing
- [x] Updated unit tests


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -83,12 +84,18 @@ use ts_rs::TS;
 // tends to use either snake_case or kebab-case.
 macro_rules! v2_enum_from_core {
     (
-        pub enum $Name:ident from $Src:path { $( $Variant:ident ),+ $(,)? }
+        $(#[$enum_meta:meta])*
```
celia-oai, line 87: what is this change for?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -83,12 +84,18 @@ use ts_rs::TS;
...
+            $( $(#[$variant_meta:meta])* $Variant:ident ),+ $(,)?
+        }
     ) => {
         #[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, JsonSchema, TS)]
+        $(#[$enum_meta])*
         #[serde(rename_all = "camelCase")]
         #[ts(export_to = "v2/")]
-        pub enum $Name { $( $Variant ),+ }
+        pub enum $Name {
+            $( $(#[$variant_meta])* $Variant ),+
```
celia-oai, line 97: same here, what is this for?

celia-oai (approved):

codex-rs/app-server-protocol/src/protocol/v2.rs
dylan-hurd-oai: This allows us to add `#[derive(Default)]` to the enum declaration. We could alternatively keep the macro simple, and do a manual `impl` for Default.

codex-rs/app-server-protocol/src/protocol/v2.rs
dylan-hurd-oai: Alongside above, allows you to declare `#[default]` on a specific variant

PR #14011, opened 2026-03-08
[apps] Fix apps enablement condition.

mzeng-openai:
- [x] Fix apps enablement condition to check both the feature flag and that the user is not an API key user.


codex-rs/tui/src/chatwidget.rs
```diff
@@ -3225,7 +3225,7 @@ impl ChatWidget {
 
         widget
             .bottom_pane
-            .set_connectors_enabled(widget.config.features.enabled(Feature::Apps));
+            .set_connectors_enabled(widget.connectors_enabled());
```
xl-openai, line 3289: Do we want to rename connector to app in TUI as well?

codex-rs/core/src/features.rs
```diff
@@ -251,6 +252,10 @@ impl Features {
         self.enabled.contains(&f)
     }
 
+    pub fn apps_enabled(&self, auth: Option<&CodexAuth>) -> bool {
+        self.enabled(Feature::Apps) && !auth.is_some_and(CodexAuth::is_api_key_auth)
```
xl-openai: should we just explicitly check is chatgpt auth?

codex-rs/core/src/features.rs
```diff
@@ -251,6 +252,10 @@ impl Features {
         self.enabled.contains(&f)
     }
 
+    pub fn apps_enabled(&self, auth: Option<&CodexAuth>) -> bool {
```
xl-openai: maybe just take an auth manager directly so caller doesn't need to do auth().await on its own?

canvrno-oai (approved):

codex-rs/core/src/tools/spec.rs
```diff
@@ -1912,9 +1917,8 @@ pub(crate) fn build_specs(
         builder.register_handler("request_user_input", request_user_input_handler);
     }
 
-    if config.search_tool
-        && let Some(app_tools) = app_tools
-    {
+    if config.search_tool {
+        let app_tools = app_tools.unwrap_or_default();
```
sayan-oai, line 1948: i think defaulting to empty on missing `app_tools` means API-auth sessions still have access to `search_tool_bm25` but will always return an empty tool list. we should disable the tool for API-auth users?

sayan-oai (commented):
additional nit: i think we should update the app-server README.md to include that ChatGPT auth is now required for apps.

PR #14012, opened 2026-03-08
fix(core): update active item logic for interleaved streamed items

dylan-hurd-oai:
## Summary

Fixes a `codex-core` streaming bug where interleaved response items could misroute live events.

The root cause is that `try_run_sampling_request` tracks only a single active streamed item. If a reasoning item and an assistant message overlapped in the same Responses stream, later deltas could be associated with the wrong item, or one item’s completion could clear the other item’s live state early. That could break downstream event propagation in both the TUI and app-server even though the session still produced persisted rollout events.

This change updates core to track active streamed items by id and route assistant vs. reasoning deltas through the correct live item.

I will admit that `ActiveResponseItems` is a little goofy. We should probably start storing & using the index numbers provided by the responses api. But if this is the root cause of the symptoms that I suspect, I'd like to land this sooner than later.

## Tests
Added an integration test in `core/tests/suite/items.rs` covering an interleaved stream. It fails consistently against main, and passes with this fix.


jif-oai:
@codex review

jif-oai (approved):
The code is ok but I must admit I'm not 100% convinced by the design. I don't have a better idea for now though

PR #14107, opened 2026-03-09
fix: keep permissions profiles forward compatible

viyatb-oai:
## Summary
- preserve unknown `:special_path` tokens, including nested entries, so older Codex builds warn and ignore instead of failing config load
- fail closed with a startup warning when a permissions profile has missing or empty filesystem entries instead of aborting profile compilation
- normalize Windows verbatim paths like `\?\C:\...` before absolute-path validation while keeping explicit errors for truly invalid paths

## Testing
- just fmt
- cargo test -p codex-core permissions_profiles_allow
- cargo test -p codex-core normalize_absolute_path_for_platform_simplifies_windows_verbatim_paths
- cargo test -p codex-protocol unknown_special_paths_are_ignored_by_legacy_bridge
- cargo clippy -p codex-core -p codex-protocol --all-targets -- -D warnings
- cargo clean

### DIFF
--- a/codex-rs/protocol/src/permissions.rs
+++ b/codex-rs/protocol/src/permissions.rs
@@ -71,7 +71,8 @@ pub enum FileSystemSpecialPath {
     /// Do not make older runtimes reject newly introduced tokens.
     /// New parser support should be additive, while unknown values must stay
     /// representable so config from a newer Codex degrades to warn-and-ignore
-    /// instead of failing to load.
+    /// instead of failing to load. Codex 0.112.0 rejected unknown values here,
+    /// which broke forward compatibility for newer config.
     /// Preserves future special-path tokens so older runtimes can ignore them
     /// without rejecting config authored by a newer release.
     Unknown {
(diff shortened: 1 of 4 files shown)


codex-rs/protocol/src/permissions.rs
```diff
@@ -67,12 +67,32 @@ pub enum FileSystemSpecialPath {
     },
     Tmpdir,
     SlashTmp,
+    /// WARNING: `:special_path` tokens are part of config compatibility.
+    /// Do not make older runtimes reject newly introduced tokens.
+    /// New parser support should be additive, while unknown values must stay
+    /// representable so config from a newer Codex degrades to warn-and-ignore
+    /// instead of failing to load.
+    /// Preserves future special-path tokens so older runtimes can ignore them
+    /// without rejecting config authored by a newer release.
```
dylan-hurd-oai, line 77: Can we also include a note about 0.112.0 rejecting values?

dylan-hurd-oai (approved):

codex-rs/protocol/src/permissions.rs
viyatb-oai: added!

PR #14125, opened 2026-03-09
Mark incomplete resumed turns interrupted when idle

guinness-oai:
Fixes a Codex app bug where quitting the app mid-run could leave the reopened thread stuck in progress and non-interactable. On cold thread resume, app-server could return an idle thread with a replayed turn still marked in progress. This marks incomplete replayed turns as interrupted unless the thread is actually active.

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -6723,6 +6723,7 @@ impl CodexMessageProcessor {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
 async fn handle_thread_listener_command(
     conversation_id: ThreadId,
     conversation: &Arc<CodexThread>,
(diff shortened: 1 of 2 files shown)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3245,12 +3245,20 @@ impl CodexMessageProcessor {
...
                         .loaded_status_for_thread(&thread.id)
                         .await,
                     false,
                 );
+                if !matches!(status, ThreadStatus::Active { .. }) {
+                    for turn in &mut thread.turns {
+                        if matches!(turn.status, TurnStatus::InProgress) {
+                            turn.status = TurnStatus::Interrupted;
+                        }
+                    }
```
euroelessar: I think this check/rewrite needs to be done before `resume_thread_with_history` call above.
We do allow multiple clients to connect to the same app server, and they should observe the same thread status/etc, but with this change only the first connected client will see interrupted turns, the rest will be stuck in limbo as before this change.

euroelessar (approved):

PR #14139, opened 2026-03-09
windows-sandbox: add runner IPC foundation for future unified_exec

iceweasel-oai:
# Summary

This PR introduces the Windows sandbox runner IPC foundation that later unified_exec work will build on.

The key point is that this is intentionally infrastructure-only. The new IPC transport, runner plumbing, and ConPTY helpers are added here, but the active elevated Windows sandbox path still uses the existing request-file bootstrap. In other words, this change prepares the transport and module layout we need for unified_exec without switching production behavior over yet.

Part of this PR is also a source-layout cleanup: some Windows sandbox files are moved into more explicit `elevated/`, `conpty/`, and shared locations so it is clearer which code is for the elevated sandbox flow, which code is legacy/direct-spawn behavior, and which helpers are shared between them. That reorganization is intentional in this first PR so later behavioral changes do not also have to carry a large amount of file-move churn.

# Why This Is Needed For unified_exec

Windows elevated sandboxed unified_exec needs a long-lived, bidirectional control channel between the CLI and a helper process running under the sandbox user. That channel has to support:

- starting a process and reporting structured spawn success/failure
- streaming stdout/stderr back incrementally
- forwarding stdin over time
- terminating or polling a long-lived process
- supporting both pipe-backed and PTY-backed sessions

The existing elevated one-shot path is built around a request-file bootstrap and does not provide those primitives cleanly. Before we can turn on Windows sandbox unified_exec, we need the underlying runner protocol and transport layer that can carry those lifecycle events and streams.

# Why Windows Needs More Machinery Than Linux Or macOS

Linux and macOS can generally build unified_exec on top of the existing sandbox/process model: the parent can spawn the child directly, retain normal ownership of stdio or PTY handles, and manage the lifetime of the sandboxed process without introducing a second control process.

Windows elevated sandboxing is different. To run inside the sandbox boundary, we cross into a different user/security context and then need to manage a long-lived process from outside that boundary. That means we need an explicit helper process plus an IPC transport to carry spawn, stdin, output, and exit events back and forth. The extra code here is mostly that missing Windows sandbox infrastructure, not a conceptual difference in unified_exec itself.

# What This PR Adds

- the framed IPC message types and transport helpers for parent <-> runner communication
- the renamed Windows command runner with both the existing request-file bootstrap and the dormant IPC bootstrap
- named-pipe helpers for the elevated runner path
- ConPTY helpers and process-thread attribute plumbing needed for PTY-backed sessions
- shared sandbox/process helpers that later PRs will reuse when switching live execution paths over
- early file/module moves so later PRs can focus on behavior rather than layout churn

# What This PR Does Not Yet Do

- it does not switch the active elevated one-shot path over to IPC yet
- it does not enable Windows sandbox unified_exec yet
- it does not remove the existing request-file bootstrap yet

So while this code compiles and the new path has basic validation, it is not yet the exercised production path. That is intentional for this first PR: the goal here is to land the transport and runner foundation cleanly before later PRs start routing real command execution through it.

# Follow-Ups

Planned follow-up PRs will:

1. switch elevated one-shot Windows sandbox execution to the new runner IPC path
2. layer Windows sandbox unified_exec sessions on top of the same transport
3. remove the legacy request-file path once the IPC-based path is live

# Validation

- `cargo build -p codex-windows-sandbox`


troycoco9-sys (commented):
Ok

codex-rs/windows-sandbox-rs/src/elevated/command_runner_win.rs
```diff
@@ -0,0 +1,739 @@
...
+        anyhow::bail!("runner: unsupported protocol version {}", msg.version);
+    }
+    match msg.message {
+        Message::SpawnRequest { payload } => Ok(*payload),
+        other => anyhow::bail!("runner: expected spawn_request, got {other:?}"),
+    }
+}
+
+/// Pick an effective CWD, using a junction if the ACL helper is active.
+fn effective_cwd(req_cwd: &Path, log_dir: Option<&Path>) -> PathBuf {
```
dylan-hurd-oai, line 188: am i misremembering or do we already have logic for this somewhere?

codex-rs/windows-sandbox-rs/src/elevated/ipc_framed.rs
```diff
@@ -0,0 +1,181 @@
+//! Framed IPC protocol used between the parent (CLI) and the elevated command runner.
+//!
+//! This module defines the JSON message schema (spawn request/ready, output, stdin,
+//! exit, error, terminate) plus length‑prefixed framing helpers for a byte stream.
+//! It is **elevated-path only**: the parent uses it to bootstrap the runner and
+//! stream unified_exec I/O over named pipes. The legacy restricted‑token path does
+//! not use this protocol, and non‑unified exec capture uses it only when running
+//! through the elevated runner.
```
dylan-hurd-oai, line 8: Parsing this out - does this mean unified_exec can only run inside the elevated sandbox? Do we / will we validate this elsewhere?

codex-rs/windows-sandbox-rs/src/elevated/runner_pipe.rs
```diff
@@ -0,0 +1,106 @@
...
+/// Generates a unique named-pipe path used to communicate with the runner process.
+pub fn pipe_pair() -> (String, String) {
+    let mut rng = SmallRng::from_entropy();
+    let base = format!(r"\\.\pipe\codex-runner-{:x}", rng.gen::<u128>());
+    (format!("{base}-in"), format!("{base}-out"))
+}
+
+/// Creates a named pipe with permissive ACLs so the sandbox user can connect.
+pub fn create_named_pipe(name: &str, access: u32) -> io::Result<HANDLE> {
+    // Allow sandbox users to connect by granting Everyone full access on the pipe.
```
dylan-hurd-oai: I'm not sure about the security implications of this - wouldn't this imply other processes could use this pipe?

dylan-hurd-oai (commented):
Couple comments, mostly for my understanding. Overall this approach feels reasonable!

codex-rs/windows-sandbox-rs/src/elevated/command_runner_win.rs
iceweasel-oai: you are remembering correctly. Part of this PR is moving code around so it's clear what sandbox code is for elevated vs legacy vs shared. `command_runner_win.rs` got moved into `/elevated/` but also changed enough such that github didn't recognize it as a move. So there is some stuff here that was pre-existing, like this

dylan-hurd-oai: ack, ty! i skimmed command_runner_win but missed it

codex-rs/windows-sandbox-rs/src/elevated/ipc_framed.rs
iceweasel-oai: no, it just means that the framed IPC approach is only used in the elevated path because we have the middleman of the command runner.exe. I can make this comment a bit clearer

codex-rs/windows-sandbox-rs/src/elevated/runner_pipe.rs
iceweasel-oai: yeah I think it does mean that. I think I can tighten this up to only allow the real user and the sandbox user to access it.

codex-rs/tui_app_server/src/multi_agents.rs
```diff
@@ -537,10 +537,13 @@ fn status_summary_line(status: &AgentStatus) -> Line<'static> {
     status_summary_spans(status).into()
 }
 
+// Allow `.yellow()`
+#[allow(clippy::disallowed_methods)]
 fn status_summary_spans(status: &AgentStatus) -> Vec<Span<'static>> {
     match status {
         AgentStatus::PendingInit => vec![Span::from("Pending init").cyan()],
         AgentStatus::Running => vec![Span::from("Running").cyan().bold()],
+        AgentStatus::Interrupted => vec![Span::from("Interrupted").yellow()],
```
dylan-hurd-oai, line 546: We use red for interruptions elsewhere - why yellow here?

dylan-hurd-oai (approved):
one comment on tui_app_server change but otherwise LGTM

codex-rs/tui_app_server/src/multi_agents.rs
iceweasel-oai: this shouldn't be here. I'm removing and re-pushing

PR #14149, opened 2026-03-10
unifying all image saves to /tmp to bug-proof

won-openai:
image-gen feature will have the model saving to /tmp by default + at all times

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -6636,7 +6636,8 @@ async fn handle_assistant_item_done_in_plan_mode(
     {
         maybe_complete_plan_item_from_message(sess, turn_context, state, item).await;
 
-        if let Some(turn_item) = handle_non_tool_response_item(item, true).await {
+        if let Some(turn_item) = handle_non_tool_response_item(sess, turn_context, item, true).await
+        {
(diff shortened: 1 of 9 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -3327,6 +3328,14 @@ impl Session {
             )
             .into_text(),
         );
+        if turn_context.features.enabled(Feature::ImageGeneration) {
+            let image_output_dir = default_image_generation_output_dir();
+            developer_sections.push(format!(
```
aibrahim-oai: we should only do that if we have image.

codex-rs/core/src/stream_events_utils.rs
```diff
@@ -212,6 +212,18 @@ pub(crate) async fn handle_output_item_done(
 
             record_completed_response_item(ctx.sess.as_ref(), ctx.turn_context.as_ref(), &item)
                 .await;
+            if record_default_image_save_developer_message {
```
aibrahim-oai: let's pass session in `handle_non_tool_response_item` and send the msg on the spot so we avoid extra complexity.

aibrahim-oai (approved):

PR #14155, opened 2026-03-10
feat: Add additional macOS Sandbox Permissions for Launch Services, Contacts, Reminders

leoshimo-oai:
Add additional macOS Sandbox Permissions levers for the following:

- Launch Services
- Contacts
- Reminders


codex-rs/core/src/restricted_read_only_platform_defaults.sbpl
```diff
@@ -27,6 +27,19 @@
   (subpath "/System/iOSSupport/System/Library/SubFrameworks")
   (subpath "/usr/lib"))
 
+; System Framework and AppKit resources
```
celia-oai, line 30: we shouldn't be changing this default here?

celia-oai: @viyatb-oai could you confirm

codex-rs/protocol/src/models.rs
```diff
@@ -174,10 +196,13 @@ pub struct MacOsSeatbeltProfileExtensions {
     pub macos_preferences: MacOsPreferencesPermission,
     #[serde(alias = "automations")]
     pub macos_automation: MacOsAutomationPermission,
+    pub macos_launch_services: bool,
```
celia-oai, line 200: should we also add alias to these ones?

leoshimo-oai: Good call - updating

viyatb-oai (approved):

codex-rs/core/src/restricted_read_only_platform_defaults.sbpl
leoshimo-oai: Updated to `restricted_read_only_platform_defaults.sbpl` per off-thread conversation

PR #14157, opened 2026-03-10
Enforce single tool output type in codex handlers

pakrym-oai:
We'll need to associate output schema with each tool. Each tool can only have on output type.


codex-rs/core/src/tools/handlers/mcp.rs
```diff
@@ -3,25 +3,53 @@ use std::sync::Arc;
...
 use crate::tools::context::ToolInvocation;
-use crate::tools::context::ToolOutputBox;
 use crate::tools::context::ToolPayload;
 use crate::tools::registry::ToolHandler;
 use crate::tools::registry::ToolKind;
 use codex_protocol::models::ResponseInputItem;
 
 pub struct McpHandler;
 
+pub enum McpHandlerOutput {
```
pakrym-oai, line 16: This can be simplified and I'll do it as a followup.

codex-rs/core/src/tools/registry.rs
```diff
@@ -52,19 +54,68 @@ pub trait ToolHandler: Send + Sync {
...
+}
+
+struct AnyToolResult {
+    preview: String,
+    success: bool,
+    response: ResponseInputItem,
+}
+
+#[async_trait]
+trait AnyToolHandler: Send + Sync {
```
pakrym-oai, line 67: This is so registry can store non-generic handlers.

aibrahim-oai (approved):

PR #14160, opened 2026-03-10
Add spawn_agent model overrides

aibrahim-oai:
- add `model` and `reasoning_effort` to the `spawn_agent` schema so the values pass through
- validate requested models against `model.model` and only check that the selected model supports the requested reasoning effort


pakrym-oai (approved):
The product change looks reasonable but the test change seems excessively custom. Can we reuse existing sub agent tests?

codex-rs/core/src/tools/handlers/multi_agents.rs
```diff
@@ -963,6 +976,106 @@ fn apply_spawn_agent_overrides(config: &mut Config, child_depth: i32) {
     }
 }
 
+async fn apply_requested_spawn_agent_model_overrides(
```
pakrym-oai, line 979: the fact that all this code needs to be written is a bit sad. our internal agent api is cumbersome

codex-rs/core/src/tools/handlers/multi_agents.rs
```diff
@@ -963,6 +976,106 @@ fn apply_spawn_agent_overrides(config: &mut Config, child_depth: i32) {
...
+    Ok(())
+}
+
+fn find_spawn_agent_model_name(
+    available_models: &[codex_protocol::openai_models::ModelPreset],
+    requested_model: &str,
+) -> Result<String, FunctionCallError> {
+    available_models
+        .iter()
+        .find(|model| model.model == requested_model || model.id == requested_model)
```
pakrym-oai: why are we checking id?

PR #14167, opened 2026-03-10
Export tools module into code mode runner

pakrym-oai:
**Summary**
- allow `code_mode` to pass enabled tools metadata to the runner and expose them via `tools.js`
- import tools inside JavaScript rather than relying only on globals or proxies for nested tool calls
- update specs, docs, and tests to exercise the new bridge and explain the tooling changes

**Testing**
- Not run (not requested)


codex-rs/core/src/tools/code_mode.rs
```diff
@@ -157,7 +161,14 @@ async fn execute_node(
         String::from_utf8_lossy(&buf).trim().to_string()
     });
 
-    write_message(&mut stdin, &HostToNodeMessage::Init { source }).await?;
+    write_message(
```
cconger, line 164: oof, we're starting a new process and initing all the tools for each js invocation :|

Ok for PoC but will be nice to have these be startup snapshots when we get to the isolates.

cconger (approved):

PR #14170, opened 2026-03-10
queue slash commands in tui

charley-oai:
## Summary
- make slash commands entered during a running turn queue instead of failing immediately
- cover queued `/review` behavior and inline slash-command variants in the TUI tests
- update the queued-command snapshot to reflect the new pending state

## Testing
- just fmt
- cargo test -p codex-tui


codex-rs/tui/src/bottom_pane/mod.rs
```diff
@@ -419,25 +410,15 @@ impl BottomPane {
...
-            // If a task is running and a status line is visible, allow Esc to
-            // send an interrupt even while the composer has focus.
-            // When a popup is active, prefer dismissing it over interrupting the task.
+            // If a task is running, allow Esc to send an interrupt when no popup is active.
+            // Final-message streaming can temporarily hide the status widget, but that should not
+            // disable interrupt.
             if key_event.code == KeyCode::Esc
                 && matches!(key_event.kind, KeyEventKind::Press | KeyEventKind::Repeat)
                 && self.is_task_running
-                && !is_agent_command
```
charley-oai, line 437: The old `&& !is_agent_command` guard was a heuristic workaround, not the real correctness condition.

Why it was added:
- In `2a5bcc053f` (`fix: esc in /agent`), `/agent ` was a special case because typing past the command name hid the slash popup, so `popup_active()` was no longer protecting that flow.
- At that point, `Esc` only interrupted in a narrower set of cases, and the guard was used to keep `/agent` draft editing from getting treated like a global interrupt.

Why it is safe to remove now:
- The rule we want now is simpler and more correct: **if a task is running and no popup/modal is active, `Esc` should interrupt**.
- That means the thing we should key off is actual UI state (`popup_active()` / modal stack), not the current text prefix.
- Once the `/agent` popup is gone, there is no longer any agent-specific local UI to dismiss. At that point `/agent ` is just ordinary composer text, and it should behave the same as any other nonempty draft.

Why the text-prefix check was the wrong abstraction:
- It special-cased one command name instead of checking the actual thing that matters: whether the user is inside a local picker/dialog flow.
- It was inconsistent with the newer behavior for other nonempty drafts, where `Esc` is intentionally allowed to interrupt.
- It would have kept `/agent` as a one-off exception even after we explicitly changed the product behavior to allow interrupt with nonempty composer text.

Why this does not lose user state:
- The interrupt path preserves the draft text; it does not clear the composer.
- We also have coverage for the non-destructive behavior higher up in `ChatWidget`, so this is not “drop the command and interrupt”, it is “interrupt, while keeping the draft available”.

So the short version is: removing `!is_agent_command` is safe because popup/modal state is the real guard. `/agent` text by itself is not a meaningful interaction mode once its picker is gone.

codex-rs/tui/src/app.rs
```diff
@@ -2439,6 +2409,10 @@ impl App {
                 self.chat_widget.set_model(&model);
                 self.refresh_status_line();
             }
+            AppEvent::HandleSlashCommandDraft(draft) => {
+                self.chat_widget.handle_serialized_slash_command(draft);
+                self.refresh_status_line();
+            }
```
charley-oai, line 2797: Allows a slash command popup (e.g. model picker popup) to then submit a text slash command that can be queued or submitted

codex-rs/tui/src/chatwidget.rs
```diff
@@ -2078,29 +2070,31 @@ impl ChatWidget {
...
                 self.restore_user_message_to_composer(combined);
             }
-        } else if let Some(combined) = self.drain_pending_messages_for_restore() {
+        } else if let Some(combined) = self.drain_restorable_messages_for_restore() {
             self.restore_user_message_to_composer(combined);
         }
         self.refresh_pending_input_preview();
+        if !started_turn_after_interrupt {
+            self.maybe_send_next_queued_input();
+        }
```
charley-oai, line 2245: - we changed interrupt behavior so queued slash actions were no longer merged back into the composer
- instead, they stayed queued and were replayed after interrupt
- to make that work, we added `drain_queued_inputs_until_blocked()` at the end of `on_interrupted_turn()`

That new replay step created the race:

- in the `submit_pending_steers_after_interrupt` path, `on_interrupted_turn()` calls `submit_user_message(...)`
- but `submit_user_message(...)` does not synchronously mark the task as running
- so the new post-interrupt replay could immediately start draining queued items before the replacement steer turn had actually started

`started_turn_after_interrupt` was added to block exactly that case.

charley-oai:
@codex review this

codex-rs/tui/src/bottom_pane/chat_composer.rs
```diff
@@ -2525,9 +2508,6 @@ impl ChatComposer {
             && let Some(cmd) =
                 slash_commands::find_builtin_command(name, self.builtin_command_flags())
         {
-            if self.reject_slash_command_if_unavailable(cmd) {
-                return Some(InputResult::None);
-            }
```
charley-oai, line 2530: This was `"'/{}' is disabled while a task is in progress.",` which we no longer emit

codex-rs/tui/src/bottom_pane/chat_composer.rs
```diff
@@ -2553,13 +2533,6 @@ impl ChatComposer {
 
         let cmd = slash_commands::find_builtin_command(name, self.builtin_command_flags())?;
 
-        if !cmd.supports_inline_args() {
-            return None;
-        }
```
charley-oai, line 2558: Every command supports inline args now

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

charley-oai:
@codex review this

PR #14172, opened 2026-03-10
fix: support split carveouts in windows restricted-token sandbox

viyatb-oai:
## Summary
- keep legacy Windows restricted-token sandboxing as the supported baseline
- support the split-policy subset that restricted-token can enforce directly today
- support full-disk read, the same writable root set as legacy `WorkspaceWrite`, and extra read-only carveouts under those writable roots via additional deny-write ACLs
- continue to fail closed for unsupported split-only shapes, including explicit unreadable (`none`) carveouts, reopened writable descendants under read-only carveouts, and writable root sets that do not match the legacy workspace roots

## Example
Given a filesystem policy like:

```toml
":root" = "read"
":cwd" = "write"
"./docs" = "read"
```

the restricted-token backend can keep the workspace writable while denying writes under `docs` by layering an extra deny-write carveout on top of the legacy workspace-write roots.

A policy like:

```toml
"/workspace" = "write"
"/workspace/docs" = "read"
"/workspace/docs/tmp" = "write"
```

still fails closed, because the unelevated backend cannot reopen the nested writable descendant safely.

## Stack
-> fix: support split carveouts in windows restricted-token sandbox #14172
fix: support split carveouts in windows elevated sandbox #14568


codex-rs/core/src/exec.rs
```diff
@@ -822,27 +816,40 @@ fn should_use_windows_restricted_token_sandbox(
...
     file_system_sandbox_policy: &FileSystemSandboxPolicy,
     network_sandbox_policy: NetworkSandboxPolicy,
+    sandbox_policy_cwd: &Path,
 ) -> Option<String> {
     if should_use_windows_restricted_token_sandbox(
         sandbox,
         sandbox_policy,
         file_system_sandbox_policy,
-    ) {
+    ) && !file_system_sandbox_policy
```
bolinfest: If `file_system_sandbox_policy
        .needs_direct_runtime_enforcement(network_sandbox_policy, sandbox_policy_cwd)` is cheap maybe we do it once and reuse it on line 840 even though there is a slim chance we don't need this value?

codex-rs/core/src/exec.rs
```diff
@@ -822,27 +816,40 @@ fn should_use_windows_restricted_token_sandbox(
...
+        return Some(
+            "windows sandbox backend cannot enforce split filesystem permissions directly; refusing to run unsandboxed"
+                .to_string(),
+        );
+    }
+
+    Some(format!(
+        "windows sandbox backend cannot enforce file_system={:?}, network={network_sandbox_policy:?}, legacy_policy={sandbox_policy:?}; refusing to run unsandboxed",
+        file_system_sandbox_policy.kind,
+    ))
```
bolinfest: consider

```suggestion
    if file_system_sandbox_policy
        .needs_direct_runtime_enforcement(network_sandbox_policy, sandbox_policy_cwd)
    {
        Some(
            "windows sandbox backend cannot enforce split filesystem permissions directly; refusing to run unsandboxed"
                .to_string(),
        )
    } else {

    Some(format!(
        "windows sandbox backend cannot enforce file_system={:?}, network={network_sandbox_policy:?}, legacy_policy={sandbox_policy:?}; refusing to run unsandboxed",
        file_system_sandbox_policy.kind,
    ))
    }
```

or even do `let reason = if ...; Some(reason)`

codex-rs/core/src/exec_tests.rs
```diff
@@ -225,6 +225,7 @@ fn windows_restricted_token_rejects_network_only_restrictions() {
                 &policy,
                 &file_system_policy,
                 NetworkSandboxPolicy::Restricted,
+                Path::new("/tmp"),
```
bolinfest: Is this OK even though this test also runs on Windows (at least I think it does?)

codex-rs/core/src/exec_tests.rs
```diff
@@ -235,14 +236,15 @@ fn windows_restricted_token_rejects_network_only_restrictions() {
...
-    let file_system_policy = FileSystemSandboxPolicy::restricted(vec![]);
+    let file_system_policy = FileSystemSandboxPolicy::from(&policy);
 
     assert_eq!(
         unsupported_windows_restricted_token_sandbox_reason(
             SandboxType::WindowsRestrictedToken,
             &policy,
             &file_system_policy,
             NetworkSandboxPolicy::Restricted,
+            Path::new("/tmp"),
```
bolinfest: and in the other places here...

codex-rs/core/src/exec_tests.rs
```diff
@@ -310,10 +399,11 @@ async fn kill_child_process_group_kills_grandchildren_on_timeout() -> Result<()>
         "-c".to_string(),
         "sleep 60 & echo $!; sleep 60".to_string(),
     ];
+    let cwd = std::env::current_dir()?;
     let env: HashMap<String, String> = std::env::vars().collect();
     let params = ExecParams {
         command,
-        cwd: std::env::current_dir()?,
+        cwd: cwd.clone(),
```
bolinfest: unnecessary clone?

bolinfest (approved):

codex-rs/core/src/sandboxing/mod.rs
```diff
@@ -74,10 +76,68 @@ pub struct ExecRequest {
     pub sandbox_policy: SandboxPolicy,
     pub file_system_sandbox_policy: FileSystemSandboxPolicy,
     pub network_sandbox_policy: NetworkSandboxPolicy,
+    pub(crate) windows_restricted_token_filesystem_overlay:
+        Option<WindowsRestrictedTokenFilesystemOverlay>,
     pub justification: Option<String>,
     pub arg0: Option<String>,
 }
 
+pub struct ExecRequestArgs {
```
bolinfest: Why are we creating a new type instead of just expanding `ExecRequest`?

PR #14173, opened 2026-03-10
fix: preserve split filesystem semantics in linux sandbox

viyatb-oai:
## Stack

   fix: fail closed for unsupported split windows sandboxing #14172
-> fix: preserve split filesystem semantics in linux sandbox #14173
   fix: align core approvals with split sandbox policies #14171
   refactor: centralize filesystem permissions precedence #14174

## Summary
- Preserve Linux split filesystem carveouts in bubblewrap by applying mount masks in the right order, so narrower rules still win under broader writable roots.
- Preserve unreadable ancestors of writable roots by masking them first and then rebinding the narrower writable descendants.
- Stop rejecting legacy-plus-split Linux configs that are sandbox-equivalent after `cwd` resolution by comparing semantics instead of raw legacy structs.
- Fail closed when callers provide partial split policies, mismatched legacy-plus-split policies, or force `--use-legacy-landlock` for split-only shapes that legacy Landlock cannot enforce.
- Add Linux regressions for overlapping writable, read-only, and denied paths, and document the supported split-policy enforcement path.

## Example
Given a split filesystem policy like:

```toml
[permissions.dev.filesystem]
":root" = "read"
"/code" = "write"
"/code/.git" = "read"
"/code/secrets" = "none"
"/code/secrets/tmp" = "write"
```

this PR makes Linux enforce the intended result under bubblewrap:

- `/code` stays writable
- `/code/.git` stays read-only
- `/code/secrets` stays denied
- `/code/secrets/tmp` can still be reopened as writable if explicitly allowed

Before this, Linux could lose one of those carveouts depending on mount order or legacy-policy fallback. This PR keeps the split-policy semantics intact and rejects configurations that legacy Landlock cannot represent safely.


codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -226,6 +232,17 @@ fn resolve_sandbox_policies(
...
     match (sandbox_policy, split_policies) {
         (Some(sandbox_policy), Some((file_system_sandbox_policy, network_sandbox_policy))) => {
+            let derived_legacy_policy = file_system_sandbox_policy
+                .to_legacy_sandbox_policy(network_sandbox_policy, sandbox_policy_cwd)
+                .unwrap_or_else(|err| {
+                    panic!(
+                        "split sandbox policies require direct runtime enforcement and cannot be paired with legacy sandbox policy: {err}"
+                    )
+                });
+            assert_eq!(
```
celia-oai: is having assert_eq in none-test-setting fine? should we have a better way of exposing error?

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -258,81 +259,68 @@ fn create_filesystem_args(
...
-            // expects a live fd number, so keep the backing file open until we
-            // exec bubblewrap below.
-            args.push("--perms".to_string());
-            args.push("000".to_string());
-            args.push("--ro-bind-data".to_string());
-            args.push(null_fd.clone());
-            args.push(path_to_string(unreadable_root));
-        }
-        preserved_files.push(null_file);
+    let mut rootless_unreadable_roots: Vec<PathBuf> = unreadable_roots
```
celia-oai, line 338: I asked codex to suggest if there could potentially be edge cases not covered. lmk if this makes sense:

>   Handled by 

> the new logic:
> 
>   - Nested write -> read -> write under a writable ancestor. That is the main fix in bwrap.rs. It sorts writable roots by depth, then applies each root’s read-only carveouts and nested denies so narrower children
>     can reopen broader parents.
>   - Same-path conflicts are not left to bwrap.rs. They are resolved upstream in permissions.rs, where precedence is “more specific path wins; for equal specificity none > write > read.” So bwrap.rs mostly receives
>     already-reduced writable/read-only/unreadable sets.
>   - Symlink and missing-path cases are explicitly handled in append_read_only_subpath_args() and append_unreadable_root_args(): if a protected path is a symlink under a writable root, or does not exist yet, it
>     mounts /dev/null on the symlink or first missing component.
> 
>   The edge case I would worry about is:
> 
>   - A broader none parent with a narrower write child, when there is no broader writable ancestor.
>   - Example policy: / = read, /a = none, /a/b = write.
>   - In the policy model, the more specific /a/b write should win.
>   - But in bwrap.rs, rootless_unreadable_roots are applied after all writable binds. So it would bind /a/b writable first, then later mask /a, which hides /a/b again.
> 
>   That pattern does not appear to be covered by the commit’s tests. So the new logic fixes the main overlapping-write/read case, but it is not obviously complete for every none/write nesting pattern.

celia-oai (approved):
does this remaining edge case makes sense:

```
- Policy: /repo = write, /repo/a = none, /repo/a/b = write

  Expected result from the policy model:

  - /repo writable
  - /repo/a denied
  - /repo/a/b writable again, because the more specific child write should win over the parent deny

  Why it may still fail on Linux bubblewrap:

  - The sandbox first binds /repo writable.
  - Then it masks /repo/a as unreadable.
  - Later it tries to bind /repo/a/b writable again.
  - But after /repo/a was masked, the path to /repo/a/b may no longer exist inside the sandbox.
  - The branch only recreates parent directories for one class of deny/write overlap, and this case does not seem to go through that path.
```

PR #14175, opened 2026-03-10
Let models opt into original image detail

fjord-oai:
## Summary

This PR narrows original image detail handling to a single opt-in feature:

- `image_detail_original` lets the model request `detail: "original"` on supported models
- Omitting `detail` preserves the default resized behavior

The model only sees `detail: "original"` guidance when the active model supports it:

- JS REPL instructions include the guidance and examples only on supported models
- `view_image` only exposes a `detail` parameter when the feature and model can use it

The image detail API is intentionally narrow and consistent across both paths:

- `view_image.detail` supports only `"original"`; otherwise omit the field
- `codex.emitImage(..., detail)` supports only `"original"`; otherwise omit the field
- Unsupported explicit values fail clearly at the API boundary instead of being silently reinterpreted
- Unsupported explicit `detail: "original"` requests fall back to normal behavior when the feature is disabled or the model does not support original detail


codex-rs/core/src/original_image_detail.rs
```diff
@@ -0,0 +1,107 @@
...
+    detail: Option<ImageDetail>,
+) -> Option<ImageDetail> {
+    match detail {
+        Some(ImageDetail::Original) if !can_request_original_image_detail(features, model_info) => {
+            None
+        }
+        other => other,
+    }
+    .or_else(|| {
+        should_force_original_image_detail(features, model_info).then_some(ImageDetail::Original)
```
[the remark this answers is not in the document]
fjord-oai: Fixed

codex-rs/core/src/codex.rs
```diff
@@ -451,6 +445,13 @@ impl Codex {
             .clone()
             .or_else(|| conversation_history.get_base_instructions().map(|s| s.text))
             .unwrap_or_else(|| model_info.get_model_instructions(config.personality));
+        let user_instructions = get_user_instructions(
+            &config,
+            Some(&model_info),
+            Some(&allowed_skills_for_implicit_invocation),
+            Some(loaded_plugins.capability_summaries()),
+        )
+        .await;
```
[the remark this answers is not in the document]
fjord-oai: added a comment

codex-rs/core/src/original_image_detail.rs
```diff
@@ -0,0 +1,112 @@
...
+        && features.enabled(Feature::ImageDetailOriginalAlways)
+}
+
+pub(crate) fn normalize_output_image_detail(
+    features: &Features,
+    model_info: &ModelInfo,
+    detail: Option<ImageDetail>,
+) -> Option<ImageDetail> {
+    if should_force_original_image_detail(features, model_info) {
+        return Some(ImageDetail::Original);
```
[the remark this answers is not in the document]
fjord-oai: added some clarifying comments

codex-rs/core/src/tools/handlers/view_image.rs
```diff
@@ -28,6 +29,13 @@ const VIEW_IMAGE_UNSUPPORTED_MESSAGE: &str =
 #[derive(Deserialize)]
 struct ViewImageArgs {
     path: String,
+    detail: Option<ViewImageDetail>,
+}
+
+#[derive(Clone, Copy, Deserialize, Eq, PartialEq)]
+#[serde(rename_all = "lowercase")]
+enum ViewImageDetail {
+    Original,
```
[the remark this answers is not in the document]
fjord-oai: clarified

codex-rs/core/src/tools/handlers/view_image.rs
```diff
@@ -66,6 +74,15 @@ impl ToolHandler for ViewImageHandler {
         };
 
         let args: ViewImageArgs = parse_arguments(&arguments)?;
+        let detail = match args.detail.as_deref() {
+            None => None,
+            Some("original") => Some(ViewImageDetail::Original),
+            Some(other) => {
+                return Err(FunctionCallError::RespondToModel(format!(
+                    "view_image.detail only supports `original`; omit `detail` for default resized behavior, got `{other}`"
+                )));
```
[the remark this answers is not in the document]
fjord-oai: The intended contract here is narrower than the raw Responses API. We do not want `view_image` to accept arbitrary `detail` strings and silently reinterpret them. The only supported explicit value is `original`; otherwise the caller should omit `detail`. In `image_detail_original_always` mode, omitting `detail` still results in `detail: "original"` on supported models, but invalid values like `low` remain invalid rather than being coerced. I updated the feature/helper comments to reflect that semantics more clearly, and added coverage locking in that `view_image` still rejects unsupported `detail` values even in always-on mode.

codex-rs/core/src/codex.rs
```diff
@@ -451,6 +445,19 @@ impl Codex {
...
+        // stale (for example JS REPL `detail: "original"` docs). Keep any
+        // model-specific guidance here advisory rather than critical: the turn
+        // should still work correctly without it, because tool exposure and
+        // runtime behavior are refreshed from current `model_info`.
+        let user_instructions = get_user_instructions(
+            &config,
+            Some(&model_info),
+            Some(&allowed_skills_for_implicit_invocation),
+            Some(loaded_plugins.capability_summaries()),
+        )
```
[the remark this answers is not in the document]
fjord-oai: known edge case, see comment next to the feature

codex-rs/core/src/tools/handlers/view_image.rs
```diff
@@ -28,6 +28,13 @@ const VIEW_IMAGE_UNSUPPORTED_MESSAGE: &str =
 #[derive(Deserialize)]
 struct ViewImageArgs {
     path: String,
+    detail: Option<String>,
```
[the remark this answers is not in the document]
fjord-oai: fixed

fjord-oai:
@codex review

codex-rs/core/src/codex.rs
```diff
@@ -451,6 +445,19 @@ impl Codex {
...
             .unwrap_or_else(|| model_info.get_model_instructions(config.personality));
+        // Resolve user instructions once for the session. If a later turn
+        // switches models, model-specific guidance in this block can become
+        // stale (for example JS REPL `detail: "original"` docs). Keep any
+        // model-specific guidance here advisory rather than critical: the turn
+        // should still work correctly without it, because tool exposure and
+        // runtime behavior are refreshed from current `model_info`.
+        let user_instructions = get_user_instructions(
+            &config,
+            Some(&model_info),
```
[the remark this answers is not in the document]
fjord-oai: yes, known edge case. see comment.

codex-rs/core/src/tools/handlers/view_image.rs
```diff
@@ -28,6 +29,13 @@ const VIEW_IMAGE_UNSUPPORTED_MESSAGE: &str =
 #[derive(Deserialize)]
 struct ViewImageArgs {
     path: String,
+    #[serde(flatten)]
+    extra: serde_json::Map<String, Value>,
```
pakrym-oai: we can define the field in the model unconditionally just not tell model about it or use it's value when the feature is off

codex-rs/core/src/codex.rs
```diff
@@ -463,6 +445,19 @@ impl Codex {
             .clone()
             .or_else(|| conversation_history.get_base_instructions().map(|s| s.text))
             .unwrap_or_else(|| model_info.get_model_instructions(config.personality));
+        // Resolve user instructions once for the session. If a later turn
+        // switches models, model-specific guidance in this block can become
+        // stale (for example JS REPL `detail: "original"` docs). Keep any
+        // model-specific guidance here advisory rather than critical: the turn
+        // should still work correctly without it, because tool exposure and
+        // runtime behavior are refreshed from current `model_info`.
+        let user_instructions = get_user_instructions(
```
pakrym-oai: I don't think we want this. We re-inject model specific instructions on model switch elsewhere

codex-rs/core/src/original_image_detail.rs
```diff
@@ -0,0 +1,91 @@
+use crate::features::Feature;
+use crate::features::Features;
+use codex_protocol::models::ImageDetail;
+use codex_protocol::openai_models::ModelInfo;
+
+pub(crate) fn can_request_original_image_detail(
```
pakrym-oai, line 6: are these two helpers worth a module?

codex-rs/core/tests/suite/view_image.rs
```diff
@@ -395,6 +396,178 @@ async fn view_image_tool_can_preserve_original_resolution_on_gpt5_3_codex() -> a
...
+
+    let req = mock.single_request();
+    let body_with_tool_output = req.body_json();
+    let output_text = req
+        .function_call_output_content_and_success(call_id)
+        .and_then(|(content, _)| content)
+        .expect("output text present");
+    assert_eq!(
+        output_text,
+        "view_image.detail does not accept null; omit `detail` for default resized behavior"
```
pakrym-oai: do we have to be this strict?

codex-rs/core/tests/suite/view_image.rs
```diff
@@ -499,6 +672,110 @@ async fn view_image_tool_keeps_legacy_behavior_below_gpt5_3_codex() -> anyhow::R
     Ok(())
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn view_image_tool_does_not_force_original_resolution_with_capability_feature_only()
```
pakrym-oai, line 692: is this a useful test? Isn't this the default behavior that's already covered?

pakrym-oai (changes requested):
Mostly nits except the change to instruction generation in the main loop.

codex-rs/core/tests/suite/view_image.rs
fjord-oai: The difference between this one and `view_image_tool_attaches_local_image` is that this specifically enables the image_detail_original feature. Once the `image_detail_original` is always on, then yes it will be redundant. Would you prefer that we keep just `view_image_tool_attaches_local_image` and turn on `image_detail_original` in that test?

codex-rs/core/tests/suite/view_image.rs
fjord-oai: yeah, that's a fair point. changed to allow null/undefined to count the same as omitting the detail arg

codex-rs/core/src/original_image_detail.rs
fjord-oai: the helpers are used in both tools/spec.rs, handlers/view_image.rs, and js_repl/mod.rs, so it seemed nicer to put them in a central location.

codex-rs/core/src/codex.rs
fjord-oai: As per offline conversation, updated the docs to reference the current state of the view_image tool to see if detail: original is available because that should always be up to date.

codex-rs/core/src/tools/handlers/view_image.rs
fjord-oai: yup, it detail: original is just dropped if it's not available.

pakrym-oai (approved):

PR #14176, opened 2026-03-10
feat: Allow sync with remote plugin status.

xl-openai:
Add forceRemoteSync to plugin/list.
When it is set to True, we will sync the local plugin status with the remote one (backend-api/plugins/list).


codex-rs/core/src/plugins/manager.rs
```diff
@@ -311,6 +424,166 @@ impl PluginsManager {
...
+                warn!(error = %err, "skipping remote plugin sync because remote state could not be fetched");
+                return Ok(RemotePluginSyncResult::default());
+            }
+        };
+        let configured_plugins = configured_plugins_from_stack(&config.config_layer_stack);
+        let curated_marketplace_root = curated_plugins_repo_path(self.codex_home.as_path());
+        let curated_marketplace_path = AbsolutePathBuf::try_from(
+            curated_marketplace_root.join(".agents/plugins/marketplace.json"),
+        )
+        .map_err(|_| PluginRemoteSyncError::LocalMarketplaceNotFound)?;
```
sayan-oai, line 439: can this race with the async curated marketplace fetch on startup? probably not a big issue

codex-rs/core/src/plugins/manager.rs
```diff
@@ -311,6 +424,166 @@ impl PluginsManager {
...
+        &self,
+        config: &Config,
+        auth: Option<&CodexAuth>,
+    ) -> Result<RemotePluginSyncResult, PluginRemoteSyncError> {
+        info!("starting remote plugin sync");
+        let remote_plugins = match fetch_remote_plugin_status(config, auth).await {
+            Ok(remote_plugins) => remote_plugins,
+            Err(err) => {
+                warn!(error = %err, "skipping remote plugin sync because remote state could not be fetched");
+                return Ok(RemotePluginSyncResult::default());
```
sayan-oai: question: callers are ok with a no-op success response on sync errors?

codex-rs/core/src/plugins/manager.rs
```diff
@@ -311,6 +424,166 @@ impl PluginsManager {
...
+                    uninstalls.push(plugin_id);
+                }
+                if is_installed || current_enabled.is_some() {
+                    result.uninstalled_plugin_ids.push(plugin_key.clone());
+                }
+                if current_enabled.is_some() {
+                    config_edits.push(ConfigEdit::ClearPath {
+                        segments: vec!["plugins".to_string(), plugin_key],
+                    });
+                }
```
sayan-oai, line 544: is the requirement that force sync makes the two plugin lists exactly the same? explicitly uninstalling local plugins that weren't in the remote list seems harsh. can we leave them alone and just have the `if` case?

sayan-oai (approved):
approving to unblock, a few small questions

codex-rs/core/src/plugins/manager.rs
xl-openai: Good call. I think some sort of client retry is needed.

codex-rs/core/src/plugins/manager.rs
xl-openai: I’ve considered this, but I think it’s more responsible to let users use whatever is already on disk rather than blocking them. I’ll add a response message that indicates whether the sync succeeded.

codex-rs/core/src/plugins/manager.rs
xl-openai: If a plugin is disabled remotely, I think it should also be disabled locally. Otherwise, things could get messy quickly. Also, if we go with a remote-only model, there’s currently no way to remove it from local once it’s been installed if we go this way.

codex-rs/core/src/plugins/manager.rs
```diff
@@ -311,6 +424,160 @@ impl PluginsManager {
...
+        }
+
+        let store = self.store.clone();
+        let store_result = tokio::task::spawn_blocking(move || {
+            for (source_path, plugin_id) in installs {
+                store.install(source_path, plugin_id)?;
+            }
+            for plugin_id in uninstalls {
+                store.uninstall(&plugin_id)?;
+            }
```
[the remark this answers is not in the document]
xl-openai: This is fine.

PR #6864, opened 2025-11-19
 fix: add custom CA certificate support for login flows

3axap4eHko:
## What

  Adds support for custom CA certificates in OAuth login flows to fix authentication failures behind corporate
  proxies with SSL/TLS inspection.

  ## Why

  The Codex CLI fails to authenticate when running behind corporate proxies that perform SSL/TLS inspection with
  custom CA certificates. The OAuth token exchange fails because `reqwest::Client::new()` only trusts system
  default certificate roots, causing it to reject connections intercepted by corporate proxies presenting
  certificates signed by custom CAs.

  This affects all Codex CLI users in enterprise environments with:
  - SSL/TLS inspection proxies (Zscaler, Palo Alto Networks, etc.)
  - Custom internal CA certificates
  - Air-gapped environments with internal certificate authorities

  Fixes #6849

  ## How

  **Implementation:**
  - Added `build_login_http_client()` function that creates an HTTP client with custom CA certificate support
  - Reads CA certificate path from environment variables (priority order):
    1. `CODEX_CA_CERTIFICATE` (primary)
    2. `SSL_CERT_FILE` (fallback, standard across many tools)
  - Updated all three authentication flows to use the custom client:
    - `exchange_code_for_tokens()` (OAuth code exchange)
    - `obtain_api_key()` (API key exchange)
    - `run_device_code_login()` (device code flow)
  - Follows the same pattern as existing `otel_provider.rs` implementation

  **Testing:**
  - Added comprehensive test coverage with 3 new tests:
    - Certificate loading via `CODEX_CA_CERTIFICATE` env var
    - Certificate loading via `SSL_CERT_FILE` fallback
    - Invalid certificate rejection with proper error handling
  - All existing tests continue to pass
  - Added `serial_test` dependency to safely test env var manipulation

  **Usage:**
  ```bash
  export CODEX_CA_CERTIFICATE=/path/to/corporate-ca.pem
  codex login
  ```
  or using the standard env var:
  ```bash
  export SSL_CERT_FILE=/path/to/corporate-ca.pem
  codex login
  ````
  
  **Testing:**

  - ✅ All tests pass (cargo test)
  - ✅ No clippy warnings (cargo clippy --tests -- -D warnings)
  - ✅ Code formatted (cargo fmt --check)
  - ✅ Tested with both environment variables
  - ✅ Invalid certificate properly rejected with clear error message

PR #14178, opened 2026-03-10
login: add custom CA support for login flows

joshka-oai:
## Stacked PRs

This work is split across three stacked PRs:

- #14178: add custom CA support for browser and device-code login flows, docs, and hermetic subprocess tests
- #14239: broaden the shared custom CA path from login to other outbound `reqwest` clients across Codex
- #14240: extend that shared custom CA handling to secure websocket TLS so websocket connections honor the same CA env vars

Review order: #14178, then #14239, then #14240.

Supersedes #6864.

Thanks to @3axap4eHko for the original implementation and investigation here. Although this version rearranges the code and history significantly, the majority of the credit for this work belongs to them.

## Problem

Login flows need to work in enterprise environments where outbound TLS is intercepted by an internal proxy or gateway. In those setups, system root certificates alone are often insufficient to validate the OAuth and device-code endpoints used during login. The change adds a login-specific custom CA loading path, but the important contracts around env precedence, PEM compatibility, test boundaries, and probe-only workarounds need to be explicit so reviewers can understand what behavior is intentional.

For users and operators, the behavior is simple: if login needs to trust a custom root CA, set `CODEX_CA_CERTIFICATE` to a PEM file containing one or more certificates. If that variable is unset, login falls back to `SSL_CERT_FILE`. If neither is set, login uses system roots. Invalid or empty PEM files now fail with an error that points back to those environment variables and explains how to recover.

## What This Delivers

Users can now make Codex login work behind enterprise TLS interception by pointing `CODEX_CA_CERTIFICATE` at a PEM bundle containing the relevant root certificates. If that variable is unset, login falls back to `SSL_CERT_FILE`, then to system roots.

This PR applies that behavior to both browser-based and device-code login flows. It also makes login tolerant of the PEM shapes operators actually have in hand: multi-certificate bundles, OpenSSL `TRUSTED CERTIFICATE` labels, and bundles that include well-formed CRLs.

## Mental model

`codex-login` is the place where the login flows construct ad hoc outbound HTTP clients. That makes it the right boundary for a narrow CA policy: look for `CODEX_CA_CERTIFICATE`, fall back to `SSL_CERT_FILE`, load every parseable certificate block in that bundle into a `reqwest::Client`, and fail early with a clear user-facing error if the bundle is unreadable or malformed.

The implementation is intentionally pragmatic about PEM input shape. It accepts ordinary certificate bundles, multi-certificate bundles, OpenSSL `TRUSTED CERTIFICATE` labels, and bundles that also contain CRLs. It does not validate a certificate chain or prove a handshake; it only constructs the root store used by login.

## Non-goals

This change does not introduce a general-purpose transport abstraction for the rest of the product. It does not validate whether the provided bundle forms a real chain, and it does not add handshake-level integration tests against a live TLS server. It also does not change login state management or OAuth semantics beyond ensuring the existing flows share the same CA-loading rules.

## Tradeoffs

The main tradeoff is keeping this logic scoped to login-specific client construction rather than lifting it into a broader shared HTTP layer. That keeps the review surface smaller, but it also means future login-adjacent code must continue to use `build_login_http_client()` or it can silently bypass enterprise CA overrides.

The `TRUSTED CERTIFICATE` handling is also intentionally a local compatibility shim. The rustls ecosystem does not currently accept that PEM label upstream, so the code normalizes it locally and trims the OpenSSL `X509_AUX` trailer bytes down to the certificate DER that `reqwest` can consume.

## Architecture

`custom_ca.rs` is now the single place that owns login CA behavior. It selects the CA file from the environment, reads it, normalizes PEM label shape where needed, iterates mixed PEM sections with `rustls-pki-types`, ignores CRLs, trims OpenSSL trust metadata when necessary, and returns either a configured `reqwest::Client` or a typed error.

The browser login server and the device-code flow both call `build_login_http_client()`, so they share the same trust-store policy. Environment-sensitive tests run through the `login_ca_probe` helper binary because those tests must control process-wide env vars and cannot reliably build a real reqwest client in-process on macOS seatbelt runs.

## Observability

The custom CA path logs which environment variable selected the bundle, which file path was loaded, how many certificates were accepted, when `TRUSTED CERTIFICATE` labels were normalized, when CRLs were ignored, and where client construction failed. Returned errors remain user-facing and include the relevant path, env var, and remediation hint.

This gives enough signal for three audiences:
- users can see why login failed and which env/file caused it
- sysadmins can confirm which override actually won
- developers can tell whether the failure happened during file read, PEM parsing, certificate registration, or final reqwest client construction

## Tests

Pure unit tests stay limited to env precedence and empty-value handling. Real client construction lives in subprocess tests so the suite remains hermetic with respect to process env and macOS sandbox behavior.

The subprocess tests verify:
- `CODEX_CA_CERTIFICATE` precedence over `SSL_CERT_FILE`
- fallback to `SSL_CERT_FILE`
- single-certificate and multi-certificate bundles
- malformed and empty-bundle errors
- OpenSSL `TRUSTED CERTIFICATE` handling
- CRL tolerance for well-formed CRL sections

The named PEM fixtures under `login/tests/fixtures/` are shared by the tests so their purpose stays reviewable.


joshka-oai (commented):
Comparison to the original PR (#6864), mainly to make review easier for @3axap4eHko:

The main behavioral intent is still the same: login can use a custom CA bundle via `CODEX_CA_CERTIFICATE`, with fallback to `SSL_CERT_FILE`, and the login flows share that client-construction path.

The biggest structural change is that most of the CA-specific code now lives in a dedicated `custom_ca.rs` module instead of adding a few hundred lines to `server.rs`. That was mainly to keep the login server code focused on the server flow and make the CA-loading logic more additive and easier to review in isolation.

Relative to the original PR, this version also leans harder into robustness and observability:
- the PEM fixtures were moved into named files and made shared across tests
- custom CA parsing now has a typed error surface and clearer user-facing failure messages
- the CA-loading path logs enough context to diagnose which env var/file won and where loading/building failed
- `TRUSTED CERTIFICATE` handling is documented as a compatibility shim, with rationale and upstream context
- CRLs are handled via `rustls-pki-types` mixed-section parsing rather than broad text stripping
- env-sensitive tests were isolated into subprocess tests, and the macOS sandbox/proxy panic was contained to the probe path so the tests stay focused on CA behavior

So from my perspective, the main delta versus #6864 is not a change in product intent so much as a reorganization around reviewability, clearer contracts, and more defensive behavior around real-world CA bundles and test environments.

Most of the credit for the actual feature still belongs to @3axap4eHko; this version rearranges and hardens the work substantially, but it is still building on the original implementation and review feedback.

pakrym-oai (approved):

PR #14225, opened 2026-03-10
Use V8 for code mode

pakrym-oai:
Link against V8 and run code mode using the V8 isolate.


codex-rs/code-mode/src/imp.rs
```diff
@@ -0,0 +1,515 @@
...
+        return Err(format_v8_exception(scope));
+    };
+    let Some(instantiated) = module.instantiate_module(scope, resolve_code_mode_module) else {
+        return Err(format_v8_exception(scope));
+    };
+    if !instantiated {
+        return Err("failed to instantiate code_mode module".to_string());
+    }
+
+    let Some(result) = module.evaluate(scope) else {
```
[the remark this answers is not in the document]
pakrym-oai: Fine for now.

defs.bzl
```diff
@@ -149,6 +149,13 @@ def codex_rust_crate(
         "BAZEL_PACKAGE": native.package_name(),
     } | rustc_env
 
+    linux_allow_multiple_definitions_flags = select({
+        "@platforms//os:linux": [
+            "-Clink-arg=-Wl,--allow-multiple-definition",
```
zbarsky-openai: why are we setting this on all crates? what is it working around?

codex-rs/code-mode/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+path = "src/lib.rs"
+
+[lints]
+workspace = true
+
+[dependencies]
+serde = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true }
+
+[target.'cfg(not(all(target_os = "linux", target_env = "musl")))'.dependencies]
```
zbarsky-openai, line 18: so does v8 not work on musl? I thought we shipped musl releases?

codex-rs/code-mode/src/libc_stack_end_shim.cc
```diff
@@ -0,0 +1,25 @@
+// Bazel's Linux linker setup does not currently resolve V8's
+// `__libc_stack_end` reference from the system runtime. Provide a weak
```
zbarsky-openai: hmm i wonder if this is resolved if we use bazeled v8 instead of this prebuilt. @adamf-oai made this work already for Rockset (and it's also what cloudflare does for workerd)

codex-rs/code-mode/Cargo.toml
pakrym-oai: Yeah, rusty_v8 has not prebuilts for musl. We'll have to figure out some way around it later.

pakrym-oai: We can in theory still maintain an optional nodejs bridge.

defs.bzl
pakrym-oai: I think we are double-linking libc++ / libc++abi / libunwind

pakrym-oai: Thinking about it it might be a pretty bad workaround.

codex-rs/code-mode/Cargo.toml
zbarsky-openai: we have build rules to build v8 from source. have you tried those yet?

pakrym-oai: Tried it locally with `bazel build @crates//:v8-146.4.0 --platforms=@llvm//platforms:linux_amd64 --action_env=V8_FROM_SOURCE=1 --action_env=PYTHON=python3`. It does switch into the source-build path, but it fails immediately in `tools/ninja_gn_binaries.py` because the Bazel build-script runfiles do not expose `./v8/DEPS` relative to the build script cwd. The generated `@crates//:v8-146.4.0` target currently has empty `build_script_data`, so the source-build path needs additional Bazel wiring before it works. I kept the smaller prebuilt-path fix here and can do a follow-up if we want to make `V8_FROM_SOURCE` work under Bazel.

zbarsky-openai: No, we don't want to use `V8_FROM_SOURCE`. We want to use the actual bazl build for v8 itself, and supply that as a "prebuilt" to the crate. It will be prebuilt from the crate's POV, but built from source as part of our build graph, and thus should work with musl, not have the duplicated libcxx ,etc

pakrym-oai: Understood. I checked the generated Bazel targets locally: today we only have the crate target plus its cargo build script, and no standalone Bazel producer for `librusty_v8.a` that we can point `RUSTY_V8_ARCHIVE` at yet. The crate side does support consuming a local archive via `RUSTY_V8_ARCHIVE`, so the path you are describing looks feasible, but it would require adding that Bazel-side archive producer first rather than just patching the current prebuilt/download path.

PR #14239, opened 2026-03-10
client: extend custom CA handling across HTTPS and websocket clients

joshka-oai:
## Stacked PRs

This work is now effectively split across two steps:

- #14178: add custom CA support for browser and device-code login flows, docs, and hermetic subprocess tests
- #14239: extend that shared custom CA handling across Codex HTTPS clients and secure websocket TLS

Note: #14240 was merged into this branch while it was stacked on top of this PR. This PR now subsumes that websocket follow-up and should be treated as the combined change.

Builds on top of #14178.

## Problem

Custom CA support landed first in the login path, but the real requirement is broader. Codex constructs outbound TLS clients in multiple places, and both HTTPS and secure websocket paths can fail behind enterprise TLS interception if they do not honor `CODEX_CA_CERTIFICATE` or `SSL_CERT_FILE` consistently.

This PR broadens the shared custom-CA logic beyond login and applies the same policy to websocket TLS, so the enterprise-proxy story is no longer split between “HTTPS works” and “websockets still fail”.

## What This Delivers

Custom CA support is no longer limited to login. Codex outbound HTTPS clients and secure websocket connections can now honor the same `CODEX_CA_CERTIFICATE` / `SSL_CERT_FILE` configuration, so enterprise proxy/intercept setups work more consistently end-to-end.

For users and operators, nothing new needs to be configured beyond the same CA env vars introduced in #14178. The change is that more of Codex now respects them, including websocket-backed flows that were previously still using default trust roots.

I also manually validated the proxy path locally with mitmproxy using:
`CODEX_CA_CERTIFICATE=~/.mitmproxy/mitmproxy-ca-cert.pem HTTPS_PROXY=http://127.0.0.1:8080 just codex`
with mitmproxy installed via `brew install mitmproxy` and configured as the macOS system proxy.

## Mental model

`codex-client` is now the owner of shared custom-CA policy for outbound TLS client construction. Reqwest callers start from the builder configuration they already need, then pass that builder through `build_reqwest_client_with_custom_ca(...)`. Websocket callers ask the same module for a rustls client config when a custom CA bundle is configured.

The env precedence is the same everywhere:
- `CODEX_CA_CERTIFICATE` wins
- otherwise fall back to `SSL_CERT_FILE`
- otherwise use system roots

The helper is intentionally narrow. It loads every usable certificate from the configured PEM bundle into the appropriate root store and returns either a configured transport or a typed error that explains what went wrong.

## Non-goals

This does not add handshake-level integration tests against a live TLS endpoint. It does not validate that the configured bundle forms a meaningful certificate chain. It also does not try to force every transport in the repo through one abstraction; it extends the shared CA policy across the reqwest and websocket paths that actually needed it.

## Tradeoffs

The main tradeoff is centralizing CA behavior in `codex-client` while still leaving adoption up to call sites. That keeps the implementation additive and reviewable, but it means the rule "outbound Codex TLS that should honor enterprise roots must use the shared helper" is still partly enforced socially rather than by types.

For websockets, the shared helper only builds an explicit rustls config when a custom CA bundle is configured. When no override env var is set, websocket callers still use their ordinary default connector path.

## Architecture

`codex-client::custom_ca` now owns CA bundle selection, PEM normalization, mixed-section parsing, certificate extraction, typed CA-loading errors, and optional rustls client-config construction for websocket TLS.

The affected consumers now call into that shared helper directly rather than carrying login-local CA behavior:
- backend-client
- cloud-tasks
- RMCP client paths that use `reqwest`
- TUI voice HTTP paths
- `codex-core` default reqwest client construction
- `codex-api` websocket clients for both responses and realtime websocket connections

The subprocess CA probe, env-sensitive integration tests, and shared PEM fixtures also live in `codex-client`, which is now the actual owner of the behavior they exercise.

## Observability

The shared CA path logs:
- which environment variable selected the bundle
- which path was loaded
- how many certificates were accepted
- when `TRUSTED CERTIFICATE` labels were normalized
- when CRLs were ignored
- where client construction failed

Returned errors remain user-facing and include the relevant env var, path, and remediation hint. That same error model now applies whether the failure surfaced while building a reqwest client or websocket TLS configuration.

## Tests

Pure unit tests in `codex-client` cover env precedence and PEM normalization behavior. Real client construction remains in subprocess tests so the suite can control process env and avoid the macOS seatbelt panic path that motivated the hermetic test split.

The subprocess coverage verifies:
- `CODEX_CA_CERTIFICATE` precedence over `SSL_CERT_FILE`
- fallback to `SSL_CERT_FILE`
- single-cert and multi-cert bundles
- malformed and empty-file errors
- OpenSSL `TRUSTED CERTIFICATE` handling
- CRL tolerance for well-formed CRL sections

The websocket side is covered by the existing `codex-api` / `codex-core` websocket test suites plus the manual mitmproxy validation above.


codex-rs/codex-client/src/custom_ca.rs
```diff
@@ -107,76 +121,71 @@ pub enum BuildLoginHttpClientError {
...
 ///
-/// Returns a [`BuildLoginHttpClientError`] when the configured CA file is unreadable, malformed,
-/// or contains a certificate block that `reqwest` cannot register as a root. Calling raw
-/// `reqwest::Client::builder()` instead would skip those user-facing errors and can make login
-/// failures in enterprise environments much harder to diagnose.
-pub fn build_login_http_client() -> Result<reqwest::Client, BuildLoginHttpClientError> {
-    build_login_http_client_with_env(&ProcessEnv, reqwest::Client::builder())
+/// Returns a [`BuildReqwestClientError`] when the configured CA file is unreadable, malformed, or
+/// contains a certificate block that `reqwest` cannot register as a root.
+pub fn build_reqwest_client_with_custom_ca(
```
xl-openai, line 179: Do we want a more generic entry point for constructing a reqwest client, in case we need to add more decorators later?

joshka-oai: Not sure what would be a decorator that we'd add there, but we're passing in the `ClientBuilder`. The builder can be pre-configured with whatever you need in the extra info. Also, code is cheap and this is fairly deep in the core (not really user facing - implementation deatil). I think if we need to augment it with other bits, telling codex to do that would be fairly simple.

I think probably ok to not make this fully open ended on extension. WDYT?

pakrym-oai (approved):

PR #14244, opened 2026-03-10
Add code_mode output helpers for text and images

pakrym-oai:
Summary
- document how code-mode can import `output_text`/`output_image` and ensure `add_content` stays compatible
- add a synthetic `@openai/code_mode` module that appends content items and validates inputs
- cover the new behavior with integration tests for structured text and image outputs

Testing
- Not run (not requested)


codex-rs/core/src/tools/code_mode_runner.cjs
```diff
@@ -146,15 +146,88 @@ function createToolsModule(context, protocol, enabledTools) {
...
+  ) {
+    return String(value);
+  }
+
+  try {
+    const serialized = JSON.stringify(value);
+    if (typeof serialized === 'string') {
+      return serialized;
+    }
+  } catch {}
```
cconger: Do we have a mechanism by which we plan on surfacing errors.  I feel like showing the model that there was a serialization error would be helpful here.

I see we have the typeerror below for image strings.

codex-rs/core/src/tools/code_mode_runner.cjs
```diff
@@ -146,15 +146,88 @@ function createToolsModule(context, protocol, enabledTools) {
...
+  if (/^(?:https?:\/\/|data:)/i.test(value)) {
+    return value;
+  }
+  throw new TypeError('output_image expects an http(s) or data URL');
+}
+
+function createCodeModeModule(context) {
+  const outputText = (value) => {
+    const item = {
+      type: 'input_text',
```
cconger, line 202: why are these keyed "input_*"

codex-rs/core/tests/suite/code_mode.rs
```diff
@@ -106,6 +106,81 @@ add_content(JSON.stringify(await exec_command({ cmd: "printf code_mode_exec_mark
...
+        success,
+        Some(false),
+        "code_mode image output failed unexpectedly"
+    );
+    assert_eq!(
+        req.custom_tool_call_output("call-1"),
+        serde_json::json!({
+            "type": "custom_tool_call_output",
+            "call_id": "call-1",
+            "output": [
```
cconger, line 315: Is it important for output to be so structured here instead of freeform?

I feel like we might also want to provide some kind of typedArray uint8 to allow byte passing in and out instead of requiring data:base64 encoding if we want to allow the model to output image data as "I want to see this" payload.

cconger (approved):
nonblocking questions inline.

codex-rs/core/tests/suite/code_mode.rs
pakrym-oai: >Is it important for output to be so structured here instead of freeform?

That's the content_item format from Responses API. That's how it distinguishes images that the model wants to see from raw text.

codex-rs/core/src/tools/code_mode_runner.cjs
pakrym-oai: image
https://developers.openai.com/api/reference/resources/responses/

codex-rs/core/src/tools/code_mode_runner.cjs
pakrym-oai: Yeah, good idea we should just let this bubble up.

PR #14246, opened 2026-03-10
Allow dynamic registration of dynamic tool calls

cconger:
Add experimental app-server support for connection-owned dynamic tool providers over both `stdio` and websocket.

Clients can now:
- register a tool with `toolProvider/register`
- unregister it with `toolProvider/unregister`

Registered tools are stored in an app-server registry, mapped back to the owning connection, and routed through the existing dynamic tool call flow. Tools with `injectIntoContext: true` are exposed to the model; tools with `injectIntoContext: false` remain callable only from `js_repl` via `codex.tool(...)`.

This also extends dynamic tool call events with `thread_id` so provider-owned tools can be invoked from newly created or resumed threads.

```mermaid
    sequenceDiagram
        participant Client as Provider Client
        participant AppServer as App Server
        participant ThreadClient as Thread Client
        participant Core as Core Thread

        Client->>AppServer: toolProvider/register(tool)
        AppServer->>AppServer: store tool -> owning connection

        ThreadClient->>AppServer: thread/start
        AppServer->>Core: load thread + sync registered tools

        ThreadClient->>AppServer: turn/start
        AppServer->>Core: submit user input
        Core-->>AppServer: DynamicToolCallRequest(thread_id, tool, args, provider_owned=true)

        AppServer->>Client: item/tool/call
        Client-->>AppServer: DynamicToolCallResponse
        AppServer->>Core: submit dynamic tool response
        Core-->>AppServer: continue turn
```


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -500,6 +500,47 @@ pub struct DynamicToolSpec {
     pub name: String,
     pub description: String,
     pub input_schema: JsonValue,
+    #[serde(default = "default_dynamic_tool_inject_into_context")]
```
apanasenko-oai: nit: `#[serde(default = "default_enabled")]`

codex-rs/app-server-protocol/src/protocol/v2.rs
apanasenko-oai, line 508: do we want to introduce namespace? pretty soon we will have problem of name uniqueness and in future integration with search, we would need to have some grouping for tools.

codex-rs/protocol/src/dynamic_tools.rs
```diff
@@ -10,15 +10,22 @@ pub struct DynamicToolSpec {
     pub name: String,
     pub description: String,
     pub input_schema: JsonValue,
+    #[serde(default = "default_inject_into_context")]
```
apanasenko-oai: #[serde(default = "default_enabled")]

codex-rs/app-server-protocol/src/protocol/v2.rs
cconger: I feel like since the names can afford `/` and `.` that we can deserialize the names into namespaces based on naming, was kinda hoping to have namespace essentially be a convention instead of a first class affordance.

PR #14247, opened 2026-03-10
Pass more params to compaction

rasmusrygaard:
Pass more params to /compact. This should give us parity with the /responses endpoint to improve caching.

I'm torn about the MCP await. Blocking will give us parity but it seems like we explicitly don't block on MCPs. Happy either way

### DIFF
--- a/codex-rs/core/src/client.rs
+++ b/codex-rs/core/src/client.rs
@@ -281,6 +281,8 @@ impl ModelClient {
         &self,
         prompt: &Prompt,
         model_info: &ModelInfo,
+        effort: Option<ReasoningEffortConfig>,
+        summary: ReasoningSummaryConfig,
         session_telemetry: &SessionTelemetry,
     ) -> Result<Vec<ResponseItem>> {
         if prompt.input.is_empty() {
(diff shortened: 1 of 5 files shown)


codex-rs/core/src/client.rs
```diff
@@ -294,10 +296,29 @@ impl ModelClient {
                 .with_telemetry(Some(request_telemetry));
 
         let instructions = prompt.base_instructions.text.clone();
+        let input = prompt.get_formatted_input();
+        let tools = create_tools_json_for_responses_api(&prompt.tools)?;
+        let reasoning = Self::build_reasoning(model_info, effort, summary);
+        let verbosity = if model_info.support_verbosity {
```
pakrym-oai, line 302: nit: consider extracting

pakrym-oai: actually, does verbosity even matter? It's used for summarization

codex-rs/core/src/compact_remote.rs
```diff
@@ -92,10 +95,20 @@ async fn run_remote_compact_task_inner_impl(
...
         .collect();
 
+    let prompt_input = history.for_prompt(&turn_context.model_info.input_modalities);
+    let tool_router = built_tools(
+        sess.as_ref(),
+        turn_context.as_ref(),
+        &prompt_input,
+        &HashSet::new(),
+        None,
+        &CancellationToken::new(),
```
pakrym-oai, line 105: I might be time to with the cancellation token through. Non blocking

codex-rs/core/src/client.rs
pakrym-oai: It does.

pakrym-oai (approved):

PR #14248, opened 2026-03-10
Add ephemeral flag support to thread fork

joeytrasatti-openai:
### Summary
This PR adds first-class ephemeral support to thread/fork, bringing it in line with thread/start. The goal is to support one-off completions on full forked threads without persisting them as normal user-visible threads.

### DIFF
--- a/codex-rs/app-server-protocol/src/protocol/v2.rs
+++ b/codex-rs/app-server-protocol/src/protocol/v2.rs
@@ -2420,6 +2420,8 @@ pub struct ThreadForkParams {
     pub base_instructions: Option<String>,
     #[ts(optional = nullable)]
     pub developer_instructions: Option<String>,
+    #[ts(optional = nullable)]
+    pub ephemeral: Option<bool>,
     /// If true, persist additional rollout EventMsg variants required to
     /// reconstruct a richer thread history on subsequent resume/fork/read.
     #[experimental("thread/fork.persistFullHistory")]
--- a/codex-rs/app-server-protocol/src/protocol/v2.rs
+++ b/codex-rs/app-server-protocol/src/protocol/v2.rs
@@ -2420,8 +2420,8 @@ pub struct ThreadForkParams {
     pub base_instructions: Option<String>,
     #[ts(optional = nullable)]
     pub developer_instructions: Option<String>,
-    #[ts(optional = nullable)]
-    pub ephemeral: Option<bool>,
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub ephemeral: bool,
     /// If true, persist additional rollout EventMsg variants required to
     /// reconstruct a richer thread history on subsequent resume/fork/read.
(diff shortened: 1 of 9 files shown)


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2420,6 +2420,8 @@ pub struct ThreadForkParams {
     pub base_instructions: Option<String>,
     #[ts(optional = nullable)]
     pub developer_instructions: Option<String>,
+    #[ts(optional = nullable)]
+    pub ephemeral: Option<bool>,
```
owenlin0: We can do this for booleans which results in slightly cleaner generated types (`field?: boolean` instead of `field?: boolean | null` and server behavior (it's in AGENTS.md):

> When you want omission to mean `false` for boolean fields, use `#[serde(default, skip_serializing_if = "std::ops::Not::not")] pub field: bool` over `Option<bool>`.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3868,41 +3865,59 @@ impl CodexMessageProcessor {
...
+                    )
+                    .await;
+                    return;
+                }
+            };
+            thread.preview = preview_from_rollout_items(&history_items);
+            thread.turns = build_turns_from_rollout_items(&history_items);
+            thread
         };
-        // forked thread names do not inherit the source thread name
```
owenlin0, line 3890: we lost this comment, could we restore?

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3868,41 +3865,59 @@ impl CodexMessageProcessor {
...
+                        format!(
+                            "failed to load rollout `{}` for thread {thread_id}: {err}",
+                            fork_rollout_path.display()
+                        ),
+                    )
+                    .await;
+                    return;
+                }
             }
+        } else {
```
owenlin0, line 3890: it's a bit hard to parse what the two branches are, could we add some comments? Or better yet, could we reuse a helper somehow? i.e. rename the `populate_resume_turns` helper (since ideally we could use it for resuming and forking a thread?) and leverage it for `thread/fork` perhaps?

codex-rs/app-server/src/codex_message_processor.rs
joeytrasatti-openai: yes will restore!

codex-rs/app-server-protocol/src/protocol/v2.rs
joeytrasatti-openai: Got it - have updated to be optional, non null

codex-rs/app-server/src/codex_message_processor.rs
joeytrasatti-openai: renamed to `populate_thread_turns` and using that!

owenlin0 (approved):

PR #14250, opened 2026-03-10
feat(core): emit turn metric for network proxy state

viyatb-oai:
## Summary
- add a per-turn `codex.turn.network_proxy` metric constant
- emit the metric from turn completion using the live managed proxy enabled state
- add focused tests for active and inactive tag emission


rreichel3-oai (approved):
lgtm!

PR #14252, opened 2026-03-10
Increase sdk workflow timeout to 15 minutes

aibrahim-oai:
- raise the sdk workflow job timeout from 10 to 15 minutes to reduce false cancellations near the current limit

### DIFF
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -7,8 +7,8 @@ on:
 
 jobs:
   sdks:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/sdk.yml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   sdks:
     runs-on: ubuntu-latest
```
pakrym-oai: change the agent instead, use the larger linux vm

pakrym-oai (approved):

PR #14259, opened 2026-03-10
Add store/load support for code mode

pakrym-oai:
adds support for transferring state across code mode invocations.


cconger (approved):
I would really prefer us to keep the isolate up between execs.

PR #14262, opened 2026-03-10
[sdk-v2] Add support for `builtinTools` and `manualToolExecution`

ashwinnathan-openai:
- Splits the built-in tool control slice out of PR #14038 , adding `thread/start.builtinTools` and `thread/start.manualToolExecution` to app-server.


codex-rs/core/src/tools/handlers/apply_patch.rs
```diff
@@ -237,6 +237,9 @@ pub(crate) async fn intercept_apply_patch(
     call_id: &str,
     tool_name: &str,
 ) -> Result<Option<FunctionToolOutput>, FunctionCallError> {
+    if !turn.tools_config.builtin_tool_enabled("apply_patch") {
```
ashwinnathan-openai, line 240: We might need this because sometimes the model might not call `apply_patch` directly, and instead emit a shell command that we intercept as `apply_patch`.

codex-rs/core/src/tools/registry.rs
```diff
@@ -334,8 +355,27 @@ impl ToolRegistryBuilder {
...
-        self.specs
-            .push(ConfiguredToolSpec::new(spec, supports_parallel_tool_calls));
+        self.specs.push(ConfiguredToolSpec::new(
+            spec,
+            supports_parallel_tool_calls,
+            false,
+        ));
+    }
+
+    pub fn push_builtin_spec(&mut self, spec: ToolSpec) {
```
ashwinnathan-openai: The prototype filtered built-ins by [mutating the registry](https://github.com/openai/codex/pull/14038/changes#diff-43f441dfd109101848f3f30f559942db74a39d31b4ff4f1aae942bb932411c0fR289) after build. This version tags specs/handlers as built-in up front, lets the router filter exposed built-in specs, and keeps runtime enforcement in the registry so now the registry doesn't mutate after build.

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -598,6 +599,18 @@ impl UnifiedExecProcessManager {
                 prefix_rule: request.prefix_rule.clone(),
             })
             .await;
+        let exec_approval_requirement = if context
```
ashwinnathan-openai, line 597: Threads through `manual_tool_execution` -- previously in prototype we would hardcode 
```
if manual_tool_execution {
        return Decision::Prompt;
    }
```
but I think having it live at the tool level allows for more granular control + semantically helps because `manual_tool_execution` != safety policy decision.

codex-rs/core/src/tools/spec.rs
```diff
@@ -239,6 +254,69 @@ impl ToolsConfig {
...
+            .is_none_or(|requested| requested.iter().any(|name| name == tool_name))
+    }
+
+    pub fn should_force_manual_tool_execution(&self) -> bool {
+        self.manual_tool_execution
+    }
+}
+
+fn builtin_tool_names() -> BTreeSet<&'static str> {
+    BTreeSet::from([
```
sdcoffey: i don't know if there's another canonical source in the codebase, but it would be good to structure this in a way that a test will fail if another built in tool is added and it's not listed here.

codex-rs/core/src/apply_patch.rs
```diff
@@ -50,9 +50,19 @@ pub(crate) async fn apply_patch(
         } => InternalApplyPatchInvocation::DelegateToExec(ApplyPatchExec {
             action,
             auto_approved: !user_explicitly_approved,
-            exec_approval_requirement: ExecApprovalRequirement::Skip {
-                bypass_sandbox: false,
-                proposed_execpolicy_amendment: None,
+            exec_approval_requirement: if turn_context
```
sdcoffey, line 53: ooo is this an inline labmda?

sdk/python/src/codex_app_server/generated/v2_all.py
```diff
@@ -2,109 +2,126 @@
 #   filename:  codex_app_server_protocol.v2.schemas.json
 
 from __future__ import annotations
-from pydantic import BaseModel, ConfigDict, Field, RootModel
-from typing import Annotated, Any, Literal
+
 from enum import Enum
+from typing import Any
```
sdcoffey: a little weird that the diff is so big in this file. I wonder if the codegen is not deterministic...

sdcoffey (commented):
looks great! just added a few comments

codex-rs/core/src/apply_patch.rs
ashwinnathan-openai: lol unsure I thought we'd need the closure syntax `|...|` for it be a lambda, I think this is just an inline if 😄

sdk/python/src/codex_app_server/generated/v2_all.py
ashwinnathan-openai: hmm also I'm unsure why codex decided to regen here, no changes in `codex_app_server` atm I'll unstage this file

codex-rs/core/src/tools/spec.rs
ashwinnathan-openai: Updated this so builtin invocation names/keys come from a single canonical registry. `push_spec` and `register_builtin_handler` now require a `BuiltinToolKey`, so passing a plain string to the builtin registration path won’t compile. This should be enough to keep builtin validation and registration in sync.

PR #14268, opened 2026-03-10
memories: exclude AGENTS and skills from stage1 input

andi-oai:
###### Why/Context/Summary
- Exclude injected AGENTS.md instructions and standalone skill payloads from memory stage 1 inputs so memory generation focuses on conversation content instead of prompt scaffolding.
- Strip only the AGENTS fragment from mixed contextual user messages during stage-1 serialization, which preserves environment context in the same message.
- Keep subagent notifications in the memory input, and add focused unit coverage for the fragment classifier, rollout policy, and stage-1 serialization path.

###### Test plan
- `just fmt`
- `cargo test -p codex-core --lib contextual_user_message`
- `cargo test -p codex-core --lib rollout::policy`
- `cargo test -p codex-core --lib memories::phase1`


wendyjiao-openai (approved):

codex-rs/core/src/contextual_user_message.rs
```diff
@@ -136,4 +143,44 @@ mod tests {
             text: "hello".to_string(),
         }));
     }
+
+    #[test]
+    fn excludes_agents_fragment_for_memories() {
```
jif-oai: can you merge all those tests instead?

codex-rs/core/src/memories/phase1.rs
```diff
@@ -484,6 +483,44 @@ mod job {
     }
 }
 
+fn sanitize_response_item_for_memories(item: &ResponseItem) -> Option<ResponseItem> {
```
jif-oai: Can you move this in the job module?

codex-rs/core/src/memories/phase1.rs
```diff
@@ -463,16 +464,14 @@ mod job {
     }
 
     /// Serializes filtered stage-1 memory items for prompt inclusion.
-    fn serialize_filtered_rollout_response_items(
+    pub(super) fn serialize_filtered_rollout_response_items(
```
jif-oai, line 467: Add a comment saying the pub is only for testing purpose

codex-rs/core/src/rollout/policy.rs
```diff
@@ -180,3 +180,37 @@ fn event_msg_persistence_mode(ev: &EventMsg) -> Option<EventPersistenceMode> {
         | EventMsg::ImageGenerationBegin(_) => None,
     }
 }
+
+#[cfg(test)]
+mod tests {
```
jif-oai: this is redundant... drop this please

codex-rs/core/src/contextual_user_message.rs
```diff
@@ -103,6 +103,13 @@ pub(crate) fn is_contextual_user_fragment(content_item: &ContentItem) -> bool {
         .any(|definition| definition.matches_text(text))
 }
 
+pub(crate) fn is_memory_excluded_contextual_user_fragment(content_item: &ContentItem) -> bool {
```
jif-oai, line 114: document this function

codex-rs/core/src/contextual_user_message.rs
andi-oai: Merged the new memory-exclusion coverage into a single table-driven test.

codex-rs/core/src/contextual_user_message.rs
andi-oai: Added a doc comment describing that this helper only omits the AGENTS/skill subset for memory stage 1.

codex-rs/core/src/memories/phase1.rs
andi-oai: Moved the sanitizer and its serialization test under the `job` module so the helper stays private.

codex-rs/core/src/memories/phase1.rs
andi-oai: Moved the sanitizer and its serialization test under the `job` module so the helper stays private.

codex-rs/core/src/rollout/policy.rs
andi-oai: Dropped the redundant rollout-policy test.

codex-rs/core/src/memories/phase1.rs
```diff
@@ -482,6 +481,120 @@ mod job {
...
+            id: id.clone(),
+            role: role.clone(),
+            content,
+            end_turn: *end_turn,
+            phase: phase.clone(),
+        })
+    }
+
+    #[cfg(test)]
+    mod tests {
```
jif-oai: this is cool but no ideal tbh... test should always be at the top end of the file
If the problem is just an import issue because you don't want to put a function as `pub`, you can do something like `pub(in crate::memories::phase1::tests)` for example for the function

andi-oai: Moved the serialization test back into the file's main bottom `tests` module and kept the helper visibility narrow with `pub(super)`.

I also tried the more specific `pub(in crate::memories::phase1::tests)` shape, but that path does not resolve in a normal build here because `phase1::tests` only exists under `#[cfg(test)]`.

PR #14271, opened 2026-03-10
chore: add a separate reject-policy flag for skill approvals

celia-oai:
## Summary
- add `skill_approval` to `RejectConfig` and the app-server v2 `AskForApproval::Reject` payload so skill-script prompts can be configured independently from sandbox and rule-based prompts
- update Unix shell escalation to reject prompts based on the actual decision source, keeping prefix rules tied to `rules`, unmatched command fallbacks tied to `sandbox_approval`, and skill scripts tied to `skill_approval`
- regenerate the affected protocol/config schemas and expand unit/integration coverage for the new flag and skill approval behavior


leoshimo-oai (approved):
lgtm - ty!

PR #14274, opened 2026-03-11
feat: search_tool migrate to bring you own tool of Responses API

apanasenko-oai:
## Why

to support a new bring your own search tool in Responses API(https://developers.openai.com/api/docs/guides/tools-tool-search#client-executed-tool-search) we migrating our bm25 search tool to use official way to execute search on client and communicate additional tools to the model. 

## What
- replace the legacy `search_tool_bm25` flow with client-executed `tool_search`
- add protocol, SSE, history, and normalization support for `tool_search_call` and `tool_search_output`
- return namespaced Codex Apps search results and wire namespaced follow-up tool calls back into MCP dispatch


codex-rs/core/src/codex.rs
```diff
@@ -3855,7 +3772,16 @@ impl Session {
...
 
-    pub(crate) async fn parse_mcp_tool_name(&self, tool_name: &str) -> Option<(String, String)> {
+    pub(crate) async fn parse_mcp_tool_name(
+        &self,
+        name: &str,
+        namespace: &Option<String>,
+    ) -> Option<(String, String)> {
+        let tool_name = if let Some(namespace) = namespace {
+            &format!("{namespace}_{name}")
+        } else {
```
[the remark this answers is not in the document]
apanasenko-oai: it should not happens.

codex-rs/core/src/context_manager/normalize.rs
```diff
@@ -38,6 +38,31 @@ pub(crate) fn ensure_call_outputs_present(items: &mut Vec<ResponseItem>) {
                     ));
                 }
             }
+            ResponseItem::ToolSearchCall {
```
pakrym-oai, line 41: good changes in this file, can we have a test for them somewhere?

codex-rs/protocol/src/models.rs
```diff
@@ -240,6 +240,13 @@ pub enum ResponseInputItem {
         call_id: String,
         output: FunctionCallOutputPayload,
     },
+    ToolSearchOutput {
+        call_id: String,
+        status: String,
+        execution: String,
+        #[ts(type = "unknown[]")]
+        tools: Vec<serde_json::Value>,
```
pakrym-oai, line 248: do we not have a tool definition type? it's ok if we don't

apanasenko-oai: I didn't find any, I can define it or leave it as is for now.

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -0,0 +1,515 @@
...
+
+pub(crate) fn split_namespace_and_name(
+    qualified_tool_name: &str,
+    tool: &ToolInfo,
+) -> (String, String) {
+    let namespace = if let Some(connector_name) = tool.connector_name.clone() {
+        format!(
+            "mcp__{}__{}",
+            tool.server_name,
+            sanitize_name(&connector_name)
```
sayan-oai: i think `connector_name` is already used elsewhere as a display name, not a tool prefix, so this might produce incorrect tool names.

codex-rs/core/src/codex.rs
```diff
@@ -3857,7 +3774,20 @@ impl Session {
             .await
     }
 
-    pub(crate) async fn parse_mcp_tool_name(&self, tool_name: &str) -> Option<(String, String)> {
+    pub(crate) async fn parse_mcp_tool_name(
+        &self,
+        name: &str,
+        namespace: &Option<String>,
+    ) -> Option<(String, String)> {
+        let tool_name = if let Some(namespace) = namespace {
```
pakrym-oai, line 3777: do we ever register mcps into namespaces?

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -0,0 +1,515 @@
...
+}
+
+#[derive(Clone)]
+struct ToolEntry {
+    name: String,
+    info: ToolInfo,
+    search_text: String,
+}
+
+impl From<HashMap<String, ToolInfo>> for ToolSearchHandler {
```
pakrym-oai: can this be a constructor instead?

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -0,0 +1,515 @@
...
+    fn from(tools: HashMap<String, ToolInfo>) -> Self {
+        Self { tools }
+    }
+}
+
+impl ToolEntry {
```
pakrym-oai: why do we need another type for ToolInfo

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -0,0 +1,515 @@
...
+                acc
+            });
+
+    let mut serialized = Vec::with_capacity(grouped.len());
+    for (namespace, tools) in grouped {
+        let Some(first_tool) = tools.first() else {
+            continue;
+        };
+
+        let description = first_tool.info.connector_description.clone().or_else(|| {
```
pakrym-oai: Why is `connector_description` so prominent in this implementation? Shouldn't this be pretty generic for any codex tool?

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -0,0 +1,515 @@
...
+                description: description.unwrap_or_default(),
+                tools,
+            },
+        ))?);
+    }
+
+    Ok(serialized)
+}
+
+pub(crate) fn split_namespace_and_name(
```
pakrym-oai: this feels like logic that should be in a shared place.

pakrym-oai: this logic == tool name generation and parsing.

codex-rs/core/src/tools/spec.rs
```diff
@@ -2180,15 +2231,24 @@ pub(crate) fn build_specs(
         builder.register_handler("request_permissions", request_permissions_handler);
     }
 
-    if config.search_tool {
-        let app_tools = app_tools.unwrap_or_default();
+    if config.search_tool
+        && let Some(app_tools) = app_tools
+    {
+        let search_tool_handler = Arc::new(ToolSearchHandler::from(app_tools.clone()));
```
pakrym-oai: Is this implementation for app tools or all defered tools?

pakrym-oai: I think it should be for all deferred tools :)

codex-rs/core/src/tools/context.rs
```diff
@@ -107,6 +112,30 @@ impl ToolOutput for CallToolResult {
     }
 }
 
+#[derive(Clone)]
+pub struct ToolSearchOutput {
+    pub tools: Vec<JsonValue>,
```
pakrym-oai: ResponsesApiTool ?

codex-rs/core/src/tools/context.rs
```diff
@@ -46,6 +47,9 @@ pub enum ToolPayload {
     Function {
         arguments: String,
     },
+    ToolSearch {
+        arguments: serde_json::Value,
```
pakrym-oai: Should we have a struct here?

codex-rs/core/src/tools/context.rs
pakrym-oai: or ToolSpec?

codex-rs/core/src/tools/js_repl/mod.rs
```diff
@@ -637,6 +637,16 @@ impl JsReplManager {
                 summary.result_is_error = Some(!output.success());
                 summary
             }
+            ResponseInputItem::ToolSearchOutput { tools, .. } => JsReplToolCallResponseSummary {
```
pakrym-oai, line 640: I don't think this is possible.

codex-rs/core/src/codex.rs
apanasenko-oai: I don't want to refactor mcp in this review, this review trying to derisk apps, and I will work in next review to extend search for mcp tools as well and properly support namespaces there.

codex-rs/core/src/tools/handlers/tool_search.rs
apanasenko-oai: this implementation focusing on search only for apps. search for mcps is separate feature.

codex-rs/core/src/tools/spec.rs
apanasenko-oai: we will extend this to all mcp tools later. this review focusing only on existing search over the apps.

codex-rs/core/src/tools/registry.rs
```diff
@@ -130,8 +132,18 @@ impl ToolRegistry {
         Self { handlers }
     }
 
-    fn handler(&self, name: &str) -> Option<Arc<dyn AnyToolHandler>> {
-        self.handlers.get(name).map(Arc::clone)
+    fn handler(&self, name: &str, namespace: Option<&str>) -> Option<Arc<dyn AnyToolHandler>> {
+        if let Some(namespace) = namespace {
+            let name = namespaced_tool_handler_name(namespace, name);
```
pakrym-oai: Why is this fallback needed?

apanasenko-oai: not all tools will have namespace.

codex-rs/core/src/tools/js_repl/mod.rs
apanasenko-oai: what do you mean? what should we do here?

codex-rs/core/src/tools/context.rs
apanasenko-oai: ToolSearchOutputTool

codex-rs/core/src/tools/js_repl/mod.rs
pakrym-oai: Eh, keep it.

codex-rs/core/src/context_manager/history_tests.rs
```diff
@@ -1357,6 +1401,59 @@ fn normalize_removes_orphan_custom_tool_call_output_panics_in_debug() {
...
+        status: "completed".to_string(),
+        execution: "client".to_string(),
+        tools: Vec::new(),
+    }];
+    let mut h = create_history_with_items(items);
+    h.normalize_history(&default_input_modalities());
+}
+
+#[test]
+fn normalize_keeps_server_tool_search_output_without_matching_call() {
```
pakrym-oai, line 1435: is this correct?

pakrym-oai: ah, debug

codex-rs/core/tests/suite/code_mode.rs
```diff
@@ -574,12 +574,12 @@ async fn code_mode_can_print_content_only_mcp_tool_result_fields() -> Result<()>
 
     let server = responses::start_mock_server().await;
     let code = r#"
-import { image_scenario } from "tools/mcp/rmcp.js";
+	import { tools } from "tools/mcp/rmcp.js";
```
pakrym-oai: unrelated? and unexpected.

apanasenko-oai: yes, sorry.

codex-rs/core/src/tools/spec.rs
```diff
@@ -1839,6 +1860,59 @@ pub(crate) fn mcp_tool_to_openai_tool(
     fully_qualified_name: String,
     tool: rmcp::model::Tool,
```
pakrym-oai, line 1861: nit, add defer to mcp_tool_to_openai_tool ?

codex-rs/core/src/tools/registry.rs
```diff
@@ -377,7 +396,14 @@ impl ToolRegistryBuilder {
     }
 }
 
-fn unsupported_tool_call_message(payload: &ToolPayload, tool_name: &str) -> String {
+fn unsupported_tool_call_message(
+    payload: &ToolPayload,
+    tool_name: &str,
+    namespace: Option<&str>,
+) -> String {
+    let tool_name = namespace
```
pakrym-oai: nit: namespaced_tool_handler_name -> tool_handler_key and let tool_handler_key take namespace: Option<&str>

codex-rs/core/src/tools/registry.rs
```diff
@@ -513,3 +546,60 @@ async fn dispatch_after_tool_use_hook(
...
+        async fn handle(
+            &self,
+            _invocation: ToolInvocation,
+        ) -> Result<Self::Output, FunctionCallError> {
+            unreachable!("test handler should not be invoked")
+        }
+    }
+
+    #[test]
+    fn handler_looks_up_namespaced_aliases_explicitly() {
```
pakrym-oai, line 575: maybe we should add an intergration test that calls a tool via namespace instead.

codex-rs/core/src/tools/spec.rs
apanasenko-oai: I did go with separate `mcp_tool_to_deferred_openai_tool` to avoid creating noise and updating every call site of `mcp_tool_to_openai_tool`.

codex-rs/core/src/codex.rs
```diff
@@ -3852,7 +3769,20 @@ impl Session {
             .await
     }
 
-    pub(crate) async fn parse_mcp_tool_name(&self, tool_name: &str) -> Option<(String, String)> {
+    pub(crate) async fn parse_mcp_tool_name(
```
pakrym-oai, line 3772: why do we need namespace here? Won't the function call come back with the correct "namespace" and "name" on it considering all the calculations are happening on mcp manager level?

pakrym-oai (changes requested):
Mostly good but something isn't right with tool naming for code mode

codex-rs/core/src/codex.rs
apanasenko-oai: mcp operates on qualified names, for apps we split them on namespace and tool name, when we do lookup for mcp handler, we need to combine them back together.

pakrym-oai (approved):

PR #14287, opened 2026-03-11
[apps] Add tool_suggest tool.

mzeng-openai:
- [x] Add tool_suggest tool.
- [x] Move chatgpt/src/connectors.rs and core/src/connectors.rs into a dedicated mod so that we have all the logic and global cache in one place.
- [x] Update TUI app link view to support rendering the installation view for mcp elicitation.


codex-rs/core/src/codex.rs
```diff
@@ -6264,6 +6057,30 @@ pub(crate) async fn built_tools(
     } else {
         None
     };
+    let auth = sess.services.auth_manager.auth().await;
+    let discoverable_connectors =
+        if turn_context.features.enabled(Feature::Apps) && turn_context.tools_config.search_tool {
```
canvrno-oai: Should this be gated by the `Feature::ToolSuggest` feature?

codex-rs/core/src/tools/handlers/tool_suggest.rs
```diff
@@ -0,0 +1,362 @@
...
+                            .to_string(),
+                    ));
+                }
+
+                let auth = session.services.auth_manager.auth().await;
+                let discoverable_connectors =
+                    connectors::list_tool_suggest_discoverable_connectors_with_auth(
+                        &turn.config,
+                        auth.as_ref(),
+                        &[],
```
canvrno-oai: If we pass an empty set here the downstream accessible-connectors filtering will not filter out connectors that are already accessible.

canvrno-oai (approved):
Tested and working, very cool stuff. Added two comments for small issues, nothing blocking. During testing I had a small hiccup (of my own making) since the connector I tested with had been installed and disabled previously. We should think about how to handle this when the connector is disabled, there are a few ways to look at this.

codex-rs/chatgpt/src/connectors.rs
```diff
@@ -1,25 +1,18 @@
-use std::collections::HashMap;
```
mzeng-openai, line 1: Move most of the logic to the connectors mod so that we can share the logic and persist global connectors cache in one place.

codex-rs/core/src/tools/handlers/tool_suggest.rs
sayan-oai: @mzeng-openai is this a problem? the comment here seems true.

codex-rs/core/src/tools/handlers/tool_suggest.rs
```diff
@@ -0,0 +1,369 @@
...
+            &connector,
+        );
+        let response = session
+            .request_mcp_server_elicitation(turn.as_ref(), request_id, params)
+            .await;
+        let completed = response
+            .as_ref()
+            .is_some_and(|response| response.action == ElicitationAction::Accept);
+
+        if completed {
```
sayan-oai, line 196: if the elicitation is accepted and the connector is installed, are the underlying mcp tools actually available in the same turn? we dont actually create new handlers for them in `ToolRouter`, and that is static for the turn.

right now in `tool_suggest_description.md`, we tell the model it can use the "newly available tool" in the same turn, but im not sure that's true. lmk if im misunderstanding.

codex-rs/core/templates/search_tool/tool_suggest_description.md
```diff
@@ -0,0 +1,22 @@
...
+
+Discoverable tools:
+{{discoverable_tools}}
+
+Workflow:
+
+1. Match the user's request against the discoverable tools list above.
+2. If one tool clearly fits, call `tool_suggest` with:
+   - `tool_type`: `connector` or `plugin`
+   - `action_type`: `install` or `enable`
```
sayan-oai, line 20: right now only `connector` + `install` is supported. i know we will fast-follow with others, but should we include those in the tool description now? dont want to confuse the model.

codex-rs/core/src/tools/handlers/tool_suggest.rs
mzeng-openai: See below, we actually hard refresh the tools (right now connectors, will quick follow with plugins.

codex-rs/core/templates/search_tool/tool_suggest_description.md
```diff
@@ -0,0 +1,22 @@
+# Tool suggestion discovery
+
+Suggests a discoverable connector or plugin when the user clearly wants a capability that is not currently available in the active `tools` list.
+
+Use this ONLY after `search_tool_bm25` fails to find a good match and the user's request strongly matches one of the discoverable tools listed below.
```
sayan-oai: im OOTL on elicitations, but wanted to double-check we have, or plan to, confirm this flow is in-distribution for the models? specifically:

looking for tools -> search_tool -> no match -> tool_suggest -> user installs -> use new tool?

mzeng-openai: You got it correctly!

codex-rs/core/src/tools/handlers/tool_suggest.rs
```diff
@@ -0,0 +1,487 @@
...
+    install_url: &str,
+) -> serde_json::Value {
+    json!({
+        TOOL_SUGGEST_APPROVAL_KIND_KEY: TOOL_SUGGEST_APPROVAL_KIND_VALUE,
+        TOOL_SUGGEST_TOOL_TYPE_KEY: tool_type.as_str(),
+        TOOL_SUGGEST_SUGGEST_TYPE_KEY: action_type.as_str(),
+        TOOL_SUGGEST_REASON_KEY: suggest_reason,
+        TOOL_SUGGEST_TOOL_ID_KEY: tool_id,
+        TOOL_SUGGEST_TOOL_NAME_KEY: tool_name,
+        TOOL_SUGGEST_INSTALL_URL_KEY: install_url,
```
sayan-oai: why not a struct we serialize/deserialize?

codex-rs/tui/src/bottom_pane/mcp_server_elicitation.rs
```diff
@@ -266,8 +299,63 @@ impl McpServerElicitationFormRequest {
...
+
+    pub(crate) fn request_id(&self) -> &McpRequestId {
+        &self.request_id
+    }
+}
+
+fn parse_tool_suggestion_request(meta: Option<&Value>) -> Option<ToolSuggestionRequest> {
+    let meta = meta?.as_object()?;
+    if meta.get(APPROVAL_META_KIND_KEY).and_then(Value::as_str)
+        != Some(APPROVAL_META_KIND_TOOL_SUGGESTION)
```
sayan-oai, line 326: it seems like because we inject extra tool suggestion metadata into the existing MCP elicitation, the TUI has to be aware of how to parse it differently. would it be possible to restructure this so tool suggestions are a first-class typed elicitation (or at least use a shared metadata type) instead of the extra custom layer?

codex-rs/tui/src/bottom_pane/app_link_view.rs
```diff
@@ -113,6 +142,31 @@ impl AppLinkView {
...
+        self.elicitation_target.is_some()
+    }
+
+    fn resolve_elicitation(&self, decision: ElicitationAction) {
+        let Some(target) = self.elicitation_target.as_ref() else {
+            return;
+        };
+        self.app_event_tx.send(AppEvent::SubmitThreadOp {
+            thread_id: target.thread_id,
+            op: Op::ResolveElicitation {
```
sayan-oai, line 155: maybe nit: should `app_link_view` know about resolving elicitations, request_id, etc? its mostly rendering/input handling, should we just return an outcome and consolidate the elicitation parsing + resolution in a single spot in the bottom_pane mod?

codex-rs/tui/src/bottom_pane/mcp_server_elicitation.rs
mzeng-openai: The MCP elicitation type works slightly differently, the closest is the URL elicitation which navigates the user to a URL. I'm planning to consolidate MCP and connector auth to both use this elicitation format in a separate PR.

I think we are ok with this one since the app-server clients also need to do a lot of custom things for connectors because connectors don't follow the exact MCP auth flow.

codex-rs/core/src/tools/handlers/tool_suggest.rs
```diff
@@ -0,0 +1,487 @@
...
+        "{tool_name} could help with this request.\n\n{suggest_reason}\n\nOpen ChatGPT to {} it, then confirm here if you finish.",
+        args.action_type.as_str()
+    );
+
+    McpServerElicitationRequestParams {
+        thread_id,
+        turn_id: Some(turn_id),
+        server_name: CODEX_APPS_MCP_SERVER_NAME.to_string(),
+        request: McpServerElicitationRequest::Form {
+            meta: Some(build_tool_suggestion_meta(
```
sayan-oai: will this be cleaned up in the follow up you mentioned [here](https://github.com/openai/codex/pull/14287#discussion_r2921914085)?

it doesnt seem great to shove all the tool suggestion stuff into this meta field like i mentioned.

sayan-oai (approved):
approving so we can start testing. i think the elicitation parsing and resolution should be consolidated, right now it feels like the TUI needs to know too much about how things are passed and structured.

codex-rs/core/templates/search_tool/tool_suggest_description.md
mzeng-openai: updated to support plugins.

PR #14295, opened 2026-03-11
Support waiting for code_mode sessions

pakrym-oai:
## Summary
- persist the code mode runner process in the session-scoped code mode store
- switch the runner protocol from `init` to `start` with explicit session ids
- handle runner-side session processing without the init waiter queue

## Validation
- just fmt
- cargo check -p codex-core
- node --check codex-rs/core/src/tools/code_mode_runner.cjs


cconger (approved):
We spoke out of band, we're likely missing a default yield time on the first exec.

I do think we should revisit if the set_yield_time and set_max_tokens makes as being driven from within the worker calling to parent instead of some other annotation/pragma style approach for encoding the data to the cjs core on the way into the runtime.

PR #14300, opened 2026-03-11
fix(otel): make HTTP trace export survive app-server runtimes

owenlin0:
## Summary

This PR fixes OTLP HTTP trace export in runtimes where the previous exporter setup was unreliable, especially around app-server usage. It also removes the old `codex_otel::otel_provider` compatibility shim and switches remaining call sites over to the crate-root `codex_otel::OtelProvider` export.

## What changed

- Use a runtime-safe OTLP HTTP trace exporter path for Tokio runtimes.
- Add an async HTTP client path for trace export when we are already inside a multi-thread Tokio runtime.
- Make provider shutdown flush traces before tearing down the tracer provider.
- Add loopback coverage that verifies traces are actually sent to `/v1/traces`:
  - outside Tokio
  - inside a multi-thread Tokio runtime
  - inside a current-thread Tokio runtime
- Remove the `codex_otel::otel_provider` shim and update remaining imports.

## Why

I hit cases where spans were being created correctly but never made it to the collector. The issue turned out to be in exporter/runtime behavior rather than the span plumbing itself. This PR narrows that gap and gives us regression coverage for the actual export path.


codex-rs/otel/src/provider.rs
```diff
@@ -321,6 +325,34 @@ fn build_tracer_provider(
         } => {
             debug!("Using OTLP Http exporter for traces: {endpoint}");
 
+            if crate::otlp::current_tokio_runtime_is_multi_thread() {
```
pakrym-oai, line 328: when is it not true?

owenlin0: our main binaries are multi-threaded, but `app-server-test-client` and unit tests marked with just `#[tokio::test]` are single-threaded

pakrym-oai (approved):

PR #14301, opened 2026-03-11
feat(app-server): keep request traces attached to async work

owenlin0:
## Summary

This PR keeps app-server request trace context alive for the full lifetime of the work that request kicks off, instead of losing trace lineage once the request handler returns or hands work off to background tasks. 

This approach is relevant for `thread/start` and other RPC handlers that run in a non-blocking way. In the near future we'll most likely want to make all app-server handlers run in a non-blocking way by default, and only queue operations that must operate in order (e.g. thread RPCs per thread?), so we want to make sure tracing in app-server just generally works.

Depends on https://github.com/openai/codex/pull/14300

**Before**
image

**After**
image

## What changed

- Keep request-scoped trace context around until we send the final response or error, or the connection closes.
- Thread that trace context through detached `thread/start` work so background startup stays attached to the originating request.
- Pass request trace context through to downstream core operations, including:
  - thread creation
  - resume/fork flows
  - turn submission
  - review
  - interrupt
  - realtime conversation operations
- Add tracing tests that verify:
  - remote W3C trace context is preserved for `thread/start`
  - remote W3C trace context is preserved for `turn/start`
  - downstream core spans stay under the originating request span
  - request-scoped tracing state is cleaned up correctly
- Clean up shutdown behavior so detached background tasks and spawned threads are drained before process exit.

## Why

Before this, traces could get disconnected from the async work they spawned, especially for `thread/start`, which is intentionally fire-and-forget from the request loop. That made it much harder to follow startup and turn execution end to end. This change keeps the request lineage intact all the way through the work it triggers.


owenlin0:
@codex review

PR #14313, opened 2026-03-11
Defer initial context insertion until the first turn

charley-oai:
## Summary
- defer fresh-session `build_initial_context()` until the first real turn instead of seeding model-visible context during startup
- rely on the existing `reference_context_item == None` turn-start path to inject full initial context on that first real turn (and again after baseline resets such as compaction)
- add a regression test for `InitialHistory::New` and update affected deterministic tests / snapshots around developer-message layout, collaboration instructions, personality updates, and compact request shapes

## Notes
- this PR does not add any special empty-thread `/compact` behavior
- most of the snapshot churn is the direct result of moving the initial model-visible context from startup to the first real turn, so first-turn request layouts no longer contain a pre-user startup copy of permissions / environment / other developer-visible context
- remote manual `/compact` with no prior user still skips the remote compact request; local first-turn `/compact` still issues a compact request, but that request now reflects the lack of startup-seeded context

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -1927,21 +1927,9 @@ impl Session {
         };
         match conversation_history {
             InitialHistory::New => {
-                // Build and record initial items (user instructions + environment context)
-                // TODO(ccunningham): Defer initial context insertion until the first real turn
-                // starts so it reflects the actual first-turn settings (permissions, etc.) and
-                // we do not emit model-visible "diff" updates before the first user message.
-                let items = self.build_initial_context(&turn_context).await;
-                self.record_conversation_items(&turn_context, &items).await;
-                {
-                    let mut state = self.state.lock().await;
-                    state.set_reference_context_item(Some(turn_context.to_turn_context_item()));
-                }
+                // Defer initial context insertion until the first real turn starts so
+                // turn/start overrides can be merged before we write model-visible context.
                 self.set_previous_turn_settings(None).await;
-                // Ensure initial items are visible to immediate readers (e.g., tests, forks).
-                if !is_subagent {
-                    self.flush_rollout().await;
-                }
             }
             InitialHistory::Resumed(resumed_history) => {
                 let rollout_items = resumed_history.history;
--- a/codex-rs/core/tests/suite/compact.rs
+++ b/codex-rs/core/tests/suite/compact.rs
@@ -500,19 +500,14 @@ async fn manual_compact_emits_api_and_local_token_usage_events() {
 
     let server = start_mock_server().await;
 
-    let first_turn = sse(vec![
-        ev_assistant_message("m0", FIRST_REPLY),
-        ev_completed_with_tokens("r0", 80),
-    ]);
-
     // Compact run where the API reports zero tokens in usage. Our local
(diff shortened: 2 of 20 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -1927,21 +1927,9 @@ impl Session {
         };
         match conversation_history {
             InitialHistory::New => {
-                // Build and record initial items (user instructions + environment context)
-                // TODO(ccunningham): Defer initial context insertion until the first real turn
-                // starts so it reflects the actual first-turn settings (permissions, etc.) and
-                // we do not emit model-visible "diff" updates before the first user message.
-                let items = self.build_initial_context(&turn_context).await;
-                self.record_conversation_items(&turn_context, &items).await;
```
charley-oai, line 1935: Deleting the startup-time `build_initial_context(...)` block is safe because we are **not** removing the “inject full initial context on the first turn” behavior; we are only moving it off the eager startup path and onto the existing per-turn path.

On every real turn, `run_turn` calls [`record_context_updates_and_set_reference_context_item(...)`](https://github.com/openai/codex/blob/1170b52fb9664ae5a5f2e34a020edce81e95b11b/codex-rs/core/src/codex.rs#L5356-L5357) **before** the user prompt is recorded. That helper explicitly checks whether `reference_context_item` is missing and, if so, calls [`build_initial_context(turn_context)`](https://github.com/openai/codex/blob/1170b52fb9664ae5a5f2e34a020edce81e95b11b/codex-rs/core/src/codex.rs#L3500-L3515) instead of diffing.

So the invariant remains:

- fresh thread: `reference_context_item == None`
- first real turn: inject full initial context via `build_initial_context(...)`
- later turns: diff against the stored baseline
- after compaction clears the baseline: the same `None` path reinjects full initial context

In other words, we are **not** relying on diffs for the first turn. We are still relying on `build_initial_context(...)` for the first turn; we are just no longer doing that work too early at startup, which is what caused the model-visible “initial context, then first-turn diff before the first user message” artifact.

codex-rs/core/tests/suite/compact.rs
```diff
@@ -516,39 +521,53 @@ async fn manual_compact_emits_api_and_local_token_usage_events() {
...
-    // Second TokenCount: from the local post-compaction estimate.
-    let last = wait_for_event_match(&codex, |ev| match ev {
-        EventMsg::TokenCount(tc) => tc
-            .info
-            .as_ref()
-            .map(|info| info.last_token_usage.total_tokens),
-        _ => None,
-    })
-    .await;
+    let mut token_counts = Vec::new();
```
pakrym-oai: why did this have to change?

charley-oai: Before, manual_compact_emits_api_and_local_token_usage_events was compacting an empty thread, so the next two TokenCount events happened to line up with the compact API usage and the local recompute. After the no-op change, that setup no longer exercised compaction at all, so I had to seed one real user turn first in codex-rs/core/tests/suite/compact.rs:503

charley-oai: Nvm reverted the empty compaction guard as discussed

pakrym-oai (approved):

PR #14351, opened 2026-03-11
Publish codex-app-server in GitHub releases

cconger:
Add codex-app-server to the existing release build, signing, and staging paths so tagged releases upload app-server binaries alongside the other published artifacts.

### DIFF
--- a/.github/workflows/rust-release.yml
+++ b/.github/workflows/rust-release.yml
@@ -214,7 +214,7 @@ jobs:
         shell: bash
         run: |
           echo "CARGO_PROFILE_RELEASE_LTO: ${CARGO_PROFILE_RELEASE_LTO}"
-          cargo build --target ${{ matrix.target }} --release --timings --bin codex --bin codex-responses-api-proxy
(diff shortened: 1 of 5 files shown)


leoshimo-oai (approved):

.github/workflows/rust-release.yml
```diff
@@ -214,7 +214,7 @@ jobs:
         shell: bash
         run: |
           echo "CARGO_PROFILE_RELEASE_LTO: ${CARGO_PROFILE_RELEASE_LTO}"
-          cargo build --target ${{ matrix.target }} --release --timings --bin codex --bin codex-responses-api-proxy
+          cargo build --target ${{ matrix.target }} --release --timings --bin codex --bin codex-responses-api-proxy --bin codex-app-server
```
bolinfest, line 217: The link time is going to kill us: we can't be adding this to our release jobs.

How much smaller is this standalone binary, anyway?

bolinfest (changes requested):

.github/workflows/rust-release.yml
cconger: 55MB versus 90MB

PR #14374, opened 2026-03-11
[telemetry] surface projection for user message type

rhan-oai:
- surface projection for user message type
- introduce item metadata feature flag
- introduce and emit item uuid as metadata, behind feature flag


codex-rs/protocol/src/models.rs
```diff
@@ -257,6 +260,63 @@ pub enum ResponseInputItem {
...
+#[serde(rename_all = "snake_case")]
+#[ts(rename_all = "snake_case")]
+pub enum UserMessageType {
+    Prompt,
+    PromptSteering,
+    PromptQueued,
+}
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Eq, JsonSchema, TS)]
+pub struct ResponseItemMetadata {
```
owenlin0: Can we define a unique type per ResponseItem variant? So `ResponseItemMessageMetadata` / `ResponseItemFunctionCallOutputMetadata`/ etc.?

I expect we'll have specific metadata fields for each variant (i.e. `user_message_type` only makes sense on Messages)

owenlin0: and looking at the response item types, I think `uuid` is only necessary on user messages since everything else has a unique `call_id`

codex-rs/protocol/src/models.rs
```diff
@@ -257,6 +260,63 @@ pub enum ResponseInputItem {
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub user_message_type: Option<UserMessageType>,
+    /// Client-visible UUID generated by Codex for this item.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub uuid: Option<String>,
```
owenlin0: if we define unique metadata types, I think we could also make `user_message_type` and `uuid` non-optional

codex-rs/protocol/src/models.rs
```diff
@@ -257,6 +260,63 @@ pub enum ResponseInputItem {
...
+}
+
+impl ResponseItem {
+    /// Ensures message metadata includes a generated UUID.
+    pub fn with_generated_metadata_uuid(self) -> Self {
```
owenlin0: Instead of introducing this helper, can we just maintain this invariant when we construct a `ResponseItemMessageMetadata` struct? This would naturally be required once we make `uuid` non-optional

codex-rs/protocol/src/models.rs
```diff
@@ -250,6 +253,63 @@ pub enum ResponseInputItem {
...
```
dylan-hurd-oai: I think it would make more sense to populate this when we create the `ResponseItemMetadata` object, e.g. with `default` macro/impl or a `ResponseItemMetadata::new` method

codex-rs/protocol/src/models.rs
```diff
@@ -250,6 +253,63 @@ pub enum ResponseInputItem {
...
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize, PartialEq, Eq, JsonSchema, TS)]
+pub struct ResponseItemMetadata {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub user_message_type: Option<UserMessageType>,
+    /// Client-visible UUID generated by Codex for this item.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub uuid: Option<String>,
```
dylan-hurd-oai: let's just call this `metadata_id` instead of uuid - it might be slightly repetitive but we have many IDs floating around with ReponseItem

codex-rs/protocol/src/models.rs
rhan-oai: was considering this, sgtm

codex-rs/protocol/src/models.rs
rhan-oai: sg on the variants, for id's uuid will be needed on tools since call_id refers to each invocation and collapses call/output which opposes item level identification

owenlin0: hmm, can you explain more why `call_id` does not work for function calls? a call_id is 1:1 with a function call

owenlin0: i think we only need to care about what codex (as a client of responsesapi) sends, not receives, so it only makes sense to attach metadata on function call outputs (not inputs/arguments, which is what responsesapi sends back to codex when the model decides to trigger a tool)

rhan-oai: looping in Mark to clarify deduping needs

codex-rs/protocol/src/models.rs
```diff
@@ -257,6 +260,33 @@ pub enum ResponseInputItem {
     },
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "snake_case")]
+#[ts(rename_all = "snake_case")]
+pub enum UserMessageType {
```
owenlin0, line 267: nit: is it worth calling this something like `UserMessageSubmissionType` and having these values?
```
Regular
Steering
Queued
```

seems more precise than `Type`

codex-rs/protocol/src/models.rs
```diff
@@ -257,6 +260,33 @@ pub enum ResponseInputItem {
...
+    PromptSteering,
+    PromptQueued,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, JsonSchema, TS)]
+pub struct ResponseItemMessageMetadata {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub user_message_type: Option<UserMessageType>,
+    /// Client-visible metadata ID generated by Codex for this item.
```
owenlin0, line 278: maybe worth documenting this field allows us to de-dupe

owenlin0 (commented):
looking good overall! just some small comments

owenlin0 (approved):
preapproving

PR #14382, opened 2026-03-11
check for large binaries in CI

owenlin0:
Prevent binaries >500KB from being committed. And maintain an allowlist if we need to bypass on a case-by-case basis.

I checked the currently tracked binary-like assets in the repo. There are only 5 obvious committed binaries by extension/MIME type:
- `.github/codex-cli-splash.png`: `838,131` bytes, about `818 KiB`
- `codex-rs/vendor/bubblewrap/bubblewrap.jpg`: `40,239` bytes, about `39 KiB`
- `codex-rs/skills/src/assets/samples/skill-creator/assets/skill-creator.png`: `1,563` bytes
- `codex-rs/skills/src/assets/samples/openai-docs/assets/openai.png`: `1,429` bytes
- `codex-rs/skills/src/assets/samples/skill-installer/assets/skill-installer.png`: `1,086` bytes

So `500 KB` looks like a good default for this repo. It would only trip on one existing intentional asset, which keeps the allowlist small and the policy easy to understand.

Here's a smoke-test from a throwaway branch that tries to commit a large binary: https://github.com/openai/codex/actions/runs/22971558828/job/66689330435?pr=14383

### DIFF
--- a/scripts/check_binary_blob_size.py
+++ b/scripts/check_binary_blob_size.py
@@ -1,0 +1,195 @@
+#!/usr/bin/env python3
+
+from __future__ import annotations
+
+import argparse
+import os
+import subprocess
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+
+
+DEFAULT_MAX_BYTES = 500 * 1024
+
+
+@dataclass(frozen=True)
+class BinaryBlob:
+    path: str
+    size_bytes: int
+    is_allowlisted: bool
+
+
+def run_git(*args: str) -> str:
+    result = subprocess.run(
+        ["git", *args],
+        check=True,
+        capture_output=True,
+        text=True,
+    )
+    return result.stdout
+
+
+def load_allowlist(path: Path) -> set[str]:
+    allowlist: set[str] = set()
+    for raw_line in path.read_text(encoding="utf-8").splitlines():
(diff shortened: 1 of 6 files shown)


scripts/check_binary_blob_size.py
```diff
@@ -0,0 +1,195 @@
...
+    parser.add_argument(
+        "--allowlist",
+        type=Path,
+        required=True,
+        help="Path to the newline-delimited allowlist file.",
+    )
+    args = parser.parse_args()
+
+    allowlist = load_allowlist(args.allowlist)
+    binary_blobs = collect_changed_binary_blobs(args.base, args.head, allowlist)
```
bolinfest: Should we only check binary blobs? Why should oversized text files be OK by default?

bolinfest: Maybe some of our codegen will hit the limit, but we should add it to the allowlist if so?

bolinfest (approved):
Note we'll have to update our GitHub ruleset to include this as a blocking check once this lands on `main`.

scripts/check_binary_blob_size.py
owenlin0: yeah good push, i updated this to check all files and updated the allowlist

PR #14005, opened 2026-03-08
Add in-process app server and wire up exec to use it

etraut-openai:
This is a subset of PR #13636. See that PR for a full overview of the architectural change.

This PR implements the in-process app server and modifies the non-interactive "exec" entry point to use the app server.

PR #14392, opened 2026-03-11
chore(app-server): stop emitting codex/event/ notifications

owenlin0:
## Description

This PR stops emitting legacy `codex/event/*` notifications from the public app-server transports.

It's been a long time coming! app-server was still producing a raw notification stream from core, alongside the typed app-server notifications and server requests, for compatibility reasons. Now, external clients should no longer be depending on those legacy notifications, so this change removes them from the stdio and websocket contract and updates the surrounding docs, examples, and tests to match.

### Caveat
I left the "in-process" version of app-server alone for now, since `codex exec` was recently based on top of app-server via this in-process form here: https://github.com/openai/codex/pull/14005

Seems like `codex exec` still consumes some legacy notifications internally, so this branch only removes `codex/event/*` from app-server over stdio and websockets. 

## Follow-up

Once `codex exec` is fully migrated off `codex/event/*` notifications, we'll be able to stop emitting them entirely entirely instead of just filtering it at the external transport boundary.

### DIFF
--- a/codex-rs/app-server-protocol/schema/json/ClientRequest.json
+++ b/codex-rs/app-server-protocol/schema/json/ClientRequest.json
@@ -813,7 +813,7 @@
           "type": "boolean"
         },
         "optOutNotificationMethods": {
-          "description": "Exact notification method names that should be suppressed for this connection (for example `codex/event/session_configured`).",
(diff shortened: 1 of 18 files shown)


celia-oai (approved):
huge!

PR #14394, opened 2026-03-11
chore(app-server): delete unused rpc methods from v1.rs

owenlin0:
## Description

This PR trims `app-server-protocol`'s v1 surface down to the small set of legacy types we still actually use. 

Unfortunately, we can't delete all of them yet because:
- a few one-off v1 RPCs are still used by the Codex app
- a few of these app-server-protocol v1 types are actually imported by core crates

This change deletes that unused RPC surface, keeps the remaining compatibility types in place, and makes the crate root re-export only the v1 structs that downstream crates still depend on.

## Why

The main goal here is to make the legacy protocol surface match reality. Leaving a large pile of dead v1 structs in place makes it harder to tell which compatibility paths are still intentional, and it keeps old schema/types around even though nothing should be building against them anymore.

This also gives us a cleaner boundary for future cleanup. Instead of re-exporting all of `protocol::v1::*`, the crate now explicitly exposes only the v1 types that are still live, which makes it much easier to see what remains and delete more safely later.

## What changed

- Deleted the unused v1 RPC/request/response structs from `app-server-protocol/src/protocol/v1.rs`.
- Kept the small set of v1 compatibility types that are still live, including:
  - `initialize`
  - `getConversationSummary`
  - `getAuthStatus`
  - `gitDiffToRemote`
  - legacy approval payloads
  - config-related structs still used by downstream crates
- Replaced the blanket `pub use protocol::v1::*` export in `app-server-protocol/src/lib.rs` with an explicit list of the remaining supported v1 types.
- Regenerated the schema/type artifacts, which also updated the `InitializeCapabilities` opt-out example to use `thread/started` instead of the old `codex/event/session_configured` example.

## Validation

## Follow-up

The next cleanup is to keep shrinking the remaining v1 compatibility surface as callers migrate off it. Once the remaining consumers stop importing these legacy types, we should be able to remove more of the v1 module and eventually stop exporting it from the crate root entirely.


celia-oai (approved):
huge

PR #14395, opened 2026-03-11
fix(network-proxy): serve HTTP proxy listener as HTTP/1

viyatb-oai:
## Summary
- switch the local HTTP proxy listener from Rama's auto server to explicit HTTP/1 so CONNECT clients skip the version-sniffing pre-read path
- move rustls crypto-provider bootstrap into the HTTP proxy runner so direct callers do not need hidden global init
- add a regression test that exercises a plain HTTP/1 CONNECT request against a live loopback listener


rreichel3-oai (approved):
Looks good!

PR #14400, opened 2026-03-11
Use a private desktop for Windows sandbox instead of Winsta0\Default

iceweasel-oai:
## Summary
- launch Windows sandboxed children on a private desktop instead of `Winsta0\Default`
- make private desktop the default while keeping `windows.sandbox_private_desktop=false` as the escape hatch
- centralize process launch through the shared `create_process_as_user(...)` path
- scope the private desktop ACL to the launching logon SID

## Why
Today sandboxed Windows commands run on the visible shared desktop. That leaves an avoidable same-desktop attack surface for window interaction, spoofing, and related UI/input issues. This change moves sandboxed commands onto a dedicated per-launch desktop by default so the sandbox no longer shares `Winsta0\Default` with the user session.

The implementation stays conservative on security with no silent fallback back to `Winsta0\Default`

If private-desktop setup fails on a machine, users can still opt out explicitly with `windows.sandbox_private_desktop=false`.

## Validation
- `cargo build -p codex-cli`
- elevated-path `codex exec` desktop-name probe returned `CodexSandboxDesktop-*`
- elevated-path `codex exec` smoke sweep for shell commands, nested `pwsh`, jobs, and hidden `notepad` launch
- unelevated-path full private-desktop compatibility sweep via `codex exec` with `-c windows.sandbox=unelevated`


codex-rs/windows-sandbox-rs/src/desktop.rs
```diff
@@ -0,0 +1,223 @@
...
+        return Err(anyhow::anyhow!(
+            "GetSecurityInfo failed for private desktop: {get_code}"
+        ));
+    }
+
+    let mut updated_dacl: *mut ACL = ptr::null_mut();
+    let set_entries_code = SetEntriesInAclW(
+        entries.len() as u32,
+        entries.as_ptr(),
+        existing_dacl,
```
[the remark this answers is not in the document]
iceweasel-oai: fixed in https://github.com/openai/codex/pull/14400/changes/1896ab11f0aab265b72d621fdabcbc9e3d6461b9

codex-rs/windows-sandbox-rs/src/desktop.rs
```diff
@@ -0,0 +1,193 @@
...
+                    err,
+                    format_last_error(err),
+                ),
+                logs_base_dir,
+            );
+            return Err(anyhow::anyhow!("CreateDesktopW failed: {err}"));
+        }
+
+        unsafe {
+            grant_desktop_access(handle, logs_base_dir)?;
```
viyatb-oai: If `grant_desktop_access(handle, logs_base_dir)?` fails here, `PrivateDesktop` has not been constructed yet, so the newly created desktop handle never reaches the `Drop` impl and is leaked. Because private desktops are now the default and the PR already expects some machines to need the escape hatch, repeated failures can accumulate leaked desktops until the parent process exits. Consider wrapping the handle in an RAII guard immediately after `CreateDesktopW` or explicitly calling `CloseDesktop(handle)` on the pre-construction error paths.

viyatb-oai (approved):
lgtm - one non blocking comment

codex-rs/windows-sandbox-rs/src/desktop.rs
iceweasel-oai: good catch - addressed in https://github.com/openai/codex/pull/14400/changes/366cafdf4442a05f0761aa6b37afd8e4f9262e8a

PR #14416, opened 2026-03-12
feature: reuse skill-scoped managed network proxies for skill scripts

celia-oai:
## Summary
This follows up the managed-network override work for skills by wiring those overrides into shell execution. Before this change, skill scripts still reused the session-level managed network proxy, so per-skill allow/deny domain overrides were not applied through a dedicated proxy.

- add `NetworkProxySpec` helpers to apply `managed_network_override` settings and derive a normalized cache key so equivalent specs reuse the same proxy
- introduce a shared `SkillNetworkProxyCache` in session services and thread it through session and subagent construction
- update unix escalation to resolve the owning skill for a script, start or reuse the matching shared proxy, and fall back to the session proxy when no skill override applies
- add unit coverage for partial override replacement and normalized cache-key reuse

Related:
- Follow-up to #14522

## Testing
tested with a skill that executes `fetch_example.sh` which fetches from www.example.com.

### case 1
with the following skill permission:
```
permissions:
  network:
    enabled: true
```

result:
```
"aggregatedOutput": "\nHTTP_STATUS:000\ncurl: (56) CONNECT tunnel failed, response 403\n",
```
That's because the program uses the allowed_domain of the default network proxy and `www.example.com` is not in the allowed_domain.

### case 2
with the following permission:
```
permissions:
  network:
    enabled: true
    allowed_domains:
      - "www.example.com"
```

result:
```
<       "aggregatedOutput": "pyenv: cannot rehash: /Users/celia/.pyenv/shims isn't writable\n<!doctype html><html lang=\"en\"><head><title>Example Domain</title><meta name=\"viewport\" content=\"width=device-width, initial-scale=1\"><style>body{background:#eee;width:60vw;margin:15vh auto;font-family:system-ui,sans-serif}h1{font-size:1.5em}div{opacity:0.8}a:link,a:visited{color:#348}</style></head><body><div><h1>Example Domain</h1><p>This domain is for use in documentation examples without needing permission. Avoid use in operations.</p><p><a href=\"https://iana.org/domains/example\">Learn more</a></p></div></body></html>\n\nHTTP_STATUS:200\n",
```
fetched the website correctly.

### DIFF
--- a/codex-rs/core/src/codex_tests.rs
+++ b/codex-rs/core/src/codex_tests.rs
@@ -44,6 +44,7 @@ use crate::protocol::UserMessageEvent;
 use crate::rollout::policy::EventPersistenceMode;
 use crate::rollout::recorder::RolloutRecorder;
 use crate::rollout::recorder::RolloutRecorderParams;
+use crate::skill_network_proxy_cache::SkillNetworkProxyCache;
 use crate::state::TaskKind;
 use crate::tasks::SessionTask;
 use crate::tasks::SessionTaskContext;
@@ -2373,7 +2374,7 @@ async fn session_new_fails_when_zsh_fork_enabled_without_zsh_path() {
         plugins_manager,
         mcp_manager,
         Arc::new(FileWatcher::noop()),
-        Arc::new(crate::skill_network_proxy_cache::SkillNetworkProxyCache::new()),
+        Arc::new(SkillNetworkProxyCache::new()),
         AgentControl::default(),
     )
     .await;
@@ -2504,11 +2505,10 @@ pub(crate) async fn make_session_and_context() -> (Session, TurnContext) {
         mcp_manager,
         file_watcher,
         agent_control,
+        #[cfg(unix)]
         network_proxy_spec: None,
         network_proxy: None,
-        skill_network_proxy_cache: Arc::new(
-            crate::skill_network_proxy_cache::SkillNetworkProxyCache::new(),
-        ),
+        skill_network_proxy_cache: Arc::new(SkillNetworkProxyCache::new()),
         network_approval: Arc::clone(&network_approval),
         state_db: None,
         model_client: ModelClient::new(
@@ -3302,11 +3302,10 @@ pub(crate) async fn make_session_and_context_with_dynamic_tools_and_rx(
         mcp_manager,
         file_watcher,
         agent_control,
+        #[cfg(unix)]
         network_proxy_spec: None,
         network_proxy: None,
-        skill_network_proxy_cache: Arc::new(
-            crate::skill_network_proxy_cache::SkillNetworkProxyCache::new(),
-        ),
+        skill_network_proxy_cache: Arc::new(SkillNetworkProxyCache::new()),
         network_approval: Arc::clone(&network_approval),
         state_db: None,
         model_client: ModelClient::new(
--- a/codex-rs/core/src/codex_tests.rs
+++ b/codex-rs/core/src/codex_tests.rs
@@ -2373,6 +2373,7 @@ async fn session_new_fails_when_zsh_fork_enabled_without_zsh_path() {
         plugins_manager,
         mcp_manager,
         Arc::new(FileWatcher::noop()),
+        Arc::new(crate::skill_network_proxy_cache::SkillNetworkProxyCache::new()),
         AgentControl::default(),
     )
     .await;
@@ -2503,7 +2504,11 @@ pub(crate) async fn make_session_and_context() -> (Session, TurnContext) {
         mcp_manager,
         file_watcher,
         agent_control,
+        network_proxy_spec: None,
         network_proxy: None,
+        skill_network_proxy_cache: Arc::new(
+            crate::skill_network_proxy_cache::SkillNetworkProxyCache::new(),
+        ),
         network_approval: Arc::clone(&network_approval),
         state_db: None,
         model_client: ModelClient::new(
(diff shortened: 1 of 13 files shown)


codex-rs/core/src/config/network_proxy_spec.rs
```diff
@@ -84,6 +88,35 @@ impl NetworkProxySpec {
         self.config.network.enable_socks5
     }
 
+    #[cfg_attr(not(unix), allow(dead_code))]
```
bolinfest: Can we use `#[cfg(unix)]` instead for these methods?

codex-rs/core/src/config/network_proxy_spec.rs
```diff
@@ -84,6 +88,35 @@ impl NetworkProxySpec {
...
+    pub(crate) fn with_skill_managed_network_override(
+        &self,
+        managed_network_override: &SkillManagedNetworkOverride,
+    ) -> Self {
+        let mut spec = self.clone();
+        if let Some(allowed_domains) = managed_network_override.allowed_domains.as_ref() {
+            spec.config.network.allowed_domains = allowed_domains.clone();
```
bolinfest, line 100: This is why we really need this to be a map rather than a list.

codex-rs/core/src/config/network_proxy_spec.rs
```diff
@@ -20,6 +21,9 @@ use codex_protocol::protocol::SandboxPolicy;
 use std::collections::HashSet;
 use std::sync::Arc;
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub(crate) struct SkillNetworkProxyKey(String);
```
bolinfest, line 27: Why not just use a regular struct for this and rely on `#[derive(Eq, Hash)]`? Though you will need a constructor like `shared_skill_proxy_key()` that ensures the fields are sorted when this is constructed.

Or if you want to distill it down to a single value, why not use `sha2::Digest` instead of an arbitrarily long `String`?

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
```diff
@@ -846,6 +830,7 @@ fn commands_for_intercepted_exec_policy(
 }
 
 struct CoreShellCommandExecutor {
+    session: Option<Arc<crate::codex::Session>>,
```
bolinfest, line 833: When is this `None`?

codex-rs/core/src/tools/runtimes/shell/unix_escalation_tests.rs
```diff
@@ -651,6 +651,7 @@ host_executable(name = "git", paths = ["{allowed_git_literal}"])
 async fn prepare_escalated_exec_turn_default_preserves_macos_seatbelt_extensions() {
     let cwd = AbsolutePathBuf::from_absolute_path(std::env::temp_dir()).unwrap();
     let executor = CoreShellCommandExecutor {
+        session: None,
```
bolinfest, line 654: If it's only `None` in tests, then ideally we would use a dummy `Session` or something so the business logic doesn't have to deal with the `None` case.

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
```diff
@@ -1075,6 +1068,26 @@ impl CoreShellCommandExecutor {
...
+        let Some(session) = self.session.as_ref() else {
+            return Ok(self.network.clone());
+        };
+        let Some(skill) =
+            find_skill_for_program(session.as_ref(), self.sandbox_policy_cwd.as_path(), program)
+                .await
+        else {
+            return Ok(self.network.clone());
+        };
+        if let Some(network) = session.get_or_start_skill_network_proxy(&skill).await? {
```
bolinfest, line 1085: I think you can collapse these `if`s such that `Ok(self.network.clone())` only appears once?

codex-rs/core/src/codex_tests.rs
```diff
@@ -2373,6 +2373,7 @@ async fn session_new_fails_when_zsh_fork_enabled_without_zsh_path() {
         plugins_manager,
         mcp_manager,
         Arc::new(FileWatcher::noop()),
+        Arc::new(crate::skill_network_proxy_cache::SkillNetworkProxyCache::new()),
```
bolinfest: I would import `SkillNetworkProxyCache` at the top?

codex-rs/core/src/skill_network_proxy_cache.rs
```diff
@@ -0,0 +1,38 @@
+use crate::config::SkillNetworkProxyKey;
+use crate::config::StartedNetworkProxy;
+use std::collections::HashMap;
+use std::future::Future;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+#[derive(Default)]
+pub(crate) struct SkillNetworkProxyCache {
+    #[cfg_attr(not(unix), allow(dead_code))]
```
bolinfest: Why shouldn't this be available on Windows? I thought what we have is cross-platform?

codex-rs/core/src/skill_network_proxy_cache.rs
```diff
@@ -0,0 +1,38 @@
...
+    where
+        F: FnOnce() -> Fut,
+        Fut: Future<Output = anyhow::Result<StartedNetworkProxy>>,
+    {
+        let mut proxies = self.proxies.lock().await;
+        if let Some(proxy) = proxies.get(&key) {
+            return Ok(Arc::clone(proxy));
+        }
+
+        let proxy = Arc::new(start().await?);
```
bolinfest, line 40: Can we find a way to represent this so that we do not have to `await` while holding the mutex for `proxies`?

codex-rs/core/src/skill_network_proxy_cache.rs
celia-oai: this is not not available on windows, it's not called because only `unix_escalation` path hits it

PR #14419, opened 2026-03-12
use scopes_supported for OAuth when present on MCP servers

### LINKED ISSUE #8889: codex ignores scopes_supported from OAuth protected resource metadata

codex-cli 0.79.0

Enterprise

_No response_

Darwin 25.2.0 arm64 arm

I am adding my MCP server, which causes codex to trigger OIDC authentication, but the URL does not have `scopes even though my protected resource metadata includes `scopes_supported`

add the MCP Server https://mcp.stage.oort.io
see it triggers authentication without scopes.

add space delimited values of scopes_supported  to a `scope` query parameter in the URL

### Additional information

_No response_

jgershen-oai:
Fixes [#8889](https://github.com/openai/codex/issues/8889).

## Summary
- Discover and use advertised MCP OAuth `scopes_supported` when no explicit or configured scopes are present.
- Apply the same scope precedence across `mcp add`, `mcp login`, skill dependency auto-login, and app-server MCP OAuth login.
- Keep discovered scopes ephemeral and non-persistent.
- Retry once without scopes for CLI and skill auto-login flows if the OAuth provider rejects discovered scopes.

## Motivation
Some MCP servers advertise the scopes they expect clients to request during OAuth, but Codex was ignoring that metadata and typically starting OAuth with no scopes unless the user manually passed `--scopes` or configured `server.scopes`.

That made compliant MCP servers harder to use out of the box and is the behavior described in [#8889](https://github.com/openai/codex/issues/8889).

This change also brings our behavior in line with the MCP authorization spec's scope selection guidance:
https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#scope-selection-strategy

## Behavior
Scope selection now follows this order everywhere:
1. Explicit request scopes / CLI `--scopes`
2. Configured `server.scopes`
3. Discovered `scopes_supported`
4. Legacy empty-scope behavior

Compatibility notes:
- Existing working setups keep the same behavior because explicit and configured scopes still win.
- Discovered scopes are never written back into config or token storage.
- If discovery is missing, malformed, or empty, behavior falls back to the previous empty-scope path.
- App-server login gets the same precedence rules, but does not add a transparent retry path in this change.

## Implementation
- Extend streamable HTTP OAuth discovery to parse and normalize `scopes_supported`.
- Add a shared MCP scope resolver in `core` so all login entrypoints use the same precedence rules.
- Preserve provider callback errors from the OAuth flow so CLI/skill flows can safely distinguish provider rejections from other failures.
- Reuse discovered scopes from the existing OAuth support check where possible instead of persisting new config.


jgershen-oai:

codex-rs/cli/src/mcp_cmd.rs
```diff
@@ -183,6 +187,51 @@ impl McpCli {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
+async fn perform_oauth_login_with_legacy_retry(
```
gpeal: The name legacy retry is pretty unclear here. Should it be `perform_oauth_login_retry_without_scopes` or something? Why is it necessary to have that fallback? Docs would be helpful, I think

jgershen-oai: I wanted to keep the fallback to no-scope auth since that's how it works today, and I don't want this to be a breaking change for anyone. Good call on the name; added some docs

gpeal (approved):

PR #14426, opened 2026-03-12
Decouple request permissions feature and tool

mousseau-oai:
Update behavior so that `reject { request_permissions: true, ... }` rejects the `request_permissions` tool and that `reject { sandbox_approval: true, ... }` rejects `with_additional_permissions`.


dylan-hurd-oai (approved):
Thanks for the rev!

PR #14427, opened 2026-03-12
feat: refactor on openai-curated plugins.

xl-openai:
- Curated repo sync now uses GitHub HTTP, not local git.
- Curated plugin cache/versioning now uses commit SHA instead of local.
- Startup sync now always repairs or refreshes curated plugin cache from tmp (auto update to the lastest)

### DIFF
--- a/codex-rs/core/src/plugins/manager.rs
+++ b/codex-rs/core/src/plugins/manager.rs
@@ -11,6 +11,7 @@ use super::marketplace::load_marketplace_summary;
 use super::marketplace::resolve_marketplace_plugin;
 use super::plugin_manifest_name;
 use super::plugin_manifest_paths;
+use super::read_curated_plugins_sha;
 use super::store::DEFAULT_PLUGIN_VERSION;
 use super::store::PluginId;
 use super::store::PluginIdError;
@@ -45,6 +46,7 @@ use std::collections::HashSet;
 use std::fs;
 use std::path::Path;
 use std::path::PathBuf;
+use std::sync::Arc;
 use std::sync::RwLock;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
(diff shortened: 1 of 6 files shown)


gverma-openai (approved):

codex-rs/core/src/plugins/manager.rs
```diff
@@ -670,26 +706,56 @@ impl PluginsManager {
...
-            })
+            .spawn(
+                move || match sync_openai_plugins_repo(codex_home.as_path()) {
+                    Ok(curated_plugin_version) => {
+                        if let Err(err) = refresh_curated_plugin_cache(
+                            codex_home.as_path(),
+                            &curated_plugin_version,
+                            &configured_curated_plugin_ids,
+                        ) {
+                            warn!("failed to refresh curated plugin cache after sync: {err}");
```
sayan-oai: do we need to reset CURATED_REPO_SYNC_STARTED in this case? otherwise its never reset

codex-rs/core/src/plugins/manager.rs
```diff
@@ -670,26 +706,56 @@ impl PluginsManager {
...
             .name("plugins-curated-repo-sync".to_string())
-            .spawn(move || {
-                if let Err(err) = sync_openai_plugins_repo(codex_home.as_path()) {
-                    CURATED_REPO_SYNC_STARTED.store(false, Ordering::SeqCst);
-                    warn!("failed to sync curated plugins repo: {err}");
-                }
-            })
+            .spawn(
+                move || match sync_openai_plugins_repo(codex_home.as_path()) {
+                    Ok(curated_plugin_version) => {
```
sayan-oai, line 745: i think `plugins_for_config` serves a cached value. should we invalidate the cache when the sync completes?

sayan-oai (approved):
approved, two small questions

PR #14430, opened 2026-03-12
[telemetry] emittance

rhan-oai:
- metadata emission for user message type
- wiring for ResponseItem

Testing
- tests prompt, prompt queue and prompt steering
`INFO     | 2026-03-12 15:12:27 | codex_backend.routers.responses | responses.create_response:459 | Responses input metadata probe for session 019ce41b-a59d-7e61-ad3d-052b69082252: [{'index': 2, 'type': 'message', 'role': 'user', 'user_message_type': 'prompt', 'metadata_keys': ['user_message_type']}] | `


codex-rs/core/src/codex.rs
```diff
@@ -692,6 +694,20 @@ impl Codex {
...
     }
 
+    /// Use sparingly: wrapper for integration tests and callers that only hold a
+    /// `Codex` handle.
+    ///
+    /// This keeps `Session` internals encapsulated while still allowing queued
+    /// response-input behavior to be exercised from public APIs.
+    /// Returns the input unchanged when there is no active turn.
+    #[doc(hidden)]
+    pub async fn inject_response_items(
```
dylan-hurd-oai: I don't think we need this API - turn_steer_returns_active_turn_id tests mid-turn steering without this, can we just do something similar? We should be certain it's absolutely required

codex-rs/core/src/codex.rs
```diff
@@ -3658,13 +3675,49 @@ impl Session {
...
+                },
+                Some(kind),
+            ) if role == "user" => {
+                let mut metadata = metadata.unwrap_or_default();
+                metadata.user_message_type = Some(kind);
+                ResponseItem::Message {
+                    id,
+                    role,
+                    content,
+                    metadata: Some(metadata),
```
dylan-hurd-oai: I don't think we need to add metadata on a `TurnItem::UserMessage`, do we?

We construct the request to the [client from the conversation history](https://github.com/openai/codex/blob/39a9f6f5395e5e2f77f83248adcb2ea3002b0149/codex-rs/core/src/codex.rs#L5663-L5668) which uses response items and are stored below this item.

It feels like it will be much cleaner to only attach metadata to `ResponseInputItem` and `ResponseItem`, so we can avoid exposing them in app-server.

cc @owenlin0 for some app-server opinions as well

dylan-hurd-oai (changes requested):
I'm a bit confused about how we're organizing metadata!

codex-rs/core/src/codex.rs
rhan-oai: this is used to test queueing, not steering, but I agree that it'd be nice to have a different option. haven't figured out a faithful integration test for queueing without using the wrappers, would be good to discuss

codex-rs/core/src/codex.rs
```diff
@@ -1000,6 +1001,18 @@ fn local_time_context() -> (String, String) {
...
+    let ResponseInputItem::Message { role, metadata, .. } = item else {
+        return;
+    };
+    if role != "user" {
+        return;
+    }
+    let mut metadata_value = metadata.take().unwrap_or_default();
+    metadata_value.user_message_type = Some(kind);
+    *metadata = Some(metadata_value);
+}
```
dylan-hurd-oai: Let's move this into a separate file please 🙏

codex-rs/core/src/state/turn.rs
```diff
@@ -56,6 +57,12 @@ pub(crate) struct RunningTask {
     pub(crate) _timer: Option<codex_otel::Timer>,
 }
 
+#[derive(Debug, Clone)]
+pub(crate) struct PendingInputItem {
+    pub(crate) input: ResponseInputItem,
+    pub(crate) user_message_type: Option<UserMessageType>,
```
dylan-hurd-oai: @owenlin0 for thoughts here - should we just add metadata to ResponseInputItem?

codex-rs/core/src/codex.rs
```diff
@@ -5595,8 +5671,19 @@ pub(crate) async fn run_turn(
...
+    let mut initial_input_for_turn: ResponseInputItem = ResponseInputItem::from(input.clone());
+    if sess.enabled(Feature::ItemMetadata) {
+        stamp_user_message_type_on_input_item(&mut initial_input_for_turn, UserMessageType::Prompt);
+    }
+    let response_item: ResponseItem = initial_input_for_turn.clone().into();
+    sess.record_user_prompt_and_emit_turn_item(
+        turn_context.as_ref(),
+        &input,
+        response_item,
+        Some(UserMessageType::Prompt),
```
dylan-hurd-oai: We've already attached the metadata above - why do we need to add it to the method here?

dylan-hurd-oai (commented):
I think something in lifecycle of ResponseInputItem > ResponseItem and stamping metadata onto the object is not adding up here for me. Let's think about how to make this cleaner - e.g. how can we structure this to avoid unnecessary repeated stampings? feels like we should align this more closely with the creation of ResponseItem.

PR #14436, opened 2026-03-12
Refactor code mode worker dispatch

pakrym-oai:
## Summary
- start a code mode worker once per turn and let it pump nested tool calls through a dedicated queue
- simplify code mode request/response dispatch around request ids and generic runner-unavailable errors
- clean up the code mode process API and runner protocol plumbing

## Testing
- cargo test -p codex-core (started, then interrupted before completion)


pakrym-oai:
Superseded by #14437 from a fresh branch based on main.

PR #14437, opened 2026-03-12
Dispatch tools when code mode is not awaited directly

pakrym-oai:
## Summary
- start a code mode worker once per turn and let it pump nested tool calls through a dedicated queue
- simplify code mode request/response dispatch around request ids and generic runner-unavailable errors
- clean up the code mode process API and runner protocol plumbing

## Testing
- not run yet

### DIFF
--- a/codex-rs/core/src/tools/code_mode.rs
+++ b/codex-rs/core/src/tools/code_mode.rs
@@ -1,5 +1,4 @@
 use std::collections::HashMap;
-use std::collections::VecDeque;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
@@ -19,7 +18,6 @@ use crate::tools::context::ToolPayload;
 use crate::tools::js_repl::resolve_compatible_node;
 use crate::tools::router::ToolCall;
 use crate::tools::router::ToolCallSource;
-use crate::tools::router::ToolRouterParams;
 use crate::truncate::TruncationPolicy;
 use crate::truncate::formatted_truncate_text_content_items_with_policy;
 use crate::truncate::truncate_function_output_items_with_policy;
@@ -33,6 +31,8 @@ use tokio::io::AsyncReadExt;
 use tokio::io::AsyncWriteExt;
 use tokio::io::BufReader;
 use tokio::sync::Mutex;
+use tokio::sync::mpsc;
+use tokio::sync::oneshot;
 use tokio::task::JoinHandle;
 use tracing::warn;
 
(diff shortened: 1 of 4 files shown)


codex-rs/core/src/tools/code_mode.rs
```diff
@@ -278,7 +345,7 @@ pub(crate) fn instructions(config: &Config) -> Option<String> {
...
...penai/code_mode\"`). `output_text(value)` surfaces text back to the model and stringifies non-string objects with `JSON.stringify(...)` when possible. `output_image(imageUrl)` appends an `input_image` content item for `http(s)` or `data:` URLs. `store(key, value)` persists JSON-serializable values across `{PUBLIC_TOOL_NAME}` calls in the current session, and `load(key)` returns a cloned stored value or `undefined`. `set_max_output_tokens_per_exec_call(value)` sets the token budget used to truncate direct `{PUBLIC_TOOL_NAME}` returns; `{WAIT_TOOL_NAME}` uses its own `max_tokens` argument instead and defaults to `10000`. `set_yield_time(value)` asks `{PUBLIC_TOOL_NAME}` to return early if the script is still running after that many milliseconds so `{WAIT_TOOL_NAME}` can resume it later. `yield_control()` returns a yielded `{PUBLIC_TOOL_NAME}` response immediately while the script keeps running in the background. The returned content starts with a separate `Script completed`, `Script failed`, or `Script running with session ID …` text item that includes wall time. When truncation happens, the final text may include `Total output lines:` and the usual `…N tokens truncated…` marker.\n",
```
cconger: I'm realizing that `yield` doesn't actually feel quite right here, its more about "detaching" focus or backgrounding.  The yield is actually just fire and forget and has no impact on the control flow in repl.  This is all the more apparent when we're allowing tool calls to run backgrounded too.

cconger (approved):
Naming nit, curious your thoughts.

codex-rs/core/src/tools/code_mode.rs
```diff
@@ -51,80 +52,110 @@ struct ExecContext {
...
+                };
+                let response = HostToNodeMessage::Response {
+                    request_id: tool_call.request_id,
+                    id: tool_call.id,
+                    code_mode_result: call_nested_tool(
+                        exec.clone(),
+                        tool_call.name,
+                        tool_call.input,
+                    )
+                    .await,
```
cconger: We actually have head of line blocking here in this loop.  Since our tool calls are async its possible that multiple could be in flight from js at a time, and here we are serially executing so a slow one blocks the rest.

cconger: Its even more exacerbated by this being shared across the process.

pakrym-oai: I agree but one step at a time. Want to keep individual changes small.

PR #14445, opened 2026-03-12
feat: add plugin/read.

xl-openai:
return more information for a specific plugin.


codex-rs/core/src/plugins/manager.rs
```diff
@@ -705,6 +725,89 @@ impl PluginsManager {
...
+            |err| match err {
+                PluginIdError::Invalid(message) => MarketplaceError::InvalidPlugin(message),
+            },
+        )?;
+        let plugin_key = plugin_id.as_key();
+        let (installed_plugins, configured_plugins) = self.configured_plugin_states(config);
+        let source_path = match &plugin.source {
+            MarketplacePluginSourceSummary::Local { path } => path.clone(),
+        };
+        let manifest = load_plugin_manifest(source_path.as_path());
```
sayan-oai: i think at runtime we error if loading the manifest fails, instead of falling back to defaults. is there a reason we want this behavior to be different in `plugin/read`?

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -5252,6 +5263,44 @@ impl CodexMessageProcessor {
...
+                    .await
+                    .unwrap_or_default()
+            }
+        };
+        let plugin_app_ids = plugin_apps
+            .iter()
+            .map(|connector_id| connector_id.0.as_str())
+            .collect::<HashSet<_>>();
+
+        filter_disallowed_connectors(merge_plugin_apps(connectors, plugin_apps.to_vec()))
```
sayan-oai: it seems weird that we have to know this filter + merge pattern in CodexMessageProcessor. i see we do something similar for CodexMessageProcessor::plugin_install as well. should this method become a helper in connectors:: that can be used by plugin_read and plugin_install?

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -5252,6 +5263,44 @@ impl CodexMessageProcessor {
         connectors::merge_connectors_with_accessible(all, accessible, all_connectors_loaded)
     }
 
+    async fn load_plugin_app_summaries(
```
sayan-oai: name is a bit misleading; this would work for any apps. we dont really use plugin_id. see point below about making this a generic helper.

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3092,6 +3107,47 @@ pub struct PluginSummary {
...
+    pub description: String,
+    pub short_description: Option<String>,
+    pub interface: Option<SkillInterface>,
+    pub path: PathBuf,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct PluginAppSummary {
```
sayan-oai: why not just use `AppSummary`?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3092,6 +3107,47 @@ pub struct PluginSummary {
...
+    pub description: Option<String>,
+    pub install_url: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct PluginMcpServerSummary {
+    pub name: String,
+}
```
sayan-oai: optional nit: could we just have a vec of McpServerNames instead of a struct with one string? i know everything else is `*Summary` though, so up to you.

codex-rs/app-server-protocol/schema/typescript/v2/PluginSkillSummary.ts
```diff
@@ -0,0 +1,6 @@
+// GENERATED CODE! DO NOT MODIFY BY HAND!
+
+// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
+import type { SkillInterface } from "./SkillInterface";
+
+export type PluginSkillSummary = { name: string, description: string, shortDescription: string | null, interface: SkillInterface | null, path: string, };
```
sayan-oai: nit: can this just be `SkillSummary`? i dont think its plugin specific.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -5252,6 +5259,30 @@ impl CodexMessageProcessor {
...
```
sayan-oai: another nit: should this be a method of `CodexMessageProcesor`, or just a private helper? wdyt?

codex-rs/core/src/plugins/manager.rs
```diff
@@ -705,6 +725,89 @@ impl PluginsManager {
...
+        Ok(PluginReadOutcome {
+            marketplace_name: marketplace.name,
+            marketplace_path: marketplace.path,
+            plugin: PluginDetailSummary {
+                id: plugin_key.clone(),
+                name: plugin.name,
+                description,
+                source: plugin.source,
+                install_policy: plugin.install_policy,
+                auth_policy: plugin.auth_policy,
```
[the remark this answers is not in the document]
sayan-oai: im assuming this is wrong lol (referring to codex review comment saying it wont compile)

sayan-oai (approved):
thanks for cleaning up. a couple of optional nits, otherwise lgtm

PR #14446, opened 2026-03-12
Add Python SDK public API and examples

shaqayeq-oai:
## TL;DR
WIP esp the examples

  Thin the Python SDK public surface so the wrapper layer returns canonical app-server generated models directly.

  - keeps `Codex` / `AsyncCodex` / `Thread` / `Turn` and input helpers, but removes alias-only type layers and custom result models
  - `metadata` now returns `InitializeResponse` and `run()` returns the generated app-server `Turn`
  - updates docs, examples, notebook, and tests to use canonical generated types and regenerates `v2_all.py` against current schema
  - keeps the pinned runtime-package integration flow and real integration coverage

  ## Validation
  - `PYTHONPATH=sdk/python/src python3 -m pytest sdk/python/tests`
  - `GH_TOKEN="$(gh auth token)" RUN_REAL_CODEX_TESTS=1 PYTHONPATH=sdk/python/src python3 -m pytest sdk/python/tests -rs`


sdk/python/src/codex_app_server/generated/v2_types.py
```diff
@@ -1,25 +1,23 @@
-"""Stable aliases over full v2 autogenerated models (datamodel-code-generator)."""
+"""Stable aliases over the canonical generated v2 models."""
```
owenlin0: Do we need to have this file since we are already trying to keep the app-server API stable and backwards compatible generally? Could we simplify & remove?

sdk/python/src/codex_app_server/public_types.py
```diff
@@ -0,0 +1,41 @@
+"""Shallow public aliases over the generated v2 wire models."""
```
owenlin0: Can also remove this too

owenlin0: I think we should consider all generated types of app-server API "public"

sdk/python/src/codex_app_server/public_api.py
```diff
@@ -0,0 +1,795 @@
...
+)
+
+
+@dataclass(slots=True)
+class TurnResult:
+    thread_id: str
+    turn_id: str
+    status: TurnStatus
+    error: TurnError | None
+    text: str
```
owenlin0: Instead of defining our own types, wherever possible it'd be great to just use the generated Python types directly. So instead of defining a `TurnResult`, could we just use the auto-generated `Turn` object? this would mean no `text` and `usage` fields, which I think is actually good IMO

my thinking is that we take a lot of thought in developing the app-server API and it should be the canonical interface, so the more we expose it directly the better

owenlin0: Same goes for the various types below

owenlin0 (commented):
this is a great start! just some initial thoughts about keeping the SDK as thin a wrapper around app-server as possible. will have to think about the public interface a bit more

sdk/python/src/codex_app_server/api.py
```diff
@@ -0,0 +1,685 @@
...
+        await self._codex._ensure_initialized()
+        return await self._codex._client.thread_set_name(self.id, name)
+
+    async def compact(self) -> ThreadCompactStartResponse:
+        await self._codex._ensure_initialized()
+        return await self._codex._client.thread_compact(self.id)
+
+
+@dataclass(slots=True)
+class TurnHandle:
```
shaqayeq-oai, line 610: WIP
I kept the wrapper layer for now since it’s still where the per-turn control methods live, but renamed it to `TurnHandle/AsyncTurnHandle` so we don’t have two different public things both called `Turn`. The generated app-server `Turn` stays the canonical returned type, and this wrapper is just the handle around that flow but lmk if we would rather do it differently.

owenlin0 (approved):
this looks quite nice! I think we can layer on some nice convenience methods (borrowed from TS SDK) that we can do as a followup to make the common case easy:
- Add a convenience `Thread.run()` / `AsyncThread.run()` method that returns the final assistant response and collected items.
- Have those methods accept plain strings as input instead of having to construct `TextInput`.

(in addition to the lower-level `thread.turn(...)` method that returns a turn handle, if the client needs to steer/interrupt/etc.)

owenlin0 (commented):
naming nit: `public_api.py` -> `api.py`?

PR #14473, opened 2026-03-12
fix(cli): support legacy use_linux_sandbox_bwrap flag

viyatb-oai:
## Summary
- restore `use_linux_sandbox_bwrap` as a removed feature key so older `--enable` callers parse again
- keep it as a no-op by leaving runtime behavior unchanged
- add regression coverage for the legacy `--enable` path

## Testing
- Not run (updated and pushed quickly)


bolinfest (approved):
Thanks for the quick fix!

PR #14491, opened 2026-03-12
Fix MCP tool calling

pakrym-oai:
Properly escape mcp tool names and make tools only available via imports.


codex-rs/core/tests/suite/code_mode.rs
```diff
@@ -1438,6 +1438,182 @@ contentLength=0"
...
+    let (output, success) = custom_tool_output_body_and_success(&req, "call-1");
+    assert_ne!(
+        success,
+        Some(false),
+        "exec global scope inspection failed unexpectedly: {output}"
+    );
+    assert_eq!(
+        serde_json::from_str::<Vec<String>>(&output)?,
+        vec![
+            "AggregateError",
```
pakrym-oai: check that we are not leaking extra stuff

cconger (approved):

PR #14501, opened 2026-03-12
dynamic tool calls: add param `exposeToContext` to optionally hide tool

cconger:
This extends dynamic_tool_calls to allow us to hide a tool from the model context but still use it as part of the general tool calling runtime (for ex from js_repl/code_mode)


codex-rs/core/src/codex.rs
```diff
@@ -6108,9 +6108,25 @@ fn build_prompt(
     turn_context: &TurnContext,
     base_instructions: BaseInstructions,
 ) -> Prompt {
+    let hidden_dynamic_tools = turn_context
+        .dynamic_tools
+        .iter()
+        .filter(|tool| !tool.expose_to_context)
```
pakrym-oai: we call this `defer_loading` elsewhere should we go with the same name?

codex-rs/core/src/codex.rs
```diff
@@ -6108,9 +6108,25 @@ fn build_prompt(
...
+        .map(|tool| tool.name.as_str())
+        .collect::<HashSet<_>>();
+    let tools = if hidden_dynamic_tools.is_empty() {
```
pakrym-oai: do we need to filter them out or just set defer_loading on them?

codex-rs/state/src/runtime/threads.rs
```diff
@@ -435,6 +437,7 @@ ON CONFLICT(thread_id, position) DO NOTHING
             .bind(tool.name.as_str())
             .bind(tool.description.as_str())
             .bind(input_schema)
+            .bind(tool.expose_to_context)
```
pakrym-oai: what happens when a model decides to run a stored tool that isn't registered anymore?

pakrym-oai (approved):
Consider renaming to `defer_loading`

codex-rs/core/src/codex.rs
cconger: They're not callable as remote tools so I would prefer not to even expose them to responses but maybe I'm missing something here.

codex-rs/state/src/runtime/threads.rs
cconger: Yeah I think this is an existing bug/problem with dynamic tool calls either way right now.  Without a registry with lifecycles we're just trusting that there's a listener for the dynamic tool call for the resumer too.

cconger:
@codex review

PR #14511, opened 2026-03-12
code_mode: Move exec params from runtime declarations to @pragma

cconger:
This change moves code_mode exec session settings out of the runtime API and into an optional first-line pragma, so instead of calling  runtime helpers like set_yield_time() or set_max_output_tokens_per_exec_call(), the model can write // @exec: {"yield_time_ms": ...,  "max_output_tokens": ...} at the top of the freeform exec source. Rust now parses that pragma before building the source, validates  it, and passes the values directly in the exec start message to the code-mode broker, which applies them at session start without any worker-runtime mutation path. The @openai/code_mode module no longer exposes those setter functions, the docs and grammar were updated to describe the pragma form, and the existing code_mode tests were converted to use pragma-based configuration instead.

### DIFF
--- a/codex-rs/core/src/tools/code_mode/runner.cjs
+++ b/codex-rs/core/src/tools/code_mode/runner.cjs
@@ -115,17 +115,27 @@ function codeModeWorkerMain() {
     return contentItems;
   }
 
-  function createToolsNamespace(callTool, enabledTools) {
+  function createGlobalToolsNamespace(callTool, enabledTools) {
     const tools = Object.create(null);
 
-    for (const tool of enabledTools) {
-      const toolNamespace = Array.isArray(tool.namespace) ? tool.namespace : [];
-      if (toolNamespace.length !== 0) {
-        continue;
-      }
+    for (const { tool_name, global_name } of enabledTools) {
+      Object.defineProperty(tools, global_name, {
+        value: async (args) => callTool(tool_name, args),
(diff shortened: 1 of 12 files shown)


codex-rs/core/src/tools/code_mode/runner.cjs
```diff
@@ -587,14 +533,18 @@ function sessionWorkerSource() {
...
+      : normalizeMaxOutputTokensPerExecCall(start.max_output_tokens);
   const session = {
     completed: false,
     content_items: [],
     default_yield_time_ms: normalizeYieldTime(start.default_yield_time_ms),
     id: start.session_id,
     initial_yield_timer: null,
     initial_yield_triggered: false,
-    max_output_tokens_per_exec_call: DEFAULT_MAX_OUTPUT_TOKENS_PER_EXEC_CALL,
+    max_output_tokens_per_exec_call: maxOutputTokensPerExecCall,
```
pakrym-oai, line 611: does JS need to know about this value?

pakrym-oai (approved):
awesome!

codex-rs/core/src/tools/code_mode/runner.cjs
cconger: Currently disconnecting is managed in the cjs level so I think when we implement this I would expect it to be done in the top level js context.

PR #14516, opened 2026-03-12
Rename reject approval policy to granular

mousseau-oai:
`reject` approval policy is now called `granular` where each property indicates if approval is *allowed*. Renamed `Feature::RequestPermissions` to `Feature::ExecPermissionApprovals` as the original name became confusing with the `request_permissions` tool.


dylan-hurd-oai (approved):
The scope of the rename looks great! My only concern is the breaking change to `AskForApproval` - users that have manually edited their config.toml to use this field will start seeing an error, and we never marked it as experimental. 

I _think_ usage/adoption would be low enough that we're okay with this, but in the future we should be clearer about marking this as experimental ahead of time.

PR #14523, opened 2026-03-12
tui: add ctrl-x composer clear

pc-oai:
## Summary
- add a dedicated `Ctrl-X` shortcut to clear the composer without affecting transcript or quit behavior
- show the new shortcut in the footer shortcut overlay
- add a regression test and update the affected snapshots

## Testing
- `cargo test -p codex-tui ctrl_x_clears_composer_without_emitting_ops -- --exact`
- `cargo test -p codex-tui footer -- --nocapture`


etraut-openai:
We already support Ctrl+C for this purpose. I don't think we need another key binding for the same action.

PR #14525, opened 2026-03-12
feature: add granular builtin tool enablement 

ashwinnathan-openai:
## Why
This PR adds a new config surface for controlling built-in tool availability from `config.tools` and from per-thread config overrides.

The final design uses grouped feature tables under `[tools.<feature>]`. This composes cleanly with layered TOML, keeps per-feature config colocated with enablement, and gives SDK/app-server callers an explicit opt-in mode when they want to start from no built-in tools.

## What Changed
- Added grouped tool controls for:
  - `shell`
  - `filesystem`
  - `javascript`
  - `agents`
  - `agent_jobs`
  - `planning`
  - `user_input`
  - `web_search`
  - `image_generation`
  - `document_generation`
- Added `tools.disable_defaults` as the top-level switch for explicit grouped mode.
- Added per-feature wrapper types in config/app-server protocol:
  - generic `ToolFeature*` wrappers with `enabled`
  - `WebSearchFeature*` wrappers with `enabled` plus existing web-search config
- Updated tool gating, config readback, app-server protocol types, generated schemas, and tests to use the grouped feature model.
- Preserved shipped legacy config behavior in default mode, including:
  - legacy `tools.view_image`
  - top-level `web_search = "cached" | "live" | "disabled"`
  - legacy `[tools] web_search = true|false` remains accepted and ignored, and the schema reflects that

## Invariants
- If `tools.disable_defaults` is omitted or `false`, Codex preserves the existing/default tool behavior and shipped legacy fallback knobs continue to participate.
- If `tools.disable_defaults = true`, grouped `[tools.<feature>]` entries are the only source of truth for built-in tool exposure.
- In explicit grouped mode, legacy standalone knobs do not re-enable tools.
- For `web_search`, top-level mode config controls `Cached` / `Live` / `Disabled`, but exposure still comes from `[tools.web_search]` when `disable_defaults = true`.

## Testing
- `just write-config-schema`
- `just fmt`
- `cargo test -p codex-core`
- `cargo test`


codex-rs/core/src/tools/sandboxing.rs
```diff
@@ -157,6 +157,19 @@ impl ExecApprovalRequirement {
             _ => None,
         }
     }
+
+    pub fn force_manual_approval(self) -> Self {
+        match self {
+            Self::Skip {
+                proposed_execpolicy_amendment,
+                ..
+            } => Self::NeedsApproval {
```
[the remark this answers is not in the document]
ashwinnathan-openai: This is known and valid, but to limit scope in this PR we will be handling this in a followup PR where we handle [tool call approvals](https://www.notion.so/openai/App-Server-Review-dynamic-tool-call-approval-3218e50b62b080c88459da9528d753c3) more generally.

codex-rs/app-server-protocol/schema/json/v2/ConfigReadResponse.json
```diff
@@ -771,7 +879,7 @@
         "web_search": {
           "anyOf": [
             {
-              "$ref": "#/definitions/WebSearchToolConfig"
+              "$ref": "#/definitions/WebSearchFeatureConfigV2"
```
ashwinnathan-openai: This new definition includes the `enabled` field in `[tools.web_search]`

codex-rs/core/src/tools/spec.rs
```diff
@@ -461,6 +485,67 @@ impl ToolsConfig {
...
+
+    pub fn is_builtin_tool_invocation_enabled(&self, tool_name: &str) -> bool {
+        let Some(tool) = builtin_tool_key(tool_name) else {
+            return false;
+        };
+        let Some(feature) = ToolFeatureKey::for_builtin_tool(tool) else {
+            return true;
+        };
+        if self.is_feature_explicitly_controlled(feature) {
+            return self.is_tool_feature_enabled(feature);
```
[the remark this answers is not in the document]
ashwinnathan-openai: The contract is that when `disable_defaults=true`, we will rely on the new enablement surface of `[tools.]` instead of legacy logic.

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -529,11 +528,38 @@ pub struct SandboxWorkspaceWrite {
     pub exclude_slash_tmp: bool,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "snake_case")]
+#[ts(export_to = "v2/")]
+pub struct ToolFeatureConfigV2 {
```
bolinfest: Why the `V2` suffix?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -529,11 +528,38 @@ pub struct SandboxWorkspaceWrite {
...
+pub struct WebSearchFeatureConfigV2 {
+    pub enabled: Option<bool>,
+    #[serde(flatten)]
+    pub config: WebSearchToolConfig,
+}
+
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
 #[serde(rename_all = "snake_case")]
 #[ts(export_to = "v2/")]
 pub struct ToolsV2 {
```
bolinfest, line 555: @celia-oai @owenlin0 there's no need for `V2` on type names, is there?

codex-rs/app-server/README.md
```diff
@@ -114,10 +114,7 @@ Example with notification opt-out:
     },
     "capabilities": {
       "experimentalApi": true,
-      "optOutNotificationMethods": [
-        "thread/started",
-        "item/agentMessage/delta"
-      ]
+      "optOutNotificationMethods": ["thread/started", "item/agentMessage/delta"]
```
bolinfest, line 118: Was this file autoformatted? I think @owenlin0 had it the other way so it was fewer lines and somewhat faster to read?

codex-rs/core/src/config/mod.rs
```diff
@@ -581,6 +570,38 @@ pub struct Config {
...
+    pub shell: Option<bool>,
+    pub filesystem: Option<bool>,
+    pub javascript: Option<bool>,
+    pub agents: Option<bool>,
+    pub agent_jobs: Option<bool>,
+    pub planning: Option<bool>,
+    pub user_input: Option<bool>,
+    pub web_search: Option<bool>,
+    pub image_generation: Option<bool>,
+    pub document_generation: Option<bool>,
```
bolinfest, line 622: I feel like these should have their own types with the associated configuration specific to each tool. `None` can mean "not enabled" so the corresponding configuration struct doesn't have to have its own `enabled` field perhaps?

codex-rs/core/src/config/mod.rs
```diff
@@ -1578,42 +1605,82 @@ pub struct RealtimeAudioToml {
...
+    fn is_enabled(&self) -> bool {
+        self.enabled.unwrap_or(true)
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq, Eq, JsonSchema)]
 #[schemars(deny_unknown_fields)]
 pub struct ToolsToml {
+    pub disable_defaults: Option<bool>,
+    pub shell: Option<ToolFeatureToml>,
```
bolinfest, line 1676: Likewise, this should have the specific TOML shape we want to parse?

codex-rs/core/src/tools/registry.rs
```diff
@@ -20,9 +21,189 @@ use codex_hooks::HookToolInput;
 use codex_hooks::HookToolInputLocalShell;
 use codex_hooks::HookToolKind;
 use codex_protocol::models::ResponseInputItem;
+use codex_protocol::models::VIEW_IMAGE_TOOL_NAME;
 use codex_utils_readiness::Readiness;
 use tracing::warn;
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub(crate) enum BuiltinToolKey {
+    CodeMode,
```
bolinfest, line 42: alpha sort?

codex-rs/core/src/tools/registry.rs
```diff
@@ -20,9 +21,189 @@ use codex_hooks::HookToolInput;
 use codex_hooks::HookToolInputLocalShell;
 use codex_hooks::HookToolKind;
 use codex_protocol::models::ResponseInputItem;
+use codex_protocol::models::VIEW_IMAGE_TOOL_NAME;
 use codex_utils_readiness::Readiness;
 use tracing::warn;
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub(crate) enum BuiltinToolKey {
```
bolinfest, line 38: docstring on this type and maybe some of the others?

codex-rs/core/src/tools/registry.rs
```diff
@@ -20,9 +21,189 @@ use codex_hooks::HookToolInput;
...
+
+    pub(crate) fn iter() -> impl Iterator<Item = Self> {
+        Self::ALL.into_iter()
+    }
+
+    pub(crate) const fn invocation_names(self) -> &'static [&'static str] {
+        match self {
+            Self::CodeMode => &[PUBLIC_TOOL_NAME],
+            Self::ExecCommand => &[
+                "exec_command",
```
bolinfest: Hmm, should move of these be `const` shared with the other code under `core/src/tools/`?

codex-rs/core/src/tools/spec.rs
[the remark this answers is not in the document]
bolinfest: Maybe put that in a comment so Codex doesn't make the same comment in the future (hopefully)?

codex-rs/core/src/tools/spec.rs
```diff
@@ -2606,57 +2730,66 @@ pub(crate) fn build_specs_with_discoverable_tools(
 
     match &config.shell_type {
         ConfigShellToolType::Default => {
-            push_tool_spec(
+            push_builtin_tool_spec_if_enabled(
                 &mut builder,
+                config,
+                "exec_command",
```
bolinfest: Hmm, is `exec_command` correct here?

codex-rs/core/src/tools/spec.rs
```diff
@@ -2606,57 +2730,66 @@ pub(crate) fn build_specs_with_discoverable_tools(
...
                 ),
                 /*supports_parallel_tool_calls*/ true,
-                config.code_mode_enabled,
             );
         }
     }
 
-    if config.shell_type != ConfigShellToolType::Disabled {
+    if config.shell_type != ConfigShellToolType::Disabled
+        && config.is_builtin_tool_invocation_enabled("exec_command")
```
bolinfest: Hmm, this thing where `exec_command` is the marker for all shell commands doesn't sit well with me...

codex-rs/core/src/tools/spec.rs
```diff
@@ -2267,6 +2352,45 @@ fn push_tool_spec(
...
+    }
+}
+
+fn push_builtin_tool_with_handler_if_enabled<H>(
+    builder: &mut ToolRegistryBuilder,
+    config: &ToolsConfig,
+    invocation_name: &str,
+    spec: ToolSpec,
+    supports_parallel_tool_calls: bool,
+    handler_name: &str,
```
bolinfest: When is `handler_name` different from `invocation_name`?

codex-rs/core/src/config/mod.rs
bolinfest: @pakrym-oai I'm particularly interested to get your thoughts on this bit!

codex-rs/core/src/tools/spec.rs
```diff
@@ -2606,57 +2730,66 @@ pub(crate) fn build_specs_with_discoverable_tools(
...
                 /*supports_parallel_tool_calls*/ true,
-                config.code_mode_enabled,
             );
         }
         ConfigShellToolType::Local => {
-            push_tool_spec(
+            push_builtin_tool_spec_if_enabled(
                 &mut builder,
+                config,
+                "local_shell",
```
bolinfest: I also don't understand how this and others are not causing https://github.com/openai/codex/pull/14652 to fail....

codex-rs/app-server-protocol/src/protocol/v2.rs
ashwinnathan-openai: Unnecessary - I think codex saw v2 for other structs and assumed there would be a collision.

codex-rs/app-server-protocol/src/protocol/v2.rs
ashwinnathan-openai: I *think* for this one we already export Tools from protocol::v1 so we need to disambiguate?

codex-rs/core/src/tools/spec.rs
ashwinnathan-openai: It's not updated to remove

codex-rs/core/src/tools/spec.rs
ashwinnathan-openai: hmm perhaps `shell` would be better then

codex-rs/core/src/tools/spec.rs
ashwinnathan-openai: perhaps `shell`?

codex-rs/core/src/config/mod.rs
ashwinnathan-openai: As in create a specific shape for each tool like `ShellFeatureToml`? For now I just kept it generic as `ToolFeatureToml` which has `enabled` and if we end up adding feature specific configuration we should be able to evolve easily right?

codex-rs/core/src/config/mod.rs
ashwinnathan-openai: iiuc we can't do `None` = `not enabled`  because in the `disable_defaults = false` case, if a thread override doesn't include `tools.shell`, we'd resolve that as `tool_feature_overrides.shell = None` and incorrectly treat it as disabled instead of inheriting from the lower layer.

I think we need `Option<bool>` for the three states:
  - `None` = inherit / no explicit override
  - `Some(true)` = enabled
  - `Some(false)` = disabled

Separately though open to having separate types with the associated config specific to each tool but don't know if that's premature atm.

codex-rs/core/src/tools/spec.rs
```diff
@@ -461,6 +491,71 @@ impl ToolsConfig {
         nested.code_mode_only_enabled = false;
         nested
     }
+
+    pub fn is_builtin_tool_invocation_enabled(&self, tool_name: &str) -> bool {
+        let Some(tool) = builtin_tool_key(tool_name) else {
+            return false;
+        };
+        let Some(feature) = ToolFeatureKey::for_builtin_tool(tool) else {
+            return true;
```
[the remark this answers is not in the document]
ashwinnathan-openai: Yes that's fine we decided to not do anything with the MCP tools for now.

codex-rs/app-server/tests/suite/v2/thread_start.rs
```diff
@@ -150,6 +151,46 @@ async fn thread_start_creates_thread_and_emits_started() -> Result<()> {
...
+        })
+        .await?;
+
+    let resp: JSONRPCResponse = timeout(
+        DEFAULT_READ_TIMEOUT,
+        mcp.read_stream_until_response_message(RequestId::Integer(req_id)),
+    )
+    .await??;
+    let ThreadStartResponse { thread, .. } = to_response::<ThreadStartResponse>(resp)?;
+    assert!(!thread.id.is_empty(), "thread id should not be empty");
```
bolinfest, line 192: So this asserts that it starts but not that `tools` was honored, correct?

I suppose we would have to put something in the `ThreadStartResponse` if we wanted stronger signal?

codex-rs/core/src/config/mod.rs
bolinfest: But `ToolFeatureOverrides` is a field of `Config` (as opposed to `ConfigToml`), so we use `ConfigToml` to derive this value so we should know the value of `disable_defaults` when deriving this struct, no?

codex-rs/core/src/config/mod.rs
bolinfest: I guess so since that's what you have for `WebSearchFeatureToml` but not the others?

codex-rs/core/src/config/mod.rs
```diff
@@ -2020,31 +2102,43 @@ fn resolve_web_search_mode(
...
+        } else {
+            None
+        }
+    };
+
+    match tool_feature_overrides.web_search {
+        Some(true) => explicit_mode().or_else(feature_default_mode),
+        Some(false) => Some(WebSearchMode::Disabled),
+        None if tool_feature_overrides.disable_defaults => Some(WebSearchMode::Disabled),
+        None => explicit_mode().or_else(feature_default_mode),
```
bolinfest: The lambdas are unexpected and I think they can be eliminated like so:

```suggestion
    match tool_feature_overrides.web_search {
        Some(false) => Some(WebSearchMode::Disabled),
        None if tool_feature_overrides.disable_defaults => Some(WebSearchMode::Disabled),
        None | Some(true) => {
           // inline the code for explicit_mode().or_else(feature_default_mode) here?
        },
```

codex-rs/core/src/config/mod.rs
```diff
@@ -2054,6 +2148,93 @@ fn resolve_web_search_config(
...
+            config_toml
+                .tools
+                .as_ref()
+                .and_then(|tools| tools.view_image)
+        })
+}
+
+fn resolve_tool_feature_overrides(
+    config_toml: &ConfigToml,
+    config_profile: &ConfigProfile,
```
bolinfest, line 2183: Are we obligated to honor profiles?

codex-rs/core/src/tools/registry.rs
```diff
@@ -20,9 +21,199 @@ use codex_hooks::HookToolInput;
...
+    ListMcpResourceTemplates,
+    ReadFile,
+    ReadMcpResource,
+    RequestPermissions,
+    RequestUserInput,
+    ReportAgentJobResult,
+    ResumeAgent,
+    SearchToolBm25,
+    SendInput,
+    SpawnAgent,
```
bolinfest, line 63: What about the other shell tools like `ShellCommand`?

codex-rs/app-server/tests/suite/v2/thread_start.rs
ashwinnathan-openai: Yes this only checks that request starts with this override shape

codex-rs/core/src/config/mod.rs
ashwinnathan-openai: Yea websearch is the bespoke case for now

codex-rs/core/src/tools/spec.rs
```diff
@@ -461,6 +491,71 @@ impl ToolsConfig {
...
+    fn is_tool_feature_enabled(&self, feature: ToolFeatureKey) -> bool {
+        if let Some(enabled) = self.tool_feature_overrides.for_feature(feature) {
+            return enabled;
+        }
+
+        if self.tool_feature_overrides.disable_defaults {
+            return false;
+        }
+
+        self.is_tool_feature_enabled_by_default(feature)
```
bolinfest: Note this is slightly more idiomatic Rust. You could go further and eliminate the `return` above, as well.

```suggestion
        if self.tool_feature_overrides.disable_defaults {
            false
        } else {
            self.is_tool_feature_enabled_by_default(feature)
        }
```

codex-rs/core/src/tools/registry.rs
ashwinnathan-openai: The aliases are here but I think it does make more sense to have a flat list of all raw tool names and control the mapping in `ToolFeatureKey` - I will update

```
Self::ExecCommand => &[
      SHELL_TOOL_NAME,
      EXEC_COMMAND_TOOL_NAME,
      CONTAINER_EXEC_TOOL_NAME,
      LOCAL_SHELL_TOOL_NAME,
      SHELL_COMMAND_TOOL_NAME,
  ],
```

codex-rs/core/src/config/mod.rs
ashwinnathan-openai: Oh I think I understand, we can correctly infer the interpretation of `None` based on `disable_defaults` e.g. if it is `false` we inherit and if `true` we unset it?

codex-rs/core/src/config/mod.rs
ashwinnathan-openai: I think so... I don't think people would really have specific tool enablement logic live in profiles but not honoring it and then trying to bake it back in would be a breaking change.

codex-rs/core/src/config/mod.rs
ashwinnathan-openai: Actually still a bit confused... disable_defaults helps with the None disambiguation but how would we represent an explicit disable?

```
[tools.shell]
enabled = false
```

ToolFeatureOverrides is not the final runtime resolution, that happens in `spec.rs`

- ToolsConfig::new(...) computes runtime defaults/capabilities from model + feature flags
- is_builtin_tool_invocation_enabled(...) determines the final expose

So if this layer only had None / Some(config), we couldn't distinguish “explicitly disable this tool” from “no explicit override; inherit if defaults are enabled.”

PR #14527, opened 2026-03-12
app-server: Add platform os and family to init response

euroelessar:
This allows the client to pick os-specific behavior while interacting with the app server, e.g. to use proper path separators.


codex-rs/app-server-protocol/src/protocol/v1.rs
```diff
@@ -56,6 +56,8 @@ pub struct InitializeCapabilities {
 #[serde(rename_all = "camelCase")]
 pub struct InitializeResponse {
     pub user_agent: String,
+    pub platform_family: String,
+    pub platform_os: String,
```
owenlin0, line 64: should we add some docstrings to document what the known values are for these?

owenlin0 (approved):

PR #14529, opened 2026-03-12
Simplify permissions available in request permissions tool

mousseau-oai:
The `request_permissions` tool can now only specific file system and network permissions.

### DIFF
--- a/codex-rs/core/src/tools/spec.rs
+++ b/codex-rs/core/src/tools/spec.rs
@@ -389,43 +389,6 @@ fn create_file_system_permissions_schema() -> JsonSchema {
     }
 }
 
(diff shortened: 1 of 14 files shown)


codex-rs/core/src/tools/spec.rs
dylan-hurd-oai: let's just consolidate and disable here as well!

dylan-hurd-oai (approved):

dylan-hurd-oai (approved):

codex-rs/core/src/tools/spec.rs
mousseau-oai: Good catch, removed the `macos` entry from the tool definition.

PR #14531, opened 2026-03-12
Add plugin usage telemetry

alexsong-oai:
adding metrics including: 
* plugin used
* plugin installed/uninstalled
* plugin enabled/disabled


codex-rs/core/src/plugins/toggles.rs
```diff
@@ -0,0 +1,206 @@
...
+                .as_ref()
+                .and_then(|config| config.plugins.get(plugin_id))
+                .map(|plugin| plugin.enabled);
+            (plugin_id.clone(), enabled)
+        })
+        .collect()
+}
+
+pub fn plugin_toggle_events_to_emit(
+    previous_states: &BTreeMap<String, Option<bool>>,
```
xl-openai: Can we simplify the logic by no checking previous_status?

codex-rs/app-server/src/message_processor.rs
```diff
@@ -206,6 +207,11 @@ impl MessageProcessor {
             .plugins_manager()
             .maybe_start_curated_repo_sync_for_config(&config);
         let cloud_requirements = Arc::new(RwLock::new(cloud_requirements));
+        let analytics_events_client =
+            AnalyticsEventsClient::new(Arc::clone(&config), Arc::clone(&auth_manager));
+        thread_manager
+            .plugins_manager()
+            .set_analytics_events_client(analytics_events_client.clone());
```
xl-openai: why not just set it when you init plugin_manager?

codex-rs/core/src/plugins/toggles.rs
alexsong-oai: We need the previous state to make sure it’s actually an update?

xl-openai: I think it’s already checked by the UI, except for a minimal race condition.

xl-openai (approved):

codex-rs/core/src/plugins/toggles.rs
```diff
@@ -0,0 +1,100 @@
...
+                if plugins == "plugins" && enabled == "enabled" && value.is_boolean() =>
+            {
+                if let Some(enabled) = value.as_bool() {
+                    pending_changes.insert(plugin_id.clone(), enabled);
+                }
+            }
+            [plugins, plugin_id] if plugins == "plugins" => {
+                if let Some(enabled) = value.get("enabled").and_then(JsonValue::as_bool) {
+                    pending_changes.insert(plugin_id.clone(), enabled);
+                }
```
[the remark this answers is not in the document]
alexsong-oai: this should be fine

codex-rs/core/src/plugins/toggles.rs
```diff
@@ -0,0 +1,100 @@
...
+            }
+            [plugins] if plugins == "plugins" => {
```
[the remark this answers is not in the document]
alexsong-oai: this should be fine

PR #14533, opened 2026-03-13
app-server: add filesystem watch support

euroelessar:
### Summary
Add the v2 app-server filesystem watch RPCs and notifications, wire them through the message processor, and implement connection-scoped watches with notify-backed change delivery. This also updates the schema fixtures, app-server documentation, and the v2 integration coverage for watch and unwatch behavior.

This allows clients to efficiently watch for filesystem updates, e.g. to react on branch changes.

### Testing
- exercise watch lifecycles for directory changes, atomic file replacement, missing-file targets, and unwatch cleanup


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2215,6 +2215,52 @@ pub struct FsCopyParams {
...
+    pub path: AbsolutePathBuf,
+}
+
+/// Created watch handle returned by `fs/watch`.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct FsWatchResponse {
+    /// Connection-scoped watch identifier used for `fs/unwatch` and `fs/changed`.
+    pub watch_id: String,
```
owenlin0, line 2319: minor: have we considered having the client operate on paths only, instead of having the client have to be concerned with watch IDs? i.e. two fs/watch calls to the same path would be idempotent and still be one subscription

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2215,6 +2215,52 @@ pub struct FsCopyParams {
...
+#[ts(export_to = "v2/")]
+pub struct FsUnwatchResponse {}
+
+/// Filesystem watch notification emitted for `fs/watch` subscribers.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct FsChangedNotification {
+    /// Watch identifier returned by `fs/watch`.
+    pub watch_id: String,
```
owenlin0, line 2345: if we do away with `watch_id`, subscriptions could be keyed off of the `watch_path` instead?

non-blocking, up to you

owenlin0 (approved):

codex-rs/app-server-protocol/src/protocol/v2.rs
euroelessar: I've given it some thought and it looks like it complicates the unwatching logic a fair bit?
we would need to track "# of watcher per path" on either client or server side, otherwise we may unsubscribe when we didn't intend to yet and stop receiving updates.
wdyt?

owenlin0: hmm, why do we need to track # of watchers per path? 

if a client operates on paths, then it's either watching a path or not, right? so multiple `fs/watch` calls to a path will be idempotent. and `fs/unwatch` will just delete the subscription, if it exists.

wouldn't watch state just be: `HashSet<(connection_id, path)>`?

euroelessar: imagine if watchers are dynamic, and are periodically created/destroyed based on use cases
clients will need to track how many watchers do they have per path on their side to avoid prematurely calling `fs/unwatch` if there is still some interest in the path in some other part of the app

owenlin0: makes sense, if it makes the client implementation simpler i am for it

euroelessar: yeah, I've pushed all the hard deduping/counting/etc logic into core's fs watcher in order to make clients simpler

owenlin0 (approved):

PR #14541, opened 2026-03-13
chore: clarify plugin + app copy in model instructions

sayan-oai:
- clarify app mentions are in user messages
- clarify what it means for tools to be provided via `codex_apps` MCP
- add plugin descriptions (with basic sanitization) to top-level `## Plugins` section alongside the corresponding plugin names
- explain that skills from plugins are prefixed with `plugin_name:` in top-level `##Plugins` section

changes to more logically organize `Apps`, `Skills`, and `Plugins` instructions will be in a separate PR, as that shuffles dev + user instructions in ways that change tests broadly.

### Tests
confirmed in local rollout, some new tests.


codex-rs/core/src/plugins/manager.rs
```diff
@@ -188,6 +189,21 @@ impl PluginCapabilitySummary {
...
+fn prompt_safe_plugin_description(description: Option<&str>) -> Option<String> {
+    let description = description?
+        .split_whitespace()
+        .collect::<Vec<_>>()
+        .join(" ");
+
+    if description.is_empty()
+        || description.chars().count() > MAX_CAPABILITY_SUMMARY_DESCRIPTION_LEN
+    {
+        None
```
pakrym-oai: Truncate instead?

pakrym-oai (approved):

PR #14542, opened 2026-03-13
[bazel] Bump up cc and rust toolchains

zbarsky-openai:
This lets us drop various patches and go all the way to a very clean setup.

In case folks are curious what was going on... we were depending on the toolchain finding stdlib headers as sibling files of `clang++`, and for linking we were providing a `-resource-dir` containing the runtime libs. However, some users of the cc toolchain (such as rust build scripts) do the equivalent of `$CC $CCFLAGS $LDFLAGS` so the `-resource-dir` was being passed when compiling, which suppressed the default stdlib header location logic. The upstream fix was to swap to using `-isystem` to pass the stdlib headers, while carefully controlling the ordering to simulate them coming from the resource-dir.


MODULE.bazel
```diff
@@ -82,16 +75,11 @@ crate.annotation(
...
-    patch_args = ["-p1"],
-    patches = [
-        "//patches:aws-lc-sys_memcmp_check.patch",
-    ],
 )
 
 inject_repo(crate, "zstd")
 
 bazel_dep(name = "bzip2", version = "1.0.8.bcr.3")
-bazel_dep(name = "libcap", version = "2.27.bcr.1")
```
zbarsky-openai, line 94: moves this to not be interspersed with the modules used in the crates

.bazelrc
```diff
@@ -56,3 +56,7 @@ common --jobs=30
 common:remote --extra_execution_platforms=//:rbe
 common:remote --remote_executor=grpcs://remote.buildbuddy.io
 common:remote --jobs=800
+# TODO(team): Evaluate if this actually helps, zbarsky is not sure, everything seems bottlenecked on `core` either way.
+# Enable pipelined compilation since we are not bound by local CPU count.
+#common:remote --@rules_rust//rust/settings:pipelined_compilation
```
bolinfest, line 61: OK, so we're leaving this commented out for now?

bolinfest (approved):
So pretty!

.bazelrc
zbarsky-openai: yeah, I wasn't sure if this was better or not, so I left a breadcrumb to make it easy to evaluate as a followup

PR #14546, opened 2026-03-13
feat(tui): start wiring app-server backend behind --app-server

fcoury:
## Problem

The TUI talks directly to codex-core, duplicating business logic that the app-server already provides to other surfaces like the VSCode extension. This means behavior can diverge between surfaces, and any fix or enhancement must be replicated in both the TUI and the app-server independently. The goal is to migrate the TUI onto the app-server so all clients share a single implementation. Skills loading (`ListSkills`) is the first operation to move — it's self-contained, has clear request/response semantics, and exercises the full conversion pipeline — making it a good proving ground for the pattern the remaining operations will follow.

## Mental model

There are now two skill-loading channels, selected by the `--app-server` CLI flag:

1. **Legacy (default):** `ChatWidget` sends `Op::ListSkills` over the codex-core op channel. Core handles the request end-to-end and fires `EventMsg::ListSkillsResponse` back.
2. **App-server:** `ChatWidget` wraps `Op::ListSkills` inside `AppEvent::CodexOp(…)` and sends it up to the `App` event loop. `App::run` intercepts it, expands empty `cwds` against the active chat session cwd, kicks off a detached typed `skills/list` RPC against the embedded app-server, converts the response from app-server protocol types into core protocol types, and pushes the result into the same `set_skills_from_response` / `refresh_plugin_mentions` path the legacy flow uses.

The decision point is `ChatWidget::request_skills_list`, which inspects `use_app_server` and routes to the appropriate channel. The `App` event loop's `select!` arm pattern-matches on `AppEvent::CodexOp(Op::ListSkills { .. })` only when `use_app_server` is true; all other events fall through to the existing `handle_event` handler.

## Non-goals

- Routing any operation *other than* `ListSkills` through the app-server. This is the first gated operation; others will follow the same pattern.
- Changing the app-server protocol types to match core types (or vice versa). The conversion functions are intentionally mechanical 1:1 mappings.
- Removing the legacy codex-core skills path. Both paths must work.

## Tradeoffs

- **~130 lines of `into_core_*` boilerplate (8 functions):** These could be derived with a macro or eliminated by unifying the protocol types, but field-by-field mapping keeps the two protocol crates decoupled and makes divergence between them compile-time visible.
- **Hardcoded `RequestId::Integer(1)`:** The detached `skills/list` path still uses a static request ID. That is acceptable for the current narrowly scoped migration, but it is an explicit limitation and should become a counter before this pattern is reused more broadly.
- **`per_cwd_extra_user_roots: None`:** The TUI does not yet support per-CWD extra roots. This is deferred, not forgotten.

## Architecture

```
CLI --app-server flag
  → lib.rs: extracted into `app_server` bool, passed to App::run
    → App::run: threaded into ChatWidgetInit.use_app_server
      → ChatWidget: stored as field, exposed via use_app_server()

ChatWidget::request_skills_list(cwds, force_reload)
  ├─ use_app_server=false → submit_op(Op::ListSkills) → codex-core op channel
  └─ use_app_server=true  → app_event_tx.send(AppEvent::CodexOp(Op::ListSkills))
                                ↓
                           App event loop select! arm
                                ↓
                           effective_skills_list_cwds
                                ↓
                           request_app_server_skills_list (tokio::spawn)
                                ↓
                           app_server.requester().request_typed(ClientRequest::SkillsList)
                                ↓
                           into_core_skills_list_response_event (+ into_core_* children)
                                ↓
                           app_event_tx.send(AppEvent::SkillsListLoaded { result })
                                ↓
                           App event loop handles SkillsListLoaded
                                ↓
                           chat_widget.set_skills_from_response / refresh_plugin_mentions
```

### App-server adapter (`app_server_adapter.rs`)

A transitional adapter module on `App` that drains the app-server event stream during the hybrid migration period. It handles `Lagged`, `ServerNotification`, and `LegacyNotification` events, and rejects any `ServerRequest` with a JSON-RPC -32000 error since the TUI does not yet handle server-initiated requests. This module should shrink and disappear as more flows move onto the app-server surface.

### Files changed outside `tui/`

The only non-TUI code touched by this branch is:

- **`app-server-client/src/lib.rs`** — Extended `InProcessAppServerClient` with typed request/response plumbing and an `InProcessAppServerRequester` handle for making RPC calls from spawned tasks.

## Observability

- `skills/list` RPC failures surface as `color_eyre` errors that propagate up and terminate the app — no silent swallowing.
- Skill-load warnings (invalid SKILL.md files) are emitted via `emit_skill_load_warnings` on the app-server path identically to the legacy path.

## Tests

- `list_skills_without_app_server_uses_core_op_channel` — verifies `SkillsUpdateAvailable` routes through the op channel when `use_app_server=false`, and nothing appears on the app-event channel.
- `list_skills_with_app_server_uses_app_event_channel` — verifies the inverse: op channel stays empty, app-event channel receives the wrapped op.
- `effective_skills_list_cwds_uses_active_chat_widget_cwd_when_request_empty` — verifies the app-server path uses the active session cwd rather than the original launch cwd when the request does not specify `cwds`.
- `effective_skills_list_cwds_preserves_explicit_cwds` — verifies explicitly provided cwd values are forwarded unchanged.
- `parses_app_server_flag` — confirms clap parses `--app-server`.
- Existing tests updated to pass `use_app_server: false` to `ChatWidgetInit` constructors, confirming no behavioral change in the default path.


codex-rs/tui/src/app.rs
```diff
@@ -191,6 +196,147 @@ fn errors_for_cwd(cwd: &Path, response: &ListSkillsResponseEvent) -> Vec<SkillEr
...
+fn request_app_server_skills_list(
+    app_event_tx: AppEventSender,
+    client: InProcessAppServerRequester,
+    cwds: Vec<PathBuf>,
+    force_reload: bool,
+) {
+    tokio::spawn(async move {
+        let result = client
+            .request_typed(ClientRequest::SkillsList {
+                request_id: RequestId::Integer(1),
```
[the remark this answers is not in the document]
fcoury: Added `NEXT_APP_SERVER_REQUEST_ID` to keep and provide the next item id atomically.

codex-rs/tui/src/app.rs
```diff
@@ -2480,6 +2650,13 @@ impl App {
             AppEvent::ConnectorsLoaded { result, is_final } => {
                 self.chat_widget.on_connectors_loaded(result, is_final);
             }
+            AppEvent::SkillsListLoaded { result } => {
+                let event = result.map_err(|err| color_eyre::eyre::eyre!(err))?;
+                let cwd = self.chat_widget.config_ref().cwd.clone();
+                emit_skill_load_warnings(&self.app_event_tx, &errors_for_cwd(&cwd, &event));
+                self.chat_widget.set_skills_from_response(&event);
```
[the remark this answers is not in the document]
fcoury: Added the list of directories to `AppEvent::SkillsListLoaded` in a way that we can check if it contains the current directory on which the `ChatWidget` is operating on. 

Also added a test to assure that a notification we get back with a list of directories that don't match it is disregarded.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -482,6 +482,9 @@ pub(crate) struct ChatWidgetInit {
     // Shared latch so we only warn once about invalid status-line item IDs.
     pub(crate) status_line_invalid_items_warned: Arc<AtomicBool>,
     pub(crate) session_telemetry: SessionTelemetry,
+    /// When true, route eligible operations (currently `ListSkills`) through
```
etraut-openai: Remove reference to "ListSkills" since that comment won't age well.

codex-rs/tui/src/app.rs
```diff
@@ -191,6 +199,154 @@ fn errors_for_cwd(cwd: &Path, response: &ListSkillsResponseEvent) -> Vec<SkillEr
         .unwrap_or_default()
 }
 
+fn effective_skills_list_cwds(cwds: Vec<PathBuf>, current_cwd: &Path) -> Vec<PathBuf> {
```
etraut-openai: Why are all these skills methods in the app.rs module? This doesn't seem like the right place for them.

codex-rs/tui/src/app.rs
```diff
@@ -191,6 +199,154 @@ fn errors_for_cwd(cwd: &Path, response: &ListSkillsResponseEvent) -> Vec<SkillEr
...
+    });
+}
+
+/// Convert an app-server protocol skills entry into the core protocol
+/// equivalent so the existing `ChatWidget` skills UI can consume it
+/// unchanged. Each nested type has a corresponding `into_core_*` helper
+/// below; these are intentionally field-by-field rather than derived to
+/// keep the two protocol crates decoupled and make field divergence a
+/// compile error.
+fn into_core_skills_list_entry(
```
etraut-openai: It looks like it takes a lot of code to convert from app server to core data structures in this case. We'll eventually want the TUI to consume app server data structures directly, so maybe it's better to convert in the other direction? Or do you think we should do that as a separate (later) step?

etraut-openai (commented):
I started to review. This was a good experiment because the ListSkills doesn't involve any events, so it's relatively straightforward. However, this highlights how challenging it's going to be to retain both legacy and app server code paths — even in the short term. It makes me question whether a feature flag is the right approach here. 

I'm going to do an experiment and create a separate parallel PR that explores the approach where we don't make it conditional and just cut over to the new app server path so we can compare.

codex-rs/tui/src/app.rs
fcoury: I have another branch where I do that. We can discuss the implementation. In this particular PR I wanted to do as little as possible to get this one action migrated.

codex-rs/tui/src/chatwidget.rs
fcoury: Ouch, agreed. Comment is now improved.

codex-rs/tui/src/app.rs
fcoury: I extracted a new `skills.rs` for the helper functions.

PR #14568, opened 2026-03-13
fix: support split carveouts in windows elevated sandbox

viyatb-oai:
## Summary
- preserve legacy Windows elevated sandbox behavior for existing policies
- add elevated-only support for split filesystem policies that can be represented as readable-root overrides, writable-root overrides, and extra deny-write carveouts
- resolve those elevated filesystem overrides during sandbox transform and thread them through setup and policy refresh
- keep failing closed for explicit unreadable (`none`) carveouts and reopened writable descendants under read-only carveouts
- for explicit read-only-under-writable-root carveouts, materialize missing carveout directories during elevated setup before applying the deny-write ACL
- document the elevated vs restricted-token support split in the core README

## Example
Given a split filesystem policy like:

```toml
":root" = "read"
":cwd" = "write"
"./docs" = "read"
"C:/scratch" = "write"
```

the elevated backend now provisions the readable-root overrides, writable-root overrides, and extra deny-write carveouts during setup and refresh instead of collapsing back to the legacy workspace-only shape.

If a read-only carveout under a writable root is missing at setup time, elevated setup creates that carveout as an empty directory before applying its deny-write ACE; otherwise the sandboxed command could create it later and bypass the carveout. This is only for explicit policy carveouts. Best-effort workspace protections like `.codex/` and `.agents/` still skip missing directories.

A policy like:

```toml
"/workspace" = "write"
"/workspace/docs" = "read"
"/workspace/docs/tmp" = "write"
```

still fails closed, because the elevated backend does not reopen writable descendants under read-only carveouts yet.


codex-rs/windows-sandbox-rs/src/identity.rs
```diff
@@ -122,16 +124,24 @@ fn select_identity(policy: &SandboxPolicy, codex_home: &Path) -> Result<Option<S
...
     env_map: &HashMap<String, String>,
     codex_home: &Path,
+    read_roots_override: Option<&[PathBuf]>,
+    write_roots_override: Option<&[PathBuf]>,
+    deny_write_paths_override: &[PathBuf],
 ) -> Result<SandboxCreds> {
     let sandbox_dir = crate::setup::sandbox_dir(codex_home);
-    let needed_read = gather_read_roots(command_cwd, policy, codex_home);
-    let needed_write = gather_write_roots(policy, policy_cwd, command_cwd, env_map);
+    let needed_read = read_roots_override
```
iceweasel-oai, line 145: `gather_read_roots()` attempts to provide read access to the sandbox user to a number of directories that we think it might need in order to actually be able to execute commands (like the real user's profile dir, some important windows directories, etc.)

I'm worried that if we just replace those with a set of config-supplied directories then some commands will fail in surprising ways. Unless, this feature is meant to be literally "the sandbox can read this directories and nothing else"?

viyatb-oai: good point, fixed it.

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
```diff
@@ -138,6 +138,7 @@ pub(super) async fn try_run_zsh_fork(
         file_system_sandbox_policy,
         network_sandbox_policy,
         windows_restricted_token_filesystem_overlay: _windows_restricted_token_filesystem_overlay,
+        windows_elevated_filesystem_overrides: _windows_elevated_filesystem_overrides,
```
iceweasel-oai: what do you think about merging these into one thing, and then each sandbox implementation (legacy vs elevated) can decide which of the fields it actually supports?

codex-rs/core/src/exec.rs
```diff
@@ -104,6 +104,18 @@ pub(crate) struct WindowsRestrictedTokenFilesystemOverlay {
     pub(crate) additional_deny_write_paths: Vec<AbsolutePathBuf>,
 }
 
+/// Resolved filesystem overrides for the elevated Windows sandbox backend.
+///
+/// Unlike the restricted-token path, elevated setup can provision explicit read
+/// and write roots up front, so we pass the exact split-policy roots and any
```
iceweasel-oai: might want to update this docstring to indicate that read roots include the `gather_read_roots` ones too

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
viyatb-oai: Yep, that shape is cleaner. I merged the restricted-token overlay and elevated overrides into one core-internal `WindowsSandboxFilesystemOverrides` struct, and each backend now just consumes the subset it supports.

codex-rs/core/src/exec.rs
```diff
@@ -300,15 +303,26 @@ pub fn build_exec_request(
...
+            resolve_windows_elevated_filesystem_overrides(
+                exec_req.sandbox,
+                &exec_req.sandbox_policy,
+                &exec_req.file_system_sandbox_policy,
+                exec_req.network_sandbox_policy,
+                sandbox_cwd,
+                exec_req.windows_sandbox_level,
+            )
+        } else {
+            resolve_windows_restricted_token_filesystem_overlay(
```
iceweasel-oai: I'd suggest renaming all overlay -> override now that they both use the same WindowsSandboxFilesystemOverrides struct

iceweasel-oai (approved):

PR #14609, opened 2026-03-13
move plugin/skill instructions into dev msg and reorder

sayan-oai:
Move the general `Apps`, `Skills` and `Plugins` instructions blocks out of `user_instructions` and into the developer message, with new `Apps -> Skills -> Plugins` order for better clarity.

Also wrap those sections in stable XML-style instruction tags (like other sections) and update prompt-layout tests/snapshots. This makes the tests less brittle in snapshot output (we can parse the sections), and it consolidates the capability instructions in one place.

#### Tests
Updated snapshots, added tests.

`<AGENTS_MD>` disappearing in snapshots is expected: before this change, the wrapped user-instructions message was kept alive by `Skills` content. Now that `Skills` and `Plugins` are in the developer message, that wrapper only appears when there is real project-doc/user-instructions content.


codex-rs/core/tests/common/context_snapshot.rs
```diff
@@ -241,6 +246,15 @@ fn canonicalize_snapshot_text(text: &str) -> String {
     if text.starts_with("<permissions instructions>") {
```
sayan-oai, line 259: one note-- the `<permissions instructions>` tag is the only one that doesn't use underscores, i.e., `<permissions_instructions>`. can bring inline, but didn't want to touch as part of these changes.

codex-rs/core/tests/suite/snapshots/all__suite__compact__manual_compact_with_history_shapes.snap
```diff
@@ -6,9 +6,7 @@ Scenario: Manual /compact with prior user history compacts existing history and
 
 ## Local Compaction Request
 00:message/developer:<PERMISSIONS_INSTRUCTIONS>
-01:message/user[2]:
-    [01] <AGENTS_MD>
```
sayan-oai, line 10: snapshots dropping `<AGENTS_MD>` is expected; `get_user_instructions()` used to carry the general Skills/Plugins sections, so the wrapped user-instructions message often existed even when there was no real project doc content. After moving those sections into the developer message, that wrapper is now only present when there is actual user/project-doc-style content, so snapshots like this legitimately drop `<AGENTS_MD>`.

codex-rs/core/src/guardian_tests.rs
```diff
@@ -290,7 +291,9 @@ async fn guardian_review_request_layout_matches_model_visible_request_snapshot()
     .await;
 
     let (mut session, mut turn) = crate::codex::make_session_and_context().await;
+    let temp_cwd = TempDir::new()?;
     let mut config = (*turn.config).clone();
+    config.cwd = temp_cwd.path().to_path_buf();
```
sayan-oai, line 419: make guardian tests hermetic; some platforms would find agents.md and others wouldnt. so now we pass a temp dir.

codex-rs/core/src/plugins/render_tests.rs
```diff
@@ -17,7 +19,9 @@ fn render_plugins_section_includes_descriptions_and_skill_naming_guidance() {
...
+    let expected = format!(
+        "{PLUGINS_INSTRUCTIONS_OPEN_TAG}\n## Plugins\nA plugin is a local bundle of skills, MCP servers, and apps. Below is the list of plugins that are enabled and available in this session.\n### Available plugins\n- `sample`: inspect sample data\n### How to use plugins\n- Discovery: The list above is the plugins available in this session.\n- Skill naming: If a plugin contributes skills, those skill entries are prefixed with `plugin_name:` in the Skills list.\n- Trigger rules: If the user explicitly names a plugin, prefer capabilities associated with that plugin for that turn.\n- Relationship to capabilities: Plugins are not invoked directly. Use their underlying skills, MCP tools, and app tools to help solve the task.\n- Preference: When a relevant plugin is available, prefer using capabilities associated with that plugin over standalone capabilities that provide similar functionality.\n- Missing/blocked: If the user requests a plugin that is not listed above, or the plugin does not have relevant callable capabilities for the task, say so briefly and continue with the best fallback.\n{PLUGINS_INSTRUCTIONS_CLOSE_TAG}"
+    );
```
charley-oai: Usually better to hardcode values like this in tests so you can notice when they change (imho)

codex-rs/core/tests/suite/snapshots/all__suite__compact__manual_compact_with_history_shapes.snap
charley-oai: Ah would be nice to update the test to still show legit agents.md but that's out of scope,  I think we should update these tests to show every kind of fragment (and like auto fold over all of them so when a new on is added it gets auto tested) but anyways

charley-oai (approved):

PR #14610, opened 2026-03-13
feat: support restricted ReadOnlyAccess in elevated Windows sandbox

viyatb-oai:
## Summary
- support legacy `ReadOnlyAccess::Restricted` on Windows in the elevated setup/runner backend
- keep the unelevated restricted-token backend on the legacy full-read model only, and fail closed for restricted read-only policies there
- keep the legacy full-read Windows path unchanged while deriving narrower read roots only for elevated restricted-read policies
- honor `include_platform_defaults` by adding backend-managed Windows system roots only when requested, while always keeping helper roots and the command `cwd` readable
- preserve `workspace-write` semantics by keeping writable roots readable when restricted read access is in use in the elevated backend
- document the current Windows boundary: legacy `SandboxPolicy` is supported on both backends, while richer split-only carveouts still fail closed instead of running with weaker enforcement

## Testing
- `cargo check -p codex-windows-sandbox --tests --target x86_64-pc-windows-msvc`
- `cargo clippy -p codex-windows-sandbox --tests --target x86_64-pc-windows-msvc -- -D warnings`
- `cargo test -p codex-core windows_restricted_token_`

## Notes
- local `cargo test -p codex-windows-sandbox` on macOS only exercises the non-Windows stubs; the Windows-targeted compile and clippy runs provide the local signal, and GitHub Windows CI exercises the runtime path

### DIFF
--- a/codex-rs/core/src/exec.rs
+++ b/codex-rs/core/src/exec.rs
@@ -34,6 +34,7 @@ use crate::spawn::spawn_child_async;
 use crate::text_encoding::bytes_to_string_smart;
 use crate::tools::sandboxing::SandboxablePreference;
 use codex_network_proxy::NetworkProxy;
+#[cfg(any(target_os = "windows", test))]
 use codex_protocol::permissions::FileSystemSandboxKind;
 use codex_protocol::permissions::FileSystemSandboxPolicy;
 use codex_protocol::permissions::NetworkSandboxPolicy;
@@ -765,13 +766,14 @@ async fn exec(
 ) -> Result<RawExecToolCallOutput> {
     #[cfg(target_os = "windows")]
     if sandbox == SandboxType::WindowsRestrictedToken {
-        if let Some(reason) = unsupported_windows_restricted_token_sandbox_reason(
+        let support = windows_restricted_token_sandbox_support(
             sandbox,
             params.windows_sandbox_level,
             sandbox_policy,
             file_system_sandbox_policy,
             network_sandbox_policy,
-        ) {
+        );
+        if let Some(reason) = support.unsupported_reason {
             return Err(CodexErr::Io(io::Error::other(reason)));
         }
         return exec_windows_sandbox(params, sandbox_policy).await;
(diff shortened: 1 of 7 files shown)


codex-rs/windows-sandbox-rs/src/lib.rs
```diff
@@ -272,11 +272,6 @@ mod windows_impl {
         ) {
             anyhow::bail!("DangerFullAccess and ExternalSandbox are not supported for sandboxing")
         }
```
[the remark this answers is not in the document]
viyatb-oai: Good catch. I narrowed this back so `windowsSandboxLevel=restricted-token` no longer accepts `ReadOnlyAccess::Restricted`; that backend remains legacy full-read only and fails closed for restricted read-only policies. This PR now only claims restricted-read support in the elevated setup/runner backend. Separately, the existing legacy workspace-write protections for `.git`, `.codex`, and `.agents` remain in place in the restricted-token path, so this does not make those subpaths newly writable.

codex-rs/core/src/exec.rs
```diff
@@ -815,6 +815,7 @@ fn should_use_windows_restricted_token_sandbox(
     file_system_sandbox_policy: &FileSystemSandboxPolicy,
 ) -> bool {
     sandbox == SandboxType::WindowsRestrictedToken
+        && sandbox_policy.has_full_disk_read_access()
```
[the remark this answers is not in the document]
viyatb-oai: Good catch. The preflight guard was still treating `SandboxType::WindowsRestrictedToken` as if it uniquely meant the unelevated backend, even though `exec_windows_sandbox()` splits again on `params.windows_sandbox_level`. I updated the guard to take `windows_sandbox_level` explicitly: restricted read-only still fails closed for the unelevated restricted-token path, but it is now allowed for the elevated setup/runner path. I also added `windows_elevated_sandbox_allows_restricted_read_only_policies` so this stays covered.

codex-rs/core/src/exec_tests.rs
```diff
@@ -190,6 +190,7 @@ fn windows_restricted_token_skips_external_sandbox_policies() {
     assert_eq!(
         should_use_windows_restricted_token_sandbox(
             SandboxType::WindowsRestrictedToken,
+            codex_protocol::config_types::WindowsSandboxLevel::Disabled,
```
bolinfest: Import `WindowsSandboxLevel` at the top?

codex-rs/core/src/exec_tests.rs
```diff
@@ -248,6 +252,40 @@ fn windows_restricted_token_allows_legacy_restricted_policies() {
...
+        access: codex_protocol::protocol::ReadOnlyAccess::Restricted {
+            include_platform_defaults: true,
+            readable_roots: vec![],
+        },
+        network_access: false,
+    };
+    let file_system_policy = FileSystemSandboxPolicy::from(&policy);
+
+    assert_eq!(
+        should_use_windows_restricted_token_sandbox(
```
bolinfest: I'm confused why `should_use_windows_restricted_token_sandbox()` and `unsupported_windows_restricted_token_sandbox_reason()` don't take the same args.

It feels like they should be one function that returns a struct/tuple with two values so `should_use_windows_restricted_token_sandbox()` is only exercised once?

codex-rs/core/src/exec.rs
bolinfest: While you're here, can you turn this into a simple `if` statement?

codex-rs/core/src/exec_tests.rs
```diff
@@ -248,6 +252,40 @@ fn windows_restricted_token_allows_legacy_restricted_policies() {
...
+            SandboxType::WindowsRestrictedToken,
+            codex_protocol::config_types::WindowsSandboxLevel::Disabled,
+            &policy,
+            &file_system_policy,
+        ),
+        false
```
bolinfest: for clarity, maybe add a string as the third argument like `"Not restricted because..."`

codex-rs/core/src/exec_tests.rs
```diff
@@ -262,6 +300,39 @@ fn windows_restricted_token_allows_legacy_workspace_write_policies() {
     assert_eq!(
         unsupported_windows_restricted_token_sandbox_reason(
             SandboxType::WindowsRestrictedToken,
+            codex_protocol::config_types::WindowsSandboxLevel::Disabled,
+            &policy,
+            &file_system_policy,
+            NetworkSandboxPolicy::Restricted,
+        ),
+        None
```
bolinfest: Same here?

bolinfest (approved):

PR #14611, opened 2026-03-13
Add auth 401 recovery observability to client bug reports

ccy-oai:
CXC-392

[With 401](https://openai.sentry.io/issues/7333870443/?project=4510195390611458&query=019ce8f8-560c-7f10-a00a-c59553740674&referrer=issue-stream)
401 auth tags in Sentry

```
- auth_401_*: preserved facts from the last unauthorized response we cared about
- auth_*: latest auth-related request facts for the uploaded session when they are not intentionally suppressed to preserve 401 context
- auth_recovery_*: what the client did after unauthorized, plus whether the follow-up succeeded
```

Without 401
happy-path auth tags in Sentry

###### Summary
- Add client-visible 401 diagnostics for auth attachment, upstream auth classification, and 401 request correlation ids.
- Record unauthorized recovery mode, phase, outcome, and retry/follow-up status without changing auth behavior.
- Surface the highest-signal auth and recovery fields on uploaded client bug reports so they are usable in Sentry.
- Preserve original unauthorized evidence under `auth_401_*` while keeping follow-up result tags separate.

###### Rationale (from spec findings)
- The dominant bucket needed proof of whether the client attached auth before send or upstream still classified the request as missing auth.
- Client uploads needed to show whether unauthorized recovery ran and what the client tried next.
- Request id and cf-ray needed to be preserved on the unauthorized response so server-side correlation is immediate.
- The bug-report path needed the same auth evidence as the request telemetry path, otherwise the observability would not be operationally useful.

###### Scope
- Add auth 401 and unauthorized-recovery observability in `codex-rs/core`, `codex-rs/codex-api`, and `codex-rs/otel`, including Sentry feedback tag surfacing.
- Keep auth semantics, refresh behavior, retry behavior, endpoint classification, and geo-denial follow-up work out of this PR.

###### Trade-offs
- This exports only safe auth evidence: header presence/name, upstream auth classification, request ids, and recovery state. It does not export token values or raw upstream bodies.
- This keeps websocket connection reuse as a transport clue because it can help distinguish stale reused sessions from fresh reconnects.
- Misroute/base-url classification and geo-denial are intentionally deferred to a separate follow-up PR so this review stays focused on the dominant auth 401 bucket.

###### Client follow-up
- PR 2 will add misroute/provider and geo-denial observability plus the matching Sentry surfacing.
- A separate host/app-server PR should log auth-decision inputs so pre-send host auth state can be correlated with client request evidence.
- `device_id` remains intentionally separate until there is a safe existing source on the feedback upload path.

###### Testing
- `cargo test -p codex-core emit_feedback_request_tags_`
- `cargo test -p codex-core emit_feedback_auth_recovery_tags_`
- `cargo test -p codex-core auth_request_telemetry_context_tracks_attached_auth_and_retry_phase`
- `cargo test -p codex-core websocket_session_tracks_connection_reuse`
- `cargo test -p codex-core extract_response_debug_context_decodes_identity_headers`
- `cargo test -p codex-core identity_auth_details`
- `cargo test -p codex-otel otel_export_routing_policy_routes_api_request_auth_observability`
- `cargo test -p codex-otel otel_export_routing_policy_routes_websocket_connect_auth_observability`
- `cargo test -p codex-otel otel_export_routing_policy_routes_websocket_request_transport_observability`


etraut-openai (approved):
Overall, looks good. The telemetry additions to the auth module, in particular, are going to come in handy for not only the 401 issue but also some of the token refresh bugs we've been playing whack-a-mole with for a while.

Before committing, there are some clippy errors to fix. Plus, there are some codex code review errors that look like legit (but perhaps edge-case) problems. I recommend running "codex review" a few more times to see if it finds anything more.

Thanks for doing this!

PR #14615, opened 2026-03-13
refactor(tui): move skills loading onto the app-server

fcoury:
## Problem

The TUI was still loading skills through a TUI/core-specific path even though the app-server already owns the skills API surface used by other clients. That left the TUI with duplicated request plumbing, duplicated event handling, and a separate in-memory skills model that had to be kept in sync with app-server behavior.

## What this change does

This PR moves TUI skills loading onto the app-server path and makes the TUI consume app-server protocol types directly.

Concretely:

- `skills/list` requests are issued as app-server RPCs from `App`
- `SkillsListLoaded` now carries `codex_app_server_protocol::SkillsListResponse`
- the TUI stores and renders `codex_app_server_protocol::SkillMetadata` directly instead of adapting it into legacy core/protocol skill structs
- the old app-server-to-core conversion layer for skills is removed
- live skills refresh in the TUI is now driven by app-server `ServerNotification::SkillsChanged`, not the legacy `EventMsg::SkillsUpdateAvailable` path
- session logging now summarizes `SkillsListLoaded` instead of dumping the full skills payload

## Mental model

There is now one TUI skills-loading path:

1. `ChatWidget` requests `Op::ListSkills`
2. `App::run` intercepts that op and calls the app-server `skills/list` RPC
3. the RPC result is returned as `AppEvent::SkillsListLoaded`
4. `App` applies the response to the chat widget and refreshes plugin mentions

For on-disk changes:

1. core’s file watcher still detects skill-file changes internally
2. app-server translates that into `ServerNotification::SkillsChanged`
3. the TUI app-server adapter turns that notification into `AppEvent::RefreshSkillsList`
4. `App` reissues `skills/list`

## Non-goals

- Routing unrelated TUI operations through the app-server in this PR
- Reworking core’s internal file-watcher ownership
- Changing the app-server protocol shape to match legacy TUI/core types

## Tradeoffs

- The TUI still has a small transitional app-server adapter module to translate app-server notifications into app events.
- Skills refresh on disk still originates from core internally, but the TUI no longer depends on the legacy core event to consume it.
- `per_cwd_extra_user_roots` remains `None`; this PR does not add per-CWD extra-root support.

## Files of interest

- `codex-rs/tui/src/app.rs`
- `codex-rs/tui/src/app/skills.rs`
- `codex-rs/tui/src/app/app_server_adapter.rs`
- `codex-rs/tui/src/app_event.rs`
- `codex-rs/tui/src/chatwidget/skills.rs`
- `codex-rs/tui/src/session_log.rs`
- `codex-rs/app-server-client/src/lib.rs`

## Tests

Updated and added coverage includes:

- `app_server_skills_changed_notification_emits_refresh_skills_list`
- `trigger_skills_list_request_uses_current_cwd_and_emits_result`
- `effective_skills_list_cwds_uses_active_chat_widget_cwd_when_request_empty`
- `effective_skills_list_cwds_preserves_explicit_cwds`
- stale-response tests for `SkillsListLoaded`
- `generated_id_typed_request_roundtrip_works` in `codex-app-server-client`

## Observability

`CODEX_TUI_RECORD_SESSION` no longer logs the full skills payload for `SkillsListLoaded`; it records a compact summary instead.


codex-rs/tui/src/app/skills.rs
```diff
@@ -0,0 +1,198 @@
...
+
+pub(super) fn effective_skills_list_cwds(cwds: Vec<PathBuf>, current_cwd: &Path) -> Vec<PathBuf> {
+    if cwds.is_empty() {
+        vec![current_cwd.to_path_buf()]
+    } else {
+        cwds
+    }
+}
+
+pub(super) fn request_app_server_skills_list(
```
etraut-openai: If we're not going to implement a conditional switch, maybe we should avoid using "app_server" and "legacy" in cases where we're just cutting over. Otherwise we'll need to clean up these names later when all code paths use app server.

codex-rs/tui/src/app/skills.rs
```diff
@@ -0,0 +1,198 @@
...
+use codex_app_server_protocol::SkillsListParams;
+use codex_app_server_protocol::SkillsListResponse;
+use codex_protocol::protocol::ListSkillsResponseEvent;
+use codex_protocol::protocol::SkillErrorInfo;
+use std::path::Path;
+use std::path::PathBuf;
+use std::sync::atomic::AtomicI64;
+use std::sync::atomic::Ordering;
+
+static NEXT_APP_SERVER_REQUEST_ID: AtomicI64 = AtomicI64::new(1);
```
etraut-openai: I think this should be in a common module, not in skills.rs. Perhaps move this to the app-server-client module?

codex-rs/tui/src/app/skills.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+}
+
+/// Convert an app-server protocol skills entry into the core protocol
+/// equivalent so the existing `ChatWidget` skills UI can consume it
+/// unchanged. Each nested type has a corresponding `into_core_*` helper
+/// below; these are intentionally field-by-field rather than derived to
+/// keep the two protocol crates decoupled and make field divergence a
+/// compile error.
+fn into_core_skills_list_entry(
```
etraut-openai: I understand that these conversion routines are here to avoid the need to modfiy the TUI code that consumes these structures, but we're going to need to change that code anyway. Is there a reason not to do it as part of this PR?

codex-rs/tui/src/chatwidget/skills.rs
```diff
@@ -142,6 +142,14 @@ impl ChatWidget {
         self.skills_all = skills;
         self.set_skills(Some(enabled_skills_for_mentions(&self.skills_all)));
     }
+
+    #[cfg(test)]
+    pub(crate) fn loaded_skill_names(&self) -> Vec<String> {
```
etraut-openai, line 142: This test isn't verifying anything other than that the call doesn't crash. Is it worth it?

codex-rs/tui/src/chatwidget/tests.rs
```diff
@@ -1774,6 +1775,17 @@ async fn make_chatwidget_manual(
     ChatWidget,
     tokio::sync::mpsc::UnboundedReceiver<AppEvent>,
     tokio::sync::mpsc::UnboundedReceiver<Op>,
+) {
+    make_chatwidget_manual_with_use_app_server(model_override, false).await
+}
+
+async fn make_chatwidget_manual_with_use_app_server(
```
etraut-openai: Nit: Can we find a better name for this function?

codex-rs/tui/src/chatwidget/tests.rs
```diff
@@ -1941,6 +1954,45 @@ fn assert_no_submit_op(op_rx: &mut tokio::sync::mpsc::UnboundedReceiver<Op>) {
     }
 }
 
+#[tokio::test]
+async fn list_skills_without_app_server_uses_core_op_channel() {
```
etraut-openai: Why are we adding tests for the legacy path that we're removing?

codex-rs/tui/src/chatwidget/tests.rs
```diff
@@ -1941,6 +1954,45 @@ fn assert_no_submit_op(op_rx: &mut tokio::sync::mpsc::UnboundedReceiver<Op>) {
...
+        Ok(Op::ListSkills {
+            cwds,
+            force_reload: true,
+        }) if cwds.is_empty()
+    );
+    assert_matches!(app_rx.try_recv(), Err(TryRecvError::Empty));
+}
+
+#[tokio::test]
+async fn list_skills_with_app_server_uses_app_event_channel() {
```
etraut-openai: If we're cutting over to app_server unconditionally, then we don't need "app_server" in these method names.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -482,6 +482,9 @@ pub(crate) struct ChatWidgetInit {
     // Shared latch so we only warn once about invalid status-line item IDs.
     pub(crate) status_line_invalid_items_warned: Arc<AtomicBool>,
     pub(crate) session_telemetry: SessionTelemetry,
+    /// When true, route operations that have migrated to the embedded
+    /// app-server RPC instead of codex-core's op channel.
+    pub(crate) use_app_server: bool,
```
etraut-openai: If we're going to do a hard cut-over, then we don't need this.

codex-rs/tui/src/lib.rs
```diff
@@ -996,12 +996,13 @@ async fn run_ratatui_app(
...
         images,
         no_alt_screen,
+        app_server,
         ..
     } = cli;
 
     let use_alt_screen = determine_alt_screen_mode(no_alt_screen, config.tui_alternate_screen);
     tui.set_alt_screen_enabled(use_alt_screen);
-    let app_server = match start_embedded_app_server(
+    let embedded_app_server = match start_embedded_app_server(
```
etraut-openai: Why is this renamed? I'd prefer to stick with "app_server". We're going to want to support remote app servers as well.

etraut-openai (commented):
First round of review comments

fcoury:
@codex review

PR #14616, opened 2026-03-13
Fix turn context reconstruction after backtracking

charley-oai:
## Summary
- reuse rollout reconstruction when applying a backtrack rollback so `reference_context_item` is restored from persisted rollout state
- build rollback replay from the flushed rollout items plus the rollback marker, avoiding the extra reread/fallback path
- add regression coverage for rollback after compaction so turn-context diffing stays aligned after backtracking

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -2008,17 +2008,9 @@ impl Session {
             }
             InitialHistory::Resumed(resumed_history) => {
                 let rollout_items = resumed_history.history;
-
-                let reconstructed_rollout = self
-                    .reconstruct_history_from_rollout(&turn_context, &rollout_items)
+                let previous_turn_settings = self
+                    .apply_rollout_reconstruction(&turn_context, &rollout_items)
                     .await;
-                let previous_turn_settings = reconstructed_rollout.previous_turn_settings.clone();
-                self.set_previous_turn_settings(previous_turn_settings.clone())
-                    .await;
-                {
-                    let mut state = self.state.lock().await;
-                    state.set_reference_context_item(reconstructed_rollout.reference_context_item);
-                }
 
                 // If resuming, warn when the last recorded model differs from the current one.
                 let curr: &str = turn_context.model_info.slug.as_str();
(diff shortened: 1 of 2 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -2040,13 +2032,6 @@ impl Session {
                     .await;
                 }
 
-                // Always add response items to conversation history
-                let reconstructed_history = reconstructed_rollout.history;
-                if !reconstructed_history.is_empty() {
-                    self.record_into_history(&reconstructed_history, &turn_context)
-                        .await;
-                }
-
```
charley-oai, line 2049: We now rely on `replace_history` in `apply_rollout_reconstruction` helper

codex-rs/core/src/codex.rs
```diff
@@ -2061,26 +2046,8 @@ impl Session {
...
-                        reconstructed_rollout.reference_context_item.clone(),
-                    );
-                }
-
-                // Always add response items to conversation history
-                let reconstructed_history = reconstructed_rollout.history;
-                if !reconstructed_history.is_empty() {
-                    self.record_into_history(&reconstructed_history, &turn_context)
-                        .await;
-                }
```
charley-oai, line 2083: We now rely on `replace_history` in `apply_rollout_reconstruction` helper

codex-rs/core/src/codex.rs
```diff
@@ -4968,29 +4940,22 @@ mod handlers {
...
+        sess.persist_rollout_items(&[RolloutItem::EventMsg(rollback_msg.clone())])
             .await;
-        sess.replace_history(
-            reconstructed.history,
-            reconstructed.reference_context_item.clone(),
-        )
-        .await;
-        sess.set_previous_turn_settings(reconstructed.previous_turn_settings)
+        sess.flush_rollout().await;
+        sess.apply_rollout_reconstruction(turn_context.as_ref(), replay_items.as_slice())
```
charley-oai, line 4952: This will properly set the `reference_turn_context`

pakrym-oai (approved):

PR #14619, opened 2026-03-13
fix: persist future network host approvals across sessions

viyatb-oai:
## Summary
- apply persisted execpolicy network rules when booting the managed network proxy
- pass the current execpolicy into managed proxy startup so host approvals selected with "allow this host in the future" survive new sessions


codex-rs/core/src/config/network_proxy_spec.rs
```diff
@@ -279,6 +296,30 @@ impl NetworkProxySpec {
...
+            host,
+        );
+    }
+    for host in denied_domains {
+        upsert_network_domain(
+            &mut config.network.denied_domains,
+            &mut config.network.allowed_domains,
+            host,
+        );
+    }
```
bolinfest: Should we be using `HashSet` or `BTreeSet` and merge collections which would be better big-O wise?

codex-rs/core/src/codex.rs
```diff
@@ -1156,12 +1156,22 @@ impl Session {
...
     async fn start_managed_network_proxy(
         spec: &crate::config::NetworkProxySpec,
+        exec_policy: &codex_execpolicy::Policy,
         sandbox_policy: &SandboxPolicy,
         network_policy_decider: Option<Arc<dyn codex_network_proxy::NetworkPolicyDecider>>,
         blocked_request_observer: Option<Arc<dyn codex_network_proxy::BlockedRequestObserver>>,
         managed_network_requirements_enabled: bool,
         audit_metadata: NetworkProxyAuditMetadata,
     ) -> anyhow::Result<(StartedNetworkProxy, SessionNetworkProxyRuntime)> {
+        let spec = match spec.with_exec_policy_network_rules(exec_policy) {
```
bolinfest: I think `.map_err()` would be cleaner here?

bolinfest (approved):

codex-rs/core/src/config/network_proxy_spec.rs
viyatb-oai: yep you're correct, fixing

PR #14622, opened 2026-03-13
Stabilize multi-agent feature flag

aibrahim-oai:
- make multi_agent stable and enabled by default
- update feature and tool-spec coverage to match the new default


codex-rs/core/src/features.rs
```diff
@@ -700,12 +700,8 @@ pub const FEATURES: &[FeatureSpec] = &[
     FeatureSpec {
         id: Feature::Collab,
         key: "multi_agent",
```
sayan-oai, line 713: feature name stays `multi_agent`, not `subagents`? i know backcompat is an issue, just checking this is fine.

sayan-oai (approved):

PR #14634, opened 2026-03-13
Use git hooks for commit attribution

gabec-openai:
## Summary

- move commit attribution back from developer-message injection to runtime git hook injection
- preserve current `commit_attribution` config semantics:
  - unset => default `Codex <noreply@openai.com>`
  - empty string => disable attribution
  - non-empty string => use verbatim
  - malformed non-string TOML => ignore the setting and fall back to the default attribution
- thread the injected git hook config through normal shell execution, unified exec, and shell snapshot env replay

### Rust checks

- `just fmt`
- `just fix -p codex-core`
- `just write-config-schema`

### Targeted Rust tests

- `cargo test -p codex-core commit_attribution -- --nocapture`
- `cargo test -p codex-core config_toml_ignores_malformed_commit_attribution -- --nocapture`
- `env -u CODEX_SANDBOX -u CODEX_SANDBOX_NETWORK_DISABLED cargo test -p codex-core unified_exec_intercepts_apply_patch_exec_command --test all -- --nocapture`

### Temp-workspace `codex exec` validation

I validated the runtime behavior using isolated temp git repos plus isolated temp `CODEX_HOME` directories so each config variant ran without cross-test contamination:

- built and used the local debug binary from `codex-rs/target/debug/codex`
- for each variant, created a fresh temp repo and a fresh temp `CODEX_HOME`
- copied `auth.json` into the temp `CODEX_HOME` so `codex exec` could run normally
- used `codex exec --json --ephemeral --full-auto ...` to create a real git commit inside the temp repo
- inspected the resulting commit message and the generated hook state

Validated config variants:

- default config with no `commit_attribution` override:
  - commit message ended with `Co-authored-by: Codex <noreply@openai.com>`
- `commit_attribution = "AgentX <agent@example.com>"`:
  - commit message ended with `Co-authored-by: AgentX <agent@example.com>`
- `commit_attribution = ""`:
  - no co-author trailer was added
  - no hook was created
- malformed `commit_attribution = true`:
  - config load remained non-fatal
  - the malformed setting was ignored
  - commit message fell back to `Co-authored-by: Codex <noreply@openai.com>`

### Broader test context

- `env -u CODEX_SANDBOX -u CODEX_SANDBOX_NETWORK_DISABLED cargo test -p codex-core --test all`
  - one unrelated test still fails locally:
    - `suite::code_mode::code_mode_exposes_normalized_illegal_mcp_tool_names`
  - `suite::unified_exec::unified_exec_intercepts_apply_patch_exec_command` failed once in the full run but passed when rerun in isolation
- `env -u CODEX_SANDBOX -u CODEX_SANDBOX_NETWORK_DISABLED cargo test`
  - workspace run was started, but unrelated failures in `codex-app-server` tracing tests showed the full local workspace suite was already red outside this change

### DIFF
--- a/codex-rs/core/src/commit_attribution.rs
+++ b/codex-rs/core/src/commit_attribution.rs
@@ -22,30 +22,28 @@ const COMMIT_HOOK_NAMES: &[&str] = &[
 pub(crate) fn configure_git_hooks_env_for_config(
     env: &mut HashMap<String, String>,
     config: &Config,
-) {
+) -> Vec<(String, String)> {
     configure_git_hooks_env(
         env,
         config.codex_home.as_path(),
         config.commit_attribution.as_deref(),
-    );
+    )
 }
 
 pub(crate) fn configure_git_hooks_env(
     env: &mut HashMap<String, String>,
     codex_home: &Path,
     config_attribution: Option<&str>,
-) {
-    let Some(value) = resolve_attribution_value(config_attribution) else {
-        return;
+) -> Vec<(String, String)> {
+    let Some((key, value)) = git_hooks_runtime_config(codex_home, config_attribution) else {
+        return Vec::new();
     };
 
-    let Ok(hooks_path) = ensure_codex_hook_scripts(codex_home, &value) else {
-        return;
-    };
-
-    set_git_runtime_config(env, "core.hooksPath", hooks_path.to_string_lossy().as_ref());
+    set_git_runtime_config(env, &key, &value);
+    vec![(key, value)]
 }
 
+#[cfg(test)]
 pub(crate) fn injected_git_config_env(env: &HashMap<String, String>) -> Vec<(String, String)> {
     let mut pairs = env
         .iter()
@@ -72,6 +70,18 @@ fn resolve_attribution_value(config_attribution: Option<&str>) -> Option<String>
         }
         None => Some(DEFAULT_ATTRIBUTION_VALUE.to_string()),
     }
+}
+
+fn git_hooks_runtime_config(
+    codex_home: &Path,
+    config_attribution: Option<&str>,
+) -> Option<(String, String)> {
+    let value = resolve_attribution_value(config_attribution)?;
+    let hooks_path = ensure_codex_hook_scripts(codex_home, &value).ok()?;
+    Some((
+        "core.hooksPath".to_string(),
+        hooks_path.to_string_lossy().into_owned(),
+    ))
 }
 
 fn ensure_codex_hook_scripts(codex_home: &Path, value: &str) -> std::io::Result<PathBuf> {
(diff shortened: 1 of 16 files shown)


codex-rs/core/src/config/mod.rs
```diff
@@ -1453,6 +1413,29 @@ enum WebSearchToolConfigInput {
     Config(WebSearchToolConfig),
 }
 
+#[derive(Deserialize)]
+#[serde(untagged)]
+enum CommitAttributionInput {
+    String(String),
+    // Ignore malformed values so a bad local override does not prevent Codex from starting.
+    Ignored(IgnoredAny),
```
pakrym-oai, line 1421: Why are we specifically ignoring invalid values for commit_attribution?

codex-rs/core/src/tools/handlers/shell.rs
```diff
@@ -332,6 +326,9 @@ impl ShellHandler {
         } = args;
 
         let mut exec_params = exec_params;
+        if session.features().enabled(Feature::CodexGitCommit) {
+            configure_git_hooks_env_for_config(&mut exec_params.env, turn.config.as_ref());
```
pakrym-oai: How does set_git_runtime_config work with git config set via shell profile in shell snapshot?

pakrym-oai (changes requested):
I might be missing something but feels like with the current setup set_git_runtime_config will ignore the shell snapshot.

codex-rs/core/src/commit_attribution.rs
```diff
@@ -28,6 +74,65 @@ fn resolve_attribution_value(config_attribution: Option<&str>) -> Option<String>
...
+        format!(
+            "\nmsg_file=\"${{1:-}}\"\nif [[ -n \"$msg_file\" && -f \"$msg_file\" ]]; then\n  git interpret-trailers \\\n    --in-place \\\n    --if-exists doNothing \\\n    --if-missing add \\\n    --trailer 'Co-authored-by={escaped_value}' \\\n    \"$msg_file\" || true\nfi\n"
+        )
+    } else {
+        String::new()
+    };
+
+    format!(
+        "#!/usr/bin/env bash\nset -euo pipefail\n\nunset GIT_CONFIG_COUNT\nwhile IFS='=' read -r name _; do\n  case \"$name\" in\n    GIT_CONFIG_KEY_*|GIT_CONFIG_VALUE_*) unset \"$name\" ;;\n  esac\ndone < <(env)\n\nexisting_hooks_path=\"$(git config --path core.hooksPath 2>/dev/null || true)\"\nif [[ -z \"$existing_hooks_path\" ]]; then\n  git_dir=\"$(git rev-parse --git-common-dir 2>/dev/null || git rev-parse --git-dir 2>/dev/null || true)\"\n  if [[ -n \"$git_dir\" ]]; then\n    existing_hooks_path=\"$git_dir/hooks\"\n  fi\nfi\n\nif [[ -n \"$existing_hooks_path\" ]]; then\n  existing_hook=\"$existing_hooks_path/{hook_name}\"\n  if [[ -x \"$existing_hook\" && \"$existing_hook\" != \"$0\" ]]; then\n    \"$existing_hook\" \"$@\"\n  fi\nfi\n{prepare_commit_msg_body}"
```
pakrym-oai: can we format this better? as a multiline string maybe? same for above

codex-rs/core/src/config/mod.rs
gabec-openai: basically the downstream apps are very brittle to config parse failures, so i wanted to have some graceful fallback. i could the argument for removing attribution altogether on config failure versus reverting to default. let me know if you feel strongly we should fail tho.

codex-rs/core/src/tools/handlers/shell.rs
gabec-openai: good catch, fixed!

codex-rs/core/tests/suite/mod.rs
```diff
@@ -68,6 +68,8 @@ mod client_websockets;
 mod code_mode;
 mod codex_delegate;
 mod collaboration_instructions;
+#[cfg(not(target_os = "windows"))]
```
pakrym-oai, line 71: what about windows?

PR #14643, opened 2026-03-13
Add exec-server transport for unified exec

starr-openai:
## Summary
- add a new `codex-exec-server` crate that exposes a stdio JSON-RPC surface for exec, write, terminate, and streamed output notifications
- gate exec-server-backed unified exec routing behind `experimental_exec_server` and teach the CLI/core session bootstrap to launch the daemon
- update unified exec process plumbing and PTY process metadata so interactive sessions can be realized through the exec server, with smoke coverage for the new path

## Testing
- Not run locally as part of this PR

## Notes
- I manually verified the runtime path locally by enabling `experimental_exec_server` and confirming spawned shells were parented by `codex-exec-server`.

### DIFF
--- a/codex-rs/core/src/unified_exec/process_manager.rs
+++ b/codex-rs/core/src/unified_exec/process_manager.rs
@@ -50,6 +50,7 @@ use crate::unified_exec::process::OutputBuffer;
 use crate::unified_exec::process::OutputHandles;
 use crate::unified_exec::process::SpawnLifecycleHandle;
 use crate::unified_exec::process::UnifiedExecProcess;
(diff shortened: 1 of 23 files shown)


codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -529,10 +530,32 @@ impl UnifiedExecProcessManager {
 
     pub(crate) async fn open_session_with_exec_env(
         &self,
+        process_id: i32,
```
starr-openai, line 533: why are we passing in a process_id here? it can only be returned. if a client-provided id it should be something else

PR #14651, opened 2026-03-14
Add argument-comment Dylint runner

bolinfest:
## Why

Call sites like `foo(false)` and `bar(None)` are often a sign that the API is not carrying enough meaning on its own. Contributors should prefer self-documenting designs such as enums, named methods, and newtypes when that is practical.

This PR adds a repo-local lint as a fallback for the places where a larger API refactor would be too heavy. The team also preferred a lighter comment shape than `clang-tidy`'s `bugprone-argument-comment`, so the convention here is exact `/*param*/` comments rather than `/*param=*/`.

## What changed

- add an isolated Dylint library under `tools/argument-comment-lint`
- add `argument_comment_mismatch`, which validates that an existing `/*param*/` comment matches the resolved parameter name
- add `uncommented_anonymous_literal_argument`, which targets opaque positional literals such as `None`, booleans, and numeric literals
- keep string and char literals exempt so obviously self-descriptive arguments do not need extra comment noise
- update the parser, diagnostics, examples, and UI coverage to use the exact `/*param*/` form
- add `tools/argument-comment-lint/run.sh`, a `just argument-comment-lint` entrypoint, and default `CARGO_INCREMENTAL=0` in the wrapper to avoid the current nightly Dylint incremental-compilation ICE in local runs
- add `AGENTS.md` guidance that pushes contributors toward better API design first and treats argument comments as the fallback when a smaller compatibility-preserving change is more appropriate

## Examples

Preferred when a positional literal still needs help at the callsite:

```rust
fn create_openai_url(base_url: Option<String>, retry_count: usize) -> String {
    String::new()
}

create_openai_url(/*base_url*/ None, /*retry_count*/ 3);
```

Rejected when the comment name is wrong:

```rust
fn create_openai_url(base_url: Option<String>) -> String {
    String::new()
}

create_openai_url(/*api_base*/ None);
```

Still intentionally allowed without comments for self-descriptive literals:

```rust
fn split_top_level(body: &str, delimiter: char) -> Vec<&str> {
    Vec::new()
}

fn describe(prefix: &str, suffix: &str) {}

split_top_level("a|b|c", '|');
describe("openai", "https://api.openai.com/v1");
```

## Verification

- `cargo test` in `tools/argument-comment-lint`
- `bash -n tools/argument-comment-lint/run.sh`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/14651).
* #14652
* __->__ #14651


etraut-openai (approved):
Thanks for adding this!

PR #14668, opened 2026-03-14
Reuse guardian session across approvals

charley-oai:
## Summary
- reuse a guardian subagent session across approvals so reviews keep a stable prompt cache key and avoid one-shot startup overhead
- clear the guardian child history before each review so prior guardian decisions do not leak into later approvals
- include the `smart_approvals` -> `guardian_approval` feature flag rename in the same PR to minimize release latency on a very tight timeline
- add regression coverage for prompt-cache-key reuse without prior-review prompt bleed

## Request
- Bug/enhancement request: internal guardian prompt-cache and latency improvement request


charley-oai:
@codex review this

charley-oai:
@codex review this

codex-rs/core/src/tools/network_approval.rs
```diff
@@ -186,8 +186,9 @@ impl Default for NetworkApprovalService {
 
 impl NetworkApprovalService {
     pub(crate) async fn copy_session_approved_hosts_to(&self, other: &Self) {
-        let approved_hosts = self.session_approved_hosts.lock().await;
+        let approved_hosts = self.session_approved_hosts.lock().await.clone();
         let mut other_approved_hosts = other.session_approved_hosts.lock().await;
+        other_approved_hosts.clear();
```
charley-oai, line 193: Before:

- `copy_session_approved_hosts_to()` only did `extend(...)`
- if the cached guardian child had old approved hosts, they stayed there forever

After:

- it clones the parent’s current approved-host set
- clears the child’s set
- repopulates it from the parent snapshot

That matters because the guardian session is now reused. Without this, the child could keep stale network approvals from earlier reviews and end
up with broader access than the parent session.

aibrahim-oai (approved):

PR #14674, opened 2026-03-14
fix: fix symlinked writable roots in sandbox policies

viyatb-oai:
## Summary
- normalize effective readable, writable, and unreadable sandbox roots after resolving special paths so symlinked roots use canonical runtime paths
- add a protocol regression test for a symlinked writable root with a denied child and update protocol expectations to canonicalized effective paths
- update macOS seatbelt tests to assert against effective normalized roots produced by the shared policy helpers

## Testing
- just fmt
- cargo test -p codex-protocol
- cargo test -p codex-core explicit_unreadable_paths_are_excluded_
- cargo clippy -p codex-protocol -p codex-core --tests -- -D warnings

## Notes
- This is intended to fix the symlinked TMPDIR bind failure in bubblewrap described in #14672.
Fixes #14672


codex-rs/protocol/src/permissions.rs
```diff
@@ -378,6 +378,7 @@ impl FileSystemSandboxPolicy {
                 .filter(|entry| self.can_read_path_with_cwd(entry.path.as_path(), cwd))
                 .map(|entry| entry.path)
                 .collect(),
+            true,
```
etraut-openai: Nit: The use of raw "true" and "false" arguments in these calls makes them hard to comprehend what's going on. Michael recently proposed that we use /*param_name*/ comments before such arguments to help with readability. He just merged https://github.com/openai/codex/pull/14651, which is a linter that enforces this. He also updated AGENTS.md to document this convention, but I think you generated this code with the old AGENTS.md.

etraut-openai (approved):

codex-rs/protocol/src/permissions.rs
viyatb-oai: good point, i'll update

PR #14678, opened 2026-03-14
v8-musl: Add custom build for v8 with musl

cconger:
Add a standalone codex-rs/v8-poc crate to prove that Cargo can link against the staged musl V8 bundle without wiring V8 into codex-core yet.

Teach rusty-v8-bazel-build to stage and content-cache the musl V8 artifacts, emit the Cargo environment needed to force the v8 crate to use our archive and binding, and build the PoC crate against that bundle in CI. This keeps the experiment isolated while proving the end-to-end consumer path we would use for a future Rust crate.


.github/scripts/rusty_v8_bazel.py
```diff
@@ -0,0 +1,641 @@
...
+    info.uid = 0
+    info.gid = 0
+    info.uname = ""
+    info.gname = ""
+    info.mtime = 0
+    with source.open("rb") as src:
+        tar.addfile(info, src)
+
+
+def write_bundle_archive(
```
zbarsky-openai, line 358: i feel like this can be done in bash more cleary/simply?

.github/scripts/rusty_v8_bazel.py
```diff
@@ -0,0 +1,641 @@
...
+                        continue
+                    relpath = header_path.relative_to(include_root.parent)
+                    add_regular_file_to_tar(
+                        tar,
+                        source=header_path,
+                        arcname=f"{bundle_root}/{relpath.as_posix()}",
+                    )
+
+
+def validate_bundle(platform: str, target: str, bundle_path: Path) -> None:
```
zbarsky-openai, line 409: what's the purpose of this function? it is hardcoding a TON of implentation details of both bazel and the llvm toolchain, it's gonna make upgrading anything a pain. can we not do this?

.github/scripts/rusty_v8_bazel.py
```diff
@@ -0,0 +1,641 @@
...
+        first_existing_path([archive], f"staged V8 archive for {target}"),
+        first_existing_path([binding], f"staged V8 binding for {target}"),
+    )
+
+
+def runtime_output_dir(output_dir: Path) -> Path:
+    return output_dir / "runtime"
+
+
+def staged_runtime_libs(output_dir: Path) -> tuple[Path, Path, Path]:
```
zbarsky-openai, line 258: why do we need all these implemtnation details

.github/scripts/rusty_v8_bazel.py
```diff
@@ -0,0 +1,641 @@
...
+    "libcxx_library_dir": "@llvm//runtimes/libcxx:libcxx_library_search_directory",
+    "libunwind_library_dir": "@llvm//runtimes/libunwind:libunwind_library_search_directory",
+    "musl_library_dir": "@llvm//runtimes/musl:musl_library_search_directory",
+    "libcxx_headers_dir": "@llvm//runtimes/libcxx:libcxx_headers_include_search_directory",
+    "libcxxabi_headers_dir": "@llvm//runtimes/libcxx:libcxxabi_headers_include_search_directory",
+    "musl_headers_dir": "@llvm//runtimes/musl:musl_headers_include_search_directory",
+}
+COMPILER_RT_BUILTINS_LABEL = "@llvm//runtimes/compiler-rt:clang_rt.builtins.static"
+COMPILER_RT_BUILTINS_FILENAME = "libclang_rt.builtins.static.a"
+V8_CACHE_KEY_INPUT_PATTERNS = [
```
zbarsky-openai, line 43: feels a little brittle, why not just kick this off on-demand? bazel remote cache should short-circuit the build if needed, github cache tends to just make things slower tbh

.github/workflows/rusty-v8-bazel-build.yml
```diff
@@ -0,0 +1,223 @@
...
+      - "patches/**"
+      - "third_party/v8/**"
+      - "**/*.bzl"
+      - "**/BUILD"
+      - "**/BUILD.bazel"
+  push:
+    branches:
+      - main
+    paths:
+      - ".bazelrc"
```
zbarsky-openai, line 23: i thought we just needed to produce these once/periodically when bumping v8, do we really need this on each main commit?

.github/workflows/rusty-v8-bazel-build.yml
```diff
@@ -0,0 +1,223 @@
...
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          version="$(python3 .github/scripts/rusty_v8_bazel.py print-version)"
+          version_suffix="${version//./_}"
+          target_suffix="${TARGET//-/_}"
+          lib_target="//third_party/v8:v8_${version_suffix}_${target_suffix}"
+          binding_target="//third_party/v8:src_binding_release_${target_suffix}"
+          runtime_targets=(
```
zbarsky-openai, line 102: i'm really confused what you're doing here :hmmmm:

third_party/v8/BUILD.bazel
```diff
@@ -0,0 +1,124 @@
...
+        -e 's|"v8/src/libplatform/default-platform.h"|"src/libplatform/default-platform.h"|' \
+        -e 's|  namespace i = v8::internal;|  (void)usage;|' \
+        -e '/using HelpOptions = i::FlagList::HelpOptions;/d' \
+        -e '/HelpOptions help_options = HelpOptions(HelpOptions::kExit, usage);/d' \
+        -e 's|  i::FlagList::SetFlagsFromCommandLine(argc, argv, true, help_options);|  v8::V8::SetFlagsFromCommandLine(argc, argv, true);|' \
+        $(location @v8_crate_146_4_0//:binding_cc) > "$@"
+    """,
+)
+
+genrule(
```
zbarsky-openai, line 22: use `copy_file` rule

third_party/v8/BUILD.bazel
```diff
@@ -0,0 +1,124 @@
...
+    name = "support_h",
+    srcs = ["@v8_crate_146_4_0//:support_h"],
+    outs = ["support.h"],
+    cmd = "cp $(location @v8_crate_146_4_0//:support_h) $@",
+)
+
+cc_library(
+    name = "v8_146_4_0_binding",
+    srcs = [":binding_cc"],
```
zbarsky-openai, line 31: nit: you can do `srcs = ["binding.cc"]`, it's a bit clearer

third_party/v8/BUILD.bazel
```diff
@@ -0,0 +1,124 @@
...
+    hdrs = [":support_h"],
+    copts = select({
+        "@v8//bazel/config:is_clang": ["-std=c++20"],
+        "@v8//bazel/config:is_gcc": ["-std=gnu++2a"],
+        "@v8//bazel/config:is_windows": ["/std:c++20"],
```
zbarsky-openai, line 36: umm what does `is_windows` actually check? have you verified? It's pretty common to assume that windows == msvc which is NOT our case. We use clang everywhere. I think you should just pass `copts = ["-std=c++20"]`, no?

third_party/v8/BUILD.bazel
```diff
@@ -0,0 +1,124 @@
...
+    deps = [
+        "@v8//:core_lib_icu",
+        "@v8//:rusty_v8_internal_headers",
+    ],
+)
+
+cc_library(
+    name = "v8_146_4_0_public_headers",
+    hdrs = ["@v8//:public_header_files"],
+    includes = ["../../external/v8+/include"],
```
zbarsky-openai, line 48: this seems off, it's including an implementation detail as the path. Does `rusty_v8_internal_headers` not provide its own proper `includes` attribute? maybe that needs patching instead?

third_party/v8/BUILD.bazel
```diff
@@ -0,0 +1,124 @@
...
+    deps = ["@v8//:rusty_v8_internal_headers"],
+)
+
+cc_binary(
+    name = "v8_146_4_0_smoke_test",
+    srcs = ["smoke_test.cc"],
+    copts = select({
+        "@v8//bazel/config:is_clang": ["-std=c++20"],
```
zbarsky-openai, line 56: ditto copts

third_party/v8/BUILD.bazel
```diff
@@ -0,0 +1,124 @@
...
+
+cc_library(
+    name = "v8_146_4_0_x86_64_unknown_linux_musl_imported",
+    linkopts = [
+        "-Wl,--no-as-needed",
+        "-ldl",
+        "-pthread",
+    ],
+    deps = [
+        ":v8_146_4_0_x86_64_unknown_linux_musl_archive",
```
zbarsky-openai, line 78: where is this coming from? Is this all importing the previously-built v8 from the release? If so - don't think we need that, just consume the targets directly as part of this build graph? Should be way simpler

MODULE.bazel
```diff
@@ -174,6 +176,253 @@ crate.annotation(
...
+    'bool dump_stack_in_signal_handler = true;\\n\\n#if HAVE_EXECINFO_H\\n// The prefix used for mangled symbols, per the Itanium C++ ABI:\\n// http://www.codesourcery.com/cxx-abi/abi.html#mangling\\nconst char kMangledSymbolPrefix[] = "_Z";\\n\\n// Characters that can be used for symbols, generated by Ruby:\\n// ((\\'a\\'..\\'z\\').to_a+(\\'A\\'..\\'Z\\').to_a+(\\'0\\'..\\'9\\').to_a + [\\'_\\']).join\\nconst char kSymbolCharacters[] =\\n    "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_";\\n\\n',
+    1,
+)
+stack_trace.write_text(text)
+
+platform_posix = Path("src/base/platform/platform-posix.cc")
+text = platform_posix.read_text()
+text = text.replace(
+    '#if defined(V8_LIBC_GLIBC)\\nextern "C" void* __libc_stack_end;\\n#endif\\n',
```
zbarsky-openai, line 361: this feels like a codex hack. Shouldn't we be tweaking the BUILD to properly pass V8_LIBC_GLIBC on glibc and not pass that on musl? Have you checked that hacking thesse out is actually the right behavior?

MODULE.bazel
```diff
@@ -174,6 +176,253 @@ crate.annotation(
 
 inject_repo(crate, "alsa_lib")
 
+bazel_dep(name = "v8", version = "14.6.202.11")
+archive_override(
+    module_name = "v8",
+    integrity = "sha256-Ju45oFJMyCWyT/r+d+MDsCbCeoDBisGaj4+KZYIIYSU=",
+    patch_cmds = ["""
```
zbarsky-openai, line 183: at the very least we should use actual patch files, not patch_cmd. But alo why do we need all these changes? Is there an explanation somewhere of your investgiation and what you've discovered, what are the main issues this PR is trying to fix, etc?

PR #14699, opened 2026-03-14
feat(tui): route fresh-session thread lifecycle through app-server

fcoury:
## Motivation

The TUI's thread lifecycle was wired directly to core, bypassing app-server entirely for event delivery. This meant fresh sessions didn't go through the app-server `thread/start` RPC, so server-side notifications — like skills-changed events that drive `$`-prompt completions and custom prompt listings — could race or never arrive, leaving users with missing skills and prompts after starting a new session.

The parent branch (`feat/tui-gated-app-server`) made app-server always-on and routed skills reloads through its notification stream. This branch extends that by moving **fresh-session thread lifecycle** onto app-server as well, so events flow through a single transport rather than a parallel direct-core path.

Resume and fork sessions are intentionally left on the direct-core transport for now. Fresh sessions are the simplest and most common path, so migrating them first keeps the blast radius small and proves the new transport before touching session-restore logic, which has its own edge cases around replay and fork metadata. Phase 2 (resume/fork migration) is tracked but not in scope here.

## Problem

Thread bootstrap was duplicated across three `ChatWidget` constructors (`new`, `new_from_existing`, and `spawn_agent` / `spawn_agent_from_existing`), each carrying ~150 lines of near-identical widget initialization. Worse, `ChatWidget` owned both event sourcing (spawning listener tasks) and session state application, making it impossible for `App` to reason about which threads have active listeners and how events reach the main loop.

## Mental model

Threads now have a **transport** that determines how their events reach `App`:

- **AppServerManaged** — fresh sessions go through `thread/start` RPC; events arrive on the app-server `LegacyNotification` stream and are routed via `legacy_notification_to_thread_event`.
- **DirectCore** — resume/fork sessions hold a direct `Arc<CodexThread>`; a spawned tokio task polls `next_event()` and forwards as `AppEvent::ThreadEvent`.

`App` owns the lifecycle (channel registration, listener spawn, activation). `ChatWidget` only configures its rendering state from a `SessionConfiguredEvent` and starts a fire-and-forget op-forwarder — it no longer subscribes to events or starts threads.

The old `on_session_configured` handler splits into `apply_session_configured_state` (pure state mutation) and `run_session_bootstrap_side_effects` (ops and redraw). Constructors call both synchronously, so the widget is fully configured before the first render frame.

## Stacking context

This PR targets `feat/tui-gated-app-server`, which made app-server always-on and routed skills reloads through server notifications. That branch established the plumbing; this one uses it for fresh-session thread lifecycle.

```
main
 └─ feat/tui-gated-app-server
     │  App-server always-on (removed --app-server flag)
     │  Skills routing through app-server notifications
     │  App-server skill types throughout TUI
     │
     └─ feat/tui-app-server-thread-events  ← this PR
          Fresh-session thread lifecycle via app-server
          Thread event ownership moved from ChatWidget → App
          Resume/fork intentionally still direct-core
```

## Non-goals

- Migrating resume/fork paths to app-server RPCs (those still use direct-core — phase 2).
- Replacing `LegacyNotification` with typed server notifications.
- Changing observable behavior — all three startup paths (fresh / resume / fork) produce identical user-visible state.

## Tradeoffs

- **Duplicated `ThreadId` parsing**: `session_configured_from_thread_start_response` re-parses the thread id even though `build_fresh_app_server_session` already parsed it. Accepted for clarity; the cost is a UUID parse.
- **`event.clone()` in `apply_session_configured_state`**: The `SessionConfiguredEvent` is cloned once for the session-info history cell. Unavoidable without restructuring the cell builder to borrow.
- **`suppress_session_configured_redraw` removed**: The timing issue it addressed (redundant redraw on resume) no longer exists because state is applied synchronously. Constructors now always request a redraw via `run_session_bootstrap_side_effects`.
- **`AppEvent::FatalExitRequest` removed**: Startup failure is now reported as an inline error message in the chat widget rather than a hard exit, which is arguably more graceful.

## Architecture

```
App::build_fresh_app_server_session()
  -> thread/start RPC -> ThreadStartResponse
  -> session_configured_from_thread_start_response() -> SessionConfiguredEvent
  -> ChatWidget::new_from_existing_app_server() -> widget

App::attach_app_server_managed_thread()
  -> register_thread_event_channel() (no listener)

App::attach_direct_thread()  [resume/fork]
  -> register_thread_event_channel()
  -> spawn_direct_thread_event_listener()

App-server event loop:
  LegacyNotification -> legacy_notification_to_thread_event()
    -> AppEvent::ThreadEvent { thread_id, event }
```

## Observability

- `tracing::debug` on direct-thread listener stop (existing).
- `tracing::warn` on rejected app-server requests (existing).
- No new metrics; existing thread-event channel machinery is reused.

## Tests

- `attach_app_server_managed_thread_registers_channel_without_listener` — channel present, no listener task.
- `attach_direct_thread_registers_channel_and_listener` — both channel and listener.
- `session_configured_from_thread_start_response_maps_thread_start_fields` — round-trip mapping correctness.
- `build_fresh_app_server_session_returns_bootstrapped_widget` — end-to-end through the app-server.
- `session_configured_bootstrap_updates_state_and_requests_prompts_and_skills` — ChatWidget state + side effects.
- `app_server_existing_thread_constructor_applies_bootstrap_state` — constructor applies session state.
- 3 tests for `legacy_notification_to_thread_event` (happy path, missing id, bad payload).


fcoury:
@codex review against fcoury:feat/tui-gated-app-server

PR #14727, opened 2026-03-15
Use workspace requirements for guardian prompt override

charley-oai:
## Summary
- move `guardian_developer_instructions` from managed config into workspace-managed `requirements.toml`
- have guardian continue using the override when present and otherwise fall back to the bundled local guardian prompt
- keep the generalized prompt-quality improvements in the shared guardian default prompt
- update requirements parsing, layering, schema, and tests for the new source of truth

## Context
This replaces the earlier managed-config / MDM rollout plan.

The intended rollout path is workspace-managed requirements, including cloud enterprise policies, rather than backend model metadata, Statsig, or Jamf-managed config. That keeps the default/fallback behavior local to `codex-rs` while allowing faster policy updates through the enterprise requirements plane.

This is intentionally an admin-managed policy input, not a user preference: the guardian prompt should come either from the bundled `codex-rs` default or from enterprise-managed `requirements.toml`, and normal user/project/session config should not override it.

## Updating The OpenAI Prompt
After this lands, the OpenAI-specific guardian prompt should be updated through the workspace Policies UI at `/codex/settings/policies` rather than through Jamf or codex-backend model metadata.

Operationally:
- open the workspace Policies editor as a Codex admin
- edit the default `requirements.toml` policy, or a higher-precedence group-scoped override if we ever want different behavior for a subset of users
- set `guardian_developer_instructions = """..."""` to the full OpenAI-specific guardian prompt text
- save the policy; codex-backend stores the raw TOML and `codex-rs` fetches the effective requirements file from `/wham/config/requirements`

When updating the OpenAI-specific prompt, keep it aligned with the shared default guardian policy in `codex-rs` except for intentional OpenAI-only additions.

## Testing
- `cargo check --tests -p codex-core -p codex-config -p codex-cloud-requirements --message-format short`
- `cargo run -p codex-core --bin codex-write-config-schema`
- `cargo fmt`
- `git diff --check`


charley-oai:
@codex review this

alexsong-oai (approved):

PR #14732, opened 2026-03-15
[apps] Improve search tool fallback.

mzeng-openai:
- [x] Bypass tool search and stuff tool specs directly into model context when either a. Tool search is not available for the model or b. There are not that many tools to search for.

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -377,6 +377,7 @@ pub(crate) const INITIAL_SUBMIT_ID: &str = "";
 pub(crate) const SUBMISSION_CHANNEL_CAPACITY: usize = 512;
 const CYBER_VERIFY_URL: &str = "https://chatgpt.com/cyber";
 const CYBER_SAFETY_URL: &str = "https://developers.openai.com/codex/concepts/cyber-safety";
+const DIRECT_APP_TOOL_EXPOSURE_THRESHOLD: usize = 100;
 
 impl Codex {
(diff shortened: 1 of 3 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -6490,6 +6489,21 @@ pub(crate) async fn built_tools(
         mcp_tools = selected_mcp_tools;
     }
 
+    // Expose app tools directly when tool_search is disabled, or when tool_search
+    // is enabled but the accessible app tool set stays below the direct-exposure threshold.
+    let expose_app_tools_directly = !turn_context.tools_config.search_tool
+        || app_tools
+            .as_ref()
+            .is_some_and(|tools| tools.len() < DIRECT_APP_TOOL_EXPOSURE_THRESHOLD);
```
apanasenko-oai, line 6498: if we want to use threshold as way to disable search tool, we need to move it to `turn_context.tools_config.search_tool` otherwise we will have situation when we have search tool and tools injected into context.

mzeng-openai: Tool search will not be enabled if app_tools is empty: https://github.com/openai/codex/blob/49edf311ac3ae84659b0ec5eacd5e471c881eee8/codex-rs/core/src/tools/spec.rs#L2734

apanasenko-oai: ah, I see, I would love this logic be much simpler since it hard to keep in mind all this invariants.

apanasenko-oai (approved):

PR #14759, opened 2026-03-15
Build v8 with rust bindings through bazel

cconger:
This PR add v8 through Bazel-first and splits it into two workflows.
1. On PRs that touch the v8/bazel suface we rebuild the artifacts and test them (eventually this will be done through an actual dep + test suite and bazel caching on v8.
2. Actual release packaging.  This will build and release bundled release pairs as artifacts that can be depended on by our cargo build pipeline.

This canary build runs a canary matrix: GNU Linux runs the Bazel test target for v8-poc, and musl Linux builds and executes the smoke_test binary. That gives us a cheap signal that the Bazel-native consumer path is still healthy before anything is released.

We also are adding a new release target (2) that will let us build a minimal dependency set per target triplet that we can depend on for cargo builds.


third_party/v8/BUILD.bazel
```diff
@@ -0,0 +1,279 @@
...
+        -e 's|"v8/src/libplatform/default-platform.h"|"src/libplatform/default-platform.h"|' \
+        -e 's|  namespace i = v8::internal;|  (void)usage;|' \
+        -e '/using HelpOptions = i::FlagList::HelpOptions;/d' \
+        -e '/HelpOptions help_options = HelpOptions(HelpOptions::kExit, usage);/d' \
+        -e 's|  i::FlagList::SetFlagsFromCommandLine(argc, argv, true, help_options);|  v8::V8::SetFlagsFromCommandLine(argc, argv, true);|' \
+        $(location @v8_crate_146_4_0//:binding_cc) > "$@"
+    """,
+)
+
+genrule(
```
cconger: I looked into replacing this, we don't depend on @bazel_skylib yet so we do not have the copy_file rule.

Also just referencing `@v8_crate_146_4_0//:support_h` from hdrs below seemed to fail.

.github/workflows/rusty-v8-release.yml
```diff
@@ -0,0 +1,178 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+
+      - name: Provide python3 shim (Windows)
+        if: runner.os == 'Windows'
+        shell: pwsh
+        run: |
+          $shim_dir = Join-Path $env:RUNNER_TEMP "python3-shim"
```
zbarsky-openai: i think you can just cross-build all of these instead of dealing with windows hosts. It should just work. (Just set `--platforms` appropriately

.github/workflows/rusty-v8-release.yml
```diff
@@ -0,0 +1,178 @@
...
+          bazel_args=(
+            build
+            "--platforms=@llvm//platforms:${PLATFORM}"
+            "${pair_target}"
+            --build_metadata=REPO_URL=https://github.com/${GITHUB_REPOSITORY}.git
+            --build_metadata=COMMIT_SHA=$(git rev-parse HEAD)
+            --build_metadata=ROLE=CI
+            --build_metadata=VISIBILITY=PUBLIC
+          )
+          if [[ -n "${BAZEL_REPO_CONTENTS_CACHE:-}" ]]; then
```
zbarsky-openai: where do these come from and why are they optional?

.github/workflows/rusty-v8-release.yml
```diff
@@ -0,0 +1,178 @@
...
+          bazel_cmd=(bazel)
+          if [[ -n "${BAZEL_STARTUP_ARGS:-}" ]]; then
+            bazel_cmd+=("${BAZEL_STARTUP_ARGS}")
+          fi
+
+          bazel_args=(
+            build
+            "--platforms=@llvm//platforms:${PLATFORM}"
+            "${pair_target}"
+            --build_metadata=REPO_URL=https://github.com/${GITHUB_REPOSITORY}.git
```
zbarsky-openai, line 78: why are these manually created annotations? Shouldn't it just come from existing ci.bazelrc or similar?

patches/v8_bazel_rules.patch
```diff
@@ -0,0 +1,226 @@
...
+-                "-Wno-invalid-offsetof",
+-                "-Wno-deprecated-this-capture",
+-                "-Wno-deprecated-declarations",
+-                "-std=c++20",
+-            ],
+             "@v8//bazel/config:is_gcc": [
+                 "-Wno-extra",
+                 "-Wno-array-bounds",
+@@ -155,7 +149,12 @@ def _default_args():
+             "@v8//bazel/config:is_windows": [
```
zbarsky-openai: i left a comment on previous version, I don't think this `is_windows` check is correct. we use clang on windows as well, not msvc.

patches/v8_bazel_rules.patch
```diff
@@ -0,0 +1,226 @@
...
++                "-Wno-invalid-offsetof",
++                "-Wno-deprecated-this-capture",
++                "-Wno-deprecated-declarations",
++                "-std=c++20",
++            ],
+         }) + select({
+             "@v8//bazel/config:is_gcc_fastbuild": [
+                 # Non-debug builds without optimizations fail because
+@@ -184,7 +183,7 @@ def _default_args():
+                 "Advapi32.lib",
```
zbarsky-openai: this probably not gonna link on windows (`.lib` is an MSVC-ism)

patches/v8_bazel_rules.patch
```diff
@@ -0,0 +1,226 @@
...
+diff --git a/orig/v8-14.6.202.11/BUILD.bazel b/mod/v8-14.6.202.11/BUILD.bazel
+index 85f31b7..7314584 100644
+--- a/orig/v8-14.6.202.11/BUILD.bazel
++++ b/mod/v8-14.6.202.11/BUILD.bazel
+@@ -303,7 +303,7 @@ v8_int(
+ # If no explicit value for v8_enable_pointer_compression, we set it to 'none'.
+ v8_string(
+     name = "v8_enable_pointer_compression",
+-    default = "none",
++    default = "False",
```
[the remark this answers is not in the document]
zbarsky-openai: so this is interesting...do we want to enable it? It's a performance/memory win as long as you can stay within the compessed size (2gb I think). Depends on what you're using this for!

patches/v8_bazel_rules.patch
```diff
@@ -0,0 +1,226 @@
...
+         "include/inspector/Debugger.h",
+@@ -4426,15 +4426,18 @@ genrule(
+         "src/inspector/protocol/Schema.cpp",
+         "src/inspector/protocol/Schema.h",
+     ],
+-    cmd = "$(location :code_generator) --jinja_dir . \
+-           --inspector_protocol_dir third_party/inspector_protocol \
++    cmd = "INSPECTOR_PROTOCOL_DIR=$$(dirname $(execpath third_party/inspector_protocol/code_generator.py)); \
++           PYTHONPATH=$$INSPECTOR_PROTOCOL_DIR:external/rules_python++pip+v8_python_deps_311_jinja2/site-packages:external/rules_python++pip+v8_python_deps_311_markupsafe/site-packages:$${PYTHONPATH-} \
++           python3 $(execpath third_party/inspector_protocol/code_generator.py) --jinja_dir . \
```
zbarsky-openai: non-hermetic python3 invocation. Should be using python toolchain instead. actually it seems like all of this should be a `py_binary` instead of hackily constructing a PYTHONPATH...

patches/v8_custom_archive_out_dir.patch
```diff
@@ -0,0 +1,141 @@
...
++      for base in [runfiles_dir.join("_main"), runfiles_dir] {
++        let candidate = base.join(&path);
++        if candidate.exists() {
++          return candidate.to_string_lossy().into_owned();
++        }
++      }
++    }
++  }
++
++  if let Ok(runfiles_dir) = env::var("RUNFILES_DIR") {
```
zbarsky-openai: do we actually need runfiles support? (this doesn't even implement it correctly., codex loves to write hacky + incorrect runfiles implementations. I guess poor training data...). I think it would be better to just pass $execpath and use the path as-is

.github/workflows/rusty-v8-release.yml
cconger: I think I kinda painted myself into a corner here, by requiring needing to run the binary on the host to validate, but I'll figure out how to get CI confidence here.

patches/v8_source_portability.patch
```diff
@@ -0,0 +1,78 @@
...
++    }
+ #else
+     return nullptr;
+ #endif  // !defined(V8_LIBC_GLIBC)
++    return nullptr;
+   }
+   void* base;
+   size_t size;
+@@ -1476,7 +1479,8 @@
+   // __libc_stack_end is process global and thus is only valid for
```
zbarsky-openai, line 56: i did not validate this part, please do the due diligence here

third_party/v8/BUILD.bazel
zbarsky-openai: you can just pull it in, it's 1 line in MODULE.bazel and you already depend on it transitively, i'm positive :)

third_party/v8/BUILD.bazel
```diff
@@ -0,0 +1,275 @@
...
+save
+end
+EOF
+      $(location @llvm//tools:llvm-ar) -M < "$(@D)/merge.mri"
+      $(location @llvm//tools:llvm-ranlib) "$@"
+    """,
+)
+
+cc_static_library(
+    name = "v8_146_4_0_x86_64_unknown_linux_musl",
```
zbarsky-openai: what are all of these per-platform static libs? Aren't they exactly the same??

third_party/v8/BUILD.bazel
```diff
@@ -0,0 +1,275 @@
...
+)
+
+cc_static_library(
+    name = "v8_146_4_0_x86_64_pc_windows_msvc",
+    deps = [":v8_146_4_0_binding"],
+    features = V8_STATIC_LIBRARY_FEATURES,
+)
+
+alias(
+    name = "v8_146_4_0_aarch64_pc_windows_gnullvm",
```
zbarsky-openai, line 198: um....I don't think this is correct. I mean it's equivalent because these are all the same, but gnullvm and msvc are quite different

third_party/v8/BUILD.bazel
```diff
@@ -0,0 +1,275 @@
...
+filegroup(
+    name = "rusty_v8_release_pair_aarch64_pc_windows_msvc",
+    srcs = [
+        ":v8_146_4_0_aarch64_pc_windows_msvc",
+        ":src_binding_release_aarch64_pc_windows_msvc",
+    ],
+)
+
+alias(
+    name = "rusty_v8_archive_for_target",
```
zbarsky-openai, line 351: I think you want to replace 90% of this file with a `--platforms` flag you pass from the CLI, or something like that. still haven't fully wrapped my head around what you're doing

BUILD.bazel
```diff
@@ -23,6 +23,38 @@ platform(
     parents = ["@platforms//host"],
 )
 
+platform(
+    name = "linux_amd64_musl_pie_off",
```
zbarsky-openai: i think you just want to use `@rules_rs//rs/experimental/platforms:linux_amd64_musl` or however it's spelled. That should do the right thing. Ditto below

MODULE.bazel
```diff
@@ -63,13 +65,48 @@ crate.from_cargo(
...
     gen_build_script = "off",
     deps = ["@zstd"],
 )
+crate.annotation(
+    build_script_data = [
+        "@@//third_party/v8:rusty_v8_archive_for_target",
+        "@@//third_party/v8:rusty_v8_binding_for_target",
+    ],
+    build_script_env = {
+        "RUSTY_V8_ARCHIVE": "$(execpath @@//third_party/v8:rusty_v8_archive_for_target)",
```
zbarsky-openai, line 99: oh, you are already passing these with execpath. So then why do you need runfiles support?

.github/workflows/rusty-v8-release.yml
cconger: Don't need this, sorry.  Cleaned now

patches/v8_bazel_rules.patch
cconger: Windows was pretty completely redone here, sorry this was in a half applied state before I disabled it.

cconger:
Closed in favor of #15021

PR #14816, opened 2026-03-16
fix: tui freeze when sub-agents are present

jif-oai:
The issue was due to a circular `Drop` schema where the embedded app-server wait for some listeners that wait for this app-server them-selves.

The fix is an explicit cleaning

**Repro:**
* Start codex
* Ask it to spawn a sub-agent
* Close Codex
* It takes 5s to exit


aibrahim-oai (approved):

etraut-openai (approved):
Thanks for finding and fixing this!

PR #14818, opened 2026-03-16
Fix: Handle invalid MDM config entries safely

greg-openai:
## Summary

MDM-managed config was brittle in a way that made a bad admin rollout disproportionately dangerous. A single invalid managed setting could prevent Codex from loading configuration at all, turning a small policy mistake into a fleet-wide outage across every machine. Because MDM settings are centrally pushed and not end-user recoverable, one bad value could effectively take Codex offline across an organization until the policy was fixed upstream as Codex exits immediately upon parsing the config.

Failure scenarios include:
- invalid base64 string in profile
- invalid configuration options
- line containing invalid toml
    
A concrete example was:

```toml
[sandbox_workspace_write]
writable_roots = ["~/code"]
```

That works when configured in a local TOML file such as `/etc/codex/config.toml`, but previously failed when delivered via MDM with:

```text
AbsolutePathBuf deserialized without a base path
```

The root cause was that local file-backed config and MDM config followed different processing paths. File-backed config had relative-path resolution and home expansion applied before typed deserialization, while the MDM layer was merged as raw TOML and later deserialized without that path-resolution context.

## Problem

Before this change, MDM config handling had two failure modes that were effectively fatal for many Codex entrypoints:

1. Invalid-but-parseable config values
   - Example: `writable_roots = ["~/code"]` or `writable_roots = ["relative/path"]`
   - Example: unknown config keys such as `non.existant.value = 23`
   - These values could survive initial raw TOML parsing, then fail later during typed config deserialization.

2. Invalid MDM payloads
   - Invalid base64
   - Invalid UTF-8 after decoding
   - Invalid TOML
   - Non-table TOML roots

In both cases, the common user-facing behavior was a configuration load error and exit. That is not a good fit for MDM, where the recipient is not the person who can fix the configuration.

## Why `writable_roots` Was Different via MDM

For local TOML files, Codex resolves path-like config values before the final typed config build. That includes handling things like `~` expansion under the existing path guard/resolution flow.

For MDM-managed config, the old code path only:
- decoded base64
- parsed raw TOML
- merged that TOML into the config layer stack

Then later, the merged config was deserialized into typed config without a base-path guard. That meant `AbsolutePathBuf` fields in MDM config could reject values like `~/code`, even though equivalent values worked from local file-backed config.

### 1. Path Handling: Safe `~` Expansion for `AbsolutePathBuf`

We updated shared `AbsolutePathBuf` deserialization to reuse the existing home-expansion logic before deciding whether a base path is required.

New behavior:
- `~` and `~/...` are expanded first
- if the expanded result is absolute, it is accepted
- plain relative paths are still rejected when no base path exists

This fixes the MDM happy path for `~/code` without broadening semantics for arbitrary relative paths like `foo/bar`.

### 2. MDM Validation and Recovery: Fail Soft Instead of Failing Closed

We added typed TOML sanitization for MDM config and MDM requirements config so bad managed values no longer take down config loading.

The sanitizer:
- attempts typed deserialization into the real config schema
- uses serde path information to find the exact failing field or array element
- removes the invalid entry
- retries until the remaining payload is valid or unrecoverable

We also prune empty arrays and tables after recovery so we do not leave behind confusing no-op shells.

If the payload is malformed at the transport or parse layer, we now:
- log the problem
- ignore the whole MDM payload
- continue startup without applying that managed payload

This applies to:
- invalid base64
- invalid UTF-8
- invalid TOML
- non-table config roots

### Happy path

Valid MDM config continues to load normally.

Example:
```toml
sandbox_mode = "workspace-write"

[sandbox_workspace_write]
writable_roots = ["~/code"]
```

Now:
- `~/code` expands to an absolute home-directory path
- the config loads successfully
- behavior matches user expectations from local TOML-backed config

### Recoverable bad values

If the MDM payload is structurally valid TOML but contains invalid config entries, we now:
- log the invalid path
- drop the bad field or array element
- continue loading the remaining valid MDM config

Examples:
- unknown keys
- invalid enum values
- wrong types
- bad individual path entries in arrays

### Unrecoverable payload issues

If the payload is malformed at the transport or parse layer, we:
- log the problem
- ignore the whole MDM payload
- continue startup without applying that managed payload

## Validation and Recovery Model

There are now three layers of validation:

1. Payload validation
   - base64 decode
   - UTF-8 validation
   - TOML parse
   - root must be a table

2. Typed schema validation with recovery
   - deserialize into `ConfigToml` / `ConfigRequirementsToml`
   - use serde-path information to localize failures
   - remove the offending field or array element
   - prune empty containers
   - retry until valid or unrecoverable

3. Normal merged config build
   - remaining config still goes through the usual typed config build
   - shared path validation still applies
   - `~` now resolves safely for `AbsolutePathBuf`

This gives us a fail-soft MDM path while still preserving strong validation for the resulting config.

### Permanent tests added

We added focused regression coverage for the path handling changes:
- `~` expansion in `AbsolutePathBuf` without a base path
- plain relative paths still failing without a base path
- MDM `sandbox_workspace_write.writable_roots = ["~/code"]` loading successfully

We also added focused regression coverage for the validation and recovery changes:
- invalid managed config entries being ignored without dropping valid sibling config
- invalid managed config payloads being ignored
- invalid managed requirements payloads being ignored

### Local validation performed

We also locally verified:
- the provided base64 MDM payload processes successfully
- the commented-out `writable_roots = ["~/code"]` path works once enabled in the payload

### Commands run

```text
cargo test -p codex-utils-absolute-path
cargo test -p codex-config
cargo test -p codex-core managed_preferences_
```

These passed.

A full `cargo test -p codex-core` still has one unrelated pre-existing failing auth test:
- `auth::tests::enforce_login_restrictions_logs_out_for_workspace_mismatch`

## Relevant Notes

- This change intentionally does not make arbitrary relative paths valid in MDM.
- Only `~`/`~/...` gains new support in no-base deserialization.
- The goal is to improve resilience for admin-managed config without weakening path validation or introducing `cwd`-dependent behavior.

## Outcome

This change improves MDM safety in two distinct ways:

1. Path handling is now consistent enough that common admin values like `~/code` work as expected.
2. Validation and recovery are now resilient enough that bad managed values are logged and ignored instead of causing unrecoverable end-user startup failures.


codex-rs/core/src/config/mod.rs
```diff
@@ -619,6 +619,17 @@ impl ConfigBuilder {
...
+        let loader_overrides = {
+            let loader_overrides_provided = loader_overrides.is_some();
+            let mut loader_overrides = loader_overrides.unwrap_or_default();
+            if !loader_overrides_provided {
+                loader_overrides.managed_preferences_base64 = Some(String::new());
+                loader_overrides.macos_managed_config_requirements_base64 = Some(String::new());
+            }
+            loader_overrides
+        };
+        #[cfg(not(all(test, target_os = "macos")))]
```
owenlin0, line 632: This seems like an anti-pattern (changing the logic of this function based on whether it's running inside a test), can we address this via proper test setup instead? i.e. make tests explicitly pass in empty overrides?

codex-rs/config/src/diagnostics.rs
```diff
@@ -134,6 +136,70 @@ pub fn config_error_from_typed_toml<T: DeserializeOwned>(
...
+                });
+            }
+            Err(err) => {
+                let path_hint = err.path().clone();
+                let toml_err: toml::de::Error = err.into_inner();
+                let Some(recovery_path) = recovery_path_for_error(&path_hint) else {
+                    return Err(io::Error::new(io::ErrorKind::InvalidData, toml_err));
+                };
+                let recovery_path_display = display_recovery_path(&recovery_path);
+                if !seen_paths.insert(recovery_path_display.clone()) {
```
owenlin0, line 180: had codex take a pass at this PR, and it flagged this case:

```
model = "gpt-5.4"

sandbox_mode = "workspace-write"

[sandbox_workspace_write]
writable_roots = ["relative/a", "relative/b", "~/code"]
```

where I think the expected behavior we'd want is to keep the valid entries in a list, so just `["~/code"]`, but somehow we drop the whole blob in this scenario. i didn't dig too much further but worth double-checking

owenlin0 (commented):
> We updated shared AbsolutePathBuf deserialization to reuse the existing home-expansion logic before deciding whether a base path is required.
> 
> New behavior:
> - ~ and ~/... are expanded first

are we sure we want to do this everywhere `AbsolutePathBuf` is deserialized? It'd be a pretty wide change since we use this type in a lot of app-server APIs as well.

PR #14829, opened 2026-03-16
[stack 3/4] Add current thread context to realtime startup

aibrahim-oai:
## Stack Position
3/4. Top-of-stack sibling built on #14830.

## Base
- #14830

## Sibling
- #14827

## Scope
- Extend the realtime startup context with a bounded summary of the latest thread turns for continuity.


charley-oai (approved):
lgtm

PR #14833, opened 2026-03-16
fix(core): fix sanitize name to use '_' everywhere

apanasenko-oai:


codex-rs/core/src/connectors.rs
```diff
@@ -918,11 +918,15 @@ fn normalize_connector_value(value: Option<&str>) -> Option<String> {
...
-    let slug = sanitize_name(name);
+    let slug = sanitize_slug(name);
     format!("https://chatgpt.com/apps/{slug}/{connector_id}")
 }
 
 pub fn sanitize_name(name: &str) -> String {
+    sanitize_slug(name).replace("-", "_")
+}
+
+pub fn sanitize_slug(name: &str) -> String {
```
pakrym-oai: does this need to be public?

pakrym-oai (approved):

codex-rs/core/src/connectors.rs
apanasenko-oai, line 471: @mzeng-openai do we need here `-` or `_`?

codex-rs/core/src/connectors.rs
apanasenko-oai: no

PR #14838, opened 2026-03-16
fix(core): prevent hanging turn/start due to websocket warming issues

owenlin0:
## Description

This PR fixes a bad first-turn failure mode in app-server when the startup websocket prewarm hangs. Before this change, `initialize -> thread/start -> turn/start` could sit behind the prewarm for up to five minutes, so the client would not see `turn/started`, and even `turn/interrupt` would block because the turn had not actually started yet.

Now, we:
- set a (configurable) timeout of 15s for websocket startup time, exposed as `websocket_startup_timeout_ms` in config.toml
- `turn/started` is sent immediately on `turn/start` even if the websocket is still connecting
- `turn/interrupt` can be used to cancel a turn that is still waiting on the websocket warmup
- the turn task will wait for the full 15s websocket warming timeout before falling back

## Why

The old behavior made app-server feel stuck at exactly the moment the client expects turn lifecycle events to start flowing. That was especially painful for external clients, because from their point of view the server had accepted the request but then went silent for minutes.

## Configuring the websocket startup timeout
Can set it in config.toml like this:
```
[model_providers.openai]
supports_websockets = true
websocket_connect_timeout_ms = 15000
```

### DIFF
--- a/codex-rs/core/src/client.rs
+++ b/codex-rs/core/src/client.rs
@@ -117,6 +117,9 @@ const RESPONSES_WEBSOCKETS_V2_BETA_HEADER_VALUE: &str = "responses_websockets=20
 const RESPONSES_ENDPOINT: &str = "/responses";
 const RESPONSES_COMPACT_ENDPOINT: &str = "/responses/compact";
 const MEMORIES_SUMMARIZE_ENDPOINT: &str = "/memories/trace_summarize";
+#[cfg(test)]
+pub(crate) const WEBSOCKET_CONNECT_TIMEOUT: Duration =
+    Duration::from_millis(crate::model_provider_info::DEFAULT_WEBSOCKET_CONNECT_TIMEOUT_MS);
 pub fn ws_version_from_features(config: &Config) -> bool {
     config
         .features
@@ -308,6 +311,27 @@ impl ModelClient {
             .cached_websocket_session
             .lock()
             .unwrap_or_else(std::sync::PoisonError::into_inner) = websocket_session;
+    }
+
+    pub(crate) fn force_http_fallback(
+        &self,
+        session_telemetry: &SessionTelemetry,
+        model_info: &ModelInfo,
+    ) -> bool {
+        let websocket_enabled = self.responses_websocket_enabled(model_info);
+        let activated =
+            websocket_enabled && !self.state.disable_websockets.swap(true, Ordering::Relaxed);
+        if activated {
+            warn!("falling back to HTTP");
+            session_telemetry.counter(
+                "codex.transport.fallback_to_http",
+                /*inc*/ 1,
+                &[("from_wire_api", "responses_websocket")],
+            );
+        }
+
+        self.store_cached_websocket_session(WebsocketSession::default());
+        activated
     }
 
     /// Compacts the current conversation history using the Compact endpoint.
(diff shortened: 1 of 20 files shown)


codex-rs/core/tests/suite/stream_no_completed.rs
```diff
@@ -61,6 +61,7 @@ async fn retries_on_early_close() {
         request_max_retries: Some(0),
         stream_max_retries: Some(1),
         stream_idle_timeout_ms: Some(2000),
+        websocket_startup_timeout_ms: None,
```
pakrym-oai: I'd do a websocket connect timeout and use it for both normal and prewarm connect.

codex-rs/core/src/tasks/regular.rs
```diff
@@ -72,6 +239,12 @@ impl SessionTask for RegularTask {
         "session_task.turn"
     }
 
+    fn emits_turn_started_on_spawn(&self) -> bool {
```
pakrym-oai: can we get rid of this completely? And await at the point where we need the session (after TaskStart)?

codex-rs/core/src/tasks/regular.rs
```diff
@@ -1,41 +1,183 @@
...
+pub(crate) struct StartupPrewarmHandle {
+    task: JoinHandle<CodexResult<ModelClientSession>>,
+    started_at: Instant,
+    timeout: Duration,
 }
 
-impl Default for RegularTask {
-    fn default() -> Self {
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+struct StartupPrewarmCancelled;
```
pakrym-oai: CancelErr?

codex-rs/core/src/client.rs
```diff
@@ -907,13 +933,20 @@ impl ModelClientSession {
                     auth_context,
                     request_route_telemetry,
                 )
-                .await?;
+                .await
+            {
+                Ok(new_conn) => new_conn,
+                Err(err) => {
+                    if matches!(err, ApiError::Transport(TransportError::Timeout)) {
+                        self.reset_websocket_session();
```
pakrym-oai, line 945: Why is this needed? Don't we auto retry?

codex-rs/core/src/tasks/regular.rs
```diff
@@ -81,8 +232,22 @@ impl SessionTask for RegularTask {
...
+            collaboration_mode_kind: ctx.collaboration_mode.mode,
+        });
+        sess.send_event(ctx.as_ref(), event).await;
+        sess.set_server_reasoning_included(false).await;
+        let prewarmed_client_session = match self
+            .take_prewarmed_session(&sess.services.session_telemetry, &cancellation_token)
+            .await
+        {
+            Ok(prewarmed_client_session) => prewarmed_client_session,
+            Err(StartupPrewarmCancelled) => return None,
```
pakrym-oai: If nobody cares about the value consider Option<>

codex-rs/core/src/client.rs
owenlin0: so a retry does happen but at the turn level, not by creating a brand new websocket client from scratch

So when a request fails, Codex retries the same turn using the same turn-scoped session object. If we do not call `reset_websocket_session()`, that retry can carry leftover websocket state from the timed-out attempt and act like it is continuing a connection that never really came up.

codex-rs/core/src/config/schema_tests.rs
```diff
@@ -6,6 +6,10 @@ use pretty_assertions::assert_eq;
 use similar::TextDiff;
 use tempfile::TempDir;
 
+fn trim_single_trailing_newline(contents: &str) -> &str {
```
pakrym-oai, line 9: this is random

pakrym-oai (approved):

PR #14842, opened 2026-03-16
Add notify to code-mode

pakrym-oai:
Allows model to send an out-of-band notification.

The notification is injected as another tool call output for the same call_id.


codex-rs/core/src/tools/code_mode/process.rs
```diff
@@ -138,9 +141,13 @@ pub(super) async fn spawn_code_mode_process(
                     NodeToHostMessage::ToolCall { tool_call } => {
                         let _ = tool_call_tx.send(tool_call);
                     }
+                    NodeToHostMessage::Notify { notify } => {
+                        let _ = notify_tx.send(notify);
```
cconger: I'm curious why this is a different channel.  It looks like you're demuxing it from stdout here to two channels, and then multiplex it back on the read side in worker.  Is there a reason you want these to be split channels?

cconger (approved):
One question about second channel.

codex-rs/core/src/tools/code_mode/process.rs
pakrym-oai: Good point. There is no reason.

PR #14846, opened 2026-03-16
use framed IPC for elevated command runner

iceweasel-oai:
## Summary
This is PR 2 of the Windows sandbox runner split.

PR 1 introduced the framed IPC runner foundation and related Windows sandbox infrastructure without changing the active elevated one-shot execution path. This PR switches that elevated one-shot path over to the new runner IPC transport and removes the old request-file bootstrap that PR 1 intentionally left in place.

After this change, ordinary elevated Windows sandbox commands still behave as one-shot executions, but they now run as the simple case of the same helper/IPC transport that later unified_exec work will build on.

## Why this is needed for unified_exec
Windows elevated sandboxed execution crosses a user boundary: the CLI launches a helper as the sandbox user and has to manage command execution from outside that security context. For one-shot commands, the old request-file/bootstrap flow was sufficient. For unified_exec, it is not.

Unified_exec needs a long-lived bidirectional channel so the parent can:
- send a spawn request
- receive structured spawn success/failure
- stream stdout and stderr incrementally
- eventually support stdin writes, termination, and other session lifecycle events

This PR does not add long-lived sessions yet. It converts the existing elevated one-shot path to use the same framed IPC transport so that PR 3 can add unified_exec session semantics on top of a transport that is already exercised by normal elevated command execution.

## Scope
This PR:
- updates `windows-sandbox-rs/src/elevated_impl.rs` to launch the runner with named pipes, send a framed `SpawnRequest`, wait for `SpawnReady`, and collect framed `Output`/`Exit` messages
- removes the old `--request-file=...` execution path from `windows-sandbox-rs/src/elevated/command_runner_win.rs`
- keeps the public behavior one-shot: no session reuse or interactive unified_exec behavior is introduced here

This PR does not:
- add Windows unified_exec session support
- add background terminal reuse
- add PTY session lifecycle management

## Why Windows needs this and Linux/macOS do not
On Linux and macOS, the existing sandbox/process model composes much more directly with long-lived process control. The parent can generally spawn and own the child process (or PTY) directly inside the sandbox model we already use.

Windows elevated sandboxing is different. The parent is not directly managing the sandboxed process in the same way; it launches across a different user/security context. That means long-lived control requires an explicit helper process plus IPC for spawn, output, exit, and later stdin/session control.

So the extra machinery here is not because unified_exec is conceptually different on Windows. It is because the elevated Windows sandbox boundary requires a helper-mediated transport to support it cleanly.

## Validation

### DIFF
--- a/codex-rs/windows-sandbox-rs/src/elevated_impl.rs
+++ b/codex-rs/windows-sandbox-rs/src/elevated_impl.rs
@@ -9,6 +9,13 @@ mod windows_impl {
     use crate::helper_materialization::resolve_helper_for_launch;
     use crate::helper_materialization::HelperExecutable;
     use crate::identity::require_logon_sandbox_creds;
+    use crate::ipc_framed::decode_bytes;
+    use crate::ipc_framed::read_frame;
+    use crate::ipc_framed::write_frame;
+    use crate::ipc_framed::FramedMessage;
+    use crate::ipc_framed::Message;
+    use crate::ipc_framed::OutputStream;
+    use crate::ipc_framed::SpawnRequest;
     use crate::logging::log_failure;
     use crate::logging::log_note;
     use crate::logging::log_start;
@@ -26,8 +33,9 @@ mod windows_impl {
     use rand::SeedableRng;
     use std::collections::HashMap;
     use std::ffi::c_void;
-    use std::fs;
+    use std::fs::File;
     use std::io;
+    use std::os::windows::io::FromRawHandle;
     use std::path::Path;
     use std::path::PathBuf;
     use std::ptr;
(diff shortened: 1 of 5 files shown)


codex-rs/windows-sandbox-rs/src/elevated_impl.rs
```diff
@@ -213,7 +210,7 @@ mod windows_impl {
         cwd: &Path,
         mut env_map: HashMap<String, String>,
         timeout_ms: Option<u64>,
-        use_private_desktop: bool,
+        _use_private_desktop: bool,
```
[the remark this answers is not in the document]
iceweasel-oai: fixed in https://github.com/openai/codex/pull/14846/changes/2f37985f6765d52d850ed670f3a5277799a09519

codex-rs/windows-sandbox-rs/src/elevated_impl.rs
```diff
@@ -390,73 +343,59 @@ mod windows_impl {
...
+                    real_codex_home: codex_home.to_path_buf(),
+                    cap_sids,
+                    timeout_ms,
+                    tty: false,
+                    stdin_open: false,
+                }),
+            },
+        };
+        write_frame(&mut pipe_write, &spawn_request)?;
+        read_spawn_ready(&mut pipe_read)?;
```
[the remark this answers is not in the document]
iceweasel-oai: fixed in https://github.com/openai/codex/pull/14846/changes/0f23dde6aa0d4770ed4e3e4e5ced5ba413134a49

codex-rs/windows-sandbox-rs/src/elevated_impl.rs
```diff
@@ -40,15 +48,12 @@ mod windows_impl {
     use windows_sys::Win32::System::Diagnostics::Debug::SetErrorMode;
     use windows_sys::Win32::System::Pipes::ConnectNamedPipe;
     use windows_sys::Win32::System::Pipes::CreateNamedPipeW;
-    // PIPE_ACCESS_DUPLEX is 0x00000003; not exposed in windows-sys 0.52, so use the value directly.
-    const PIPE_ACCESS_DUPLEX: u32 = 0x0000_0003;
+    const PIPE_ACCESS_INBOUND: u32 = 0x0000_0001;
+    const PIPE_ACCESS_OUTBOUND: u32 = 0x0000_0002;
```
dylan-hurd-oai, line 52: sanity-check: PIPE_ACCESS_DUPLEX is not exposed in windows-sys 0.52, but are these?

codex-rs/windows-sandbox-rs/src/elevated_impl.rs
```diff
@@ -390,106 +343,110 @@ mod windows_impl {
...
+                    CloseHandle(pi.hThread);
+                }
+                if pi.hProcess != 0 {
+                    CloseHandle(pi.hProcess);
+                }
             }
+            return Err(err.into());
         }
 
+        let result = (|| -> Result<CaptureResult> {
```
dylan-hurd-oai, line 373: since we're doing a read loop in this block, should it be on its own thread? or are we planning to handle that in a follow-up unified exec PR?

dylan-hurd-oai (approved):

codex-rs/windows-sandbox-rs/src/elevated_impl.rs
iceweasel-oai: they also are not

codex-rs/windows-sandbox-rs/src/elevated_impl.rs
iceweasel-oai: yep - this is for the non-pty/unified_exec version only

PR #14847, opened 2026-03-16
feat: add websocket auth for app-server

viyatb-oai:
## Summary
This change adds websocket authentication at the app-server transport boundary and enforces it before JSON-RPC `initialize`, so authenticated deployments reject unauthenticated clients during the websocket handshake rather than after a connection has already been admitted.

During rollout, websocket auth is opt-in for non-loopback listeners so we do not break existing remote clients. If `--ws-auth ...` is configured, the server enforces auth during websocket upgrade. If auth is not configured, non-loopback listeners still start, but app-server logs a warning and the startup banner calls out that auth should be configured before real remote use.

The server supports two auth modes: a file-backed capability token, and a standard HMAC-signed JWT/JWS bearer token verified with the `jsonwebtoken` crate, with optional issuer, audience, and clock-skew validation. Capability tokens are normalized, hashed, and compared in constant time. Short shared secrets for signed bearer tokens are rejected at startup. Requests carrying an `Origin` header are rejected with `403` by transport middleware, and authenticated clients present credentials as `Authorization: Bearer <token>` during websocket upgrade.

## Validation
- `cargo test -p codex-app-server transport::auth`
- `cargo test -p codex-cli app_server_`
- `cargo clippy -p codex-app-server --all-targets -- -D warnings`
- `just bazel-lock-check`

Note: in the broad `cargo test -p codex-app-server connection_handling_websocket` run, the touched websocket auth cases passed, but unrelated Unix shutdown tests failed with a timeout in this environment.


etraut-openai:
@codex review

codex-rs/Cargo.toml
```diff
@@ -169,13 +167,15 @@ async-stream = "0.3.6"
 async-trait = "0.1.89"
 axum = { version = "0.8", default-features = false }
 base64 = "0.22.1"
+biscuit = "0.7.0"
```
etraut-openai: I'm not that familiar with crypto/auth crates in Rust, but I'm curious why you're using biscuit rather than a more popular crate like jsonwebtoken. It looks like biscuit is well maintained, so it's probably fine.

etraut-openai (approved):
The implementation looks good to me.

I have some concerns about this from a usability standpoint. The "signed bearer token" auth method is very advanced and will not be appropriate for most codex users and use cases. The capability token approach is easier to configure, but it still requires multiple steps. We'll need to document this very well to help users get through the steps.

PR #14850, opened 2026-03-16
Materialize MCP tool images into local refs

Aismit:
## Summary

This keeps the change at the producer seam for MCP tool results.

When an MCP tool returns image content, Codex now materializes the image bytes into a temp file before the next model sample and rewrites the MCP content block to a `local_image` reference. The existing function-call output conversion then loads that local image back into model-facing image content for the same turn.

## Why this change

The monorepo observer path is too late for this behavior. By the time `/Users/aismit/code/openai/codex/wham_agent_v2/wham_agent_v2/rollout_fn.py` is iterating emitted app-server events, the tool result has already been consumed by the Codex loop.

The earliest place where we still have the raw mutable `CallToolResult` and can affect the same turn is `codex-rs/core/src/mcp_tool_call.rs`, right after `sess.call_tool(...)` returns and before the result is packaged into `McpToolOutput`.

## What changed

- `codex-rs/core/src/mcp_tool_call.rs`
  - Materializes MCP `image` content blocks into temp files under `codex-mcp-tool-images`
  - Rewrites those blocks to `{"type":"local_image","path":"..."}` when the model supports image input
  - Preserves the existing text fallback for models that do not support image input
- `codex-rs/protocol/src/models.rs`
  - Teaches MCP tool output conversion to treat `local_image` content blocks as image-bearing output by loading the file and emitting the existing function-call image content items

## Why these files

These are the two minimal places needed for the producer path:
- `codex-rs/core` owns the same-turn MCP tool result before the model sees it
- `codex-rs/protocol` already owns conversion from MCP tool output into function-call output payloads

No Hoopa, GAAS, or monorepo observer changes are included here.

## Testing

- `cargo fmt --all`
- `cargo test -p codex-protocol serializes_local_image_outputs_as_array`
- `cargo test -p codex-protocol local_image_output_missing_file_becomes_text_placeholder`
- `cargo test -p codex-core sanitize_mcp_tool_result_for_model_rewrites_image_content`
- `cargo test -p codex-core sanitize_mcp_tool_result_for_model_materializes_image_when_supported`


Aismit:
Superseded by #14852, which replays the same two-file change on a fresh branch from main.

PR #14851, opened 2026-03-16
Add exit helper to code mode scripts

pakrym-oai:
- **Summary**
  - expose `exit` through the code mode bridge and module so scripts can stop mid-flight
  - surface the helper in the description documentation
  - add a regression test ensuring `exit()` terminates execution cleanly
- **Testing**
  - Not run (not requested)


cconger (approved):
Unsure if it already has access to `close()` as its running in a worker context.  But fine if we want to have a consistent and named interface to manage this.

PR #14852, opened 2026-03-16
Materialize MCP tool images into local refs

Aismit:
## Summary

Fresh branch from `main` with only the MCP tool image producer changes.

When an MCP tool returns image content, Codex now materializes those image bytes into a temp file before the next model sample and rewrites the MCP content block to a `local_image` reference. The existing function-call output conversion then loads that local image back into model-facing image content for the same turn.

## Why this change

The observer path in the monorepo is too late to affect the same turn. By the time emitted app-server events are being consumed, the tool result has already been used by the Codex loop.

The earliest mutable seam is `codex-rs/core/src/mcp_tool_call.rs`, immediately after `sess.call_tool(...)` returns and before the result is packaged into `McpToolOutput`.

## Files changed

- `codex-rs/core/src/mcp_tool_call.rs`
  - materialize MCP `image` content blocks into temp files
  - rewrite them to `{"type":"local_image","path":"..."}` for image-capable models
  - keep the existing text fallback for non-image-capable models
- `codex-rs/protocol/src/models.rs`
  - treat `local_image` MCP content as image-bearing tool output by loading the file and emitting the existing function-call image content items

## Testing

- `cargo fmt --all`
- `cargo test -p codex-protocol serializes_local_image_outputs_as_array`
- `cargo test -p codex-protocol local_image_output_missing_file_becomes_text_placeholder`
- `cargo test -p codex-core sanitize_mcp_tool_result_for_model_rewrites_image_content`
- `cargo test -p codex-core sanitize_mcp_tool_result_for_model_materializes_image_when_supported`


Aismit:
Superseded by #14854, which is rebuilt directly on top of current origin/main with the isolated MCP tool image changes.

PR #14853, opened 2026-03-16
Wire remote app-server auth through the client

etraut-openai:
For app-server websocket auth, support the two server-side mechanisms from
PR #14847:

- `--ws-auth capability-token --ws-token-file /abs/path`
- `--ws-auth signed-bearer-token --ws-shared-secret-file /abs/path`
  with optional `--ws-issuer`, `--ws-audience`, and
  `--ws-max-clock-skew-seconds`

On the client side, add interactive remote support via:

- `--remote ws://host:port` or `--remote wss://host:port`
- `--remote-auth-token-env <ENV_VAR>`

Codex reads the bearer token from the named environment variable and sends it
as `Authorization: Bearer <token>` during the websocket handshake. Remote auth
tokens are only allowed for `wss://` URLs or loopback `ws://` URLs.

Testing:
- tested both auth methods manually to confirm connection success and rejection for both auth types

### DIFF
--- a/codex-rs/app-server-client/src/remote.rs
+++ b/codex-rs/app-server-client/src/remote.rs
@@ -48,6 +48,9 @@ use tokio_tungstenite::MaybeTlsStream;
 use tokio_tungstenite::WebSocketStream;
 use tokio_tungstenite::connect_async;
 use tokio_tungstenite::tungstenite::Message;
+use tokio_tungstenite::tungstenite::client::IntoClientRequest;
+use tokio_tungstenite::tungstenite::http::HeaderValue;
+use tokio_tungstenite::tungstenite::http::header::AUTHORIZATION;
 use tracing::warn;
 use url::Url;
 
@@ -57,6 +60,7 @@ const INITIALIZE_TIMEOUT: Duration = Duration::from_secs(10);
 #[derive(Debug, Clone)]
 pub struct RemoteAppServerConnectArgs {
     pub websocket_url: String,
+    pub auth_token: Option<String>,
     pub client_name: String,
     pub client_version: String,
     pub experimental_api: bool,
@@ -83,6 +87,16 @@ impl RemoteAppServerConnectArgs {
             },
             capabilities: Some(capabilities),
         }
+    }
+}
+
+pub(crate) fn websocket_url_supports_auth_token(url: &Url) -> bool {
+    match (url.scheme(), url.host()) {
+        ("wss", Some(_)) => true,
+        ("ws", Some(url::Host::Domain(domain))) => domain.eq_ignore_ascii_case("localhost"),
+        ("ws", Some(url::Host::Ipv4(addr))) => addr.is_loopback(),
+        ("ws", Some(url::Host::Ipv6(addr))) => addr.is_loopback(),
+        _ => false,
     }
 }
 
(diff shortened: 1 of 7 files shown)


etraut-openai:
@codex review

etraut-openai:
@codex review

codex-rs/app-server-client/src/remote.rs
```diff
@@ -56,6 +59,7 @@ const INITIALIZE_TIMEOUT: Duration = Duration::from_secs(10);
 #[derive(Debug, Clone)]
 pub struct RemoteAppServerConnectArgs {
     pub websocket_url: String,
+    pub auth_token: Option<String>,
```
fcoury, line 63: If we ever log `RemoteAppServerConnectArgs` the auth_token could leak. I advise you implement a custom version of Debug trait where the `auth_token` is obfuscated.

codex-rs/cli/src/main.rs
```diff
@@ -530,6 +530,10 @@ struct InteractiveRemoteOptions {
     /// Accepted forms: `ws://host:port` or `wss://host:port`.
     #[arg(long = "remote", value_name = "ADDR")]
     remote: Option<String>,
+
+    /// Bearer token to send to a remote app server websocket.
+    #[arg(long = "remote-auth-token", value_name = "TOKEN")]
+    remote_auth_token: Option<String>,
```
fcoury: The `--remote-auth-token` param will be leaked in process listings like `ps aux`. A better alternative would be to use an environment variable to pass in the token, since it doesn't appear in process listings.

codex-rs/tui_app_server/src/lib.rs
```diff
@@ -284,7 +284,10 @@ async fn start_embedded_app_server(
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub(crate) enum AppServerTarget {
     Embedded,
-    Remote(String),
+    Remote {
+        websocket_url: String,
+        auth_token: Option<String>,
```
fcoury, line 289: Same potential issue: the `auth_token` can leak if we log the debug version of this enum variant.

fcoury (commented):
Looks good aside from those minor potential issues.

codex-rs/tui_app_server/src/lib.rs
etraut-openai: I'm not that worried about it. We don't currently log this, and it would be a bug if we did in the future.

codex-rs/app-server-client/src/remote.rs
etraut-openai: I'm not that worried about it. We don't currently log this, and it would be a bug if we did in the future.

codex-rs/cli/src/main.rs
etraut-openai: This is a good point. I'm changing it to use an env variable. The updated parameter name is `--remote-auth-token-env`.

fcoury: You can also standardize the env var name and remove the parameter.

etraut-openai: We've done that in a few other cases and have regretted it. Basing behaviors on hard-coded env variable names results in surprising behaviors and support issues. For example, we recently needed to deprecate `OPENAI_BASE_URL` because it created many support issues.

fcoury: Good insight. Never thought of that.

But env vars in general are bad. Hard to test, and akin to global state. But I don't see another safer way.

etraut-openai: There's precedent for this pattern in the CLI. We use it for the `--with-api-key <env-var>` command-line option. In an earlier version, we supported `--api-key <token>`, but we needed to change that for security — the same concern that you're raising here.

PR #14856, opened 2026-03-16
[apps][tool_suggest] Remove tool_suggest's dependency on tool search.

mzeng-openai:
- [x] Remove tool_suggest's dependency on tool search.


codex-rs/core/templates/search_tool/tool_suggest_description.md
```diff
@@ -3,8 +3,7 @@
 Suggests a discoverable connector or plugin when the user clearly wants a capability that is not currently available in the active `tools` list.
 
 Use this ONLY when:
-- There's no available tool to handle the user's request
-- And tool_search fails to find a good match
+- You've already tried to find a matching available tool for the user's request but couldn't find a good match. This includes `tool_search` (if available) and other means.
```
sayan-oai, line 6: now that `tool_search` is not always available when `tool_suggest` is available, i wonder if we should conditionally render the bit about `tool_search` rather than saying `(if available)` to avoid confusing the model?

sayan-oai:
in similar fashion to my review comment, should we avoid assuming `tool_search` is present in the description we render for `apps` [here](https://sg.apps.openai.org/github.com/openai/codex/-/blob/codex-rs/core/src/apps/render.rs?L7)?

sayan-oai (approved):

PR #14859, opened 2026-03-16
Feat: CXA-1831 Persist latest model and reasoning effort in sqlite

shijie-oai:
### Summary
The goal is for us to get the latest turn model and reasoning effort on thread/resume is no override is provided on the thread/resume func call. This is the part 1 which we write the model and reasoning effort for a thread to the sqlite db and there will be a followup PR to consume the two new fields on thread/resume.

[part 2 PR is currently WIP](https://github.com/openai/codex/pull/14888) and this one can be merged independently.


codex-rs/state/migrations/0020_threads_model_reasoning_effort.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE threads ADD COLUMN model TEXT;
+ALTER TABLE threads ADD COLUMN reasoning_effort TEXT;
```
shijie-oai, line 2: Capture the latest model and reasoning effort per thread.

codex-rs/protocol/src/openai_models.rs
```diff
@@ -48,6 +49,22 @@ pub enum ReasoningEffort {
...
+            "none" => Ok(Self::None),
+            "minimal" => Ok(Self::Minimal),
+            "low" => Ok(Self::Low),
+            "medium" => Ok(Self::Medium),
+            "high" => Ok(Self::High),
+            "xhigh" => Ok(Self::XHigh),
+            _ => Err(format!("invalid reasoning_effort: {s}")),
+        }
+    }
+}
```
shijie-oai, line 59: A simple string to enum parser so that read from db can be correctly converted. Better to keep such parser on the enum def side so that we can reuse down the line.

codex-rs/protocol/src/openai_models.rs
```diff
@@ -48,6 +49,22 @@ pub enum ReasoningEffort {
     XHigh,
 }
 
+impl FromStr for ReasoningEffort {
+    type Err = String;
```
shijie-oai, line 53: Generic string type error - open to have a custom error type if necessary but feel like an overkill.

shijie-oai:
@codex review

charley-oai (approved):
Seems reasonable

codex-rs/state/migrations/0020_threads_model_reasoning_effort.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE threads ADD COLUMN model TEXT;
```
jif-oai, line 1: @charley-oai if there is a migration here, maybe good to package yours in the same release

codex-rs/protocol/src/openai_models.rs
```diff
@@ -48,6 +49,22 @@ pub enum ReasoningEffort {
     XHigh,
 }
 
+impl FromStr for ReasoningEffort {
```
jif-oai, line 52: Why not just using serde?

codex-rs/state/src/runtime/threads.rs
```diff
@@ -756,6 +780,69 @@ mod tests {
         assert_eq!(memory_mode.as_deref(), Some("polluted"));
     }
 
+    #[tokio::test]
+    async fn apply_rollout_items_without_turn_context_persists_null_model_and_reasoning_effort() {
```
jif-oai: the 3 tests in this file are a bit redundant. Can we merge them please? 
We shouldn't have test for every combinatorial possibility and try to build smarter tests instead. The CI pipeline is already pretty slow

codex-rs/state/src/extract.rs
```diff
@@ -325,6 +328,71 @@ mod tests {
         assert_eq!(metadata.cwd, PathBuf::from("/fallback/workspace"));
     }
 
+    #[test]
+    fn turn_context_sets_model_and_reasoning_effort() {
```
jif-oai, line 332: Mostly redundant as well

codex-rs/protocol/src/openai_models.rs
shijie-oai: Serde works but wanted to be more intentional with the mapping and intentionally throw here for invalid value.

shijie-oai: actually will switch to serde.

codex-rs/state/src/extract.rs
shijie-oai: Addressed

codex-rs/state/src/runtime/threads.rs
shijie-oai: Addressed

PR #14861, opened 2026-03-16
Add marketplace display names to plugin/list

xl-openai:
Add display_name support to marketplace.json.

### DIFF
--- a/codex-rs/app-server-protocol/schema/json/codex_app_server_protocol.schemas.json
+++ b/codex-rs/app-server-protocol/schema/json/codex_app_server_protocol.schemas.json
@@ -9372,6 +9372,12 @@
       },
       "PluginMarketplaceEntry": {
         "properties": {
+          "displayName": {
+            "type": [
+              "string",
+              "null"
+            ]
+          },
           "name": {
             "type": "string"
           },
--- a/codex-rs/app-server-protocol/schema/json/codex_app_server_protocol.v2.schemas.json
+++ b/codex-rs/app-server-protocol/schema/json/codex_app_server_protocol.v2.schemas.json
@@ -6160,6 +6160,12 @@
     },
     "PluginMarketplaceEntry": {
       "properties": {
+        "displayName": {
+          "type": [
+            "string",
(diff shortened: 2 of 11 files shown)


codex-rs/app-server/README.md
```diff
@@ -164,14 +164,14 @@ Example with notification opt-out:
...
 - `skills/list` — list skills for one or more `cwd` values (optional `forceReload`).
-- `plugin/list` — list discovered plugin marketplaces and plugin state, including effective marketplace install/auth policy metadata. `interface.category` uses the marketplace category when present; otherwise it falls back to the plugin manifest category. Pass `forceRemoteSync: true` to refresh curated plugin state before listing (**under development; do not call from production clients yet**).
+- `plugin/list` — list discovered plugin marketplaces and plugin state, including marketplace-level `displayName` plus effective marketplace install/auth policy metadata. Plugin `interface.category` uses the marketplace category when present; otherwise it falls back to the plugin manifest category. Pass `forceRemoteSync: true` to refresh curated plugin state before listing (**under development; do not call from production clients yet**).
```
sayan-oai: do we really need this in the readme? seems a bit too schema-specific, up to you

codex-rs/core/src/plugins/store.rs
```diff
@@ -157,11 +171,13 @@ impl PluginStore {
         }
         validate_plugin_segment(&plugin_version, "plugin version")
             .map_err(PluginStoreError::Invalid)?;
+        let disallowed_skill_dirs = disallowed_skill_dirs(source_path.as_path(), product)?;
```
sayan-oai: do we need to also key the install cache by `product` now? it seems like a `ChatGPT` or `API` install could be overwritten by a later `Codex` refresh/sync, so the cache would serve the wrong values. LMK if im misunderstanding

xl-openai:
Repurpose to support add marketplace display names to plugin/list only as product-aware plugin support requires more thinking.

sayan-oai (approved):

PR #14862, opened 2026-03-16
Add exec-server stub server and protocol docs

starr-openai:
## Summary

This PR adds the standalone `codex-exec-server` stdio JSON-RPC crate.

It includes:
- the `codex-exec-server` binary and Rust client
- smoke tests for stdio initialization, output streaming, and writes
- a crate-local README with the current transport, protocol, and API definition

It intentionally does not wire the crate into the CLI or unified-exec yet.

## Validation

- `just bazel-lock-check`

## Notes

- The README documents the current wire protocol as implemented in this PR.
- The follow-up wiring into higher-level launch paths can land separately.

### DIFF
--- a/codex-rs/exec-server/src/bin/codex-exec-server.rs
+++ b/codex-rs/exec-server/src/bin/codex-exec-server.rs
@@ -1,0 +1,7 @@
+#[tokio::main]
+async fn main() {
+    if let Err(err) = codex_exec_server::run_main().await {
+        eprintln!("{err}");
+        std::process::exit(1);
+    }
+}
--- a/codex-rs/exec-server/src/client.rs
+++ b/codex-rs/exec-server/src/client.rs
@@ -1,9 +1,6 @@
 use std::collections::HashMap;
-use std::path::PathBuf;
-use std::process::Stdio;
 use std::sync::Arc;
 use std::sync::Mutex as StdMutex;
-use std::sync::atomic::AtomicBool;
 use std::sync::atomic::AtomicI64;
 use std::sync::atomic::Ordering;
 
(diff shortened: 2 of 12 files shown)


codex-rs/exec-server/src/bin/codex-exec-server.rs
```diff
@@ -0,0 +1,23 @@
+use clap::Parser;
+use codex_exec_server::ExecServerTransport;
+
+#[derive(Debug, Parser)]
+struct ExecServerArgs {
+    /// Transport endpoint URL. Supported values: `stdio://` (default),
```
pakrym-oai: Do we need stdin transport?

codex-rs/exec-server/src/server/handler.rs
```diff
@@ -0,0 +1,830 @@
...
+            )
+            .await
+        }
+        .map_err(|err| internal_error(err.to_string()))?;
+
+        let process_id = params.process_id.clone();
+        {
+            let mut process_map = self.processes.lock().await;
+            if process_map.contains_key(&process_id) {
+                spawned.session.terminate();
```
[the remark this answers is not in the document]
starr-openai: Fixed in `adfc5aa0c`: duplicate `processId` is now rejected before spawning, with a regression test that verifies no side-effecting command runs on the duplicate path.

codex-rs/exec-server/src/server/handler.rs
```diff
@@ -0,0 +1,830 @@
...
+    }
+
+    async fn handle_terminate_request(
+        &self,
+        params: crate::protocol::TerminateParams,
+    ) -> Result<TerminateResponse, codex_app_server_protocol::JSONRPCErrorError> {
+        let process = {
+            let mut process_map = self.processes.lock().await;
+            process_map.remove(&params.process_id)
+        };
```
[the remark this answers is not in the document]
starr-openai: Fixed in `adfc5aa0c`: terminate no longer removes the map entry immediately. The ID stays reserved until exit cleanup, with a regression test covering that behavior.

codex-rs/exec-server/src/client.rs
```diff
@@ -0,0 +1,1162 @@
...
+        let process_id = params.process_id.clone();
+        let status = Arc::new(RemoteProcessStatus::new());
+        let (output_tx, output_rx) = broadcast::channel(256);
+        self.inner.processes.lock().await.insert(
+            process_id.clone(),
+            RegisteredProcess {
+                output_tx,
+                status: Arc::clone(&status),
+            },
+        );
```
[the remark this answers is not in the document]
starr-openai: Fixed in `adfc5aa0c`: `start_process` now rejects duplicate local IDs before inserting into the client map, and there is a regression test proving the original process still receives output.

codex-rs/exec-server/src/bin/codex-exec-server.rs
starr-openai: I think yes for now: stdio is the local transport for the helper that launches a colocated server process, while websocket is the remote path. Keeping the server transport-agnostic lets us support both without coupling that choice into the client API.

codex-rs/exec-server/src/server/handler.rs
```diff
@@ -0,0 +1,938 @@
...
+                            .handle_exec_request(params)
+                            .await
+                            .map(ExecServerResponseMessage::Exec),
+                        Err(err) => Err(err),
+                    },
+                )
+                .await;
+            }
+            ExecServerRequest::Write { request_id, params } => {
+                self.send_request_result(
```
pakrym-oai: feels like there is quite a bit of boilerplate with match self.require_initialized and self.send_request_result and actual work is a small part of each branch.

codex-rs/exec-server/src/protocol.rs
```diff
@@ -0,0 +1,134 @@
...
+use serde::Serialize;
+
+pub const INITIALIZE_METHOD: &str = "initialize";
+pub const INITIALIZED_METHOD: &str = "initialized";
+pub const EXEC_METHOD: &str = "command/exec";
+pub const EXEC_WRITE_METHOD: &str = "command/exec/write";
+pub const EXEC_TERMINATE_METHOD: &str = "command/exec/terminate";
+pub const EXEC_OUTPUT_DELTA_METHOD: &str = "command/exec/outputDelta";
+pub const EXEC_EXITED_METHOD: &str = "command/exec/exited";
+pub const PROTOCOL_VERSION: &str = "exec-server.v0";
```
pakrym-oai, line 6: consider removing versioning until we design this.

codex-rs/exec-server/src/protocol.rs
```diff
@@ -0,0 +1,134 @@
...
+impl From<Vec<u8>> for ByteChunk {
+    fn from(value: Vec<u8>) -> Self {
+        Self(value)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct InitializeParams {
+    pub client_name: String,
```
pakrym-oai, line 11: is this needed?

codex-rs/exec-server/src/server/handler.rs
```diff
@@ -0,0 +1,1032 @@
...
+    }
+
+    pub(crate) async fn exec(
+        &self,
+        params: crate::protocol::ExecParams,
+    ) -> Result<ExecResponse, codex_app_server_protocol::JSONRPCErrorError> {
+        self.require_initialized()?;
+        let process_id = params.process_id.clone();
+        {
+            let process_map = self.processes.lock().await;
```
starr-openai: race condition?

codex-rs/exec-server/src/server/handler.rs
```diff
@@ -0,0 +1,1032 @@
...
+        loop {
+            let response = {
+                let process_map = self.processes.lock().await;
+                let process = process_map.get(&params.process_id).ok_or_else(|| {
+                    invalid_request(format!("unknown process id {}", params.process_id))
+                })?;
+
+                let mut chunks = Vec::new();
+                let mut total_bytes = 0;
+                for retained in process.output.iter().filter(|chunk| chunk.seq > after_seq) {
```
starr-openai: will we block a process from writing stdout if we don't read stdout here?

codex-rs/exec-server/src/server/handler.rs
```diff
@@ -0,0 +1,1032 @@
...
+                        break;
+                    }
+                    total_bytes += chunk_len;
+                    chunks.push(ProcessOutputChunk {
+                        seq: retained.seq,
+                        stream: retained.stream,
+                        chunk: retained.chunk.clone().into(),
+                    });
+                    if total_bytes >= max_bytes {
+                        break;
```
starr-openai: lets log a warning / emit an event to indicate truncation?

codex-rs/exec-server/src/server/handler.rs
```diff
@@ -0,0 +1,1032 @@
...
+                    "stdin is closed for process {}",
+                    params.process_id
+                )));
+            }
+            process.session.writer_sender()
+        };
+
+        writer_tx
+            .send(params.chunk.into_inner())
+            .await
```
starr-openai: do we need to worry about this blocking other pending RPCs?

codex-rs/exec-server/src/protocol.rs
starr-openai: its a stub for now but I imagine we will need auth/identity later on

PR #14532, opened 2026-03-13
[hooks] stop continuation & stop_hook_active mechanics

eternal-openai:
Stop hooks now receive `stop_hook_active` and enable stop hooks to loop forever if they'd like to. In the initial hooks PR, we  implemented a simpler mechanic that the stop-blocking could only happen once in a row

- support stop hook adding a continuation prompt to add a further task
- if multiple stop-blocks happen that have continuation prompts, they are concatenated

example run:
```
› hey :)


• Running SessionStart hook: lighting the observatory

SessionStart hook (completed)
  warning: Hi, I'm a session start hook for wizard-tower (startup).
  hook context: Startup context: A wimboltine stonpet is an exotic cuisine from hyperspace

• Aloha :) Happy to jam with you. What are we building today?

• Running Stop hook: updating the guards

Stop hook (blocked)
  warning: Wizard Tower Stop hook continuing conversation
  feedback: cook the stonpet

• Aloha, here’s the hyperspace move for cooking a wimboltine stonpet:

  1. Sear the stonpet in a hot pan with moon-oil until the edges shimmer.
  2. Add star-lime, black salt, and a little fermented nebula paste.
  3. Lower the heat and let it braise for 8 cosmic minutes with a splash of comet broth.
  4. Finish with sky herbs and serve over warm asteroid rice.

  The vibe: crispy outside, tender center, deep interdimensional savor.

• Running Stop hook: updating the guards

Stop hook (completed)
  warning: Wizard Tower Stop hook saw a second pass and stayed calm to avoid a loop.
```

.codex/config.toml
```
[features]
codex_hooks = true
```

.codex/hooks.json
```
{
  "hooks": {
    "SessionStart": [
      {
        "matcher": "startup|resume",
        "hooks": [
          {
            "type": "command",
            "command": "/usr/bin/python3 .codex/hooks/session_start_demo.py",
            "timeoutSec": 10,
            "statusMessage": "lighting the observatory"
          }
        ]
      }
    ],
    "Stop": [
      {
        "hooks": [
          {
            "type": "command",
            "command": "/usr/bin/python3 .codex/hooks/stop_demo_block.py",
            "timeoutSec": 10,
            "statusMessage": "updating the guards"
          }
        ]
      }
    ]
  }
}
```

.codex/hooks/session_start_demo.py
```
#!/usr/bin/env python3

import json
import sys
from pathlib import Path


def main() -> int:
    payload = json.load(sys.stdin)
[...]

PR #14867, opened 2026-03-17
[hooks] use a user message > developer message for prompt continuation

eternal-openai:
## Summary

Persist Stop-hook continuation prompts as `user` messages instead of hidden `developer` messages + some requested integration tests

This is a followup to @pakrym 's comment in https://github.com/openai/codex/pull/14532 to make sure stop-block continuation prompts match training for turn loops

- Stop continuation now writes `<hook_prompt hook_run_id="...">stop hook's user prompt<hook_prompt>` 
- Introduces quick-xml dependency, though we already indirectly depended on it anyway via syntect
- This PR only has about 500 lines of actual logic changes, the rest is tests/schema

## Testing

Example run (with a sessionstart hook and 3 stop hooks) - this shows context added by session start, then two stop hooks sending their own additional prompts in a new turn. The model responds with a single message addressing both. Then when that turn ends, the hooks detect that they just ran using `stop_hook_active` and decide not to infinite loop

test files for this (unzip, move codex -> .codex): [codex.zip](https://github.com/user-attachments/files/26075806/codex.zip)

```
› cats


• Running SessionStart hook: lighting the observatory

SessionStart hook (completed)
  warning: Hi, I'm a session start hook for wizard-tower (startup).
  hook context: A wimboltine stonpet is an exotic cuisine from hyperspace

• Cats are tiny zen wizards, my friend: equal parts nap, mystery, and chaos. If you want, we can talk
  cat facts, cat breeds, cat names, or build something cat-themed in this repo.

• Running Stop hook: checking the tower wards

• Running Stop hook: sacking the guards

• Running Stop hook: hiring the guards

Stop hook (completed)
  warning: Wizard Tower Stop hook reviewed the completed reply (177 chars).

Stop hook (blocked)
  warning: Wizard Tower Stop hook continuing conversation
  feedback: cook the stonpet

Stop hook (blocked)
  warning: Wizard Tower Stop hook continuing conversation
  feedback: eat the cooked stonpet

• Stonpet’s cooked, aloha style: flash-seared over a blue quasiflame, glazed with nebula salt, and
  rested until the hyperspace juices settle.

  Now we eat with gratitude, my friend. One mindful bite in, and the flavor is pure cosmic surf:
  smoky, bright, and totally out of this dimension.

• Running Stop hook: checking the tower wards

• Running Stop hook: sacking the guards

• Running Stop hook: hiring the guards

Stop hook (completed)
  warning: Wizard Tower Stop hook reviewed the completed reply (285 chars).

Stop hook (completed)
  warning: Wizard Tower Stop hook saw a second pass and stayed calm to avoid a loop.

Stop hook (completed)
  warning: Wizard Tower Stop hook saw a second pass and stayed calm to avoid a loop.
```


codex-rs/core/src/compact_remote.rs
```diff
@@ -208,7 +208,7 @@ fn should_keep_compacted_history_item(item: &ResponseItem) -> bool {
         ResponseItem::Message { role, .. } if role == "user" => {
             matches!(
                 crate::event_mapping::parse_turn_item(item),
-                Some(TurnItem::UserMessage(_))
+                Some(TurnItem::UserMessage(_) | TurnItem::HookPrompt(_))
```
sayan-oai, line 211: should we also persist `HookPrompt` turn items for the local compaction used by non OpenAI providers [here](https://github.com/openai/codex/blob/20f2a216df3e2d534069438ca7126811de9ff89a/codex-rs/core/src/compact.rs#L257-L264)?

sayan-oai (approved):

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -3794,4 +3848,59 @@ mod tests {
...
+            OutgoingMessage::AppServerNotification(ServerNotification::ItemCompleted(
+                notification,
+            )) => {
+                assert_eq!(notification.thread_id, conversation_id.to_string());
+                assert_eq!(notification.turn_id, "turn-1");
+                assert_eq!(
+                    notification.item,
+                    ThreadItem::HookPrompt {
+                        id: notification.item.id().to_string(),
+                        fragments: vec![
```
pakrym-oai, line 3888: Consider `content` to align with user message.

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -3794,4 +3848,59 @@ mod tests {
...
+        let msg = recv_broadcast_message(&mut rx).await?;
+        match msg {
+            OutgoingMessage::AppServerNotification(ServerNotification::ItemCompleted(
+                notification,
+            )) => {
+                assert_eq!(notification.thread_id, conversation_id.to_string());
+                assert_eq!(notification.turn_id, "turn-1");
+                assert_eq!(
+                    notification.item,
+                    ThreadItem::HookPrompt {
```
pakrym-oai, line 3886: Consider HookMessage

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -4257,10 +4263,19 @@ pub enum ThreadItem {
     ContextCompaction { id: String },
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(rename_all = "camelCase", export_to = "v2/")]
+pub struct HookPromptFragment {
+    pub text: String,
+    pub hook_run_id: String,
```
pakrym-oai, line 4271: Do we expose this ID anywhere else?

codex-rs/core/src/compact_remote.rs
pakrym-oai: I think they should be persisted in both types of compaction. For remote compaction we filter out "special" user messages like content information but hook prompts might need to be persisted because they start the turn.

codex-rs/protocol/src/items.rs
```diff
@@ -199,6 +227,91 @@ impl UserMessageItem {
...
+    }
+
+    Some(HookPromptFragment { text, hook_run_id })
+}
+
+fn serialize_hook_prompt_fragment(text: &str, hook_run_id: &str) -> Option<String> {
+    if hook_run_id.trim().is_empty() {
+        return None;
+    }
+    to_xml_string(&HookPromptXml {
```
pakrym-oai, line 308: we do not typically use real XML serialization for context injection, we don't care about value escaping, just rough boundaries.

codex-rs/protocol/src/items.rs
```diff
@@ -199,6 +227,91 @@ impl UserMessageItem {
...
+        text: text.to_string(),
+        hook_run_id: hook_run_id.to_string(),
```
pakrym-oai, line 310: Will this end up in model context? Is it useful there?

codex-rs/app-server/src/bespoke_event_handling.rs
eternal-openai: Will do in a followup

codex-rs/app-server/src/bespoke_event_handling.rs
eternal-openai: Will do in a followup

codex-rs/app-server-protocol/src/protocol/v2.rs
eternal-openai: Not in the UIs, this is just to make the data model clean

codex-rs/protocol/src/items.rs
eternal-openai: The idea is just to be able to link the hook run to the prompt, but if you don't think that's important, I can remove it

codex-rs/protocol/src/items.rs
eternal-openai: Hmm, I see. Probably sounds like it's worth doing in case there are xml tags in the new user prompt. I imagine that's a bit paranoid, but probably safe?

PR #14878, opened 2026-03-17
feat: support remote_sync for plugin install/uninstall.

xl-openai:
- Added forceRemoteSync to plugin/install and plugin/uninstall.
- With forceRemoteSync=true, we update the remote plugin status first, then apply the local change only if the backend call succeeds.
- Kept plugin/list(forceRemoteSync=true) as the main recon path, and for now it treats remote enabled=false as uninstall. We
  will eventually migrate to plugin/installed for more precise state handling.


codex-rs/core/src/plugins/remote.rs
```diff
@@ -0,0 +1,266 @@
...
+        return Err(RemotePluginMutationError::UnexpectedStatus { url, status, body });
+    }
+
+    let parsed: RemotePluginMutationResponse =
+        serde_json::from_str(&body).map_err(|source| RemotePluginMutationError::Decode {
+            url: url.clone(),
+            source,
+        })?;
+    let expected_enabled = action == "enable";
+    if parsed.id != plugin_id {
```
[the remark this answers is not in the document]
xl-openai: it is OK.

codex-rs/core/src/plugins/remote.rs
```diff
@@ -0,0 +1,266 @@
...
+        .await
+        .map_err(|source| RemotePluginMutationError::Request {
+            url: url.clone(),
+            source,
+        })?;
+    let status = response.status();
+    let body = response.text().await.unwrap_or_default();
+    if !status.is_success() {
+        return Err(RemotePluginMutationError::UnexpectedStatus { url, status, body });
+    }
```
sayan-oai, line 224: codex says the backend uninstall endpoint returns 404 when the plugin isn't installed remotely. so if you have the plugin installed locally but not remotely and call uninstall with `forceRemoteSync=true`, it will fail here before local removal

codex-rs/core/src/plugins/manager.rs
```diff
@@ -655,10 +700,13 @@ impl PluginsManager {
                 );
                 continue;
             }
-            if remote_enabled_by_name
-                .insert(plugin.name.clone(), plugin.enabled)
-                .is_some()
-            {
+            // For now, sync treats remote `enabled = false` as uninstall rather than a distinct
+            // disabled state.
+            // TODO: Switch sync to `plugins/installed` so install and enable states stay distinct.
```
sayan-oai, line 705: clarifying, this means new app-server endpoint?

sayan-oai (approved):
approving to unblock, but left a comment with idempotency concerns.

codex-rs/core/src/plugins/manager.rs
```diff
@@ -545,6 +567,27 @@ impl PluginsManager {
 
     pub async fn uninstall_plugin(&self, plugin_id: String) -> Result<(), PluginUninstallError> {
         let plugin_id = PluginId::parse(&plugin_id)?;
+        self.uninstall_plugin_id(plugin_id).await
+    }
+
+    pub async fn uninstall_plugin_with_remote_sync(
```
sayan-oai, line 573: optional nit: maybe these methods should have `curated` in the name, or something to clearly convey this is only with the curated plugins (we dont support remote-syncing other marketplaces, or local plugins)?

codex-rs/core/src/plugins/remote.rs
xl-openai: which is OK? ideally plugins/list will re-sync?

codex-rs/core/src/plugins/manager.rs
xl-openai: no it is a chat endpoint. we will move from calling chat plugins/list to just plugins/installed.

codex-rs/core/src/plugins/manager.rs
xl-openai: Actually not really—we also have private plugins (with remote sync). The real question is whether desktop can support purely local plugins, which I don’t have an answer for yet.

PR #14888, opened 2026-03-17
Feat: reuse persisted model and reasoning effort on thread resume

shijie-oai:
## Summary

This PR makes `thread/resume` reuse persisted thread model metadata when the caller does not explicitly override it.

Changes:
- read persisted thread metadata from SQLite during `thread/resume`
- reuse persisted `model` and `model_reasoning_effort` as resume-time defaults
- fetch persisted metadata once and reuse it later in the resume response path
- keep thread summary loading on the existing rollout path, while reusing persisted metadata when available
- document the resume fallback behavior in the app-server README

## Why

Before this change, resuming a thread without explicit overrides derived `model` and `model_reasoning_effort` from current config, which could drift from the thread’s last persisted values. That meant a resumed thread could report and run with different model settings than the ones it previously used.

## Behavior

Precedence on `thread/resume` is now:
1. explicit resume overrides
2. persisted SQLite metadata for the thread
3. normal config resolution for the resumed cwd

### DIFF
--- a/codex-rs/app-server/README.md
+++ b/codex-rs/app-server/README.md
@@ -228,7 +228,9 @@ Start a fresh thread when you need a new Codex conversation.
 
 Valid `personality` values are `"friendly"`, `"pragmatic"`, and `"none"`. When `"none"` is selected, the personality placeholder is replaced with an empty string.
 
-To continue a stored session, call `thread/resume` with the `thread.id` you previously recorded. The response shape matches `thread/start`, and no additional notifications are emitted. You can also pass the same configuration overrides supported by `thread/start`, including `approvalsReviewer`. When `model` and `config.model_reasoning_effort` are not explicitly overridden, resume reuses the latest values persisted for that thread:
+To continue a stored session, call `thread/resume` with the `thread.id` you previously recorded. The response shape matches `thread/start`, and no additional notifications are emitted. You can also pass the same configuration overrides supported by `thread/start`, including `approvalsReviewer`.
+
+When no model-related resume override is provided, resume reuses the latest persisted `model` and `reasoningEffort` values for that thread. Supplying any of `model`, `modelProvider`, `config.model`, or `config.model_reasoning_effort` disables that persisted fallback and uses the explicit overrides plus normal config resolution instead:
 
 ```json
 { "method": "thread/resume", "id": 11, "params": {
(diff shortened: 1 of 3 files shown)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3481,18 +3489,22 @@ impl CodexMessageProcessor {
...
-                let Some(mut thread) = self
+                let mut thread = match self
                     .load_thread_from_resume_source_or_send_internal(
-                        request_id.clone(),
                         thread_id,
                         thread.as_ref(),
                         &response_history,
                         rollout_path.as_path(),
                         fallback_model_provider.as_str(),
+                        persisted_resume_metadata.as_ref(),
```
shijie-oai, line 3468: Pass down `persisted_resume_metadata` if available to avoid an additional db read.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -7496,6 +7519,36 @@ fn collect_resume_override_mismatches(
...
 }
 
+fn merge_persisted_resume_metadata(
+    request_overrides: &mut Option<HashMap<String, serde_json::Value>>,
+    typesafe_overrides: &mut ConfigOverrides,
+    persisted_metadata: &ThreadMetadata,
+) {
+    if has_model_resume_override(request_overrides.as_ref(), typesafe_overrides) {
+        return;
+    }
```
shijie-oai, line 7446: If any override is detected, we not merged the persisted model settings.

codex-rs/app-server/README.md
```diff
@@ -229,7 +229,9 @@ Start a fresh thread when you need a new Codex conversation.
 
 Valid `personality` values are `"friendly"`, `"pragmatic"`, and `"none"`. When `"none"` is selected, the personality placeholder is replaced with an empty string.
 
-To continue a stored session, call `thread/resume` with the `thread.id` you previously recorded. The response shape matches `thread/start`, and no additional notifications are emitted. You can also pass the same configuration overrides supported by `thread/start`, including `approvalsReviewer`:
+To continue a stored session, call `thread/resume` with the `thread.id` you previously recorded. The response shape matches `thread/start`, and no additional notifications are emitted. You can also pass the same configuration overrides supported by `thread/start`, including `approvalsReviewer`.
+
+When no model-related resume override is provided, resume reuses the latest persisted `model` and `reasoningEffort` values for that thread. Supplying any of `model`, `modelProvider`, `config.model`, or `config.model_reasoning_effort` disables that persisted fallback and uses the explicit overrides plus normal config resolution instead:
```
owenlin0: nit: codex language can be a bit robotic sounding haha

maybe something like this would be better
```
By default, resume uses the latest `model` and `reasoningEffort` values associated with the thread. Supplying any of...
```

owenlin0 (approved):

Issue #14875, opened 2026-03-17
Bubblewrap regression on Linux/WSL: missing writable_roots fail command startup and detached children are reaped


escape0707:
### What version of Codex CLI is running?

0.115.0-alpha.11+ (last known good: 0.115.0-alpha.9)

### What subscription do you have?

Pro

### Which model were you using?

gpt-5.4

### What platform is your computer?

Linux 6.6.87.2-microsoft-standard-WSL2 x86_64 unknown

### What terminal emulator and version are you using (if applicable)?

VSCode integrated terminal on WSL, Windows Terminal

### What issue are you seeing?

After <https://github.com/openai/codex/commit/04892b4ceb3ba6516eddd21c3f7ed2920a977442> (`refactor: make bubblewrap the default Linux sandbox`), Linux/WSL `workspace-write` behavior regressed in two ways:

1. If `sandbox_workspace_write.writable_roots` contains a well-formed path that does not exist, that sandboxed command invocation fails during setup.
2. If a normal non-elevated shell tool call starts a detached child process, that child is reaped when the one-shot tool call ends.

The first regression makes it harder to share or combine Windows Codex App and WSL Codex homes, because a Windows-only or misparsed `writable_root` is no longer harmless on Linux.

The second regression is what makes Playwright extension sessions unstable: the `Playwright MCP Bridge` bar appears, then disappears a few seconds later because the detached bridge process does not survive.

Environment:
- Host: Codex CLI, VSCode Codex Extension on WSL2
- OS: ArchLinux on WSL2
- Sandbox mode: `workspace-write`

Regression window:
- Last known good published version: `0.115.0-alpha.9`
- First known bad published version: `0.115.0-alpha.11`
- Exact last known good source commit: `b5f927b97327ab62047dba9eaf0e998bfaa7dab3`
- Likely introducing commit: `04892b4ceb3ba6516eddd21c3f7ed2920a977442`
- First runnable bad source commit: `e99e8e4a6bc2959266d4dd34c34e2b84e472ac52`

### Regression A: Missing `writable_root` fails command startup

1. Configure `workspace-write` with one existing and one missing writable root:

```toml
sandbox_mode = "workspace-write"

[sandbox_workspace_write]
network_access = true
writable_roots = [
  "~/.cache/codex-writable-root-existing",
  "~/.cache/codex-writable-root-missing",
]
```

Assume `~/.cache/codex-writable-root-existing` exists and `~/.cache/codex-writable-root-missing` does not.

[...]

PR #14890, opened 2026-03-17
fix(linux-sandbox): ignore missing writable roots

viyatb-oai:
## Summary
- skip nonexistent `workspace-write` writable roots in the Linux bubblewrap mount builder instead of aborting sandbox startup
- keep existing writable roots mounted normally so mixed Windows/WSL configs continue to work
- add unit and Linux integration regression coverage for the missing-root case

## Context
This addresses regression A from #14875. Regression B will be handled in a separate PR.

The old bubblewrap integration added `ensure_mount_targets_exist` as a preflight guard because bubblewrap bind targets must exist, and failing early let Codex return a clearer error than a lower-level mount failure.

That policy turned out to be too strict once bubblewrap became the default Linux sandbox: shared Windows/WSL or mixed-platform configs can legitimately contain a well-formed writable root that does not exist on the current machine. This PR keeps bubblewrap's existing-target requirement, but changes Codex to skip missing writable roots instead of treating them as fatal configuration errors.

### DIFF
--- a/codex-rs/linux-sandbox/src/bwrap.rs
+++ b/codex-rs/linux-sandbox/src/bwrap.rs
@@ -211,8 +211,14 @@ fn create_filesystem_args(
     file_system_sandbox_policy: &FileSystemSandboxPolicy,
     cwd: &Path,
 ) -> Result<BwrapArgs> {
-    let writable_roots =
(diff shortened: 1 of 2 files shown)


etraut-openai (approved):
Looks good!

PR #14891, opened 2026-03-17
fix(linux-sandbox): preserve detached children

viyatb-oai:
## Summary
- make application-generated Linux sandbox invocations explicitly allow intentionally detached descendants to survive after the initial one-shot helper exits
- omit bubblewrap `--die-with-parent` only for that opt-in path, while preserving the default behavior and retaining `--unshare-user` and `--unshare-pid`
- add Linux regression coverage for a detached child that continues writing after the parent exits

## Context
This addresses regression B from #14875. Regression A was fixed separately by #14890.

The regression is not caused by the PID namespace by itself. An intentionally detached child is torn down when bubblewrap is asked to tie sandbox lifetime to the original parent process. The normal application Linux-sandbox command path now opts into detached-child lifetime handling, and the Linux sandbox omits `--die-with-parent` only for that explicit request. Direct/default helper behavior remains unchanged.

This keeps the fix within sandboxing and Linux-sandbox code rather than adding lifetime behavior in `core/exec`, while retaining user and PID namespace isolation.

## Validation
- `just fmt`
- `just test -p codex-sandboxing`
- `cargo clippy -p codex-sandboxing -p codex-linux-sandbox --all-targets -- -D warnings`
- `just argument-comment-lint -p codex-sandboxing -p codex-linux-sandbox`
- `cargo clean`

Note: this checkout is macOS, so the Linux-only detached-child runtime regression test runs in Linux CI rather than locally.


codex-rs/core/src/exec.rs
```diff
@@ -198,6 +198,7 @@ pub async fn process_exec_tool_call(
         codex_linux_sandbox_exe,
         use_legacy_landlock,
     )?;
+    exec_req.allow_detached_children_in_linux_sandbox();
```
bolinfest: Why isn't this done as part of `build_exec_request()` so we don't have to create it as `mut`?

codex-rs/linux-sandbox/tests/suite/landlock.rs
```diff
@@ -310,6 +310,58 @@ async fn test_writable_root() {
...
+        "nohup sh -c 'echo \"$$\" > \"{pid}\"; while true; do printf x >> \"{log}\"; sleep 0.2; done' >/dev/null 2>&1 </dev/null &",
+        pid = pid_path.to_string_lossy(),
+        log = log_path.to_string_lossy(),
+    );
+
+    let output = run_cmd_result_with_writable_roots(
+        &["bash", "-lc", &command],
+        &[tempdir.path().to_path_buf()],
+        LONG_TIMEOUT_MS,
+        false,
```
bolinfest: linter will probably require comments for these?

codex-rs/linux-sandbox/tests/suite/landlock.rs
```diff
@@ -310,6 +310,58 @@ async fn test_writable_root() {
...
+        true,
+    )
+    .await
+    .expect("sandboxed command should execute");
+
+    assert_eq!(output.exit_code, 0);
+
+    tokio::time::sleep(std::time::Duration::from_millis(700)).await;
```
bolinfest: Import `Duration` above maybe?

codex-rs/core/src/exec.rs
```diff
@@ -203,6 +203,28 @@ pub async fn process_exec_tool_call(
...
+fn build_process_exec_tool_request(
+    params: ExecParams,
+    sandbox_policy: &SandboxPolicy,
+    file_system_sandbox_policy: &FileSystemSandboxPolicy,
+    network_sandbox_policy: NetworkSandboxPolicy,
+    sandbox_cwd: &Path,
+    codex_linux_sandbox_exe: &Option<PathBuf>,
+    use_legacy_landlock: bool,
+) -> Result<ExecRequest> {
+    let mut exec_req = build_exec_request(
```
bolinfest: Did the `mut` just move to here? This one method on `ExecRequest` smells funny to me (as does the `mut`).

viyatb-oai: yeah fixing this correctly.

escape0707:
> Closing this pull request because it has had no updates for more than 14 days. If you plan to continue working on it, feel free to reopen or open a new PR.

That's unfortunate.

PR #14892, opened 2026-03-17
Fix tui_app_server: ignore duplicate legacy stream events

etraut-openai:
The in-process app-server currently emits both typed `ServerNotification`s and legacy `codex/event/*` notifications for the same live turn updates. `tui_app_server` was consuming both paths, so message deltas and completed items could be enqueued twice and rendered as duplicated output in the transcript.

Ignore legacy notifications for event types that already have typed (app server) notification handling, while keeping legacy fallback behavior for events that still only arrive on the old path. This preserves compatibility without duplicating streamed commentary or final agent output.

We will remove all of the legacy event handlers over time; they're here only during the short window where we're moving the tui to use the app server.


etraut-openai:
@codex review

dylan-hurd-oai (approved):

PR #14896, opened 2026-03-17
[plugins] Support plugin installation elicitation.

mzeng-openai:
It now supports:

- Connectors that are from installed and enabled plugins that are not installed yet
- Plugins that are on the allowlist that are not installed yet.


codex-rs/core/src/tools/handlers/tool_suggest.rs
```diff
@@ -59,7 +59,8 @@ struct ToolSuggestMeta<'a> {
     suggest_reason: &'a str,
     tool_id: &'a str,
     tool_name: &'a str,
-    install_url: &'a str,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    install_url: Option<&'a str>,
```
sayan-oai, line 64: do we need to update the TUI consumer? it still says `install_url` is a string [here](https://sg.apps.openai.org/github.com/openai/codex/-/blob/codex-rs/tui/src/bottom_pane/mcp_server_elicitation.rs?L152).

codex-rs/core/src/tools/handlers/tool_suggest_tests.rs
```diff
@@ -1,5 +1,107 @@
...
+use codex_app_server_protocol::AppInfo;
+use codex_utils_absolute_path::AbsolutePathBuf;
 use pretty_assertions::assert_eq;
+use serde_json::json;
+use std::collections::BTreeMap;
+use std::fs;
+use std::path::Path;
+use tempfile::tempdir;
+
+fn write_file(path: &Path, contents: &str) {
```
sayan-oai: these helpers are duplicated across a few files, and are similar to the ones in plugins/manager_test.rs. can we consolidate?

codex-rs/core/src/tools/discoverable.rs
```diff
@@ -109,3 +117,20 @@ impl From<PluginCapabilitySummary> for DiscoverablePluginInfo {
         }
     }
 }
+
+impl From<PluginDetailSummary> for DiscoverablePluginInfo {
+    fn from(value: PluginDetailSummary) -> Self {
+        Self {
+            id: value.id,
+            name: value.name,
+            description: value.description,
```
sayan-oai: i think we do some normalization on the description in PluginCapabilitySummary to make it more prompt-friendly (normalization/truncation). PluginDetailSummary doesn't. Could we consolidate these types somehow? all three are similar. We could also convert PluginDetailSummary -> PluginCapabilitySummary to keep the normalization, and only use that for DiscoverablePluginInfo.

codex-rs/core/src/tools/handlers/tool_suggest.rs
```diff
@@ -279,18 +231,74 @@ fn build_tool_suggestion_meta<'a>(
...
+                Err(err) => {
+                    warn!(
+                        "failed to refresh codex apps tools cache after tool suggestion for {}: {err:#}",
+                        connector.id
+                    );
+                    false
+                }
+            }
+        }
+        DiscoverableTool::Plugin(plugin) => {
```
sayan-oai, line 282: does this assume the client has already run the local `plugin/install` before accepting? the app does, but the TUI just opens the browser, refreshes connectors, and resolves the elicitation, so will this check ever be true there?

codex-rs/core/src/connectors.rs
```diff
@@ -116,13 +106,24 @@ pub(crate) async fn list_tool_suggest_discoverable_tools_with_auth(
...
+        &connector_ids,
+    )
+    .into_iter()
+    .map(DiscoverableTool::from);
+    let discoverable_plugins = list_tool_suggest_discoverable_plugins(config)?
+        .into_iter()
+        .map(DiscoverablePluginInfo::from)
+        .map(DiscoverableTool::from);
+    Ok(discoverable_connectors
+        .chain(discoverable_plugins)
```
sayan-oai, line 125: nit: this seems like a surprising place to chain plugins. should this function live outside of connectors.rs?

codex-rs/core/src/tools/handlers/tool_suggest.rs
mzeng-openai: I've disabled it for TUI because there's no plugin installation flow here yet.

codex-rs/core/src/connectors.rs
```diff
@@ -377,6 +375,16 @@ fn filter_tool_suggest_discoverable_tools(
     connectors
 }
 
+fn tool_suggest_connector_ids(config: &Config) -> HashSet<String> {
+    PluginsManager::new(config.codex_home.clone())
+        .plugins_for_config(config)
```
sayan-oai, line 380: confirming-- we only want to suggest connectors that are associated with plugins? if they have no plugins installed, this would be empty, right?

codex-rs/core/src/tools/handlers/tool_suggest.rs
```diff
@@ -95,15 +97,16 @@ impl ToolHandler for ToolSuggestHandler {
                 "suggest_reason must not be empty".to_string(),
             ));
         }
-        if args.tool_type == DiscoverableToolType::Plugin {
+        if args.action_type != DiscoverableToolAction::Install {
             return Err(FunctionCallError::RespondToModel(
-                "plugin tool suggestions are not currently available".to_string(),
+                "tool suggestions currently support only action_type=\"install\"".to_string(),
```
sayan-oai, line 102: should we not tell the model that `enable` is an option [here](https://sg.apps.openai.org/github.com/openai/codex/-/blob/codex-rs/core/src/tools/spec.rs?L1763)?

sayan-oai (approved):

PR #14954, opened 2026-03-17
Add dynamic tool call support to tui_app_server

etraut-openai:
- add a `tui_app_server` dynamic tool registry and dispatch path for TUI-owned dynamic tools
- register dynamic tool specs on `thread/start` and resolve incoming `DynamicToolCall` requests through the existing app-server request handle
- keep dynamic tool transcript rendering unchanged while removing the old unsupported-request rejection path
- factor the new app-layer integration logic out of `app.rs` into focused adapter/session code
- add request-handle helpers in `codex-app-server-client` and update lockfiles for the new test dependencies

### DIFF
--- a/codex-rs/Cargo.lock
+++ b/codex-rs/Cargo.lock
@@ -2585,6 +2585,7 @@ name = "codex-tui-app-server"
 version = "0.0.0"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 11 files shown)


etraut-openai:
@codex review

PR #14956, opened 2026-03-17
Bound Windows PowerShell parser hangs

aibrahim-oai:
## What is flaky
`codex-rs/shell-command/src/command_safety/windows_safe_commands.rs` had intermittent Windows failures in `recognizes_safe_powershell_wrappers` and `accepts_full_path_powershell_invocations`.

## Why it was flaky
Those tests exercise the same PowerShell AST helper that `is_safe_command` uses to classify wrapper commands.

- The helper shells out to PowerShell and waits for it to parse the command AST.
- On some Windows CI runners, PowerShell startup can stall during process initialization.
- After the first timeout hardening, the helper no longer used `Command::output()`, which meant stdin was inherited unless it was set explicitly.
- An inherited stdin let the parser process sit around waiting for input until the timeout budget expired, even for commands that should have parsed immediately.

So the tests were not actually flaking on parsing logic. They were flaking because the helper process lifecycle depended on runner startup conditions and stdin wiring.

## How this PR fixes it
- Keep the parser bounded by a strict 5 second cap and fail closed if the child does not finish.
- Poll and terminate a stuck parser child instead of waiting indefinitely.
- Add a regression test for the stuck-child path.
- Set the parser subprocess stdin to `Stdio::null()` so it matches the old `Command::output()` behavior and never waits on inherited input.

## Why this fix fixes the flakiness
The safe-command classifier now has deterministic process semantics: the helper either exits quickly with parse output or it is treated as failed within the bounded window. There is no longer a path where a safe wrapper depends on ambient stdin state or on a half-started PowerShell process lingering until timeout.

### DIFF
--- a/codex-rs/shell-command/src/command_safety/windows_safe_commands.rs
+++ b/codex-rs/shell-command/src/command_safety/windows_safe_commands.rs
@@ -3,6 +3,7 @@ use base64::engine::general_purpose::STANDARD as BASE64_STANDARD;
 use serde::Deserialize;
 use std::io::Read;
 use std::path::Path;
+use std::process::Child;
 use std::process::Command;
 use std::process::Output;
 use std::process::Stdio;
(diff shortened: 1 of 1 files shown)


codex-rs/shell-command/src/command_safety/windows_safe_commands.rs
```diff
@@ -1,11 +1,19 @@
...
+use std::process::Output;
+use std::process::Stdio;
 use std::sync::LazyLock;
+use std::thread;
+use std::time::Duration;
+use std::time::Instant;
 
 const POWERSHELL_PARSER_SCRIPT: &str = include_str!("powershell_parser.ps1");
+const POWERSHELL_PARSER_TIMEOUT: Duration = Duration::from_secs(5);
+const POWERSHELL_PARSER_POLL_INTERVAL: Duration = Duration::from_millis(10);
```
iceweasel-oai, line 17: this could probably be a bit higher, but I don't care that much

iceweasel-oai (approved):

aibrahim-oai:
@codex review

PR #14960, opened 2026-03-17
Add FS abstraction and use in view_image

pakrym-oai:
Adds an environment crate and environment + file system abstraction.

Environment is a combination of attributes and services specific to environment the agent is connected to:
File system, process management, OS, default shell.

The goal is to move most of agent logic that assumes environment to work through the environment abstraction.


codex-rs/environment/src/fs.rs
```diff
@@ -0,0 +1,316 @@
...
+    ) -> FileSystemResult<()>;
+}
+
+#[derive(Clone, Default)]
+pub(crate) struct LocalFileSystem;
+
+#[async_trait]
+impl FileSystem for LocalFileSystem {
+    async fn read_file(&self, path: &Path) -> FileSystemResult<Vec<u8>> {
+        tokio::fs::read(path).await
```
starr-openai: do we need to guard against reading a file that is too large here?

codex-rs/environment/src/fs.rs
```diff
@@ -0,0 +1,316 @@
...
+                "fs/copy only supports regular files, directories, and symlinks",
+            ))
+        })
+        .await
+        .map_err(|err| io::Error::other(format!("filesystem task failed: {err}")))?
+    }
+}
+
+fn copy_dir_recursive(source: &Path, target: &Path) -> io::Result<()> {
+    std::fs::create_dir_all(target)?;
```
bolinfest, line 207: This file seems to contain a mix of `std::fs` and `tokio::fs`: why isn't it async all the way through?

codex-rs/environment/src/fs.rs
```diff
@@ -0,0 +1,316 @@
...
+    pub file_name: String,
+    pub is_directory: bool,
+    pub is_file: bool,
+}
+
+pub type FileSystemResult<T> = io::Result<T>;
+
+#[async_trait]
+pub trait FileSystem: Send + Sync {
+    async fn read_file(&self, path: &Path) -> FileSystemResult<Vec<u8>>;
```
bolinfest: Can we favor `AbsolutePath` over `Path`?

codex-rs/core/src/state/service.rs
```diff
@@ -61,4 +62,5 @@ pub(crate) struct SessionServices {
     /// Session-scoped model client shared across turns.
     pub(crate) model_client: ModelClient,
     pub(crate) code_mode_service: CodeModeService,
+    pub(crate) environment: Arc<Environment>,
```
bolinfest, line 65: This is intended to be an `ExecutorEnvironment`, is that right?

codex-rs/core/src/tools/handlers/view_image.rs
```diff
@@ -87,22 +88,41 @@ impl ToolHandler for ViewImageHandler {
...
-        let metadata = fs::metadata(&abs_path).await.map_err(|error| {
-            FunctionCallError::RespondToModel(format!(
-                "unable to locate image at `{}`: {error}",
-                abs_path.display()
-            ))
-        })?;
-
-        if !metadata.is_file() {
+        let abs_path =
+            AbsolutePathBuf::try_from(turn.resolve_path(Some(args.path))).map_err(|error| {
```
bolinfest, line 92: Hmm, so ultimately, this `view_image` tool should be run on the executor? So `resolve_path()` will work since it will be speaking in terms of its own local path?

Though I don't know if it will ultimately have `TurnContext`?

codex-rs/environment/src/fs.rs
```diff
@@ -0,0 +1,332 @@
...
+                io::ErrorKind::InvalidInput,
+                "fs/copy only supports regular files, directories, and symlinks",
+            ))
+        })
+        .await
+        .map_err(|err| io::Error::other(format!("filesystem task failed: {err}")))?
+    }
+}
+
+fn copy_dir_recursive(source: &Path, target: &Path) -> io::Result<()> {
```
bolinfest, line 206: I see, given the recursion, I guess sync makes this all a bit simpler. Maybe we add a `_blocking` suffix to this name?

codex-rs/environment/src/fs.rs
```diff
@@ -0,0 +1,332 @@
...
+        }
+    }
+    Ok(())
+}
+
+fn destination_is_same_or_descendant_of_source(
+    source: &Path,
+    destination: &Path,
+) -> io::Result<bool> {
+    let source = std::fs::canonicalize(source)?;
```
bolinfest, line 229: Hmm, this is tricky. Canonicalization is not always what you want here?

codex-rs/environment/src/fs.rs
bolinfest, line 206: Should we just use `fs_extra`?

codex-rs/core/src/tools/handlers/view_image.rs
pakrym-oai: We'll move resolve_path to be remote path aware.

codex-rs/core/src/state/service.rs
pakrym-oai: yes, do you want it renamed? type or type and field?

bolinfest (approved):

PR #14963, opened 2026-03-17
fix(linux-sandbox): prefer system /usr/bin/bwrap when available

viyatb-oai:
## Problem
Ubuntu/AppArmor hosts started failing in the default Linux sandbox path after the switch to vendored/default bubblewrap in `0.115.0`.

The clearest report is in [#14919](https://github.com/openai/codex/issues/14919), especially [this investigation comment](https://github.com/openai/codex/issues/14919#issuecomment-4076504751): on affected Ubuntu systems, `/usr/bin/bwrap` works, but a copied or vendored `bwrap` binary fails with errors like `bwrap: setting up uid map: Permission denied` or `bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted`.

The root cause is Ubuntu's `/etc/apparmor.d/bwrap-userns-restrict` profile, which grants `userns` access specifically to `/usr/bin/bwrap`. Once Codex started using a vendored/internal bubblewrap path, that path was no longer covered by the distro AppArmor exception, so sandbox namespace setup could fail even when user namespaces were otherwise enabled and `uidmap` was installed.

## What this PR changes
- prefer system `/usr/bin/bwrap` whenever it is available
- keep vendored bubblewrap as the fallback when `/usr/bin/bwrap` is missing
- when `/usr/bin/bwrap` is missing, surface a Codex startup warning through the app-server/TUI warning path instead of printing directly from the sandbox helper with `eprintln!`
- use the same launcher decision for both the main sandbox execution path and the `/proc` preflight path
- document the updated Linux bubblewrap behavior in the Linux sandbox and core READMEs

## Why this fix
This still fixes the Ubuntu/AppArmor regression from [#14919](https://github.com/openai/codex/issues/14919), but it keeps the runtime rule simple and platform-agnostic: if the standard system bubblewrap is installed, use it; otherwise fall back to the vendored helper.

The warning now follows that same simple rule. If Codex cannot find `/usr/bin/bwrap`, it tells the user that it is falling back to the vendored helper, and it does so through the existing startup warning plumbing that reaches the TUI and app-server instead of low-level sandbox stderr.

## Testing
- `cargo test -p codex-app-server --lib`
- `cargo test -p codex-tui-app-server tests::embedded_app_server_start_failure_is_returned`
- `cargo clippy -p codex-linux-sandbox --all-targets`
- `cargo clippy -p codex-app-server --all-targets`
- `cargo clippy -p codex-tui-app-server --all-targets`


codex-rs/linux-sandbox/src/launcher.rs
```diff
@@ -0,0 +1,158 @@
+use std::ffi::CString;
+use std::fs::File;
+use std::os::fd::AsRawFd;
+use std::os::raw::c_char;
+use std::os::unix::ffi::OsStrExt;
+use std::path::Path;
+use std::path::PathBuf;
```
bolinfest: Consider `AbsolutePathBuf` for good measure?

codex-rs/linux-sandbox/src/launcher.rs
```diff
@@ -0,0 +1,136 @@
...
+                    "warning: Codex could not find system bubblewrap at {SYSTEM_BWRAP_PATH}. Please install bubblewrap with your package manager. Codex will use the vendored bubblewrap in the meantime."
+                );
+            }
+            exec_vendored_bwrap(argv, preserved_files)
+        }
+    }
+}
+
+fn preferred_bwrap_launcher() -> BubblewrapLauncher {
+    if Path::new(SYSTEM_BWRAP_PATH).exists() {
```
bolinfest: Slightly more pedantic?

```suggestion
    if Path::new(SYSTEM_BWRAP_PATH).is_file() {
```

codex-rs/linux-sandbox/src/launcher.rs
```diff
@@ -0,0 +1,136 @@
...
+        BubblewrapLauncher::System(PathBuf::from(SYSTEM_BWRAP_PATH))
+    } else {
+        BubblewrapLauncher::Vendored
+    }
+}
+
+fn exec_system_bwrap(program: &Path, argv: Vec<String>, preserved_files: Vec<File>) -> ! {
+    make_files_inheritable(&preserved_files);
```
bolinfest, line 44: Why is this being done?

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -434,7 +433,11 @@ fn run_bwrap_with_proc_fallback(
         command_cwd,
         options,
     );
-    exec_vendored_bwrap(bwrap_args.args, bwrap_args.preserved_files);
+    exec_bwrap(
+        bwrap_args.args,
+        bwrap_args.preserved_files,
```
bolinfest: Oh, I see, we already had this special list.

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -434,7 +433,11 @@ fn run_bwrap_with_proc_fallback(
...
+        /*emit_missing_system_bwrap_warning*/ true,
```
bolinfest: The linter works!!!

bolinfest: But also, why should we categorically display the warning? I think the model is likely to see it, but not the user?

codex-rs/linux-sandbox/src/launcher.rs
viyatb-oai: adding a comment in there

codex-rs/linux-sandbox/src/linux_run_main.rs
viyatb-oai: I think we should display in the warning in the notifications that are shown in the tui

codex-rs/linux-sandbox/src/launcher.rs
```diff
@@ -0,0 +1,136 @@
...
+pub(crate) fn exec_bwrap(
+    argv: Vec<String>,
+    preserved_files: Vec<File>,
+    emit_missing_system_bwrap_warning: bool,
+) -> ! {
+    match preferred_bwrap_launcher() {
+        BubblewrapLauncher::System(program) => exec_system_bwrap(&program, argv, preserved_files),
+        BubblewrapLauncher::Vendored => {
+            if emit_missing_system_bwrap_warning {
+                eprintln!(
```
bolinfest: It feels like the warning should be sent upstream of this point? Via `EventMsg::Warning` or the app-server equivalent?

And when we make that decision, should we perform a runtime test with our vendored `bwrap` to see if we think it works? Because if it does, we shouldn't bother the user with this?

viyatb-oai: Moved the missing-system-bwrap notice out of the launcher and into startup warnings that surface through the app-server and TUI notification path. I also narrowed the warning to Linux hosts that actually ship `/etc/apparmor.d/bwrap-userns-restrict` and are missing `/usr/bin/bwrap`, so we only bother users on the AppArmor-affected fallback path instead of every host that can run with vendored bubblewrap.

codex-rs/app-server/src/lib.rs
```diff
@@ -310,6 +316,24 @@ fn project_config_warning(config: &Config) -> Option<ConfigWarningNotification>
     })
 }
 
+#[cfg(target_os = "linux")]
+fn missing_system_bwrap_warning() -> Option<String> {
```
etraut-openai: Why are we issuing an error message from the `tui_app_server` code? Doesn't this condition potentially affect all surfaces, including legacy `tui`  mode, non-interactive `exec`, and app server clients? This doesn't seem like the right place to put this error message.

etraut-openai (changes requested):

codex-rs/app-server/src/lib.rs
```diff
@@ -435,7 +436,7 @@ pub async fn run_main_with_transport(
     };
     let loader_overrides_for_config_api = loader_overrides.clone();
     let mut config_warnings = Vec::new();
-    let config = match ConfigBuilder::default()
+    let mut config = match ConfigBuilder::default()
```
bolinfest: I am always a bit suspicious when something as important as `Config` has not been `mut` for a long time and then we need to change that invariant...

viyatb-oai: fixed.

codex-rs/app-server/src/lib.rs
viyatb-oai: fixed

etraut-openai (approved):
I didn't review the sandbox changes (I presume bolinfest did). The error reporting looks good to me now.

bolinfest (approved):
Thanks for jumping on this!

PR #14970, opened 2026-03-17
Simple directory mentions

canvrno-oai:
- Adds simple support for directory mentions in the TUI.
- Codex App/VS Code will require minor change to recognize a directory mention as such and change the link behavior.
- Directory mentions have a trailing slash to differentiate from extensionless files

image
image


codex-rs/file-search/src/lib.rs
```diff
@@ -448,14 +449,13 @@ fn walker_worker(
...
-            if entry.file_type().is_some_and(|ft| ft.is_dir()) {
-                return ignore::WalkState::Continue;
-            }
             let path = entry.path();
             let Some(full_path) = path.to_str() else {
                 return ignore::WalkState::Continue;
             };
-            if let Some((_, relative_path)) = get_file_path(path, &search_directories) {
+            if let Some((_, relative_path)) = get_file_path(path, &search_directories)
+                && !relative_path.is_empty()
```
dylan-hurd-oai: @codex when would relative_path potentially be empty?

codex-rs/file-search/src/lib.rs
```diff
@@ -552,9 +552,13 @@ fn matcher_worker(
                             } else {
                                 None
                             };
+                            let mut display_path = relative_path.to_string();
+                            if Path::new(full_path).is_dir() && !display_path.ends_with('/') {
+                                display_path.push('/');
+                            }
```
dylan-hurd-oai: We should have a comment for this - why is it required? If we're using a trailing slash to differentiate between files and directories, could we do so with structured data instead? e.g. add a `DirectoryMatch` alongside a `FileMatch`, or include a `type` field in the match object

canvrno-oai: The trailing slash is indeed to differentiate between files and directories. Two motivations behind this:
- Make it clear in context that this is a directory
- Avoiding a match type for simplicity

I can add a match type and update the App PR to utilize this, but should we keep the trailing slash in context?

dylan-hurd-oai: Not opposed to keeping the trailing slash! Just feels risky to make it load-bearing

dylan-hurd-oai (approved):

PR #14988, opened 2026-03-17
Add thread/shellCommand to app server API surface

etraut-openai:
This PR adds a new `thread/shellCommand` app server API so clients can implement `!` shell commands. These commands are executed within the sandbox, and the command text and output are visible to the model.

The internal implementation mirrors the current TUI `!` behavior.
- persist shell command execution as `CommandExecution` thread items, including source and formatted output metadata
- bridge live and replayed app-server command execution events back into the existing `tui_app_server` exec rendering path

This PR also wires `tui_app_server` to submit `!` commands through the new API.


etraut-openai:
@codex review

codex-rs/app-server/README.md
```diff
@@ -410,6 +411,23 @@ While compaction is running, the thread is effectively in a turn so clients shou
...
+
+Use `thread/shellCommand` for the TUI `!` workflow. The request returns immediately with `{}`.
+
+Progress is emitted as standard `item/*` notifications on the same `threadId`:
+
+- `item/started` with `item: { "type": "commandExecution", "source": "userShell", ... }`
+- zero or more `item/commandExecution/outputDelta`
+- `item/completed` with the same `commandExecution` item id
+
+If the thread already has an active turn, the command runs as an auxiliary action on that turn and the formatted output is injected into the turn’s message stream with the same behavior as legacy TUI `!` commands.
```
owenlin0: Hm, what happens if there isn't an active turn? Does it also emit `turn/started` and `turn/completed`? All items are associated with a turn, so it'd be weird if it didn't. But from testing the TUI, a standalone `!` doesn't look like a turn?

owenlin0: Or... does it "queue" the commandExecution item for when the next turn starts?

etraut-openai: It creates a standalone turn when there isn’t already an active one. This mirrors how it works today in the TUI if you use a `!` command.

`StandaloneTurn` mode emits the normal turn lifecycle itself, so you get a real turn with `turn/started`, the `item/*` notifications for the `commandExecution`, and then `turn/completed`.

The reason it may not look like a normal turn in the TUI is just presentation. For `!` we render the shell execution as the exec cell rather than as a user message.

owenlin0: ah okay, if this is how it works in core then that makes sense to me. probably worth updating the README to explain what happens when there isn't an active turn (i.e. it starts one, emits a single item for commandExecution and ends?)

etraut-openai: Good call. README updated.

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2874,6 +2876,19 @@ pub struct ThreadCompactStartParams {
 #[ts(export_to = "v2/")]
 pub struct ThreadCompactStartResponse {}
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ThreadShellCommandParams {
+    pub thread_id: String,
+    pub command: String,
```
owenlin0, line 2888: hmm, we use this for `command/exec`, should we be consistent?

```
    /// Command argv vector. Empty arrays are rejected.
    pub command: Vec<String>,
```

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -4139,13 +4154,18 @@ pub enum ThreadItem {
...
         status: CommandExecutionStatus,
         /// A best-effort parsing of the command to understand the action(s) it will perform.
         /// This returns a list of CommandAction objects because a single shell command may
         /// be composed of many commands piped together.
         command_actions: Vec<CommandAction>,
         /// The command's output, aggregated from stdout and stderr.
         aggregated_output: Option<String>,
+        /// The command's output after applying the model-facing formatter/truncation policy.
+        #[serde(default)]
+        formatted_output: String,
```
owenlin0: Do we need to expose this? I'd avoid, since this could bloat the payload quite a bit. Could we just rely on the existing `aggregated_output`?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -4865,14 +4896,19 @@ pub struct TerminalInteractionNotification {
...
 pub struct CommandExecutionOutputDeltaNotification {
     pub thread_id: String,
     pub turn_id: String,
     pub item_id: String,
-    pub delta: String,
+    /// Raw bytes from the command stream (may not be valid UTF-8).
+    #[serde_as(as = "serde_with::base64::Base64")]
+    #[schemars(with = "String")]
+    #[ts(type = "string")]
+    pub delta: Vec<u8>,
```
owenlin0: hmm this seems like it could be a breaking change

codex-rs/app-server-protocol/src/protocol/v2.rs
etraut-openai: `command/exec` requires the caller to do their own shell expansion. This new call assumes that the string will be sent to the shell "as is" for interpretation by the shell itself. I don't think it makes sense to make this a vector of strings in this case.

codex-rs/app-server-protocol/src/protocol/v2.rs
etraut-openai: Good point. I've changed this.

codex-rs/app-server-protocol/src/protocol/v2.rs
etraut-openai: After thinking about this more, I don't think that clients are generally going to be able to do much with raw byte output anyway. We'll stick with best-effort conversion to u8 text. I've undone the change to this structure.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2972,6 +2978,58 @@ impl CodexMessageProcessor {
...
+                self.outgoing.send_error(request_id, error).await;
+                return;
+            }
+        };
+
+        match self
+            .submit_core_op(
+                &request_id,
+                thread.as_ref(),
+                Op::RunUserShellCommand { command },
```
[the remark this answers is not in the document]
etraut-openai: This is correct behavior, not a bug.

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -1580,7 +1582,6 @@ pub(crate) async fn apply_bespoke_event_handling(
         }
         EventMsg::ExecCommandOutputDelta(exec_command_output_delta_event) => {
             let item_id = exec_command_output_delta_event.call_id.clone();
-            let delta = String::from_utf8_lossy(&exec_command_output_delta_event.chunk).to_string();
```
owenlin0, line 1583: nit: felt cleaner to keep it at the top. non-blocking though

owenlin0 (approved):
lgtm!

PR #15011, opened 2026-03-18
Forward session and turn headers to MCP HTTP requests

nicholasclark-openai:
## Summary
- forward request-scoped task headers through MCP tool metadata lookups and tool calls
- apply those headers to streamable HTTP initialize, tools/list, and tools/call requests
- update affected rmcp/core tests for the new request_headers plumbing

## Testing
- cargo test -p codex-rmcp-client
- cargo test -p codex-core (fails on pre-existing unrelated error in core/src/auth_env_telemetry.rs: missing websocket_connect_timeout_ms in ModelProviderInfo initializer)
- just fix -p codex-rmcp-client
- just fix -p codex-core (hits the same unrelated auth_env_telemetry.rs error)
- just fmt


chenglong-openai (approved):

codex-rs/core/src/mcp_connection_manager.rs
```diff
@@ -378,7 +378,7 @@ struct ManagedClient {
 }
 
 impl ManagedClient {
-    fn listed_tools(&self) -> Vec<ToolInfo> {
+    fn listed_tools(&self, _request_headers: Option<reqwest::header::HeaderMap>) -> Vec<ToolInfo> {
```
pakrym-oai: why unused parameter?

codex-rs/core/src/mcp_connection_manager.rs
```diff
@@ -425,9 +425,42 @@ struct AsyncManagedClient {
     client: Shared<BoxFuture<'static, Result<ManagedClient, StartupOutcomeError>>>,
     startup_snapshot: Option<Vec<ToolInfo>>,
     startup_complete: Arc<AtomicBool>,
+    startup_request_headers: Arc<StdMutex<Option<reqwest::header::HeaderMap>>>,
     tool_plugin_provenance: Arc<ToolPluginProvenance>,
 }
 
+struct StartupRequestHeadersGuard {
```
pakrym-oai: what is this for?

codex-rs/rmcp-client/src/rmcp_client.rs
```diff
@@ -82,15 +83,37 @@ const JSON_MIME_TYPE: &str = "application/json";
...
+            request = request.header(name, value.clone());
+        }
+    }
+    request
+}
 
 #[derive(Clone)]
 struct StreamableHttpResponseClient {
     inner: reqwest::Client,
+    request_headers_state: Arc<StdMutex<Option<HeaderMap>>>,
```
pakrym-oai, line 115: Are we reusing the same client while mutating headers from various calls? Is there a data race?

pakrym-oai: I.e. if two threads call client in parallel, who's headers win?

nicholasclark-openai: Yeah you're right. Let me figure out that right way to get the session/turn headers in here without resorting to a mutex.

codex-rs/core/src/tasks/mod.rs
```diff
@@ -153,6 +153,8 @@ impl Session {
     ) {
         self.abort_all_tasks(TurnAbortReason::Replaced).await;
         self.clear_connector_selection().await;
+        self.sync_mcp_request_headers_for_turn(turn_context.as_ref())
```
nicholasclark-openai, line 156: The MCP client lifecycle is different than that of a turn and can issues parallel calls, so there's an inherent race condition changing headers inside of a turn. Given turns seem sequential, we set the headers as a turn is starting and unset them once done. This allows to have the headers in the MCP calls without needing blocking concurrency or introducing a race.

chenglong-openai (approved):

PR #15020, opened 2026-03-18
fix: harden plugin feature gating

xl-openai:
**BAD PR**
I messed up the commits in this PR and accidentally merged incorrect changes.


codex-rs/core/src/plugins/manager.rs
```diff
@@ -444,34 +440,29 @@ impl PluginsManager {
...
 
     pub fn plugins_for_layer_stack(
         &self,
-        cwd: &Path,
-        config_layer_stack: &ConfigLayerStack,
+        config: &Config,
         force_reload: bool,
     ) -> PluginLoadOutcome {
-        if !plugins_feature_enabled_from_stack(config_layer_stack) {
+        if !config.features.enabled(Feature::Plugins) {
```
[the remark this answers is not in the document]
xl-openai: that‘s ok.

codex-rs/core/src/thread_manager.rs
```diff
@@ -175,7 +175,10 @@ impl ThreadManager {
             .cloned()
             .unwrap_or_else(|| ModelProviderInfo::create_openai_provider(/*base_url*/ None));
         let (thread_created_tx, _) = broadcast::channel(THREAD_CREATED_CHANNEL_CAPACITY);
-        let plugins_manager = Arc::new(PluginsManager::new(codex_home.clone()));
+        let plugins_manager = Arc::new(PluginsManager::new(
+            codex_home.clone(),
+            config.features.enabled(crate::features::Feature::Plugins),
+        ));
```
sayan-oai: this instance is reused for all threads/sessions over the lifetime of the process. so for app-server, changing config after the process is started wouldn't update the value, and passing thread-level config that changes the value wouldn't be respected.

lmk if im misunderstanding, but this would basically make it static until you restart app-server

xl-openai: Good call rollback to original approach.

codex-rs/core/src/plugins/manager.rs
```diff
@@ -444,49 +440,36 @@ impl PluginsManager {
...
-    }
-
-    pub fn plugins_for_layer_stack(
-        &self,
-        cwd: &Path,
-        config_layer_stack: &ConfigLayerStack,
-        force_reload: bool,
-    ) -> PluginLoadOutcome {
-        if !plugins_feature_enabled_from_stack(config_layer_stack) {
+        if !config.features.enabled(Feature::Plugins) {
```
sayan-oai: this reads cloud requirements, yes?

codex-rs/core/src/plugins/manager.rs
sayan-oai, line 1302: as we discussed offline, this just reads the current enablement state of the plugins directly from config, without considering whether `features.plugins` is enabled. do we not want to gate this as well?

sayan-oai: looked a bit more, `configured_plugins_from_stack` seems like the main way existing functionality loads plugins. gating here seems like it would allow us to avoid gating on many of the higher level functions you added the gating to, as long as those work correctly when no plugins are returned (empty HashMap).

it seems that the only logic that needs to keep its **own** gate is the marketplace-reading logic in `list_marketplaces_for_config`.

im not sure if code-organization-wise pushing down the gate check is better (less random gate checks in higher-level functions, dont have to remember to add those checks yourself) or worse (less clear that those functions will return empty when `features.plugin = false`).

LMK what you think.

codex-rs/core/src/plugins/marketplace.rs
```diff
@@ -238,7 +242,16 @@ fn list_marketplaces_with_home(
...
-        marketplaces.push(load_marketplace(&marketplace_path)?);
+        match load_marketplace(&marketplace_path) {
+            Ok(marketplace) => marketplaces.push(marketplace),
+            Err(err) => {
+                warn!(
+                    path = %marketplace_path,
+                    error = %err,
+                    "skipping marketplace that failed to load"
+                );
+            }
```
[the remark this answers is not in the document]
xl-openai: This is OK.

sayan-oai (approved):
approving to unblock as some of these fixes are urgent, but I think we should think about [this comment](https://github.com/openai/codex/pull/15020#discussion_r2951581691) to see which option is cleaner (it was already hard to debug this because of messy code 🙃).

and i know this is tedious/a nit, but it would be great if we could split this pr based on the 4 points in the description. the changes are all small but high impact on functionality, and I'm worried about rolling back 3 good fixes because they were bundled with one bug.

codex-rs/core/src/plugins/manager.rs
xl-openai: correct.

codex-rs/core/src/plugins/manager.rs
xl-openai: I think configured_plugins_from_stack is a low-level help function and it makes sense to parse the config stack directly (no depends on config). But I do cleanup the configured_plugin_states more to get rid of the plugins_for_config call.

PR #15021, opened 2026-03-18
V8 Bazel Build

cconger:
Alternative approach, we use rusty_v8 for all platforms that its predefined, but lets build from source a musl v8 version with bazel for x86 and aarch64 only.  We would need to release this on github and then use the release.


patches/v8_bazel_rules.patch
```diff
@@ -0,0 +1,226 @@
...
+diff --git a/orig/v8-14.6.202.11/BUILD.bazel b/mod/v8-14.6.202.11/BUILD.bazel
+index 85f31b7..7314584 100644
+--- a/orig/v8-14.6.202.11/BUILD.bazel
++++ b/mod/v8-14.6.202.11/BUILD.bazel
+@@ -303,7 +303,7 @@ v8_int(
+ # If no explicit value for v8_enable_pointer_compression, we set it to 'none'.
+ v8_string(
+     name = "v8_enable_pointer_compression",
+-    default = "none",
++    default = "False",
```
cconger, line 64: As pointed out in other PRs I would like to revisit this.

codex-rs/docs/bazel.md
```diff
@@ -42,6 +42,14 @@ In some cases, an upstream crate may need a patch or a `crate.annotation` in `..
 to have it build in Bazel's sandbox or make it cross-compilation-friendly. If you see issues,
 feel free to ping zbarsky or mbolin.
 
+One current example is `rusty_v8`: musl workspace builds should consume explicit, version-aligned
```
pakrym-oai: nit: not sure this detail is useful for casual bazel user. rusty is a pretty special case even for bazel in this repo.

.github/workflows/rusty-v8-release.yml
```diff
@@ -0,0 +1,199 @@
...
+            "--platforms=@llvm//platforms:${PLATFORM}"
+            "${pair_target}"
+            "${extra_targets[@]}"
+            --build_metadata=REPO_URL=https://github.com/${GITHUB_REPOSITORY}.git
+            --build_metadata=COMMIT_SHA=$(git rev-parse HEAD)
+            --build_metadata=ROLE=CI
+            --build_metadata=VISIBILITY=PUBLIC
+          )
+
+          if [[ -n "${BUILDBUDDY_API_KEY:-}" ]]; then
```
pakrym-oai: pretty sure it's always set

MODULE.bazel
```diff
@@ -132,6 +133,25 @@ crate.annotation(
...
+    ],
+    build_script_env = {
+        "RUSTY_V8_ARCHIVE": "$(execpath //third_party/v8:rusty_v8_archive_for_target)",
+        "RUSTY_V8_SRC_BINDING_PATH": "$(execpath //third_party/v8:rusty_v8_binding_for_target)",
+    },
+    crate = "v8",
+    gen_build_script = "on",
+    patch_args = ["-p1"],
+    patches = [
+        "//patches:v8_custom_archive_out_dir.patch",
```
pakrym-oai: this patch is for rusty, correct? should we name them accordingly?

patches/v8_custom_archive_out_dir.patch
```diff
@@ -0,0 +1,141 @@
+# What: teach the Rust v8 crate build script to consume Bazel-provided prebuilt
+# archive and binding paths, including runfiles path resolution.
+# Scope: build.rs only; affects prebuilt artifact discovery and link setup, not
+# V8 engine source or runtime behavior.
+
+--- a/build.rs
```
pakrym-oai: should this be in your other pr (where we consume the prebuilt)?

cconger: Yes, this is patching for the consumption side I can drop this for now, and rebuild it on the consumption PR.  Was helpful for the E2E test.

MODULE.bazel
cconger: Gonna drop for now, will have the name be disambiguated on the consumption side PR

pakrym-oai (approved):

PR #15035, opened 2026-03-18
fix(apply-patch): handle CRLF gracefully

dylan-hurd-oai:
## Why

`apply_patch` currently parses patches with `str::lines`, which strips carriage returns from CRLF input. That can cause updates to lose the file's existing line endings.

## What changed

- Split patch text on `\n` so CRLF payload lines retain their `\r`.
- Ignore trailing `\r` only when parsing patch syntax markers.
- Add scenario fixtures for update, add, delete-only, move, and multi-chunk CRLF patches.

## Testing

- `just test -p codex-apply-patch`


dylan-hurd-oai:
> My one question is, does this mean that we risk having one line-ending in the patch body and another line ending in the file, and risking a weird result with inconsistent line endings?

The intent of this PR is to eliminate that case! This will allow the model to preserve line endings in the file.

PR #15036, opened 2026-03-18
fix(core) disable command_might_be_dangerous when unsandboxed

dylan-hurd-oai:
## Summary
If we are in a mode that is already explicitly un-sandboxed, then `ApprovalPolicy::Never` should not block dangerous commands.

## Testing
- [x] Existing unit test covers old behavior
- [x] Added a unit test for this new case

### DIFF
--- a/codex-rs/core/src/exec_policy.rs
+++ b/codex-rs/core/src/exec_policy.rs
@@ -553,8 +553,8 @@ pub fn render_decision_for_unmatched_command(
         cfg!(windows) && matches!(sandbox_policy, SandboxPolicy::ReadOnly { .. });
 
     let sandbox_is_explicitly_disabled = matches!(
-        file_system_sandbox_policy.kind,
-        FileSystemSandboxKind::Unrestricted | FileSystemSandboxKind::ExternalSandbox
+        sandbox_policy,
+        SandboxPolicy::DangerFullAccess | SandboxPolicy::ExternalSandbox { .. }
     );
 
     // If the command is flagged as dangerous or we have no sandbox protection,
--- a/codex-rs/core/src/exec_policy.rs
+++ b/codex-rs/core/src/exec_policy.rs
@@ -549,8 +549,13 @@ pub fn render_decision_for_unmatched_command(
 
     // On Windows, ReadOnly sandbox is not a real sandbox, so special-case it
     // here.
-    let runtime_sandbox_provides_safety =
+    let relying_on_runtime_sandbox_for_safety =
         cfg!(windows) && matches!(sandbox_policy, SandboxPolicy::ReadOnly { .. });
+
+    let sandbox_is_explicitly_disabled = matches!(
+        file_system_sandbox_policy.kind,
+        FileSystemSandboxKind::Unrestricted | FileSystemSandboxKind::ExternalSandbox
+    );
 
     // If the command is flagged as dangerous or we have no sandbox protection,
     // we should never allow it to run without approval.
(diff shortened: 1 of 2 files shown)


codex-rs/core/src/exec_policy.rs
```diff
@@ -549,18 +509,30 @@ pub fn render_decision_for_unmatched_command(
...
     // We prefer to prompt the user rather than outright forbid the command,
     // but if the user has explicitly disabled prompts, we must
     // forbid the command.
-    if command_might_be_dangerous(command) || runtime_sandbox_provides_safety {
+    if command_might_be_dangerous(command) || relying_on_runtime_sandbox_for_safety {
         return match approval_policy {
-            AskForApproval::Never => Decision::Forbidden,
+            AskForApproval::Never => {
+                if sandbox_is_explicitly_disabled {
+                    // If the sandbox is explicitly disabled, we should allow the command to run
```
bolinfest, line 569: This could also be because no sandbox is available, like for Windows users who have not installed our sandbox?

dylan-hurd-oai: Ah I might be misunderstanding the intent of `FileSystemSandboxKind`. Should we instead use the following logic?

```
let sandbox_is_explicitly_disabled =  matches!(
    sandbox_policy,
    SandboxPolicy::DangerFullAccess | SandboxPolicy::ExternalSandbox { .. },
)
```

I've updated to switch to this instead

codex-rs/core/src/exec_policy_tests.rs
```diff
@@ -1686,3 +1596,96 @@ async fn verify_approval_requirement_for_unsafe_powershell_command() {
             (unless AskForApproval::Never is specified)."#
     );
 }
+
+struct ExecApprovalRequirementScenario {
```
bolinfest, line 1741: I find this a bit confusing since most of the fields are "inputs" to the assertion whereas `ExecApprovalRequirement` is the only output.

What about dropping `expected_requirement` from the struct and doing:

```rust
async fn assert_exec_approval_requirement_for_command(
  scenario: ExecApprovalRequirementScenario,
  expected_requirement: ExecApprovalRequirement,
) {
```

I think that makes it clearer from the call site what is being asserted.

Or even have `assert_exec_approval_requirement_for_command()` return `ExecApprovalRequirement` and then have an `assert_eq!()` so you can pass a string as the third argument explaining the assertion.

codex-rs/core/src/exec_policy_tests.rs
```diff
@@ -1686,3 +1596,96 @@ async fn verify_approval_requirement_for_unsafe_powershell_command() {
...
+    policy_src: Option<String>,
```
bolinfest, line 1743: Consider commenting to it's immediately obvious:

```suggestion
    /// Source for Starlark .rules file.
    policy_src: Option<String>,
```

codex-rs/core/src/exec_policy_tests.rs
bolinfest, line 1741: nit: I would list the tests before the scaffolding

codex-rs/core/src/exec_policy.rs
```diff
@@ -549,18 +509,30 @@ pub fn render_decision_for_unmatched_command(
 
     // On Windows, ReadOnly sandbox is not a real sandbox, so special-case it
     // here.
-    let runtime_sandbox_provides_safety =
+    let relying_on_runtime_sandbox_for_safety =
         cfg!(windows) && matches!(sandbox_policy, SandboxPolicy::ReadOnly { .. });
 
+    let sandbox_is_explicitly_disabled = matches!(
```
bolinfest: Push this down into `AskForApproval::Never` since that's the only place it's used?

codex-rs/core/src/exec_policy.rs
```diff
@@ -549,18 +509,30 @@ pub fn render_decision_for_unmatched_command(
 
     // On Windows, ReadOnly sandbox is not a real sandbox, so special-case it
     // here.
-    let runtime_sandbox_provides_safety =
+    let relying_on_runtime_sandbox_for_safety =
```
bolinfest: I'm not sure if the old or new variable name is quite right here, or at least I don't feel like it reflects the comment. It feels like it should be `environment_lacks_sandbox_protections`.

bolinfest (approved):

codex-rs/core/src/exec_policy.rs
dylan-hurd-oai: definitely better, updated

codex-rs/core/src/exec_policy_tests.rs
dylan-hurd-oai: done

codex-rs/core/src/exec_policy.rs
dylan-hurd-oai: moved

codex-rs/core/src/exec_policy_tests.rs
dylan-hurd-oai: updated

PR #15041, opened 2026-03-18
feat: support product-scoped plugins.

xl-openai:
1. Added SessionSource::Custom(String) and --session-source.
  2. Enforced plugin and skill products by session_source.
  3. Applied the same filtering to curated background refresh.


sayan-oai (dismissed):
1. `codex-rs/core/src/plugins/manager.rs:666-815` still reconciles every curated remote plugin without checking `policy.products`, so `plugin/list(forceRemoteSync=true)` can reinstall and re-enable ChatGPT-only plugins in an Atlas/Codex session even though install/list/background refresh now hide them.

2. `codex-rs/core/src/plugins/manager.rs:477-501` and `codex-rs/core/src/plugins/manager.rs:1205-1408` intentionally skip product filtering for already-configured plugins, so a wrong-product plugin that is still present in config or cache will continue to load its apps/MCP servers at runtime; that means this PR does not actually enforce plugin scoping by `session_source`, only marketplace admission.

3. Follow-up: `codex-rs/protocol/src/protocol.rs:2310-2324`, `codex-rs/core/src/rollout/mod.rs:7-15`, and `codex-rs/core/src/rollout/list.rs:701-705` make custom session-source resume/history matching case-sensitive. `--session-source ChatGPT` gets ChatGPT product gating, but the resulting thread will not show up in the interactive resume filters because those allowlists are hard-coded to lowercase custom values.

codex-rs/core/src/codex.rs
```diff
@@ -4770,15 +4736,16 @@ mod handlers {
...
+        let cwd = {
             let state = sess.state.lock().await;
-            vec![state.session_configuration.cwd.clone()]
-        } else {
-            cwds
+            state.session_configuration.cwd.clone()
         };
+        if cwds.is_empty() {
+            cwds = vec![cwd];
+        }
```
sayan-oai: why do we need this change? to make `cwds` mut? this also grabs the session state lock regardless of whether `cwds` is empty, which seems mildly worse.

codex-rs/core/src/skills/manager.rs
```diff
@@ -167,8 +186,10 @@ impl SkillsManager {
                     scope: SkillScope::User,
                 }),
         );
-        let outcome = load_skills_from_roots(roots);
-        let outcome = finalize_skill_outcome(outcome, &config_layer_stack);
+        let outcome = crate::skills::filter_skill_load_outcome_for_product(
+            finalize_skill_outcome(load_skills_from_roots(roots), &config_layer_stack),
+            self.restriction_product,
+        );
```
sayan-oai: nit: it seems a bit tedious to have this pattern of calling 3 helpers to get the outcome. you have to remember too much as the user. is it possible to clean up, or make one wrapper?

codex-rs/core/src/plugins/manager.rs
sayan-oai, line 813: do we need to check for product restrictions here too?

sayan-oai, line 1037: this also could go either way, but do we want to enforce product restrictions on `plugin/read`?

codex-rs/protocol/src/protocol.rs
```diff
@@ -2302,13 +2299,31 @@ impl fmt::Display for SessionSource {
...
+        }
+
+        let normalized = trimmed.to_ascii_lowercase();
+        Ok(match normalized.as_str() {
+            "cli" => SessionSource::Cli,
+            "vscode" => SessionSource::VSCode,
+            "exec" => SessionSource::Exec,
+            "mcp" | "appserver" | "app-server" | "app_server" => SessionSource::Mcp,
+            "unknown" => SessionSource::Unknown,
+            _ => SessionSource::Custom(trimmed.to_string()),
```
sayan-oai: do we want to coerce to lowercase? the values we match against in `rollout/mod.rs` is lowercase

codex-rs/core/src/plugins/manager.rs
```diff
@@ -1228,7 +1274,14 @@ fn refresh_curated_plugin_cache(
...
             MarketplacePluginSource::Local { path } => path,
         };
-        plugin_sources.insert(plugin_name, source_path);
+        if plugin.policy.products.is_empty()
+            || restriction_product
+                .is_some_and(|product| product.matches_product_restriction(&plugin.policy.products))
+        {
+            plugin_sources.insert(plugin_name, source_path);
+        } else {
+            product_restricted_plugin_names.insert(plugin_name);
```
sayan-oai: `product_restricted_plugin_names` is a bit confusing IMO; maybe `disallowed_plugin_names_for_product`? because `restriction_product` means `the product for this session` and we are reusing `restricted` in a confusing way.

sayan-oai (commented):
small clarity nit:

maybe for readers it'd be nice to include a doc comment somewhere explaining that for skills:

- if the plugin doesnt match the requested product, its filtered out
- if the plugin matches but some bundled skills dont match, theyre filtered out
- separately, if there are individual skills that dont match, theyre filtered out

you have to bounce around a few files to figure that out right now.

codex-rs/core/src/plugins/manager.rs
xl-openai: Good call add some logic to the actual sync time.

xl-openai: Good call will add PluginNotFound

codex-rs/core/src/plugins/manager.rs
```diff
@@ -1238,6 +1256,15 @@ fn refresh_curated_plugin_cache(
...
 
         let Some(source_path) = plugin_sources.get(&plugin_id.plugin_name).cloned() else {
+            if product_restricted_plugin_names.contains(&plugin_id.plugin_name) {
+                info!(
+                    plugin = plugin_id.plugin_name,
+                    marketplace = OPENAI_CURATED_MARKETPLACE_NAME,
+                    session_source = %session_source,
+                    "skipping curated plugin cache refresh for product-restricted plugin"
+                );
+                continue;
```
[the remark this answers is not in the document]
xl-openai: I think it is OK.

codex-rs/protocol/src/protocol.rs
```diff
@@ -2332,6 +2351,24 @@ impl SessionSource {
...
     }
+    pub fn restriction_product(&self) -> Option<Product> {
+        match self {
+            SessionSource::Custom(source) => Product::from_session_source_name(source),
+            SessionSource::Cli
+            | SessionSource::VSCode
+            | SessionSource::Exec
+            | SessionSource::Mcp
+            | SessionSource::Unknown => Some(Product::Codex),
+            SessionSource::SubAgent(_) => None,
```
sayan-oai, line 2362: confirming, we are ok with subagents not inheriting the restriction product of parent thread? IIUC this means subagents cant use product-scoped plugins like their parents can

sayan-oai (approved):

codex-rs/protocol/src/protocol.rs
xl-openai: it is ok the ThreadManager is only inited once and the product should be set already.

PR #15056, opened 2026-03-18
feat: add graph representation of agent network

jif-oai:
Add a representation of the agent graph. This is now used for:
* Cascade close agents (when I close a parent, it close the kids)
* Cascade resume (oposite)

Later, this will also be used for post-compaction stuffing of the context

Direct fix for: https://github.com/openai/codex/issues/14458


codex-rs/core/src/tools/handlers/multi_agents/close_agent.rs
```diff
@@ -72,7 +72,7 @@ impl ToolHandler for Handler {
             session
                 .services
                 .agent_control
-                .shutdown_agent(agent_id)
+                .close_agent(agent_id)
                 .await
                 .map_err(|err| collab_agent_error(agent_id, err))
                 .map(|_| ())
```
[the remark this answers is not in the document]
jif-oai: this is not correct

codex-rs/core/src/agent/control.rs
```diff
@@ -231,6 +244,84 @@ impl AgentControl {
...
+            return Ok(resumed_thread_id);
+        };
+        let Some(state_db_ctx) = resumed_thread.state_db() else {
+            return Ok(resumed_thread_id);
+        };
+
+        let mut resume_queue = VecDeque::from([(thread_id, root_depth)]);
+        while let Some((parent_thread_id, parent_depth)) = resume_queue.pop_front() {
+            let child_ids = match state_db_ctx
+                .list_thread_spawn_children_with_status(
```
charley-oai, line 262: Could maybe just call `list_thread_spawn_descendants_matching` and then share code with `live_thread_spawn_children` to traverse the tree in level order except a separate sqlite query per level. However I imagine the num levels will normally be small so that's prob not necessary.

Maybe vague benefit would be to share more code btwn the in-memory traversals and the sqlite traversals

The other slightly confusing thing about reviewing this diff is how the names of functions for traversing sqlite-stored graph vs in-memory graph are similar, so it's a bit unclear when we're querying sqlite vs when we're pulling from in-memory state (maybe a footgun for future editors?)

codex-rs/Cargo.lock
```diff
@@ -9367,6 +9368,9 @@ name = "strum"
 version = "0.27.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "af23d6f6c1a224baef9d3f61e287d2761385a5b88fdab4eb4c6f11aeb54c4bcf"
+dependencies = [
+ "strum_macros 0.27.2",
```
charley-oai, line 9372: Not sure if this matters but we will now have 2 different strum_macros versions loaded

- the workspace pins strum = "0.27.2" and strum_macros = "0.28.0" in `codex-rs/Cargo.toml:275`
- enabling strum’s derive feature pulls in the matching strum_macros 0.27.2 transitively

charley-oai (approved):
Looks graphical to me

codex-rs/core/src/agent/control.rs
jif-oai: Not really a simplification because we need the depth during resume while the other is just a BFS
I will try to unify in a follow-up

PR #15067, opened 2026-03-18
Protect first-time project .codex creation across Linux and macOS sandboxes

rreichel3-oai:
## Problem

Codex already treated an existing top-level project `./.codex` directory as protected, but there was a gap on first creation.

If `./.codex` did not exist yet, a turn could create files under it, such as `./.codex/config.toml`, without going through the same approval path as later modifications. That meant the initial write could bypass the intended protection for project-local Codex state.

## What this changes

This PR closes that first-creation gap in the Unix enforcement layers:

- `codex-protocol`
  - treat the top-level project `./.codex` path as a protected carveout even when it does not exist yet
  - avoid injecting the default carveout when the user already has an explicit rule for that exact path
- macOS Seatbelt
  - deny writes to both the exact protected path and anything beneath it, so creating `./.codex` itself is blocked in addition to writes inside it
- Linux bubblewrap
  - preserve the same protected-path behavior for first-time creation under `./.codex`
- tests
  - add protocol regressions for missing `./.codex` and explicit-rule collisions
  - add Unix sandbox coverage for blocking first-time `./.codex` creation
  - tighten Seatbelt policy assertions around excluded subpaths

## Scope

This change is intentionally scoped to protecting the top-level project `.codex` subtree from agent writes.

It does not make `.codex` unreadable, and it does not change the product behavior around loading project skills from `.codex` when project config is untrusted.

## Why this shape

The fix is pointed rather than broad:
- it preserves the current model of “project `.codex` is protected from writes”
- it closes the security-relevant first-write hole
- it avoids folding a larger permissions-model redesign into this PR

## Validation

- `cargo test -p codex-protocol`
- `cargo test -p codex-sandboxing seatbelt`
- `cargo test -p codex-exec --test all sandbox_blocks_first_time_dot_codex_creation -- --nocapture`


codex-rs/sandboxing/src/seatbelt_tests.rs
```diff
@@ -142,8 +150,10 @@ fn explicit_unreadable_paths_are_excluded_from_full_disk_read_and_write_access()
         "expected read carveout parameter in args: {args:#?}"
     );
     assert!(
-        args.iter()
-            .any(|arg| arg == &format!("-DWRITABLE_ROOT_0_RO_0={}", unreadable_root.display())),
+        args.iter().any(|arg| {
+            arg.starts_with("-DWRITABLE_ROOT_0_RO_")
+                && arg.ends_with(&format!("={}", unreadable_root.display()))
+        }),
```
bolinfest: This isn't a huge deal, but while we're here, we could create the suffix once:

```suggestion
    let expected_suffix = format!("={}", unreadable_root.display());
    assert!(
        args.iter().any(|arg| {
            arg.starts_with("-DWRITABLE_ROOT_0_RO_")
                && arg.ends_with(&expected_suffix)
        }),
```

Though taking a step back, the `unreadable_root` is supposed to be associated with `WRITABLE_ROOT_0_RO_`?

codex-rs/sandboxing/src/seatbelt_tests.rs
```diff
@@ -637,31 +654,24 @@ fn create_seatbelt_args_with_read_only_git_and_codex_subpaths() {
     .map(std::string::ToString::to_string)
     .collect();
     let args = create_seatbelt_command_args(shell_command.clone(), &policy, &cwd, false, None);
-
```
bolinfest, line 649: Can we get back to this? It's much clearer to have one `assert_eq!()`.

codex-rs/sandboxing/src/seatbelt_tests.rs
```diff
@@ -142,8 +150,10 @@ fn explicit_unreadable_paths_are_excluded_from_full_disk_read_and_write_access()
         "expected read carveout parameter in args: {args:#?}"
     );
     assert!(
-        args.iter()
-            .any(|arg| arg == &format!("-DWRITABLE_ROOT_0_RO_0={}", unreadable_root.display())),
+        args.iter().any(|arg| {
```
bolinfest: Is there any reason the arg list is unstable? Shouldn't we be able to match `arg` exactly instead of using prefix/suffix checks?

codex-rs/core/src/seatbelt.rs
```diff
@@ -388,6 +388,12 @@ fn build_seatbelt_access_policy(
                 normalize_path_for_sandbox(excluded_subpath.as_path()).unwrap_or(excluded_subpath);
             let excluded_param = format!("{param_prefix}_{index}_RO_{excluded_index}");
```
bolinfest: Hmm, so `_RO_` means "read-only," but are we also using it for "unreadable" subpaths here?

codex-rs/core/src/seatbelt_tests.rs
```diff
@@ -797,6 +812,82 @@ fn create_seatbelt_args_with_read_only_git_and_codex_subpaths() {
...
+        "mkdir -p \"$1\" && echo 'sandbox_mode = \"danger-full-access\"' > \"$2\"",
+        "bash",
+        dot_codex.to_string_lossy().as_ref(),
+        config_toml.to_string_lossy().as_ref(),
+    ]
+    .iter()
+    .map(std::string::ToString::to_string)
+    .collect();
+    let args =
+        create_seatbelt_command_args(shell_command, &policy, repo_root.as_path(), false, None);
```
bolinfest: This is a reminder to myself that we really need to get away from `SandboxPolicy` in favor of `FileSystemPermissions`.

codex-rs/sandboxing/src/seatbelt_tests.rs
```diff
@@ -797,6 +812,82 @@ fn create_seatbelt_args_with_read_only_git_and_codex_subpaths() {
     );
 }
 
+#[test]
+fn create_seatbelt_args_block_first_time_dot_codex_creation() {
```
bolinfest: I really like this test, though I feel it should ideally live in `codex-rs/exec/tests/suite/sandbox.rs` because both our Mac and Linux sandboxing should be enforcing this, right?

codex-rs/protocol/src/permissions.rs
```diff
@@ -1047,19 +1076,46 @@ fn default_read_only_subpaths_for_writable_root(
...
 
-    // Make .agents/skills and .codex/config.toml and related files read-only
-    // to the agent, by default.
-    for subdir in &[".agents", ".codex"] {
-        #[allow(clippy::expect_used)]
-        let top_level_codex = writable_root.join(subdir).expect("valid relative path");
-        if top_level_codex.as_path().is_dir() {
-            subpaths.push(top_level_codex);
-        }
+    #[allow(clippy::expect_used)]
```
bolinfest, line 1122: For these, we should check to see if the user has an existing rule.

Though I'll be honest: I think I need to go back and change `FileSystemPermissions` to use a map like we do in the config file.

codex-rs/protocol/src/permissions.rs
```diff
@@ -278,6 +278,30 @@ impl FileSystemSandboxPolicy {
                     FileSystemPath::Special { .. } => true,
                 }
             });
+
+            let cwd_root = AbsolutePathBuf::from_absolute_path(cwd)
+                .expect("workspace-write cwd should already be absolute");
```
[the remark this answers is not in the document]
rreichel3-oai: This one is obsolete on the current tip. I fixed it in `7796e6069` and added regression coverage in `legacy_workspace_write_projection_accepts_relative_cwd` in `codex-rs/protocol/src/permissions.rs`.

codex-rs/protocol/src/permissions.rs
rreichel3-oai: Updated! 

I think that would be helpful but would prefer to keep this PR focused

codex-rs/sandboxing/src/seatbelt_tests.rs
rreichel3-oai: Moved!

codex-rs/core/src/seatbelt.rs
rreichel3-oai: Renamed

codex-rs/sandboxing/src/seatbelt_tests.rs
rreichel3-oai: done!

codex-rs/sandboxing/src/seatbelt_tests.rs
rreichel3-oai: Yes, fixed. This now matches the generated arg exactly instead of using prefix/suffix checks.

codex-rs/exec/tests/suite/sandbox.rs
```diff
@@ -312,6 +312,78 @@ async fn sandbox_distinguishes_command_and_policy_cwds() {
     assert!(allowed_exists, "allowed path should exist");
 }
 
+#[tokio::test]
+async fn sandbox_blocks_first_time_dot_codex_creation() {
+    core_test_support::skip_if_sandbox!();
+    #[cfg(target_os = "linux")]
+    let sandbox_env = match linux_sandbox_test_env().await {
+        Some(env) => env,
+        None => return,
```
bolinfest, line 322: Is this an error: should it `panic!`?

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -778,6 +778,9 @@ mod tests {
                 "--bind".to_string(),
                 "/".to_string(),
                 "/".to_string(),
+                "--ro-bind".to_string(),
+                "/dev/null".to_string(),
+                "/.codex".to_string(),
```
bolinfest, line 789: This is `.codex` in the filesystem root?

codex-rs/protocol/src/protocol.rs
```diff
@@ -1099,14 +1107,20 @@ fn default_read_only_subpaths_for_writable_root(
...
-        #[allow(clippy::expect_used)]
-        let top_level_codex = writable_root.join(subdir).expect("valid relative path");
-        if top_level_codex.as_path().is_dir() {
-            subpaths.push(top_level_codex);
-        }
+    #[allow(clippy::expect_used)]
+    let top_level_agents = writable_root.join(".agents").expect("valid relative path");
+    if top_level_agents.as_path().is_dir() {
+        subpaths.push(top_level_agents);
+    }
```
bolinfest, line 1116: Can/should we treat `.agents` the same way if it is missing as we do with `.codex`?

codex-rs/sandboxing/src/seatbelt_tests.rs
```diff
@@ -818,6 +848,60 @@ fn create_seatbelt_args_with_read_only_git_and_codex_subpaths() {
...
+        "-c",
+        "mkdir -p \"$1\" && echo 'sandbox_mode = \"danger-full-access\"' > \"$2\"",
+        "bash",
+        dot_codex.to_string_lossy().as_ref(),
+        config_toml.to_string_lossy().as_ref(),
+    ]
+    .iter()
+    .map(std::string::ToString::to_string)
+    .collect();
+    let args = create_seatbelt_command_args_with_extensions(
```
bolinfest, line 885: Should we run the command and verify that it's blocked? Or that's covered by `codex-rs/exec/tests/suite/sandbox.rs`?

codex-rs/sandboxing/src/seatbelt_tests.rs
```diff
@@ -818,6 +848,60 @@ fn create_seatbelt_args_with_read_only_git_and_codex_subpaths() {
...
+        shell_command,
+        &policy,
+        repo_root.as_path(),
+        /*enforce_managed_network*/ false,
+        None,
```
bolinfest: Hmm, why doesn't the linter flag this `None` as needing a comment. Will have to ask Codex. I've seen this other times, too.

codex-rs/linux-sandbox/src/bwrap.rs
bolinfest: Incidentally, I find this hard to read: we should probably have a comment before each group of three to clarify what it's doing?

bolinfest (approved):

codex-rs/sandboxing/src/seatbelt_tests.rs
rreichel3-oai: I think this is just a lint gap rather than an intentional exemption.

I added a comment for clarity.

codex-rs/sandboxing/src/seatbelt_tests.rs
rreichel3-oai: That runtime behavior is covered by `codex-rs/exec/tests/suite/sandbox.rs`.

The Seatbelt test here is intentionally narrower: it only checks the policy shape for denying both the exact `.codex` path and descendants. The end-to-end “command is actually blocked” coverage lives in the exec sandbox test.

codex-rs/protocol/src/protocol.rs
rreichel3-oai: Afaik, we don't exec code directly out of `.agents` the same way MCP's can be spawned in `config.toml`.  The config allows a sandbox-bypass.

codex-rs/linux-sandbox/src/bwrap.rs
rreichel3-oai: Yes. In this test the writable root is `/`, so the default protected subpath under that root shows up as `/.codex` in the generated bwrap args.

Added some comments!

codex-rs/exec/tests/suite/sandbox.rs
rreichel3-oai: I don’t think this should panic.

`None` here means the Linux capability probe determined that Landlock is not actually enforceable on the current host, so this test should skip rather than fail. The helper is being used as a skip gate, not as an assertion that the environment must support Linux sandbox enforcement.

PR #15076, opened 2026-03-18
Add a startup deprecation warning for custom prompts

etraut-openai:
## Summary
- detect custom prompts in `$CODEX_HOME/prompts` during TUI startup
- show a deprecation notice only when prompts are present, with guidance to use `$skill-creator`
- add TUI tests and snapshot coverage for present, missing, and empty prompts directories

## Testing
- Manually tested

### DIFF
--- a/codex-rs/tui/src/snapshots/codex_tui__app__tests__startup_custom_prompt_deprecation_notice.snap
+++ b/codex-rs/tui/src/snapshots/codex_tui__app__tests__startup_custom_prompt_deprecation_notice.snap
@@ -1,0 +1,7 @@
+---
+source: tui/src/app.rs
+expression: rendered
+---
+⚠ Custom prompts are deprecated and will soon be removed.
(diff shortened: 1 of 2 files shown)


codex-rs/tui/src/snapshots/codex_tui__app__tests__startup_custom_prompt_deprecation_notice.snap
```diff
@@ -0,0 +1,7 @@
+---
+source: tui/src/app.rs
+expression: rendered
+---
+⚠ Custom prompts are deprecated and will soon be removed.
+Detected 1 custom prompt in `$CODEX_HOME/prompts`. Use the `$skill-creator` skill to convert each custom prompt into
+a skill.
```
joshka-oai, line 7: What about making this have a directly copy paste-able prompt? (`$skill-creator turn ~/.codex/prompts/foobar.md into a skill`)

joshka-oai (approved):

PR #15087, opened 2026-03-18
Route unified exec process startup through remote exec-server

starr-openai:
## Summary
- add experimental config for a websocket exec-server endpoint plus optional remote workspace root mapping
- route unified exec process/session creation through a remote backend when configured
- wrap remote exec-server sessions behind `UnifiedExecProcess` so `exec_command` can stream output, write stdin, and terminate over the existing protocol
- sketch the filesystem end-state by making `codex_environment::Environment` pluggable and adding a remote exec-server filesystem adapter
- add a narrow smoke test for `codex exec` using a local websocket exec-server as the remote stand-in

## Scope
- this draft now includes both the remote `exec_command` MVP and a filesystem-backend shape sketch so we can review the end state before splitting into smaller PRs
- no project-doc, skills, or app-server command routing changes in this draft
- remote mode fails fast for sandboxed exec and inherited FDs instead of falling back to local

## Validation
- `cargo check -p codex-core --lib`
- `cargo check -p codex-exec-server --lib`
- `cargo check -p codex-app-server --lib`
- `just fmt`
- `just fix -p codex-core -p codex-exec-server -p codex-app-server`
- earlier on the first commit: `cargo test -p codex-core --lib --no-run --message-format=short`
- earlier on the first commit: `cargo test -p codex-exec --test all remote_exec_server -- --nocapture`


codex-rs/core/src/config/mod.rs
```diff
@@ -2768,6 +2797,8 @@ impl Config {
             web_search_mode: constrained_web_search_mode.value,
             web_search_config,
             use_experimental_unified_exec_tool,
+            experimental_unified_exec_exec_server_websocket_url,
```
pakrym-oai, line 2800: This should be `exec_server_url`

PR #15088, opened 2026-03-18
Add Python SDK thread.run convenience methods

shaqayeq-oai:
## TL;DR
Add `thread.run(...)` / `async thread.run(...)` convenience methods to the Python SDK for the common case.

- add `RunInput = Input | str` and `RunResult` with `final_response`, collected `items`, and optional `usage`
- keep `thread.turn(...)` strict and lower-level for streaming, steering, interrupting, and raw generated `Turn` access
- update Python SDK docs, quickstart examples, and tests for the sync and async convenience flows

## Validation
- `python3 -m pytest sdk/python/tests/test_public_api_signatures.py sdk/python/tests/test_public_api_runtime_behavior.py`
- `python3 -m pytest sdk/python/tests/test_real_app_server_integration.py -k 'thread_run_convenience or async_thread_run_convenience'` (skipped in this environment)


sdk/python/src/codex_app_server/api.py
```diff
@@ -87,6 +101,92 @@ def _to_wire_input(input: Input) -> list[JsonObject]:
...
+
+    if completed is None:
+        raise RuntimeError("turn completed event not received")
+
+    _raise_for_failed_turn(completed.turn)
+    return RunResult(
+        final_response=_final_assistant_response_from_items(items),
+        items=items,
+        usage=usage,
+    )
```
owenlin0: IMO it'd be great to introduce some structure so `api.py` doesn't become too monolithic (and also so Codex itself can follow a better file structure in future changes!)

Let's keep `api.py` minimal as the public facade and where the generated `Codex` / `Thread` methods live


Codex suggests this:
> Extract the handwritten convenience logic into private modules:
> - `_inputs.py`: TextInput, ImageInput, LocalImageInput, SkillInput, MentionInput, Input, RunInput, _to_wire_input, _normalize_run_input
> - `_run.py`: RunResult, _collect_run_result, _collect_async_run_result, _raise_for_failed_turn, assistant-text extraction
> - optionally _`turn_handles.py`: TurnHandle and AsyncTurnHandle if you want the bottom half slimmer too
> 

thoughts?

owenlin0 (commented):
API looks super clean in the examples, love to see it!

just a comment about organization, but pre-approving

sdk/python/src/codex_app_server/api.py
```diff
@@ -87,6 +101,92 @@ def _to_wire_input(input: Input) -> list[JsonObject]:
...
+
+
+def _assistant_text_from_item(item: ThreadItem) -> tuple[bool, str]:
+    thread_item = item.root if hasattr(item, "root") else item
+    if isinstance(thread_item, AgentMessageThreadItem):
+        return True, thread_item.text
+    return False, ""
+
+
+def _final_assistant_response_from_items(items: list[ThreadItem]) -> str:
```
owenlin0: actually, I wonder if we should be checking for `agent_message.phase == "finalAnswer"` if it's set

sdk/python/src/codex_app_server/api.py
```diff
@@ -87,6 +101,92 @@ def _to_wire_input(input: Input) -> list[JsonObject]:
...
+    final_response = ""
+    for item in items:
```
owenlin0: just an optimization, we could search from the back since it's pretty much always the last item

owenlin0 (approved):

sdk/python/src/codex_app_server/api.py
owenlin0: non-blocking, i think it's fine as is.

shaqayeq-oai: doing this:

> if an AgentMessageThreadItem has phase == MessagePhase.final_answer, return that text immediately
> otherwise keep the last phase-less assistant message as the compatibility fallback
> ignore explicit commentary for final_response
> So the effective selection order is:
> 
> last final_answer
> else last phase is None
> else ""

thoughts?

owenlin0: sounds good!

owenlin0: Should it optional instead of returning `""`?

sdk/python/src/codex_app_server/api.py
```diff
@@ -65,6 +71,14 @@ class MentionInput:
 
 InputItem = TextInput | ImageInput | LocalImageInput | SkillInput | MentionInput
 Input = list[InputItem] | InputItem
+RunInput = Input | str
+
+
+@dataclass(slots=True)
+class RunResult:
```
owenlin0: Actually, now that I'm looking at this again, should `RunResult` just be a facade over `AppServerTurn`? 

i.e. it mostly just exposes `AppServerTurn`'s fields (including `status`, `error`, and `items`), but also has convenience accessors like `final_response` and `usage`

that way you don't need to write a method that accumulates thread items, you can simply access them via `AppServerTurn`

sdk/python/src/codex_app_server/api.py
shaqayeq-oai: I reverted the `_turn_handles.py` split because it introduced an awkward `api.py` <-> `_turn_handles.py` circular typing dependency around AsyncCodex / AsyncTurnHandle. We can make it work with `TYPE_CHECKING` or a private protocol, but that felt like more complexity than value

sdk/python/src/codex_app_server/api.py
shaqayeq-oai: I think we’d still need to accumulate `item/completed` notifications for `thread.run()`. The current schema says Turn.items is only populated for `thread/resume` / `thread/fork` and otherwise empty [v2_all.py](https://sg.apps.openai.org/github.com/openai/codex/-/blob/sdk/python/src/codex_app_server/generated/v2_all.py?L5204)
[bespoke_event_handling.rs](https://sg.apps.openai.org/github.com/openai/codex/-/blob/codex-rs/app-server/src/bespoke_event_handling.rs?L1892)

sdk/python/src/codex_app_server/api.py
shaqayeq-oai: yea good point

sdk/python/src/codex_app_server/api.py
shaqayeq-oai: but maybe i missed sth lmk and i can fix it in a follow-up pr. merging to reduce churn for @.sdcoffey 's PR

owenlin0: actually you're right, totally forgot we don't populate items on a completed turn lol

PR #15090, opened 2026-03-18
Add exec-server process and filesystem RPCs

starr-openai:
Rebased onto current `main` after #15091 landed separately.

This PR now carries the surviving exec-server process RPC and filesystem RPC implementation on top of the current exec-server stub / websocket-only / environment-abstraction shape on `main`.

It also includes the `rust-ci` sccache guard for `macos-15-xlarge` + `x86_64-apple-darwin` so this branch does not reproduce the same mixed-architecture `ring` failure that hit the earlier stack.


codex-rs/exec-server/src/server/handler.rs
```diff
@@ -42,4 +110,335 @@ impl ExecServerHandler {
...
+            };
+            let seq = process.next_seq;
+            process.next_seq += 1;
+            process.retained_bytes += chunk.len();
+            process.output.push_back(RetainedOutputChunk {
+                seq,
+                stream,
+                chunk: chunk.clone(),
+            });
+            while process.retained_bytes > RETAINED_OUTPUT_BYTES_PER_PROCESS {
```
pakrym-oai, line 450: unified exec needs the tail. we might need the same "drop middle strategy" as used in core.

pakrym-oai (approved):

PR #15095, opened 2026-03-18
[WIP] fix: rollout recorder not adding turns to rollout after an error until next restart

JaviSoto:
I have been chasing this bug for quite some time. I noticed that sometimes when I closed the Codex app, or I tried to resume a CLI thread, that a bunch of messages were missing.

Turns out that this can happen if the rollout recorder hits an error (for example if you’re out of disk space). After that happens, it will stop writing to disk.

I’m not confident about the Codex-written fix, but I’m sharing it as-is at least to highlight the problem and start the conversation!

### DIFF
--- a/codex-rs/core/src/rollout/recorder.rs
+++ b/codex-rs/core/src/rollout/recorder.rs
@@ -5,6 +5,12 @@ use std::fs::{self};
 use std::io::Error as IoError;
 use std::path::Path;
 use std::path::PathBuf;
+#[cfg(test)]
+use std::sync::Arc;
+#[cfg(test)]
+use std::sync::atomic::AtomicUsize;
+#[cfg(test)]
+use std::sync::atomic::Ordering;
 
 use chrono::SecondsFormat;
 use chrono::Utc;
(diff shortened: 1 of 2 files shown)


codex-rs/core/src/rollout/recorder.rs
```diff
@@ -946,6 +960,10 @@ async fn sync_thread_state_after_write(
 
 struct JsonlWriter {
     file: tokio::fs::File,
+    #[cfg(test)]
+    fail_next_writes: Option<Arc<AtomicUsize>>,
+    #[cfg(test)]
+    fail_next_flushes: Option<Arc<AtomicUsize>>,
```
charley-oai: Can we avoid adding test-only fields to a prod type? Or is there a reason we need this for tests?

codex-rs/core/src/rollout/recorder.rs
```diff
@@ -956,7 +974,87 @@ struct RolloutLineRef<'a> {
...
+        }
+    }
+
     async fn write_rollout_item(&mut self, rollout_item: &RolloutItem) -> std::io::Result<()> {
+        self.write_rollout_items(std::slice::from_ref(rollout_item))
+            .await
+    }
+
+    async fn write_rollout_items(&mut self, rollout_items: &[RolloutItem]) -> std::io::Result<()> {
+        #[cfg(test)]
```
charley-oai: This makes the prod code a bit hard to read, do we have to nest test-only code in a prod function like this?

codex-rs/core/src/rollout/recorder.rs
```diff
@@ -956,7 +974,87 @@ struct RolloutLineRef<'a> {
...
+        let file_len_before_write = self.file.metadata().await?.len();
+        let mut json = String::new();
+        for rollout_item in rollout_items {
+            json.push_str(&Self::rollout_line_json(rollout_item)?);
+            json.push('\n');
+        }
+
+        let result = async {
+            self.file.write_all(json.as_bytes()).await?;
+            #[cfg(test)]
```
charley-oai: Same here

PR #15101, opened 2026-03-18
[i18n] Add i18n scaffolding.

mzeng-openai:
- [x] Add i18n scaffolding, initially supporting elicitation messages.


codex-rs/core/src/config/config_tests.rs
```diff
@@ -6155,6 +6160,59 @@ speaker = "Desk Speakers"
     Ok(())
 }
 
+#[test]
+fn locale_loads_from_config_toml() -> std::io::Result<()> {
+    let cfg: ConfigToml = toml::from_str(
+        r#"
+locale = "zh-CN"
```
ja-openai, line 6167: nit: can we reuse those locale tags exactly. It's what we have in Mojito and 1:1 mapping is a lot better/simpler.  `zh-CN -> zh`


```
am ar bg bn bs ca cs da de el es es-419 et fa fi fr fr-CA gu hi hr hu hy id is it ja ka kk kn ko lt lv mk ml mn mr ms my nb nl pa pl pt pt-PT ro ru sk sl so sq sr sv sw ta te th tl tr uk ur vi zh zh-HK zh-Hant 
```

codex-rs/i18n/src/locales/en/approval.ftl
```diff
@@ -0,0 +1,98 @@
+approval-param-action = Action
```
ja-openai, line 17: I realize that in the monorepo/python we only use the fluent library to do message formatting and not to load the strings. 

Mojito does not support `.ftl`, I need to see if it is easy to add, and/or if there are alternative format we can use.

mzeng-openai: per Codex, we can add comments to these strings, and they can be converted to the exact same format as other repos (id, defaultMessage, description)

codex-rs/core/src/config/config_tests.rs
mzeng-openai: Added

codex-rs/i18n/src/locales/en/approval.ftl
ja-openai: Ideally, I want first-party support in Mojito:

- an FTL parser/generator implemented as an Okapi filter (the l10n library Mojito usually relies on), so it can plug in like any other format
- formatted message rendering in the frontend, so linguists can view the final result

We could limit it to a subset of FTL at first, such as single strings with no dependencies on other strings, but full support would be much better.

Started to poke around with Codex...

codex-rs/i18n/src/locales/en/approval.ftl
```diff
@@ -0,0 +1,114 @@
+approval-header = Approve app tool call?
```
ja-openai, line 1: let's just call that directory: `en` , that's the default locale in Mojito

docs/config.md
```diff
@@ -26,6 +26,19 @@ Codex can run a notification hook when the agent finishes a turn. See the config
...
+Set top-level `locale` in `config.toml` to choose the preferred locale for
+localized user-facing strings. For example:
+
+```toml
+locale = "zh"
+```
+
+Codex resolves locales to canonical locale codes, preferring specific locale
+codes first, then generic language codes, and finally `en-US`. For example,
+`zh-HK` falls back to `zh` when there is no `zh-HK` resource bundle.
```
ja-openai, line 40: nit: probably comes from the current files present in the scaffold, but might be a bit misleading for later. we'd want `zh-hk` to fallback to `zh-hant` in the list of Mojito locales not `zh`

ja-openai (approved):
looks good! just one change for the `en` directory (and a tiny comment on locale fallback)

PR #15104, opened 2026-03-18
fix: harden plugin feature gating

xl-openai:
Resubmit https://github.com/openai/codex/pull/15020 with correct content.

1. Use requirement-resolved config.features as the plugin gate.
2. Guard plugin/list, plugin/read, and related flows behind that gate.
3. Skip bad marketplace.json files instead of failing the whole list.
4. Simplify plugin state and caching.


sayan-oai (approved):
stamping as I already approved the original #15020.

PR #15114, opened 2026-03-18
fix: allow restricted filesystem profiles to read helper executables

celia-oai:
## Summary

This PR fixes restricted filesystem permission profiles so Codex's runtime-managed helper executables remain readable without requiring explicit user configuration.

- add implicit readable roots for the configured `zsh` helper path and the main execve wrapper
- allowlist the shared `$CODEX_HOME/tmp/arg0` root when the execve wrapper lives there, so session-specific helper paths keep working
- dedupe injected paths and avoid adding duplicate read entries to the sandbox policy
- add regression coverage for restricted read mode with helper executable overrides

## Testing 
before this change: got this error when executing a shell command via zsh fork:
```
"sandbox error: sandbox denied exec error, exit code: 127, stdout: , stderr: /etc/zprofile:11: operation not permitted: /usr/libexec/path_helper\nzsh:1: operation not permitted: .codex/skills/proxy-a/scripts/fetch_example.sh\n"
```

saw this change went away after this change, meaning the readable roots and injected correctly.

### DIFF
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -97,6 +97,7 @@ use std::path::Path;
 use std::path::PathBuf;
 
 use crate::config::permissions::compile_permission_profile;
+use crate::config::permissions::get_readable_roots_required_for_codex_runtime;
 use crate::config::permissions::network_proxy_config_from_profile_network;
 use crate::config::profile::ConfigProfile;
 use codex_network_proxy::NetworkProxyConfig;
@@ -1917,91 +1918,6 @@ fn resolve_permission_config_syntax(
     })
 }
 
-fn add_additional_file_system_writes(
-    file_system_sandbox_policy: &mut FileSystemSandboxPolicy,
-    additional_writable_roots: &[AbsolutePathBuf],
-) {
-    for path in additional_writable_roots {
-        let exists = file_system_sandbox_policy.entries.iter().any(|entry| {
-            matches!(
-                &entry.path,
-                codex_protocol::permissions::FileSystemPath::Path { path: existing }
-                    if existing == path && entry.access == codex_protocol::permissions::FileSystemAccessMode::Write
-            )
-        });
-        if !exists {
-            file_system_sandbox_policy.entries.push(
-                codex_protocol::permissions::FileSystemSandboxEntry {
-                    path: codex_protocol::permissions::FileSystemPath::Path { path: path.clone() },
-                    access: codex_protocol::permissions::FileSystemAccessMode::Write,
-                },
-            );
-        }
-    }
-}
-
-/// Returns implicit readable roots for runtime-managed helper executables that
-/// must stay usable under restricted filesystem policies without requiring
-/// explicit user permission-profile entries.
-///
-/// When the execve wrapper lives under `$CODEX_HOME/tmp/arg0`, we allowlist the
-/// shared arg0 temp root instead of a single session-specific wrapper path so
-/// future helper paths created in that tree remain readable too.
-fn helper_readable_roots(
-    codex_home: &Path,
-    zsh_path: Option<&PathBuf>,
-    main_execve_wrapper_exe: Option<&PathBuf>,
-) -> Vec<AbsolutePathBuf> {
-    let arg0_root = AbsolutePathBuf::from_absolute_path(codex_home.join("tmp").join("arg0")).ok();
-    let zsh_path = zsh_path.and_then(|path| AbsolutePathBuf::from_absolute_path(path).ok());
-    let execve_wrapper_root = main_execve_wrapper_exe.and_then(|path| {
-        let path = AbsolutePathBuf::from_absolute_path(path).ok()?;
-        if let Some(arg0_root) = arg0_root.as_ref()
-            && path.as_path().starts_with(arg0_root.as_path())
-        {
-            Some(arg0_root.clone())
-        } else {
-            Some(path)
-        }
-    });
-
-    let mut readable_roots = Vec::new();
-    if let Some(zsh_path) = zsh_path {
-        readable_roots.push(zsh_path);
-    }
-    if let Some(execve_wrapper_root) = execve_wrapper_root {
-        readable_roots.push(execve_wrapper_root);
-    }
-    readable_roots
-}
-
-fn add_additional_file_system_reads(
-    file_system_sandbox_policy: &mut FileSystemSandboxPolicy,
-    additional_readable_roots: &[AbsolutePathBuf],
-) {
-    if file_system_sandbox_policy.has_full_disk_read_access() {
-        return;
-    }
-
-    for path in additional_readable_roots {
-        let exists = file_system_sandbox_policy.entries.iter().any(|entry| {
-            matches!(
-                &entry.path,
-                codex_protocol::permissions::FileSystemPath::Path { path: existing }
-                    if existing == path && entry.access == codex_protocol::permissions::FileSystemAccessMode::Read
-            )
-        });
-        if !exists {
-            file_system_sandbox_policy.entries.push(
-                codex_protocol::permissions::FileSystemSandboxEntry {
-                    path: codex_protocol::permissions::FileSystemPath::Path { path: path.clone() },
-                    access: codex_protocol::permissions::FileSystemAccessMode::Read,
-                },
-            );
-        }
-    }
-}
-
 /// Optional overrides for user configuration (e.g., from CLI flags).
 #[derive(Default, Debug, Clone)]
 pub struct ConfigOverrides {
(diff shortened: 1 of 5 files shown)


codex-rs/core/src/config/mod.rs
```diff
@@ -1927,6 +1897,75 @@ fn add_additional_file_system_writes(
     }
 }
 
+fn dedup_absolute_paths(paths: Vec<AbsolutePathBuf>) -> Vec<AbsolutePathBuf> {
```
bolinfest: Should we use https://docs.rs/indexmap/latest/indexmap/set/struct.IndexSet.html for this instead so insertion order is preserved and we know everything is unique by construction?

codex-rs/core/src/config/mod.rs
```diff
@@ -1927,6 +1897,75 @@ fn add_additional_file_system_writes(
...
+/// Returns implicit readable roots for runtime-managed helper executables that
+/// must stay usable under restricted filesystem policies without requiring
+/// explicit user permission-profile entries.
+///
+/// When the execve wrapper lives under `$CODEX_HOME/tmp/arg0`, we allowlist the
+/// shared arg0 temp root instead of a single session-specific wrapper path so
+/// future helper paths created in that tree remain readable too.
+fn helper_readable_roots(
+    codex_home: &Path,
+    zsh_path: Option<&PathBuf>,
```
bolinfest: Hmm, this seems like my mistake: `zsh_path` is an `AbsolutePathBuf` on `ConfigToml` but not `Config`?

codex-rs/core/src/config/mod.rs
```diff
@@ -1927,6 +1897,75 @@ fn add_additional_file_system_writes(
...
+    let arg0_root = AbsolutePathBuf::from_absolute_path(codex_home.join("tmp").join("arg0")).ok();
+    let zsh_path = zsh_path.and_then(|path| AbsolutePathBuf::from_absolute_path(path).ok());
+    let execve_wrapper_root = main_execve_wrapper_exe.and_then(|path| {
+        let path = AbsolutePathBuf::from_absolute_path(path).ok()?;
+        if let Some(arg0_root) = arg0_root.as_ref()
+            && path.as_path().starts_with(arg0_root.as_path())
+        {
+            return Some(arg0_root.clone());
+        }
+        Some(path)
```
bolinfest:
```suggestion
        {
            Some(arg0_root.clone())
        } else {
            Some(path)
        }
```

codex-rs/core/src/config/mod.rs
```diff
@@ -1927,6 +1897,75 @@ fn add_additional_file_system_writes(
...
+    });
+
+    dedup_absolute_paths(
```
bolinfest: These are just two paths and then can never be the same, can they?

codex-rs/core/src/config/mod.rs
```diff
@@ -1927,6 +1897,75 @@ fn add_additional_file_system_writes(
...
+fn add_additional_file_system_reads(
+    file_system_sandbox_policy: &mut FileSystemSandboxPolicy,
+    additional_readable_roots: &[AbsolutePathBuf],
+) {
+    if file_system_sandbox_policy.has_full_disk_read_access() {
+        return;
+    }
+
+    for path in additional_readable_roots {
+        let exists = file_system_sandbox_policy.entries.iter().any(|entry| {
```
bolinfest: Why iterate entries instead of doing a lookup?

bolinfest (commented):
I flagged a number of big-O issues in this PR. Let's take a step back and think about this design some more.

codex-rs/core/src/config/mod.rs
celia-oai: ah good point. I guess this seems more preventative, removing for now as it's not needed

codex-rs/core/src/config/mod.rs
celia-oai: I guess this was pattern-matched from `add_additional_file_system_writes`, will fix both

celia-oai: oh i guess it doesn't matter cus file_system_sandbox_policy.entries is `Vec<FileSystemSandboxEntry>` type. should I change it to a hashset?

codex-rs/core/src/config/mod.rs
celia-oai: this is where the type became PathBuf: https://github.com/openai/codex/blob/dev/cc/fix-3/codex-rs/core/src/config/mod.rs#L2589. Happy to refactor in a followup so that everything is `AbsolutePathBuf` if needed

bolinfest: Thanks, yes, please do that in a follow-up!

codex-rs/core/src/config/mod.rs
```diff
@@ -1927,6 +1897,68 @@ fn add_additional_file_system_writes(
...
 }
 
+/// Returns implicit readable roots for runtime-managed helper executables that
+/// must stay usable under restricted filesystem policies without requiring
+/// explicit user permission-profile entries.
+///
+/// When the execve wrapper lives under `$CODEX_HOME/tmp/arg0`, we allowlist the
+/// shared arg0 temp root instead of a single session-specific wrapper path so
+/// future helper paths created in that tree remain readable too.
+fn helper_readable_roots(
```
bolinfest: I would favor a longer name that is explicit because all of this permissions stuff can be a bit confusing. Maybe:

```rust
/// Returns a list of paths that must be readable by shell tools in order
/// for Codex to function. These should always be added to the
/// `FileSystemSandboxPolicy` for a thread. 
fn get_readable_roots_required_for_codex_runtime(
```

codex-rs/core/src/config/mod.rs
```diff
@@ -1927,6 +1897,68 @@ fn add_additional_file_system_writes(
...
+    codex_home: &Path,
+    zsh_path: Option<&PathBuf>,
+    main_execve_wrapper_exe: Option<&PathBuf>,
+) -> Vec<AbsolutePathBuf> {
+    let arg0_root = AbsolutePathBuf::from_absolute_path(codex_home.join("tmp").join("arg0")).ok();
```
bolinfest: Over in `codex-rs/arg0/src/lib.rs`, we update the `PATH` env var to include the _specific_ subfolder of `~/.codex/tmp/arg0` to include. Today, that includes `apply_patch` in addition to `codex-execve-wrapper`. I realize it requires some extra plumbing (and maybe merits an extra field in `Arg0DispatchPaths`), but it feels like we should be adding the specific folder (e.g., `~/.codex/tmp/arg0/codex-arg0gJ8bFl`) rather than just `main_execve_wrapper_exe` to the required readable roots?

codex-rs/core/src/config/mod.rs
```diff
@@ -1927,6 +1897,68 @@ fn add_additional_file_system_writes(
     }
 }
 
+/// Returns implicit readable roots for runtime-managed helper executables that
```
bolinfest: `codex-rs/core/src/config/mod.rs` is quite large: is there another file where this new code could live instead?

codex-rs/core/src/config/mod.rs
```diff
@@ -2643,12 +2679,17 @@ impl Config {
                     &resolved_cwd,
                 )
             };
+        add_additional_file_system_reads(
```
bolinfest: I'm not sure this is the right place to do this.

`FileSystemSandboxPolicy::get_readable_roots_with_cwd()` seems more appropriate?

codex-rs/core/src/config/mod.rs
celia-oai: will move to codex-rs/core/src/config/permissions.rs

codex-rs/core/src/config/permissions.rs
```diff
@@ -190,6 +190,37 @@ pub(crate) fn compile_permission_profile(
...
+/// for Codex to function. These should always be added to the
+/// `FileSystemSandboxPolicy` for a thread.
+pub(crate) fn get_readable_roots_required_for_codex_runtime(
+    codex_home: &Path,
+    zsh_path: Option<&PathBuf>,
+    main_execve_wrapper_exe: Option<&PathBuf>,
+) -> Vec<AbsolutePathBuf> {
+    let arg0_root = AbsolutePathBuf::from_absolute_path(codex_home.join("tmp").join("arg0")).ok();
+    let zsh_path = zsh_path.and_then(|path| AbsolutePathBuf::from_absolute_path(path).ok());
+    let execve_wrapper_root = main_execve_wrapper_exe.and_then(|path| {
```
bolinfest, line 203: Ah, I see, so we are relying on getting the parent folder of `codex-execve-wrapper` to find the right folder of `~/.codex/tmp/arg0`. This certainly works for now, but in a follow-up, I would prefer we do something more robust!

Let's add a TODO and then in a subsequent PR, we should:

- update `Arg0DispatchPaths` to include the path entry added by `prepend_path_entry_for_codex_aliases()`
- thread all that through to here so that we can add that entry to the return value instead of relying on the parent of `main_execve_wrapper_exe`

Also, I see these comments and I think they are out of date:

https://github.com/openai/codex/blob/3aaee0f0612d4f7c03fbce02f13094d517a58605/codex-rs/core/src/config/mod.rs#L452-L463

That is, I do not think we should be relying on `ConfigOverrides` to thread the values of `Arg0DispatchPaths` through. Ideally we would try to clean that up as part of the TODO, as well.

codex-rs/core/src/config/permissions.rs
```diff
@@ -190,6 +190,37 @@ pub(crate) fn compile_permission_profile(
...
+        if let Some(arg0_root) = arg0_root.as_ref()
+            && path.as_path().starts_with(arg0_root.as_path())
+        {
+            path.parent()
+        } else {
+            Some(path)
+        }
+    });
+
+    let mut readable_roots = Vec::new();
```
bolinfest, line 214: protip (not worth waiting for CI for this one small change, tho)

```suggestion
    let mut readable_roots = Vec::with_capacity(2);
```

bolinfest (approved):

PR #15120, opened 2026-03-19
chore: refactor network permissions to use explicit domain and unix socket rule maps

celia-oai:
## Summary

This PR replaces the legacy network allow/deny list model with explicit rule maps for domains and unix sockets across managed requirements, permissions profiles, the network proxy config, and the app server protocol.

Concretely, it:

- introduces typed domain (`allow` / `deny`) and unix socket permission (`allow` / `none`) entries instead of separate `allowed_domains`, `denied_domains`, and `allow_unix_sockets` lists
- updates config loading, managed requirements merging, and exec-policy overlays to read and upsert rule entries consistently
- exposes the new shape through protocol/schema outputs, debug surfaces, and app-server config APIs
- rejects the legacy list-based keys and updates docs/tests to reflect the new config format

## Why

The previous representation split related network policy across multiple parallel lists, which made merging and overriding rules harder to reason about. Moving to explicit keyed permission maps gives us a single source of truth per host/socket entry, makes allow/deny precedence clearer, and gives protocol consumers access to the full rule state instead of derived projections only.

### Backward compatible

- Managed requirements still accept the legacy `experimental_network.allowed_domains`, `experimental_network.denied_domains`, and `experimental_network.allow_unix_sockets` fields. They are normalized into the new canonical `domains` and `unix_sockets` maps internally.
- App-server v2 still deserializes legacy `allowedDomains`, `deniedDomains`, and `allowUnixSockets` payloads, so older clients can continue reading managed network requirements.
- App-server v2 responses still populate `allowedDomains`, `deniedDomains`, and `allowUnixSockets` as legacy compatibility views derived from the canonical maps.
- `managed_allowed_domains_only` keeps the same behavior after normalization. Legacy managed allowlists still participate in the same enforcement path as canonical `domains` entries.

### Not backward compatible

- Permissions profiles under `[permissions.<profile>.network]` no longer accept the legacy list-based keys. Those configs must use the canonical `[domains]` and `[unix_sockets]` tables instead of `allowed_domains`, `denied_domains`, or `allow_unix_sockets`.
- Managed `experimental_network` config cannot mix canonical and legacy forms in the same block. For example, `domains` cannot be combined with `allowed_domains` or `denied_domains`, and `unix_sockets` cannot be combined with `allow_unix_sockets`.
- The canonical format can express explicit `"none"` entries for unix sockets, but those entries do not round-trip through the legacy compatibility fields because the legacy fields only represent allow/deny lists.
## Testing
`/target/debug/codex sandbox macos --log-denials /bin/zsh -c 'curl https://www.example.com' ` gives 200 with config 
```
[permissions.workspace.network.domains]
"www.example.com" = "allow"
```
and fails when set to deny: `curl: (56) CONNECT tunnel failed, response 403`.

Also tested backward compatibility path by verifying that adding the following to `/etc/codex/requirements.toml` works:
```
[experimental_network]
allowed_domains = ["www.example.com"]
```


codex-rs/config/src/config_requirements.rs
```diff
@@ -132,41 +132,137 @@ pub struct McpServerRequirement {
...
+        let permission = match self {
+            Self::Allow => "allow",
+            Self::None => "none",
+        };
+        f.write_str(permission)
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq, Eq)]
 pub struct NetworkRequirementsToml {
```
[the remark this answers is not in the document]
celia-oai: this is fine

codex-rs/core/src/config/permissions.rs
```diff
@@ -114,14 +211,11 @@ impl NetworkToml {
...
-        }
-        if let Some(denied_domains) = self.denied_domains.as_ref() {
-            config.network.denied_domains = denied_domains.clone();
+        if let Some(domains) = self.domains.as_ref() {
+            overlay_network_domain_permissions(config, domains);
         }
-        if let Some(allow_unix_sockets) = self.allow_unix_sockets.as_ref() {
-            config.network.allow_unix_sockets = allow_unix_sockets.clone();
+        if self.unix_sockets.is_some() {
+            config.network.unix_sockets = self.unix_sockets.as_ref().map(to_proxy_unix_sockets);
```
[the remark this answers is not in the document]
celia-oai: hmm, should I just get rid of `none`?

celia-oai:
@codex review

codex-rs/core/src/config/permissions.rs
```diff
@@ -56,6 +59,98 @@ pub enum FilesystemPermissionToml {
...
+            Self::Allow => "allow",
+            Self::None => "none",
+        };
+        f.write_str(permission)
+    }
+}
+
 #[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq, Eq, JsonSchema)]
 #[schemars(deny_unknown_fields)]
 pub struct NetworkToml {
```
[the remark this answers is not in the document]
celia-oai: I think this is fine since it's an experimental feature

codex-rs/app-server/src/config_api.rs
```diff
@@ -410,20 +412,78 @@ fn map_residency_requirement_to_api(
...
+        .as_ref()
+        .map(codex_core::config_loader::NetworkUnixSocketPermissionsToml::allow_unix_sockets)
+        .filter(|entries| !entries.is_empty());
+
     NetworkRequirements {
         enabled: network.enabled,
         http_port: network.http_port,
         socks_port: network.socks_port,
         allow_upstream_proxy: network.allow_upstream_proxy,
         dangerously_allow_non_loopback_proxy: network.dangerously_allow_non_loopback_proxy,
```
viyatb-oai, line 434: can you make sure `managed_allowed_domains_only` is propagated and respected here? I think I missed adding it here...

codex-rs/config/src/config_requirements.rs
```diff
@@ -132,41 +133,237 @@ pub struct McpServerRequirement {
...
+fn legacy_unix_socket_permissions_from_list(
+    allow_unix_sockets: Option<Vec<String>>,
+) -> Option<NetworkUnixSocketPermissionsToml> {
+    let entries = allow_unix_sockets
+        .unwrap_or_default()
+        .into_iter()
+        .map(|path| (path, NetworkUnixSocketPermissionToml::Allow))
+        .collect::<BTreeMap<_, _>>();
+
+    (!entries.is_empty()).then_some(NetworkUnixSocketPermissionsToml { entries })
```
viyatb-oai, line 340: [P2] Preserve explicit empty legacy network lists

These helpers collapse explicit legacy `allowed_domains = []` / `allow_unix_sockets = []` to `None`, which is not equivalent to the old structs. In `danger-full-access`, `allowed_domains = []` used to clear and pin the allowlist to empty, but now it is treated as absent and leaves user allowlist entries intact. Likewise, `allow_unix_sockets = []` used to keep `dangerously_allow_all_unix_sockets` constrained off via `constraints.allow_unix_sockets = Some([])`, but that constraint now disappears.

viyatb-oai (approved):

codex-rs/config/src/config_requirements.rs
celia-oai: there is unfortunately some lossiness expected after migrating to the new schema. I added a docstring here to explain this: https://github.com/openai/codex/pull/15120/changes#diff-6dea1db0d0ce5a6f3b19ccbc66cb916ad15d3bbc8ce347cff9a934445f9ce56eR311-R313

celia-oai: I feel the better thing to do here is to get rid of `dangerously_allow_all_unix_sockets` down the road, and restrict `danger-full-access` behavior?

viyatb-oai: let me take a stab at seeing if we can preserve the behavior

viyatb-oai: you can land this though and we can do a fast follow

PR #15121, opened 2026-03-19
Add experimental network delegation transport

sdcoffey:
**note to reviewers:** Most of this PR is generated types

## Summary
- add the experimental app-server protocol surface for delegated model requests, stream metadata, request failures, and unary compaction
- route delegated model execution through app-server into core, including persisted thread-level `networkDelegation` config and delegated remote compaction
- teach the Python SDK to answer `model/request` / `model/compact` and send delegated stream notifications back to Codex

## Testing
- cargo test -p codex-app-server-protocol
- cargo test -p codex-app-server delegated_model::tests::delegated_compact_request_round_trips_output
- cargo test -p codex-core client::tests::delegated_compact_conversation_history_routes_through_transport


codex-rs/app-server-protocol/schema/json/ModelRequestParams.json
```diff
@@ -0,0 +1,998 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "definitions": {
```
pakrym-oai, line 3: I think we should make request opaque and not pull in all these types.

codex-rs/app-server-protocol/schema/json/ModelCompactParams.json
```diff
@@ -0,0 +1,967 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "definitions": {
+    "ContentItem": {
```
pakrym-oai, line 4: I think we should make request opaque and not pull in all these types.

codex-rs/app-server-protocol/schema/json/ClientRequest.json
```diff
@@ -1243,6 +1243,32 @@
       ],
       "type": "string"
     },
+    "NetworkDelegationConfig": {
+      "properties": {
+        "mode": {
+          "$ref": "#/definitions/NetworkDelegationMode"
+        },
+        "streamIdleTimeoutMs": {
```
pakrym-oai, line 1251: Do we care? The app-server client can handle this.

codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -555,7 +555,16 @@ macro_rules! server_request_definitions {
...
-        #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+        #[derive(
+            Serialize,
+            Deserialize,
+            Debug,
+            Clone,
+            PartialEq,
+            JsonSchema,
+            TS,
+            ExperimentalApi,
```
pakrym-oai, line 566: intentional?

codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -701,7 +710,16 @@ macro_rules! client_notification_definitions {
...
-        #[derive(Serialize, Deserialize, Debug, Clone, JsonSchema, TS, Display)]
+        #[derive(
+            Serialize,
+            Deserialize,
+            Debug,
+            Clone,
+            JsonSchema,
+            TS,
+            Display,
+            ExperimentalApi,
```
pakrym-oai, line 721: intentional?

codex-rs/app-server/tests/suite/v2/command_exec.rs
```diff
@@ -744,11 +744,14 @@ async fn command_exec_process_ids_are_connection_scoped_and_disconnect_terminate
     )
     .await?;
 
-    let delta = read_command_exec_delta_ws(&mut ws1).await?;
-    assert_eq!(delta.process_id, "shared-process");
-    assert_eq!(delta.stream, CommandExecOutputStream::Stdout);
-    let delta_text = String::from_utf8(STANDARD.decode(&delta.delta_base64)?)?;
-    assert!(delta_text.contains("ready"));
+    let stdout = read_command_exec_output_ws_until_contains(
```
pakrym-oai, line 747: unrelated?

codex-rs/core/src/delegated_model_transport.rs
```diff
@@ -0,0 +1,46 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub enum DelegatedModelEvent {
+    StreamMetadata(HashMap<String, String>),
+    StreamEvent(Value),
+    RequestFailed(ModelRequestError),
+}
+
+#[async_trait]
+pub trait DelegatedModelTransport: Send + Sync {
+    async fn start_model_request(
```
pakrym-oai, line 37: this method shape feels a bit hard to use.

Might be be better to have Result<ModelResponse, ModelRequestError>

Where ModelResponse has `mpsc::Receiver<DelegatedModelEvent>` and metadata.

While error returns an error.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2021,6 +2033,15 @@ impl CodexMessageProcessor {
                     session_configured,
                     ..
                 } = new_conv;
+                if let Err(error) =
+                    Self::set_network_delegation(thread.as_ref(), network_delegation).await
```
pakrym-oai, line 2037: I'd love if we pushed this setting much deeper (all the way to app server startup config?). I don't think there is a point to support changing it per thread.

pakrym-oai (commented):
PR needs many more integration tests (search for `test_codex()`)

codex-rs/core/src/client.rs
```diff
@@ -382,6 +439,55 @@ impl ModelClient {
             None
         };
         let text = create_text_param_for_request(verbosity, &prompt.output_schema);
+        let mut extra_headers = self.build_subagent_headers();
+        extra_headers.extend(build_conversation_headers(Some(
+            self.state.conversation_id.to_string(),
+        )));
+
+        if self
```
pakrym-oai, line 447: I don't think we should fork logic here. Is there a lower layer we can plug implementation into? Transport? Or making ApiCompactClient a trait?

The point is to keep exactly the same request building and sending code and only fork inside the transport implementation

codex-rs/core/src/client.rs
```diff
@@ -821,6 +927,115 @@ impl ModelClientSession {
...
+            stream: request.stream,
+            include: request.include,
+            service_tier: request.service_tier,
+            prompt_cache_key: request.prompt_cache_key,
+            text: request.text.map(serde_json::to_value).transpose()?,
+            request_headers: Some(request_headers),
+        })
+    }
+
+    async fn stream_delegated(
```
pakrym-oai, line 994: same comment as above.

delegate request ~~~=== normal http flow in terms of request and consuming events. The only different is that events don't come via http.

codex-rs/core/src/client.rs
```diff
@@ -142,6 +164,8 @@ struct ModelClientState {
     beta_features_header: Option<String>,
     disable_websockets: AtomicBool,
     cached_websocket_session: StdMutex<WebsocketSession>,
+    delegated_model_transport: DelegatedModelTransportSlot,
```
pakrym-oai, line 167: Wonder if we can push `delegated` flag into ModelProviderInfo. That's where existing transport switch is coming from.

codex-rs/app-server-protocol/schema/json/ClientRequest.json
sdcoffey: will remove

PR #15127, opened 2026-03-19
[Codex][Codex Auth] Handle stale managed auth refresh races before relogin

ccy-oai:
CODEX-5156

###### Summary
- Managed ChatGPT auth now reloads compatible local auth before spending a stale cached refresh token on the normal `auth()` path.
- A `refresh_token_reused` failure now does one final guarded reload before forcing relogin.
- Guarded reload stays pinned to compatible `account_id` and `chatgpt_user_id` when an account boundary is known.
- Legacy auth without `account_id` keeps the old compatibility path instead of being stranded on stale credentials.

###### Rationale (from spec findings)
- The 401 incident spec captured a concrete stale-refresh split-brain cohort: one local client advanced auth while another local session failed with `refresh_token_reused`.
- The old proactive stale path went straight to the token authority and could spend an already-rotated refresh lineage.
- Same-store local auth advancement should not force relogin when newer compatible auth is already available on disk.
- Recovery still has to reject cross-user adoption when another local login changed the user identity inside the same workspace/account.

###### Scope
- Changes `codex-rs/login/src/auth/manager.rs`, `codex-rs/login/src/auth/auth_tests.rs`, and the two cloud-requirements expectations in `codex-rs/cloud-requirements/src/lib.rs`.
- Leaves storage/schema changes, cross-process locking, app-hosted `chatgptAuthTokens`, and cross-device auth divergence unchanged.

###### Trade-offs
- Recovery is limited to same-store local auth changes; it does not solve cases where the newer auth exists only on another machine or another store.
- When cached auth already has `account_id`, reload adoption now requires compatible `account_id` and `chatgpt_user_id` instead of accepting the account boundary alone.
- Unauthorized recovery now evaluates guarded reloads against the auth snapshot that triggered the 401, not mutable cache state.

###### Client follow-up
- Add observability for retained `refresh_token_reused` volume after the same-store race fix.
- Tackle login/bootstrap failures such as `token_exchange_failed` separately.
- Re-evaluate whether any later local refresh serialization is still needed after this resilience pass.

###### Testing
- `cargo test -p codex-login stale_proactive_refresh_uses_newer_local_auth_before_spending_cached_refresh_token -- --nocapture`
- `cargo test -p codex-login stale_proactive_refresh_does_not_adopt_a_different_user_in_same_account -- --nocapture`
- `cargo test -p codex-login stale_proactive_refresh_without_account_id_adopts_auth_with_account_id -- --nocapture`
- `cargo test -p codex-login stale_proactive_refresh_without_identity_claims_still_attempts_refresh -- --nocapture`
- `cargo test -p codex-login unauthorized_recovery_reload_uses_original_auth_snapshot_for_same_account_user_guard -- --nocapture`
- `cargo test -p codex-login refresh_token_reused_recovers_if_guarded_reload_finds_newer_auth -- --nocapture`
- `cargo test -p codex-login refresh_token_reused_without_account_id_adopts_auth_when_disk_auth_gains_account_id -- --nocapture`
- `cargo test -p codex-login --no-run --message-format short`
- `cargo test -p codex-cloud-requirements fetch_cloud_requirements_uses_reloaded_auth_before_unauthorized_retry -- --nocapture`
- `cargo test -p codex-cloud-requirements fetch_cloud_requirements_rejects_different_user_in_same_account -- --nocapture`
- `cargo test -p codex-cloud-requirements --no-run --message-format short`
- `just fmt`
- Manual built-CLI validation in isolated `CODEX_HOME=/tmp/codex-auth-resilience-smoke` using `/Users/ccy/code/codex/codex-rs/target/debug/codex`.
- Manual case 1: started a long-lived session, rewrote `auth.json` stale, restored newer same-user auth on disk, then prompted again in the same session; it succeeded without relogin.
- Manual case 2: repeated the flow with cached auth missing `tokens.account_id`, then restored current same-user auth on disk; the same session still succeeded without relogin.

### DIFF
--- a/codex-rs/login/src/auth/manager.rs
+++ b/codex-rs/login/src/auth/manager.rs
@@ -825,8 +825,10 @@ enum ReloadOutcome {
     ReloadedChanged,
     /// Reload was performed and the cached auth remained the same
     ReloadedNoChange,
-    /// Reload was skipped (missing or mismatched account id)
-    Skipped,
+    /// Reload was skipped because the on-disk auth is for a different account/workspace boundary.
+    SkippedDifferentIdentity,
+    /// Reload was skipped because the cached auth lacks enough identity to guard the reload.
+    SkippedUnknownIdentity,
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
@@ -852,6 +854,7 @@ pub struct UnauthorizedRecovery {
     manager: Arc<AuthManager>,
     step: UnauthorizedRecoveryStep,
     expected_account_id: Option<String>,
+    recovery_auth: Option<CodexAuth>,
     mode: UnauthorizedRecoveryMode,
 }
 
@@ -886,6 +889,7 @@ impl UnauthorizedRecovery {
             manager,
             step,
             expected_account_id,
+            recovery_auth: cached_auth,
             mode,
         }
     }
@@ -958,10 +962,10 @@ impl UnauthorizedRecovery {
 
         match self.step {
             UnauthorizedRecoveryStep::Reload => {
-                match self
-                    .manager
-                    .reload_if_account_id_matches(self.expected_account_id.as_deref())
-                {
+                match self.manager.reload_if_account_id_matches(
+                    self.recovery_auth.as_ref(),
+                    self.expected_account_id.as_deref(),
+                ) {
                     ReloadOutcome::ReloadedChanged => {
                         self.step = UnauthorizedRecoveryStep::RefreshToken;
                         return Ok(UnauthorizedRecoveryStepResult {
@@ -974,7 +978,8 @@ impl UnauthorizedRecovery {
                             auth_state_changed: Some(false),
                         });
                     }
-                    ReloadOutcome::Skipped => {
+                    ReloadOutcome::SkippedDifferentIdentity
+                    | ReloadOutcome::SkippedUnknownIdentity => {
                         self.step = UnauthorizedRecoveryStep::Done;
                         return Err(RefreshTokenError::Permanent(RefreshTokenFailedError::new(
                             RefreshTokenFailedReason::Other,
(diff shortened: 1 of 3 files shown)


codex-rs/core/src/auth.rs
```diff
@@ -1166,6 +1166,49 @@ impl AuthManager {
         }
     }
 
+    fn reload_for_refresh(&self, cached_auth: &CodexAuth) -> ReloadOutcome {
+        match cached_auth.get_account_id() {
+            Some(expected_account_id) => {
+                self.reload_if_account_id_matches(Some(expected_account_id.as_str()))
+            }
+            None => {
```
joshka-oai: nit: This tends to read better as:

```suggestion
        if let Some(expected_account_id) = cached_auth.get_account_id() {
            self.reload_if_account_id_matches(Some(expected_account_id.as_str()))
        } else {
```

codex-rs/core/src/auth.rs
```diff
@@ -1166,6 +1166,49 @@ impl AuthManager {
...
+                tracing::info!(
+                    "Reloading auth with legacy identity guard for refresh because account id is unavailable."
```
joshka-oai: What's a "legacy identity guard"? Add docs to clarify this to the function (why / expectations etc.)

codex-rs/core/src/auth.rs
```diff
@@ -1166,6 +1166,49 @@ impl AuthManager {
...
+                );
+                let new_auth = self.load_auth_from_storage();
+                if !new_auth
+                    .as_ref()
+                    .is_some_and(|new_auth| Self::same_refresh_identity(cached_auth, new_auth))
+                {
+                    tracing::info!(
+                        "Skipping auth reload due to legacy identity mismatch during refresh."
```
joshka-oai: same - legacy? Anchor this language with docs

codex-rs/login/src/auth/manager.rs
```diff
@@ -1166,6 +1166,49 @@ impl AuthManager {
         }
     }
 
+    fn reload_for_refresh(&self, cached_auth: &CodexAuth) -> ReloadOutcome {
```
joshka-oai, line 1164: Would be generally nice to have a list of expected outcomes - complex method, comments are simplifications / ways to drop the cognitive load on this.

codex-rs/login/src/auth/manager.rs
```diff
@@ -1166,6 +1166,49 @@ impl AuthManager {
...
+                if auth_changed {
+                    ReloadOutcome::ReloadedChanged
+                } else {
+                    ReloadOutcome::ReloadedNoChange
+                }
+            }
+        }
+    }
+
+    fn same_refresh_identity(cached_auth: &CodexAuth, new_auth: &CodexAuth) -> bool {
```
joshka-oai, line 1206: Doc would cut down the need to have to read this code to understand what same means (has the same userid or email)

codex-rs/login/src/auth/manager.rs
```diff
@@ -1313,15 +1356,28 @@ impl AuthManager {
...
+                        if error.reason == RefreshTokenFailedReason::Exhausted =>
+                    {
+                        match self.reload_for_refresh(&auth) {
+                            ReloadOutcome::ReloadedChanged => Ok(()),
+                            ReloadOutcome::ReloadedNoChange | ReloadOutcome::Skipped => {
+                                Err(RefreshTokenError::Permanent(error))
+                            }
+                        }
+                    }
+                    other => other,
```
joshka-oai, line 1387: It's definitely worth documenting why we're doing this here. Anyone reading this would have questions about the process. Anyone trying to understand it later and judge whether to change it would have similar concerns (i.e. why this code)

codex-rs/login/src/auth/manager.rs
ccy-oai: Updated this for readability.

- Removed the large explanatory block.
- Replaced the vague `legacy` / `same-user fallback` wording.
- Added one explicit line at the guard: ``account_id`` is the ChatGPT workspace/account boundary; if it is missing, we only reload from disk when `chatgpt_user_id` or `email` still match.

That keeps the code mostly self-explanatory while still making the boundary clear.

codex-rs/core/src/auth.rs
ccy-oai: Took this nit in the latest push while fixing the proactive stale-refresh issue.

codex-rs/core/src/auth.rs
ccy-oai: Updated this area to remove the old `legacy identity guard` wording. The code now uses one explicit boundary comment: `account_id` is the ChatGPT workspace/account boundary, and if it is missing we only reload from disk when `chatgpt_user_id` or `email` still match.

codex-rs/core/src/auth.rs
ccy-oai: Anchored this in code rather than adding a larger block comment. The current wording names the exact boundary and fallback signals instead of using `legacy` as a term of art.

codex-rs/login/src/auth/manager.rs
ccy-oai: I kept this to short boundary comments instead of a larger outcome block. The latest code makes the branches explicit: strict `account_id` match when present, fallback to `chatgpt_user_id` / `email` only when `account_id` is missing, and proactive refresh now bails out instead of persisting stale auth over a different on-disk account.

codex-rs/login/src/auth/manager.rs
ccy-oai: Added the short comment on `same_refresh_identity()` to make `same` concrete here: this path falls back to stable same-user claims (`chatgpt_user_id` or `email`) only when older auth files lack `account_id`.

codex-rs/login/src/auth/manager.rs
ccy-oai: Added a short reason comment here as well: this reload exists because another local client may already have refreshed the same auth lineage before we see `refresh_token_reused`.

codex-rs/login/src/auth/manager.rs
ccy-oai: Updated this again in the latest push. The mixed-shape recovery path is gone now: if cached auth has `account_id`, we require the same `account_id`; if cached auth does not have `account_id`, we only reload when the on-disk auth also lacks `account_id` and the same-user claims still match. That keeps the workspace boundary pinned instead of adopting a different workspace/account shape from disk.

ccy-oai: Updated this again in the latest push. Final rule: if cached auth has `account_id`, we only reload when both `account_id` and `chatgpt_user_id` remain compatible; if cached auth has no `account_id`, we keep the old behavior path and allow same-`chatgpt_user_id` on-disk auth to be adopted even if the newer file now includes `account_id`. `email` is no longer used as a reload boundary.

codex-rs/core/src/auth.rs
ccy-oai: Updated this again in the latest push. The code now avoids the old `legacy` terminology entirely: with `account_id` present we require compatible `account_id` + `chatgpt_user_id`; without `account_id` we stay on the old path and only use `chatgpt_user_id` compatibility.

codex-rs/login/src/auth/manager.rs
ccy-oai: Updated this in the latest push as well: `same` now means `chatgpt_user_id` compatibility. We do not use `email` for reload adoption anymore.

PR #15131, opened 2026-03-19
Add sandboxed upload_file and download_file tools

caseychow-oai:
## Summary
- add static Codex `upload_file` and `download_file` tools backed by a new `codex-file-transfer` helper crate
- preserve bundled experimental tool exposure when remote model metadata omits those tool flags
- fix ChatGPT file route construction and follow the JSON download-link response to the authenticated content fetch

## Key Decisions
- `upload_file` and `download_file` are built-in tools, not experimental tools; they are registered unconditionally rather than being gated on `experimental_supported_tools`
- the second hop to a backend-provided `download_url` does not forward ChatGPT auth headers; only the initial authenticated metadata request does
- download destinations use strict `cp`-style semantics, including basename sanitization when writing into an existing directory
- helper-side local filesystem failures return structured tool errors instead of collapsing into generic helper failures when possible
- upload and download stream file contents instead of buffering whole files in memory
- the new helper crate is wired into Bazel so Cargo-only success does not hide build-system regressions

## Validation
- cargo test -p codex-file-transfer -- --nocapture
- cargo test -p codex-core refresh_available_models_preserves_local_experimental_tools_when_remote_empty -- --nocapture
- cargo test -p codex-core test_models_json_default_model_includes_file_transfer_tools -- --nocapture
- cargo test -p codex-core parse_helper_output_preserves_request_context_on_nonzero_exit -- --nocapture
- just fix -p codex-file-transfer
- just fix -p codex-core
- live smoke via `just exec --json --sandbox danger-full-access --skip-git-repo-check ...` for upload and download round-trip

## Smoke notes
- `just exec` runs with cwd at `codex-rs`, so relative file paths resolve there during smoke tests
- verified a fresh upload returned an `openai-file://v1/...` URI and an immediate download wrote matching bytes (`some data`) to `/tmp/codex-download-smoke.txt`


codex-rs/core/src/models_manager/manager.rs
```diff
@@ -478,7 +478,17 @@ impl ModelsManager {
                 .iter()
                 .position(|existing| existing.slug == model.slug)
             {
-                existing_models[existing_index] = model;
```
caseychow-oai: This should not be necessary as a change.

PR #15134, opened 2026-03-19
Dismiss stale app-server requests after remote resolution

ebrevdo:
Dismiss stale TUI app-server approvals after remote resolution

When an approval, user-input prompt, or elicitation request is resolved by another client, the TUI now dismisses the matching local UI instead of leaving stale prompts behind and emitting a misleading local cancellation.

This change teaches pending app-server request tracking to map `serverRequest/resolved` notifications back to the concrete request type and stable request key, then propagates that resolved request into TUI prompt state. Approval, request-user-input, and MCP elicitation overlays now drop the resolved current or queued request quietly, advance to the next queued request when present, and avoid emitting abort/cancel events for stale UI.

The latest update also retires matching prompts while they are still deferred behind active streaming and suppresses buffered active-thread requests whose app-server request id has already been resolved before drain. `ChatWidget` removes a resolved request from both the deferred interrupt queue and the materialized bottom-pane stack, while active-thread request handling verifies the app-server request is still pending before showing a prompt. Lifecycle events such as exec begin/end remain queued so approved work can still render normally.

Tests cover resolved-request mapping, overlay dismissal behavior, deferred prompt pruning for same-turn user input, exec approval IDs, lifecycle-event retention, and the buffered active-thread ordering regression.

Validation:
- `just fmt`
- `git diff --check`
- `cargo test -p codex-tui resolved_buffered_approval_does_not_become_actionable_after_drain`
- `cargo test -p codex-tui enqueue_primary_thread_session_replays_buffered_approval_after_attach`
- `cargo test -p codex-tui chatwidget::interrupts`
- `just fix -p codex-tui`

### DIFF
--- a/codex-rs/tui/src/app/app_server_adapter.rs
+++ b/codex-rs/tui/src/app/app_server_adapter.rs
@@ -158,8 +158,12 @@ impl App {
     ) {
         match &notification {
             ServerNotification::ServerRequestResolved(notification) => {
-                self.pending_app_server_requests
-                    .resolve_notification(&notification.request_id);
+                if let Some(request) = self
+                    .pending_app_server_requests
+                    .resolve_notification(&notification.request_id)
+                {
+                    self.chat_widget.dismiss_app_server_request(&request);
+                }
             }
             ServerNotification::McpServerStatusUpdated(_) => {
                 self.refresh_mcp_startup_expected_servers_from_config();
--- a/codex-rs/tui/src/bottom_pane/request_user_input/mod.rs
+++ b/codex-rs/tui/src/bottom_pane/request_user_input/mod.rs
@@ -781,8 +781,8 @@ impl RequestUserInputOverlay {
 
         let queue_len = self.queue.len();
         self.queue
-            .retain(|queued_request| queued_request.call_id != *id);
-        if self.request.call_id == *id {
+            .retain(|queued_request| queued_request.turn_id != *id);
+        if self.request.turn_id == *id {
             self.advance_queue_or_complete();
             return true;
         }
@@ -1574,7 +1574,7 @@ mod tests {
 
         assert!(
             overlay.dismiss_app_server_request(&ResolvedAppServerRequest::UserInput {
-                id: "call-1".to_string(),
+                id: "turn-1".to_string(),
             })
         );
         assert!(overlay.done, "resolved request should close the overlay");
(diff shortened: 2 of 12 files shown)


codex-rs/tui/src/bottom_pane/request_user_input/mod.rs
```diff
@@ -759,14 +771,23 @@ impl RequestUserInputOverlay {
...
+
+    fn dismiss_resolved_request(&mut self, request: &ResolvedAppServerRequest) -> bool {
+        let ResolvedAppServerRequest::UserInput { id } = request else {
+            return false;
+        };
+
+        let queue_len = self.queue.len();
+        self.queue
+            .retain(|queued_request| queued_request.turn_id != *id);
+        if self.request.turn_id == *id {
```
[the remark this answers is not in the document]
ebrevdo: addressing

codex-rs/tui/src/app/app_server_adapter.rs
```diff
@@ -158,8 +158,12 @@ impl App {
     ) {
         match &notification {
             ServerNotification::ServerRequestResolved(notification) => {
-                self.pending_app_server_requests
-                    .resolve_notification(&notification.request_id);
+                if let Some(request) = self
```
fcoury-oai, line 161: I think there is still one stale-prompt ordering case here, and cross checked with Codex.

This can happen when the user is using two surfaces at the same time: one of them resolves the approval/input request, while the TUI has already queued the "show this prompt" event, just not displayed it.

The current code tries to dismiss the prompt immediately when `ServerRequestResolved` arrives, but there's nothing visible to dismiss just yet. Then the queued event is processed and the approval/input prompt is displayed. In this scenario the TUI will still show the prompt that was handled elsewhere already.

The downside is that the user can still interact with the prompt that is no longer valid, which can be very confusing.

I think the best way to address it would be to make the request display and resolved-request dismissal happen in the same ordered path.

Or we could retain resolved-request state and suppress a matching queued request when it is later
processed. 

A useful regression test would enqueue a `ServerRequest`, enqueue the matching `ServerRequestResolved` before draining active thread events, and assert that no prompt becomes actionable.

What do you think?

ebrevdo: let me add the regression test first then see what the cleanest solution would be there.  thanks for catching this race!

fcoury-oai (approved):
Looks good!

PR #15143, opened 2026-03-19
Add nickname_candidates to config.toml

baumann-oai:
## Summary
- Allow `config.toml` to define a global `nickname_candidates` list for spawned agents, with role-specific candidates still taking precedence when present.
- Keep the fallback behavior narrow: role config first, then global config, then built-in defaults.
- Update config loading/validation and the generated schema to use the same key, and cover user-vs-project precedence plus empty/duplicate cleanup in tests.

## Testing
- `just write-config-schema`
- `cargo test -p codex-core`

### DIFF
--- a/codex-rs/core/src/agent/control_tests.rs
+++ b/codex-rs/core/src/agent/control_tests.rs
@@ -962,6 +962,94 @@ async fn spawn_thread_subagent_uses_role_specific_nickname_candidates() {
 }
 
 #[tokio::test]
+async fn spawn_thread_subagent_uses_global_nickname_pool_when_role_has_no_candidates() {
+    let mut harness = AgentControlHarness::new().await;
+    harness.config.agent_nickname_pool = vec!["Scout".to_string()];
+    harness.config.agent_roles.insert(
+        "researcher".to_string(),
+        AgentRoleConfig {
+            description: Some("Research role".to_string()),
+            config_file: None,
+            nickname_candidates: None,
+        },
+    );
+    let (parent_thread_id, _parent_thread) = harness.start_thread().await;
+
+    let child_thread_id = harness
+        .control
+        .spawn_agent(
+            harness.config.clone(),
+            text_input("hello child"),
+            Some(SessionSource::SubAgent(SubAgentSource::ThreadSpawn {
+                parent_thread_id,
+                depth: 1,
+                agent_nickname: None,
+                agent_role: Some("researcher".to_string()),
+            })),
+        )
+        .await
+        .expect("child spawn should succeed");
+
+    let child_thread = harness
+        .manager
+        .get_thread(child_thread_id)
+        .await
+        .expect("child thread should be registered");
+    let snapshot = child_thread.config_snapshot().await;
+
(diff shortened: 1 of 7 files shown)


codex-rs/core/src/agent/guards.rs
```diff
@@ -139,7 +139,7 @@ impl Guards {
                 active_agents.nickname_reset_count += 1;
                 if let Some(metrics) = codex_otel::metrics::global() {
                     let _ = metrics.counter(
-                        "codex.multi_agent.nickname_pool_reset",
+                        "codex.multi_agent.nickname_candidates_reset",
```
jif-oai, line 142: why?

codex-rs/core/src/agent/control_tests.rs
```diff
@@ -961,6 +961,94 @@ async fn spawn_thread_subagent_uses_role_specific_nickname_candidates() {
...
+    let SessionSource::SubAgent(SubAgentSource::ThreadSpawn { agent_nickname, .. }) =
+        snapshot.session_source
+    else {
+        panic!("expected thread-spawn sub-agent source");
+    };
+    assert_eq!(agent_nickname, Some("Scout".to_string()));
+}
+
+#[tokio::test]
+async fn spawn_thread_subagent_prefers_role_candidates_over_global_nickname_candidates() {
```
jif-oai, line 1009: this seems a bit useless

codex-rs/core/src/config/agent_roles.rs
```diff
@@ -441,6 +441,47 @@ fn normalize_agent_role_nickname_candidates(
     Ok(Some(normalized_candidates))
 }
 
+pub(crate) fn normalize_global_agent_nickname_candidates(
```
jif-oai, line 444: can you make sure we don't already have code doing the exact same thing?

Also, I don't think this should be `pub(crate)`

codex-rs/core/src/config/config_tests.rs
```diff
@@ -1579,6 +1597,153 @@ profile = "project"
     Ok(())
 }
 
+#[tokio::test]
+async fn user_agents_nickname_candidates_are_used_without_project_override() -> std::io::Result<()>
```
jif-oai, line 1601: Let's avoid adding useless tests. Most of them are very low value. Let's try to have one very good test instead

codex-rs/core/src/config/mod.rs
```diff
@@ -1666,6 +1669,29 @@ pub struct AgentsToml {
...
+    /// ```
+    #[serde(
+        default,
+        deserialize_with = "deserialize_optional_agent_nickname_candidates"
+    )]
+    pub nickname_candidates: Option<AgentNicknameCandidatesToml>,
+
+    /// Named global fallback nickname candidate packs keyed by pack name.
+    #[serde(default)]
+    pub nickname_packs: BTreeMap<String, Vec<String>>,
```
jif-oai, line 1693: Do we really need this? This sounds going a bit too far IMO

Customisation is cool but up to a given limit IMO

PR #15154, opened 2026-03-19
adding full imagepath to tui

won-openai:
adding full path to TUI so image is open-able in the TUI after being generated. LImited to VSCode Terminal for now.


aibrahim-oai (approved):
add a snapshot

PR #15172, opened 2026-03-19
feat: cascade logs for multi-agents in `/feedback`

jif-oai:


codex-rs/feedback/src/lib.rs
```diff
@@ -398,6 +427,144 @@ impl FeedbackSnapshot {
...
 }
 
+fn merge_attachment_paths(
+    extra_attachment_paths: &[PathBuf],
+    discovered_attachment_paths: &[PathBuf],
+) -> Vec<PathBuf> {
+    let mut merged = Vec::with_capacity(
+        extra_attachment_paths
+            .len()
+            .saturating_add(discovered_attachment_paths.len()),
```
charley-oai, line 437: we could reuse this capacity for the hashset? Or if it doesn't matter just make them both `Vec::new()` and `HashSet..::new()`

charley-oai (approved):
I guess this is already a problem, but maybe more likely with a bunch of subagents; what happens when our sqlite exceeds the 20MiB Sentry upload limit due to rollout files?

Also importantly, can we include the subagent name in the rollout filename or somehow mark the subagent rollout files as subagents so it's easier to understand which file is which in Sentry?

PR #15190, opened 2026-03-19
Plumb MCP turn metadata through _meta

nicholasclark-openai:
## Summary

Some background. We're looking to instrument GA turns end to end. Right now a big gap is grouping mcp tool calls with their codex sessions. We send session id and turn id headers to the responses call but not the mcp/wham calls. 

Ideally we could pass the args as headers like with responses, but given the setup of the rmcp client, we can't send as headers without either changing the rmcp package upstream to allow per request headers or introducing a mutex which break concurrency. An earlier attempt made the assumption that we had 1 client per thread, which allowed us to set headers at the start of a turn. @pakrym mentioned that this assumption might break in the near future.

So the solution now is to package the turn metadata/session id into the _meta field in the post body and pull out in codex-backend.

- send turn metadata to MCP servers via `tools/call` `_meta` instead of assuming per-thread request headers on shared clients
- preserve the existing `_codex_apps` metadata while adding `x-codex-turn-metadata` for all MCP tool calls
- extend tests to cover both custom MCP servers and the codex apps search flow


codex-rs/core/src/codex.rs
```diff
@@ -1285,6 +1285,7 @@ impl Session {
 
     #[allow(clippy::too_many_arguments)]
     fn make_turn_context(
+        conversation_id: ThreadId,
```
nicholasclark-openai, line 1289: We need to pass along session id as well to get the full picture

owenlin0 (approved):

PR #15197, opened 2026-03-19
Add Codex Apps sediment file remapping

caseychow-oai:
## Summary
- bridge Codex Apps tools that declare `_meta["openai/fileParams"]` through the OpenAI file upload flow
- mask those file params in model-visible tool schemas so the model provides absolute local file paths instead of raw file payload objects
- rewrite those local file path arguments client-side into `ProvidedFilePayload`-shaped objects before the normal MCP tool call

## Details
- applies to scalar and array file params declared in `openai/fileParams`
- Codex uploads local files directly to the backend and uses the uploaded file metadata to build the MCP tool arguments locally
- this PR is input-only

## Verification
- `just fmt`
- `cargo test -p codex-core mcp_tool_call -- --nocapture`


caseychow-oai:
> there's one real bug here: in `rewrite_single_argument_string` we parse bare `file_*/file-*` as OpenAI file IDs before trying to resolve a local path, so a local file like `file_report.csv` gets rewritten to `sediment://file_report.csv` instead of being uploaded. can we either require `sediment://` for remote handles in this path, or check local path existence first?

will investigate

> also, can we split out the unrelated cleanup from this PR? the nextest / approval-matrix / code-mode timing changes, `cargo-bin` refactor, `apply_patch` fallback, and app-server `fileChange` test relaxation don't seem tied to the sediment bridge.

wilco, was trying to remove as much of that as possible but kept breaking CI. if i can make it work without i'll try to remove

caseychow-oai:
Removing local fallback entirely, we should always assume it's `sediment://` for remote files here.

codex-rs/tui/tests/suite/model_availability_nux.rs
```diff
@@ -140,7 +140,12 @@ trust_level = "trusted"
     let writer_tx = session.writer_sender();
     let interrupt_writer = writer_tx.clone();
     let interrupt_task = tokio::spawn(async move {
-        sleep(Duration::from_secs(2)).await;
+        let interrupt_delay = if cfg!(target_os = "macos") {
```
mzeng-openai: Can we clean up the unrelated diffs here and in other places.

codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -51,12 +60,18 @@ use std::path::Path;
...
+    pub(crate) namespace: Option<&'a str>,
+}
+
 /// Handles the specified tool call dispatches the appropriate
 /// `McpToolCallBegin` and `McpToolCallEnd` events to the `Session`.
 pub(crate) async fn handle_mcp_tool_call(
     sess: Arc<Session>,
     turn_context: &Arc<TurnContext>,
     call_id: String,
+    qualified_tool: QualifiedMcpTool<'_>,
```
mzeng-openai: Why do we need change the core flow here? Can we move the file transformation logic out of the core flow as much as possible? Essentially you only need to rewrite a few special params and I don't think you have to change the whole core flow to achieve that.

codex-rs/core/src/tools/handlers/mcp.rs
```diff
@@ -23,6 +24,8 @@ impl ToolHandler for McpHandler {
             session,
             turn,
             call_id,
+            tool_name,
```
mzeng-openai: Same here, do we have to change the core flow?

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -21,14 +21,18 @@ use crate::client_common::tools::ResponsesApiTool;
...
+    openai_file_bridge_enabled: bool,
 }
 
 pub(crate) const TOOL_SEARCH_TOOL_NAME: &str = "tool_search";
 pub(crate) const DEFAULT_LIMIT: usize = 8;
 
 impl ToolSearchHandler {
-    pub fn new(tools: HashMap<String, ToolInfo>) -> Self {
-        Self { tools }
+    pub fn new(tools: HashMap<String, ToolInfo>, openai_file_bridge_enabled: bool) -> Self {
```
mzeng-openai: I don't think you need to add it to tool search.

codex-rs/core/src/tools/spec.rs
```diff
@@ -1031,6 +1038,44 @@ fn create_view_image_tool(can_request_original_image_detail: bool) -> ToolSpec {
     })
 }
 
+fn create_download_openai_file_tool() -> ToolSpec {
```
mzeng-openai: Do we still need to add a model-facing tool when you are already manually transforming the tool params before/after tool calls?

codex-rs/core/tests/suite/shell_snapshot.rs
```diff
@@ -72,7 +72,12 @@ async fn wait_for_snapshot(codex_home: &Path) -> Result<PathBuf> {
 }
 
 async fn wait_for_file_contents(path: &Path) -> Result<String> {
-    let deadline = Instant::now() + Duration::from_secs(5);
+    let file_wait_timeout = if cfg!(target_os = "macos") {
```
mzeng-openai: Also please clean up here.

codex-rs/core/src/codex.rs
```diff
@@ -1152,6 +1154,41 @@ pub(crate) struct SessionSettingsUpdate {
 }
 
 impl Session {
+    fn add_session_openai_file_root_to_sandbox(
```
mzeng-openai: Can you confirm with @bolinfest if this is ok with the sandbox policy?

codex-rs/core/src/tools/spec.rs
caseychow-oai: Per discussion, I'll remove this on the theory that the model can retry the underlying tool call instead of just the download phase.

codex-rs/core/src/tools/handlers/tool_search.rs
caseychow-oai: There's two main things that we have to make sure to expose to the model, and the tool results seems to be the best thing. First is that we need to modify the parameter so that it is showing the model file shape instead of the path. The second is that we need to indicate to the model that that is what is happening on both the upload and the download flows.

I've gone ahead and tried to streamline it as much as possible to isolate it to just the case that we're caring about. If you feel like we can go further I'm happy to take a shot.

codex-rs/core/src/mcp_connection_manager.rs
```diff
@@ -1374,6 +1452,13 @@ async fn start_server_task(
         .await
         .map_err(StartupOutcomeError::from)?;
 
+    let server_supports_sandbox_state_capability =
+        server_supports_experimental_capability(&initialize_result, MCP_SANDBOX_STATE_CAPABILITY);
+    let server_supports_openai_file_bridge_capability = server_supports_experimental_capability(
```
bolinfest: This new capability is in the other direction, right?

- `MCP_SANDBOX_STATE_CAPABILITY` is a capability expressed by the _server_ so that the client knows it should broadcast sandbox-changed notifications
- `MCP_OPENAI_FILE_BRIDGE_CAPABILITY` is a capability expressed by the _client_ so the server knows it can send custom upload requests to the client

caseychow-oai: resolve offline

codex-rs/core/src/tools/spec.rs
```diff
@@ -2406,8 +2424,12 @@ fn dynamic_tool_to_openai_tool(
 }
 
 fn mcp_tool_to_openai_tool_parts(
-    tool: rmcp::model::Tool,
+    tool_info: &ToolInfo,
 ) -> Result<(String, JsonSchema, Option<JsonValue>), serde_json::Error> {
+    let mut tool = tool_info.tool.clone();
+    if tool_info.server_name == CODEX_APPS_MCP_SERVER_NAME {
```
bolinfest: Why is this special-cased?

codex-rs/core/src/tools/spec.rs
```diff
@@ -2406,8 +2424,12 @@ fn dynamic_tool_to_openai_tool(
 }
 
 fn mcp_tool_to_openai_tool_parts(
-    tool: rmcp::model::Tool,
+    tool_info: &ToolInfo,
 ) -> Result<(String, JsonSchema, Option<JsonValue>), serde_json::Error> {
```
bolinfest: create a struct for this return type and share it with `plain_mcp_tool_to_openai_tool_parts()`?

codex-rs/core/src/tools/spec.rs
```diff
@@ -2460,6 +2506,122 @@ fn mcp_call_tool_result_output_schema(structured_content_schema: JsonValue) -> J
...
 }
 
+/// Sanitize a JSON Schema (as serde_json::Value) so it can fit our limited
+/// JsonSchema enum. This function:
+/// - Ensures every schema object has a "type". If missing, infers it from
+///   common keywords (properties => object, items => array, enum/const/format => string)
+///   and otherwise defaults to "string".
+/// - Fills required child fields (e.g. array items, object properties) with
+///   permissive defaults when absent.
+fn sanitize_json_schema(value: &mut JsonValue) {
```
bolinfest: I'm a bit suspicious we need this: what is it about this PR that introduces this need? Is there any chance `schemars::JsonSchema` supports this already?

Also, I don't see any test coverage for this (or is it a side effect from the tests that exercise `mcp_tool_to_openai_tool()`?)

codex-rs/core/src/codex.rs
```diff
@@ -6602,14 +6603,19 @@ pub(crate) async fn built_tools(
...
     Ok(Arc::new(ToolRouter::from_config(
         &turn_context.tools_config,
         ToolRouterParams {
             mcp_tools: has_mcp_servers.then(|| {
                 mcp_tools
                     .into_iter()
-                    .map(|(name, tool)| (name, tool.tool))
+                    .map(|(name, tool_info)| {
+                        let mut tool = tool_info.tool;
+                        if tool_info.server_name == CODEX_APPS_MCP_SERVER_NAME {
```
bolinfest: Again, why the special-casing?

codex-rs/core/src/mcp_openai_file.rs
```diff
@@ -0,0 +1,187 @@
+use serde_json::Map;
```
bolinfest: Can you please add an overview docstring (in Rust, that's `//!`) that explains the strategy of what we're trying to do here.

codex-rs/core/src/mcp_openai_file.rs
```diff
@@ -0,0 +1,187 @@
...
+
+const META_OPENAI_FILE_PARAMS: &str = "openai/fileParams";
+
+pub(crate) fn declared_openai_file_input_param_names(
+    meta: Option<&Map<String, JsonValue>>,
+) -> Vec<String> {
+    declared_top_level_fields(meta, META_OPENAI_FILE_PARAMS)
+}
+
+pub(crate) fn mask_input_schema_for_file_path_params(
```
bolinfest: We appear to be mutating the `JsonValue` in place, which doesn't seem great to me. Can we just inspect the `_meta` and then decide whether we need to associate some sort of transformation function with this tool call?

codex-rs/core/src/mcp_openai_file.rs
```diff
@@ -0,0 +1,187 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use pretty_assertions::assert_eq;
+
+    #[test]
+    fn declared_openai_file_fields_ignore_nested_names() {
+        let meta = serde_json::json!({
+            "openai/fileParams": ["file", "nested.value", "files[0]", "attachments"],
```
bolinfest: so we need to invent our own "selector"-like language to support this? we have to support this at arbitrary depths?

is there any way to avoid that?

codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -453,6 +453,44 @@ fn record_server_fields(span: &Span, url: Option<&str>) {
     }
 }
 
+async fn execute_mcp_tool_call(
+    sess: &Session,
+    turn_context: &TurnContext,
+    invocation: McpToolExecutionInvocation<'_>,
+) -> Result<CallToolResult, String> {
+    let rewritten_arguments = rewrite_mcp_tool_arguments_for_openai_files(
```
bolinfest, line 461: This would be cleaner if some "rewrite" function could be associated with the `invocation`. Having this special-case in the middle of our stack is not great.

codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -1027,9 +1073,120 @@ pub(crate) async fn lookup_mcp_tool_metadata(
...
             .and_then(serde_json::Value::as_object)
             .cloned(),
+        openai_file_input_params: Some(declared_openai_file_input_param_names(
+            tool_info.tool.meta.as_deref(),
+        ))
+        .filter(|params| !params.is_empty()),
     })
 }
 
+async fn rewrite_mcp_tool_arguments_for_openai_files(
```
bolinfest: This feature-specific logic should not be in a general location like `mcp_tool_call.rs`.

codex-rs/codex-api/src/files.rs
```diff
@@ -0,0 +1,350 @@
+use std::path::Path;
```
bolinfest, line 1: Please do not add this to `codex-rs/core`: it should go in some other location.

codex-rs/core/src/tools/spec.rs
caseychow-oai: I don't feel strongly here, this is a Codex recommendation: https://github.com/openai/codex/pull/15197#discussion_r2998167069

In general, it's a bit safer because we don't want all MCPs to get access to internal file upload mechanisms, especially because they may have a separate file upload mechanism; that part I think is something we can peel out as we build out more of the MCP spec for file uploads beyond what exists today.

codex-rs/core/src/codex.rs
caseychow-oai: Discussed above: https://github.com/openai/codex/pull/15197#discussion_r3001463798

codex-rs/core/src/mcp_openai_file.rs
caseychow-oai: This is pushing for the opposite actually--it denies nested values. I'll change it so that it treats those JSON keys literally.

codex-rs/core/src/tools/spec.rs
caseychow-oai: I dug into this locally and I think the broad helper was more than this PR needed. The general schema sanitization already lives in `codex-rs/tools` inside `parse_tool_input_schema()` and has dedicated tests there. For this PR, the only extra compatibility fix that still looked justified was the narrow top-level `properties: null -> {}` normalization for the plain MCP conversion path. I trimmed `spec.rs` down to that narrower normalization and added a focused regression test for `plain_mcp_tool_to_openai_tool()` so the compatibility case is covered directly.

codex-rs/codex-mcp/src/mcp_connection_manager.rs
```diff
@@ -1372,6 +1394,8 @@ async fn start_server_task(
         .await
         .map_err(StartupOutcomeError::from)?;
 
+    let server_supports_sandbox_state_capability =
+        server_supports_experimental_capability(&initialize_result, MCP_SANDBOX_STATE_CAPABILITY);
```
pakrym-oai: is this an unrelated change? same q for call_tool_with_client

codex-rs/core/tests/suite/unified_exec.rs
```diff
@@ -171,7 +171,7 @@ async fn create_workspace_directory(
     test: &TestCodex,
```
pakrym-oai, line 189: the new features needs an integration test (like other features in `suite`)

AGENTS.md
```diff
@@ -23,6 +23,7 @@ In the codex-rs folder where the rust code lives:
 - When making a change that adds or changes an API, ensure that the documentation in the `docs/` folder is up to date if applicable.
 - Prefer private modules and explicitly exported public crate API.
 - If you change `ConfigToml` or nested config types, run `just write-config-schema` to update `codex-rs/core/config.schema.json`.
+- For MCP tool-call interception work, prefer capability-driven custom MCP requests over broad generic call-path rewrites when the server can own the transformation logic.
```
pakrym-oai: What does this mean? I don't think I understand it.

codex-rs/core/src/mcp_openai_file.rs
```diff
@@ -0,0 +1,654 @@
...
+            "openai/fileOutputs": ["output", "artifacts/0"]
+        });
+        let meta = meta.as_object().expect("meta object");
+
+        assert_eq!(
+            declared_openai_file_input_param_names(Some(meta)),
+            vec![
+                "file".to_string(),
+                "nested.value".to_string(),
+                "files[0]".to_string(),
```
pakrym-oai: what does this mean? do we actually support only one item of the array being a file?

codex-rs/core/src/mcp_openai_file.rs
```diff
@@ -0,0 +1,654 @@
...
+
+use crate::codex::Session;
+use crate::codex::TurnContext;
+use codex_api::OpenAiFileUploadAuth;
+use codex_api::upload_local_file;
+use codex_login::CodexAuth;
+use serde_json::Map;
+use serde_json::Value as JsonValue;
+
+const META_OPENAI_FILE_PARAMS: &str = "openai/fileParams";
```
pakrym-oai: ooc why not use a schema annotation instead of the meta? like a `x-openai-file` ?

codex-rs/codex-api/src/files.rs
```diff
@@ -0,0 +1,341 @@
...
+        source: serde_json::Error,
+    },
+    #[error("OpenAI file upload for `{file_id}` is not ready yet")]
+    UploadNotReady { file_id: String },
+    #[error("OpenAI file upload for `{file_id}` failed: {message}")]
+    UploadFailed { file_id: String, message: String },
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct OpenAiFileUploadAuth {
```
pakrym-oai: is there no existing type that can be used for this?

codex-rs/codex-api/src/lib.rs
```diff
@@ -49,6 +50,13 @@ pub use crate::endpoint::ResponsesOptions;
 pub use crate::endpoint::ResponsesWebsocketClient;
 pub use crate::endpoint::ResponsesWebsocketConnection;
 pub use crate::error::ApiError;
+pub use crate::files::OPENAI_FILE_UPLOAD_LIMIT_BYTES;
```
pakrym-oai: do all of these need to be public?

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -62,6 +64,11 @@ impl ToolHandler for ToolSearchHandler {
 
         let mut entries: Vec<(String, ToolInfo)> = self.tools.clone().into_iter().collect();
         entries.sort_by(|a, b| a.0.cmp(&b.0));
+        for (_name, tool) in &mut entries {
+            if tool.server_name == CODEX_APPS_MCP_SERVER_NAME {
+                mask_model_visible_tool_input_schema(&mut tool.tool);
+            }
+        }
```
pakrym-oai: we should nto be spreading these special cases around. Ideally the schema rewrite and param rewrite  logic is encapsulated in McpManager

codex-rs/core/src/codex.rs
```diff
@@ -6867,14 +6868,19 @@ pub(crate) async fn built_tools(
...
-
     Ok(Arc::new(ToolRouter::from_config(
         &turn_context.tools_config,
         ToolRouterParams {
             mcp_tools: has_mcp_servers.then(|| {
                 mcp_tools
                     .into_iter()
-                    .map(|(name, tool)| (name, tool.tool))
+                    .map(|(name, tool_info)| {
+                        let mut tool = tool_info.tool;
```
pakrym-oai: codex shouldn't know about these rewrites let's have mcp manager own the whole logic.

codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -981,10 +1023,17 @@ pub(crate) async fn lookup_mcp_tool_metadata(
...
         .list_all_tools()
         .await;
-
     let tool_info = tools
         .into_values()
         .find(|tool_info| tool_info.server_name == server && tool_info.tool.name == tool_name)?;
+    mcp_tool_approval_metadata_from_tool_info(turn_context, server, tool_info).await
+}
+
+async fn mcp_tool_approval_metadata_from_tool_info(
```
pakrym-oai: why split?

codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -453,6 +454,46 @@ fn record_server_fields(span: &Span, url: Option<&str>) {
...
+    sanitize_mcp_tool_result_for_model(
+        turn_context
+            .model_info
+            .input_modalities
+            .contains(&InputModality::Image),
+        Ok(result),
+    )
+}
+
+struct McpToolExecutionInvocation<'a> {
```
pakrym-oai: is this purely to avoid parameters?

pakrym-oai (changes requested):
My main concern is that the knowledge about schema/argument rewrite is spread all across the codebase. The result of mcp_manager.tools can't be safely consumed without having to remember to call rewrite.

Also we must have integration tests for this.

We might also need some retries for upload requests.

codex-rs/core/src/mcp_tool_call.rs
caseychow-oai: Accepted. This was mostly just bundling parameters, so I removed the `McpToolExecutionInvocation` wrapper locally and now pass the values directly to `execute_mcp_tool_call`. I'll include it in the next push with the rest of the review fixes.

codex-rs/codex-api/src/lib.rs
caseychow-oai: ack, trimming

codex-rs/core/src/mcp_openai_file.rs
caseychow-oai: Accepted. The test was only trying to prove that openai/fileParams entries are treated as literal field names, not that we support array-index addressing. I changed the fixture locally to use plain field names so it no longer suggests only one array item can be a file.

codex-rs/core/src/mcp_tool_call.rs
caseychow-oai: Codex slop that made its way through many rereads haha, inlining

codex-rs/codex-api/src/files.rs
caseychow-oai: Accepted. There is already an API-layer auth abstraction here, so I removed the file-upload-specific auth struct locally. upload_local_file now accepts AuthProvider, the core caller passes CoreAuthProvider, and codex-api still stays independent of codex-login.

codex-rs/codex-mcp/src/mcp_connection_manager.rs
caseychow-oai: Accepted. This was unrelated cleanup churn, so I removed it locally: call_tool_with_client is inlined back into call_tool, and the experimental capability check is back inline at startup.

codex-rs/core/src/tools/handlers/tool_search.rs
caseychow-oai: Accepted. I moved the model-visible file-param schema masking into codex-mcp via ToolInfo::model_visible_tool, so tool_search no longer knows about the Codex Apps special case; it just asks each ToolInfo for its model-visible form.

codex-rs/core/src/codex.rs
caseychow-oai: Accepted. I removed the schema rewrite branch from codex.rs too. Router construction now converts MCP ToolInfo values through ToolInfo::model_visible_tool, with the Codex Apps file-param masking owned in codex-mcp next to the MCP tool inventory logic.

codex-rs/core/src/mcp_openai_file.rs
caseychow-oai: there's a lot of shapes we can do here, I'm going with what's already been done and shipped into the ecosystem to avoid inventing new things unnecessarily: https://developers.openai.com/apps-sdk/reference#_meta-fields-on-tool-descriptor

AGENTS.md
caseychow-oai: Goal is to try to nudge codex to hit a smaller surface area for MCP mods. I can tweak

codex-rs/core/tests/suite/unified_exec.rs
caseychow-oai: ack, writing

caseychow-oai: Accepted. I added a core suite integration test that boots the Codex Apps MCP test server, advertises an app tool with openai/fileParams metadata, has the model call it with a local report.txt path, mocks the OpenAI file create/blob PUT/finalize endpoints, and verifies the Apps tools/call request receives the uploaded-file object. It also asserts the /v1/responses tool schema exposes that parameter as a local path string.

codex-rs/codex-mcp/src/mcp_connection_manager.rs
```diff
@@ -194,6 +196,94 @@ pub struct ToolInfo {
     pub connector_description: Option<String>,
 }
 
+impl ToolInfo {
+    pub fn model_visible_tool(&self) -> Tool {
```
pakrym-oai: why do we ever want to see the unprocessed tool?

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -62,6 +62,9 @@ impl ToolHandler for ToolSearchHandler {
 
         let mut entries: Vec<(String, ToolInfo)> = self.tools.clone().into_iter().collect();
         entries.sort_by(|a, b| a.0.cmp(&b.0));
+        for (_name, tool) in &mut entries {
+            tool.tool = tool.model_visible_tool();
```
pakrym-oai: can we make it so tools returned from McpManager are already in correct state without re-processing them?

codex-rs/codex-mcp/src/mcp_connection_manager.rs
```diff
@@ -194,6 +196,94 @@ pub struct ToolInfo {
...
+        if self.server_name == CODEX_APPS_MCP_SERVER_NAME {
+            mask_model_visible_tool_input_schema(&mut tool);
+        }
+        tool
+    }
+}
+
+const META_OPENAI_FILE_PARAMS: &str = "openai/fileParams";
+
+pub fn declared_openai_file_input_param_names(
```
pakrym-oai, line 204: should this be a method on Tool?

codex-rs/codex-mcp/src/mcp_connection_manager_tests.rs
```diff
@@ -60,6 +61,85 @@ fn create_codex_apps_tools_cache_context(
     }
 }
 
+#[test]
+fn declared_openai_file_fields_treat_names_literally() {
+    let meta = serde_json::json!({
+        "openai/fileParams": ["file", "input_file", "attachments"],
+        "openai/fileOutputs": ["output", "artifact"]
```
pakrym-oai: nit: rem?

pakrym-oai (approved):
Can we process tools before they returned from McpManager?

codex-rs/codex-mcp/src/mcp_connection_manager.rs
caseychow-oai: I kept this as a private helper rather than adding a trait / wrapper around rmcp's `Tool`: it is local to Codex Apps schema masking, needs no shared behavior from `Tool` itself, and is now applied at the MCP manager return boundaries so callers receive the model-visible shape without remembering to reprocess it.

PR #15203, opened 2026-03-19
Add v8-poc consumer of our new built v8

cconger:
This adds a dummy v8-poc project that in Cargo links against our prebuilt binaries and the ones provided by rusty_v8 for non musl platforms.  This demonstrates that we can successfully link and use v8 on all platforms that we want to target.

In bazel things are slightly more complicated.  Since the libraries as published have libc++ linked in already we end up with a lot of double linked symbols if we try to use them in bazel land.  Instead we fall back to building rusty_v8 and v8 from source (cached of course) on the platforms we ship to.

There is likely some compatibility drift in the windows bazel builder that we'll need to reconcile before we can re-enable them.  I'm happy to be on the hook to unwind that.


codex-rs/v8-poc/src/lib.rs
```diff
@@ -0,0 +1,65 @@
+//! Bazel-wired proof-of-concept crate reserved for future V8 experiments.
+
+/// Returns the Bazel label for this proof-of-concept crate.
+#[must_use]
+pub fn bazel_target() -> &'static str {
```
pakrym-oai, line 5: do we need this?

pakrym-oai (approved):

PR #15209, opened 2026-03-19
chat history image

won-openai:
- **Restore image generation items in resumed thread history**
- **changed save directory to codex_home**


codex-rs/app-server-protocol/src/protocol/thread_history.rs
```diff
@@ -186,34 +187,46 @@ impl ThreadHistoryBuilder {
...
+            } => {
+                if role != "user" {
+                    return;
+                }
 
-        if role != "user" {
-            return;
+                let Some(hook_prompt) = parse_hook_prompt_message(id.as_ref(), content) else {
+                    return;
+                };
```
won-openai, line 216: this isn't my change, this is the change on main that i rebased onto

aibrahim-oai: rebase

PR #15216, opened 2026-03-19
Move terminal module to terminal-detection crate

aibrahim-oai:
- Move core/src/terminal.rs and its tests into a standalone terminal-detection workspace crate.
- Update direct consumers to depend on codex-terminal-detection and import terminal APIs directly.


codex-rs/terminal/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "codex-terminal"
```
joshka-oai: nit: this module is probably more about terminal detection / info than it is about the terminal as a whole (unless you plan to move the other terminal specific related stuff in here too).

joshka-oai (approved):

codex-rs/exec-server/src/server/filesystem.rs
```diff
@@ -36,7 +36,7 @@ pub(crate) struct ExecServerFileSystem {
 impl Default for ExecServerFileSystem {
     fn default() -> Self {
         Self {
-            file_system: Arc::new(Environment.get_filesystem()),
+            file_system: Arc::new(Environment::default().get_filesystem()),
```
joshka-oai, line 39: nit: maybe worth splitting / doing as a separate PR, not a big deal but this is likely annoying other people and would be something that other PRs maybe fix - easier to rebase when needed when there's an obvious single commit that fixes it.

joshka-oai (approved):

PR #15226, opened 2026-03-19
core: eagerly initialize guardian sessions

charley-oai:
## Summary
- eagerly initialize the guardian trunk session on guardian-routed turns, including legacy `UserInput` turns, instead of waiting for the first approval request
- keep a single cached guardian trunk and fork on demand when the trunk is busy or a stale trunk cannot be replaced immediately; there is intentionally no guardian-specific fork cap in this PR
- preserve mainline guardian follow-up-review behavior while simplifying the trunk preparation flow into explicit eager-init vs real-review helpers
- do not send a separate guardian-specific prewarm Responses request; this change only moves guardian session creation earlier and relies on existing session startup prewarm behavior


charley-oai:
@codex review

charley-oai:
@codex review for correctness AND any simplification opportunities

codex-rs/core/src/guardian/review_session.rs
```diff
@@ -429,17 +769,21 @@ impl GuardianReviewSessionManager {
...
+            fork_permit,
+        );
 
-        let (outcome, _) = run_review_on_session(review_session.as_ref(), &params, deadline).await;
-        if let Some(review_session) = self.take_active_ephemeral(&review_session).await {
-            cleanup.disarm();
+        let execution_result =
+            run_review_on_session(review_session.as_ref(), &params, deadline).await;
+        if let Some((review_session, fork_permit)) = cleanup.into_parts() {
+            self.fork_pool.unregister(&review_session).await;
```
jif-oai: What if `shutdown_and_wait` is slow and not over yet? That sounds a bit racy...

codex-rs/core/src/codex.rs
```diff
@@ -4458,6 +4459,7 @@ mod handlers {
     }
 
     pub async fn user_input_or_turn(sess: &Arc<Session>, sub_id: String, op: Op) {
+        let is_user_turn = matches!(&op, Op::UserTurn { .. });
```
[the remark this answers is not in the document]
jif-oai: Minimal but fair

jif-oai (commented):
All of this sounds correct but it looks a bit like re-inventing what already exist in multiple places in the codebase... 
Are we sure about the global design of this guardian?

codex-rs/core/src/guardian/review_session.rs
charley-oai: Deleted this code

codex-rs/core/src/guardian/review_session.rs
```diff
@@ -442,17 +614,30 @@ impl GuardianReviewSessionManager {
...
-            EphemeralReviewCleanup::new(Arc::clone(&self.state), Arc::clone(&review_session));
+        {
+            let mut state = self.state.lock().await;
+            if state.shutdown_started {
+                drop(state);
+                review_session.shutdown_in_background();
+                return GuardianReviewSessionOutcome::Aborted;
+            }
+            state.active_forks.push(Arc::clone(&review_session));
+        }
```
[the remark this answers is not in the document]
charley-oai: this was intentional

codex-rs/core/src/codex.rs
```diff
@@ -4529,6 +4530,10 @@ mod handlers {
             // new_turn_with_sub_id already emits the error event.
             return;
         };
+        if routes_approval_to_guardian(current_context.as_ref()) {
```
jif-oai: I think this eager prewarm runs before we know whether this input will actually start a new turn. 
E.g. if the input gets steered into an already-active turn, this background spawn can still take guardian's shared spawn lock and delay the active turn's real approval review

codex-rs/core/src/guardian/review_session.rs
```diff
@@ -322,28 +351,215 @@ impl GuardianReviewSessionManager {
...
+    /// Returns true only when eager init should try to spawn a trunk now.
+    ///
+    /// Eager init is strictly best-effort: if any trunk already exists, or shutdown has started,
+    /// the caller should do nothing and let the real review path handle trunk reuse/replacement.
+    async fn prepare_trunk_for_eager_init(&self) -> bool {
+        let state = self.state.lock().await;
+        if state.shutdown_started {
+            return false;
+        }
+        state.trunk.is_none()
```
jif-oai: Eager init only checks whether any trunk exists, not whether the existing trunk still matches the current guardian config. That means the eager path intentionally skips the trunk-refresh so config updates won't get picked up unless cold start

codex-rs/core/src/guardian/review_session.rs
```diff
@@ -228,18 +248,57 @@ impl Drop for EphemeralReviewCleanup {
...
+            match resolve_guardian_review_config(parent_session.as_ref(), parent_turn.as_ref())
+                .await
+            {
+                Ok(resolved) => resolved,
+                Err(err) => {
+                    warn!("failed to resolve guardian review config: {err}");
+                    return;
+                }
+            };
+        let params = GuardianReviewSessionParams {
```
jif-oai: this seems odd to trigger an empty review just to start the eager?

codex-rs/core/src/guardian/review_session.rs
```diff
@@ -250,65 +309,35 @@ impl GuardianReviewSessionManager {
...
-                }
-
-                if state.trunk.is_none() {
-                    let spawn_cancel_token = CancellationToken::new();
-                    let review_session = match run_before_review_deadline_with_cancel(
+            Ok(GuardianTrunkState::Ready(trunk)) => trunk,
+            Ok(GuardianTrunkState::ShutdownStarted) => {
+                return GuardianReviewSessionOutcome::Aborted;
+            }
+            Ok(GuardianTrunkState::NeedsSpawn) => {
```
jif-oai: This should be unreachable right?

codex-rs/core/src/guardian/review_session.rs
```diff
@@ -322,28 +351,215 @@ impl GuardianReviewSessionManager {
...
+        external_cancel: Option<&CancellationToken>,
+    ) -> Result<GuardianTrunkState, GuardianReviewSessionOutcome> {
+        match self.prepare_trunk(next_reuse_key).await {
+            state @ (GuardianTrunkState::Ready(_) | GuardianTrunkState::ShutdownStarted) => {
+                return Ok(state);
+            }
+            GuardianTrunkState::NeedsSpawn => {}
+        }
+
+        let spawn_guard =
```
jif-oai: this logic seems to be a partial duplicate of what is above. In general this code is not super clear. There are a lot of different branches and a lot of small logic duplication

jif-oai (commented):
The code is not super clear. Mainly in review_session.rs

I think it needs better documentation and maybe a refactor (at least of the structure) to make this more streamlined

codex-rs/core/src/codex.rs
charley-oai: Resolved. Eager guardian init now only starts in the `NoActiveTurn` path after `steer_input(...)` fails

codex-rs/core/src/guardian/review_session.rs
charley-oai: I think config updates will still get applied in the review path, even if the prewarming didn't pick those up (will just be a latency hit but not frequent)

codex-rs/core/src/guardian/review_session.rs
charley-oai: Resolved. The eager path no longer constructs fake `GuardianReviewSessionParams` with empty prompt/schema. It now resolves config and directly spawns a guardian trunk session via the shared spawn helper.

codex-rs/core/src/guardian/review_session.rs
charley-oai: Fixed, the public review helper no longer returns `GuardianTrunkState`; `NeedsSpawn` is now internal to trunk preparation. `run_review(...)` gets `Ok(Some(trunk)), Ok(None)` for shutdown, or an error outcome, so that unreachable caller branch is gone

codex-rs/core/src/guardian/review_session.rs
charley-oai: Refactored the code a bit, collapsed the repeated “spawn guardian session under deadline/cancel” code into one helper

Hardanish-Singh (commented):
LGTM

PR #15227, opened 2026-03-19
Protect first-time project .codex creation in Windows sandbox

rreichel3-oai:
## Problem

Codex already treated an existing top-level project `./.codex` directory as protected, but there was still a first-creation gap on Windows.

If `./.codex` did not exist yet, the Windows sandbox could not attach a deny ACL to it, which meant a turn could create files under it, such as `./.codex/config.toml`, without going through the same protection as later modifications.

## What this changes

This PR closes that first-creation gap in the Windows enforcement layer:

- Windows sandbox policy
  - treat the top-level project `./.codex` path as denied even when it does not exist yet
  - keep `.git` and `.agents` behavior unchanged
- Windows ACL setup
  - reserve a missing top-level project `./.codex` directory before applying the deny ACL, so first-time creation cannot bypass protection
  - apply the deny ACL to the reserved directory for both the command CWD workspace and other writable roots
- tests
  - add Windows-focused regressions for missing `./.codex` handling in writable roots
  - cover the reserve-before-deny behavior for top-level project `.codex`

## Scope

This change is intentionally scoped to protecting the top-level project `.codex` subtree from agent writes on Windows.

It does not make `.codex` unreadable, and it does not change the product behavior around loading project skills from `.codex` when project config is untrusted.

## Why this shape

The fix is pointed rather than broad:
- it preserves the current model of `project .codex is protected from writes`
- it closes the Windows-specific first-write hole in the current ACL-based sandbox
- it avoids folding a larger permissions-model redesign into this PR

One Windows-specific detail is that the sandbox may create an empty top-level `./.codex/` so there is an existing filesystem object to attach the deny ACL to.

## Validation

### DIFF
--- a/codex-rs/app-server/tests/suite/v2/plugin_read.rs
+++ b/codex-rs/app-server/tests/suite/v2/plugin_read.rs
@@ -44,6 +44,11 @@ use tokio::task::JoinHandle;
 use tokio::time::timeout;
 
 const DEFAULT_TIMEOUT: Duration = Duration::from_secs(10);
+// Windows CI runners can take longer to finish the app-server startup handshake.
+#[cfg(windows)]
+const DEFAULT_INIT_TIMEOUT: Duration = Duration::from_secs(25);
+#[cfg(not(windows))]
+const DEFAULT_INIT_TIMEOUT: Duration = DEFAULT_TIMEOUT;
 
 #[tokio::test]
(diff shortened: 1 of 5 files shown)


viyatb-oai (approved):
lgtm - have you had a chance to test it out?

PR #15244, opened 2026-03-19
Pin Python SDK app-server stdio to UTF-8 on Windows

### LINKED ISSUE #14311: [codex-app-server Python SDK] UnicodeDecodeError with `Popen(text=True)` on non-ASCII Windows locales
### What version of the Codex App are you using (From “About Codex” dialog)?

0.114.0

plus

Windows 11

## Summary

`AppServerClient.start()` uses `subprocess.Popen(text=True)` without specifying `encoding`, causing `UnicodeDecodeError` on Windows systems with non-UTF-8 default encodings (e.g. cp949 for Korean, cp932 for Japanese, cp936 for Chinese).

## Environment

- OS: Windows 11 (Korean locale, default codepage cp949)
- Python: 3.12+
- codex-app-server SDK

## Error

```
UnicodeDecodeError: 'cp949' codec can't decode byte 0xed in position 170: illegal multibyte sequence
```

```
Turn error: 'cp949' codec can't decode byte 0x80 in position 2460: illegal multibyte sequence
```

This occurs when the codex subprocess outputs UTF-8 text (e.g. Korean characters in conversation) and Python tries to decode it using the system default encoding (cp949).

In `client.py` line 178:

```python
self._proc = subprocess.Popen(
    args,
    stdin=subprocess.PIPE,
    stdout=subprocess.PIPE,
    stderr=subprocess.PIPE,
    text=True,          # <-- uses locale.getpreferredencoding() as encoding
    cwd=self.config.cwd,
    env=env,
    bufsize=1,
)
```

`text=True` without an explicit `encoding` parameter defaults to `locale.getpreferredencoding(False)`, which returns `cp949` on Korean Windows, `cp932` on Japanese Windows, etc. The codex CLI (Node.js) outputs UTF-8, so any non-ASCII content causes a decode error.

## Suggested Fix

Add `encoding="utf-8"` to the `Popen` call:

```python
self._proc = subprocess.Popen(
    args,
    stdin=subprocess.PIPE,
    stdout=subprocess.PIPE,
    stderr=subprocess.PIPE,
    text=True,
    encoding="utf-8",   # <-- explicit UTF-8
    cwd=self.config.cwd,
    env=env,
    bufsize=1,
)
```

## Workaround

Downstream users can reconfigure the streams after `client.start()`:

```python
client = AppServerClient(config)
client.start()

proc = client._proc
if proc is not None:
    if proc.stdout and hasattr(proc.stdout, "reconfigure"):
        proc.stdout.reconfigure(encoding="utf-8")
    if proc.stderr and hasattr(proc.stderr, "reconfigure"):
        proc.stderr.reconfigure(encoding="utf-8")
    if proc.stdin and hasattr(proc.stdin, "reconfigure"):
        proc.stdin.reconfigure(encoding="utf-8")

client.initialize()
```

shaqayeq-oai:
## TL;DR
Pin the Python app-server SDK subprocess pipes to UTF-8 so Windows users on non-UTF-8 locales do not hit `UnicodeDecodeError` when the `codex` child emits UTF-8 text.

- add `encoding="utf-8"` to the `subprocess.Popen(...)` call in `AppServerClient.start()`
- add a focused regression test that asserts the client launches the subprocess with UTF-8 text I/O
- validates with `python -m pytest sdk/python/tests/test_client_rpc_methods.py sdk/python/tests/test_client_process_launch.py sdk/python/tests/test_public_api_runtime_behavior.py`

Fixes #14311.


owenlin0 (approved):

sdk/python/tests/test_client_process_launch.py
```diff
@@ -0,0 +1,38 @@
...
+
+from typing import Any
+
+import pytest
+
+import codex_app_server.client as client_module
+from codex_app_server.client import AppServerClient, AppServerConfig
+
+
+def test_start_launches_subprocess_with_utf8_text_io(
```
owenlin0: i find that codex can add tests that are not super high signal. imo this is one of them (just asserting that we pass `encoding = "utf-8"` to Popen)

shaqayeq-oai: yea i agree. will remove

PR #15256, opened 2026-03-20
[MCP caching] split MCP connections to a backend + lightweight handle

jgershen-oai:
## Summary
Split MCP connection ownership into a (to-be-shared) backend and a lightweight session-local client. 
This PR keeps all other behavior unchanged; the motivation is to be able to cache the connection.

## Testing
- cargo check -p codex-core
- cargo test -p codex-core mcp_connection_manager::tests::

### DIFF
--- a/codex-rs/core/src/mcp_connection_manager.rs
+++ b/codex-rs/core/src/mcp_connection_manager.rs
@@ -595,50 +595,27 @@ pub struct SandboxState {
 }
 
 /// A thin wrapper around a set of running [`RmcpClient`] instances.
-pub(crate) struct McpConnectionManager {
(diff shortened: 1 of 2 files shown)


codex-rs/core/src/mcp_connection_manager.rs
```diff
@@ -1708,3 +1884,656 @@ mod mcp_init_error_display_tests {}
 #[cfg(test)]
 #[path = "mcp_connection_manager_tests.rs"]
 mod tests;
+#[cfg(any())]
+mod old_tests {
```
pakrym-oai, line 1888: all tests should probably go into mcp_connection_manager_tests.

PR #15258, opened 2026-03-20
[MCP caching]: pool MCP backends in thread manager

jgershen-oai:
Depends on #15256.

## Summary
This is the PR that actually makes MCP backend reuse happen across loaded threads.

- add a `ThreadManager`-owned pool of shared MCP backends for enabled `Stdio` MCP servers
- keep `StreamableHttp` MCPs, including `codex_apps`, per-thread/local
- force-rebuild the pooled stdio half on refresh, and reacquire it when sandbox state changes
- add stdio-focused app-server lifecycle coverage for unsubscribe, archive, and archive -> unarchive -> resume

## Why
The concrete bug here is duplicate local MCP processes when multiple loaded threads use the same stdio MCP configuration. Pooling only stdio backends fixes that without turning `codex_apps` into a shared HTTP bottleneck or introducing cross-thread elicitation routing.

## Testing
- cargo check -p codex-core
- cargo test -p codex-core mcp_connection_manager::tests::
- cargo test -p codex-core pooled_new_turn_keeps_local_mcp_startup_token_when_sandbox_state_changes
- cargo test -p codex-app-server thread_mcp_pool
- cargo test -p codex-app-server mcp_server_elicitation


codex-rs/core/src/mcp_connection_manager.rs
```diff
@@ -1104,6 +1110,224 @@ impl SharedMcpBackend {
...
+    #[allow(clippy::too_many_arguments)]
+    async fn acquire_or_create(
+        &self,
+        key: SharedMcpBackendCacheKey,
+        acquire_mode: SharedMcpBackendAcquireMode,
+        mcp_servers: &HashMap<String, McpServerConfig>,
+        store_mode: OAuthCredentialsStoreMode,
+        auth_entries: HashMap<String, McpAuthStatusEntry>,
+        tx_event: Sender<Event>,
+        initial_sandbox_state: SandboxState,
```
pakrym-oai, line 1257: Is all sandboxing stuff purely for `codex/sandbox-state/update`? If so the shell mcp is going to get removed today/tomorrow so we can shed the sandboxing complexity.

codex-rs/core/src/mcp_connection_manager.rs
```diff
@@ -1104,6 +1110,224 @@ impl SharedMcpBackend {
...
+            serde_json::Value::Object(canonical)
+        }
+        other => other,
+    }
+}
+
+impl SharedMcpBackendCacheKey {
+    pub(crate) fn new(
+        mcp_servers: &HashMap<String, McpServerConfig>,
+        store_mode: OAuthCredentialsStoreMode,
```
pakrym-oai, line 1139: Does this ever change mid session?

codex-rs/core/src/state/service.rs
```diff
@@ -33,6 +34,7 @@ use tokio_util::sync::CancellationToken;
 pub(crate) struct SessionServices {
     pub(crate) mcp_connection_manager: Arc<RwLock<McpConnectionManager>>,
     pub(crate) mcp_startup_cancellation_token: Mutex<CancellationToken>,
+    pub(crate) shared_mcp_backend_pool: Option<Arc<SharedMcpBackendPool>>,
```
pakrym-oai, line 37: Having both mcp_manager and shared_mcp_backend_pool here feels like bad API that leaks implementation details.

codex-rs/core/src/codex.rs
```diff
@@ -1924,17 +1943,40 @@ impl Session {
...
-            tx_event.clone(),
-            sandbox_state,
-            config.codex_home.clone(),
-            codex_apps_tools_cache_key(auth),
-            tool_plugin_provenance,
-        )
+        let (mcp_connection_manager, cancel_token) = async {
+            match sess.services.shared_mcp_backend_pool.as_ref() {
+                Some(shared_mcp_backend_pool) => {
+                    McpConnectionManager::new_with_pool(
```
pakrym-oai, line 1949: this logic shouldn't be in codex. Pooling decisions should be hidden by McpConnectionManager or SharedMcpBackendPool

codex-rs/core/src/mcp_connection_manager.rs
```diff
@@ -1104,6 +1110,224 @@ impl SharedMcpBackend {
...
+
+            serde_json::Value::Object(canonical)
+        }
+        other => other,
+    }
+}
+
+impl SharedMcpBackendCacheKey {
+    pub(crate) fn new(
+        mcp_servers: &HashMap<String, McpServerConfig>,
```
pakrym-oai, line 1138: I'd expect a key to be per server not per group of them, what am I missing?

codex-rs/core/src/mcp_connection_manager.rs
```diff
@@ -1104,6 +1110,224 @@ impl SharedMcpBackend {
...
+    }
+
+    fn backend(&self) -> Arc<SharedMcpBackend> {
+        Arc::clone(&self.inner.backend)
+    }
+}
+
+struct SharedMcpBackendLeaseInner {
+    backend: Arc<SharedMcpBackend>,
+    cancel_token: CancellationToken,
```
pakrym-oai, line 1223: this is the startup cancellation, right? how do you drop the lease/free the server?

PR #15261, opened 2026-03-20
Store guardian transcript boundary on review session

charley-oai:
## Summary
- store the parent transcript checkpoint on the cached guardian review session instead of rollout reconstruction or parent session state
- slice guardian transcript evidence from that checkpoint so follow-up reviews only include transcript since the last terminal guardian review
- preserve the checkpoint when the cached guardian trunk is recreated for a config change and add coverage for approved and denied follow-up reviews

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -3392,7 +3392,7 @@ impl Session {
             state.replace_history(items, reference_context_item);
         }
         self.guardian_review_session
-            .set_parent_history_boundary(None)
+            .set_parent_history_boundary(/*boundary*/ None)
             .await;
     }
 
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -3387,8 +3387,13 @@ impl Session {
         items: Vec<ResponseItem>,
         reference_context_item: Option<TurnContextItem>,
     ) {
-        let mut state = self.state.lock().await;
-        state.replace_history(items, reference_context_item);
+        {
+            let mut state = self.state.lock().await;
+            state.replace_history(items, reference_context_item);
+        }
+        self.guardian_review_session
+            .set_parent_history_boundary(None)
+            .await;
     }
 
     pub(crate) async fn replace_compacted_history(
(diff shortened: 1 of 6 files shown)


Hardanish-Singh (commented):
LGTM

Raymondriter:
I'd close this as superseded rather than request a rebase.

The follow-up transcript slice this PR wanted is already on main as `GuardianTranscriptCursor` + `GuardianPromptMode::{Full, Delta}`, stored on the review session (`last_reviewed_transcript_cursor`) and copied through fork snapshots. Main also version-checks against parent history (so compaction/rewrite falls back to a full prompt), keeps root authorization out of the delta, and after #41215 resets the cursor when reviewer compaction drops the original transcript.

This branch is from March (`be7bc5c`, last code Mar 24) and is `mergeable_state: dirty` against current main. The design here is a raw `raw_items().len()` index on the session manager, which is weaker than what landed and has its own holes even on that snapshot:

- Sequential same-turn reviews (two tools, no history append) slice `[len..]` and send `<no retained transcript entries>`. Tests only cover follow-up after a new user turn.
- Abort / timeout / fail-closed still call `mark_terminal_guardian_review_boundary`, so a review that never assessed the current history still hides it from the next one.
- Only `replace_history` clears the checkpoint; `replace_compacted_history` is untouched.
- Config-change trunk recreate keeps the parent slice while spawning with `initial_history: None`, so the new reviewer loses both prior assessments and original user authorization. Main starts a new trunk in Full mode instead.

CI was green on Mar 24 against then-main; that does not apply now. No maintainer review (the Aug 1 LGTM is unaffiliated). Please reopen from current main only if some delta-cursor case is still missing. I don't think this patch is the vehicle.

sylvesterkaczmarek (commented):
Moving the boundary into the review-session manager is cleaner than reconstructing it from rollout state, and using the parent history index makes the next guardian prompt deterministic. One concurrency edge seems worth pinning: `replace_history()` resets the boundary, while an older in-flight guardian can later mark a terminal boundary against the replaced history. If replacement/compaction can race a review, that stale completion could suppress newly replaced history from the next review. A generation/token check around the terminal marker, or a regression proving this race is impossible, would make the ownership model much stronger.

PR #15275, opened 2026-03-20
feat: prefer git for curated plugin sync 

xl-openai:
start with git clone, fallback to http.


alexsong-oai (approved):
wondering why git clone is preferred?

PR #15276, opened 2026-03-20
Code mode on v8

cconger:
Moves Code Mode to a new crate with no dependencies on codex.  This create encodes the code mode semantics that we want for lifetime, mounting, tool calling.

The model-facing surface is mostly unchanged. `exec` still runs raw JavaScript, `wait` still resumes or terminates a `cell_id`, nested tools are still available through `tools.*`, and helpers like `text`, `image`, `store`, `load`, `notify`, `yield_control`, and `exit` still exist.

The major change is underneath that surface:

- Old code mode was an external Node runtime.
- New code mode is an in-process V8 runtime embedded directly in Rust.
- Old code mode managed cells inside a long-lived Node runner process.
- New code mode manages cells in Rust, with one V8 runtime thread per active `exec`.
- Old code mode used JSON protocol messages over child stdin/stdout plus Node worker-thread messages.
- New code mode uses Rust channels and direct V8 callbacks/events.

This PR also fixes the two migration regressions that fell out of that substrate change:

- `wait { terminate: true }` now waits for the V8 runtime to actually stop before reporting termination.
- synchronous top-level `exit()` now succeeds again instead of surfacing as a script error.

---

- `core/src/tools/code_mode/*` is now mostly an adapter layer for the public `exec` / `wait` tools.
- `code-mode/src/service.rs` owns cell sessions and async control flow in Rust.
- `code-mode/src/runtime/*.rs` owns the embedded V8 isolate and JavaScript execution.
- each `exec` spawns a dedicated runtime thread plus a Rust session-control task.
- helper globals are installed directly into the V8 context instead of being injected through a source prelude.
- helper modules like `tools.js` and `@openai/code_mode` are synthesized through V8 module resolution callbacks in Rust.

---

Also added a benchmark for showing the speed of init and use of a code mode env:
```
$ cargo bench -p codex-code-mode --bench exec_overhead -- --samples 30 --warm-iterations 25 --tool-counts 0,32,128
Finished [`bench` profile [optimized]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 0.18s
     Running benches/exec_overhead.rs (target/release/deps/exec_overhead-008c440d800545ae)
exec_overhead: samples=30, warm_iterations=25, tool_counts=[0, 32, 128]
scenario       tools samples    warmups      iters      mean/exec       p95/exec       rssΔ p50       rssΔ max
cold_exec          0      30          0          1         1.13ms         1.20ms        8.05MiB        8.06MiB
warm_exec          0      30          1         25       473.43us       512.49us      912.00KiB        1.33MiB
cold_exec         32      30          0          1         1.03ms         1.15ms        8.08MiB        8.11MiB
warm_exec         32      30          1         25       509.73us       545.76us      960.00KiB        1.30MiB
cold_exec        128      30          0          1         1.14ms         1.19ms        8.30MiB        8.34MiB
warm_exec        128      30          1         25       575.08us       591.03us      736.00KiB      864.00KiB
memory uses a fresh-process max RSS delta for each scenario
```


codex-rs/code-mode/src/description.rs
```diff
@@ -0,0 +1,630 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub struct ToolDefinition {
+    pub name: String,
+    pub description: String,
+    pub kind: CodeModeToolKind,
+    pub input_schema: Option<JsonValue>,
+    pub output_schema: Option<JsonValue>,
+}
+
+impl CodeModeToolDefinition for ToolDefinition {
```
pakrym-oai: Do we need a trait?

codex-rs/code-mode/src/description.rs
```diff
@@ -0,0 +1,630 @@
...
+    fn name(&self) -> &str;
+    fn description(&self) -> &str;
+    fn kind(&self) -> CodeModeToolKind;
+    fn input_schema(&self) -> Option<&JsonValue>;
+    fn output_schema(&self) -> Option<&JsonValue>;
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ToolReference {
+    pub module_path: String,
```
pakrym-oai: I think this can be removed now. All tools are global

codex-rs/code-mode/src/description.rs
```diff
@@ -0,0 +1,630 @@
...
+    pub namespace: Vec<String>,
```
pakrym-oai: Same with namespace.

codex-rs/code-mode/src/description.rs
```diff
@@ -0,0 +1,630 @@
...
+    }
+
+    sections.join("\n\n")
+}
+
+pub fn build_wait_tool_description() -> &'static str {
+    WAIT_DESCRIPTION_TEMPLATE
+}
+
+pub fn tool_reference(tool_name: &str) -> ToolReference {
```
pakrym-oai: Pretty sure we can get rid of this.

codex-rs/core/src/tools/code_mode_description.rs
```diff
@@ -1,299 +1,51 @@
...
-
-    if let Some(items) = map.get("prefixItems").and_then(serde_json::Value::as_array) {
-        let item_types = items
-            .iter()
-            .map(render_json_schema_to_typescript_inner)
-            .collect::<Vec<_>>();
-        if !item_types.is_empty() {
-            return format!("[{}]", item_types.join(", "));
+        ToolSpec::Function(tool) => {
+            let definition = codex_code_mode::ToolDefinition {
```
pakrym-oai: This back and fourth is a bit heavy considering we only trying to append to desription.

codex-rs/code-mode/benches/exec_overhead.rs
```diff
@@ -0,0 +1,550 @@
+use std::collections::HashMap;
```
pakrym-oai: Do we need to check this in?

codex-rs/code-mode/src/description.rs
cconger: Huge this will simplify a lot of the code to not have to support virtual modules.

codex-rs/core/src/tools/js_repl/mod_tests.rs
```diff
@@ -2365,8 +2365,12 @@ async fn js_repl_imported_local_files_can_access_repl_globals() -> anyhow::Resul
         return Ok(());
     }
```
pakrym-oai, line 2367: Why did this have to change?

codex-rs/code-mode/src/runtime/value.rs
```diff
@@ -0,0 +1,190 @@
...
+use crate::response::ImageDetail;
+
+pub(super) fn serialize_output_text(
+    scope: &mut v8::PinScope<'_, '_>,
+    value: v8::Local<'_, v8::Value>,
+) -> Result<String, String> {
+    if value.is_string() {
+        return Ok(value.to_rust_string_lossy(scope));
+    }
+    if value.is_undefined()
```
pakrym-oai, line 10: nit: combine with the above?

codex-rs/code-mode/src/runtime/value.rs
```diff
@@ -0,0 +1,190 @@
...
+        (value.to_rust_string_lossy(scope), None)
+    } else if value.is_object() && !value.is_array() {
+        let Ok(object) = v8::Local::<v8::Object>::try_from(value) else {
+            throw_type_error(
+                scope,
+                "image expects a non-empty image URL string or an object with image_url and optional detail",
+            );
+            return Err(());
+        };
+        let Some(image_url_key) = v8::String::new(scope, "image_url") else {
```
pakrym-oai: oh wow, we have to allocate these!

codex-rs/code-mode/src/runtime/value.rs
```diff
@@ -0,0 +1,190 @@
...
+
+pub(super) fn normalize_output_image(
+    scope: &mut v8::PinScope<'_, '_>,
+    value: v8::Local<'_, v8::Value>,
+) -> Result<FunctionCallOutputContentItem, ()> {
+    let (image_url, detail) = if value.is_string() {
+        (value.to_rust_string_lossy(scope), None)
+    } else if value.is_object() && !value.is_array() {
+        let Ok(object) = v8::Local::<v8::Object>::try_from(value) else {
+            throw_type_error(
```
pakrym-oai: can we structure this method so we use Result<FunctionCallOutputContentItem, string> when parsing and then have one place that calls throw_type_error on failure?

Might make for a more standard-looking rust.

codex-rs/code-mode/src/runtime/module_loader.rs
```diff
@@ -0,0 +1,333 @@
...
+    };
+    let global = v8::Global::new(scope, module);
+    if let Some(state) = scope.get_slot_mut::<RuntimeState>() {
+        state.module_cache.insert(specifier.to_string(), global);
+    }
+    Some(module)
+}
+
+fn helper_module_source(enabled_tools: &[EnabledToolMetadata], specifier: &str) -> Option<String> {
+    if specifier == "tools.js" {
```
pakrym-oai: I think we can kill tools.js and import from `@openai/code_mode`

codex-rs/code-mode/src/runtime/module_loader.rs
```diff
@@ -0,0 +1,333 @@
...
+) -> bool {
+    scope
+        .get_slot::<RuntimeState>()
+        .map(|state| state.exit_requested)
+        .unwrap_or(false)
+        && exception.is_string()
+        && exception.to_rust_string_lossy(scope) == EXIT_SENTINEL
+}
+
+pub(super) fn resolve_tool_response(
```
pakrym-oai, line 66: nit: this method almost doesn't fit in this file.

codex-rs/code-mode/benches/exec_overhead.rs
cconger: Can drop it. 👍 Was pleased that drastically increasing tool count doesn't seem to have a lot of weight increase on the startup time.

codex-rs/code-mode/src/runtime/mod.rs
```diff
@@ -0,0 +1,376 @@
...
+
+#[derive(Clone, Debug)]
+pub struct WaitRequest {
+    pub cell_id: String,
+    pub yield_time_ms: u64,
+    pub terminate: bool,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum RuntimeResponse {
```
pakrym-oai, line 42: ToolResponse maybe?

codex-rs/code-mode/src/runtime/mod.rs
```diff
@@ -0,0 +1,376 @@
...
+    Terminated {
+        cell_id: String,
+        content_items: Vec<FunctionCallOutputContentItem>,
+    },
+    Result {
+        cell_id: String,
+        content_items: Vec<FunctionCallOutputContentItem>,
+        stored_values: HashMap<String, JsonValue>,
+        error_text: Option<String>,
+        max_output_tokens_per_exec_call: Option<usize>,
```
pakrym-oai: this shouldn't be here, right? In both cases we parse it from tool call and have it on the other side

codex-rs/code-mode/src/runtime/mod.rs
```diff
@@ -0,0 +1,376 @@
...
+    ToolCall {
+        id: String,
+        name: String,
+        input: Option<JsonValue>,
+    },
+    Notify {
+        call_id: String,
+        text: String,
+    },
+    Result {
```
pakrym-oai, line 95: There is some overlap between this and RuntimeResponse. Both have stored valis and notify.

codex-rs/code-mode/src/runtime/mod.rs
```diff
@@ -0,0 +1,376 @@
...
+    let config = RuntimeConfig {
+        tool_call_id: request.tool_call_id,
+        enabled_tools,
+        source: request.source,
+        stored_values: request.stored_values,
+    };
+
+    thread::spawn(move || {
+        let runtime_event_tx = event_tx.clone();
+        let result = std::panic::catch_unwind(AssertUnwindSafe(|| {
```
pakrym-oai: Who is planning to panic?

codex-rs/code-mode/src/runtime/mod.rs
```diff
@@ -0,0 +1,376 @@
...
+            run_runtime(config, runtime_event_tx, command_rx, isolate_handle_tx)
+        }));
+        if let Err(panic_payload) = result {
```
pakrym-oai: This panic handling makes me uneasy. We shouldn't panic.

codex-rs/code-mode/src/runtime/mod.rs
```diff
@@ -0,0 +1,376 @@
...
+        Ok(pending_promise) => pending_promise,
+        Err(error_text) => {
+            let stored_values = scope
+                .get_slot::<RuntimeState>()
+                .map(|state| state.stored_values.clone())
+                .unwrap_or_default();
+            let _ = event_tx.send(RuntimeEvent::Result {
+                stored_values,
+                error_text: Some(error_text),
+            });
```
pakrym-oai: can we extract the send error boilderplate?

pakrym-oai: Maybe have normal rust error handling in inner method and outer method that has a single place to report errors.

codex-rs/code-mode/src/runtime/mod.rs
```diff
@@ -0,0 +1,376 @@
...
+            return;
+        }
+    };
+
+    match module_loader::completion_state(scope, pending_promise.as_ref()) {
```
pakrym-oai, line 210: do we need the pre-check or can we let the loop run and hit completion case naturally?

codex-rs/code-mode/src/runtime/globals.rs
```diff
@@ -0,0 +1,187 @@
...
+                .map(|tool| {
+                    serde_json::json!({
+                        "name": tool.global_name,
+                        "description": tool.description,
+                    })
+                })
+                .collect::<Vec<_>>()
+        })
+        .unwrap_or_default();
+    json_to_v8(scope, &JsonValue::Array(all_tools))
```
pakrym-oai: it it easier to build the dict directly?

codex-rs/code-mode/src/runtime/globals.rs
```diff
@@ -0,0 +1,187 @@
...
+        .ok_or_else(|| "failed to build ALL_TOOLS metadata".to_string())
+}
+
+fn build_console_object<'s>(
+    scope: &mut v8::PinScope<'s, '_>,
+) -> Result<v8::Local<'s, v8::Object>, String> {
+    let console = v8::Object::new(scope);
```
pakrym-oai: is there a built-in console object? if not let's just nix this. 

I was overriding node vm console object to noop only because there was a default implementation.

codex-rs/code-mode/src/runtime/callbacks.rs
```diff
@@ -0,0 +1,246 @@
...
+
+    let Some(resolver) = v8::PromiseResolver::new(scope) else {
+        throw_type_error(scope, "failed to create tool promise");
+        return;
+    };
+    let promise = resolver.get_promise(scope);
+
+    let resolver = v8::Global::new(scope, resolver);
+    let Some(state) = scope.get_slot_mut::<RuntimeState>() else {
+        throw_type_error(scope, "runtime state unavailable");
```
pakrym-oai, line 39: same nit about native rust error handling that turns into throw_type_error in one location. Maybe .map_err(throw_type_error) in the caller

codex-rs/code-mode/src/runtime/callbacks.rs
```diff
@@ -0,0 +1,246 @@
...
+        throw_type_error(scope, "failed to allocate text output item");
+        return;
+    };
+    let Some(text_value) = v8::String::new(scope, &text) else {
+        throw_type_error(scope, "failed to allocate text output item");
+        return;
+    };
+    item.set(scope, type_key.into(), type_value.into());
+    item.set(scope, text_key.into(), text_value.into());
+    retval.set(item.into());
```
pakrym-oai: did we return content item from `text` . We should maybe stop.

codex-rs/code-mode/src/runtime/callbacks.rs
```diff
@@ -0,0 +1,246 @@
...
+        let _ = state.event_tx.send(RuntimeEvent::Notify {
+            call_id: state.tool_call_id.clone(),
+            text: text.clone(),
+        });
+    }
+    let Some(value) = v8::String::new(scope, &text) else {
+        throw_type_error(scope, "failed to allocate notify result");
+        return;
+    };
+    retval.set(value.into());
```
pakrym-oai: we probably shouldn't return anything.

pakrym-oai: Any chance I can ask you to extend tests to check for semantics we missed? Like return values of helper methods. 

It's ok if not in this PR.

codex-rs/code-mode/src/runtime/callbacks.rs
```diff
@@ -0,0 +1,246 @@
...
+    scope: &mut v8::PinScope<'_, '_>,
+    args: v8::FunctionCallbackArguments,
+    mut retval: v8::ReturnValue<v8::Value>,
+) {
+    let value = if args.length() == 0 {
+        v8::undefined(scope).into()
+    } else {
+        args.get(0)
+    };
+    let text = match serialize_output_text(scope, value) {
```
pakrym-oai, line 168: Good question, do we serialize the argument here? Or do we expect text? Might make sense to align with `text` method as you did here.

codex-rs/core/src/tools/code_mode/mod.rs
```diff
@@ -29,155 +31,213 @@ use crate::truncate::TruncationPolicy;
...
+            .map_err(|_| {
+                format!("failed to inject exec notify message for cell {cell_id}: no active turn")
             })
-            .collect::<Vec<_>>()
-            .join("\n\n");
-        sections.push(nested_tool_reference);
     }
+}
+
+pub(crate) fn is_code_mode_nested_tool(tool_name: &str) -> bool {
```
pakrym-oai: can we drop these proxies?

codex-rs/core/src/tools/code_mode/mod.rs
```diff
@@ -29,155 +31,213 @@ use crate::truncate::TruncationPolicy;
...
-                FunctionToolOutput::from_content(delta_items, Some(success)),
+            Ok(FunctionToolOutput::from_content(
+                content_items,
+                Some(success),
             ))
         }
     }
 }
 
+enum CodeModeExecutionStatus {
```
pakrym-oai: Isn't this very similar to RuntimeResponse ?

codex-rs/code-mode/src/service.rs
```diff
@@ -0,0 +1,604 @@
...
+            }),
+        }
+    }
+
+    pub async fn stored_values(&self) -> HashMap<String, JsonValue> {
+        self.inner.stored_values.lock().await.clone()
+    }
+
+    pub async fn replace_stored_values(&self, values: HashMap<String, JsonValue>) {
+        *self.inner.stored_values.lock().await = values;
```
pakrym-oai, line 75: nit for later. this should probably be merge.

codex-rs/code-mode/src/service.rs
```diff
@@ -0,0 +1,604 @@
...
+    }
+}
+
+pub struct CodeModeTurnWorker {
+    shutdown_tx: Option<oneshot::Sender<()>>,
+}
+
+impl Drop for CodeModeTurnWorker {
+    fn drop(&mut self) {
+        if let Some(shutdown_tx) = self.shutdown_tx.take() {
```
pakrym-oai, line 224: do we want to join the task? or abort the task?

pakrym-oai: bit worrysome to let it hang

codex-rs/code-mode/src/service.rs
```diff
@@ -0,0 +1,604 @@
...
+    let mut response_tx = Some(initial_response_tx);
+    let mut termination_requested = false;
+    let mut runtime_closed = false;
+    let mut yield_timer: Option<std::pin::Pin<Box<tokio::time::Sleep>>> = None;
+
+    loop {
+        tokio::select! {
+            maybe_event = async {
+                if runtime_closed {
+                    std::future::pending::<Option<RuntimeEvent>>().await
```
pakrym-oai, line 309: if runtime is closed what are we waiting for?

codex-rs/code-mode/src/service.rs
```diff
@@ -0,0 +1,594 @@
...
+                        if termination_requested {
+                            if let Some(response_tx) = response_tx.take() {
+                                let _ = response_tx.send(RuntimeResponse::Terminated {
+                                    cell_id: cell_id.clone(),
+                                    content_items: std::mem::take(&mut content_items),
+                                });
+                            }
+                            break;
+                        }
+                        let result = PendingResult {
```
pakrym-oai, line 387: this branch and the                 `let Some(event) = maybe_event else {` above are very similar

codex-rs/code-mode/src/runtime/callbacks.rs
cconger: Yes, I want to throw the semantics doc at this and build out some more tests.

codex-rs/code-mode/src/service.rs
cconger: Yeah agreed.  I think I probably need to have stricter lifecycle management for the Worker and channel.  I need to think on it some.

codex-rs/code-mode/src/runtime/globals.rs
cconger: Not really haha, but tried it as an exercise.

codex-rs/code-mode/src/service.rs
cconger: If the runtime queue gets closed its possible that there was already a Result to be processed in our event_rx or if one was pending.  We keep listening until the v8 thread resolves which it should quickly after we hit a terminate_execution on the handle.

codex-rs/code-mode/src/runtime/mod.rs
cconger: RuntimeEvent is js => code_mode
RuntimeResponse is code_mode => Core

I think there's a lot of overlap but I think we likely shouldn't coalesce these types.

codex-rs/code-mode/src/runtime/value.rs
cconger: Yeah... Part of why doing the bulk json decodes are nice.

codex-rs/code-mode/src/runtime/mod.rs
cconger: I like this idea, I'd like to think on it some more and maybe follow up.

codex-rs/code-mode/src/runtime/mod.rs
cconger: Hmm, I think we need this in the cases where the script finished right away, since the loop is blocked on the command_rx.recv().

pakrym-oai (approved):

PR #15338, opened 2026-03-20
fix: fall back to vendored bubblewrap when system bwrap lacks --argv0

bolinfest:
## Why

Fixes [#15283](https://github.com/openai/codex/issues/15283), where sandboxed tool calls fail on older distro `bubblewrap` builds because `/usr/bin/bwrap` does not understand `--argv0`. The upstream [bubblewrap v0.9.0 release notes](https://github.com/containers/bubblewrap/releases/tag/v0.9.0) explicitly call out `Add --argv0`. Flipping `use_legacy_landlock` globally works around that compatibility bug, but it also weakens the default Linux sandbox and breaks proxy-routed and split-policy cases called out in review.

The follow-up Linux CI failure was in the new launcher test rather than the launcher logic: the fake `bwrap` helper stayed open for writing, so Linux would not exec it. This update also closes the user-visibility gap from review by surfacing the same startup warning when `/usr/bin/bwrap` is present but too old for `--argv0`, not only when it is missing.

## What Changed

- keep `use_legacy_landlock` default-disabled
- teach `codex-rs/linux-sandbox/src/launcher.rs` to fall back to the vendored bubblewrap build when `/usr/bin/bwrap` does not advertise `--argv0` support
- add launcher tests for supported, unsupported, and missing system `bwrap`
- write the fake `bwrap` test helper to a closed temp path so the supported-path launcher test works on Linux too
- extend the startup warning path so Codex warns when `/usr/bin/bwrap` is missing or too old to support `--argv0`
- mirror the warning/fallback wording across `codex-rs/linux-sandbox/README.md` and `codex-rs/core/README.md`, including that the fallback is the vendored bubblewrap compiled into the binary
- cite the upstream `bubblewrap` release that introduced `--argv0`

## Verification

- `bazel test --config=remote --platforms=//:rbe //codex-rs/linux-sandbox:linux-sandbox-unit-tests --test_filter=launcher::tests::prefers_system_bwrap_when_help_lists_argv0 --test_output=errors`
- `cargo test -p codex-core system_bwrap_warning`
- `cargo check -p codex-exec -p codex-tui -p codex-tui-app-server -p codex-app-server`

### DIFF
--- a/codex-rs/core/README.md
+++ b/codex-rs/core/README.md
@@ -61,9 +61,11 @@ cases like `/repo = write`, `/repo/a = none`, `/repo/a/b = write`, where the
 more specific writable child must reopen under a denied parent.
 
 The Linux sandbox helper prefers `/usr/bin/bwrap` whenever it is available and
-falls back to the vendored bubblewrap path otherwise. When `/usr/bin/bwrap` is
-missing, Codex also surfaces a startup warning through its normal notification
-path instead of printing directly from the sandbox helper.
+supports the required argv-rewrite flags, and falls back to the vendored
+bubblewrap path compiled into the binary otherwise. When `/usr/bin/bwrap` is
+missing or too old to support the required flags, Codex also surfaces a startup
+warning through its normal notification path instead of printing directly from
+the sandbox helper.
 
 ### Windows
 
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -96,6 +96,8 @@ use std::collections::HashMap;
 use std::io::ErrorKind;
 use std::path::Path;
 use std::path::PathBuf;
+#[cfg(target_os = "linux")]
+use std::process::Command;
 
 use crate::config::permissions::compile_permission_profile;
 use crate::config::permissions::network_proxy_config_from_profile_network;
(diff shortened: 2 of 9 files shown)


codex-rs/linux-sandbox/src/launcher.rs
```diff
@@ -24,17 +26,41 @@ pub(crate) fn exec_bwrap(argv: Vec<String>, preserved_files: Vec<File>) -> ! {
...
-    if !Path::new(SYSTEM_BWRAP_PATH).is_file() {
+    static LAUNCHER: OnceLock<BubblewrapLauncher> = OnceLock::new();
+    LAUNCHER
+        .get_or_init(|| preferred_bwrap_launcher_for_path(Path::new(SYSTEM_BWRAP_PATH)))
+        .clone()
+}
+
+fn preferred_bwrap_launcher_for_path(system_bwrap_path: &Path) -> BubblewrapLauncher {
+    if !system_bwrap_supports_argv0(system_bwrap_path) {
         return BubblewrapLauncher::Vendored;
```
evawong-oai, line 37: This makes sense to me. One thing I would love to confirm before we ship is whether we have an end to end repro showing the vendored fallback succeeds on one of the older distro environments from #15283. My only hesitation is that #14919 and #14963 taught us that vendored `bwrap` can behave differently on some Ubuntu AppArmor hosts, so it would be reassuring to know we validated the fallback on the target environments and are not only avoiding the `Unknown option --argv0` failure.

codex-rs/linux-sandbox/README.md
```diff
@@ -8,17 +8,19 @@ This crate is responsible for producing:
   - this should also be true of the `codex` multitool CLI
 
 On Linux, the bubblewrap pipeline prefers the system `/usr/bin/bwrap` whenever
-it is available. If `/usr/bin/bwrap` is missing, the helper still falls back to
+it is available and supports the required argv-rewrite flags. If `/usr/bin/bwrap`
+is missing or too old to support the required flags, the helper falls back to
 the vendored bubblewrap path compiled into this binary.
 Codex also surfaces a startup warning when `/usr/bin/bwrap` is missing so users
 know it is falling back to the vendored helper.
```
evawong-oai: Could we also surface the same startup warning when `/usr/bin/bwrap` is present but too old for `--argv0`, not only when it is missing? I think that would help users understand why Codex is skipping the system package and falling back to the vendored helper.

codex-rs/core/README.md
```diff
@@ -61,9 +61,10 @@ cases like `/repo = write`, `/repo/a = none`, `/repo/a/b = write`, where the
 more specific writable child must reopen under a denied parent.
 
 The Linux sandbox helper prefers `/usr/bin/bwrap` whenever it is available and
-falls back to the vendored bubblewrap path otherwise. When `/usr/bin/bwrap` is
-missing, Codex also surfaces a startup warning through its normal notification
-path instead of printing directly from the sandbox helper.
+supports the required argv-rewrite flags, and falls back to the vendored
+bubblewrap path otherwise. When `/usr/bin/bwrap` is missing, Codex also
```
evawong-oai: Small docs question. Could we mirror the linux sandbox README wording here and say that the fallback is the vendored bubblewrap source compiled into the binary? I think that makes the trust model a bit clearer for readers who land on this README first.

evawong-oai (commented):
The overall direction here makes sense to me, and I appreciate the targeted compatibility fix. I left three small questions focused on validation, user visibility, and docs clarity.

codex-rs/linux-sandbox/README.md
bolinfest: Updated this in the follow-up commit. Codex now emits the startup warning when `/usr/bin/bwrap` is missing or when it exists but is too old to support `--argv0`, and the warning comes through the normal notification path from `codex-core`.

codex-rs/core/README.md
bolinfest: Updated the docs wording as part of the same follow-up. `core/README.md` now matches the Linux sandbox README and explicitly says the fallback is the vendored bubblewrap path compiled into the binary.

codex-rs/core/src/config/mod.rs
```diff
@@ -154,21 +155,46 @@ const RESERVED_MODEL_PROVIDER_IDS: [&str; 3] = [
...
+fn system_bwrap_supports_argv0(system_bwrap_path: &Path) -> bool {
+    // bubblewrap added `--argv0` in v0.9.0:
+    // https://github.com/containers/bubblewrap/releases/tag/v0.9.0
+    let output = match Command::new(system_bwrap_path).arg("--help").output() {
+        Ok(output) => output,
+        Err(_) => return false,
+    };
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    let stderr = String::from_utf8_lossy(&output.stderr);
+    stdout.contains("--argv0") || stderr.contains("--argv0")
```
rreichel3-oai, line 195: Feels like there should be a better way to do this, but I'm not even sure what I'd suggest? I think this is good to merge for now!

codex-rs/linux-sandbox/src/launcher.rs
rreichel3-oai, line 37: Looks like we already do, but just confirming we already vendor bubblewrap? Why not just always use the vendored version for consistency?

rreichel3-oai (approved):

codex-rs/linux-sandbox/src/launcher.rs
bolinfest: @rreichel3-oai good question, but see https://github.com/openai/codex/pull/14963

PR #15342, opened 2026-03-20
Plugins TUI install/uninstall

canvrno-oai:
- Add install/uninstall actions to the TUI plugins menu
- Wire plugin install/uninstall through both TUI and `tui_app_server`
- Refresh config/plugin state after changes so the UI updates immediately
- Add a post-install app setup flow for plugins that require additional app auth

Screenshot 2026-03-20 at 4 08 44 PM
Screenshot 2026-03-20 at 4 08 54 PM
Screenshot 2026-03-20 at 4 09 07 PM
Screenshot 2026-03-20 at 4 09 24 PM

Note/known issue: The /plugin install flow fails in `tui_app_server` because after a successful install it tries to trigger a ReloadUserConfig operation, but `tui_app_server` has not yet implemented transport for that operation, so it falls through to the generic “Not available in app-server TUI yet” stub.

### DIFF
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -45,10 +45,14 @@ use codex_app_server_client::InProcessClientStartArgs;
 use codex_app_server_protocol::ClientRequest;
 use codex_app_server_protocol::ConfigLayerSource;
 use codex_app_server_protocol::ConfigWarningNotification;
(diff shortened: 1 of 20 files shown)


daniel-oai (approved):
I tested this and it worked great. One piece of feedback would be that we should keep the installed plugins at the top of the list in alphabetical form so you can find your installed plugins. However, everything seems to be working really well.

PR #15348, opened 2026-03-20
Pass platform param to featured plugins

alexsong-oai:


codex-rs/core/src/plugins/remote.rs
```diff
@@ -168,6 +168,7 @@ pub async fn fetch_remote_featured_plugin_ids(
     let client = build_reqwest_client();
     let mut request = client
         .get(&url)
+        .query(&[("platform", "codex")])
```
xl-openai: better read from "product".

xl-openai (approved):

PR #15408, opened 2026-03-21
Update config handling to rely on async ExecutorFileSystem

pakrym-oai:
Summary
- Align config-related changes with the requested async ExecutorFileSystem loading flow
- Ensure no sync ::fs:: reads remain in core/config outside tests

Testing
- Not run (not requested)


codex-rs/core/src/config/mod.rs
```diff
@@ -901,8 +917,45 @@ pub(crate) fn deserialize_config_toml_with_base(
         .map_err(|e| std::io::Error::new(std::io::ErrorKind::InvalidData, e))
 }
 
-fn load_catalog_json(path: &AbsolutePathBuf) -> std::io::Result<ModelsResponse> {
-    let file_contents = std::fs::read_to_string(path)?;
+fn absolute_path_for_filesystem_read(path: &Path) -> std::io::Result<AbsolutePathBuf> {
+    let absolute_path = if path.is_absolute() {
```
bolinfest, line 921: I believe `is_absolute()` is platform-dependent.

codex-rs/core/src/config/mod.rs
```diff
@@ -901,8 +917,45 @@ pub(crate) fn deserialize_config_toml_with_base(
...
+            ErrorKind::InvalidData,
+            format!(
+                "failed to decode UTF-8 from {}: {err}",
+                absolute_path.display()
+            ),
+        )
+    })
+}
+
+async fn read_text_file(
```
bolinfest, line 947: Why is there a helper for this? Shouldn't `ExecutorFileSystem` provide this directly?

codex-rs/core/src/config/edit.rs
```diff
@@ -743,27 +744,33 @@ pub fn apply_blocking(
         return Ok(());
     }
 
-    write_atomically(&write_paths.write_path, &document.doc.to_string()).with_context(|| {
```
bolinfest, line 746: So we lose atomic updates here, correct?

PR #15424, opened 2026-03-21
Finish moving codex exec to app-server

etraut-openai:
This PR completes the conversion of non-interactive `codex exec` to use app server rather than directly using core events and methods.

### Summary
- move `codex-exec` off exec-owned `AuthManager` and `ThreadManager` state
- route exec bootstrap, resume, and auth refresh through existing app-server paths
- replace legacy `codex/event/*` decoding in exec with typed app-server notification handling
- update human and JSONL exec output adapters to translate existing app-server notifications only
- clean up "app server client" layer by eliminating support for legacy notifications; this is no longer needed
- remove exposure of `authManager` and `threadManager` from "app server client" layer

### Testing
- `exec` has pretty extensive unit and integration tests already, and these all pass
- In addition, I asked Codex to put together a comprehensive manual set of tests to cover all of the `codex exec` functionality (including command-line options), and it successfully generated and ran these tests


codex-rs/app-server/src/transport.rs
```diff
@@ -584,13 +578,10 @@ fn should_skip_notification_for_connection(
     connection_state: &OutboundConnectionState,
     message: &OutgoingMessage,
 ) -> bool {
-    if !connection_state.allow_legacy_notifications
-        && matches!(message, OutgoingMessage::Notification(_))
-    {
+    if matches!(message, OutgoingMessage::Notification(_)) {
```
bolinfest: So will this go away in a follow-up?

codex-rs/app-server/src/transport.rs
```diff
@@ -1008,7 +998,6 @@ mod tests {
                 Arc::new(AtomicBool::new(true)),
                 Arc::new(AtomicBool::new(true)),
                 Arc::new(RwLock::new(HashSet::new())),
-                false,
```
bolinfest, line 1011: Curious why this doesn't have a comment given `argument-comment-lint`.

codex-rs/exec/src/event_processor_with_human_output.rs
```diff
@@ -1,1160 +1,424 @@
...
-    progress_done: bool,
+    final_message_rendered: bool,
+    last_total_token_usage: Option<ThreadTokenUsage>,
 }
 
 impl EventProcessorWithHumanOutput {
     pub(crate) fn create_with_ansi(
         with_ansi: bool,
-        cursor_ansi: bool,
+        _cursor_ansi: bool,
```
bolinfest: Why is this no longer honored?

codex-rs/exec/src/event_processor_with_human_output.rs
```diff
@@ -1,1160 +1,424 @@
...
-            "session id",
-            session_configured_event.session_id.to_string(),
-        ));
-
-        for (key, value) in entries {
-            eprintln!("{} {}", format!("{key}:").style(self.bold), value);
+    fn print_usage(&self) {
+        if let Some(usage) = &self.last_total_token_usage {
+            eprintln!(
+                "{} input={} cached={} output={}",
```
bolinfest: It's hard for me to tell because there is a lot of movement in this PR, but I'm pretty sure some users rely on how token usage is displayed (and parse whatever we currently do) though we should have a stable event printed to stdout if we don't already...

codex-rs/exec/src/event_processor_with_human_output.rs
```diff
@@ -1,1160 +1,424 @@
...
-    /// for the session. This mirrors the information shown in the TUI welcome
-    /// screen.
-    fn print_config_summary(
-        &mut self,
-        config: &Config,
-        prompt: &str,
-        session_configured_event: &SessionConfiguredEvent,
-    ) {
-        const VERSION: &str = env!("CARGO_PKG_VERSION");
-        ts_msg!(
```
bolinfest, line 188: The point of `ts_msg` was to include a timestamp: this seems to be lost?

codex-rs/exec/src/lib.rs
```diff
@@ -1067,71 +989,198 @@ fn session_configured_from_thread_response(
...
+async fn latest_thread_cwd(thread: &AppServerThread) -> PathBuf {
+    if let Some(path) = thread.path.as_deref()
+        && let Some(cwd) = parse_latest_turn_context_cwd(path).await
+    {
+        return cwd;
+    }
+    thread.cwd.clone()
+}
+
+async fn parse_latest_turn_context_cwd(path: &Path) -> Option<PathBuf> {
```
bolinfest, line 1113: We don't have an existing library call for this?

bolinfest (approved):

codex-rs/exec/src/event_processor_with_human_output.rs
etraut-openai: This parameter is no longer needed because it was tied to the single-line progress display for background progress messages. The app server event stream doesn't provide the details required for this feature. My plan is to drop it and see if we get feedback from users. If necessary, we could add the necessary plumbing in app server to support it, but my best guess is that this feature isn't used.

codex-rs/exec/src/event_processor_with_human_output.rs
etraut-openai: Good point. This introduced an unnecessary change in the formatting. I've updated to remain consistent with the output formatting that exists prior to this PR.

codex-rs/exec/src/event_processor_with_human_output.rs
etraut-openai: Despite the name and comment associated with ts_msg, it has never emitted a timestamp.

codex-rs/exec/src/lib.rs
etraut-openai: Not exactly. There's opportunity for some refactoring to share some logic here, but there's no library call that does this exactly.

codex-rs/exec/src/lib.rs
```diff
@@ -1320,6 +1306,15 @@ async fn handle_server_request(
             )
             .await
         }
+        ServerRequest::ChatgptAuthTokensRefresh { request_id, .. } => {
+            reject_server_request(
+                client,
+                request_id,
+                &method,
+                "chatgpt auth token refresh is not supported in exec mode".to_string(),
+            )
```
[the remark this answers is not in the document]
etraut-openai: Yes, this is expected.

codex-rs/exec/src/lib.rs
```diff
@@ -1067,71 +972,198 @@ fn session_configured_from_thread_response(
...
+        ServerNotification::TurnDiffUpdated(notification) => {
+            notification.thread_id == thread_id && notification.turn_id == turn_id
+        }
+        ServerNotification::TurnPlanUpdated(notification) => {
+            notification.thread_id == thread_id && notification.turn_id == turn_id
+        }
+        ServerNotification::TurnStarted(notification) => {
+            notification.thread_id == thread_id && notification.turn.id == turn_id
+        }
+        _ => false,
```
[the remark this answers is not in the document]
etraut-openai: This is not dropping functionality that was previously provided by exec, so this isn't a regression.

PR #15426, opened 2026-03-21
chore(core) Add approvals reviewer to UserTurn

dylan-hurd-oai:
## Summary
Adds support for approvals_reviewer to `Op::UserTurn` so we can migrate `[CodexMessageProcessor::turn_start]` to use Op::UserTurn

## Testing
- [x] Adds quick test for the new field


codex-rs/protocol/src/protocol.rs
```diff
@@ -254,6 +254,11 @@ pub enum Op {
         /// Policy to use for command approval.
         approval_policy: AskForApproval,
 
+        /// Reviewer to use for approval requests raised during this turn.
+        ///
+        /// When omitted, the session keeps the current setting
```
charley-oai, line 259: what if it's the first turn?

charley-oai (approved):
Thanks

codex-rs/protocol/src/protocol.rs
dylan-hurd-oai: Would just be the session default

PR #15465, opened 2026-03-22
CodeMode: Manage turn worker shutdown

cconger:
We only want one turn worker processing at a time, this codifies the shutdown of one before resolving the turn.

Drop still does best effort shutdown.

I recommend viewing this one omitting whitespace: https://github.com/openai/codex/pull/15465/changes?w=1

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -6182,7 +6182,7 @@ async fn run_sampling_request(
         Arc::clone(&turn_context),
         Arc::clone(&turn_diff_tracker),
     );
-    let _code_mode_worker = sess
+    let code_mode_worker = sess
         .services
         .code_mode_service
         .start_turn_worker(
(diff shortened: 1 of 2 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -6192,93 +6192,98 @@ async fn run_sampling_request(
...
+            } else {
+                break Err(err);
             }
-            tokio::time::sleep(delay).await;
-        } else {
-            return Err(err);
         }
     }
+    .await;
+    if let Some(code_mode_worker) = code_mode_worker {
```
pakrym-oai, line 6283: why don't we shutdown on drop?

pakrym-oai: ah, it's because it's async.

pakrym-oai: I just don't like making this already complicated method even deeper.

PR #15439, opened 2026-03-22
Attach network approval reviews to parent tool items

charley-oai:
## Summary
- carry an optional parent tool item id on `GuardianApprovalRequest::NetworkAccess`
- use that parent id when app-server populates guardian review `targetItemId` for attributed `network_access` reviews
- keep the existing fallback to the guardian review id when no parent tool item can be determined
- add focused tests and update the app-server README

## Testing
- `just fmt`
- started focused `cargo test -p codex-core guardian_request_turn_id_prefers_network_access_owner_turn` verification on a cold cache; stopped before completion while opening this PR

PR #15469, opened 2026-03-22
Type parent item ids for network approval reviews

charley-oai:
## Summary
- add typed `reviewId` and `parentToolItemId` fields to the legacy app-server guardian review notifications
- thread `parent_tool_item_id` through `GuardianAssessmentEvent` so app-server clients do not need to recover parent linkage from untyped `action` JSON
- always populate `parentToolItemId` for direct guardian-reviewed tool actions, where it matches `reviewId`
- keep `reviewId` distinct from `parentToolItemId` for attributed `network_access` reviews, so concurrent network reviews stay uniquely identifiable
- keep `targetItemId` as a deprecated alias for `reviewId` for backwards compatibility

## Notes
- This makes the parent tool item relationship explicit in the protocol instead of relying on notification ordering.
- `parentToolItemId` can still be absent when a `network_access` review cannot be attributed to a parent tool item.

Supersedes #15439.


codex-rs/core/tests/suite/shell_snapshot.rs
```diff
@@ -528,7 +528,7 @@ async fn shell_command_snapshot_still_intercepts_apply_patch() -> Result<()> {
     let script = "apply_patch <<'EOF'\n*** Begin Patch\n*** Add File: snapshot-apply.txt\n+hello from snapshot\n*** End Patch\nEOF\n";
     let args = json!({
         "command": script,
-        "timeout_ms": 1_000,
+        "timeout_ms": 3_000,
```
charley-oai, line 531: CI was flaking on me but I can pull this into a separate PR if you prefer

PR #15478, opened 2026-03-23
[codex-cli][app-server] Update self-serve business usage limit copy in error returned

dhruvgupta-oai:
## Summary
- update the self-serve business usage-based limit message to direct users to their admin for additional credits
- add a focused unit test for the self_serve_business_usage_based plan branch

Added also: 

If you are at a rate limit but you still have credits, codex cli would tell you to switch the model. We shouldnt do this if you have credits so fixed this.

## Test
- launched the source-built CLI and verified the updated message is shown for the self-serve business usage-based plan

Test screenshot


codex-rs/core/src/error.rs
```diff
@@ -460,6 +460,12 @@ impl std::fmt::Display for UsageLimitReachedError {
                 "You've hit your usage limit.{}",
                 retry_suffix(self.resets_at.as_ref())
             ),
+            Some(PlanType::Unknown(plan))
+                if plan.eq_ignore_ascii_case("self_serve_business_usage_based") =>
+            {
+                "You've hit your usage limit. Contact your admin to add credits to continue."
```
bwanner-oai: Do we want to differentiate out of credits and spend controls or are we fine to roll it into one message?

codex-rs/core/src/error.rs
```diff
@@ -460,6 +460,12 @@ impl std::fmt::Display for UsageLimitReachedError {
                 "You've hit your usage limit.{}",
                 retry_suffix(self.resets_at.as_ref())
             ),
+            Some(PlanType::Unknown(plan))
```
bwanner-oai, line 463: What's the cost of adding another `PlanType` and just add the handling for it instead of doing the string match?

bwanner-oai (approved):

codex-rs/core/src/error.rs
dhruvgupta-oai: let me see...this in rust and im not familar so I wanted to make as little changes as possible

dhruvgupta-oai: turns out: very costly a lot of places code has to change. Not comfortable with that so making the smallest change possible

PR #15530, opened 2026-03-23
chore: stop app-server auth refresh storms after permanent token failure

celia-oai:
built from #14256. PR description from @etraut-openai:

This PR addresses a hole in [PR 11802](https://github.com/openai/codex/pull/11802). The previous PR assumed that app server clients would respond to token refresh failures by presenting the user with an error ("you must log in again") and then not making further attempts to call network endpoints using the expired token. While they do present the user with this error, they don't prevent further attempts to call network endpoints and can repeatedly call `getAuthStatus(refreshToken=true)` resulting in many failed calls to the token refresh endpoint.

There are three solutions I considered here:
1. Change the getAuthStatus app server call to return a null auth if the caller specified "refreshToken" on input and the refresh attempt fails. This will cause clients to immediately log out the user and return them to the log in screen. This is a really bad user experience. It's also a breaking change in the app server contract that could break third-party clients.
2. Augment the getAuthStatus app server call to return an additional field that indicates the state of "token could not be refreshed". This is a non-breaking change to the app server API, but it requires non-trivial changes for all clients to properly handle this new field properly.
3. Change the getAuthStatus implementation to handle the case where a token refresh fails by marking the AuthManager's in-memory access and refresh tokens as "poisoned" so it they are no longer used. This is the simplest fix that requires no client changes.

I chose option 3.

Here's Codex's explanation of this change:

When an app-server client asks `getAuthStatus(refreshToken=true)`, we may try to refresh a stale ChatGPT access token. If that refresh fails permanently (for example `refresh_token_reused`, expired, or revoked), the old behavior was bad in two ways:

1. We kept the in-memory auth snapshot alive as if it were still usable.
2. Later auth checks could retry refresh again and again, creating a storm of doomed `/oauth/token` requests and repeatedly surfacing the same failure.

This is especially painful for app-server clients because they poll auth status and can keep driving the refresh path without any real chance of recovery.

This change makes permanent refresh failures terminal for the current managed auth snapshot without changing the app-server API contract.

What changed:
- `AuthManager` now poisons the current managed auth snapshot in memory after a permanent refresh failure, keyed to the unchanged `AuthDotJson`.
- Once poisoned, later refresh attempts for that same snapshot fail fast locally without calling the auth service again.
- The poison is cleared automatically when auth materially changes, such as a new login, logout, or reload of different auth state from storage.
- `getAuthStatus(includeToken=true)` now omits `authToken` after a permanent refresh failure instead of handing out the stale cached bearer token.

This keeps the current auth method visible to clients, avoids forcing an immediate logout flow, and stops repeated refresh attempts for credentials that cannot recover.


codex-rs/login/src/auth/manager.rs
```diff
@@ -1285,7 +1346,12 @@ impl AuthManager {
             Some(auth) => auth,
             None => return Ok(()),
         };
-        match auth {
+        if let Some(error) = self.permanent_refresh_failure_for_auth(&auth) {
```
pakrym-oai: isn't the parameter to permanent_refresh_failure_for_auth always auth_cached?

pakrym-oai: that is later compared to auth_cached again?

celia-oai: that's a good point - seems like a redundant check and just simplified the logic here

codex-rs/login/src/auth/manager.rs
```diff
@@ -796,6 +796,15 @@ struct CachedAuth {
     auth: Option<CodexAuth>,
     /// Callback used to refresh external auth by asking the parent app for new tokens.
     external_refresher: Option<Arc<dyn ExternalAuthRefresher>>,
+    /// Permanent refresh failure cached for the current auth snapshot so
+    /// later refresh attempts for the same credentials fail fast without network.
+    permanent_refresh_failure: Option<AuthScopedRefreshFailure>,
+}
+
+#[derive(Clone)]
+struct AuthScopedRefreshFailure {
```
pakrym-oai, line 805: Should we store this as the value of `auth: Option<CodexAuth>,` ?

codex-rs/login/src/auth/manager.rs
```diff
@@ -1166,6 +1195,25 @@ impl AuthManager {
...
 
+    /// Records a permanent refresh failure only if the failed refresh was
+    /// attempted against the auth snapshot that is still cached.
+    fn record_permanent_refresh_failure_if_unchanged(
+        &self,
+        attempted_auth: &CodexAuth,
+        error: &RefreshTokenFailedError,
+    ) {
+        if let Ok(mut guard) = self.inner.write() {
+            let current_auth_matches =
```
pakrym-oai, line 1206: do you need this condition? The failure encodes the auth and it's compared on read.

codex-rs/login/src/auth/manager.rs
```diff
@@ -1180,6 +1228,11 @@ impl AuthManager {
         if let Ok(mut guard) = self.inner.write() {
             let previous = guard.auth.as_ref();
             let changed = !AuthManager::auths_equal(previous, new_auth.as_ref());
+            let auth_changed_for_refresh =
```
pakrym-oai, line 1231: Storing failure along the main auth might naturally perform the cleaning.

pakrym-oai (approved):

codex-rs/login/src/auth/manager.rs
celia-oai: CodexAuth is strictly identity related so hopefully we leave it as-is

codex-rs/login/src/auth/manager.rs
celia-oai: this is still useful, just in case auth actually gets reloaded we don't record a false positive of permanent error

codex-rs/login/src/auth/manager.rs
celia-oai: Codex raises a good point: permanent refresh failure is manager-owned state, not part of the auth or identity model itself. That makes the manager the right place to track it. Pushing it down into the auth layer is not a clean fit, because that layer has a broader surface area and more code paths that would need to preserve and propagate the state correctly.

PR #15531, opened 2026-03-23
[hooks] add non-streaming (non-stdin style) shell-only PostToolUse support

eternal-openai:
CHAINED PR - note that base is eternal/hooks-pretooluse-bash, not main -- so the following PR should be first

Matching post-tool hook to the pre-tool functionality here: https://github.com/openai/codex/pull/15211

So, PreToolUse calls for plain shell calls, allows blocking. This PostToolUse call runs after the command executed

example run:

```
› as a test, run in parallel the following commands:
  - echo 'one'
  - echo '[block-pre-tool-use]'
  - echo '[block-post-tool-use]'


⚠ MCP startup incomplete (failed: notion, linear)

• Cruising through those three commands in parallel now, and I’ll share the exact outputs right after
  they land.

• Running PreToolUse hook: checking the observatory runes

• Running PreToolUse hook: checking the observatory runes

• Running PreToolUse hook: checking the observatory runes

PreToolUse hook (blocked)
  warning: wizard-tower PreToolUse demo blocked a Bash command on purpose.
  feedback: PreToolUse demo blocked the command. Remove [block-pre-tool-use] to continue.

PreToolUse hook (completed)
  warning: wizard-tower PreToolUse demo inspected Bash: echo 'one'

PreToolUse hook (completed)
  warning: wizard-tower PreToolUse demo inspected Bash: echo '[block-post-tool-use]'

• Ran echo '[block-post-tool-use]'
  └ [block-post-tool-use]

• Ran echo 'one'
  └ one

• Running PostToolUse hook: reading the observatory aftermath

• Running PostToolUse hook: reading the observatory aftermath

PostToolUse hook (blocked)
  warning: wizard-tower PostToolUse demo blocked the Bash follow-through on purpose.
  feedback: PostToolUse demo blocked the result after execution. Remove [block-post-tool-use] to
continue.

PostToolUse hook (completed)
  warning: wizard-tower PostToolUse demo inspected Bash: echo 'one'
  hook context: PostToolUse demo saw the Bash result. Response preview: one

──────────────────────────────────────────────────────────────────────────────────────────────────────

• Ran them in parallel. Little wave report:

  - echo 'one' completed and returned one.
  - echo '[block-pre-tool-use]' was blocked before execution by the PreToolUse hook.
  - echo '[block-post-tool-use]' ran, but its result was blocked afterward by the PostToolUse hook.

  The hook messages were:

  - PreToolUse demo blocked the command. Remove [block-pre-tool-use] to continue.
  - PostToolUse demo blocked the result after execution. Remove [block-post-tool-use] to continue.


› for the blocked messages (correctly blocked in this test) did the tool output land back into
  context?


• Yeah, from this run:

  - PreToolUse block: the command itself did not execute, so no command output landed in context. Only
    the block message did.
  - PostToolUse block: the command did execute, but the actual stdout did not come back into context.
    Only the post-tool block message did.

  So the blocked notifications landed in context, but the blocked command payloads themselves did not.
  Nice clean guardrail, aloha.
```


codex-rs/core/src/tools/registry.rs
```diff
@@ -303,6 +337,28 @@ impl ToolRegistry {
...
+                    invocation.call_id.clone(),
+                    post_tool_use_payload.command,
+                    post_tool_use_payload.tool_response,
+                )
+                .await,
+            )
+        } else {
+            None
+        };
         let hook_abort_error = dispatch_after_tool_use_hook(AfterToolUseHookDispatch {
```
pakrym-oai, line 382: Are `post tool use` and `after tool use` different hooks?

codex-rs/core/src/tools/registry.rs
```diff
@@ -317,6 +373,38 @@ impl ToolRegistry {
...
+        if let Some(outcome) = &post_tool_use_outcome {
+            record_additional_contexts(
+                &invocation.session,
+                &invocation.turn,
+                outcome.additional_contexts.clone(),
+            )
+            .await;
+
+            if outcome.should_stop {
+                let replacement_text = outcome
```
pakrym-oai: both branches of this if look very similar

codex-rs/core/src/tools/registry.rs
```diff
@@ -317,6 +373,38 @@ impl ToolRegistry {
...
+                    .feedback_message
+                    .clone()
+                    .or_else(|| outcome.stop_reason.clone())
+                    .unwrap_or_else(|| "PostToolUse hook stopped execution".to_string());
+                let mut guard = response_cell.lock().await;
+                if let Some(result) = guard.as_mut() {
+                    result.result = Box::new(FunctionToolOutput::from_text(
```
pakrym-oai, line 418: what if the tool is CustomTool and needs custom response. 
+ Why are we highjacking the tool result? 

Shouldn't the model still get the real tool result?

codex-rs/core/src/tools/context.rs
```diff
@@ -84,6 +84,10 @@ pub trait ToolOutput: Send {
 
     fn to_response_item(&self, call_id: &str, payload: &ToolPayload) -> ResponseInputItem;
 
+    fn post_tool_use_response(&self, _call_id: &str, _payload: &ToolPayload) -> Option<JsonValue> {
```
pakrym-oai, line 87: what should this property contain for a notmal tool output?

codex-rs/core/src/tools/handlers/unified_exec.rs
```diff
@@ -117,6 +156,14 @@ impl ToolHandler for UnifiedExecHandler {
         !is_known_safe_command(&command)
     }
 
+    fn pre_tool_use_payload(&self, invocation: &ToolInvocation) -> Option<PreToolUsePayload> {
+        exec_command_pre_tool_use_payload(&invocation.tool_name, &invocation.payload)
+    }
+
+    fn post_tool_use_payload(&self, result: &AnyToolResult) -> Option<PostToolUsePayload> {
```
pakrym-oai: AnyToolResult is what ToolResult is wrapped into at higher levels of execution to arase the concrete handler output type, it's a bit circular to pass it back in.

codex-rs/core/src/tools/handlers/unified_exec.rs
```diff
@@ -81,6 +84,42 @@ fn default_tty() -> bool {
     false
 }
 
+fn exec_command_pre_tool_use_payload(
+    tool_name: &str,
+    payload: &ToolPayload,
+) -> Option<PreToolUsePayload> {
+    if tool_name != "exec_command" {
```
pakrym-oai: can this be shared with normal arg parsing?

codex-rs/core/src/tools/handlers/shell.rs
```diff
@@ -48,6 +52,51 @@ pub struct ShellCommandHandler {
...
+fn shell_pre_tool_use_payload(payload: &ToolPayload) -> Option<PreToolUsePayload> {
+    shell_payload_command(payload).map(|command| PreToolUsePayload { command })
+}
+
+fn shell_command_payload_command(payload: &ToolPayload) -> Option<String> {
+    let ToolPayload::Function { arguments } = payload else {
+        return None;
+    };
+
+    serde_json::from_str::<ShellCommandToolCallParams>(arguments)
```
pakrym-oai: is parse_arguments appropriate here?

codex-rs/core/src/tools/handlers/shell.rs
```diff
@@ -492,8 +557,25 @@ impl ShellHandler {
             &call_id,
             /*turn_diff_tracker*/ None,
         );
-        let content = emitter.finish(event_ctx, out).await?;
-        Ok(FunctionToolOutput::from_text(content, Some(true)))
+        let (content, post_tool_use_response) = match out {
+            Ok(output) => {
+                let post_tool_use_response =
+                    crate::tools::format_exec_output_str(&output, turn.truncation_policy);
+                let content = emitter.finish(event_ctx, Ok(output)).await?;
```
pakrym-oai: can                 let content = emitter.finish(event_ctx, Err(error)).await?;
line be pulled out of the if and only `post_tool_use_response` calculated separately?

pakrym-oai: why don't we use content directly?

codex-rs/core/src/tools/registry.rs
eternal-openai: Yeah, this was introduced by @gt-oai in the old version. I'm maintaining support for now, but it'll get deprecated/removed later

codex-rs/core/src/tools/registry.rs
eternal-openai: Fixing

codex-rs/core/src/tools/registry.rs
eternal-openai: CustomTool support out of scope for this PR, I'm trying to just handle "Bash" (i.e. the name for shell-like stuff in the spec)

Killing the tool result is intentional, if the hook "blocks" the output, it basically has the opportunity to rewrite it with its own context

codex-rs/core/src/tools/context.rs
eternal-openai: By default it should be None. A tool only populates post_tool_use_response if it intentionally supports a stable PostToolUse payload shape

codex-rs/core/src/tools/handlers/unified_exec.rs
eternal-openai: Agreed. I cleaned this up so handlers no longer receive the AnyToolResult just to shape the hook payload. They now get call_id, payload, and &dyn ToolOutput, which keeps the boundary better

Small note, the spec is pretty bad around tool_response (and tool_input): "...The exact schema for both depends on the tool." ... with no further elaboration :)

So basically we have to handle "any" and I've been reading how this works in the wild as the source of truth ¯\_(ツ)_/¯

codex-rs/core/src/tools/handlers/unified_exec.rs
eternal-openai: Yep, agreed, fixed

codex-rs/core/src/tools/handlers/shell.rs
eternal-openai: I think so, yes, fixed

codex-rs/core/src/tools/handlers/shell.rs
eternal-openai: Agree, simplified. emitter.finish now sits outside the branch, and only the post_tool_use_response calculation stays conditional

---

I kept that separate on purpose. content seems like more model-facing rendered shell text, which is more transcript/UI-oriented. tool_response for PostToolUse is meant to be the stable tool-specific result payload, so I didn’t want to couple the hook input to the transcript formatting

codex-rs/core/src/tools/registry.rs
```diff
@@ -303,6 +352,33 @@ impl ToolRegistry {
             Err(err) => (err.to_string(), false),
         };
         emit_metric_for_tool_read(&invocation, success).await;
+        let post_tool_use_payload = if success {
```
pakrym-oai, line 355: for the future it would be nice to offload hook processing into it's own routine. non blocking

codex-rs/core/src/tools/registry.rs
```diff
@@ -431,33 +534,8 @@ fn sandbox_policy_tag(policy: &SandboxPolicy) -> &'static str {
...
-                .map(|params| params.command)
-        }
-        ("exec_command", ToolPayload::Function { arguments }) => {
-            serde_json::from_str::<PreToolUseExecCommandArgs>(arguments)
-                .ok()
-                .map(|params| params.cmd)
-        }
-        _ => None,
-    }
+fn replace_result_with_feedback(result: &mut AnyToolResult, text: String) {
```
pakrym-oai: not worth a helper.

codex-rs/core/src/tools/handlers/shell.rs
```diff
@@ -178,6 +226,19 @@ impl ToolHandler for ShellHandler {
         }
     }
 
+    fn pre_tool_use_payload(&self, invocation: &ToolInvocation) -> Option<PreToolUsePayload> {
+        shell_pre_tool_use_payload(&invocation.payload)
```
pakrym-oai: why don't we inline these methods?

pakrym-oai: goes for all of them

codex-rs/core/src/tools/handlers/unified_exec.rs
```diff
@@ -81,6 +84,44 @@ fn default_tty() -> bool {
...
+    payload: &ToolPayload,
+    result: &dyn ToolOutput,
+) -> Option<PostToolUsePayload> {
+    let ToolPayload::Function { arguments } = payload else {
+        return None;
+    };
+
+    let args = parse_arguments::<ExecCommandArgs>(arguments).ok()?;
+    if args.tty {
+        return None;
```
pakrym-oai: why?

pakrym-oai (approved):

codex-rs/core/src/tools/handlers/unified_exec.rs
```diff
@@ -81,6 +84,44 @@ fn default_tty() -> bool {
...
+    let ToolPayload::Function { arguments } = payload else {
+        return None;
+    };
+
+    parse_arguments::<ExecCommandArgs>(arguments)
+        .ok()
+        .map(|args| PreToolUsePayload { command: args.cmd })
+}
+
+fn exec_command_post_tool_use_payload(
```
pakrym-oai: inline?

codex-rs/core/src/tools/handlers/unified_exec.rs
eternal-openai: Was just playing it safe for now, I didn't handle streaming/interactive shell usage in this PR, and that mirrors the PreToolUse one. Definitely needs a followup, but it doesn't fit cleanly into the spec, so I've been thinking about what to do about it (maybe a spec extension)

codex-rs/core/src/tools/registry.rs
eternal-openai: Ok, noted for a future followup

PR #15545, opened 2026-03-23
chore:  use access token expiration for proactive auth refresh

celia-oai:
Follow up to #15357 by making proactive ChatGPT auth refresh depend on the access token's JWT expiration instead of treating `last_refresh` age as the primary source of truth.


codex-rs/login/src/auth/manager.rs
```diff
@@ -1333,6 +1333,11 @@ impl AuthManager {
             Some(auth_dot_json) => auth_dot_json,
             None => return false,
         };
+        if let Some(tokens) = auth_dot_json.tokens.as_ref()
+            && let Ok(Some(expires_at)) = parse_jwt_expiration(&tokens.access_token)
+        {
+            return expires_at <= Utc::now();
```
pakrym-oai, line 1339: do we want a buffer? TOKEN_REFRESH_INTERVAL?

celia-oai: why do we need a buffer? `expired_at` is a timestamp? we still have the fallback code path that uses `Token_refresh_interval` after this in case `exp` field doesn't exist in token?

codex-rs/login/src/token_data.rs
```diff
@@ -117,6 +120,12 @@ struct AuthClaims {
     chatgpt_account_id: Option<String>,
 }
 
+#[derive(Deserialize)]
+struct StandardJwtClaims {
```
pakrym-oai, line 124: why a separate struct? can we put this onto AuthClaims?

pakrym-oai: or IdClaims ?

codex-rs/login/src/auth/manager.rs
pakrym-oai: Do we still need that codepath, can "exp" not be there?

pakrym-oai (approved):

codex-rs/login/src/token_data.rs
celia-oai: this function is supposed to be generic so it can parse any jwt, not just id / access tokens I think

codex-rs/login/src/auth/manager.rs
celia-oai: technically, jwt token doesn't necessarily have the 'exp' field. In reality, the  auth token returned should always have this field. We can also throw an error here if exp doesn't exist, but I think having this silent fallback is safer

PR #15547, opened 2026-03-23
app-server: Add back pressure and batching to `command/exec`

euroelessar:
* Add `OutgoingMessageSender::send_server_notification_to_connection_and_wait` which returns only once message is written to websocket (or failed to do so)
* Use this mechanism to apply back pressure to stdout/stderr streams of processes spawned by `command/exec`, to limit them to at most one message in-memory at a time
* Use back pressure signal to also batch smaller chunks into ≈64KiB ones

This should make commands execution more robust over high-latency/low-throughput networks


codex-rs/app-server/src/outgoing_message.rs
```diff
@@ -527,6 +546,28 @@ impl OutgoingMessageSender {
         }
     }
 
+    pub(crate) async fn send_server_notification_to_connection_and_wait(
+        &self,
+        connection_id: ConnectionId,
+        notification: ServerNotification,
+    ) {
+        tracing::trace!("app-server event: {notification}");
```
owenlin0, line 554: nit: if this was just for temporary debugging, can we remove? IIRC we've had issues with noisy trace logs in the past

euroelessar: it's replicated from existing `send_server_notification_to_connections`, should I remove it from both?

euroelessar: was added by #12695

owenlin0: ah ok, fine to keep then. seems deliberate

owenlin0 (approved):

PR #15548, opened 2026-03-23
Extract rollout into its own crate

aibrahim-oai:
- move rollout recorder, listing, metadata, policy, session index, and rollout-owned DB code into `codex-rollout`
- keep `codex-core` behavior intact while renaming the leftover core-only files so the move is obvious in GitHub diffs


codex-rs/core/src/lib.rs
```diff
@@ -122,11 +122,14 @@ pub mod project_doc;
 mod rollout;
 pub(crate) mod safety;
 pub mod seatbelt;
+mod session_rollout_init_error;
 pub mod shell;
 pub mod shell_snapshot;
 pub mod skills;
 pub mod spawn;
+#[path = "state_db_bridge.rs"]
```
bolinfest: Can you do a separate rename in a follow-up (should be a mechanical change) so this unusual use of `#[path]` doesn't stick around forever?

codex-rs/core/src/rollout.rs
```diff
@@ -0,0 +1,64 @@
+use crate::config::Config;
+
+impl codex_rollout::RolloutConfigView for Config {
```
bolinfest, line 17: Why is this listed before the imports?

codex-rs/core/src/rollout.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn codex_home(&self) -> &std::path::Path {
```
bolinfest, line 18: Separately, we should be moving all of these things to `AbsolutePathBuf`.

codex-rs/core/src/rollout.rs
```diff
@@ -0,0 +1,64 @@
+use crate::config::Config;
```
bolinfest, line 1: Relatedly, will all these re-exports get removed in a mechanical follow-up PR?

codex-rs/rollout/src/lib.rs
```diff
@@ -1,9 +1,27 @@
-//! Rollout module: persistence and discovery of session rollout files.
+//! Rollout persistence and discovery for Codex session files.
 
 use std::sync::LazyLock;
 
 use codex_protocol::protocol::SessionSource;
 
+pub mod config;
+#[path = "state_db.rs"]
```
bolinfest: Here too.

codex-rs/rollout/src/path_utils.rs
```diff
@@ -0,0 +1,97 @@
...
+pub fn normalize_for_path_comparison(path: impl AsRef<Path>) -> std::io::Result<PathBuf> {
+    let canonical = path.as_ref().canonicalize()?;
+    Ok(normalize_for_wsl(canonical))
+}
+
+fn normalize_for_wsl(path: PathBuf) -> PathBuf {
+    normalize_for_wsl_with_flag(path, is_wsl())
+}
+
+fn is_wsl() -> bool {
```
bolinfest: This feels like a weird place for some of this stuff to live. How is it rollout-related?

codex-rs/rollout/src/path_utils.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    if !is_wsl_case_insensitive_path(&path) {
+        return path;
+    }
+
+    lower_ascii_path(path)
+}
+
+fn is_wsl_case_insensitive_path(path: &Path) -> bool {
```
bolinfest: docstring?

codex-rs/rollout/src/path_utils.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+}
+
+#[cfg(target_os = "linux")]
+fn ascii_eq_ignore_case(left: &[u8], right: &[u8]) -> bool {
+    left.len() == right.len()
+        && left
+            .iter()
+            .zip(right)
+            .all(|(lhs, rhs)| lhs.to_ascii_lowercase() == *rhs)
```
bolinfest: shouldn't `rhs.to_ascii_lowercase()` also?

codex-rs/rollout/src/path_utils.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+#[cfg(target_os = "linux")]
+fn lower_ascii_path(path: PathBuf) -> PathBuf {
```
bolinfest: What is all this? This does not seem right.

codex-rs/rollout/src/path_utils.rs
```diff
@@ -0,0 +1,97 @@
...
+    use std::ffi::OsString;
+    use std::os::unix::ffi::OsStrExt;
+    use std::os::unix::ffi::OsStringExt;
+
+    let bytes = path.as_os_str().as_bytes();
```
bolinfest: An `OsString` is not guaranteed to be UTF-8: we don't know what the encoding is. As such, I don't think this transform is guaranteed to be safe/valid.

codex-rs/rollout/src/policy.rs
```diff
@@ -12,7 +12,7 @@ pub enum EventPersistenceMode {
 /// Whether a rollout `item` should be persisted in rollout files for the
 /// provided persistence `mode`.
 #[inline]
```
bolinfest: I don't think we should be using `#[inline]` btw: we should let the compiler do its job.

codex-rs/core/src/rollout.rs
aibrahim-oai: yes

codex-rs/rollout/src/path_utils.rs
aibrahim-oai: makes sense

bolinfest (approved):

PR #15549, opened 2026-03-23
Allow global network allowlist wildcard

rreichel3-oai:
## Problem

Today `codex-network-proxy` rejects a global `*` in `network.allowed_domains`, so there is no static way to configure a denylist-only posture for public hosts. Users have to enumerate broad allowlist patterns instead.

## Approach

- Make global wildcard acceptance field-specific: `allowed_domains` can use `*`, while `denied_domains` still rejects a global wildcard.
- Keep the existing evaluation order, so explicit denies still win first and local/private protections still apply unless separately enabled.
- Add coverage for the denylist-only behavior and update the README to document it.

## Validation

- `just fmt`
- `cargo test -p codex-network-proxy` (full run had one unrelated flaky telemetry test: `network_policy::tests::emit_block_decision_audit_event_emits_non_domain_event`; reran in isolation and it passed)
- `cargo test -p codex-network-proxy network_policy::tests::emit_block_decision_audit_event_emits_non_domain_event -- --exact --nocapture`
- `just fix -p codex-network-proxy`


codex-rs/network-proxy/src/runtime.rs
```diff
@@ -1089,6 +1113,28 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn host_blocked_global_wildcard_allowlist_allows_public_hosts_except_denylist() {
```
rreichel3-oai, line 1117: @bolinfest this test confirms this usecase functions as expected!

codex-rs/network-proxy/src/policy.rs
```diff
@@ -151,19 +151,38 @@ pub(crate) fn is_global_wildcard_domain_pattern(pattern: &str) -> bool {
...
+fn compile_globset_with_policy(
+    patterns: &[String],
+    global_wildcard: GlobalWildcard,
+) -> Result<GlobSet> {
     let mut builder = GlobSetBuilder::new();
     let mut seen = HashSet::new();
     for pattern in patterns {
         ensure!(
-            !is_global_wildcard_domain_pattern(pattern),
+            matches!(global_wildcard, GlobalWildcard::Allow)
```
bolinfest: Can we deMorgan this to:

```rust
if global_wildcard == GlobalWildcard::Reject && is_global_wildcard_domain_pattern(pattern) {
  return Err(...)
}
```

Because then it's clearer that the error message follows from the condition.

bolinfest (approved):

PR #15580, opened 2026-03-23
Remove filter from plugins/list result

canvrno-oai:
Show all plugin marketplaces in the /plugins popup by removing the `openai-curated` marketplace filter, and update plugin popup copy/tests/snapshots to match the new behavior in both TUI codepaths.


codex-rs/core/src/agent/control_tests.rs
```diff
@@ -963,6 +963,7 @@ async fn spawn_child_completion_notifies_parent_history() {
 }
 
 #[tokio::test]
+#[ignore = "flaky on: rust-ci / Tests — windows-arm64 - aarch64-pc-windows-msvc"]
```
mzeng-openai, line 968: Seems unrelated?

mzeng-openai (approved):

PR #15583, opened 2026-03-24
[codex] reuse shared state DB across cold lookup paths

charley-oai:
## What changed
- eagerly initialize an optional shared `StateDbHandle` at app-server startup and reuse it for cold `thread/read`, `thread/resume`, summary, and rollout-path lookup paths
- keep SQLite optional: if startup init fails, log a warning and continue; cold lookup paths retry initialization on demand
- thread the optional state DB through rollout lookup helpers and agent-resume flows so those callsites can reuse an existing handle instead of reopening SQLite
- preserve filesystem fallback and read-repair when rollout paths in the DB are stale or missing, including the empty-DB-page fallback in thread listing
- update app-server tests to match DB-backed `cwd` normalization on Windows while keeping direct-rollout expectations unchanged

## Why
Cold lookup paths were repeatedly reopening SQLite through helper calls that ultimately reached `StateRuntime::init()`. That initialization opens both `state_5.sqlite` and `logs_1.sqlite`, so on machines with a very large `logs_1.sqlite`, metadata-only thread opens could still pay repeated logs DB open and migration cost.

This change keeps the fix narrow: reuse one shared state DB handle where app-server already has long-lived state, and thread that handle through adjacent lookup codepaths instead of introducing a broader global cache. SQLite remains optional and existing filesystem fallback behavior stays intact.

## Impact
- reduces repeated SQLite runtime initialization on cold lookup paths
- avoids making SQLite a hard startup dependency
- keeps session-owned persistence unchanged
- avoids regressions in list/read/summary behavior while the DB path is exercised more often

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -210,16 +210,17 @@ use codex_core::exec::ExecCapturePolicy;
 use codex_core::exec::ExecExpiration;
 use codex_core::exec::ExecParams;
 use codex_core::exec_env::create_env;
-use codex_core::find_archived_thread_path_by_id_str_with_state_db;
+use codex_core::find_archived_thread_path_by_id_str;
 use codex_core::find_thread_name_by_id;
 use codex_core::find_thread_names_by_ids;
-use codex_core::find_thread_path_by_id_str_with_state_db;
+use codex_core::find_thread_path_by_id_str;
 use codex_core::mcp::auth::discover_supported_scopes;
 use codex_core::mcp::auth::resolve_oauth_scopes;
 use codex_core::mcp::collect_mcp_snapshot;
 use codex_core::mcp::group_tools_by_server;
 use codex_core::models_manager::collaboration_mode_presets::CollaborationModesConfig;
 use codex_core::parse_cursor;
+use codex_core::path_utils::normalize_for_path_comparison;
 use codex_core::plugins::MarketplaceError;
 use codex_core::plugins::MarketplacePluginSource;
 use codex_core::plugins::OPENAI_CURATED_MARKETPLACE_NAME;
@@ -234,7 +235,6 @@ use codex_core::read_session_meta_line;
 use codex_core::rollout_date_parts;
 use codex_core::sandboxing::SandboxPermissions;
 use codex_core::state_db::StateDbHandle;
-use codex_core::state_db::get_state_db;
 use codex_core::state_db::reconcile_rollout;
 use codex_core::windows_sandbox::WindowsSandboxLevelExt;
 use codex_core::windows_sandbox::WindowsSandboxSetupMode as CoreWindowsSandboxSetupMode;
@@ -300,7 +300,6 @@ use std::sync::atomic::Ordering;
 use std::time::Duration;
 use std::time::SystemTime;
 use tokio::sync::Mutex;
-use tokio::sync::OnceCell;
 use tokio::sync::broadcast;
 use tokio::sync::oneshot;
 use tokio::sync::watch;
@@ -385,7 +384,7 @@ pub(crate) struct CodexMessageProcessor {
     pending_fuzzy_searches: Arc<Mutex<HashMap<String, Arc<AtomicBool>>>>,
     fuzzy_search_sessions: Arc<Mutex<HashMap<String, FuzzyFileSearchSession>>>,
     background_tasks: TaskTracker,
-    app_server_state_db_cell: OnceCell<StateDbHandle>,
+    state_db: StateDbHandle,
     feedback: CodexFeedback,
     log_db: Option<LogDbLayer>,
 }
@@ -432,6 +431,7 @@ pub(crate) struct CodexMessageProcessorArgs {
     pub(crate) cloud_requirements: Arc<RwLock<CloudRequirementsLoader>>,
     pub(crate) feedback: CodexFeedback,
     pub(crate) log_db: Option<LogDbLayer>,
+    pub(crate) state_db: StateDbHandle,
 }
 
 impl CodexMessageProcessor {
@@ -485,60 +485,6 @@ impl CodexMessageProcessor {
         Ok((thread_id, thread))
     }
 
-    async fn app_server_state_db(&self) -> Option<StateDbHandle> {
-        let config = Arc::clone(&self.config);
-        self.app_server_state_db_cell
-            .get_or_try_init(|| async move { get_state_db(&config).await.ok_or(()) })
-            .await
-            .ok()
-            .cloned()
-    }
-
-    async fn find_thread_path_by_id_with_state_db(
-        &self,
-        thread_id: ThreadId,
-        state_db_ctx: Option<StateDbHandle>,
-    ) -> std::io::Result<Option<PathBuf>> {
-        find_thread_path_by_id_str_with_state_db(
-            &self.config.codex_home,
-            &thread_id.to_string(),
-            state_db_ctx,
-        )
-        .await
-    }
-
-    async fn find_archived_thread_path_by_id_with_state_db(
-        &self,
-        thread_id: ThreadId,
-        state_db_ctx: Option<StateDbHandle>,
-    ) -> std::io::Result<Option<PathBuf>> {
-        find_archived_thread_path_by_id_str_with_state_db(
-            &self.config.codex_home,
-            &thread_id.to_string(),
-            state_db_ctx,
-        )
-        .await
-    }
-
-    async fn find_thread_path_by_id(
-        &self,
-        thread_id: ThreadId,
-    ) -> std::io::Result<Option<PathBuf>> {
-        self.find_thread_path_by_id_with_state_db(thread_id, self.app_server_state_db().await)
-            .await
-    }
-
-    async fn find_archived_thread_path_by_id(
-        &self,
-        thread_id: ThreadId,
-    ) -> std::io::Result<Option<PathBuf>> {
-        self.find_archived_thread_path_by_id_with_state_db(
-            thread_id,
-            self.app_server_state_db().await,
-        )
-        .await
-    }
-
     pub fn new(args: CodexMessageProcessorArgs) -> Self {
         let CodexMessageProcessorArgs {
             auth_manager,
@@ -551,6 +497,7 @@ impl CodexMessageProcessor {
             cloud_requirements,
             feedback,
             log_db,
+            state_db,
         } = args;
         Self {
             auth_manager,
@@ -569,7 +516,7 @@ impl CodexMessageProcessor {
             pending_fuzzy_searches: Arc::new(Mutex::new(HashMap::new())),
             fuzzy_search_sessions: Arc::new(Mutex::new(HashMap::new())),
             background_tasks: TaskTracker::new(),
-            app_server_state_db_cell: OnceCell::new(),
+            state_db,
             feedback,
             log_db,
         }
@@ -2294,7 +2241,13 @@ impl CodexMessageProcessor {
             }
         };
 
-        let rollout_path = match self.find_thread_path_by_id(thread_id).await {
+        let rollout_path = match find_thread_path_by_id_str(
+            &self.config.codex_home,
+            &thread_id.to_string(),
+            Some(self.state_db.clone()),
+        )
+        .await
+        {
             Ok(Some(p)) => p,
             Ok(None) => {
                 let error = JSONRPCErrorError {
@@ -2442,7 +2395,13 @@ impl CodexMessageProcessor {
             return;
         }
 
-        let thread_exists = match self.find_thread_path_by_id(thread_id).await {
+        let thread_exists = match find_thread_path_by_id_str(
+            &self.config.codex_home,
+            &thread_id.to_string(),
+            Some(self.state_db.clone()),
+        )
+        .await
+        {
             Ok(Some(_)) => true,
             Ok(None) => false,
             Err(err) => {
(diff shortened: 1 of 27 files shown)


charley-oai:
@codex review

codex-rs/core/src/state_db.rs
```diff
@@ -14,33 +14,36 @@ use codex_protocol::protocol::SessionSource;
...
 /// Core-facing handle to the SQLite-backed state runtime.
 pub type StateDbHandle = Arc<codex_state::StateRuntime>;
 
+#[derive(Clone)]
+struct CachedStateDb {
+    default_provider: String,
+    runtime: Weak<codex_state::StateRuntime>,
+}
+
+static STATE_DB_CACHE: LazyLock<StdMutex<HashMap<PathBuf, CachedStateDb>>> =
```
jif-oai: This heavily smell like a bad design to have a lock on a mutex. Even more with a downgraded Arc in it...

codex-rs/core/src/state_db.rs
```diff
@@ -84,13 +86,73 @@ pub async fn open_if_present(codex_home: &Path, default_provider: &str) -> Optio
...
+        && !default_provider.is_empty()
+        && !cached.default_provider.is_empty()
+        && cached.default_provider != default_provider
+    {
+        return None;
+    }
+    Some(runtime)
+}
+
+fn prune_stale_cached_runtimes(cache: &mut HashMap<PathBuf, CachedStateDb>) {
```
jif-oai: When would we have some stale cache?

codex-rs/core/src/state_db.rs
```diff
@@ -14,33 +14,36 @@ use codex_protocol::protocol::SessionSource;
...
+use std::sync::Mutex as StdMutex;
+use std::sync::Weak;
 use tracing::warn;
 use uuid::Uuid;
 
 /// Core-facing handle to the SQLite-backed state runtime.
 pub type StateDbHandle = Arc<codex_state::StateRuntime>;
 
+#[derive(Clone)]
+struct CachedStateDb {
```
jif-oai: This is not really a cache? This is most of a globally available thing like a singleton

jif-oai (commented):
I don't think this is a good design... in practice, we should init once the DB connections and then carry it on. 

In the current implementation, the DB is initialised in Session and we init some ad-hoc connection pool from the TUI or the app-server based on the config when we need them

I think we should move the DB init one step higher, in the app-server and carry it on everywhere from there
For the TUI, afaik, this is only at init for logs so it's fine to just re-open a dedicated connection pool

codex-rs/core/src/state_db.rs
charley-oai: Refactored and deleted this code

codex-rs/core/src/state_db.rs
charley-oai: Refactored and deleted this code

codex-rs/core/src/state_db.rs
charley-oai: Refactored and deleted this code

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -381,6 +382,7 @@ pub(crate) struct CodexMessageProcessor {
     pending_fuzzy_searches: Arc<Mutex<HashMap<String, Arc<AtomicBool>>>>,
     fuzzy_search_sessions: Arc<Mutex<HashMap<String, FuzzyFileSearchSession>>>,
     background_tasks: TaskTracker,
+    app_server_state_db_cell: OnceCell<StateDbHandle>,
```
jif-oai: Why on `OnceCell`? That should be an `Arc`

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -478,6 +480,61 @@ impl CodexMessageProcessor {
...
+    async fn app_server_state_db(&self) -> Option<StateDbHandle> {
+        let config = Arc::clone(&self.config);
+        self.app_server_state_db_cell
+            .get_or_try_init(|| async move { get_state_db(&config).await.ok_or(()) })
+            .await
+            .ok()
+            .cloned()
+    }
+
+    async fn find_thread_path_by_id_with_state_db(
```
jif-oai: Do we really need those one line wrappers? (Same for the others below)

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -506,6 +563,7 @@ impl CodexMessageProcessor {
             pending_fuzzy_searches: Arc::new(Mutex::new(HashMap::new())),
             fuzzy_search_sessions: Arc::new(Mutex::new(HashMap::new())),
             background_tasks: TaskTracker::new(),
+            app_server_state_db_cell: OnceCell::new(),
```
jif-oai: this is why you have the OnceCell... just populate it directly. It will be cleaner

codex-rs/rollout/src/list.rs
```diff
@@ -1251,15 +1255,49 @@ pub async fn find_thread_path_by_id_str(
     codex_home: &Path,
     id_str: &str,
```
jif-oai, line 1258: WTF is this. Unrelated but this is not clean

codex-rs/core/src/rollout/list.rs
```diff
@@ -1251,15 +1255,49 @@ pub async fn find_thread_path_by_id_str(
     codex_home: &Path,
     id_str: &str,
 ) -> io::Result<Option<PathBuf>> {
-    find_thread_path_by_id_str_in_subdir(codex_home, SESSIONS_SUBDIR, id_str).await
+    find_thread_path_by_id_str_in_subdir(
+        codex_home,
+        SESSIONS_SUBDIR,
+        id_str,
+        /*state_db_ctx*/ None,
```
jif-oai: this is the kind of call site where we need to carry on the DB. As below. We should just try to get rid to all the non "with_db"

codex-rs/core/src/rollout/list.rs
```diff
@@ -1251,15 +1255,49 @@ pub async fn find_thread_path_by_id_str(
...
+    )
+    .await
+}
+
+/// Locate a recorded thread rollout file by its UUID string using an already-open SQLite handle
+/// when available. Returns `Ok(Some(path))` if found, `Ok(None)` if not present or the id is
+/// invalid.
+pub async fn find_thread_path_by_id_str_with_state_db(
```
jif-oai: those names are becoming quite long 😓

codex-rs/app-server/src/codex_message_processor.rs
jif-oai: Actually StateDbHandle is already an `Arc` so you can just have `state_db: StateDbHandle` (no need to have `app_server` in the name btw... it's redundant

charley-oai: Ah ok I think that was guarding the at-most-once initialization of the DB... but we can just do that at startup

codex-rs/app-server/src/in_process.rs
```diff
@@ -693,13 +702,17 @@ mod tests {
...
-                .expect("default config should load"),
-        }
+        let codex_home = TempDir::new().expect("create temp dir");
+        let config = ConfigBuilder::default()
+            .codex_home(codex_home.path().to_path_buf())
+            .build()
+            .await
+            .expect("test config should build");
+        std::mem::forget(codex_home);
+        config
```
charley-oai, line 719: This change makes the test stop using your real default Codex home and instead use an isolated temp .codex directory

codex-rs/rollout/src/list.rs
charley-oai: I think we can refactor to pass around a `ThreadId` instead of `&str` (is that what you mean?), but let's do that in a separate PR?

codex-rs/core/src/state_db.rs
```diff
@@ -23,9 +23,11 @@ use uuid::Uuid;
 /// Core-facing handle to the SQLite-backed state runtime.
 pub type StateDbHandle = Arc<codex_state::StateRuntime>;
 
-/// Initialize the state runtime for thread state persistence and backfill checks. To only be used
-/// inside `core`. The initialization should not be done anywhere else.
```
charley-oai: @jif-oai is it fine to ignore the warning `To only be used inside `core`. The initialization should not be done anywhere else.`? We are now calling this from app-server

codex-rs/core/src/rollout/list.rs
charley-oai: Removed the `_with_state_db` wrappers entirely

codex-rs/core/src/rollout/list.rs
charley-oai: Removed the `_with_state_db` wrappers entirely

codex-rs/app-server/src/codex_message_processor.rs
charley-oai: Removed the `_with_state_db` wrappers entirely

charley-oai:
@codex review

codex-rs/rollout/src/recorder.rs
```diff
@@ -287,6 +287,12 @@ impl RolloutRecorder {
         )
         .await
         {
+            if search_term.is_none() && db_page.items.is_empty() && !fs_page.items.is_empty() {
+                tracing::warn!(
+                    "state db discrepancy during list_threads_with_db_fallback: empty_db_page"
+                );
+                return Ok(truncate_fs_page(fs_page, page_size, sort_key));
+            }
```
charley-oai: @jif-oai I can't tell whether this diff is model-hallucinated slop, or important. Justification:
> This PR now eagerly initializes SQLite at app-server startup ([`lib.rs`](https://github.com/openai/codex/blob/codex/cache-state-db-lookups/codex-rs/app-server/src/lib.rs#L492-L497)), and that init path can kick off state-db backfill earlier ([`state_db.rs`](https://github.com/openai/codex/blob/codex/cache-state-db-lookups/codex-rs/rollout/src/state_db.rs#L31-L65)). That makes it more likely that later `thread/list` calls will actually take the DB-backed path in [`list_threads_with_db_fallback()`](https://github.com/openai/codex/blob/codex/cache-state-db-lookups/codex-rs/rollout/src/recorder.rs#L216-L301).
> 
> In that function, we already preserve filesystem behavior when SQLite is unavailable ([lines 259-264](https://github.com/openai/codex/blob/codex/cache-state-db-lookups/codex-rs/rollout/src/recorder.rs#L259-L264)), and we already compute/repair from the filesystem page before querying SQLite ([lines 266-275](https://github.com/openai/codex/blob/codex/cache-state-db-lookups/codex-rs/rollout/src/recorder.rs#L266-L275)). Without this extra guard, if SQLite is available but returns an empty page while the filesystem page is non-empty, we would return the empty DB page and hide real rollout files on disk ([lines 277-296](https://github.com/openai/codex/blob/codex/cache-state-db-lookups/codex-rs/rollout/src/recorder.rs#L277-L296)).
> 
> So I don’t mean this as a new product behavior; it’s a discrepancy fallback to preserve the preexisting filesystem-visible result in the specific case where enabling SQLite earlier would otherwise regress thread listing. The `search_term.is_none()` check is intentional there, since an empty DB page for a search query can be the correct result.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -430,9 +431,17 @@ pub(crate) struct CodexMessageProcessorArgs {
...
     pub(crate) feedback: CodexFeedback,
     pub(crate) log_db: Option<LogDbLayer>,
+    pub(crate) state_db: Option<StateDbHandle>,
 }
 
 impl CodexMessageProcessor {
+    async fn shared_state_db(&self) -> Option<StateDbHandle> {
+        match self.state_db.as_ref() {
+            Some(state_db) => Some(state_db.clone()),
+            None => init(self.config.as_ref()).await,
```
jif-oai: Ultra Rust nit: `init` here is pretty hard to understand... A good implementation would be to have
In the import:
```
use codex_core::state_db;
```

Here:
```
state_db::init(...);
```

It makes the code easier to read

codex-rs/app-server/src/in_process.rs
```diff
@@ -325,7 +327,14 @@ impl InProcessClientHandle {
 /// the runtime is shut down and an `InvalidData` error is returned.
 pub async fn start(args: InProcessStartArgs) -> IoResult<InProcessClientHandle> {
     let initialize = args.initialize.clone();
-    let client = start_uninitialized(args);
+    let state_db = init(args.config.as_ref()).await;
```
jif-oai: same comment

codex-rs/app-server/src/lib.rs
```diff
@@ -489,6 +490,13 @@ pub async fn run_main_with_transport(
     }
 
     let feedback = CodexFeedback::new();
+    let state_db = init(&config).await;
```
jif-oai, line 493: good for now but here you init the full state just to get the log DB. I think we should decouple them at some point if we don't plan to have 10 more DBs

codex-rs/rollout/src/recorder.rs
jif-oai: This is correct but EXTREMELY unlikely and will become even more unlikely with time

jif-oai (approved):
IMO this implementation is way better

codex-rs/rollout/src/recorder.rs
charley-oai: Ok I'll cut it out

PR #15587, opened 2026-03-24
Add use_agent_identity feature flag

adrian-openai:
## Stack
- [PR1: #15587 Add use_agent_identity feature flag](https://github.com/openai/codex/pull/15587) <- this PR
- [PR2: #15588 Register agent identities behind use_agent_identity](https://github.com/openai/codex/pull/15588)
- [PR3: #15589 Register agent tasks behind use_agent_identity](https://github.com/openai/codex/pull/15589)
- [PR4: #15590 Use AgentAssertion downstream behind use_agent_identity](https://github.com/openai/codex/pull/15590)

## Summary
- add the under-development `use_agent_identity` feature gate
- expose the new flag in the config schema and feature tests
- keep the initial PR limited to config plumbing so the rollout can stack cleanly

## Testing
- stack validation was run on the top branch after the full flow landed

PR #15588, opened 2026-03-24
Register agent identities behind use_agent_identity

adrian-openai:
## Stack
- [PR1: #15587 Add use_agent_identity feature flag](https://github.com/openai/codex/pull/15587)
- [PR2: #15588 Register agent identities behind use_agent_identity](https://github.com/openai/codex/pull/15588) <- this PR
- [PR3: #15589 Register agent tasks behind use_agent_identity](https://github.com/openai/codex/pull/15589)
- [PR4: #15590 Use AgentAssertion downstream behind use_agent_identity](https://github.com/openai/codex/pull/15590)

## Summary
- add `AgentIdentityManager` and the agent-registration flow behind `use_agent_identity`
- persist the registered agent identity and key material per ChatGPT workspace/account
- keep this PR focused on long-lived identity registration only so task/assertion work can layer on top

## Testing
- stack validation was run on the top branch after the full flow landed


codex-rs/core/src/agent_identity.rs
```diff
@@ -0,0 +1,685 @@
...
+                SecretsBackendKind::Local,
+            ),
+            chatgpt_base_url: config.chatgpt_base_url.clone(),
+            feature_enabled: config.features.enabled(Feature::UseAgentIdentity),
+            abom: build_abom(session_source),
+            ensure_lock: Arc::new(Mutex::new(())),
+        }
+    }
+
+    pub(crate) async fn ensure_registered_identity(&self) -> Result<Option<StoredAgentIdentity>> {
```
nicksteele-oai, line 114: Could be a future thing, but If the agent is unable to register, shouldn't codex hang until it _is_ capable of registering? I'd assume that an enterprise user wouldn't be able to do much with an unregistered agent.

codex-rs/core/src/codex.rs
```diff
@@ -1938,6 +1961,7 @@ impl Session {
 
         // Start the watcher after SessionConfigured so it cannot emit earlier events.
         sess.start_file_watcher_listener();
+        sess.start_agent_identity_registration();
```
nicksteele-oai, line 1997: We should probably try to start registration after the user has gone through the onboarding flow and/or logged in, because no credentials would be available yet for this session. In this current flow, the user would need to restart the codex client in order for the session to get the credentials and actually succeed in registering an agent.

codex-rs/core/src/agent_identity.rs
```diff
@@ -0,0 +1,685 @@
...
+            .decode(&self.private_key_pkcs8_base64)
+            .context("stored agent identity private key is not valid base64")?;
+        SigningKey::from_pkcs8_der(&private_key)
+            .context("stored agent identity private key is not valid PKCS#8")
+    }
+}
+
+impl AgentIdentityBinding {
+    fn from_auth(auth: &CodexAuth, forced_workspace_id: Option<String>) -> Option<Self> {
+        if !auth.is_chatgpt_auth() {
```
nicksteele-oai, line 313: Does this mean that we don't support AuthTokens either?

codex-rs/core/src/agent_identity.rs
```diff
@@ -0,0 +1,685 @@
...
+            if !is_last_candidate
+                && matches!(
+                    status,
+                    StatusCode::NOT_FOUND | StatusCode::METHOD_NOT_ALLOWED
+                )
+            {
+                debug!(
+                    url = %url,
+                    status = %status,
+                    "agent identity registration endpoint unavailable at candidate URL; trying fallback"
```
nicksteele-oai: I don't think it falls back

codex-rs/core/src/codex.rs
adrian-openai: > We should probably try to start registration after the user has gone through the onboarding flow and/or logged in, because no credentials would be available yet for this session. In this current flow, the user would need to restart the codex client in order for the session to get the credentials and actually succeed in registering an agent.

Interesting! That's a good catch. Let me update that.

codex-rs/core/src/agent_identity.rs
adrian-openai: Well, it falls back to the other URL, right? Probably don't need to do that, though.

codex-rs/core/src/agent_identity.rs
adrian-openai: I think that's pretty reasonable. But only if the feature flag is on, of course!

codex-rs/core/src/agent_identity.rs
adrian-openai: We support only chatgpt auth, which is the auth token, right?

nicksteele-oai: My understanding is that there were [two Auth flows](https://github.com/openai/codex/blob/f55f5c258f0588dd9acc38811a2260d5debc9bd3/codex-rs/login/src/auth/manager.rs#L38) that used tokens in codex, where one was used more directly. Looking at the current version of the auth, I'd say this is fine.

nicksteele-oai: It looks like both of these flows are considered.. the [same auth flow](https://github.com/openai/codex/blob/f55f5c258f0588dd9acc38811a2260d5debc9bd3/codex-rs/login/src/auth/manager.rs#L180)? So it's fine!

PR #15592, opened 2026-03-24
Extract landlock helpers into codex-sandboxing

pakrym-oai:
## Summary
- add a new `codex-sandboxing` crate for sandboxing extraction work
- move the pure Linux sandbox argv builders and their unit tests out of `codex-core`
- keep `core::landlock` as the spawn wrapper and update direct callers to use `codex_sandboxing::landlock`

## Testing
- `cargo test -p codex-core landlock`

## Notes
- this is step 1 of the move plan aimed at minimizing per-PR diffs
- no re-exports or no-op proxy methods were added


codex-rs/sandboxing/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod landlock;
```
bolinfest, line 1: Can we cfg gate this to Linux?

bolinfest (approved):

codex-rs/sandboxing/src/lib.rs
pakrym-oai: Will take as a followup

PR #15601, opened 2026-03-24
[app-server] Add a method to override feature flags.

mzeng-openai:
- [x] Add a method to override feature flags globally and not just thread level.


codex-rs/app-server/src/config_api.rs
```diff
@@ -56,7 +62,8 @@ impl UserConfigReloader for ThreadManager {
 #[derive(Clone)]
 pub(crate) struct ConfigApi {
     codex_home: PathBuf,
-    cli_overrides: Vec<(String, TomlValue)>,
+    cli_overrides: Arc<RwLock<Vec<(String, TomlValue)>>>,
+    feature_flag_overrides: Arc<RwLock<BTreeMap<String, bool>>>,
```
pakrym-oai: why can't we do this via cli overrides?

mzeng-openai: I'll update based on our conversation, this will be a one-shot API that directly alters the value and shouldn't persist previous overrides

codex-rs/app-server/src/config_api.rs
```diff
@@ -24,15 +26,20 @@ use codex_core::config_loader::SandboxModeRequirement as CoreSandboxModeRequirem
...
 use codex_protocol::protocol::Op;
 use serde_json::json;
+use std::collections::BTreeSet;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::sync::RwLock;
 use toml::Value as TomlValue;
 use tracing::warn;
 
+const SUPPORTED_EXPERIMENTAL_FEATURE_ENABLEMENT: &[&str] = &["apps", "plugins"];
```
mzeng-openai, line 43: Added an allowlist here.

pakrym-oai (approved):

PR #13191, opened 2026-03-01
keyring: split-entry keyring storage for auth token

mikhail-oai:
This changes Codex CLI auth persistence on keyring-backed platforms to avoid Windows Credential Manager size failures during browser-based login. Previously we serialized the full auth payload into one JSON string and stored it with set_password, which caused large OAuth tokens to exceed the Windows limit after UTF-16 expansion. This PR switches keyring storage to binary secrets via set_secret/get_secret, stores auth as a versioned split-entry layout under the existing CODEX_HOME-scoped key namespace, and reconstructs the in-memory auth object on load.

## Summary

This change introduces a reusable, generalized keyring storage format for JSON-backed credentials and consolidates both MCP OAuth and CLI auth onto the same storage mechanism.

Instead of each credential type owning its own bespoke keyring serialization logic, we now use a shared library that can persist arbitrary `serde_json::Value` data in a versioned split-entry layout. Each credential record is stored under its own base key, with an `active` revision pointer, a manifest describing the JSON structure, and one keyring entry per JSON leaf value. This keeps the format dynamic and schema-agnostic while preserving JSON types.

## What Changed

- Added a shared split-JSON keyring storage helper in `codex-keyring-store`
- Introduced a versioned layout using a short `v1` marker
- Flattened JSON dynamically by path instead of hardcoding credential-specific fields
- Stored each JSON leaf value in its own keyring entry, with manifest-based reconstruction
- JSON splitting is available as a storage method, currently used only on windows.
- Reused the shared helper for:
  - MCP OAuth credentials
  - CLI auth credentials

## Why

This removes duplicated credential storage logic, gives MCP OAuth and CLI auth the same persistence model, and makes future credential formats easier to support without writing new keyring serialization code for every struct.

## Manual testing
Manual human testing was performed on 
 * macOS 26
 * Windows 11 Pro (25H2) in Azure Cloud
 * Ubuntu Forky 25.10. 

Test executed `cargo run --bin codex login` followed by `cargo run --bin codex exec hello` to ensure credentials were available and used. 

When `cli_auth_credentials_store="keyring"` is set in the `config.toml`, confirmed credential was stored in 
 * Key Chain on macOS
 * Windows Credential Manager
 * Passwords and Keys

Addressing issue https://github.com/openai/codex/issues/10353

PR #15632, opened 2026-03-24
[keyring-store] split JSON storage

mikhail-oai:
This change introduces a reusable, generalized keyring storage format for JSON-backed credentials and consolidates both MCP OAuth and CLI auth onto the same storage mechanism.

Instead of each credential type owning its own bespoke keyring serialization logic, we now use a shared library that can persist arbitrary serde_json::Value data in a versioned split-entry layout. Each credential record is stored under its own base key, with an active revision pointer, a manifest describing the JSON structure, and one keyring entry per JSON leaf value. This keeps the format dynamic and schema-agnostic while preserving JSON types.

Continuation of https://github.com/openai/codex/pull/13191

### DIFF
--- a/codex-rs/keyring-store/src/json_store_full.rs
+++ b/codex-rs/keyring-store/src/json_store_full.rs
@@ -79,7 +79,8 @@ fn load_secret_from_keyring<K: KeyringStore + ?Sized>(
     field: &str,
 ) -> Result<Option<Vec<u8>>, FullJsonKeyringError> {
     keyring_store
-        .load_secret(service, key)
+        .load(service, key)
+        .map(|value| value.map(String::into_bytes))
         .map_err(|err| credential_store_error("load", field, err))
 }
 
@@ -90,8 +91,11 @@ fn save_secret_to_keyring<K: KeyringStore + ?Sized>(
     value: &[u8],
     field: &str,
 ) -> Result<(), FullJsonKeyringError> {
+    let value = std::str::from_utf8(value).map_err(|err| {
+        FullJsonKeyringError::new(format!("failed to encode {field} as UTF-8: {err}"))
+    })?;
     keyring_store
-        .save_secret(service, key, value)
+        .save(service, key, value)
         .map_err(|err| credential_store_error("write", field, err))
 }
 
@@ -145,8 +149,8 @@ mod tests {
             .expect("JSON should exist");
         assert_eq!(loaded, expected);
         assert_eq!(
-            store.saved_secret(BASE_KEY),
-            Some(serde_json::to_vec(&expected).expect("JSON should serialize")),
+            store.saved_value(BASE_KEY),
+            Some(serde_json::to_string(&expected).expect("JSON should serialize")),
         );
     }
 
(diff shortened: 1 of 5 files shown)


codex-rs/keyring-store/src/json_store_full.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+impl std::error::Error for FullJsonKeyringError {}
+
+pub fn load_json_from_keyring<K: KeyringStore + ?Sized>(
+    keyring_store: &K,
+    service: &str,
+    base_key: &str,
+) -> Result<Option<Value>, JsonKeyringError> {
+    if let Some(bytes) = load_secret_from_keyring(keyring_store, service, base_key, "JSON record")?
```
celia-oai, line 34: can we get rid of this "field" variable?

codex-rs/keyring-store/src/json_store_full.rs
```diff
@@ -0,0 +1,196 @@
...
+
+pub fn delete_json_from_keyring<K: KeyringStore + ?Sized>(
+    keyring_store: &K,
+    service: &str,
+    base_key: &str,
+) -> Result<bool, JsonKeyringError> {
+    delete_keyring_entry(keyring_store, service, base_key, "JSON record")
+}
+
+fn load_secret_from_keyring<K: KeyringStore + ?Sized>(
```
celia-oai, line 75: can we just collapse `load_secret_from_keyring`, `save_secret_to_keyring` and `delete_keyring_entry` in the json functions above for now instead of having separate functions?

celia-oai: & can we just keep the value to be string instead of vec8? just trying to decrease the delta of this new code and old code. I think we could probably just copy over old code from e.g. `load_from_keyring` & `save_to_keyring` instead of reinventing the wheel.

celia-oai (commented):
I would still prefer a common struct or similar shared abstraction around the JSON keyring API surface here. Both [`json_store_full`](./codex-rs/keyring-store/src/json_store_full.rs) and [`json_store_split`](./codex-rs/keyring-store/src/json_store_split.rs) currently implement the same necessary functions, including `load_json_from_keyring`, `save_json_to_keyring`, and `delete_json_from_keyring`, and I think we should make that relationship explicit so the two implementations cannot drift. That would give us one place to define the required behavior and make it clearer that both backends are expected to support the same contract.

Something roughly like:

```rust
trait JsonKeyringStore {
    fn load_json(&self, service: &str, base_key: &str) -> Result<Option<Value>, JsonKeyringError>;
    fn save_json(&self, service: &str, base_key: &str, value: &Value) -> Result<(), JsonKeyringError>;
    fn delete_json(&self, service: &str, base_key: &str) -> Result<bool, JsonKeyringError>;
}

struct FullJsonStore;
impl JsonKeyringStore for FullJsonStore { /* ... */ }

struct SplitJsonStore;
impl JsonKeyringStore for SplitJsonStore { /* ... */ }
```

codex-rs/keyring-store/src/json_store_split.rs
celia-oai, line 1: can we add a docstring in this file explaining the mechanism of collapsing json into key/values, and how to recover? ideally by providing an example of json -> manifest -> leaf entries

PR #15659, opened 2026-03-24
Add MCP tool call spans

nicholasclark-openai:
## Summary
- add an explicit `mcp.tools.call` span around MCP tool execution in core
- keep MCP span validation local to `mcp_tool_call_tests` instead of broadening the integration test suite
- inline the turn/session correlation fields directly in the span initializer

## Included Changes
- `codex-rs/core/src/mcp_tool_call.rs`: wrap the existing MCP tool call in `mcp.tools.call` and inline `conversation.id`, `session.id`, and `turn.id` in the span initializer
- `codex-rs/core/src/mcp_tool_call_tests.rs`: assert the MCP span records the expected correlation and server fields

## Testing
- `cargo test -p codex-core`
- `just fmt`

## Notes
- `cargo test -p codex-core` still hits existing unrelated failures in guardian-config tests and the sandboxed JS REPL `mktemp` test
- metric work moved to sibling stacked PR #15805
- transport-level RMCP spans and trace propagation remain in sibling stacked PR #15792
- full workspace `cargo test` was not run


codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -273,6 +318,66 @@ pub(crate) async fn handle_mcp_tool_call(
...
+        conversation.id = Empty,
+        session.id = Empty,
+        turn.id = Empty,
+        server.address = Empty,
+        server.port = Empty,
+    );
+    let conversation_id = session.conversation_id.to_string();
+    span.record("conversation.id", conversation_id.as_str());
+    span.record("session.id", conversation_id.as_str());
+    span.record("turn.id", turn_context.sub_id.as_str());
```
owenlin0: nit: interesting, why do it like this instead of defining inline above?

owenlin0 (approved):

codex-rs/core/src/mcp_tool_call.rs
nicholasclark-openai: Codex decision that I missed. It was just a compile-time/runtime split

PR #15667, opened 2026-03-24
[codex] Handle guardian review timeouts explicitly

charley-oai:
## What changed

- increased the guardian reviewer timeout from 90s to 120s
- added a distinct guardian `TimedOut` terminal status through core, protocol, app-server, and TUI wiring
- surfaced guardian review timeouts as normal model-visible failure text instead of the rejection/decline path for exec, network, MCP, and escalation approvals
- added timeout-path coverage and new TUI snapshot baselines

## Why

Guardian reviews already failed closed on timeout, but the failure surfaced the same way as an intentional rejection. That made the model-visible text misleading and hid the fact that a retry could succeed.

## Impact

- the model can now distinguish reviewer timeouts from intentional denials
- app-server clients and both TUIs receive an explicit `timedOut` status
- retry guidance is clearer without weakening the fail-closed approval behavior

## Root cause

Guardian timeout handling collapsed into the same denial path used for intentional rejections, so downstream approval flows normalized both cases as rejected-by-user or denied-by-policy outcomes.


codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
```diff
@@ -429,25 +432,8 @@ impl CoreShellActionProvider {
...
-                            tool_name: tool_name.to_string(),
-                            program: program.to_string_lossy().into_owned(),
-                            argv: argv.to_vec(),
-                            cwd: workdir,
-                            additional_permissions,
-                        },
-                        /*retry_reason*/ None,
-                    )
-                    .await;
-                }
```
charley-oai, line 450: Guardian-routed execve reviews short-circuit in `process_decision()` before calling this helper. `prompt()` is only the human approval path.

PR #15672, opened 2026-03-24
tui_app_server: open ChatGPT login in the local browser

etraut-openai:
## Summary

Fixes ChatGPT login in `tui_app_server` so the local browser opens again during in-process login flows.

## Root cause

The app-server backend intentionally starts ChatGPT login with browser auto-open disabled, expecting the TUI client to open the returned `auth_url`. The app-server TUI was not doing that, so the login URL was shown in the UI but no browser window opened.

## Changes

- Add a helper that opens the returned ChatGPT login URL locally
- Call it from the main ChatGPT login flow
- Call it from the device-code fallback-to-browser path as well
- Limit auto-open to in-process app-server handles so remote sessions do not try to open a browser against a remote localhost callback


fcoury:
@etraut-openai looks good!

PR #15675, opened 2026-03-24
Add client-defined metadata to threads

guinness-oai:
## Summary
- add an optional arbitrary metadata map to `thread/start` and `thread/fork`, and return it on `Thread`
- persist thread metadata in state DB as JSON so it survives list/read/resume/fork flows
- thread metadata through core session/rollout plumbing instead of introducing a separate side channel
- update generated schemas, docs, and app-server tests to cover the new field


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2713,6 +2715,8 @@ pub struct ThreadForkParams {
     pub base_instructions: Option<String>,
     #[ts(optional = nullable)]
     pub developer_instructions: Option<String>,
+    #[ts(optional = nullable)]
+    pub metadata: Option<BTreeMap<String, JsonValue>>,
```
owenlin0: Can we make metadata a `BTreeMap<String, String>` instead, similar to responsesapi? [reference](https://developers.openai.com/api/reference/resources/$shared#(resource)%20%24shared%20%3E%20(model)%20metadata%20%3E%20(schema))

owenlin0 (commented):
Supportive of client-defined metadata, but some general thoughts:
- Can we expose metadata as a simpler string->string key value store? That way we're not allowing JSON objects of arbitrary depth.
- Let's add validation. For reference, responsesapi supports up to 16 key-value pairs and each string must be <= 512 chars.
- Can we exclusively rely on SQLite to store/retrieve this field? This is exactly the kind of data we should stop reading from the rollout file.

Can be done as a followup PR, but would be awesome to have `thread/metadata/update` be able to update a thread's metadata.

PR #15689, opened 2026-03-24
app-server: Return codex home in initialize response

euroelessar:
This allows clients to get enough information to interact with the codex skills/configuration/etc.


codex-rs/app-server-protocol/src/protocol/v1.rs
```diff
@@ -56,6 +56,8 @@ pub struct InitializeCapabilities {
 #[serde(rename_all = "camelCase")]
 pub struct InitializeResponse {
     pub user_agent: String,
+    /// Absolute path to the server's Codex home directory.
+    pub codex_home: PathBuf,
```
owenlin0: let's use `AbsolutePathBuf`?

owenlin0 (approved):

codex-rs/app-server-protocol/src/protocol/v1.rs
```diff
@@ -56,6 +56,8 @@ pub struct InitializeCapabilities {
 #[serde(rename_all = "camelCase")]
 pub struct InitializeResponse {
     pub user_agent: String,
+    /// Absolute path to the server's Codex home directory.
```
owenlin0: nit: `Absolute path to the server's $CODEX_HOME directory.`?

PR #15691, opened 2026-03-24
feat: exec-server prep for unified exec

jif-oai:
This PR partially rebase `unified_exec` on the `exec-server` and adapt the `exec-server` accordingly.

## What changed in `exec-server`

1. Replaced the old "broadcast-driven; process-global" event model with process-scoped session events. The goal is to be able to have dedicated handler for each process.
2. Add to protocol contract to support explicit lifecycle status and stream ordering:
  - `WriteResponse` now returns `WriteStatus` (Accepted, UnknownProcess, StdinClosed, Starting) instead of a bool.
  - Added seq fields to output/exited notifications.
  - Added terminal process/closed notification.
3. Demultiplexed remote notifications into per-process channels. Same as for the event sys
4. Local and remote backends now both implement ExecBackend.
5. Local backend wraps internal process ID/operations into per-process ExecProcess objects.
6. Remote backend registers a session channel before launch and unregisters on failed launch.

## What changed in `unified_exec`

1. Added unified process-state model and backend-neutral process wrapper. This will probably disappear in the future, but it makes it easier to keep the work flowing on both side.
  - `UnifiedExecProcess` now handles both local PTY sessions and remote exec-server processes through a shared `ProcessHandle`.
  - Added `ProcessState` to track has_exited, exit_code, and terminal failure message consistently across backends.
2. Routed write and lifecycle handling through process-level methods.

## Some rationals

1. The change centralizes execution transport in exec-server while preserving policy and orchestration ownership in core, avoiding duplicated launch approval logic. This comes from internal discussion.
2. Session-scoped events remove coupling/cross-talk between processes and make stream ordering and terminal state explicit (seq, closed, failed).
3. The failure-path surfacing (remote launch failures, write failures, transport disconnects) makes command tool output and cleanup behavior deterministic

## Follow-ups:
* Unify the concept of thread ID behind an obfuscated struct
* FD handling
* Full zsh-fork compatibility
* Full network sandboxing compatibility
* Handle ws disconnection


codex-rs/exec-server/src/client.rs
```diff
@@ -307,6 +307,31 @@ impl ExecServerClient {
             .map_err(Into::into)
     }
 
+    pub(crate) async fn register_session(
+        &self,
+        process_id: &str,
```
bolinfest, line 333: Given this is a str and not an int, so this is not a `PID`?

codex-rs/exec-server/src/client.rs
```diff
@@ -378,12 +403,33 @@ async fn handle_server_notification(
         EXEC_OUTPUT_DELTA_METHOD => {
             let params: ExecOutputDeltaNotification =
                 serde_json::from_value(notification.params.unwrap_or(Value::Null))?;
-            let _ = inner.events_tx.send(ExecServerEvent::OutputDelta(params));
+            let events_tx = { inner.sessions.lock().await.get(&params.process_id).cloned() };
```
bolinfest: This is going to be a common operation, no? So will this mutex be acquired/released frequently?

codex-rs/exec-server/src/client.rs
jif-oai: No, this was never a PID actually. Having it as an int has been a tech debts I have been wanting to fix in `unified_exec`
We might also fully move a `struct ProcessId` to be even more flexible

codex-rs/exec-server/src/client.rs
jif-oai: Yes, I'm working on a rework here. I don't have a perfect solution yet

codex-rs/exec-server/Cargo.toml
```diff
@@ -15,6 +15,7 @@ path = "src/bin/codex-exec-server.rs"
 workspace = true
 
 [dependencies]
+arc-swap = "1.8.2"
```
jif-oai: Ooooops, RR. Will fix

codex-rs/exec-server/src/protocol.rs
```diff
@@ -99,10 +100,19 @@ pub struct WriteParams {
     pub chunk: ByteChunk,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum WriteStatus {
+    Accepted,
+    UnknownProcess,
```
pakrym-oai, line 109: not very importatnt. do we need a generic method level error mechanism?

codex-rs/exec-server/src/protocol.rs
```diff
@@ -129,6 +139,7 @@ pub enum ExecOutputStream {
 #[serde(rename_all = "camelCase")]
 pub struct ExecOutputDeltaNotification {
     pub process_id: String,
+    pub seq: u64,
```
pakrym-oai, line 144: curious, why we need seq considering the protocol is ordered

jif-oai: This is for ws reconnection

PR #13860, opened 2026-03-07
Add Smart Approvals guardian review across core, app-server, and TUI

charley-oai:
## Summary
- add `approvals_reviewer = "user" | "guardian_subagent"` as the runtime control for who reviews approval requests
- route Smart Approvals guardian review through core for command execution, file changes, managed-network approvals, MCP approvals, and delegated/subagent approval flows
- expose guardian review in app-server with temporary unstable `item/autoApprovalReview/{started,completed}` notifications carrying `targetItemId`, `review`, and `action`
- update the TUI so Smart Approvals can be enabled from `/experimental`, aligned with the matching `/approvals` mode, and surfaced clearly while reviews are pending or resolved

## Runtime model
This PR does not introduce a new `approval_policy`.

Instead:
- `approval_policy` still controls when approval is needed
- `approvals_reviewer` controls who reviewable approval requests are routed to:
  - `user`
  - `guardian_subagent`

`guardian_subagent` is a carefully prompted reviewer subagent that gathers relevant context and applies a risk-based decision framework before approving or denying the request.

The `smart_approvals` feature flag is a rollout/UI gate. Core runtime behavior keys off `approvals_reviewer`.

When Smart Approvals is enabled from the TUI, it also switches the current `/approvals` settings to the matching Smart Approvals mode so users immediately see guardian review in the active thread:
- `approval_policy = on-request`
- `approvals_reviewer = guardian_subagent`
- `sandbox_mode = workspace-write`

Users can still change `/approvals` afterward.

Config-load behavior stays intentionally narrow:
- plain `smart_approvals = true` in `config.toml` remains just the rollout/UI gate and does not auto-set `approvals_reviewer`
- the deprecated `guardian_approval = true` alias migration does backfill `approvals_reviewer = "guardian_subagent"` in the same scope when that reviewer is not already configured there, so old configs preserve their original guardian-enabled behavior

ARC remains a separate safety check. For MCP tool approvals, ARC escalations now flow into the configured reviewer instead of always bypassing guardian and forcing manual review.

## Config stability
The runtime reviewer override is stable, but the config-backed app-server protocol shape is still settling.

- `thread/start`, `thread/resume`, and `turn/start` keep stable `approvalsReviewer` overrides
- the config-backed `approvals_reviewer` exposure returned via `config/read` (including profile-level config) is now marked `[UNSTABLE]` / experimental in the app-server protocol until we are more confident in that config surface

## App-server surface
This PR intentionally keeps the guardian app-server shape narrow and temporary.

[...]

PR #15693, opened 2026-03-24
fix: fix old system bubblewrap compatibility without falling back to vendored bwrap

viyatb-oai:
Fixes #15283.

## Summary
Older system bubblewrap builds reject `--argv0`, which makes our Linux sandbox fail before the helper can re-exec. This PR keeps using system `/usr/bin/bwrap` whenever it exists and only falls back to vendored bwrap when the system binary is missing. That matters on stricter AppArmor hosts, where the distro bwrap package also provides the policy setup needed for user namespaces.

For old system bwrap, we avoid `--argv0` instead of switching binaries:
- pass the sandbox helper a full-path `argv0`,
- keep the existing `current_exe() + --argv0` path when the selected launcher supports it,
- otherwise omit `--argv0` and re-exec through the helper's own `argv[0]` path, whose basename still dispatches as `codex-linux-sandbox`.

Also updates the launcher/warning tests and docs so they match the new behavior: present-but-old system bwrap uses the compatibility path, and only absent system bwrap falls back to vendored.

### Validation

1. Install Ubuntu 20.04 in a VM
2. Compile codex and run without bubblewrap installed - see a warning about falling back to the vendored bwrap
3. Install bwrap and verify version is 0.4.0 without `argv0` support
4. run codex and use apply_patch tool without errors

Screenshot 2026-03-25 at 11 48 36 PM
Screenshot 2026-03-25 at 11 47 32 PM
Screenshot 2026-03-25 at 11 45 45 PM
Screenshot 2026-03-25 at 11 43 56 PM


codex-rs/core/src/config/config_tests.rs
```diff
@@ -5604,10 +5604,8 @@ exit 1
 "#,
     );
     let fake_bwrap_path: &Path = fake_bwrap.as_ref();
-    let warning = system_bwrap_warning_for_path(fake_bwrap_path)
-        .expect("old system bwrap should emit a warning");
 
-    assert!(warning.contains("too old to support `--argv0`"));
+    assert_eq!(system_bwrap_warning_for_path(fake_bwrap_path), None);
```
bolinfest: Maybe:

```suggestion
    assert_eq!(system_bwrap_warning_for_path(fake_bwrap_path), None, "Do not warn even if bwrap does not support `--argv0`");
```

Or possibly we delete this test since I'm not sure what additional confidence it adds at this point?

codex-rs/arg0/src/lib.rs
```diff
@@ -176,6 +173,18 @@ where
     })
 }
 
+fn linux_sandbox_exe_path(
+    path_entry: Option<&Arg0PathEntryGuard>,
```
bolinfest: I think we should rename this `path_entry_guard` in `arg0_dispatch_or_else()`, as well.

```suggestion
    path_entry_guard: Option<&Arg0PathEntryGuard>,
```

codex-rs/arg0/src/lib.rs
```diff
@@ -159,11 +160,7 @@ where
         let current_exe = std::env::current_exe().ok();
         let paths = Arg0DispatchPaths {
             codex_linux_sandbox_exe: if cfg!(target_os = "linux") {
-                current_exe.or_else(|| {
```
bolinfest, line 162: Hmm, I'm trying to remember why we would prefer `current_exe` in the first place.  Does it save one symlink read? And does `arg0` have to be explicitly set to `codex-linux-sandbox` for it to work in that case?

codex-rs/core/src/config/config_tests.rs
bolinfest: I would still add a third string arg to this `assert_eq!()` to underscore what we are verifying here.

codex-rs/core/src/tools/handlers/apply_patch.rs
```diff
@@ -195,6 +195,8 @@ impl ToolHandler for ApplyPatchHandler {
                         );
                         emitter.begin(event_ctx).await;
 
+                        #[cfg(all(not(target_os = "windows"), not(target_os = "linux")))]
```
bolinfest: This seems odd...

codex-rs/arg0/src/lib.rs
viyatb-oai: correct, it saved one symlink read. 
re arg0: Yes. If we use `current_exe()`, `arg0` must be explicitly set to `codex-linux-sandbox`. Otherwise the process starts with basename codex, and `arg0` dispatch won’t enter sandbox-helper mode.

codex-rs/core/src/landlock.rs
```diff
@@ -12,6 +12,8 @@ use std::path::Path;
 use std::path::PathBuf;
 use tokio::process::Child;
 
+const LINUX_SANDBOX_ARG0: &str = "codex-linux-sandbox";
```
bolinfest: Can we use the one in `codex-arg0`?

Though I admit `codex-arg0` depends on a bunch of stuff that maybe this does not. Can you look into moving the const into a new, common location as a follow-up to this PR?

codex-rs/core/src/landlock.rs
```diff
@@ -48,11 +50,20 @@ where
...
+        .and_then(|name| name.to_str())
+        == Some(LINUX_SANDBOX_ARG0)
+    {
+        codex_linux_sandbox_exe.to_string_lossy().into_owned()
+    } else {
+        LINUX_SANDBOX_ARG0.to_string()
+    };
     spawn_child_async(SpawnChildRequest {
-        program: codex_linux_sandbox_exe.as_ref().to_path_buf(),
+        program: codex_linux_sandbox_exe.to_path_buf(),
```
bolinfest, line 67: Outside the scope of this PR, but I would like to change `program` to be `AsRef<OsStr>` instead of `String` so we don't have to use `to_string_lossy()` quite so much.

codex-rs/core/src/landlock.rs
```diff
@@ -48,11 +50,20 @@ where
         use_legacy_landlock,
         allow_network_for_proxy(/*enforce_managed_network*/ false),
     );
-    let arg0 = Some("codex-linux-sandbox");
+    let codex_linux_sandbox_exe = codex_linux_sandbox_exe.as_ref();
+    let arg0 = if codex_linux_sandbox_exe
```
bolinfest, line 55: Add a comment on this one?

bolinfest (approved):
If you did get an Ubuntu 20 to test on, can you please update the PR body to explain what you tested?

codex-rs/core/src/landlock.rs
viyatb-oai: there was a circular dependency in codex-arg0 but let me try again.

codex-rs/apply-patch/src/lib.rs
```diff
@@ -25,6 +25,10 @@ use crate::invocation::ExtractHeredocError;
 /// Detailed instructions for gpt-4.1 on how to use the `apply_patch` tool.
 pub const APPLY_PATCH_TOOL_INSTRUCTIONS: &str = include_str!("../apply_patch_tool_instructions.md");
 
+/// Basename used when the Codex executable self-invokes as the Linux sandbox
+/// helper.
+pub const CODEX_LINUX_SANDBOX_ARG0: &str = "codex-linux-sandbox";
```
bolinfest: `apply-patch` should have no knowledge of this: it doesn't seem like the right place for it to live?

bolinfest: I get that `arg0` already depends on `apply-patch` so it doesn't bloat its dependencies, but I don't think that's sufficient justification for moving this here.

viyatb-oai: yeah trying to find a better place for it

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -483,6 +475,46 @@ fn build_bwrap_argv(
...
+) {
+    let command_separator_index = argv
+        .iter()
+        .position(|arg| arg == "--")
+        .unwrap_or_else(|| panic!("bubblewrap argv is missing command separator '--'"));
+
+    if supports_argv0 {
+        argv.splice(
+            command_separator_index..command_separator_index,
+            ["--argv0".to_string(), "codex-linux-sandbox".to_string()],
```
bolinfest: Should use the constant here?

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -483,6 +475,46 @@ fn build_bwrap_argv(
...
 
+fn apply_inner_command_argv0(argv: &mut Vec<String>) {
+    apply_inner_command_argv0_for_launcher(
+        argv,
+        preferred_bwrap_supports_argv0(),
+        current_process_argv0(),
+    );
+}
+
+fn apply_inner_command_argv0_for_launcher(
```
bolinfest, line 487: Do existing unit tests cover this function?

viyatb-oai: Yep, this is covered directly by `inserts_bwrap_argv0_before_command_separator`, `rewrites_inner_command_path_when_bwrap_lacks_argv0`, and `rewrites_bwrap_helper_command_not_nested_user_command_when_current_exe_appears_later` in `linux_run_main_tests.rs`.

codex-rs/core/src/tools/runtimes/apply_patch.rs
```diff
@@ -134,13 +129,13 @@ impl Approvable<ApplyPatchRequest> for ApplyPatchRuntime {
         let approval_keys = self.approval_keys(req);
         let changes = req.changes.clone();
         Box::pin(async move {
+            if req.permissions_preapproved && retry_reason.is_none() {
+                return ReviewDecision::Approved;
+            }
             if routes_approval_to_guardian(turn) {
```
bolinfest, line 152: @dylan-hurd-oai it appears `routes_approval_to_guardian()` was introduced in https://github.com/openai/codex/pull/13860 and that it was added _before_ the `req.permissions_preapproved` check, but this PR swaps the ordering.

Codex made this change and it had the following justification, but I wanted to run it by you:

> I moved it earlier because `permissions_preapproved` means “this approval was already satisfied upstream”, not “skip only the user-facing prompt but still run guardian”. The runtime trace made that clear: after the permission grant, the test still emitted `GuardianAssessment`, guardian denied the patch, and the tool returned a rejection instead of patching. So the check has to happen before any approval mechanism, including guardian, or the granted permission has no effect.

dylan-hurd-oai: LGTM, thanks for checking!

codex-rs/arg0/src/lib.rs
```diff
@@ -19,6 +19,7 @@ const TOKIO_WORKER_STACK_SIZE_BYTES: usize = 16 * 1024 * 1024;
 
 #[derive(Clone, Debug, Default, Eq, PartialEq)]
 pub struct Arg0DispatchPaths {
+    pub codex_self_exe: Option<PathBuf>,
```
bolinfest, line 27: I feel like this needs a comment to explain why to use this instead of `std::env::current_exe()`.

codex-rs/core/tests/suite/mod.rs
```diff
@@ -19,7 +22,18 @@ const CODEX_HOME_ENV_VAR: &str = "CODEX_HOME";
 // based on the arg0.
 // NOTE: this doesn't work on ARM
 #[ctor]
-pub static CODEX_ALIASES_TEMP_DIR: TestCodexAliasesGuard = unsafe {
+pub static CODEX_ALIASES_TEMP_DIR: Option<TestCodexAliasesGuard> = unsafe {
```
bolinfest: This use of `unsafe` is overly broad, no?

codex-rs/core/tests/suite/mod.rs
```diff
@@ -19,7 +22,18 @@ const CODEX_HOME_ENV_VAR: &str = "CODEX_HOME";
...
+    let mut args = std::env::args_os();
```
bolinfest, line 26: I feel like this new code block deserves a comment.

PR #15695, opened 2026-03-24
Migrate config loading to be async

pakrym-oai:
Start with a method deep in config loading code (`resolve_root_git_project_for_trust`) and force the call chain to be async


codex-rs/git-utils/src/info.rs
```diff
@@ -618,7 +618,7 @@ async fn diff_against_sha(cwd: &Path, sha: &GitSha) -> Option<String> {
 /// `[get_git_repo_root]`, but resolves to the root of the main
 /// repository. Handles worktrees via filesystem inspection without invoking
 /// the `git` executable.
-pub fn resolve_root_git_project_for_trust(cwd: &Path) -> Option<PathBuf> {
+pub async fn resolve_root_git_project_for_trust(cwd: &Path) -> Option<PathBuf> {
```
bolinfest, line 621: I read the PR description, but to clarify: this is now `async`, yet does not use `await` (as of this PR).

Is this because threading `FileSystem` through all these callsites is planned in the next PR?

Alternatively, would it make sense to create the new version of this function (and all the helper functions it relies on) to use `FileSystem` first?

I'll leave it up to you.

codex-rs/tui/src/onboarding/onboarding_screen.rs
```diff
@@ -74,7 +75,7 @@ pub(crate) struct OnboardingResult {
 }
 
 impl OnboardingScreen {
-    pub(crate) fn new(tui: &mut Tui, args: OnboardingScreenArgs) -> Self {
+    pub(crate) async fn new(tui: &mut Tui, args: OnboardingScreenArgs) -> Self {
```
bolinfest, line 81: We generally try to avoid `async` for code related to rendering, no? (Particularly because `WidgetRef` methods are not `async`?)

codex-rs/tui_app_server/src/onboarding/onboarding_screen.rs
```diff
@@ -438,7 +442,7 @@ pub(crate) async fn run_onboarding_app(
 ) -> Result<OnboardingResult> {
     use tokio_stream::StreamExt;
 
-    let mut onboarding_screen = OnboardingScreen::new(tui, args);
+    let mut onboarding_screen = OnboardingScreen::new(tui, args).await;
```
bolinfest: OK, the blast radius of changing this to `async` isn't that wide...

codex-rs/exec-server/src/lib.rs
```diff
@@ -58,3 +59,5 @@ pub use server::DEFAULT_LISTEN_URL;
 pub use server::ExecServerListenUrlParseError;
 pub use server::run_main;
 pub use server::run_main_with_listen_url;
+
+pub const LOCAL_FS: LocalFileSystem = LocalFileSystem;
```
bolinfest: Do you expect this to continue to be stateless such that it can be a `const` like this? I guess if it changes, we introduce `OnceLock`?

bolinfest (approved):

codex-rs/tui/src/onboarding/onboarding_screen.rs
pakrym-oai: Yeah, as you noticed below it's pretty narrowly scoped.

codex-rs/exec-server/src/lib.rs
pakrym-oai: Correct.

codex-rs/git-utils/src/info.rs
pakrym-oai: >Is this because threading FileSystem through all these callsites is planned in the next PR?

Yes. 

This function was used as a way to pull async through the config loading in preparation to threading FS. Otherwise the change was too large to handle in one go.

PR #15710, opened 2026-03-25
Use AbsolutePathBuf for cwd state

pakrym-oai:
Migrate `cwd` and related session/config state to `AbsolutePathBuf` so downstream consumers consistently see absolute working directories.

Add test-only `.abs()` helpers for `Path`, `PathBuf`, and `TempDir`, and update branch-local tests to use them instead of `AbsolutePathBuf::try_from(...)`.

For the remaining TUI/app-server snapshot coverage that renders absolute cwd values, keep the snapshots unchanged and skip the Windows-only cases where the platform-specific absolute path layout differs.


codex-rs/core/src/guardian/tests.rs
```diff
@@ -25,7 +25,8 @@ use codex_protocol::protocol::EventMsg;
 use codex_protocol::protocol::GuardianAssessmentStatus;
 use codex_protocol::protocol::GuardianRiskLevel;
 use codex_protocol::protocol::ReviewDecision;
-use codex_utils_absolute_path::AbsolutePathBuf;
+use core_test_support::PathBufExt;
```
bolinfest, line 28: This is a nice addition, so I would mention these, and anything else like it, in the PR Body.

codex-rs/tui/src/test_support.rs
```diff
@@ -0,0 +1,32 @@
+use codex_utils_absolute_path::AbsolutePathBuf;
+use std::path::Path;
+use std::path::PathBuf;
+
+pub(crate) trait PathExt {
```
bolinfest, line 5: Should this live in `tui` or a general, reusable location?

bolinfest (approved):

PR #15714, opened 2026-03-25
Fix stale turn steering fallback in tui_app_server

etraut-openai:
This PR adds code to recover from a narrow app-server timing race where a follow-up can be sent after the previous turn has already ended but before the TUI has observed that completion.

Instead of surfacing turn/steer failed: no active turn to steer, the client now treats that as a stale active-turn cache and falls back to starting a fresh turn, matching the intended submit behavior more closely. This is similar to the strategy employed by other app server clients (notably, the IDE extension and desktop app).

This race exists because the current app-server API makes the client choose between two separate RPCs, turn/steer and turn/start, based on its local view of whether a turn is still active. That view is replicated from asynchronous notifications, so it can be stale for a brief window. The server may already have ended the turn while the client still believes it is in progress. Since the choice is made client-side rather than atomically on the server, tui_app_server can occasionally send turn/steer for a turn that no longer exists.

### DIFF
--- a/codex-rs/tui_app_server/src/app.rs
+++ b/codex-rs/tui_app_server/src/app.rs
@@ -621,6 +621,10 @@ impl ThreadEventStore {
     fn active_turn_id(&self) -> Option<&str> {
         self.active_turn_id.as_deref()
     }
+
+    fn clear_active_turn_id(&mut self) {
+        self.active_turn_id = None;
+    }
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


codex-rs/tui_app_server/src/app.rs
```diff
@@ -2021,23 +2032,32 @@ impl App {
...
                                 if !self.chat_widget.enqueue_rejected_steer() {
                                     self.chat_widget.add_error_message(turn_error.message);
                                 }
+                                return Ok(true);
+                            } else if active_turn_missing_steer_error(&error) {
+                                if let Some(channel) = self.thread_event_channels.get(&thread_id) {
+                                    let mut store = channel.store.lock().await;
+                                    store.clear_active_turn_id();
+                                }
+                                should_start_turn = true;
```
dylan-hurd-oai, line 2053: nice graceful recovery

dylan-hurd-oai (approved):

PR #15722, opened 2026-03-25
[plugins] Add a flag for tool search.

mzeng-openai:
- [x] Add a flag for tool search.

### DIFF
--- a/codex-rs/core/tests/suite/search_tool.rs
+++ b/codex-rs/core/tests/suite/search_tool.rs
@@ -86,7 +86,7 @@ fn tool_search_output_tools(request: &ResponsesRequest, call_id: &str) -> Vec<Va
         .unwrap_or_default()
 }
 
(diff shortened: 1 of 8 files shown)


canvrno-oai (approved):

codex-rs/core/tests/suite/search_tool.rs
```diff
@@ -105,6 +105,14 @@ fn configure_apps(config: &mut Config, apps_base_url: &str) {
     config.model_catalog = Some(model_catalog);
 }
 
+fn configure_apps(config: &mut Config, apps_base_url: &str) {
```
morozow, line 108: A mutable 'config' argument breaks platform consistency. It is better to keep data changes at a single, direct entry point. At the same time, binding this function directly to 'Config' type limits future extensibility. A trait-based generic would preserve the same flow while allowing extended mutable config types later.

PR #15730, opened 2026-03-25
fix: harden symlinked project config writes

viyatb-oai:
## Summary
- Explicitly protect `.codex/config.toml` as a read-only leaf in split filesystem policies.
- Preserve intentional pre-existing project config symlinks.
- In bubblewrap, first remount `.codex/` read-only to prevent retargeting, then remount the config symlink's stable target read-only.
- Fail closed for dangling config symlinks or targets that traverse another writable symlink.

## Root Cause
Bubblewrap protects `.codex/` as a read-only directory, but writes to `.codex/config.toml` can still follow a symlink to a writable target outside the protected subtree. Rejecting all symlinked project configs would also reject intentional user configurations. The sandbox instead needs to stabilize the symlink entry before protecting its target.

## Scope
The separate `--output-last-message` hardening has moved to #25956. This PR contains only the `.codex/config.toml` sandbox protection.

## Validation
- `just fmt`
- `just test -p codex-protocol writable_roots_preserve_symlinked_project_config_file`
- Bubblewrap-only tests are Linux-gated and will run in CI


bolinfest:
> Bubblewrap protects .codex/ as a read-only directory, but writes to .codex/config.toml can still follow a symlink to a writable target outside the protected subtree.

Is it possible the user created `.codex/config.toml` as a symlink and they did so intentionally such that we should honor it? That should be acceptable so long as Codex cannot create `.codex/config.toml` as a symlink, correct?

viyatb-oai:
> > Bubblewrap protects .codex/ as a read-only directory, but writes to .codex/config.toml can still follow a symlink to a writable target outside the protected subtree.
> 
> Is it possible the user created `.codex/config.toml` as a symlink and they did so intentionally such that we should honor it? That should be acceptable so long as Codex cannot create `.codex/config.toml` as a symlink, correct?

correct, fixed in https://github.com/openai/codex/pull/15730/commits/b32bd8d514e4ef6c340ad6c4f0c928fd9e798706

PR #15731, opened 2026-03-25
[codex] Require full access justification from requirements

charley-oai:
## Summary
- add an enterprise-only `require_full_access_justification` flag to `requirements.toml`
- require a written justification before switching to full access mode in `tui` and `tui_app_server`
- thread the requirement and enablement details into OTEL and the app-server requirements API/schema

## Why
- admins need a way to require an explicit operator justification before enabling full access mode
- this setting must come from managed requirements, not user `config.toml`

## Impact
- enterprise admins can set `require_full_access_justification = true` in `requirements.toml`
- users are prompted for a justification before enabling full access when the requirement is on
- telemetry now records whether justification was required and whether one was provided

## Root Cause
- full access mode had no admin-controlled requirement for a written justification, and the requirement state was not exposed in telemetry

## Known Issues
- full `cargo test -p codex-core` still has one unrelated sandbox-specific `tools::js_repl::tests::js_repl_imported_local_files_can_access_repl_globals` failure
- full `cargo test -p codex-app-server` still has five unrelated timeout failures in existing websocket and interrupt tests
- per repo instructions, tests were not rerun after `just fix` or `just fmt`

### DIFF
--- a/codex-rs/app-server-protocol/schema/json/codex_app_server_protocol.schemas.json
+++ b/codex-rs/app-server-protocol/schema/json/codex_app_server_protocol.schemas.json
@@ -6871,6 +6871,12 @@
               "object",
               "null"
             ]
+          },
+          "requireFullAccessJustification": {
+            "type": [
+              "boolean",
+              "null"
+            ]
           }
         },
         "type": "object"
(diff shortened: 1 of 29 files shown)


charley-oai:
Disregard both temporary watcher comments above; continuing to babysit CI silently from here.

PR #15772, opened 2026-03-25
Make fuzzy file search case insensitive

meyers-oai:
Makes fuzzy file search use case-insensitive matching instead of smart-case in `codex-file-search`. I find smart-case to be a poor user experience -using the wrong case for a letter drops its match so significantly, it often drops off the results list, effectively making a search case-sensitive.


dylan-hurd-oai (changes requested):
I wouldn't consider this a sufficient argument for moving away from smart case.

dylan-hurd-oai (approved):
Discussed offline - let's ship it!

PR #15791, opened 2026-03-25
fix: resolve bwrap from trusted PATH entry

viyatb-oai:
## Summary
- resolve system bwrap from PATH instead of hardcoding /usr/bin/bwrap
- skip PATH entries that resolve inside the current workspace before launching the sandbox helper
- keep the vendored bubblewrap fallback when no trusted system bwrap is found

## Validation
- cargo test -p codex-core bwrap --lib
- cargo test -p codex-linux-sandbox
- just fix -p codex-core
- just fix -p codex-linux-sandbox
- just fmt
- just argument-comment-lint
- cargo clean

### DIFF
--- a/codex-rs/core/src/config/config_tests.rs
+++ b/codex-rs/core/src/config/config_tests.rs
@@ -5704,8 +5704,8 @@ shell_tool = true
 #[cfg(target_os = "linux")]
 #[test]
 fn system_bwrap_warning_reports_missing_system_bwrap() {
-    let warning = system_bwrap_warning_for_path(Path::new("/definitely/not/a/bwrap"))
-        .expect("missing system bwrap should emit a warning");
+    let warning =
+        system_bwrap_warning_for_lookup(None).expect("missing system bwrap should emit a warning");
 
     assert!(warning.contains("could not find system bubblewrap"));
 }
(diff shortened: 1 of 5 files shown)


codex-rs/core/src/config/mod.rs
```diff
@@ -145,7 +145,7 @@ pub(crate) const DEFAULT_AGENT_JOB_MAX_RUNTIME_SECONDS: Option<u64> = None;
 pub const CONFIG_TOML_FILE: &str = "config.toml";
 const OPENAI_BASE_URL_ENV_VAR: &str = "OPENAI_BASE_URL";
 #[cfg(target_os = "linux")]
-const SYSTEM_BWRAP_PATH: &str = "/usr/bin/bwrap";
+const SYSTEM_BWRAP_PROGRAM: &str = "bwrap";
```
bolinfest, line 148: As a follow-up, we should move all the bwrap stuff out of this file (which is already quite large) so we end up with less `#[cfg(target_os = "linux")]` in here.

codex-rs/core/src/config/config_tests.rs
```diff
@@ -5725,12 +5725,48 @@ exit 1
...
+
+    assert_eq!(
+        find_system_bwrap_in_search_paths(vec![first_dir, second_dir], &cwd),
+        Some(expected_bwrap)
+    );
+}
+
+#[cfg(target_os = "linux")]
+#[test]
+fn skips_workspace_local_bwrap_in_search_paths() {
```
bolinfest, line 5755: All these tests should go in their own file, as well.

bolinfest (approved):

codex-rs/core/src/config/mod.rs
viyatb-oai: Moved this cleanup to follow-up PR #15898 to keep this PR focused on the behavior change.

codex-rs/core/src/config/config_tests.rs
viyatb-oai: Moved this cleanup to follow-up PR #15898 to keep this PR focused on the behavior change.

PR #15798, opened 2026-03-25
Avoid duplicate auth refreshes in `getAuthStatus`

etraut-openai:
I've seen several intermittent failures of `get_auth_status_returns_token_after_proactive_refresh_recovery` today. I investigated, and I found a couple of issues.

First, `getAuthStatus(refreshToken=true)` could refresh twice in one request: once via `refresh_token_if_requested()` and again via the proactive refresh path inside `auth_manager.auth()`. In the permanent-failure case this produced an extra `/oauth/token` call and made the app-server auth tests flaky. Use `auth_cached()` after an explicit refresh request so the handler reuses the post-refresh auth state instead of immediately re-entering proactive refresh logic. Keep the existing proactive path for `refreshToken=false`.

Second, serialize auth refresh attempts in `AuthManager` have a startup/request race. One proactive refresh could already be in flight while a `getAuthStatus(refreshToken=false)` request entered `auth().await`, causing a second `/oauth/token` call before the first failure or refresh result had been recorded. Guarding the refresh flow with a single async lock makes concurrent callers share one refresh result, which prevents duplicate refreshes and stabilizes the proactive-refresh auth tests.

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -1393,12 +1393,6 @@ impl CodexMessageProcessor {
         // then no auth step is required; otherwise, default to requiring auth.
         let requires_openai_auth = self.config.model_provider.requires_openai_auth;
 
-        let auth = if do_refresh {
-            self.auth_manager.auth_cached()
-        } else {
-            self.auth_manager.auth().await
-        };
-
         let response = if !requires_openai_auth {
             GetAuthStatusResponse {
                 auth_method: None,
@@ -1406,6 +1400,11 @@ impl CodexMessageProcessor {
                 requires_openai_auth: Some(false),
             }
         } else {
+            let auth = if do_refresh {
+                self.auth_manager.auth_cached()
+            } else {
+                self.auth_manager.auth().await
+            };
             match auth {
                 Some(auth) => {
                     let permanent_refresh_failure =
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -1393,6 +1393,12 @@ impl CodexMessageProcessor {
         // then no auth step is required; otherwise, default to requiring auth.
         let requires_openai_auth = self.config.model_provider.requires_openai_auth;
 
+        let auth = if do_refresh {
+            self.auth_manager.auth_cached()
+        } else {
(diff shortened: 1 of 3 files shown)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -1393,14 +1393,20 @@ impl CodexMessageProcessor {
         // then no auth step is required; otherwise, default to requiring auth.
         let requires_openai_auth = self.config.model_provider.requires_openai_auth;
 
+        let auth = if do_refresh {
```
etraut-openai: Why is this now being done outside of the "else" block below?

codex-rs/login/src/auth/manager.rs
```diff
@@ -1307,6 +1312,11 @@ impl AuthManager {
     /// can assume that some other instance already refreshed it. If the persisted
     /// token is the same as the cached, then ask the token authority to refresh.
     pub async fn refresh_token(&self) -> Result<(), RefreshTokenError> {
+        let _refresh_guard = self.refresh_lock.lock().await;
+        self.refresh_token_locked().await
+    }
+
+    async fn refresh_token_locked(&self) -> Result<(), RefreshTokenError> {
```
pakrym-oai: quite unfortunate to have two of these methods (one of which doesn't record the error)

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -1400,7 +1400,12 @@ impl CodexMessageProcessor {
                 requires_openai_auth: Some(false),
             }
         } else {
-            match self.auth_manager.auth().await {
+            let auth = if do_refresh {
+                self.auth_manager.auth_cached()
+            } else {
+                self.auth_manager.auth().await
```
pakrym-oai, line 1406: do we still need the         self.refresh_token_if_requested(do_refresh).await; call above?

pakrym-oai (approved):

codex-rs/app-server/src/codex_message_processor.rs
etraut-openai: Yes, it's still needed.

codex-rs/login/src/auth/manager.rs
etraut-openai: Good point. I'll refactor to eliminate the redundancy.

PR #15803, opened 2026-03-25
Update keyring-store JSON helpers to use secret APIs

mikhail-oai:
## Summary
- switch JSON keyring helpers back to `load_secret` and `save_secret`
- keep full and split JSON storage paths aligned with secret-based keyring behavior
- update keyring-store tests and related dependency/lockfile changes

## Testing
- `cargo test -p codex-keyring-store`


celia-oai (changes requested):
let's do this migration another time & keep the surface area to json split for now

PR #15805, opened 2026-03-25
Add MCP connector metrics

nicholasclark-openai:
## Summary
- enrich `codex.mcp.call` with `tool`, `connector_id`, and sanitized `connector_name` for actual MCP executions
- record `codex.mcp.call.duration_ms` for actual MCP executions so connector-level latency is visible in metrics
- keep skipped, blocked, declined, and cancelled paths on the plain status-only `codex.mcp.call` counter

## Included Changes
- `codex-rs/core/src/mcp_tool_call.rs`: add connector-sliced MCP count and duration metrics only for executed tool calls, while leaving non-executed outcomes as status-only counts
- `codex-rs/core/src/mcp_tool_call_tests.rs`: cover metric tag shaping, connector-name sanitization, and the new duration metric tags

## Testing
- `cargo test -p codex-core`
- `just fix -p codex-core`
- `just fmt`

## Notes
- `cargo test -p codex-core` still hits existing unrelated failures in approvals-reviewer config tests and the sandboxed JS REPL `mktemp` test
- full workspace `cargo test` was not run


codex-rs/core/src/mcp_tool_call_tests.rs
```diff
@@ -123,6 +199,77 @@ fn approval_question_text_prepends_safety_reason() {
...
+        tags,
+        vec![
+            ("status", "error".to_string()),
+            ("tool", "echo".to_string()),
+        ]
+    );
+}
+
+#[test]
+fn mcp_call_duration_metric_records_connector_metadata_when_available() {
```
owenlin0: is this test actually useful? it essentially just asserts that `session_telemetry.record_duration(...)` actually does the right thing?

codex-rs/core/src/mcp_tool_call_tests.rs
```diff
@@ -65,6 +77,70 @@ fn prompt_options(
...
+
+fn attributes_to_map<'a>(
+    attributes: impl Iterator<Item = &'a KeyValue>,
+) -> BTreeMap<String, String> {
+    attributes
+        .map(|kv| (kv.key.as_str().to_string(), kv.value.as_str().to_string()))
+        .collect()
+}
+
+fn duration_metric_point(
```
owenlin0: this is all a lot of boilerplate to assert that our OTEL layer actually works as expected, feels unnecessary. can we remove?

codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -214,51 +219,84 @@ pub(crate) async fn handle_mcp_tool_call(
...
+        let tag_refs: Vec<(&str, &str)> = tags.as_ref().map_or_else(Vec::new, |tags| {
+            tags.iter()
+                .map(|(key, value)| (*key, value.as_str()))
+                .collect()
+        });
+        let count_tags = if call_duration.is_some() {
+            tag_refs.as_slice()
+        } else {
+            status_only_tags.as_slice()
+        };
```
owenlin0: this is pretty messy. looks like this complexity is due to conditionally applying tags. for `codex.mcp.call`, can we always tag it with status, tool, connector_id, connector_name?

owenlin0: and reuse the same tags for both metrics?

nicholasclark-openai: It was done to not pollute metrics for tool calls that didn't execute declines/safety/other, but agree it just makes things messy. I'll just skip the duration metric if we didn't execute and keep the tags the same (with status) otherwise

codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -322,13 +352,46 @@ pub(crate) async fn handle_mcp_tool_call(
...
+        .iter()
+        .map(|(key, value)| (*key, value.as_str()))
+        .collect();
+    turn_context
+        .session_telemetry
+        .counter(MCP_CALL_COUNT_METRIC, /*inc*/ 1, &tag_refs);
     turn_context
         .session_telemetry
-        .counter("codex.mcp.call", /*inc*/ 1, &[("status", status)]);
+        .record_duration(MCP_CALL_DURATION_METRIC, duration, &tag_refs);
```
owenlin0: nit: looks like we do this in another place above, we could extract this into a helper function

owenlin0 (approved):
much cleaner, ty!

PR #15806, opened 2026-03-25
Add ReloadUserConfig to tui_app_server

canvrno-oai:
- Adds ReloadUserConfig to `tui_app_server`


etraut-openai (approved):
Looks good. Thanks for adding this!

PR #9193, opened 2026-01-14
fix: `writable_roots` doesn't recognize home directory symbol in non-windows OS

tiffanycitra:
Fixes:
```
[sandbox_workspace_write]
writable_roots = ["~/code/"]
```

translates to
```
/Users/ccunningham/.codex/~/code
```
(i.e. the home dir symbol isn't recognized)

PR #15817, opened 2026-03-26
Expand home-relative paths on Windows

tiffanycitra:
Follow up to: https://github.com/openai/codex/pull/9193, also support this for Windows.


codex-rs/utils/absolute-path/src/lib.rs
```diff
@@ -26,18 +26,19 @@ impl AbsolutePathBuf {
         let Some(path_str) = path.to_str() else {
             return path.to_path_buf();
         };
-        if cfg!(not(target_os = "windows"))
-            && let Some(home) = home_dir()
+        if let Some(home) = home_dir()
```
bolinfest: FYI, you can collapse things so `path.to_path_buf()` only appears once:

```rust
    fn maybe_expand_home_directory(path: &Path) -> PathBuf {
        if let Some(path_str) = path.to_str()
            && let Some(home) = home_dir()
            && let Some(rest) = path_str.strip_prefix('~')
        {
            if rest.is_empty() {
                return home;
            } else if let Some(rest) = rest.strip_prefix('/') {
                return home.join(rest.trim_start_matches('/'));
            } else if cfg!(windows)
                && let Some(rest) = rest.strip_prefix('\\')
            {
                return home.join(rest.trim_start_matches('\\'));
            }
        }

        path.to_path_buf()
    }
```

codex-rs/utils/absolute-path/src/lib.rs
```diff
@@ -26,18 +26,19 @@ impl AbsolutePathBuf {
...
-                    return home;
-                }
-                return home.join(rest);
+            if let Some(rest) = rest.strip_prefix('/') {
+                return home.join(rest.trim_start_matches('/'));
+            }
+            if cfg!(windows)
+                && let Some(rest) = rest.strip_prefix('\\')
+            {
+                return home.join(rest.trim_start_matches('\\'));
```
bolinfest, line 37: So the concern is someone has a double backslash like

```
instructions = "~\\\\instructions.md"
```

which is four backslashes because of the escaping?

Is this something Windows users do?

codex-rs/utils/absolute-path/src/lib.rs
```diff
@@ -26,18 +26,19 @@ impl AbsolutePathBuf {
...
                 return home;
             }
-            if let Some(rest) = path_str.strip_prefix("~/") {
-                let rest = rest.trim_start_matches('/');
-                if rest.is_empty() {
-                    return home;
-                }
-                return home.join(rest);
+            if let Some(rest) = rest.strip_prefix('/') {
+                return home.join(rest.trim_start_matches('/'));
```
bolinfest, line 33: To defend against?

```rust
instructions = "~//instructions.md"
```

I think we should just be strict and if someone does that, it's on them to fix.

codex-rs/utils/absolute-path/src/lib.rs
```diff
@@ -301,16 +299,16 @@ mod tests {
...
+            return;
+        };
         let temp_dir = tempdir().expect("base dir");
-        let base_dir = temp_dir.path();
         let abs_path_buf = {
-            let _guard = AbsolutePathBufGuard::new(base_dir);
-            serde_json::from_str::<AbsolutePathBuf>("\"~/code\"").expect("failed to deserialize")
+            let _guard = AbsolutePathBufGuard::new(temp_dir.path());
+            let input = serde_json::to_string(r#"~\code"#).expect("failed to serialize input");
+            serde_json::from_str::<AbsolutePathBuf>(&input).expect("failed to deserialize")
```
bolinfest: For `expect()`, please express the condition in the positive case rather than the negative one (or just use `unwrap()`):

```suggestion
            let input = serde_json::to_string(r#"~\code"#).expect("string should serialize as JSON");
            serde_json::from_str::<AbsolutePathBuf>(&input).expect("is valid abs path")
```

bolinfest (approved):
Thanks for cleaning this up!

codex-rs/utils/absolute-path/src/lib.rs
tiffanycitra: ah nice, looks much better, will do that.

codex-rs/utils/absolute-path/src/lib.rs
tiffanycitra: yeah, probably not intentionally? it's to protect against the same thing as the linux case that you commented below

codex-rs/utils/absolute-path/src/lib.rs
tiffanycitra: yeah, mostly to protect against `~//` that will be resolved to `/`. we added this safeguard here: https://github.com/openai/codex/pull/9621. thoughts?

PR #15826, opened 2026-03-26
Make AGENTS.md discovery FS-aware

pakrym-oai:
## Summary
- make AGENTS.md discovery and loading fully FS-aware and remove the non-FS discover helper
- migrate remote-aware codex-core tests to use TestEnv workspace setup instead of syncing a local workspace copy
- add AGENTS.md corner-case coverage, including directory fallbacks and remote-aware integration coverage

## Testing
- cargo test -p codex-core project_doc -- --nocapture
- cargo test -p codex-core hierarchical_agents -- --nocapture
- cargo test -p codex-core agents_md -- --nocapture
- cargo test -p codex-tui status -- --nocapture
- cargo test -p codex-tui-app-server status -- --nocapture
- just fix
- just fmt
- just bazel-lock-update
- just bazel-lock-check
- just argument-comment-lint
- remote Linux executor tests in progress via scripts/test-remote-env.sh


codex-rs/core/src/project_doc.rs
```diff
@@ -145,16 +169,22 @@ pub async fn read_project_docs(config: &Config) -> std::io::Result<Option<String
...
+            Ok(_) => {}
+            Err(err) if err.kind() == io::ErrorKind::NotFound => continue,
+            Err(err) => return Err(err),
+        }
 
-        let size = file.metadata().await?.len();
-        let mut reader = tokio::io::BufReader::new(file).take(remaining);
-        let mut data: Vec<u8> = Vec::new();
-        reader.read_to_end(&mut data).await?;
+        let mut data = match fs.read_file(&p).await {
```
bolinfest, line 179: Should `FileSystem` get a `read_head_bytes()` method or something so we don't have to sacrifice the benefits of the shorter reads we have here?

codex-rs/core/src/project_doc.rs
```diff
@@ -211,14 +244,14 @@ pub fn discover_project_doc_paths(config: &Config) -> std::io::Result<Vec<PathBu
     if !project_root_markers.is_empty() {
         for ancestor in dir.ancestors() {
             for marker in &project_root_markers {
-                let marker_path = ancestor.join(marker);
-                let marker_exists = match std::fs::metadata(&marker_path) {
+                let marker_path = AbsolutePathBuf::try_from(ancestor.join(marker))?;
```
bolinfest, line 251: Check out https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=1bcbe382dd8a0d5c75f22d4dc41a392c:

```rust
use std::path::PathBuf;

fn main() {
    let p = PathBuf::from(r#"C:\Users\Alice"#);
    println!("is absolute? {}", p.is_absolute());
}
```

I believe the playground runs on Linux, so this prints `false`, but if it ran on Windows, it would print `true`.

I don't know that we can use things like `AbsolutePathBuf` in their current form with the `FileSystem` abstraction.

codex-rs/tui/src/status/helpers.rs
```diff
@@ -16,6 +19,24 @@ fn normalize_agents_display_path(path: &Path) -> String {
     dunce::simplified(path).display().to_string()
 }
 
+fn discover_local_project_doc_paths(config: &Config) -> io::Result<Vec<AbsolutePathBuf>> {
```
jif-oai: This still uses LOCAL_FS, so in remote exec-server sessions the status card can show different AGENTS files than the ones actually loaded into the model

codex-rs/core/src/project_doc.rs
```diff
@@ -183,10 +213,13 @@ pub async fn read_project_docs(config: &Config) -> std::io::Result<Option<String
 /// contents. The list is ordered from project root to the current working
 /// directory (inclusive). Symlinks are allowed. When `project_doc_max_bytes`
 /// is zero, returns an empty list.
-pub fn discover_project_doc_paths(config: &Config) -> std::io::Result<Vec<PathBuf>> {
-    let mut dir = config.cwd.to_path_buf();
+pub async fn discover_project_doc_paths(
```
jif-oai, line 216: This no longer honors project_doc_max_bytes == 0

codex-rs/tui/src/status/helpers.rs
```diff
@@ -16,6 +19,24 @@ fn normalize_agents_display_path(path: &Path) -> String {
...
+    let config = config.clone();
+    let discover = move || {
```
jif-oai: This is doing per-call thread spawning plus a fresh `tokio` runtime just to bridge the new async discovery API into a synchronous UI helper? This seems pretty odd tbh

codex-rs/core/src/project_doc.rs
pakrym-oai: What am I missing, we still do in the call above read_project_docs_with_fs

jif-oai: The comment says:
```
When `project_doc_max_bytes` is zero, returns an empty list.
```
But this is not true... it will just work as normally

codex-rs/tui/src/status/helpers.rs
pakrym-oai: I'll add a tracking issue.

codex-rs/core/src/project_doc.rs
pakrym-oai: Now it is!

codex-rs/core/src/project_doc.rs
pakrym-oai: I think for AGENTS.md it's not super critical I doubt there setups where agents are large enough to cause concern.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -7360,33 +7360,55 @@ impl ChatWidget {
     }
 
     pub(crate) fn add_status_output(&mut self) {
```
jif-oai: this doesn't seem to follow the remote for the AGENTS... 🤔 where is it?

codex-rs/exec-server/src/local_file_system.rs
```diff
@@ -17,6 +19,9 @@ use crate::RemoveOptions;
 
 const MAX_READ_FILE_BYTES: u64 = 512 * 1024 * 1024;
 
+pub static LOCAL_FS: LazyLock<Arc<dyn ExecutorFileSystem>> =
```
jif-oai, line 22: ultra nit: this create a direct dependency between exec-server and tui. Can we re-export this somewhere else? Either protocol or the app-server or a new thing?

jif-oai (approved):
Maybe an integration tests up to the TUI level?

codex-rs/exec-server/src/local_file_system.rs
pakrym-oai: ooc, why is this a concern?

codex-rs/tui/src/chatwidget.rs
pakrym-oai: It doesn't file a followup.

PR #15888, opened 2026-03-26
Log upstream inline image limit rejections

fjord-oai:
## Summary

Add observability for upstream inline image limit rejections.

This change detects Responses API bad requests that look like inline image byte-limit or image-count-limit failures, logs them with structured fields, and records a Statsig-backed counter so we can monitor how often this is happening before the graceful local handling lands.

## Changes

- Added `inline_image_request_limit_bad_request_observation(...)` in `core/src/api_bridge.rs` to recognize upstream bad-request copy for:
  - inline image byte limit
  - image count limit
- Added structured warning logs in `map_api_error(...)` for those upstream rejections, including:
  - `bytes_exceeded`
  - `images_exceeded`
  - `response_status`
  - `response_body`
- Added client-side metric emission in `core/src/client.rs` when an upstream request is rejected for an inline image limit:
  - metric: `codex.responses.inline_image_limit`
  - outcome: `upstream_rejected`
  - tags: `bytes_exceeded`, `images_exceeded`
- Covered both the parser and metric emission with focused tests in:
  - `core/src/api_bridge_tests.rs`
  - `core/src/client_tests.rs`

- 👉 `1` https://github.com/openai/codex/pull/15888
- ⏳ `2` https://github.com/openai/codex/pull/15569
- ⏳ `3` https://github.com/openai/codex/pull/15887

### DIFF
--- a/codex-rs/core/src/api_bridge.rs
+++ b/codex-rs/core/src/api_bridge.rs
@@ -10,6 +10,7 @@ use codex_login::token_data::PlanType;
 use http::HeaderMap;
 use serde::Deserialize;
 use serde_json::Value;
+use tracing::warn;
 
 use crate::auth::CodexAuth;
 use crate::error::CodexErr;
@@ -17,6 +18,12 @@ use crate::error::RetryLimitReachedError;
 use crate::error::UnexpectedResponseError;
 use crate::error::UsageLimitReachedError;
 use crate::model_provider_info::ModelProviderInfo;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) struct InlineImageRequestLimitBadRequestObservation {
+    pub(crate) bytes_exceeded: bool,
+    pub(crate) images_exceeded: bool,
+}
 
 pub(crate) fn map_api_error(err: ApiError) -> CodexErr {
     match err {
@@ -63,6 +70,17 @@ pub(crate) fn map_api_error(err: ApiError) -> CodexErr {
                         .contains("The image data you provided does not represent a valid image")
                     {
                         CodexErr::InvalidImageRequest()
+                    } else if let Some(observation) =
+                        inline_image_request_limit_bad_request_observation(&body_text)
+                    {
+                        warn!(
+                            response_status = %status,
+                            bytes_exceeded = observation.bytes_exceeded,
+                            images_exceeded = observation.images_exceeded,
+                            response_body = %body_text,
+                            "responses request rejected by upstream inline image limit"
+                        );
+                        CodexErr::InvalidRequest(body_text)
                     } else {
                         CodexErr::InvalidRequest(body_text)
                     }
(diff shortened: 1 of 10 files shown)


codex-rs/core/src/client.rs
```diff
@@ -1185,7 +1233,14 @@ impl ModelClientSession {
...
                 .await
-                .map_err(map_api_error)?;
+                .map_err(|err| {
+                    let mapped = map_api_error(err);
+                    Self::record_upstream_inline_image_request_limit_observation(
+                        session_telemetry,
+                        &mapped,
+                    );
+                    mapped
+                })?;
```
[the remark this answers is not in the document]
fjord-oai: fixed

codex-rs/core/src/client_tests.rs
```diff
@@ -115,3 +179,112 @@ fn auth_request_telemetry_context_tracks_attached_auth_and_retry_phase() {
...
+            &session_telemetry,
+        )
+        .await
+        .expect_err("compact request should be rejected upstream");
+    assert!(matches!(err, crate::error::CodexErr::InvalidRequest(_)));
+
+    let snapshot = session_telemetry
+        .snapshot_metrics()
+        .expect("runtime metrics snapshot");
+    let (attrs, value) = metric_point(&snapshot, INLINE_IMAGE_REQUEST_LIMIT_METRIC);
```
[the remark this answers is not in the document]
fjord-oai: fixed

codex-rs/core/src/client.rs
```diff
@@ -1674,6 +1685,69 @@ fn api_error_http_status(error: &ApiError) -> Option<u16> {
...
+fn record_upstream_inline_image_request_limit_observation(
+    session_telemetry: &SessionTelemetry,
+    observation: crate::api_bridge::InlineImageRequestLimitBadRequestObservation,
+) {
+    session_telemetry.counter(
+        INLINE_IMAGE_REQUEST_LIMIT_METRIC_NAME,
+        /*inc*/ 1,
+        &[
+            (
+                "outcome",
```
pakrym-oai: Does this value add anything?

codex-rs/core/src/api_bridge.rs
```diff
@@ -138,6 +156,24 @@ fn extract_request_tracking_id(headers: Option<&HeaderMap>) -> Option<String> {
     extract_request_id(headers).or_else(|| extract_header(headers, CF_RAY_HEADER))
 }
 
+pub(crate) fn inline_image_request_limit_bad_request_observation(
```
pakrym-oai, line 159: let's add a "well_known_error" enum field to `codex.api_request` metric that is already logged in `record_api_request`. 

we can start with `TooManyImages` or `RequestSizeExceeded` values.

codex-rs/core/src/api_bridge.rs
```diff
@@ -138,6 +156,24 @@ fn extract_request_tracking_id(headers: Option<&HeaderMap>) -> Option<String> {
...
+    body: &str,
+) -> Option<InlineImageRequestLimitBadRequestObservation> {
+    let body = body.to_ascii_lowercase();
+    let bytes_exceeded = body.contains("image data") && body.contains("byte limit");
+    let images_exceeded = body.contains("image limit")
```
pakrym-oai: Can we do much stricter matching of expected errors?

fjord-oai: Found the error codes used for number of images and switched to those instead of string matching. Unfortunately there's not a specific error code for the size limit, so I think this is the best we can do there. I couldn't find the exact place where the size limit is implemented (might be a more general ingress rule?), so I don't think it's easy to add an error code there.

codex-rs/core/src/api_bridge.rs
fjord-oai: done

codex-rs/core/src/client.rs
fjord-oai: This is mostly useful for the next commit in the stack, I removed it from here.

PR #15891, opened 2026-03-26
[plugins] Polish tool suggest prompts.

mzeng-openai:
- [x] Polish tool suggest prompts to distinguish between missing connectors and discoverable plugins, and be very precise about the triggering conditions.

### DIFF
--- a/codex-rs/core/src/tools/spec_tests.rs
+++ b/codex-rs/core/src/tools/spec_tests.rs
@@ -2090,11 +2090,9 @@ fn tool_suggest_can_be_registered_without_search_tool() {
     let ToolSpec::Function(ResponsesApiTool { description, .. }) = &tool_suggest.spec else {
         panic!("expected function tool");
     };
-    assert!(
-        description.contains(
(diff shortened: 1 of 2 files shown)


codex-rs/core/src/tools/spec_tests.rs
```diff
@@ -2092,10 +2092,12 @@ fn tool_suggest_can_be_registered_without_search_tool() {
     };
     assert!(
         description.contains(
-            "You've already tried to find a matching available tool for the user's request"
+            "Suggests a missing connector in an installed plugin, or in narrower cases a discoverable plugin"
         )
```
shijie-oai: discoverable plugins cover installed plugins? maybe we dont need the first part?

mzeng-openai: These are two different things actually:
- Discoverable plugins are plugins that are not installed but discoverable, the intent should match precisely here in order to suggest.
- Missing connectors are connectors that are not installed but needed from an installed plugin

shijie-oai (approved):

owenlin0 (approved):

PR #15898, opened 2026-03-26
chore: move bwrap config helpers into dedicated module

viyatb-oai:
## Summary
- move the bwrap PATH lookup and warning helpers out of config/mod.rs
- move the related tests into a dedicated bwrap_tests.rs file

## Validation
- git diff --check
- skipped heavier local tests per request

Follow-up to #15791.


codex-rs/core/src/config/mod.rs
```diff
@@ -106,6 +106,7 @@ use toml::Value as TomlValue;
 use toml_edit::DocumentMut;
 
 pub(crate) mod agent_roles;
+mod bwrap;
```
bolinfest: Can we put `#[cfg(target_os = "linux")]` on `mod bwrap` and effectively remove all the other ones?

codex-rs/core/src/config/bwrap.rs
```diff
@@ -0,0 +1,58 @@
+#[cfg(target_os = "linux")]
```
bolinfest: Is there a good home for this outside of `codex-core` altogether? We're trying to slim it down.

bolinfest (approved):

PR #15909, opened 2026-03-26
fix(network-proxy): fail closed on network-proxy DNS lookup errors

viyatb-oai:
## Summary

Fail closed when the network proxy's local/private IP pre-check hits a DNS lookup error or timeout, instead of treating the hostname as public and allowing the request.

## Root cause

`host_resolves_to_non_public_ip()` returned `false` on resolver failure, which created a fail-open path in the `allow_local_binding = false` boundary. The eventual connect path performs its own DNS resolution later, so a transient pre-check failure is not evidence that the destination is public.

## Changes

- Treat DNS lookup errors/timeouts as local/private for blocking purposes
- Add a regression test for an allowlisted hostname that fails DNS resolution

## Validation

- `cargo clippy -p codex-network-proxy --all-targets -- -D warnings`
- `just fmt`

### DIFF
--- a/codex-rs/network-proxy/src/runtime.rs
+++ b/codex-rs/network-proxy/src/runtime.rs
@@ -698,9 +698,8 @@ async fn host_resolves_to_non_public_ip(host: &str, port: u16) -> bool {
         return is_non_public_ip(ip);
     }
 
-    // If DNS lookup fails, fail closed rather than allowing the request. The eventual connect path
-    // performs its own resolution, so a transient pre-check failure cannot be treated as proof that
-    // the destination is public.
+    // Block the request if this DNS lookup fails. We resolve the hostname again when we connect,
+    // so a failed check here does not prove the destination is public.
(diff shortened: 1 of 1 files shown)


codex-rs/network-proxy/src/runtime.rs
```diff
@@ -698,12 +698,23 @@ async fn host_resolves_to_non_public_ip(host: &str, port: u16) -> bool {
         return is_non_public_ip(ip);
     }
 
-    // If DNS lookup fails, default to "not local/private" rather than blocking. In practice, the
-    // subsequent connect attempt will fail anyway, and blocking on transient resolver issues would
-    // make the proxy fragile. The allowlist/denylist remains the primary control plane.
+    // If DNS lookup fails, fail closed rather than allowing the request. The eventual connect path
```
rreichel3-oai: Is there a simpler way to say this? Its kinda tough to parse

rreichel3-oai (approved):
LGTM!

PR #15914, opened 2026-03-26
permissions: start migrating runtime state to PermissionProfile

bolinfest:
## Why

`SandboxPolicy` is still acting as the thread-level source of truth for permissions, but it collapses richer filesystem and network state back into a lossy legacy enum. That makes it hard to preserve special filesystem entries and per-turn runtime overrides as we move toward `PermissionProfile` being the primary permission abstraction throughout the codebase.

This PR starts that migration by teaching `PermissionProfile` to represent the current runtime permissions, while keeping the existing constrained runtime policies as the authoritative state in `core`. That avoids introducing duplicated or ambiguous permission fields before `PermissionProfile` is lossless enough to fully replace the legacy runtime model.

## What changed

- replace `FileSystemPermissions { read, write }` with a canonical `entries: Vec<FileSystemSandboxEntry>` representation in `protocol`, while preserving legacy `read`/`write` serde compatibility for existing wire shapes
- add helpers to derive `PermissionProfile` from runtime filesystem and network sandbox policies, plus compatibility helpers to project back to legacy `SandboxPolicy` when needed
- keep `core` runtime state authoritative on the constrained split permission fields (`sandbox_policy`, filesystem, network, and macOS seatbelt extensions) rather than storing a second `permission_profile` field alongside them
- preserve canonical filesystem approval scope across the app-server boundary by emitting `entries` when special paths or deny rules cannot be represented losslessly as legacy `read`/`write` lists
- update the TUI and `tui_app_server` approval overlays to render special filesystem grants and deny entries instead of silently omitting them

## Testing

- `cargo test -p codex-core default_permissions_profile_populates_runtime_sandbox_policy -- --nocapture`
- `cargo test -p codex-core session_configuration_apply_preserves_split_file_system_policy_on_cwd_only_update -- --nocapture`
- `cargo test -p codex-tui -p codex-tui-app-server approval_overlay -- --nocapture`
- `just fix -p codex-app-server-protocol -p codex-tui -p codex-tui-app-server`


codex-rs/sandboxing/src/policy_transforms.rs
```diff
@@ -115,12 +127,13 @@ pub fn intersect_permission_profiles(
...
         .file_system
         .map(|requested_file_system| {
-            let granted_file_system = granted.file_system.unwrap_or_default();
-            let read =
-                intersect_permission_paths(requested_file_system.read, granted_file_system.read);
-            let write =
-                intersect_permission_paths(requested_file_system.write, granted_file_system.write);
-            FileSystemPermissions { read, write }
+            let granted_entries = granted.file_system.unwrap_or_default().entries;
+            let entries = requested_file_system
```
viyatb-oai: [P1] Preserve narrowing entries when clamping approval grants

This exact-entry intersection is no longer a safe cap now that PermissionProfile can contain semantic narrowing entries. If the request is `:root=write` plus `/tmp/secret.txt=none`, and an app-server client responds with only `:root=write`, the returned profile keeps the root write entry and drops the deny. Because `request_permissions_response_from_client_result` relies on this helper to clamp client responses to the original request, a partial or buggy client response can grant more filesystem access than Codex requested. The intersection needs to be semantic, or broad grants must be rejected unless the requested carveouts they dominate are preserved.

codex-rs/core/src/tools/handlers/mod.rs
```diff
@@ -195,13 +195,18 @@ pub(super) async fn apply_granted_turn_permissions(
...
         granted_permissions.as_ref(),
     );
-    let permissions_preapproved = match (effective_permissions.as_ref(), granted_permissions) {
-        (Some(effective_permissions), Some(granted_permissions)) => {
-            intersect_permission_profiles(effective_permissions.clone(), granted_permissions)
-                == *effective_permissions
-        }
-        _ => false,
-    };
+    let session_cwd = session.cwd().await;
```
viyatb-oai: [P2] Do not let preapproved requests narrow sticky grants

The new semantic preapproval check can decide that existing sticky grants cover a fresh request, but the effective permissions passed to the runtime still contain both the fresh request and the sticky grant. With a session grant like `/workspace=write` and a command request like `/workspace/docs=read`, preapproval succeeds, but the effective profile still contains the more-specific read entry; the runtime policy resolves the child path as read-only, so the command loses write access that the session grant already provided. Sticky grants should continue to widen command permissions, not be revoked by a narrower explicit request unless that is a deliberate deny operation.

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1087,22 +1090,46 @@ impl From<CoreNetworkApprovalContext> for NetworkApprovalContext {
...
         }
     }
 }
 
 impl From<AdditionalFileSystemPermissions> for CoreFileSystemPermissions {
     fn from(value: AdditionalFileSystemPermissions) -> Self {
-        Self {
-            read: value.read,
-            write: value.write,
+        match value.entries {
```
viyatb-oai, line 1170: [P2] Do not let empty entries fall back to stale legacy grants

The server can emit mixed permission profiles with both canonical `entries` and legacy `read`/`write` populated. On the response path, though, `entries: []` is treated the same as absent entries and falls back to `read`/`write`. If a client edits or clears the canonical list but leaves the duplicate legacy fields in place, Codex can grant paths the client no longer meant to grant. Treat a present `entries` field as authoritative even when empty, or reject/normalize responses where `entries` and `read`/`write` disagree.

PR #15917, opened 2026-03-26
Support Codex CLI stdin piping for `codex exec`

jliccini:
# Summary

Claude Code supports a useful prompt-plus-stdin workflow:

```bash
echo "complex input..." | claude -p "summarize concisely"
```

Codex previously did not support the equivalent `codex exec` form. While `codex exec` could read the prompt from stdin, it could not combine piped input with an explicit prompt argument.

This change adds that missing workflow:

```bash
echo "complex input..." | codex exec "summarize concisely"
```

With this change, when `codex exec` receives both a positional prompt and piped stdin, the prompt remains the instruction and stdin is passed along as structured `<stdin>...</stdin>` context.

Example:

```bash
curl https://jsonplaceholder.typicode.com/comments \
  | codex exec "format the top 20 items into a markdown table" \
  > table.md
```

This PR also adds regression coverage for:
- prompt argument + piped stdin
- legacy stdin-as-prompt behavior
- `codex exec -` forced-stdin behavior
- empty-stdin error cases

### DIFF
--- a/codex-rs/exec/src/lib.rs
+++ b/codex-rs/exec/src/lib.rs
@@ -1593,12 +1593,17 @@ fn prompt_with_stdin_context(prompt: &str, stdin_text: &str) -> String {
 fn resolve_prompt(prompt_arg: Option<String>) -> String {
     match prompt_arg {
         Some(p) if p != "-" => p,
-        maybe_dash => read_prompt_from_stdin(if matches!(maybe_dash.as_deref(), Some("-")) {
-            StdinPromptBehavior::Forced
-        } else {
-            StdinPromptBehavior::RequiredIfPiped
-        })
-        .expect("required stdin prompt should produce content"),
+        maybe_dash => {
+            let behavior = if matches!(maybe_dash.as_deref(), Some("-")) {
+                StdinPromptBehavior::Forced
+            } else {
+                StdinPromptBehavior::RequiredIfPiped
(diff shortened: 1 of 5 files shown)


codex-rs/exec/src/lib.rs
```diff
@@ -1528,43 +1540,89 @@ fn decode_utf16(
...
+        };
+    }
 
-            let buffer = match decode_prompt_bytes(&bytes) {
-                Ok(s) => s,
-                Err(e) => {
-                    eprintln!("Failed to read prompt from stdin: {e}");
-                    std::process::exit(1);
-                }
+    Some(buffer)
```
bolinfest: Consider eliminating the explicit `return`:

```suggestion
    if buffer.trim().is_empty() {
        match behavior {
            StdinPromptBehavior::OptionalAppend => None,
            StdinPromptBehavior::RequiredIfPiped | StdinPromptBehavior::Forced => {
                eprintln!("No prompt provided via stdin.");
std::process::exit(1);
}
        }
    } else {
        Some(buffer)
    }
```

bolinfest (approved):
Nice work: thanks for the thorough tests!

codex-rs/exec/src/lib.rs
jliccini: ack will do!

PR #15918, opened 2026-03-26
permissions: remove macOS seatbelt extension profiles

bolinfest:
## Why

`PermissionProfile` should only describe the per-command permissions we still want to grant dynamically. Keeping `MacOsSeatbeltProfileExtensions` in that surface forced extra macOS-only approval, protocol, schema, and TUI branches for a capability we no longer want to expose.

## What changed

- Removed the macOS-specific permission-profile types from `codex-protocol`, the app-server v2 API, and the generated schema/TypeScript artifacts.
- Deleted the core and sandboxing plumbing that threaded `MacOsSeatbeltProfileExtensions` through execution requests and seatbelt construction.
- Simplified macOS seatbelt generation so it always includes the fixed read-only preferences allowlist instead of carrying a configurable profile extension.
- Removed the macOS additional-permissions UI/docs/test coverage and deleted the obsolete macOS permission modules.
- Tightened `request_permissions` intersection handling so explicitly empty requested read lists are preserved only when that field was actually granted, avoiding zero-grant responses being stored as active permissions.


celia-oai:
@codex review

codex-rs/sandboxing/src/seatbelt.rs
```diff
@@ -16,12 +15,18 @@ use std::path::PathBuf;
 use tracing::warn;
 use url::Url;
 
-use crate::seatbelt_permissions::build_seatbelt_extensions;
-
 const MACOS_SEATBELT_BASE_POLICY: &str = include_str!("seatbelt_base_policy.sbpl");
 const MACOS_SEATBELT_NETWORK_POLICY: &str = include_str!("seatbelt_network_policy.sbpl");
 const MACOS_RESTRICTED_READ_ONLY_PLATFORM_DEFAULTS: &str =
     include_str!("restricted_read_only_platform_defaults.sbpl");
+const MACOS_DEFAULT_PREFERENCES_POLICY: &str = r#"; allow readonly user preferences
```
celia-oai, line 22: let's just include this as part of MACOS_SEATBELT_BASE_POLICY?

celia-oai (approved):

codex-rs/sandboxing/src/seatbelt.rs
bolinfest: addressing in follow-up https://github.com/openai/codex/pull/15931

PR #15921, opened 2026-03-26
[app-server-protocol] introduce generic ClientResponse for app-server-protocol

rhan-oai:
- introduces `ClientResponse` as the symmetrical typed response union to `ClientRequest` for app-server-protocol 
- enables scalable event stream ingestion for use cases such as analytics
- no runtime behavior changes, protocol/schema plumbing only


owenlin0 (commented):
can we avoid exporting `ClientResponse` actually? clients haven't needed it so I think we  can avoid the bloat in the generated JSON schema / TypeScript types. Let's just use `ClientResponse` internally for the analytics wiring

codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -120,6 +120,41 @@ macro_rules! client_request_definitions {
...
+        }
+
+        impl ClientResponse {
+            pub fn id(&self) -> &RequestId {
+                match self {
+                    $(Self::$variant { request_id, .. } => request_id,)*
+                }
+            }
+
+            pub fn request_method(&self) -> String {
```
owenlin0: nit: just `method` instead of `request_method` imo

owenlin0 (approved):

PR #15922, opened 2026-03-26
Remove the legacy TUI split

etraut-openai:
This is the part 1 of 2 PRs that will delete the `tui` / `tui_app_server` split. This part simply deletes the existing `tui` directory and marks the `tui_app_server` feature flag as removed. I left the `tui_app_server` feature flag in place for now so its presence doesn't result in an error. It is simply ignored.

Part 2 will rename the `tui_app_server` directory `tui`. I did this as two parts to reduce visible code churn.


codex-rs/features/src/tests.rs
```diff
@@ -31,8 +31,11 @@ fn default_enabled_features_are_stable() {
     for spec in crate::FEATURES {
         if spec.default_enabled {
             assert!(
-                matches!(spec.stage, Stage::Stable | Stage::Removed),
-                "feature `{}` is enabled by default but is not stable/removed ({:?})",
+                matches!(
```
etraut-openai: This change is incorrect. Undo this change. Instead, mark the tui_app_server feature flag as Stage::Removed rather than deprecated.

codex-rs/core/src/config/config_tests.rs
```diff
@@ -5692,13 +5693,136 @@ shell_tool = true
     Ok(())
 }
 
+#[cfg(target_os = "linux")]
+#[test]
+fn system_bwrap_warning_reports_missing_system_bwrap() {
```
etraut-openai: Why is this part of this PR? This seems unrelated.

codex-rs/core/src/config/config_tests.rs
```diff
@@ -5692,13 +5693,136 @@ shell_tool = true
...
+    let path = dir.join("bwrap");
+    fs::write(&path, "#!/bin/sh\n").expect("write fake bwrap");
+    let permissions = fs::Permissions::from_mode(0o755);
+    fs::set_permissions(&path, permissions).expect("chmod fake bwrap");
+    path
+}
+
+// Keep these tests independent from any managed/admin config on the host
+// machine so they only exercise the config written under the test codex home.
+fn isolated_loader_overrides(codex_home: &Path) -> LoaderOverrides {
```
etraut-openai: Is this really needed for this PR? Seems unrelated.

codex-rs/tui_app_server/src/bin/codex-tui.rs
```diff
@@ -0,0 +1,42 @@
+use clap::Parser;
```
etraut-openai, line 4: It would be good to include a block header comment explaining the purpose of this new module.

codex-rs/tui_app_server/src/chatwidget/tests.rs
```diff
@@ -192,6 +193,12 @@ async fn test_config() -> Config {
     let codex_home = std::env::temp_dir();
     ConfigBuilder::default()
         .codex_home(codex_home.clone())
+        .loader_overrides(LoaderOverrides {
```
etraut-openai: Why is this change needed in this PR? Seems unrelated.

codex-rs/tui_app_server/tests/suite/model_availability_nux.rs
```diff
@@ -10,7 +10,7 @@ use tokio::time::sleep;
 use tokio::time::timeout;
 
 #[tokio::test]
-async fn resume_startup_does_not_consume_model_availability_nux_count() -> Result<()> {
+async fn resume_startup_ignores_deprecated_tui_app_server_disable_flag() -> Result<()> {
```
etraut-openai: I don't understand why this test was renamed. Unless there's a good reason, revert the name to reduce code churn.

codex-rs/tui_app_server/tests/suite/model_availability_nux.rs
```diff
@@ -112,7 +112,7 @@ trust_level = "trusted"
         "resume".to_string(),
         "--last".to_string(),
         "--no-alt-screen".to_string(),
-        "--enable".to_string(),
```
etraut-openai, line 115: The `tui_app_server` feature flag is a no-op, so remove it entirely from this command-line.

codex-rs/Cargo.toml
```diff
@@ -150,8 +149,7 @@ codex-skills = { path = "skills" }
 codex-state = { path = "state" }
 codex-stdio-to-uds = { path = "stdio-to-uds" }
 codex-terminal-detection = { path = "terminal-detection" }
-codex-tui = { path = "tui" }
-codex-tui-app-server = { path = "tui_app_server" }
+codex-tui = { package = "codex-tui-app-server", path = "tui_app_server" }
```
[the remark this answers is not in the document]
etraut-openai: No need to do this. We're going to follow up this PR shortly with another that will rename the `codex-tui-app-server` package to `codex-tui`.

codex-rs/tui_app_server/src/chatwidget/tests.rs
```diff
@@ -189,9 +189,8 @@ use toml::Value as TomlValue;
 
 async fn test_config() -> Config {
     // Use base defaults to avoid depending on host state.
-    let codex_home = std::env::temp_dir();
```
etraut-openai, line 202: Why did you make this change. Seems like unnecessary churn.

codex-rs/tui_app_server/src/cli.rs
```diff
@@ -32,6 +32,10 @@ pub struct Cli {
     #[clap(skip)]
     pub resume_show_all: bool,
 
+    /// Internal: include non-interactive sessions in resume listings.
+    #[clap(skip)]
+    pub resume_include_non_interactive: bool,
```
bolinfest, line 37: Should this and the changes around this be part of this PR?

etraut-openai: Yes. This is preserving existing `codex resume --include-non-interactive` behavior from the legacy `codex-rs/tui` while this PR removes that implementation. `origin/main` already has the same field in `codex-rs/tui/src/cli.rs`, and the legacy TUI was using it in the resume filtering path, so carrying it into `tui_app_server` is part of the TUI collapse rather than new scope.

bolinfest (approved):
Thanks for the high attention to detail here!

PR #15937, opened 2026-03-27
feat(hooks): add managed hooks

viyatb-oai:
## Summary

Adds a managed-policy control for lifecycle hooks and surfaces a startup warning when hooks are loaded.

## Changes

- Add `allow_managed_hooks_only` / `allowManagedHooksOnly` to config requirements and app-server protocol surfaces.
- Apply the new requirement to `requirements.toml`, MDM-managed `requirements_toml_base64`, and cloud requirements; legacy `managed_config.toml` remains limited to `approval_policy` and `sandbox_mode`.
- Skip non-managed `hooks.json` files when the managed-only requirement is enabled.
- Emit a startup warning that hooks run arbitrary shell commands outside the sandbox and list the loaded `hooks.json` files.
- Show the new requirement in TUI debug config output and document it in `docs/config.md`.

## Why

Hooks are privileged by design and run outside the sandbox. This makes that behavior explicit at startup and gives admins a lockdown option that still allows managed hooks.


eternal-openai (approved):
Amazing work !

PR #15946, opened 2026-03-27
Normalize /mcp tool grouping for hyphenated server names

pakrym-oai:
Fix display for servers with special characters.


etraut-openai (approved):
Thanks for the fix!

PR #15954, opened 2026-03-27
chore: move pty and windows sandbox to Rust 2024

bolinfest:
## Why

`codex-utils-pty` and `codex-windows-sandbox` were the remaining crates in `codex-rs` that still overrode the workspace's Rust 2024 edition. Moving them forward in a separate PR keeps the baseline edition update isolated from the follow-on Bazel clippy workflow in #15955, while making linting and formatting behavior consistent with the rest of the workspace.

This PR also needs Cargo and Bazel to agree on the edition for `codex-windows-sandbox`. Without the Bazel-side sync, the experimental Bazel app-server builds fail once they compile `windows-sandbox-rs`.

## What changed

- switch `codex-rs/utils/pty` and `codex-rs/windows-sandbox-rs` to `edition = "2024"`
- update `codex-utils-pty` callsites and tests to use the collapsed `if let` form that Clippy expects under the new edition
- fix the Rust 2024 fallout in `windows-sandbox-rs`, including the reserved `gen` identifier, `unsafe extern` requirements, and new Clippy findings that surfaced under the edition bump
- keep the edition bump separate from a larger unsafe cleanup by temporarily allowing `unsafe_op_in_unsafe_fn` in the Windows entrypoint modules that now report it under Rust 2024
- update `codex-rs/windows-sandbox-rs/BUILD.bazel` to `crate_edition = "2024"` so Bazel compiles the crate with the same edition as Cargo

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/15954).
* #15976
* #15955
* __->__ #15954

PR #15955, opened 2026-03-27
ci: add Bazel clippy workflow for codex-rs

bolinfest:
## Why
`bazel.yml` already builds and tests the Bazel graph, but `rust-ci.yml` still runs `cargo clippy` separately. This PR starts the transition to a Bazel-backed lint lane for `codex-rs` so we can eventually replace the duplicate Rust build, test, and lint work with Bazel while explicitly keeping the V8 Bazel path out of scope for now.

To make that lane practical, the workflow also needs to look like the Bazel job we already trust. That means sharing the common Bazel setup and invocation logic instead of hand-copying it, and covering the arm64 macOS path in addition to Linux.

Landing the workflow green also required fixing the first lint findings that Bazel surfaced and adding the matching local entrypoint.

## What changed
- add a reusable `build:clippy` config to `.bazelrc` and export `codex-rs/clippy.toml` from `codex-rs/BUILD.bazel` so Bazel can run the repository's existing Clippy policy
- add `just bazel-clippy` so the local developer entrypoint matches the new CI lane
- extend `.github/workflows/bazel.yml` with a dedicated Bazel clippy job for `codex-rs`, scoped to `//codex-rs/... -//codex-rs/v8-poc:all`
- run that clippy job on Linux x64 and arm64 macOS
- factor the shared Bazel workflow setup into `.github/actions/setup-bazel-ci/action.yml` and the shared Bazel invocation logic into `.github/scripts/run-bazel-ci.sh` so the clippy and build/test jobs stay aligned
- fix the first Bazel-clippy findings needed to keep the lane green, including the cross-target `cmsghdr::cmsg_len` normalization in `codex-rs/shell-escalation/src/unix/socket.rs` and the no-`voice-input` dead-code warnings in `codex-rs/tui` and `codex-rs/tui_app_server`

## Verification
- `just bazel-clippy`
- `RUNNER_OS=macOS ./.github/scripts/run-bazel-ci.sh -- build --config=clippy --build_metadata=COMMIT_SHA=local-check --build_metadata=TAG_job=clippy -- //codex-rs/... -//codex-rs/v8-poc:all`
- `bazel build --config=clippy //codex-rs/shell-escalation:shell-escalation`
- `CARGO_TARGET_DIR=/tmp/codex4-shell-escalation-test cargo test -p codex-shell-escalation`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/bazel.yml"); YAML.load_file(".github/actions/setup-bazel-ci/action.yml")'`

## Notes
- `CARGO_TARGET_DIR=/tmp/codex4-tui-app-server-test cargo test -p codex-tui-app-server` still hits existing guardian-approvals test and snapshot failures unrelated to this PR's Bazel-clippy changes.

Related: #15954

### DIFF
--- a/.github/actions/setup-bazel-ci/action.yml
+++ b/.github/actions/setup-bazel-ci/action.yml
@@ -1,0 +1,61 @@
+name: setup-bazel-ci
+description: Prepare a Bazel CI runner with shared caches and optional test prerequisites.
+inputs:
+  target:
+    description: Target triple used for cache namespacing.
+    required: true
+  install-test-prereqs:
+    description: Install Node.js and DotSlash for Bazel-backed test jobs.
+    required: false
+    default: "false"
+outputs:
+  cache-hit:
+    description: Whether the Bazel repository cache key was restored exactly.
+    value: ${{ steps.cache_bazel_repository_restore.outputs.cache-hit }}
+
+runs:
+  using: composite
+  steps:
+    - name: Set up Node.js for js_repl tests
+      if: inputs.install-test-prereqs == 'true'
+      uses: actions/setup-node@v6
+      with:
+        node-version-file: codex-rs/node-version.txt
+
+    # Some integration tests rely on DotSlash being installed.
+    # See https://github.com/openai/codex/pull/7617.
+    - name: Install DotSlash
+      if: inputs.install-test-prereqs == 'true'
+      uses: facebook/install-dotslash@v2
+
+    - name: Make DotSlash available in PATH (Unix)
+      if: inputs.install-test-prereqs == 'true' && runner.os != 'Windows'
+      shell: bash
+      run: cp "$(which dotslash)" /usr/local/bin
+
+    - name: Make DotSlash available in PATH (Windows)
+      if: inputs.install-test-prereqs == 'true' && runner.os == 'Windows'
+      shell: pwsh
+      run: Copy-Item (Get-Command dotslash).Source -Destination "$env:LOCALAPPDATA\Microsoft\WindowsApps\dotslash.exe"
+
+    - name: Set up Bazel
+      uses: bazelbuild/setup-bazelisk@v3
+
+    # Restore bazel repository cache so we don't have to redownload all the external dependencies
+    # on every CI run.
+    - name: Restore bazel repository cache
+      id: cache_bazel_repository_restore
+      uses: actions/cache/restore@v5
+      with:
+        path: |
+          ~/.cache/bazel-repo-cache
+        key: bazel-cache-${{ inputs.target }}-${{ hashFiles('MODULE.bazel', 'codex-rs/Cargo.lock', 'codex-rs/Cargo.toml') }}
+        restore-keys: |
+          bazel-cache-${{ inputs.target }}
+
+    - name: Configure Bazel output root (Windows)
+      if: runner.os == 'Windows'
+      shell: pwsh
+      run: |
+        # Use a very short path to reduce argv/path length issues.
+        "BAZEL_OUTPUT_USER_ROOT=C:\" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
--- a/.github/scripts/run-bazel-ci.sh
+++ b/.github/scripts/run-bazel-ci.sh
@@ -1,0 +1,178 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+print_failed_bazel_test_logs=0
+use_node_test_env=0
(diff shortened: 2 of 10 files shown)


aibrahim-oai (approved):

.github/actions/setup-bazel-ci/action.yml
```diff
@@ -0,0 +1,61 @@
...
+    # Some integration tests rely on DotSlash being installed.
+    # See https://github.com/openai/codex/pull/7617.
+    - name: Install DotSlash
+      if: inputs.install-test-prereqs == 'true'
+      uses: facebook/install-dotslash@v2
+
+    - name: Make DotSlash available in PATH (Unix)
+      if: inputs.install-test-prereqs == 'true' && runner.os != 'Windows'
+      shell: bash
+      run: cp "$(which dotslash)" /usr/local/bin
```
jif-oai, line 34: That sounds a bit hacky no? Why not just updating PATH? 🤔 
Also, I though `which` was considered flaky in bash (we might not care as we are in controlled env though)

.github/actions/setup-bazel-ci/action.yml
```diff
@@ -0,0 +1,61 @@
...
+
+    # Restore bazel repository cache so we don't have to redownload all the external dependencies
+    # on every CI run.
+    - name: Restore bazel repository cache
+      id: cache_bazel_repository_restore
+      uses: actions/cache/restore@v5
+      with:
+        path: |
+          ~/.cache/bazel-repo-cache
+        key: bazel-cache-${{ inputs.target }}-${{ hashFiles('MODULE.bazel', 'codex-rs/Cargo.lock', 'codex-rs/Cargo.toml') }}
```
jif-oai, line 52: OOC, why not using `MODULE.bazel.lock` ?

.github/scripts/run-bazel-ci.sh
```diff
@@ -0,0 +1,179 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+set -o pipefail
```
jif-oai: OOC, isn't it redundant with the previous line?

.github/scripts/run-bazel-ci.sh
```diff
@@ -0,0 +1,179 @@
...
+done
+
+if [[ $# -eq 0 ]]; then
+  echo "Usage: $0 [--print-failed-test-logs] [--use-node-test-env] -- <bazel args> -- <targets>" >&2
+  exit 1
+fi
+
+bazel_startup_args=()
+if [[ -n "${BAZEL_STARTUP_ARGS:-}" ]]; then
+  read -r -a bazel_startup_args <<< "${BAZEL_STARTUP_ARGS}"
```
jif-oai: I think this will break if you have space in the startup args

Something like
```
--host_jvm_args=-Dfoo="bar baz"
```

Probably not a big deal though

jif-oai (approved):

.github/actions/setup-bazel-ci/action.yml
bolinfest: Fwiw, we were already doing this in `.github/workflows/bazel.yml`.

I believe I did this originally because updating `PATH` here doesn't propagate to subsequent steps and the bookkeeping to do it that way isn't worth it.

Or at least, `facebook/install-dotslash@v2` normally does seem to work that way, but something (Bazel?) I think uses a more sanitized version of `PATH` that excludes the entry where GitHub Actions normally introduce executables.

.github/actions/setup-bazel-ci/action.yml
bolinfest: @siggisim This is from https://github.com/openai/codex/pull/14495, can you speak to this?

.github/scripts/run-bazel-ci.sh
bolinfest: Good catch!

.github/scripts/run-bazel-ci.sh
bolinfest: Great catch! Asked Codex to fix.

Though maybe the real fix is to rewrite this file in Python...

.github/actions/setup-bazel-ci/action.yml
siggisim: Either should be fine, especially since the lock file is checked in here. `MODULE.bazel.lock` will probably get you more granular buckets which is nice (but since items in the repository cache are keyed under a hash of contents, it shouldn't be a big deal either way).

PR #15977, opened 2026-03-27
fix(permissions): preserve managed deny-read during escalation

viyatb-oai:
## Why

Managed filesystem `deny_read` requirements are administrator-enforced restrictions on specific paths. Once those requirements are active, Codex should not drop them just because an execution path would otherwise leave the sandbox.

Before this change, an explicit escalation, a prefix-rule allow, a sandbox-denial retry, or an app-server legacy sandbox override could rebuild the runtime policy without those managed read-deny entries and expose a path the administrator had marked unreadable.

This is narrower than general sandbox-mode constraints. If an enterprise only sets `allowed_sandbox_modes`, a trusted `prefix_rule(..., decision = "allow")` can still run its matching command unsandboxed; this PR only preserves managed filesystem `deny_read` restrictions across those paths.

## What Changed

- Mark filesystem policies built from managed `deny_read` requirements so callers can tell when those deny entries must survive escalation.
- Preserve managed deny-read entries when runtime permission profiles are rebuilt through protocol, app-server, or legacy sandbox-policy compatibility paths.
- Keep managed deny-read attempts inside the selected sandbox on the first attempt and after sandbox-denial retries.
- Preserve the same behavior in the zsh-fork escalation path, including prefix-rule-driven escalation.
- Add a regression test showing the opposite case too: without managed deny-read, a prefix-rule allow still chooses unsandboxed execution.

## Verification

Targeted automated verification:

```shell
cargo test -p codex-core shell_request_escalation_execution_is_explicit -- --nocapture
cargo test -p codex-core prefix_rule_uses_unsandboxed_execution_without_managed_deny_read -- --nocapture
cargo test -p codex-core prefix_rule_preserves_managed_deny_read_escalation -- --nocapture
cargo test -p codex-protocol permission_profile_round_trip_preserves_filesystem_policy_metadata -- --nocapture
cargo test -p codex-protocol preserving_deny_entries_keeps_unrestricted_policy_enforceable -- --nocapture
cargo test -p codex-app-server-protocol permission_profile_file_system_permissions_preserves_policy_metadata -- --nocapture
cargo check -p codex-app-server -p codex-tui
```

Smoke-test invocations:

```shell
# macOS exact deny + allowed control
codex exec --skip-git-repo-check -C "$ROOT" \
  -c 'default_permissions="deny_read_smoke"' \
  -c 'permissions.deny_read_smoke.filesystem={":minimal"="read",":project_roots"={"."="write","secrets"="none","future-secret"="none","**/*.env"="none"}}' \
  'Run shell commands only. Print the contents of allowed.txt. Then test whether reading secrets/exact-secret.txt succeeds without printing that file if it does. End with exactly two lines: allowed=<contents> and exact_secret=<BLOCKED or READABLE>.'

# Linux exact deny + allowed control
codex exec --skip-git-repo-check -C "$ROOT" \
  -c 'default_permissions="deny_read_smoke"' \
  -c 'permissions.deny_read_smoke.filesystem={":minimal"="read",glob_scan_max_depth=3,":project_roots"={"."="write","secrets"="none","future-secret"="none","**/*.env"="none"}}' \
  'Run shell commands only. Print the contents of allowed.txt. Then test whether reading secrets/exact-secret.txt succeeds without printing that file if it does. End with exactly two lines: allowed=<contents> and exact_secret=<BLOCKED or READABLE>.'
```

Observed manual smoke matrix:

| Case | macOS Seatbelt | Linux bubblewrap |
| --- | --- | --- |
| `cat allowed.txt` | Pass | Pass |
| `cat secrets/exact-secret.txt` | Blocked | Blocked |
| `cat envs/root.env` | Blocked | Blocked |
| `cat envs/nested/one.env` | Blocked | Blocked |
| `cat envs/nested/two.env` | Blocked | Blocked |
| `cat alias-to-secrets/exact-secret.txt` | Blocked | Blocked |
| Missing denied path | A file created after sandbox setup remained unreadable | Creation was blocked by the reserved missing-path placeholder, and the placeholder was cleaned up after exit |
| Real `codex exec` shell turn | Pass | Pass |

Notes:

- The Linux smoke run used the fallback glob walker because the devbox did not have `rg` installed.
- The smoke matrix verifies the end-to-end filesystem behavior on macOS and Linux; the escalation-specific behavior is covered by the focused tests above.


codex-rs/core/src/tools/sandboxing_tests.rs
```diff
@@ -104,7 +109,41 @@ fn guardian_bypasses_sandbox_for_explicit_escalation_on_first_attempt() {
...
+    assert_eq!(
+        sandbox_override_for_first_attempt(
+            SandboxPermissions::RequireEscalated,
+            &ExecApprovalRequirement::Skip {
+                bypass_sandbox: false,
+                proposed_execpolicy_amendment: None,
+            },
+            &file_system_policy,
+        ),
+        SandboxOverride::NoOverride
```
bolinfest: Please add messages to each `assert_eq!()` to clarify why the expected value is what it is.

codex-rs/core/src/exec_policy_tests.rs
```diff
@@ -1479,6 +1479,36 @@ async fn proposed_execpolicy_amendment_is_suppressed_when_policy_matches_allow()
     .await;
 }
 
+#[tokio::test]
+async fn allow_rule_does_not_bypass_sandbox_when_deny_read_paths_exist() {
+    assert_exec_approval_requirement_for_command(
+        ExecApprovalRequirementScenario {
+            policy_src: Some(r#"prefix_rule(pattern=["echo"], decision="allow")"#.to_string()),
```
bolinfest: `echo` is not a great example here: choose something that is not in `is_known_safe_command()`?

codex-rs/core/src/exec_policy_tests.rs
```diff
@@ -1479,6 +1479,36 @@ async fn proposed_execpolicy_amendment_is_suppressed_when_policy_matches_allow()
     .await;
 }
 
+#[tokio::test]
+async fn allow_rule_does_not_bypass_sandbox_when_deny_read_paths_exist() {
```
bolinfest: But the point of `prefix_rule()` is that it allows a command to run without a sandbox, no?

I feel like it's quite likely that `**/*.env` could be in the deny list, but a command with a `prefix_rule()` may very well need to be able to read those files?

viyatb-oai: agreed, fixed!

codex-rs/app-server-protocol/src/protocol/v2.rs
bolinfest: Incidentally, I think we should be eliminating this level of detail from the app server API. Clients should only know about the permission profile name. Also whether it is a custom profile or a built-in one, though that is derivable from the `:` prefix today.

codex-rs/core/src/tools/runtimes/shell.rs
```diff
@@ -209,8 +207,8 @@ impl Approvable<ShellRequest> for ShellRuntime {
         ))
     }
 
-    fn sandbox_mode_for_first_attempt(&self, req: &ShellRequest) -> SandboxOverride {
-        sandbox_override_for_first_attempt(req.sandbox_permissions, &req.exec_approval_requirement)
+    fn sandbox_permissions(&self, req: &ShellRequest) -> SandboxPermissions {
```
bolinfest, line 211: This function doesn't really seem to do anything at this point: can we inline all calls to it and eliminate it?

codex-rs/core/src/tools/runtimes/unified_exec.rs
```diff
@@ -211,8 +209,8 @@ impl Approvable<UnifiedExecRequest> for UnifiedExecRuntime<'_> {
         ))
     }
 
-    fn sandbox_mode_for_first_attempt(&self, req: &UnifiedExecRequest) -> SandboxOverride {
-        sandbox_override_for_first_attempt(req.sandbox_permissions, &req.exec_approval_requirement)
+    fn sandbox_permissions(&self, req: &UnifiedExecRequest) -> SandboxPermissions {
```
bolinfest, line 213: Same here?

codex-rs/core/src/tools/orchestrator.rs
```diff
@@ -148,6 +149,11 @@ impl ToolOrchestrator {
         let requirement = tool.exec_approval_requirement(req).unwrap_or_else(|| {
             default_exec_approval_requirement(approval_policy, &file_system_sandbox_policy)
         });
+        let sandbox_override = sandbox_override_for_first_attempt(
```
bolinfest: Can we compute this just before it is used (so after the `match` statement)?

codex-rs/core/src/tools/sandboxing_tests.rs
```diff
@@ -134,7 +139,43 @@ fn guardian_bypasses_sandbox_for_explicit_escalation_on_first_attempt() {
...
 }
+
+#[test]
+fn deny_read_blocks_explicit_escalation_but_preserves_policy_bypass() {
+    let file_system_policy = FileSystemSandboxPolicy::restricted(vec![FileSystemSandboxEntry {
+        path: FileSystemPath::Special {
+            value: FileSystemSpecialPath::Root,
+        },
+        access: FileSystemAccessMode::None,
+    }]);
```
bolinfest, line 154: Not that it's terribly important, but this isn't a realistic example, though I get it's easy to make the test work cross-platform this way.

bolinfest (approved):

PR #15979, opened 2026-03-27
feat(permissions): add glob deny-read policy support

viyatb-oai:
## Summary
- adds first-class filesystem policy entries for deny-read glob patterns
- parses config such as :project_roots { "**/*.env" = "none" } into pattern entries
- enforces deny-read patterns in direct read/list helpers
- fails closed for sandbox execution until platform backends enforce glob patterns in #18096
- preserves split filesystem policy in turn context only when it cannot be reconstructed from legacy sandbox policy

## Stack
1. This PR - glob deny-read policy/config/direct-tool support
2. #18096 - macOS and Linux sandbox enforcement
3. #17740 - managed deny-read requirements

## Verification
- just fmt
- cargo check -p codex-core -p codex-sandboxing --tests


codex-rs/core/src/config/config_tests.rs
```diff
@@ -657,6 +657,74 @@ async fn default_permissions_profile_populates_runtime_sandbox_policy() -> std::
     Ok(())
 }
 
+#[tokio::test]
+async fn project_root_glob_none_compiles_to_filesystem_pattern_entry() -> std::io::Result<()> {
+    let codex_home = TempDir::new()?;
+    let cwd = TempDir::new()?;
+    std::fs::write(cwd.path().join(".git"), "gitdir: nowhere")?;
```
bolinfest: nit: prefer `tokio::fs` when using an `async` test?

codex-rs/core/src/config/permissions.rs
```diff
@@ -54,8 +57,24 @@ pub(crate) fn compile_permission_profile(
                 missing_filesystem_entries_warning(profile_name),
             );
         } else {
+            if cfg!(any(target_os = "linux", target_os = "windows")) {
```
bolinfest: This is slightly more accurate, no?

```suggestion
            if cfg!(not(target_os = "macos")) {
```

codex-rs/core/src/config/permissions.rs
```diff
@@ -118,26 +136,71 @@ fn compile_network_sandbox_policy(network: Option<&NetworkToml>) -> NetworkSandb
...
+                    // exact-path parser so existing path semantics stay intact.
+                    let entry = FileSystemSandboxEntry {
+                        path: FileSystemPath::Pattern {
+                            pattern: compile_scoped_filesystem_pattern(
+                                path, subpath, *access, cwd,
+                            )?,
+                        },
+                        access: *access,
+                    };
+                    if !entries.iter().any(|existing| existing == &entry) {
```
bolinfest: On `main`, it appears `compile_permission_profile()` is the only caller of this function and it passes in an empty `Vec::new()`, so perhaps we should create the `Vec` within this function so we can be sure that, by construction, we are never adding duplicates and can skip this check?

codex-rs/core/src/config/permissions.rs
```diff
@@ -118,26 +136,71 @@ fn compile_network_sandbox_policy(network: Option<&NetworkToml>) -> NetworkSandb
...
+    access: FileSystemAccessMode,
+    startup_warnings: &mut Vec<String>,
+) -> io::Result<FileSystemPath> {
+    if !contains_glob_chars(path) {
+        return compile_filesystem_path(path, startup_warnings);
+    }
+
+    if access == FileSystemAccessMode::None {
+        return Ok(FileSystemPath::Pattern {
+            pattern: parse_absolute_path(path)?.to_string_lossy().into_owned(),
```
bolinfest, line 197: So even though an entry in `config.toml` could be a relative path, `path` has been resolved to an absolute path at this point, is that correct?

codex-rs/core/src/config/permissions.rs
```diff
@@ -186,6 +249,91 @@ fn compile_scoped_filesystem_path(
     Ok(FileSystemPath::Path { path })
 }
 
+fn compile_scoped_filesystem_pattern(
+    path: &str,
+    subpath: &str,
+    access: FileSystemAccessMode,
+    cwd: &Path,
```
bolinfest: Is this `cwd` or `project_root`?

codex-rs/core/src/config/permissions_tests.rs
```diff
@@ -216,3 +222,87 @@ fn network_toml_overlays_unix_socket_permissions_by_path() {
...
+                FilesystemPermissionToml::Access(FileSystemAccessMode::Read),
+            ),
+            (
+                "/tmp/cache/**".to_string(),
+                FilesystemPermissionToml::Access(FileSystemAccessMode::Write),
+            ),
+            (
+                ":project_roots".to_string(),
+                FilesystemPermissionToml::Scoped(BTreeMap::from([
+                    ("**/*.env".to_string(), FileSystemAccessMode::None),
```
bolinfest, line 241: Why is this supported?

codex-rs/core/src/config/permissions_tests.rs
```diff
@@ -216,3 +222,87 @@ fn network_toml_overlays_unix_socket_permissions_by_path() {
...
+    )?;
+
+    assert_eq!(
+        file_system_policy,
+        FileSystemSandboxPolicy::restricted(vec![FileSystemSandboxEntry {
+            path: FileSystemPath::Special {
+                value: FileSystemSpecialPath::project_roots(Some("docs".into())),
+            },
+            access: FileSystemAccessMode::Read,
+        }])
```
bolinfest: Add a message to clarify what we are asserting?

codex-rs/core/src/tools/handlers/list_dir.rs
```diff
@@ -96,23 +99,46 @@ impl ToolHandler for ListDirHandler {
...
+            list_dir_slice_with_policy(&path, offset, limit, depth, read_deny_matcher.as_ref())
+                .await?;
         let mut output = Vec::with_capacity(entries.len() + 1);
         output.push(format!("Absolute path: {}", path.display()));
         output.extend(entries);
         Ok(FunctionToolOutput::from_text(output.join("\n"), Some(true)))
     }
 }
 
+#[cfg(test)]
```
bolinfest: Can/should this be moved into `list_dir_tests.rs` instead?

codex-rs/core/src/codex.rs
```diff
@@ -1082,6 +1082,14 @@ impl TurnContext {
     }
 
     pub(crate) fn to_turn_context_item(&self) -> TurnContextItem {
+        let legacy_file_system_sandbox_policy = FileSystemSandboxPolicy::from_legacy_sandbox_policy(
+            self.sandbox_policy.get(),
+            &self.cwd,
+        );
+        let file_system_sandbox_policy = (self.file_system_sandbox_policy
```
bolinfest, line 1093: Maybe add a comment about why we're doing this and note that we're hoping to eliminate this shortly?

codex-rs/protocol/src/permissions.rs
```diff
@@ -154,14 +156,97 @@ struct FileSystemSemanticSignature {
...
+
+/// Runtime matcher for read-deny entries in a filesystem sandbox policy.
+pub struct ReadDenyMatcher {
+    denied_candidates: Vec<Vec<PathBuf>>,
+    deny_read_matchers: Vec<GlobMatcher>,
+    invalid_pattern: bool,
+}
+
+impl ReadDenyMatcher {
+    /// Builds a matcher from exact deny-read roots and deny-read glob entries.
```
bolinfest, line 170: Expand docstring: what does it mean when this function returns `None`?

codex-rs/protocol/src/permissions.rs
```diff
@@ -154,14 +156,97 @@ struct FileSystemSemanticSignature {
...
+            .collect();
+        Some(Self {
+            denied_candidates,
+            deny_read_matchers,
+            invalid_pattern,
+        })
+    }
+
+    /// Returns whether `path` is denied by the policy used to build this matcher.
+    pub fn is_read_denied(&self, path: &Path) -> bool {
```
bolinfest, line 210: Can be done later, but I'm curious if we will do any caching around this function because we need to check this every time we configure a sandbox for a shell tool call, right?

codex-rs/protocol/src/permissions.rs
```diff
@@ -154,14 +156,97 @@ struct FileSystemSemanticSignature {
...
 #[ts(tag = "type")]
 pub enum FileSystemPath {
-    Path { path: AbsolutePathBuf },
-    Special { value: FileSystemSpecialPath },
+    Path {
+        path: AbsolutePathBuf,
+    },
+    /// A git-style glob pattern. Pattern entries currently support
+    /// FileSystemAccessMode::None only.
+    Pattern {
```
bolinfest: Would prefer to name this `GlobPattern` so it's clear when reading.

codex-rs/protocol/src/permissions.rs
bolinfest, line 1427: In a follow-up: we should convert this crate so that tests are in a separate `_tests.rs` file.

codex-rs/protocol/src/permissions.rs
```diff
@@ -2061,4 +2239,186 @@ mod tests {
...
+        std::fs::create_dir_all(&denied_dir).expect("create denied dir");
+        std::fs::write(&nested, "secret").expect("write secret");
+
+        let policy = deny_policy(&denied_dir);
+        assert_eq!(is_read_denied(&denied_dir, &policy, temp.path()), true);
+        assert_eq!(is_read_denied(&nested, &policy, temp.path()), true);
+        assert_eq!(
+            is_read_denied(&temp.path().join("other.txt"), &policy, temp.path()),
+            false
+        );
```
bolinfest: Using `assert!()` is more canonical Rust, I think (here and below):

```suggestion
        assert!(is_read_denied(&denied_dir, &policy, temp.path()));
        assert!(is_read_denied(&nested, &policy, temp.path()));
        assert!(!is_read_denied(&temp.path().join("other.txt"), &policy, temp.path()));
```

codex-rs/protocol/src/permissions.rs
```diff
@@ -2061,4 +2239,186 @@ mod tests {
...
+
+    #[test]
+    fn glob_patterns_deny_matching_paths() {
+        let temp = TempDir::new().expect("tempdir");
+        let denied = temp.path().join("private").join("secret1.txt");
+        std::fs::create_dir_all(denied.parent().expect("parent")).expect("create parent");
+        std::fs::write(&denied, "secret").expect("write secret");
+
+        let policy = default_policy_with_unreadable_glob(format!(
+            "{}/private/secret?.txt",
```
bolinfest, line 2368: I'm curious if `?` is important to support, in practice. If not, maybe we should skip it so it's easier to support this contract across all three OSes?

bolinfest (approved):

Issue #15725, opened 2026-03-25
Linux sandbox fails with "bwrap: Can't create file ... Is a directory" when protected workspace path is a symlinked directory


Harag:
### What version of Codex CLI is running?

codex-cli 0.116.0

### What subscription do you have?

Pro

### Which model were you using?

gpt-5.4

### What platform is your computer?

Linux 6.17.0-19-generic x86_64 GNU/Linux

### What terminal emulator and version are you using (if applicable)?

bash on Ubuntu Linux

### What issue are you seeing?

In my repro, the workspace root contains:

  .agents -> some-folder/.agents

  When Codex runs a normal shell tool in sandboxed mode, bubblewrap fails immediately with:

  bwrap: Can't create file at .../.agents: Is a directory

  This is not a normal permission denial. The problem is that the Linux sandbox is treating the protected alias like a file path and tries to mask it with a file-style bind, but the
  alias resolves to a directory. bubblewrap rejects that path-type mismatch before the user command runs.

  This may be in the same general Linux/bwrap/workspace-write area as other sandbox failures, but the exact symptom here is different from namespace creation failures: this repro is
  specifically an `Is a directory` error caused by a symlinked directory alias at the workspace root.

### What steps can reproduce the bug?

1. On Linux, create a workspace where the root contains a symlink to a directory, for example:
     mkdir -p some-folder/.agents
     ln -s some-folder/.agents .agents

  2. Start Codex CLI in that workspace with a sandboxed mode that uses the Linux bubblewrap backend.

  3. Run any normal shell command through Codex, for example:
     pwd
     rg --files
     sed -n '1,20p' some-file
     git status

  4. Observe that the command fails before execution with:
     bwrap: Can't create file at .../.agents: Is a directory

  5. Switch to danger-full-access, or remove/rename the root `.agents` symlink, and the same command succeeds.

### What is the expected behavior?

Sandboxed shell execution should still work when a protected workspace-root alias is a symlink to a directory.

  If Codex needs to make that path read-only/unreadable inside bubblewrap, it should preserve the path type and bind the resolved directory appropriately instead of trying to file-mask
  the alias. Routine shell commands should not fail before execution because a protected alias points to a directory.

### Additional information

I tested a local fix and it resolves the repro.

  Tested diff:
  - Fork branch: https://github.com/Harag/codex/tree/local/bwrap-symlink-directory-fix
  - Commit: a143a5d52
  - Repo: https://github.com/Harag/codex.git

  Local patch artifacts:
[...]

Issue #14694, opened 2026-03-14
apply_patch fails when ~/.codex is a symlinked directory


rebroad:
## Summary
When `~/.codex` is a symlink to another partition, `apply_patch` fails in the sandbox. This started after the memories feature was introduced (it appears to create/use `~/.codex/memories`).

## Version
`codex` v0.115.0-alpha.19

## Repro
1. Make `~/.codex` a symlink to another partition (e.g. `/mnt/p8/@home/rebroad/.codex`).
2. Run any `apply_patch` operation.

## Actual
`bwrap` errors out during sandbox setup:

```
bwrap: Can't mkdir /home/rebroad/.codex/memories: No such file or directory
```

Earlier it also failed with:

```
bwrap: Can't bind mount /oldroot/mnt/p8/@home/rebroad/SunOS on /newroot/home/rebroad/SunOS: Unable to mount source on destination: No such file or directory
```

## Expected
`apply_patch` should work regardless of whether `~/.codex` is a symlink. The sandbox setup should resolve symlinks or create needed paths inside the sandbox appropriately.

## Notes
This appears to have regressed after the new memories feature that uses `~/.codex/memories`.
The commit that introduced `create_dir_all(&memories_root)` and added the memories root to workspace-write is:
`f72ab43fd193b31208cd3c306293b1b71a52a709` (2026-03-04 13:00:26 +0000, "feat: memories in workspace write (#13467)").

etraut-openai:
Ah, good point. Reopening this one.

PR #15981, opened 2026-03-27
fix(permissions): fix symlinked writable roots in sandbox permissions

### LINKED ISSUE #17079: Linux bwrap fails when protected symlinked carveout points outside writable roots (No such file or directory)
### Summary
Linux sandbox startup can fail when a protected read-only carveout path (for example `.git`) is a symlink under a writable root and that symlink points to an absolute target outside writable roots.

`bwrap` is currently invoked with a mask like:

`--ro-bind /dev/null /var/tmp/src-builds/codex.other/.git`

In this scenario, preflight can fail before any command/tool runs with:

`bwrap: Can't create file at /var/tmp/src-builds/codex.other/.git: No such file or directory`

### Repro shape
- `workspace-write` sandbox with writable root that canonicalizes to `/var/tmp/src-builds/codex.other`
- protected carveouts include `.../.git`
- `.../.git` is a symlink to an absolute path outside writable roots

Example:
- `/var/tmp/src-builds/codex.other/.git -> /home/rebroad/src/codex/.git`

### Observed behavior
- non-escalated commands (including `apply_patch`) fail at sandbox preflight
- error appears before command execution

### Expected behavior
Sandbox should start successfully. It should not hard-fail when masking a symlinked protected path in this layout.

### Related issues
- #15725 (symlinked directory carveout variant)
- #16161 (`.git` symlink `Is a directory` variant)

This appears to be a related but distinct failure mode (`No such file or directory`) triggered by symlink target layout and mount semantics.

viyatb-oai:
## Summary
- preserve logical symlink paths during permission normalization and config cwd handling
- bind real targets for symlinked readable/writable roots in bwrap and remap carveouts and unreadable roots there
- add regressions for symlinked carveouts and nested symlink escape masking

## Root cause
Permission normalization canonicalized symlinked writable roots and cwd to their real targets too early. That drifted policy checks away from the logical paths the sandboxed process can actually address, while bwrap still needed the real targets for mounts. The mismatch caused shell and apply_patch failures on symlinked writable roots.

## Impact
Fixes #15781.

Also fixes #17079:
- #17079 is the protected symlinked carveout side: bwrap now binds the real symlinked writable-root target and remaps carveouts before masking.

Related to #15157:
- #15157 is the broader permission-check side of this path-identity problem. This PR addresses the shared logical-vs-canonical normalization issue, but the reported Darwin prompt behavior should be validated separately before auto-closing it.

This should also fix #14672, #14694, #14715, and #15725:
- #14672, #14694, and #14715 are the same Linux symlinked-writable-root/bwrap family as #15781.
- #15725 is the protected symlinked workspace path variant; the PR preserves the protected logical path in policy space while bwrap applies read-only or unreadable treatment to the resolved target so file-vs-directory bind mismatches do not abort sandbox setup.

## Notes
- Added Linux-only regressions for symlinked writable ancestors and protected symlinked directory targets, including nested symlink escape masking without rebinding the escape target writable.


etraut-openai (approved):
Whew, keeping track of when paths are already canonicalized and when they are not is difficult. We should think about whether there's a way to make this clearer through variable names or even types, but that's beyond the scope of this PR. I'm going to have to place some trust in codex's reviews because I can't say with confidence that there's no bug here. I did spend quite a bit of time reviewing, and I didn't find any issues. This will be a nice fix to get in place!

Issue #15984, opened 2026-03-27
0.117.0 regression: UserPromptSubmit hook output is not rendered in TUI/app-server path


92645417d9e5c763259dbebc306e3e:
### What version of Codex CLI is running?

0.117.0

### What subscription do you have?

Plus

### What issue are you seeing?

  In Codex CLI 0.117.0, `UserPromptSubmit` hook output is not rendered in the TUI when running through the app-server path.

  This is visible in at least two cases:
  1. Cold start in default mode, where the first prompt is blocked by a `UserPromptSubmit` hook.
  2. Resuming an old conversation in default mode and then sending input.

  Expected hook UI such as:

  - `Running UserPromptSubmit hook: checking go-workflow input policy`
  - `UserPromptSubmit hook (stopped)`
  - `warning: go-workflow must start from PlanMode...`

  does not appear at all. Instead, the UI immediately returns to a new input box with no visible hook message.

  On 0.116, the same setup does show the hook output correctly.

  I locally patched 0.117.0 from source and confirmed the regression appears to have two parts:
  - app-server `HookStarted` / `HookCompleted` notifications are dropped during conversion before rendering
  - in the no-primary-thread cold-start path, hook notifications can also be buffered and never shown

  So this does not appear to be only a “first prompt” issue.

### What steps can reproduce the bug?

  1. Configure a `UserPromptSubmit` hook that blocks input and emits visible stop/warning text.
  2. Start Codex CLI 0.117.0 in default mode.
  3. Submit a prompt that triggers the hook, for example `$go-workflow`.
  4. Observe that no hook text is rendered and the UI immediately shows another input prompt.

  Also reproducible when:
  1. Resume an existing conversation in default mode.
  2. Submit input that triggers the same `UserPromptSubmit` hook.
  3. Observe that no hook output is shown.

### What is the expected behavior?

  The TUI should render the hook status and stop output, similar to 0.116, for example:

  - `Running UserPromptSubmit hook: ...`
  - `UserPromptSubmit hook (stopped)`
  - the emitted warning/stop lines

### Additional information

  Regression observed between 0.116 and 0.117.0.

  If useful, I have a local source patch against `rust-v0.117.0` that restores visibility by:
  - explicitly converting app-server hook notification payloads instead of relying on JSON shape conversion
  - rendering hook notifications immediately when no primary thread/session exists yet

[...]

PR #16013, opened 2026-03-27
Fix tui_app_server hook notification rendering and replay

etraut-openai:
Addresses #15984

HookStarted/HookCompleted notifications were being translated through a fragile JSON bridge, so hook status/output never reached the renderer. Early hook notifications could also be dropped during session refresh before replay.

This PR fixes `tui_app_server` by mapping app-server hook notifications into TUI hook events explicitly and preserving buffered hook notifications across refresh, so cold-start and resumed sessions render the same hook UI as the legacy TUI.

### DIFF
--- a/codex-rs/tui_app_server/src/app.rs
+++ b/codex-rs/tui_app_server/src/app.rs
@@ -518,7 +518,12 @@ struct ThreadEventStore {
 
 impl ThreadEventStore {
     fn event_survives_session_refresh(event: &ThreadBufferedEvent) -> bool {
-        matches!(event, ThreadBufferedEvent::Request(_))
+        matches!(
+            event,
+            ThreadBufferedEvent::Request(_)
+                | ThreadBufferedEvent::Notification(ServerNotification::HookStarted(_))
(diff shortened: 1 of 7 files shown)


codex-rs/tui_app_server/src/app.rs
```diff
@@ -8452,6 +8520,31 @@ guardian_approval = true
...
+                ServerNotification::HookStarted(_)
+            ))
+        ));
+        assert!(matches!(
+            snapshot.events.get(1),
+            Some(ThreadBufferedEvent::Notification(
+                ServerNotification::HookCompleted(_)
+            ))
+        ));
+    }
```
bolinfest: Can you do (with `_` replaced with expected value)?

```suggestion
        assert_eq!(
          snapshot.events,
          vec![
            ThreadBufferedEvent::Notification(
                ServerNotification::HookStarted(_)
            ),
            ThreadBufferedEvent::Notification(
                ServerNotification::HookCompleted(_)
            )
          ]
        );
    }
```

bolinfest (approved):

PR #16021, opened 2026-03-27
Fix /copy regression in tui_app_server turn completion

etraut-openai:
Addresses #16019

`tui_app_server` renders completed assistant messages from item notifications, but it only updated `/copy` state from `turn/completed`. After the app-server migration, turn completion no longer repeats the final assistant text, so `/copy` could stay unavailable even after the first normal response.

This PR track the last completed final-answer agent message during an active app-server turn and promote it into the `/copy` cache when the turn completes. This restores the pre-migration behavior without changing rollback handling.


codex-rs/tui_app_server/src/chatwidget.rs
```diff
@@ -2253,9 +2258,10 @@ impl ChatWidget {
...
     fn on_task_complete(&mut self, last_agent_message: Option<String>, from_replay: bool) {
         self.submit_pending_steers_after_interrupt = false;
-        if let Some(message) = last_agent_message.as_ref()
-            && !message.trim().is_empty()
-        {
+        let completed_turn_output = last_agent_message
+            .filter(|message| !message.trim().is_empty())
+            .or_else(|| self.pending_turn_copyable_output.take());
+        if let Some(message) = completed_turn_output.as_ref() {
             self.last_copyable_output = Some(message.clone());
```
bolinfest, line 2265: To confirm: `last_copyable_output` shouldn't be set to `completed_turn_output`?

bolinfest (approved):

codex-rs/tui_app_server/src/chatwidget.rs
etraut-openai: Yes. `completed_turn_output` is the value we assign into `last_copyable_output` in `on_task_complete()` before using it for the completion notification. The fallback is only there to source that value from the completed `AgentMessageItem` when `turn/completed` omits `last_agent_message`.

etraut-openai: Yes, it should. I renamed the local to `copyable_turn_output` to make that clearer.

PR #16041, opened 2026-03-27
Fix app-server TUI MCP startup warnings regression

etraut-openai:
This addresses #16038

The default `tui_app_server` path stopped surfacing MCP startup failures during cold start, even though the legacy TUI still showed warnings like `MCP startup incomplete (...)`. The app-server bridge emitted per-server startup status notifications, but `tui_app_server` ignored them, so failed MCP handshakes could look like a clean startup.

This change teaches `tui_app_server` to consume MCP startup status notifications, preserve the immediate per-server failure warning, and synthesize the same aggregate startup warning the legacy TUI shows once startup settles.


codex-rs/tui/src/chatwidget.rs
```diff
@@ -2827,29 +2911,113 @@ impl ChatWidget {
...
+
     #[cfg(test)]
-    fn on_mcp_startup_complete(&mut self, ev: McpStartupCompleteEvent) {
-        let mut parts = Vec::new();
-        if !ev.failed.is_empty() {
-            let failed_servers: Vec<_> = ev.failed.iter().map(|f| f.server.clone()).collect();
-            parts.push(format!("failed: {}", failed_servers.join(", ")));
-        }
-        if !ev.cancelled.is_empty() {
+    fn on_mcp_startup_update(&mut self, ev: McpStartupUpdateEvent) {
```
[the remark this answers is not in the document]
etraut-openai: This one looks non-actionable. `handle_codex_event` and `handle_codex_event_replay` are both `#[cfg(test)]` in [`codex-rs/tui/src/chatwidget.rs`], and those are the only call sites for `on_mcp_startup_update` / `on_mcp_startup_complete`. In production the app-server path uses `on_mcp_server_status_updated`, so keeping the legacy MCP startup helpers test-only is intentional and `cargo check -p codex-tui` still passes on the branch.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -2780,15 +2793,87 @@ impl ChatWidget {
         self.request_redraw();
     }
 
-    #[cfg(test)]
-    fn on_mcp_startup_update(&mut self, ev: McpStartupUpdateEvent) {
-        let mut status = self.mcp_startup_status.take().unwrap_or_default();
-        if let McpStartupStatus::Failed { error } = &ev.status {
-            self.on_warning(error);
+    fn update_mcp_startup_status(
```
fcoury, line 2805: I found this new method difficult to understand and verify. I think the behavior is correct, but the shape of the code makes it hard to reason about.

It would be beneficial to either split it up or at least add documentation about what it does, maybe even inline comments throughout the function.

I told Codex that I found it hard to follow, and asked it to suggest some improvements. Here's the list it came up with:

- mixes three distinct responsibilities in one function: ingesting an update, deciding whether it belongs to the current round or a buffered next round, and deciding whether startup is now complete. That makes it hard to verify invariants like “when do stale updates get dropped?” and “what exactly reactivates a round?” I would push to split this into smaller helpers with named concepts.
- relies on a cluster of booleans and side effects (mcp_startup_ignore_updates_until_next_start, mcp_startup_allow_terminal_only_next_round, mcp_startup_pending_next_round_saw_starting) rather than an explicit state enum. That makes valid state combinations implicit. In review, I’d recommend a small enum like Active, IgnoringStale, BufferingNextRound { allow_terminal_only, saw_starting } so illegal combinations are unrepresentable.
- encode the completion/reactivation rules inline. The logic is correct-looking, but the business rules are buried in conditionals. I’d ask for small predicate helpers such as pending_round_is_ready_to_activate(...) and startup_round_is_settled(...) with doc comments.

The main recommendation would be to refactor before merge into something like:

- record_startup_update(...)
- maybe_activate_buffered_round(...)
- maybe_finish_startup_round(...)
- render_mcp_startup_header(...)

fcoury (commented):
I tested the changes on this PR for the following conditions:
- MCP timeout
- MCP crash
- MCP failure with a reason

All of them worked as expected.

The changes are approved, but I left a small optional nit regarding the complexity of the `update_mcp_startup_status` method. I found it hard to follow during review.

codex-rs/tui/src/chatwidget.rs
etraut-openai: Thanks, good suggestions. I had Codex add comments to make it clearer.

PR #16044, opened 2026-03-27
[mcp] Bypass read-only tool checks.

mzeng-openai:
- [x] Auto / unspecified approval mode: read-only tools now skip before guardian routing.
- [x] Approve / always-allow mode: read-only tools still skip, now via the shared early return.
- [x] Prompt mode: read-only tools no longer skip; they continue to approval.


dylan-hurd-oai (approved):
I'm slightly confused about the control flow used for maybe_request_mcp_tool_approval, but this specific change LGTM

owenlin0 (approved):

PR #16055, opened 2026-03-27
Force forked agents to inherit parent model settings

friel-openai:
## Summary
- make explicit forked spawns (`fork_context = true` / `fork_turns`) ignore child `model` and `reasoning_effort` overrides
- keep forked agents on the spawning thread's effective model, provider, reasoning, service tier, verbosity, context-window, auto-compact, and profile settings
- preserve forked-child prompt-cache behavior by reusing the parent prompt cache key
- preserve MCP tool-surface parity for forked children by inheriting the parent MCP connection manager instead of starting a fresh manager
- add v1/v2 regression tests for model inheritance, cache-key inheritance, and MCP inheritance

## Test
- `cargo +1.93.1 test -p codex-core spawn_agent_fork_context_ignores_child_model_overrides --lib`
- `cargo +1.93.1 test -p codex-core multi_agent_v2_spawn_fork_turns_ignores_child_model_overrides --lib`
- `cargo +1.93.1 test -p codex-core spawn_agent_can_fork_parent_thread_history_with_sanitized_items --lib`


jif-oai (commented):
This is way too large. Forked agent should just inherit the full config of the parent agent. We shouldn't need to touch any other components

Btw profiles will disappear so feel free to ignore them all

codex-rs/core/src/codex.rs
```diff
@@ -1909,16 +1929,21 @@ impl Session {
...
-            // instance. It will be replaced with one created via
-            // McpConnectionManager::new() once all its constructor args are
-            // available. This also ensures `SessionConfigured` is emitted
-            // before any MCP-related events. It is reasonable to consider
-            // changing this to use Option or OnceCell, though the current
-            // setup is straightforward enough and performs well.
-            mcp_connection_manager: Arc::new(RwLock::new(McpConnectionManager::new_uninitialized(
-                &config.permissions.approval_policy,
-            ))),
+            mcp_connection_manager: session_configuration
```
jif-oai, line 1932: We can't do this. MCP manager get replaced/modified if we have sandbox updates, elicitation, ...
In general, we can't share the manager without re-designing it
If I get it correctly, the only reason we do this is because we want a stability over `built_tools`. First, this shouldn't drift most of the time. Second, we should use a snapshot around it to fix the issue then

codex-rs/core/src/agent/control.rs
```diff
@@ -524,6 +541,12 @@ impl AgentControl {
         let inherited_exec_policy = self
             .inherited_exec_policy_for_source(&state, Some(&session_source), &config)
             .await;
+        let inherited_prompt_cache_key = self
```
jif-oai, line 544: This resume path should not opportunistically inherit parent cache/MCP state from a live parent. `SubAgentSource::ThreadSpawn` does not persist whether this child was originally forked, so the same resumed agent will get parent-shared cache/MCP only when the parent is currently loaded, and isolated state otherwise... we can't have such non-determinism

codex-rs/core/src/codex_delegate.rs
```diff
@@ -94,6 +94,10 @@ pub(crate) async fn run_codex_thread_interactive(
         inherited_shell_snapshot: None,
         user_shell_override: None,
         inherited_exec_policy: Some(Arc::clone(&parent_session.services.exec_policy)),
+        inherited_prompt_cache_key: Some(parent_session.prompt_cache_key()),
```
jif-oai, line 97: Those are not fork

codex-rs/core/src/thread_manager.rs
```diff
@@ -770,6 +773,8 @@ impl ThreadManagerState {
         metrics_service_name: Option<String>,
         inherited_shell_snapshot: Option<Arc<ShellSnapshot>>,
```
jif-oai, line 774: this is a lot of optional inherited things. We should have a small builder pattern for the inherited things

codex-rs/core/src/tools/handlers/multi_agents_common.rs
```diff
@@ -225,7 +225,11 @@ fn build_agent_shared_config(turn: &TurnContext) -> Result<Config, FunctionCallE
     let mut config = (*base_config).clone();
```
jif-oai, line 225: All of this should be cleared IMO
We should just re-use the full same config... otherwise this will become brittle over time

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -9212,13 +9212,18 @@ mod tests {
             model: "gpt-5".to_string(),
             model_provider_id: "openai".to_string(),
             service_tier: Some(codex_protocol::config_types::ServiceTier::Flex),
+            plan_mode_reasoning_effort: None,
```
jif-oai, line 9215: Same as last comment on all those fields. We should drop all of those things. They are unrelated

codex-rs/core/src/tools/handlers/multi_agents_common.rs
```diff
@@ -225,7 +225,11 @@ fn build_agent_shared_config(turn: &TurnContext) -> Result<Config, FunctionCallE
...
     config.model = Some(turn.model_info.slug.clone());
     config.model_provider = turn.provider.clone();
-    config.model_reasoning_effort = turn.reasoning_effort;
+    // Forked children must preserve the spawning turn's effective model settings, including a
```
jif-oai, line 228: this code path is not only for forked agents

codex-rs/core/src/codex.rs
```diff
@@ -426,6 +426,8 @@ pub(crate) struct CodexSpawnArgs {
     pub(crate) metrics_service_name: Option<String>,
     pub(crate) inherited_shell_snapshot: Option<Arc<ShellSnapshot>>,
     pub(crate) inherited_exec_policy: Option<Arc<ExecPolicyManager>>,
+    pub(crate) inherited_prompt_cache_key: Option<ThreadId>,
+    pub(crate) inherited_mcp_connection_manager: Option<Arc<RwLock<McpConnectionManager>>>,
```
jif-oai, line 430: Up Option of an Arc of RwLock looks like a code smell in Rust... there are better alternatives for hot swap shared references

codex-rs/core/src/tools/handlers/multi_agents_tests.rs
```diff
@@ -366,6 +425,154 @@ async fn spawn_agent_uses_explorer_role_and_preserves_approval_policy() {
     assert_eq!(snapshot.model_provider_id, "ollama");
 }
 
+#[tokio::test]
+async fn spawn_agent_fork_context_ignores_child_model_overrides() {
```
jif-oai, line 429: this test is kind of pointless IMO
A better test would be an integration test that makes sure the full context is the same at the API level

codex-rs/core/src/tools/handlers/multi_agents_tests.rs
```diff
@@ -90,6 +94,61 @@ fn thread_manager() -> ThreadManager {
...
+[profiles.role-profile]
+model_provider = "role-provider"
+"#,
+    )
+    .await
+    .expect("role config should be written");
+
+    let mut config = (*turn.config).clone();
+    let mut role_provider =
+        built_in_model_providers(/* openai_base_url */ /*openai_base_url*/ None)["openai"].clone();
```
jif-oai, line 129: dup in the comment

codex-rs/core/src/tools/handlers/multi_agents/spawn.rs
```diff
@@ -61,17 +61,23 @@ impl ToolHandler for Handler {
...
             build_agent_spawn_config(&session.get_base_instructions().await, turn.as_ref())?;
-        apply_requested_spawn_agent_model_overrides(
-            &session,
-            turn.as_ref(),
-            &mut config,
-            args.model.as_deref(),
-            args.reasoning_effort,
-        )
-        .await?;
+        if !args.fork_context {
```
jif-oai, line 64: we should probably error to the model if it specify fork with a model and reasoning effort. Or at least emit a warning

jif-oai (changes requested):
I agree this would be good to share more cache here but I don't think this PR fits at the right level.
I would suggest 3 different small PRs:
1. Fully re-use the parent config in case of a fork + return a warning or error to the agent if it tries to set anything else
2. Shared cache key, do a proper wiring that let us "override" the cache key for a given thread. That we fix at the creation of the thread and gets set directly into the API client
3. Make the MCP tools listing stable. Either by definition or using a snapshot

PR #16122, opened 2026-03-28
In code mode, rename tools global to functions

andmis:


codex-rs/core/tests/suite/code_mode.rs
```diff
@@ -50,7 +50,7 @@ fn custom_tool_output_items(req: &ResponsesRequest, call_id: &str) -> Vec<Value>
 }
 
 fn tool_names(body: &Value) -> Vec<String> {
-    body.get("tools")
```
pakrym-oai, line 53: this refers to response api field and should stay tools

pakrym-oai (approved):

PR #16184, opened 2026-03-29
Route TUI `/feedback` submission through the app server

etraut-openai:
The TUI’s `/feedback` flow was still uploading directly through the local feedback crate, which bypassed app-server behavior such as auth-derived feedback tags like chatgpt_user_id and made TUI feedback handling diverge from other clients. It also meant that remove TUI sessions failed to upload the correct feedback logs and session details.

Testing: Manually tested `/feedback` flow and confirmed that it didn't regress.


codex-rs/tui/src/bottom_pane/feedback_view.rs
```diff
@@ -678,7 +649,7 @@ mod tests {
     fn feedback_view_with_connectivity_diagnostics() {
         let (tx_raw, _rx) = tokio::sync::mpsc::unbounded_channel::<AppEvent>();
         let tx = AppEventSender::new(tx_raw);
-        let diagnostics = FeedbackDiagnostics::new(vec![
+        let _diagnostics = FeedbackDiagnostics::new(vec![
```
etraut-openai: Can we remove this variable if it's not used?

PR #16191, opened 2026-03-29
feat: refresh non-curated cache from plugin list.

xl-openai:
1. Use versions for non-curated plugin (defined in plugin.json) for cache refresh
2. Trigger refresh from plugin/list roots

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -479,19 +479,6 @@ impl CodexMessageProcessor {
         self.thread_manager.skills_manager().clear_cache();
     }
 
-    pub(crate) async fn maybe_start_plugin_startup_tasks_for_latest_config(&self) {
-        match self.load_latest_config(/*fallback_cwd*/ None).await {
-            Ok(config) => self
-                .thread_manager
-                .plugins_manager()
-                .maybe_start_plugin_startup_tasks_for_config(
-                    &config,
-                    self.thread_manager.auth_manager(),
-                ),
-            Err(err) => warn!("failed to load latest config for plugin startup tasks: {err:?}"),
-        }
-    }
-
     fn current_account_updated_notification(&self) -> AccountUpdatedNotification {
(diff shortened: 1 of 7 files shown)


codex-rs/app-server/src/message_processor.rs
```diff
@@ -851,9 +851,6 @@ impl MessageProcessor {
         match self.config_api.write_value(params).await {
             Ok(response) => {
                 self.codex_message_processor.clear_plugin_related_caches();
-                self.codex_message_processor
```
xl-openai, line 871: The original intent was to run plugin_startup_tasks when config changes enabled plugin features. Since plugin features are now default-on, this path is no longer very useful.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -5755,6 +5742,7 @@ impl CodexMessageProcessor {
             force_remote_sync,
         } = params;
         let roots = cwds.unwrap_or_default();
+        plugins_manager.maybe_start_non_curated_plugin_cache_refresh_for_roots(&roots);
```
xl-openai, line 5972: start_non_curated_plugin_cache_refresh runs from plugin/list instead of startup because it needs the
  roots (cwds) to locate marketplace.json. We do not persist that information in config today.

codex-rs/core/src/plugins/manager.rs
```diff
@@ -1076,6 +1141,55 @@ impl PluginsManager {
...
+                state.requested_roots.clone()
+            };
+
+            let Some(roots) = roots else {
+                let mut state = match self.non_curated_cache_refresh_state.write() {
+                    Ok(state) => state,
+                    Err(err) => err.into_inner(),
+                };
+                state.in_flight = false;
+                return;
```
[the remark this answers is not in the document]
xl-openai: In the normal worker path, requested_roots is only cleared when the worker is already about to set in_flight = false and exit.

alexsong-oai (approved):

sayan-oai (approved):

PR #16202, opened 2026-03-29
Fix paste-driven bottom pane completion teardown

etraut-openai:
Fix paste-driven bottom-pane completion teardown (#16192)

`BottomPane::handle_paste()` could leave a completed modal flow mounted while re-enabling the composer, putting the TUI in an inconsistent state where stale views could still affect rendering and input routing. Align the paste path with the existing key-driven completion logic by tearing down the active modal flow before restoring composer input, and add a regression test covering the stacked-view case that exposed the bug.

Big thanks to @iqdoctor for identifying the root cause for this issue.

### DIFF
--- a/codex-rs/tui/src/bottom_pane/mod.rs
+++ b/codex-rs/tui/src/bottom_pane/mod.rs
@@ -477,9 +477,14 @@ impl BottomPane {
     }
 
     pub fn handle_paste(&mut self, pasted: String) {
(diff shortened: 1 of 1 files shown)


fcoury-oai (approved):
Could not replicate this issue - did you find a way to replicate it in manual testing?

However the test case seems enough to replicate the possible issue. Approved.

PR #16273, opened 2026-03-30
Fix regression: "not available in TUI" error message

etraut-openai:
Addresses a recent TUI regression

Problem: Pressing Ctrl+C during early TUI startup could route an interrupt with no active turn into the generic unsupported-op fallback, showing “Not available in app-server TUI yet for thread …” repeatedly.

Solution: Treat interrupt requests as handled when no active turn exists yet, preventing fallback error spam during startup, and add a regression test covering interrupt-without-active-turn behavior.


fcoury-oai (approved):
Tested before and after and was able to replicate the behavior and confirm the fix.

Looks good.

PR #16284, opened 2026-03-30
Fix TUI app-server permission profile conversions

etraut-openai:
Addresses #16283

Problem: TUI app-server permission approvals could drop filesystem grants because request and response payloads were round-tripped through mismatched camelCase and snake_case JSON shapes.
Solution: Replace the lossy JSON round-trips with typed app-server/core permission conversions so requested and granted permission profiles, including filesystem paths and scope, are preserved end to end.


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1081,6 +1081,15 @@ impl From<CoreNetworkApprovalContext> for NetworkApprovalContext {
     }
 }
 
+impl From<NetworkApprovalContext> for CoreNetworkApprovalContext {
```
etraut-openai: I don't want this PR to modify the app server protocol. Move this to the TUI directory.

codex-rs/tui/src/chatwidget/tests/approval_requests.rs
```diff
@@ -0,0 +1,308 @@
+use super::*;
```
etraut-openai, line 4: Add a block comment at the top of this new module.

fcoury-oai (approved):
I was able to replicate it manually. Approved.

PR #16344, opened 2026-03-31
feat: move exec-server ownership

jif-oai:
This introduces session-scoped ownership for exec-server so ws disconnects no longer immediately kill running remote exec processes, and it prepares the protocol for reconnect-based resume.
  - add session_id / resume_session_id to the exec-server initialize handshake
  - move process ownership under a shared session registry
  - detach sessions on websocket disconnect and expire them after a TTL instead of killing processes immediately (we will resume based on this)
  - allow a new connection to resume an existing session and take over notifications/ownership
  - I use UUID to make them not predictable as we don't have auth for now
  - make detached-session expiry authoritative at resume time so teardown wins at the TTL boundary
  - reject long-poll process/read calls that get resumed out from under an older attachment


codex-rs/exec-server/src/server/handler.rs
```diff
@@ -16,123 +21,217 @@ use codex_app_server_protocol::JSONRPCErrorError;
...
     }
 
     pub(crate) async fn shutdown(&self) {
-        self.process.shutdown().await;
+        self.session.detach().await;
     }
 
-    pub(crate) fn initialize(&self) -> Result<InitializeResponse, JSONRPCErrorError> {
-        self.process.initialize()
+    pub(crate) fn is_current_attachment(&self) -> bool {
```
starr-openai: nit on naming - `is_current_session` or `is_session_attached`?

codex-rs/exec-server/src/server/handler.rs
```diff
@@ -16,123 +21,217 @@ use codex_app_server_protocol::JSONRPCErrorError;
...
+                .as_millis() as u64;
+            let response = self
+                .session
+                .process()
+                .exec_read(ReadParams {
+                    wait_ms: Some(wait_ms),
+                    ..params.clone()
+                })
+                .await?;
+            self.require_current_attachment()?;
```
starr-openai: should this be before we do the exec_read() call?

codex-rs/exec-server/src/server/handler.rs
```diff
@@ -16,123 +21,217 @@ use codex_app_server_protocol::JSONRPCErrorError;
...
-        self.process.exec_read(params).await
+        self.require_initialized_for("exec")?;
+
+        let total_wait = Duration::from_millis(params.wait_ms.unwrap_or(0));
+        if total_wait.is_zero() {
+            return self.session.process().exec_read(params).await;
+        }
+
+        let deadline = Instant::now() + total_wait;
+        loop {
```
starr-openai: why do we loop here? the underlying process().exec_read() call should already be respecting the wait_ms, so it would only return if timeout/exit or bytes read no?

codex-rs/exec-server/src/server/handler.rs
```diff
@@ -16,123 +21,217 @@ use codex_app_server_protocol::JSONRPCErrorError;
...
+        if self.is_current_attachment() {
+            Ok(())
+        } else {
+            Err(invalid_request(
+                "session has been resumed by another connection".to_string(),
+            ))
+        }
+    }
+
+    fn require_current_attachment_string(&self) -> Result<(), String> {
```
starr-openai: nit - any way to share these two? I see `require_current_attachment_string` is just used for the string of the initialized() call - wonder if we can just return an error and then convert to str within that method only?

codex-rs/exec-server/src/server/processor.rs
```diff
@@ -40,6 +46,12 @@ pub(crate) async fn run_connection(connection: JsonRpcConnection) {
 
     // Process inbound events sequentially to preserve initialize/initialized ordering.
     while let Some(event) = incoming_rx.recv().await {
+        if let Some(current_handler) = handler.as_ref()
+            && !current_handler.is_current_attachment()
+        {
+            debug!("exec-server connection evicted after session resume");
+            break;
```
starr-openai, line 64: should we return with an error, or send an event? a second incoming connection while an existing one is active seems like an error, whereas a simple reconnect is a more normal operation

codex-rs/exec-server/src/server/processor.rs
```diff
@@ -56,8 +68,88 @@ pub(crate) async fn run_connection(connection: JsonRpcConnection) {
             }
             JsonRpcConnectionEvent::Message(message) => match message {
                 codex_app_server_protocol::JSONRPCMessage::Request(request) => {
-                    if let Some(route) = router.request_route(request.method.as_str()) {
-                        let message = route(handler.clone(), request).await;
+                    if request.method == INITIALIZE_METHOD {
```
starr-openai: this logic feels like it should be within the handler itself - we're already guarding that init is called there.  my guess is you do this so the SessionHandle is always bootstrapped at handler construction time?  imo it might still be better to make the SessionHandle init in an empty state, and we move its creation into the handler init() call.

just feels weird to peak into the rpc logic here

codex-rs/exec-server/src/server/session_registry.rs
```diff
@@ -0,0 +1,194 @@
...
+use crate::server::process_handler::ProcessHandler;
+
+#[cfg(test)]
+const DETACHED_SESSION_TTL: Duration = Duration::from_millis(200);
+#[cfg(not(test))]
+const DETACHED_SESSION_TTL: Duration = Duration::from_secs(10);
+
+#[derive(Clone, Default)]
+pub(crate) struct SessionRegistry {
+    inner: Arc<SessionRegistryInner>,
```
starr-openai: maybe a rust paradigm i'm not used to seeing - why the redirected `inner` class? could we not just have an Arc<SessionRegistry> instead?

codex-rs/exec-server/src/server/session_registry.rs
```diff
@@ -0,0 +1,194 @@
...
+    attachment_id: u64,
+}
+
+impl SessionRegistry {
+    pub(crate) async fn attach(
+        &self,
+        resume_session_id: Option<String>,
+        notifications: RpcNotificationSender,
+    ) -> Result<SessionHandle, JSONRPCErrorError> {
+        let attachment_id = self.inner.next_attachment_id.fetch_add(1, Ordering::SeqCst) + 1;
```
starr-openai: I wonder if it is simpler to just have a UUID per connection as a connection_id, and then use that.

might make debugging more straightforward since we could log the connection id on connect, and then can always correlate a connection => session_id, connection_id in the logs

codex-rs/exec-server/src/server/session_registry.rs
```diff
@@ -0,0 +1,194 @@
...
+        let mut expired_entry = None;
+        let mut expired_error = None;
+        let entry = {
+            let mut sessions = self.inner.sessions.lock().await;
+            let entry = if let Some(session_id) = resume_session_id {
+                let entry = sessions
+                    .get(&session_id)
+                    .cloned()
+                    .ok_or_else(|| invalid_request(format!("unknown session id {session_id}")))?;
+                let expired = entry
```
starr-openai: make an `entry.is_expired()` methdo?

codex-rs/exec-server/src/server/session_registry.rs
```diff
@@ -0,0 +1,194 @@
...
+                    expired_error =
+                        Some(invalid_request(format!("unknown session id {session_id}")));
+                    expired_entry = sessions.remove(&session_id);
+                    None
+                } else {
+                    Some(entry)
+                }
+            } else {
+                let session_id = Uuid::new_v4().to_string();
+                let entry = Arc::new(SessionEntry {
```
starr-openai: SessionEntry::new method perhaps?

codex-rs/exec-server/src/server/session_registry.rs
```diff
@@ -0,0 +1,194 @@
...
+        self.entry
+            .process
+            .set_notification_sender(/*notifications*/ None);
+
+        let registry = self.registry.clone();
+        let session_id = self.entry.session_id.clone();
+        let attachment_id = self.attachment_id;
+        tokio::spawn(async move {
+            registry.expire_if_detached(session_id, attachment_id).await;
+        });
```
starr-openai, line 257: note to self: need to do a deeper re-read of this fn above, but i suspect there's a way to simplify..

codex-rs/exec-server/src/server/transport.rs
```diff
@@ -58,18 +59,23 @@ async fn run_websocket_listener(
...
-                    run_connection(JsonRpcConnection::from_websocket(
-                        websocket,
-                        format!("exec-server websocket {peer_addr}"),
-                    ))
+                    run_connection(
+                        JsonRpcConnection::from_websocket(
+                            websocket,
+                            format!("exec-server websocket {peer_addr}"),
+                        ),
+                        session_registry,
```
starr-openai: yeah again, having session_registry here seems a bit off to me

codex-rs/exec-server/src/server/processor.rs
```diff
@@ -57,7 +80,13 @@ pub(crate) async fn run_connection(connection: JsonRpcConnection) {
             JsonRpcConnectionEvent::Message(message) => match message {
                 codex_app_server_protocol::JSONRPCMessage::Request(request) => {
                     if let Some(route) = router.request_route(request.method.as_str()) {
-                        let message = route(handler.clone(), request).await;
+                        let message = tokio::select! {
+                            message = route(Arc::clone(&handler), request) => message,
+                            _ = disconnected_rx.changed() => {
+                                debug!("exec-server transport disconnected while handling request");
+                                break;
+                            }
```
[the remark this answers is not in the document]
jif-oai: This will be fixed in a follow-up

starr-openai (approved):

PR #16356, opened 2026-03-31
Refactor external auth to use a single trait

etraut-openai:
## Summary
- Replace the separate external auth enum and refresher trait with a single `ExternalAuth` trait in login auth flow
- Move bearer token auth behind `BearerTokenRefresher` and update `AuthManager` and app-server wiring to use the generic external auth API


codex-rs/login/src/auth/auth_tests.rs
```diff
@@ -433,7 +518,7 @@ exit 1
         serde_json::from_value(json!({
             "command": self.command,
             "args": self.args,
-            "timeout_ms": 1000,
+            "timeout_ms": 5000,
```
etraut-openai: Why did you change this timeout? If it's not required, leave it as is.

etraut-openai: Reverted that timeout change in `8a26b7b42e`, so the test is back to the original value. The only thing I kept was the `AuthManager` `Debug` impl, because removing it breaks downstream `#[derive(Debug)]` sites that hold `Arc<AuthManager>`.

codex-rs/login/src/auth/external_bearer.rs
```diff
@@ -48,9 +52,31 @@ impl ExternalBearerAuth {
...
 
-impl fmt::Debug for ExternalBearerAuth {
+#[async_trait]
+impl ExternalAuth for BearerTokenRefresher {
+    fn auth_mode(&self) -> crate::AuthMode {
+        crate::AuthMode::ApiKey
+    }
+
+    async fn resolve(&self) -> io::Result<Option<ExternalAuthTokens>> {
+        let access_token = self.resolve_access_token().await?;
```
etraut-openai: Can't we just inline resolve_access_token?

codex-rs/login/src/auth/external_bearer.rs
```diff
@@ -48,9 +52,31 @@ impl ExternalBearerAuth {
...
+        Ok(Some(ExternalAuthTokens::access_token_only(access_token)))
+    }
+
+    async fn refresh(
+        &self,
+        _context: ExternalAuthRefreshContext,
+    ) -> io::Result<ExternalAuthTokens> {
+        self.refresh_after_unauthorized().await?;
+        let access_token = self.resolve_access_token().await?;
```
etraut-openai: Same here. Can't we just inline resolve_access_token?

codex-rs/login/src/auth/auth_tests.rs
```diff
@@ -321,6 +324,94 @@ async fn unauthorized_recovery_uses_external_refresh_for_bearer_manager() {
     assert_eq!(refreshed_token.as_deref(), Some("refreshed-provider-token"));
 }
 
+#[tokio::test]
+async fn external_chatgpt_auth_manager_refresh_persists_ephemeral_auth() {
```
etraut-openai: I don't think this new external_chatgpt_auth_manager_refresh_persists_ephemeral_auth test is adding much value. Remove it.

codex-rs/login/src/auth/auth_tests.rs
```diff
@@ -447,14 +538,42 @@ struct AuthFileParams {
     chatgpt_account_id: Option<String>,
 }
 
+fn fake_jwt(
```
etraut-openai: If we remove external_chatgpt_auth_manager_refresh_persists_ephemeral_auth, I think we can re-inline this function. Try to minimize code churn in this file.

codex-rs/login/src/auth/external_bearer.rs
etraut-openai: Inlined that logic in `274a34d770` and dropped the thin helper.

codex-rs/login/src/auth/external_bearer.rs
etraut-openai: Done in `274a34d770`. `refresh()` now updates the cache directly instead of bouncing through the helper.

codex-rs/login/src/auth/auth_tests.rs
etraut-openai: Removed that test in `274a34d770`.

codex-rs/login/src/auth/auth_tests.rs
etraut-openai: Yep. With that test gone, I re-inlined the helper in `274a34d770` and trimmed the extra scaffolding.

codex-rs/login/src/auth/manager.rs
```diff
@@ -1200,8 +1195,10 @@ impl AuthManager {
             auth_credentials_store_mode: AuthCredentialsStoreMode::File,
             forced_chatgpt_workspace_id: RwLock::new(None),
             refresh_lock: AsyncMutex::new(()),
-            external_auth: RwLock::new(Some(ExternalAuth::Bearer(ExternalBearerAuth::new(config)))),
-        })
+            external_auth: RwLock::new(None),
+        });
+        manager.set_external_auth(Arc::new(BearerTokenRefresher::new(config)));
```
bolinfest: Why isn't this done as part of `Self {}`?

codex-rs/login/src/auth/manager.rs
```diff
@@ -143,7 +143,9 @@ pub struct ExternalAuthRefreshContext {
 }
 
 #[async_trait]
-pub trait ExternalAuthRefresher: Send + Sync {
+pub trait ExternalAuth: Send + Sync {
```
bolinfest, line 150: Somewhat outside the scope of this PR, but I think traits are more deserving of docstrings than other types, so if you don't mind adding them while you're here, I think it would help clarify the role of `ExternalAuth`.

bolinfest (approved):
Thanks: this feels better!

codex-rs/login/src/auth/manager.rs
etraut-openai: Updated in `9659a88b59` so `external_bearer_only()` installs the external auth directly in `Self { ... }`.

codex-rs/login/src/auth/manager.rs
etraut-openai: Added trait docs in `9659a88b59` to spell out how `ExternalAuth` is meant to be used.

codex-rs/login/src/auth/manager.rs
```diff
@@ -143,11 +143,21 @@ pub struct ExternalAuthRefreshContext {
 }
 
 #[async_trait]
-pub trait ExternalAuthRefresher: Send + Sync {
+/// Pluggable auth provider used by `AuthManager` for externally managed auth flows.
```
bolinfest, line 146: Thanks @etraut-openai, this is particularly helpful!

etraut-openai: Appreciate it. I wanted the trait docs to make the abstraction swap easier to follow for the next person reading this code.

PR #16361, opened 2026-03-31
Refactor chatwidget tests into topical modules

etraut-openai:
Problem: `chatwidget/tests.rs` had grown into a single oversized test blob that was hard to maintain and exceeded the repo's blob size limit.

Solution: split the chatwidget tests into topical modules with a thin root `tests.rs`, shared helper utilities, preserved snapshot naming, and hermetic test config so the refactor stays stable and passes the `codex-tui` test suite.


codex-rs/tui/src/chatwidget/tests/app_server.rs
```diff
@@ -0,0 +1,588 @@
+use super::*;
+#[allow(unused_imports)]
```
bolinfest: Did you add the `allow(unused_imports)` automatically for all files? Because `assert_eq` is used here and elsewhere...

codex-rs/tui/src/chatwidget/tests/exec_flow.rs
```diff
@@ -0,0 +1,1803 @@
+use super::*;
+#[allow(unused_imports)]
+use pretty_assertions::assert_eq;
+
+// (removed experimental resize snapshot test)
```
bolinfest: remove comment?

bolinfest (approved):

PR #16387, opened 2026-04-01
feat(config) Add ConfigEdit permission profile merge

dylan-hurd-oai:
## Summary
Adds support for ConfigEdit::MergePermissionProfile to unblock support for an "Always Allow" flow in request permission flows.

## Testing
- [x] Adds unit tests

### DIFF
--- a/codex-rs/core/src/config/edit.rs
+++ b/codex-rs/core/src/config/edit.rs
@@ -7,7 +7,10 @@ use codex_features::FEATURES;
 use codex_protocol::config_types::Personality;
 use codex_protocol::config_types::ServiceTier;
 use codex_protocol::config_types::TrustLevel;
+use codex_protocol::models::FileSystemPermissions;
 use codex_protocol::openai_models::ReasoningEffort;
+use codex_protocol::permissions::FileSystemAccessMode;
+use codex_protocol::protocol::PersistPermissionProfileAction;
 use std::collections::BTreeMap;
 use std::collections::HashMap;
 use std::path::Path;
@@ -59,6 +62,8 @@ pub enum ConfigEdit {
         segments: Vec<String>,
         value: TomlItem,
     },
+    /// Merge permissions into a named profile.
+    MergePermissionProfile(PersistPermissionProfileAction),
     /// Remove the value stored at the exact dotted path.
     ClearPath { segments: Vec<String> },
 }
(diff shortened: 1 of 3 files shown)


aibrahim-oai (commented):
LGTM

codex-rs/core/src/config/edit.rs
```diff
@@ -461,6 +467,173 @@ impl ConfigDocument {
         self.remove(segments)
     }
 
+    fn merge_permission_profile(&mut self, action: &PersistPermissionProfileAction) -> bool {
+        let mut mutated = false;
+
+        for (path, access) in filesystem_path_access(action.permissions.file_system.as_ref()) {
+            if access == FileSystemAccessMode::Read
+                && self.has_writable_filesystem_ancestor(&action.profile_name, &path)
```
jif-oai, line 475: This skip is wrong when a more-specific \none` entry exists under the writable parent; in that case "Always Allow" becomes a no-op because we never persist the narrower read grant

codex-rs/core/src/config/edit.rs
```diff
@@ -461,6 +467,173 @@ impl ConfigDocument {
...
+                continue;
+            }
+
+            let segments = vec![
+                "permissions".to_string(),
+                action.profile_name.clone(),
+                "filesystem".to_string(),
+                path.clone(),
+            ];
+            let merged_access = match self.get_item(&segments).and_then(TomlItem::as_str) {
```
jif-oai, line 486: I only see a checks for a flat filesystem keys. What if the existing rule is stored in scoped form? we'll add a duplicate flat entry here, but the scoped still wins by precedence..

codex-rs/core/src/config/edit.rs
```diff
@@ -461,6 +467,173 @@ impl ConfigDocument {
...
+            "filesystem".to_string(),
+        ];
+        let Some(filesystem) = self.descend(&filesystem_segments, TraversalMode::Existing) else {
+            return false;
+        };
+
+        let path = Path::new(path);
+        for (key, item) in filesystem.iter() {
+            if matches!(item.as_str(), Some("write")) {
+                let candidate = Path::new(key);
```
jif-oai, line 535: We should have a path normalization here so make sure we don't have issues on windows

jif-oai (approved):

PR #16448, opened 2026-04-01
fix(guardian): make GuardianAssessmentEvent.action strongly typed

owenlin0:
## Description

Previously the `action` field on `EventMsg::GuardianAssessment`, which describes what Guardian is reviewing, was typed as an arbitrary JSON blob. This PR cleans it up and defines a sum type representing all the various actions that Guardian can review.

This is a breaking change (on purpose), which is fine because:
- the Codex app / VSCE does not actually use `action` at the moment
- the TUI code that consumes `action` is updated in this PR as well
- rollout files that serialized old `EventMsg::GuardianAssessment` will just silently drop these guardian events
- the contract is defined as unstable, so other clients have a fair warning :)

This will make things much easier for followup Guardian work.

## Why

The old guardian review payloads worked, but they pushed too much shape knowledge into downstream consumers. The TUI had custom JSON parsing logic for commands, patches, network requests, and MCP calls, and the app-server protocol was effectively just passing through an opaque blob.

Typing this at the protocol boundary makes the contract clearer.


codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -2904,10 +2905,11 @@ mod tests {
...
     fn guardian_assessment_started_uses_event_turn_id_fallback() {
         let conversation_id = ThreadId::new();
-        let action = json!({
-            "tool": "shell",
-            "command": "rm -rf /tmp/example.sqlite",
-        });
+        let action = codex_protocol::protocol::GuardianAssessmentAction::Command {
+            source: codex_protocol::protocol::GuardianCommandSource::Shell,
+            command: "rm -rf /tmp/example.sqlite".to_string(),
+            cwd: "/tmp".into(),
```
owenlin0, line 2911: FYI a number of the arbitrary JSON payloads in tests were actually malformed due to lack of typing (e.g. missing `cwd`)

codex-rs/protocol/src/approvals.rs
```diff
@@ -98,6 +98,85 @@ pub enum GuardianAssessmentStatus {
...
+    pub tool_name: String,
+    pub connector_id: Option<String>,
+    pub connector_name: Option<String>,
+    pub tool_title: Option<String>,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, JsonSchema, TS)]
+#[serde(tag = "type", rename_all = "snake_case")]
+#[ts(tag = "type", rename_all = "snake_case")]
+pub enum GuardianAssessmentAction {
```
dylan-hurd-oai, line 111: I think we might be exploding the types a bit here? Do we need all of GuardianAssessmentAction::Command, GuardianCommandAssessmentAction, GuardianApprovalRequest::Shell, and ExecApprovalRequestEvent?

I'm wondering whether there's a structure/interface here that we can use across Guardian and normal escalation/approval flows. Or at least, have fewer Guardian-specific types.

codex-rs/tui/src/chatwidget/tests/mcp_startup.rs
dylan-hurd-oai, line 1: nit: why is this file called mcp_guardian.rs if there are non-mcp calls in here?

dylan-hurd-oai (commented):
Would love to think about deduplicating these types - if we're planning to do so in a follow-up, that also feels reasonable!

codex-rs/protocol/src/approvals.rs
owenlin0: good catch, while iterating on the types I accidentally left dead types around such as `GuardianCommandAssessmentAction` 

cleaned them up! all the ones we need are `GuardianAssessmentAction::*`

codex-rs/tui/src/chatwidget/tests/mcp_startup.rs
owenlin0: good flag. since i'm here already, made the split more logical:
- Rename the guardian-focused tests into `guardian.rs.`
- Move the MCP startup block into `mcp_startup.rs.`

PR #16462, opened 2026-04-01
fix(guardian): fix ordering of guardian events

owenlin0:
Guardian events were emitted a bit out of order for CommandExecution items. This would make it hard for the frontend to render a guardian auto-review, which has this payload:
```
pub struct ItemGuardianApprovalReviewStartedNotification {
    pub thread_id: String,
    pub turn_id: String,
    pub target_item_id: String,
    pub review: GuardianApprovalReview,
    // FYI this is no longer a json blob
    pub action: Option<JsonValue>,
}
```

There is a `target_item_id` the auto-approval review is referring to, but the actual item had not been emitted yet.

Before this PR:
- `item/autoApprovalReview/started`
- `item/autoApprovalReview/completed`, and if approved...
- `item/started`
- `item/completed`

After this PR:
- `item/started`
- `item/autoApprovalReview/started`
- `item/autoApprovalReview/completed`
- `item/completed`

This lines up much better with existing patterns (i.e. human review in `Default mode`, where app-server would send a server request to prompt for user approval after `item/started`), and makes it easier for clients to render what guardian is actually reviewing.

We do this following a similar pattern as `FileChange` (aka apply patch) items, where we create a FileChange item and emit `item/started` if we see the apply patch approval request, before the actual apply patch call runs.


codex-rs/app-server-protocol/src/protocol/item_builders.rs
```diff
@@ -0,0 +1,303 @@
...
+    Begin(&'a PatchApplyBeginEvent),
+    End(&'a PatchApplyEndEvent),
+}
+
+/// Core event payloads that can be converted into a [`ThreadItem::CommandExecution`].
+pub enum CommandExecutionEvent<'a> {
+    ApprovalRequest(&'a ExecApprovalRequestEvent),
+    Begin(&'a ExecCommandBeginEvent),
+    End(&'a ExecCommandEndEvent),
+}
```
dylan-hurd-oai, line 86: i don't think lifetimes here do much for us?

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -1991,9 +1910,52 @@ async fn complete_file_change_item(
         .await;
 }
 
+#[allow(clippy::too_many_arguments)]
+async fn start_command_execution_item(
+    conversation_id: &ThreadId,
+    turn_id: String,
+    item_id: String,
+    command: String,
```
dylan-hurd-oai, line 1917: `too_many_arguments` with three ids and three Strings in a row feels fairly sketchy, but a struct to construct another struct feels a bit overkill....

dylan-hurd-oai (approved):
2 non-blocking comments here, lgtm

codex-rs/app-server/src/bespoke_event_handling.rs
owenlin0: yeah, neither seems ideal...

PR #16492, opened 2026-04-02
Auto-trust cwd on thread start

aibrahim-oai:
- Persist trusted cwd state during thread/start when the resolved sandbox is elevated.
- Add app-server coverage for trusted root resolution and confirm turn/start does not mutate trust.

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -235,6 +235,7 @@ use codex_features::Feature;
 use codex_features::Stage;
 use codex_feedback::CodexFeedback;
 use codex_git_utils::git_diff_to_remote;
+use codex_git_utils::resolve_root_git_project_for_trust;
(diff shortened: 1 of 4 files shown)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2211,6 +2214,56 @@ impl CodexMessageProcessor {
...
+            ) {
+                let error = JSONRPCErrorError {
+                    code: INTERNAL_ERROR_CODE,
+                    message: format!("failed to persist trusted project state: {err}"),
+                    data: None,
+                };
+                listener_task_context
+                    .outgoing
+                    .send_error(request_id, error)
+                    .await;
```
owenlin0, line 2241: nit: you can do `self.send_internal_error(request_id, format!("failed to persist trusted project state: {err}")).await;`

owenlin0 (approved):

Issue #2558, opened 2025-08-21
Codex client output truncated when scrolling in Zellij


ForSeason:
### What version of Codex is running?

codex-cli 0.23.0

### Which model were you using?

gpt5

### What platform is your computer?

Darwin 24.6.0 arm64 arm

### What steps can reproduce the bug?

When running the Codex client inside a Zellij session, the output history does not display correctly. Attempting to scroll within the Zellij pane shows truncated content instead of the full history. This makes it difficult to review previously generated output.

<img width="1072" height="905" alt="Image" src="https://github.com/user-attachments/assets/9755c5de-dbdd-46cb-be23-c258606e19ab" />

<img width="954" height="937" alt="Image" src="https://github.com/user-attachments/assets/dc87d8c0-187b-443b-aa8c-05c1e4685580" />

Steps to reproduce it:

```
cargo install zellij
zellij
codex
```

Then try scrolling back using Zellij’s scroll mode.

### What is the expected behavior?

Scrolling in Zellij should display the full history of Codex output without truncation.

### What do you see instead?

Scrolling only shows partial output — earlier content is cut off.

alexff8505:
I get the same issue in multiple Mac based Terminals, inc the default.  "Unlimited Scrollback" options can solve the issue if your software has it.

brunopeartree:
Hi, I am experiencing the same issue. Currently this prevents me from switching from claude code.

Artawower:
Same here, this behavior only started showing up very recently

warrenchortle:
Also seeing this on linux

luccahuguet:
have this as well, does not seem to be a problem with zellij. since claude code has no such issue

luccahuguet:
i've got some feedback from zellij's author that might be helpful 

> with the strong disclaimer that I have not tried to reproduce this myself and this is just an educated guess - this happens because the application is relying on the scrollbuffer in "Alternate Screen" (eg. with control code 1049), which the specification mentions should not exist (source: https://invisible-island.net/xterm/ctlseqs/ctlseqs.html under "Alternate Screen Buffer").

> I have tried in the past to be lenient with this behavior, because this is a common issue with some apps (can't blame them, not everyone spends time reading xterm specs), but allowing scrollbuffer in these cases causes a lot of other problems in apps that do follow the spec. Feel free to quote me on this.

yogurt80:
I have the same issue.

luccahuguet:
[...]

PR #16578, opened 2026-04-02
fix(tui): handle zellij redraw and composer rendering

fcoury-oai:
## TL;DR

Fixes the issues when using Codex CLI with Zellij multiplexer. Before this PR there would be no scrollback when using it inside a zellij terminal.

## Problem

Addresses #2558

Zellij does not support ANSI scroll-region manipulation (`DECSTBM` / Reverse Index) or the alternate screen buffer in the way traditional terminals do. When codex's TUI runs inside Zellij, two things break: (1) inline history insertion corrupts the display because the scroll-region escape sequences are silently dropped or mishandled, and (2) the composer textarea renders with inherited background/foreground styles that produce unreadable text against Zellij's pane chrome.

## Mental model

The fix introduces a **Zellij mode** — a runtime boolean detected once at startup via `codex_terminal_detection::terminal_info().is_zellij()` — that gates two subsystems onto Zellij-safe terminal strategies:

- **History insertion** (`insert_history.rs`): Instead of using `DECSTBM` scroll regions and Reverse Index (`ESC M`) to slide content above the viewport, Zellij mode scrolls the screen by emitting `\n` at the bottom row and then writes history lines at absolute positions. This avoids every escape sequence Zellij mishandles.
- **Viewport expansion** (`tui.rs`): When the viewport grows taller than available space, the standard path uses `scroll_region_up` on the backend. Zellij mode instead emits newlines at the screen bottom to push content up, then invalidates the ratatui diff buffer so the next draw is a full repaint.
- **Composer rendering** (`chat_composer.rs`, `textarea.rs`): All text rendering in the input area uses an explicit `base_style` with `Color::Reset` foreground, preventing Zellij's pane styling from bleeding into the textarea. The prompt chevron (`›`) and placeholder text use explicit color constants instead of relying on `.bold()` / `.dim()` modifiers that render inconsistently under Zellij.

## Non-goals

- This change does not fix or improve Zellij's terminal emulation itself.
- It does not rearchitect the inline viewport model; it adds a parallel code path gated on detection.
- It does not touch the alternate-screen disable logic (that already existed and continues to use `is_zellij` via the same detection).

## Tradeoffs

- **Code duplication in `insert_history.rs`**: The Zellij and Standard branches share the line-rendering loop (color setup, span merging, `write_spans`) but differ in the scrolling preamble. The duplication is intentional — merging them would force a complex conditional state machine that's harder to reason about than two flat sequences.
- **`invalidate_viewport` after every Zellij history flush or viewport expansion**: This forces a full repaint on every draw cycle in Zellij, which is more expensive than ratatui's normal diff-based rendering. This is necessary because Zellij's lack of scroll-region support means the diff buffer's assumptions about what's on screen are invalid after we manually move content.
- **Explicit colors vs semantic modifiers**: Replacing `.bold()` / `.dim()` with `Color::Cyan` / `Color::DarkGray` / `Color::White` in the Zellij branch sacrifices theme-awareness for correctness. If the project ever adopts a theming system, Zellij styling will need to participate.

## Architecture

The Zellij detection flag flows through three layers:

1. **`codex_terminal_detection`** — `TerminalInfo::is_zellij()` (new convenience method) reads the already-detected `Multiplexer` variant.
2. **`Tui` struct** — caches `is_zellij` at construction; passes it into `update_inline_viewport`, `flush_pending_history_lines`, and `insert_history_lines_with_mode`.
3. **`ChatComposer` struct** — independently caches `is_zellij` at construction; uses it in `render_textarea` for style decisions.

The two caches (`Tui.is_zellij` and `ChatComposer.is_zellij`) are read from the same global `OnceLock<TerminalInfo>`, so they always agree.

## Observability

No new logging, metrics, or tracing is introduced. Diagnosis depends on:
- Whether `ZELLIJ` or `ZELLIJ_SESSION_NAME` env vars are set (the detection heuristic).
- Visual inspection of the rendered TUI inside Zellij vs a standard terminal.
- The insta snapshot `zellij_empty_composer` captures the Zellij-mode render path.

## Tests

- `terminal_info_reports_is_zellij` — unit test in `terminal-detection` confirming the convenience method.
- `zellij_empty_composer_snapshot` — insta snapshot in `chat_composer` validating the Zellij render path for an empty composer.
- `vt100_zellij_mode_inserts_history_and_updates_viewport` — integration test in `insert_history` verifying that Zellij-mode history insertion writes content and shifts the viewport.

### DIFF
--- a/codex-rs/terminal-detection/src/lib.rs
+++ b/codex-rs/terminal-detection/src/lib.rs
@@ -204,6 +204,11 @@ impl TerminalInfo {
 
         sanitize_header_value(raw)
     }
+
+    /// Returns whether the active terminal multiplexer is Zellij.
+    pub fn is_zellij(&self) -> bool {
+        matches!(self.multiplexer, Some(Multiplexer::Zellij {}))
+    }
 }
 
 static TERMINAL_INFO: OnceLock<TerminalInfo> = OnceLock::new();
--- a/codex-rs/terminal-detection/src/terminal_tests.rs
+++ b/codex-rs/terminal-detection/src/terminal_tests.rs
@@ -126,18 +126,18 @@ fn detects_term_program() {
 fn terminal_info_reports_is_zellij() {
     let zellij = terminal_info(
         TerminalName::Unknown,
-        None,
-        None,
-        None,
+        /*term_program*/ None,
+        /*version*/ None,
+        /*term*/ None,
         Some(Multiplexer::Zellij {}),
     );
     assert!(zellij.is_zellij());
 
     let non_zellij = terminal_info(
         TerminalName::Unknown,
-        None,
-        None,
-        None,
+        /*term_program*/ None,
+        /*version*/ None,
+        /*term*/ None,
         Some(Multiplexer::Tmux { version: None }),
     );
     assert!(!non_zellij.is_zellij());
--- a/codex-rs/terminal-detection/src/terminal_tests.rs
+++ b/codex-rs/terminal-detection/src/terminal_tests.rs
@@ -123,6 +123,27 @@ fn detects_term_program() {
 }
 
 #[test]
+fn terminal_info_reports_is_zellij() {
+    let zellij = terminal_info(
+        TerminalName::Unknown,
+        None,
+        None,
+        None,
+        Some(Multiplexer::Zellij {}),
+    );
+    assert!(zellij.is_zellij());
+
+    let non_zellij = terminal_info(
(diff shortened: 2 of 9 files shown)


etraut-openai (approved):
It's unfortunate that we need to add hacks like this to accommodate a non-compliant terminal emulator that's missing key features, but given the popularity of Zellij, I don't think we can ignore it. I don't see a better solution. I worry about the maintainability of this approach.

PR #16582, opened 2026-04-02
feat: single app-server bootstrap in TUI

jif-oai:
Before this, the TUI was starting 2 app-server. One to check the login status and one to actually start the session

This PR make only one app-server startup and defer the login check in async, outside of the frame rendering path


codex-rs/tui/src/onboarding/onboarding_screen.rs
```diff
@@ -519,9 +519,6 @@ pub(crate) async fn run_onboarding_app(
             }
         }
     }
-    if let Some(app_server) = app_server {
```
bolinfest, line 536: Should the caller of `run_onboarding_app()` do this instead?

codex-rs/tui/src/lib.rs
```diff
@@ -979,29 +980,32 @@ async fn run_ratatui_app(
     // Initialize high-fidelity session event logging if enabled.
     session_log::maybe_init(&initial_config);
 
+    let mut app_server = Some(
```
bolinfest, line 1020: I guess the main question is: is it ever a problem that we unconditionally start the app server?

codex-rs/tui/src/onboarding/onboarding_screen.rs
jif-oai: I'm reworking this in another way

codex-rs/tui/src/lib.rs
jif-oai: I think we might need to restart it (after onboarding for example) but I don't think that just starting it might be a problem 🤔

fcoury-oai (approved):
Did some overall testing, specially around /fork and /resume and everything seems to work as expected.

Code looks good too, approved.

Issue #16389, opened 2026-04-01
Interactive `/review` follow-up can crash codex-cli 0.118.0 with active turn ID mismatch


toku345:
### What version of Codex CLI is running?

codex-cli 0.118.0

### What subscription do you have?

ChatGPT Business, ChatGPT Plus

### Which model were you using?

gpt-5.4

### What platform is your computer?

Darwin 24.6.0 arm64 arm, Linux 6.17.0-1014-nvidia aarch64 aarch64

### What terminal emulator and version are you using (if applicable)?

iTerm2, tmux 3.6a

### What issue are you seeing?

I searched existing issues first and found some related reports, but not an identical one:

- #14077
- #15842
- #13892
- #16339

This report is specifically about the interactive CLI `/review` slash command: if I send a follow-up message while the review is still running, codex-cli can crash with an active turn ID mismatch.

While `/review` was still running, I sent an additional follow-up message to refine the review context.

The follow-up message was:

`Issue #26 も参考にしてください`

Immediately after that, codex-cli crashed with:

```text
Error: turn/steer failed: expected active turn id `019d4673-d45f-7a51-8129-1ac29e120433` but found `019d4673-d38f-7121-a7dc-d13f0dcb9f1b`
```

This does not look like a normal `Review was interrupted` case, because the CLI process itself crashed.

I also reproduced the same issue on a second machine running:
`Linux 6.17.0-1014-nvidia aarch64 aarch64`

In the Linux reproduction, the follow-up input was:

`see Issue #9`

and codex-cli crashed with:

```text
Error: turn/steer failed: expected active turn id `019d46c1-a2ce-7a41-844d-30e5bde8ac2e` but found `019d46c1-a2ae-79b3-b928-7b6c2fa76471`
```

So this does not appear to be macOS-specific.

### What steps can reproduce the bug?

1. Start the interactive CLI with `codex` in a repository.
2. Run the `/review` slash command.
3. While the review is still actively running, send a follow-up message instead of waiting for the review to finish.
4. In my reproductions, the follow-up messages were:
   - `Issue #26 も参考にしてください`
   - `see Issue #9`
5. Observe that codex-cli crashes with an active turn ID mismatch, for example:

   ```text
   Error: turn/steer failed: expected active turn id `019d4673-d45f-7a51-8129-1ac29e120433` but found `019d4673-d38f-7121-a7dc-d13f0dcb9f1b`
   ```

Affected local thread IDs from reproduced sessions:

[...]

PR #14717, opened 2026-03-14
Move TUI on top of app server (parallel code)

etraut-openai:
This PR replicates the `tui` code directory and creates a temporary parallel `tui_app_server` directory. It also implements a new feature flag `tui_app_server` to select between the two tui implementations.

Once the new app-server-based TUI is stabilized, we'll delete the old `tui` directory and feature flag.

PR #16588, opened 2026-04-02
Fix stale turn steering during TUI review follow-ups

etraut-openai:
Addresses #16389

Problem: `/review` follow-ups can crash when app-server TUI steers with a stale active turn id; #14717 introduced the client-side race, and #15714 only handled the “no active turn” half.

Solution: Treat turn-id mismatch as stale cached state too, sync to the server’s current turn id, retry once, and let review turns fall into the existing queue path.


fcoury-oai (approved):
I was able to replicate this issue and check it doesn't happen in this branch.

Approved!

PR #16590, opened 2026-04-02
Fix non-determinism in rules_rs/crate_git_repository.bzl

tyler-french:
Running multiple builds with no changes causes some differences, we see that https://app.buildbuddy.io/compare/a9719629-1660-4735-a477-d66357f234fb...df85310b-eb5c-4c10-8b79-4d0449ba6cdd#file shows the file-differences between two Bazel builds.

These differences are caused by a non-deterministic `.git` entry in the rules_rs crates that are created with `crate_git_repository`.

As a way to make these deterministic, we can remove this entry after we download the git source, so that the input to the compile action is deterministic.

### CLA

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -71,6 +71,7 @@ single_version_override(
     patch_strip = 1,
     patches = [
         "//patches:rules_rs_windows_gnullvm_exec.patch",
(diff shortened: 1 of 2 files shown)


tyler-french:

bolinfest (approved):

Issue #16560, opened 2026-04-02
Forked session id on /status command


hessel-vo:
### What variant of Codex are you using?

CLI

### What feature would you like to see?

Until recently, the /status overview in an interactive session also included the session id of the original session after a fork.
This was removed and the /status command now only shows the current session id.
Please add back the session id in the /status overview, it was useful for tracking forking of sessions.

etraut-openai:
Thanks for the bug report. This will be fixed in the next release.

PR #16596, opened 2026-04-02
Fix fork source display in /status (expose forked_from_id in app server)

etraut-openai:
Addresses #16560

Problem: `/status` stopped showing the source thread id in forked TUI sessions after the app-server migration.

Solution: Carry fork source ids through app-server v2 thread data and the TUI session adapter, and update TUI fixtures so `/status` matches the old TUI behavior.


fcoury-oai (commented):
Replicated the issue and tested the fix. 

Approved.

fcoury-oai (approved):

PR #16601, opened 2026-04-02
Fix resume picker stale thread names

etraut-openai:
Addresses #16562

Problem: Resume picker could keep a stale backend-provided thread title instead of the latest name from session_index.jsonl.

Solution: Always backfill/override picker row names from local session_index.jsonl and cover stale-name replacement with a regression test.


fcoury-oai (approved):
Looks good, approved.

PR #16638, opened 2026-04-03
[codex-analytics] add protocol-native turn timestamps

rhan-oai:
- introduces core-centric timestamp definitions for turns (`started_at`, `completed_at`, `duration_ms`)
- adds `started_at` time to TurnSummary for access at completion

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/16638).
* #16870
* #16706
* #16659
* #16641
* #16640
* __->__ #16638


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -5043,6 +5043,9 @@ pub struct ThreadNameUpdatedNotification {
 pub struct TurnStartedNotification {
     pub thread_id: String,
     pub turn: Turn,
+    /// Unix timestamp (in seconds) when the turn started.
+    #[ts(type = "number")]
+    pub created_at: i64,
```
owenlin0: Could we update the Turn object itself to have `created_at`, `completed_at`, and `duration_ms` fields? That way these are all persisted and exposed through all the various APIs like `thread/read`, `thread/resume`, etc.

That likely means we need to update EventMsg::TurnStarted and EventMsg::TurnCompleted to start storing timestamps, and unfortunately we'd probably have to also make the timestamps optional since historical turns won't have them. But I think the tradeoffs are worth it

owenlin0 (commented):
these fields are going to be super helpful to have :)

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -1942,11 +1961,17 @@ async fn handle_turn_plan_update(
...
+    completed_at: Option<i64>,
+    duration_ms: Option<i64>,
+}
+
 async fn emit_turn_completed_with_status(
     conversation_id: ThreadId,
     event_turn_id: String,
     status: TurnStatus,
     error: Option<TurnError>,
+    turn_completion_metadata: TurnCompletionMetadata,
```
owenlin0, line 1899: nit: `status` and `error` also feel like `TurnCompletionMetadata`, should we lump them together?

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -1956,6 +1981,9 @@ async fn emit_turn_completed_with_status(
             items: vec![],
             error,
             status,
+            created_at: None,
```
owenlin0: doesn't feel great that this would always be `None` for all `turn/completed` notifications. Is there a way to populate it?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3691,6 +3691,15 @@ pub struct Turn {
     pub status: TurnStatus,
     /// Only populated when the Turn's status is failed.
     pub error: Option<TurnError>,
+    /// Unix timestamp (in seconds) when the turn started.
+    #[ts(type = "number | null")]
+    pub created_at: Option<i64>,
```
owenlin0: nit: should we rename this to be `started_at`? The reason is that there are two possible timestamps to be using:
1. the timestamp of the `turn/start` RPC call
2. the timestamp when the core turn operation (i.e. `Op::UserInput`) actually starts - core operations are queued

And it represents 2)

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -6929,6 +6932,9 @@ impl CodexMessageProcessor {
             items,
             error: None,
             status: TurnStatus::InProgress,
+            created_at: None,
```
owenlin0: We should be populating `created_at` whenever we can. I see a lot of places where it's hardcoded to `None` 🤔

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -2901,6 +2946,24 @@ mod tests {
...
+            duration_ms: None,
+        }
+    }
+
+    fn turn_aborted_event(turn_id: &str) -> TurnAbortedEvent {
+        TurnAbortedEvent {
+            turn_id: Some(turn_id.to_string()),
+            reason: codex_protocol::protocol::TurnAbortReason::Interrupted,
+            completed_at: None,
+            duration_ms: None,
```
owenlin0: nit: these are test helpers, right? could we populate these to some non-None values to be more representative?

codex-rs/app-server/src/codex_message_processor.rs
rhan-oai: given the semantics of `started_at`, it doesn't make sense to populate these in app-server before core turn start as defined in turn_timing.rs is established

owenlin0 (approved):

codex-rs/app-server/src/codex_message_processor.rs
owenlin0: yes, fair point!

PR #16640, opened 2026-04-03
[codex-analytics] feature plumbing and emittance

rhan-oai:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/16640).
* #16870
* #16706
* #16641
* __->__ #16640

### DIFF
--- a/codex-rs/analytics/src/reducer.rs
+++ b/codex-rs/analytics/src/reducer.rs
@@ -8,6 +8,8 @@ use crate::events::CodexPluginUsedEventRequest;
 use crate::events::CodexRuntimeMetadata;
 use crate::events::CodexTurnEventParams;
 use crate::events::CodexTurnEventRequest;
+use crate::events::CodexTurnSteerEventParams;
+use crate::events::CodexTurnSteerEventRequest;
 use crate::events::GuardianReviewEventParams;
 use crate::events::GuardianReviewEventPayload;
 use crate::events::GuardianReviewEventRequest;
@@ -26,6 +28,7 @@ use crate::events::subagent_source_name;
 use crate::events::subagent_thread_started_event_request;
 use crate::events::thread_source_name;
 use crate::facts::AnalyticsFact;
+use crate::facts::AnalyticsJsonRpcError;
 use crate::facts::AppMentionedInput;
 use crate::facts::AppUsedInput;
 use crate::facts::CodexCompactionEvent;
@@ -38,12 +41,17 @@ use crate::facts::SubAgentThreadStartedInput;
 use crate::facts::ThreadInitializationMode;
 use crate::facts::TurnResolvedConfigFact;
 use crate::facts::TurnStatus;
+use crate::facts::TurnSteerRejectionReason;
+use crate::facts::TurnSteerResult;
+use crate::facts::TurnTokenUsageFact;
+use crate::now_unix_seconds;
 use codex_app_server_protocol::ClientRequest;
 use codex_app_server_protocol::ClientResponse;
 use codex_app_server_protocol::CodexErrorInfo;
 use codex_app_server_protocol::InitializeParams;
 use codex_app_server_protocol::RequestId;
 use codex_app_server_protocol::ServerNotification;
+use codex_app_server_protocol::TurnSteerResponse;
 use codex_app_server_protocol::UserInput;
 use codex_git_utils::collect_git_info;
 use codex_git_utils::get_git_repo_root;
@@ -54,6 +62,7 @@ use codex_protocol::config_types::ReasoningSummary;
 use codex_protocol::protocol::SandboxPolicy;
 use codex_protocol::protocol::SessionSource;
 use codex_protocol::protocol::SkillScope;
+use codex_protocol::protocol::TokenUsage;
 use sha1::Digest;
 use std::collections::HashMap;
 use std::path::Path;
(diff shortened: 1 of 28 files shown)


codex-rs/protocol/src/protocol.rs
```diff
@@ -2263,6 +2263,14 @@ pub enum InitialHistory {
 }
 
 impl InitialHistory {
+    pub fn scan_rollout_items(&self, mut predicate: impl FnMut(&RolloutItem) -> bool) -> bool {
+        match self {
+            InitialHistory::New => false,
```
rhan-oai: determining is_first_turn for resumed or forked threads is tricky without explicit metadata, predicate looks for first user turn so performance cost should be negligible on average

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -1942,11 +1967,18 @@ async fn handle_turn_plan_update(
     }
 }
 
+struct TurnCompletionMetadata {
+    completed_at: Option<i64>,
+    duration_ms: Option<i64>,
+}
```
rhan-oai: to avoid too many args warning / group data

codex-rs/core/src/thread_manager.rs
```diff
@@ -219,6 +221,24 @@ impl ThreadManager {
...
+            config,
+            auth_manager,
+            session_source,
+            collaboration_modes_config,
+            environment_manager,
+            /*analytics_events_client*/ None,
+        )
+    }
+
+    pub fn new_with_analytics_events_client(
```
owenlin0: codex tends to do this a lot.... instead of defining a new function with a new parameter, can we just update `new` + the callsites? otherwise we will have function sprawl every time a new param is added

codex-rs/core/src/codex.rs
```diff
@@ -6321,6 +6347,42 @@ pub(crate) async fn run_turn(
     last_agent_message
 }
 
+async fn track_turn_resolved_config_analytics(
```
owenlin0, line 6502: should this function no-op if features.general_analytics is not enabled?

codex-rs/analytics/src/reducer.rs
```diff
@@ -275,6 +469,143 @@ impl AnalyticsReducer {
...
+        SandboxPolicy::ReadOnly { .. } => "read_only",
+        SandboxPolicy::WorkspaceWrite { .. } => "workspace_write",
+        SandboxPolicy::ExternalSandbox { .. } => "external_sandbox",
+    }
+}
+
+fn collaboration_mode_mode(mode: ModeKind) -> &'static str {
+    match mode {
+        ModeKind::Plan => "plan",
+        ModeKind::Default | ModeKind::PairProgramming | ModeKind::Execute => "default",
```
owenlin0, line 956: hmm why do we coerce PairProgramming and Execute to "default"?

codex-rs/core/src/codex.rs
```diff
@@ -817,6 +822,17 @@ pub(crate) fn session_loop_termination_from_handle(
     .shared()
 }
 
+fn initial_history_has_prior_user_turns(conversation_history: &InitialHistory) -> bool {
+    conversation_history.scan_rollout_items(rollout_item_is_user_turn_boundary)
+}
+
+fn rollout_item_is_user_turn_boundary(item: &RolloutItem) -> bool {
```
owenlin0: nit: can we move these two functions to `codex-rs/core/src/thread_rollout_truncation.rs`? 

can probably have something like:
```
// codex-rs/core/src/thread_rollout_truncation.rs
pub(crate) fn initial_history_has_user_turns(history: &InitialHistory) -> bool {
    match history {
        InitialHistory::New => false,
        InitialHistory::Resumed(resumed) => rollout_has_user_turns(&resumed.history),
        InitialHistory::Forked(items) => rollout_has_user_turns(items),
    }
}

pub(crate) fn rollout_has_user_turns(items: &[RolloutItem]) -> bool {
    !user_turn_positions_in_rollout(items).is_empty()
}
```

owenlin0 (commented):
Optional follow-up re: tests - we could probably move the failed/interrupted turn analytics permutations into the reducer tests and keep just one app-server integration test for the end-to-end turn event. would keep the app-server integration tests a bit more focused.

owenlin0 (approved):

codex-rs/analytics/src/reducer.rs
rhan-oai: both serialize to default as per protocol/src/config_types.rs

```
pub enum ModeKind {
    Plan,
    #[default]
    #[serde(
        alias = "code",
        alias = "pair_programming",
        alias = "execute",
        alias = "custom"
    )]
    Default,
    #[doc(hidden)]
    #[serde(skip_serializing, skip_deserializing)]
    #[schemars(skip)]
    #[ts(skip)]
    PairProgramming,
    #[doc(hidden)]
    #[serde(skip_serializing, skip_deserializing)]
    #[schemars(skip)]
    #[ts(skip)]
    Execute,
}

pub const TUI_VISIBLE_COLLABORATION_MODES: [ModeKind; 2] = [ModeKind::Default, ModeKind::Plan];
```

owenlin0: oh interesting. ok ignore me

PR #16641, opened 2026-04-03
[codex-analytics] add token usage metadata

rhan-oai:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/16641).
* #16870
* #16706
* __->__ #16641
* #16640


owenlin0 (commented):
hmm this is a bit trickier than it initially looks. had codex double-check and codex's comment looks right to me:

> I think this is slightly off as-is. `ThreadTokenUsageUpdatedNotification.token_usage.last` looks like “last completed token usage update / API response,” not necessarily “all tokens used by this turn.”
> 
> That works for simple turns with one model response, but it can undercount turns that loop through tools or otherwise make multiple sampling requests under the same `turn_id`. In that case each notification’s `last` is just the most recent response, while the turn event fields sound like they should represent the whole turn.
> 
> Since the notification includes both `turn_id` and cumulative `token_usage.total`, I think the cleaner fix is to compute the turn delta from the cumulative totals instead of copying `last` directly:
> 
> - On the first token usage update for a turn, derive a baseline as `total - last`.
> - On each update for that same turn, set the turn usage to `latest_total - baseline`.
> - Use saturating field-wise subtraction.
> 
> That keeps this using the existing app-server notification, but avoids treating the thread-level notification’s `last` field as turn-level usage.

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -2204,6 +2211,11 @@ async fn handle_token_count_event(
             turn_id,
             token_usage,
         };
+        if let Some(analytics_events_client) = analytics_events_client {
```
owenlin0: do we still need these changes now that we're emitting a custom turn token fact?

owenlin0 (approved):

PR #16646, opened 2026-04-03
Fix stale thread-name resume lookups

etraut-openai:
Addresses #15943

Problem: Name-based resume could stop on a newer session_index entry whose rollout was never persisted, shadowing an older saved thread with the same name.

Solution: Materialize rollouts before indexing thread names and make name lookup skip unresolved entries until it finds a persisted rollout.

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -2217,14 +2217,19 @@ impl Session {
         }
     }
 
-    pub(crate) async fn ensure_rollout_materialized(&self) {
+    pub(crate) async fn try_ensure_rollout_materialized(&self) -> std::io::Result<()> {
         let recorder = {
             let guard = self.services.rollout.lock().await;
             guard.clone()
         };
-        if let Some(rec) = recorder
-            && let Err(e) = rec.persist().await
-        {
+        if let Some(rec) = recorder {
+            rec.persist().await?;
+        }
+        Ok(())
+    }
+
+    pub(crate) async fn ensure_rollout_materialized(&self) {
+        if let Err(e) = self.try_ensure_rollout_materialized().await {
             warn!("failed to materialize rollout recorder: {e}");
         }
     }
@@ -5540,6 +5545,18 @@ mod handlers {
             sess.send_event_raw(event).await;
             return;
         };
+
+        if let Err(e) = sess.try_ensure_rollout_materialized().await {
+            let event = Event {
+                id: sub_id,
+                msg: EventMsg::Error(ErrorEvent {
+                    message: format!("Failed to set thread name: {e}"),
+                    codex_error_info: Some(CodexErrorInfo::Other),
+                }),
+            };
+            sess.send_event_raw(event).await;
+            return;
+        }
 
         let codex_home = sess.codex_home().await;
         if let Err(e) =
--- a/codex-rs/rollout/src/session_index.rs
+++ b/codex-rs/rollout/src/session_index.rs
@@ -111,11 +111,12 @@ pub async fn find_thread_names_by_ids(
     Ok(names)
 }
 
-/// Find the most recently updated thread id for a thread name, if any.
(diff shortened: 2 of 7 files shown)


codex-rs/core/src/codex.rs
```diff
@@ -5375,6 +5375,8 @@ mod handlers {
             return;
         };
 
+        sess.ensure_rollout_materialized().await;
```
jif-oai: This needs to observe/propagate materialization failure before appending to session_index.jsonl; otherwise we can still index a name for a non-resumable thread.

codex-rs/rollout/src/session_index.rs
```diff
@@ -136,10 +136,32 @@ pub async fn find_thread_path_by_name_str(
...
-    super::list::find_thread_path_by_id_str(codex_home, &thread_id.to_string()).await
+    }
+    let path = session_index_path(codex_home);
+    if !path.exists() {
+        return Ok(None);
+    }
+    let name = name.to_string();
+    // Collect all matching thread ids newest-first instead of stopping at the first name hit:
+    // the newest entry may point at a thread whose rollout was never materialized.
+    let thread_ids =
```
jif-oai: Can we stream candidates and resolve newest-first instead of preloading every matching ID? This can make common name resume do full-index work plus many per-ID file searches...

Ideally, most of this should be migrated to the DB as well

codex-rs/rollout/src/session_index.rs
```diff
@@ -136,10 +136,32 @@ pub async fn find_thread_path_by_name_str(
...
+        tokio::task::spawn_blocking(move || collect_thread_ids_from_end_by_name(&path, &name))
+            .await
+            .map_err(std::io::Error::other)??;
+
+    for thread_id in thread_ids {
+        // Keep walking until a matching id resolves to an existing rollout so an unsaved rename
+        // cannot shadow an older persisted session with the same name.
+        if let Some(path) =
```
jif-oai, line 137: This should validate that the candidate rollout is loadable or at least contains session metadata,...otherwise a zero-byte/partially-created rollout can still shadow an older valid thread

codex-rs/core/src/codex.rs
```diff
@@ -5498,6 +5503,18 @@ mod handlers {
             return;
         };
 
+        if let Err(e) = sess.try_ensure_rollout_materialized().await {
```
jif-oai, line 5549: This new materialization-error path still races with the TUI’s optimistic success message: the TUI inserts “Thread renamed ... codex resume ...” before `SetThreadName` is acked, so a persist failure here can still leave a false resume hint.
Can we move that success/resume hint to `ThreadNameUpdated` or add a pending/error state?

codex-rs/rollout/src/session_index.rs
```diff
@@ -111,35 +111,41 @@ pub async fn find_thread_names_by_ids(
...
-) -> std::io::Result<Option<PathBuf>> {
-    let Some(thread_id) = find_thread_id_by_name(codex_home, name).await? else {
-        return Ok(None);
-    };
-    super::list::find_thread_path_by_id_str(codex_home, &thread_id.to_string()).await
+    while let Some(thread_id) = rx.recv().await {
+        // Keep walking until a matching id resolves to a loadable rollout so an unsaved or partial
+        // rename cannot shadow an older persisted session with the same name.
+        if let Some(path) =
+            super::list::find_thread_path_by_id_str(codex_home, &thread_id.to_string()).await?
```
jif-oai, line 138: this can be extremely expensive... we have to migrate to a DB look-up. Adding on my todo

jif-oai (approved):

codex-rs/rollout/src/session_index.rs
etraut-openai: Agreed. I kept this PR to the streaming newest-first fallback; a DB-backed lookup is the right follow-up for the broader cost issue.

Issue #15282, opened 2026-03-20
Suppress the warning `⚠ Codex could not find system bubblewrap at /usr/bin/bwrap` when sandbox is bypassed.


your-diary:
### What variant of Codex are you using?

CLI

### What feature would you like to see?

After upgrading to `v0.116.0`, I get this warning every time I start `codex`:

```
⚠ Codex could not find system bubblewrap at /usr/bin/bwrap. Please install bubblewrap with your package manager. Codex will use the vendored bubblewrap in the meantime.
```

I want this to be shown only when sandboxing is enabled.

## Why?

I'm using `codex` exclusively inside a Docker container with `--dangerously-bypass-approvals-and-sandbox`.

If I understand correctly, `bwrap` is not used in this setup.

If so, being forced to install `bwrap` seems not reasonable.

### Additional information

Probably related entry in the [changelog](https://github.com/openai/codex/releases/tag/rust-v0.116.0):

> Improved Linux sandbox startup on symlinked checkouts, missing writable roots, and Ubuntu/AppArmor hosts by preferring system bwrap when available. (https://github.com/openai/codex/pull/14849, https://github.com/openai/codex/pull/14890, https://github.com/openai/codex/pull/14963)

SomwareHR:
I, too, ran into this warning message.
Currently I am not willing to install bubblewrap.
Can `use_legacy_landlock = true` in `~/.codex/config.toml` help in this situation? I didn't implement it yet, I am not sure if it would make any meaningful change.
Thank you
BR
sw

snabb:
Just a hint that can potentially benefit others that come here after searching with the error message. I am on Ubuntu 24.04 and encountered this issue even though I had bubblewrap installed.

The following was needed to make bubblewrap work:
```
sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
```

If you want to have it applied automatically on next reboot, you can do something like this:
```
echo 'kernel.apparmor_restrict_unprivileged_userns=0' | sudo tee /etc/sysctl.d/99-codex-userns.conf
```

TonyGravagno:
Still seeing message from CLI but I don't care.

seakros:
From their page (https://developers.openai.com/codex/concepts/sandboxing):

[...]

PR #16667, opened 2026-04-03
Suppress bwrap warning when sandboxing is bypassed

etraut-openai:
Addresses #15282

Problem: Codex warned about missing system bubblewrap even when sandboxing was disabled.

Solution: Gate the bwrap warning on the active sandbox policy and skip it for danger-full-access and external-sandbox modes.


bolinfest (approved):
Thanks for the clean fix!

Issue #11555, opened 2026-02-12
TUI composer sometimes auto-fills with `codex(pid) MallocSt...` text


0xWerz:
## Summary
Codex CLI TUI input sometimes auto-fills with text like:
`codex(11033) MallocSt...`

I did not type or paste this. The PID part changes over time (e.g. `codex(8582)`, `codex(10358)`, `codex(11033)`).

## Screenshot
<img width="947" height="185" alt="image" src="https://github.com/user-attachments/assets/f814fef4-ed71-43f8-92fa-5412fbffc4f2" />

## Environment
- macOS
- Codex CLI TUI

### Runtime repro (intermittent)
1. Start Codex CLI TUI.
2. Keep using it for a while (normal interaction, tool runs, shell output, etc.).
3. Eventually the composer may auto-fill with `codex(<pid>) MallocSt...` text.

## Expected behavior
Input should only contain user-typed or user-pasted text.

## Actual behavior
Composer gets unexpected text that looks like allocator diagnostics.

## Analysis / hypothesis
- The text appears to match macOS malloc stack-logging diagnostics.
- The changing PID supports that it is process-generated noise, not static input.
- In terminals without reliable bracketed paste signaling, fast char streams can be classified as paste bursts. This likely lets diagnostic text leak into composer input as if it were pasted text.

## Potential fix direction
- Add a guard in paste-burst flush handling for this diagnostic shape.
- Avoid silent data loss: if uncertain, quarantine as placeholder-backed pending paste rather than dropping content outright.

I can help validate fixes quickly on macOS if needed.

jiabaogithub:
does it fixed? i also found this issue in ver v0.121.0(current latest version)

LiuDeng:
This issue still exists in v0.124.0

LiuDeng:
@etraut-openai

jamesx0416:
Still happens in v0.125.0

yuvrajangadsingh:
@etraut-openai the fix from #16699 is on main and shipped in v0.119.0+ (commit a71fc47cf8 — strips `MallocStackLogging*` and `MallocLogFile*` in `pre_main_hardening_macos`). but the recent reports here are all post-fix:

- @jiabaogithub on v0.121.0 (apr 17)
- @LiuDeng on v0.124.0 (apr 25)
- @jamesx0416 on v0.125.0 (may 3)

[...]

PR #16699, opened 2026-04-03
Fix macOS malloc diagnostics leaking into TUI composer

etraut-openai:
Addresses #11555

Problem: macOS malloc stack-logging diagnostics could leak into the TUI composer and get misclassified as pasted user input.

Solution: Strip `MallocStackLogging*` and `MallocLogFile*` during macOS pre-main hardening and document the additional env cleanup.

### DIFF
--- a/codex-rs/process-hardening/src/lib.rs
+++ b/codex-rs/process-hardening/src/lib.rs
@@ -185,32 +185,4 @@ mod tests {
         assert_eq!(keys.len(), 1);
         assert_eq!(keys[0].as_os_str(), ld_test_var);
     }
-
-    #[test]
-    fn env_keys_with_prefix_matches_malloc_stack_logging_variants() {
-        let vars = vec![
-            (OsString::from("MallocStackLogging"), OsString::from("1")),
-            (
-                OsString::from("MallocStackLoggingNoCompact"),
-                OsString::from("1"),
-            ),
-            (
-                OsString::from("MallocLogFile"),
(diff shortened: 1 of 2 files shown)


codex-rs/process-hardening/src/lib.rs
```diff
@@ -96,13 +86,12 @@ pub(crate) fn pre_main_hardening_macos() {
...
-    let dyld_keys = env_keys_with_prefix(std::env::vars_os(), b"DYLD_");
+    remove_env_vars_with_prefix(b"DYLD_");
 
-    for key in dyld_keys {
-        unsafe {
-            std::env::remove_var(key);
-        }
-    }
+    // Remove macOS malloc stack-logging controls so allocator diagnostics from
+    // Codex or inherited child processes do not get sprayed into the TUI.
```
bolinfest: I think this is a reasonable fix, though I don't love it in the event that someone is trying to use these environment variables for legitimate purposes, and also, this is specific to the TUI and is not an issue for `codex app-server`, for example. That is, this fix changes behavior for more cases than would be ideal, especially since it seems to be specific to how paste bursts are processed.

Nevertheless, given all the folks who have commented on https://github.com/openai/codex/issues/11555, this seems like the right thing to do.

My only request is to amend this comment to link to https://github.com/openai/codex/issues/11555 explicitly to justify this code to the reader.

bolinfest (approved):
Thanks for fixing this up!

Issue #16124, opened 2026-03-28
Support remote `--cd` for `codex --remote` command


AdithyanI:
### What issue are you seeing?

Human-readable setup:
- app-server is running on a remote Mac mini
- the `codex --remote ...` client is running on a separate MacBook
- the client can connect to the remote app-server normally
- the failure only happens when passing `--cd` with a path that exists on the remote machine but not on the local machine

`codex --remote ... --cd <remote-path>` fails with:

```text
Error: No such file or directory (os error 2)
```

This happens even when the remote app-server is running and the same `cwd` is valid on the remote host.

I verified that the remote app-server itself accepts the exact same path over the websocket protocol via `thread/start` with `cwd` set, and returns that `cwd` in the response. That suggests the issue is in the stock CLI's `--remote` / `--cd` handling rather than app-server `cwd` support.

Environment:
- local client: `codex-cli 0.117.0`
- remote server: `codex-cli 0.117.0`
- local client machine: MacBook
- remote server machine: Mac mini
- local OS: macOS arm64
- remote OS: macOS arm64
- transport: websocket (`codex app-server --listen ws://<REMOTE_HOST>:<PORT>`)

### What steps can reproduce the bug?

1. Start a remote app-server on a machine where a target project directory exists:

```bash
codex app-server --listen ws://<REMOTE_HOST>:<PORT>
```

2. From another machine, verify the server is reachable:

```bash
curl -fsS http://<REMOTE_HOST>:<PORT>/readyz >/dev/null && echo ready
```

3. Try to launch the TUI against that remote server with a remote-only working directory:

```bash
codex --remote ws://<REMOTE_HOST>:<PORT> --no-alt-screen --cd /path/on/remote/host/project 'say hello'
```

4. Observe the failure:

```text
Error: No such file or directory (os error 2)
```

5. As a control, connect to the same app-server directly over websocket and send `thread/start` with the same `cwd`:

```json
{
  "method": "thread/start",
  "id": 2,
  "params": {
    "model": "gpt-5.4",
    "cwd": "/path/on/remote/host/project",
    "approvalPolicy": "never"
  }
}
```

The server accepts it and returns:

```json
{
  "accepted_cwd": "/path/on/remote/host/project",
  "thread_cwd": "/path/on/remote/host/project"
}
```

### What is the expected behavior?

If `--remote` is being used, I would expect `--cd` to be forwarded as the remote thread/session `cwd` without requiring that path to exist on the local machine.

[...]

PR #16700, opened 2026-04-03
Add remote --cd forwarding for app-server sessions

etraut-openai:
Addresses #16124

Problem: `codex --remote --cd <path>` canonicalized the path locally and then omitted it from remote thread lifecycle requests, so remote-only working directories failed or were ignored.

Solution: Keep remote startup on the local cwd, forward explicit `--cd` values verbatim to `thread/start`, `thread/resume`, and `thread/fork`, and cover the behavior with `codex-tui` tests.

Testing: I manually tested `--remote --cd` with both absolute and relative paths and validated correct behavior.

---

Update based on code review feedback:

Problem: Remote `--cd` was forwarded to `thread/resume` and `thread/fork`, but not to `thread/list` lookups, so `--resume --last` and picker flows could select a session from the wrong cwd; relative cwd filters also failed against stored absolute paths.

Solution: Apply explicit remote `--cd` to `thread/list` lookups for `--last` and picker flows, normalize relative cwd filters on the app-server before exact matching, and document/test the behavior.

### DIFF
--- a/codex-rs/app-server/README.md
+++ b/codex-rs/app-server/README.md
@@ -273,7 +273,7 @@ Experimental API: `thread/start`, `thread/resume`, and `thread/fork` accept `per
 - `modelProviders` — restrict results to specific providers; unset, null, or an empty array will include all providers.
 - `sourceKinds` — restrict results to specific sources; omit or pass `[]` for interactive sessions only (`cli`, `vscode`).
 - `archived` — when `true`, list archived threads only. When `false` or `null`, list non-archived threads (default).
-- `cwd` — restrict results to threads whose session cwd exactly matches this path.
+- `cwd` — restrict results to threads whose session cwd exactly matches this path. Relative paths are resolved against the app-server process cwd before matching.
 - `searchTerm` — restrict results to threads whose extracted title contains this substring (case-sensitive).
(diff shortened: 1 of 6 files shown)


fcoury-oai:
From a local Codex review:

1. You have one remote app-server hosting multiple repos, for example /srv/repo-a and /srv/repo-b.
2. Yesterday you worked in /srv/repo-a, and that’s currently the most recently updated thread overall.
3. Today you run:

codex --remote ws://host:3000 --cd /srv/repo-b --resume --last

4. What you expect: Codex resumes the latest thread whose working directory is /srv/repo-b.
5. What can happen with this PR as written: the “find latest thread” lookup still ignores --cd in remote mode, so it may pick the latest / srv/repo-a thread instead. Then the new --cd /srv/repo-b override gets sent to thread/resume for that /srv/repo-a thread.

So the user-visible symptom would be: “I asked to resume my latest session in repo B, but Codex opened some other conversation from repo A, possibly now running it in repo B’s directory.”

That’s risky because it’s not just a confusing selection bug: the resumed conversation’s context and the working directory can now be mismatched, which could lead the agent to edit the wrong checkout or interpret prior discussion against the wrong repo.

etraut-openai:
@codex review

fcoury-oai (approved):
Looks good!

PR #16709, opened 2026-04-03
Sanitize forked child history

aibrahim-oai:
- Keep only parent system/developer/user messages plus assistant final-answer messages in forked child history.
- Strip parent tool/reasoning items and remove the unmatched synthetic spawn output.

### DIFF
--- a/codex-rs/core/src/agent/control_tests.rs
+++ b/codex-rs/core/src/agent/control_tests.rs
@@ -14,6 +14,7 @@ use codex_login::CodexAuth;
 use codex_protocol::AgentPath;
 use codex_protocol::config_types::ModeKind;
 use codex_protocol::models::ContentItem;
+use codex_protocol::models::MessagePhase;
 use codex_protocol::models::ResponseItem;
 use codex_protocol::protocol::ErrorEvent;
 use codex_protocol::protocol::EventMsg;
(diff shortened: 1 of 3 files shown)


codex-rs/core/src/agent/control_tests.rs
```diff
@@ -616,10 +654,19 @@ async fn spawn_agent_can_fork_parent_thread_history() {
...
+            role: "user".to_string(),
+            content: vec![ContentItem::InputText {
+                text: "parent seed context".to_string(),
+            }],
+            end_turn: None,
+            phase: None,
+        },
+        assistant_message("parent final answer", Some(MessagePhase::FinalAnswer)),
+    ];
+    assert_eq!(history.raw_items(), &expected_history);
```
bolinfest: Suggestion: add an assertion message to emphasize what we expect (note what should be filtered)?

codex-rs/core/src/agent/control_tests.rs
```diff
@@ -910,7 +854,7 @@ async fn spawn_agent_fork_last_n_turns_keeps_only_recent_turns() {
         history.raw_items(),
         "queued message"
     ));
-    assert!(history_contains_text(
+    assert!(!history_contains_text(
```
bolinfest: Again, add assertion message for clarity?

bolinfest (approved):

PR #16710, opened 2026-04-03
fix(tui): sort skill mentions by display name first

fcoury-oai:
## Summary

The skill list opened by '$' shows `interface.display_name` preferably if available but the sorting order of the search results use the `skill.name` for sorting the results regardless.

This can be clearly seen in this example below: I expected with "pr" as the search term to have "PR Babysitter" be the first item, but instead it's way down the list.

The reason is because "PR Babysitter" skill name is "babysit-pr" and therefore it doesn't rank as high as "pr-review-triage".

This PR fixes this behavior.

| Before | After |
| --- | --- |
| image | image |

## Testing

- `just fmt`
- `cargo test -p codex-tui bottom_pane::skill_popup::tests::display_name_match_sorting_beats_worse_secondary_search_term_matches --lib -- --exact`
- `cargo test -p codex-tui`

### DIFF
--- a/codex-rs/tui/src/bottom_pane/skill_popup.rs
+++ b/codex-rs/tui/src/bottom_pane/skill_popup.rs
@@ -161,11 +161,14 @@ impl SkillPopup {
                     .sort_rank
                     .cmp(&self.mentions[b.0].sort_rank)
             } else {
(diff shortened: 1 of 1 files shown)


etraut-openai (approved):
Looks good now!

PR #16735, opened 2026-04-03
[codex] allow disabling prompt instruction blocks

tibo-openai:
This PR adds root and profile config switches to omit the generated `<permissions instructions>` and `<apps_instructions>` prompt blocks while keeping both enabled by default, and it gates both the initial developer-context injection and later permissions diff injection so turning the permissions block off stays effective across turn-context overrides.

Also added a prompt debug tool that can be used as `codex debug prompt-input "hello"` and dumps the constructed items list.

### DIFF
--- a/codex-rs/core/src/prompt_debug.rs
+++ b/codex-rs/core/src/prompt_debug.rs
@@ -1,0 +1,152 @@
+use std::collections::HashSet;
+use std::sync::Arc;
+
+use codex_exec_server::EnvironmentManager;
+use codex_features::Feature;
+use codex_login::AuthManager;
+use codex_models_manager::collaboration_mode_presets::CollaborationModesConfig;
+use codex_protocol::error::Result as CodexResult;
+use codex_protocol::models::ResponseInputItem;
+use codex_protocol::models::ResponseItem;
+use codex_protocol::protocol::SessionSource;
+use codex_protocol::user_input::UserInput;
+use tokio_util::sync::CancellationToken;
+
+use crate::codex::Session;
+use crate::codex::build_prompt;
+use crate::codex::built_tools;
+use crate::config::Config;
+use crate::thread_manager::ThreadManager;
+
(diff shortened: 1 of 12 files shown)


codex-rs/core/tests/suite/client.rs
```diff
@@ -1208,47 +1216,19 @@ async fn includes_apps_guidance_as_developer_message_for_chatgpt_auth() {
     wait_for_event(&codex, |ev| matches!(ev, EventMsg::TurnComplete(_))).await;
 
     let request = resp_mock.single_request();
-    let request_body = request.body_json();
-    let input = request_body["input"].as_array().expect("input array");
     let apps_snippet =
         "Apps (Connectors) can be explicitly triggered in user messages in the format";
 
-    let has_developer_apps_guidance = input.iter().any(|item| {
```
aibrahim-oai, line 1216: why do we edit already existing tests given that config doesn't change behavior?

codex-rs/core/src/prompt_debug.rs
```diff
@@ -0,0 +1,152 @@
+use std::collections::HashSet;
```
aibrahim-oai, line 1: I think we can edit the exec to have a mode that prints the rollout files filtered by responses item. Instead of maintaining a light version of it

aibrahim-oai (approved):

codex-rs/core/tests/suite/client.rs
aibrahim-oai: Ideally we should have all tests pass and new tests for the new behavior

tibo-openai: Was just drive-by cleanup / simplification.

codex-rs/core/src/prompt_debug.rs
tibo-openai: Ack. Let's clean this up afterwards.

PR #16736, opened 2026-04-03
Move unified-exec sandbox launch to exec-server

starr-openai:
## Summary
- add a shared `SandboxLaunchConfig` that can be serialized across the exec-server protocol
- move remote unified-exec sandbox argv construction into `codex-exec-server`
- keep zsh-fork and managed-network gaps explicit with TODOs while preserving current behavior

## Validation
- `just fmt`
- remote via `codex-applied-devbox`: `cargo test -p codex-exec-server`
- remote via `codex-applied-devbox`: `cargo test -p codex-sandboxing`
- remote via `codex-applied-devbox`: `cargo test -p codex-core unified_exec`
- remote Docker smoke against containerized exec-server: verified `sandbox=None` -> `SandboxType::None`, `auto` selects `LinuxSeccomp` when policy requires it, and `require` forces `LinuxSeccomp` across env boundary (restrictive auto case in Docker still depends on host kernel userns support)


codex-rs/sandboxing/src/seatbelt_tests.rs
```diff
@@ -27,9 +27,9 @@ use tempfile::TempDir;
 
 fn assert_seatbelt_denied(stderr: &[u8], path: &Path) {
     let stderr = String::from_utf8_lossy(stderr);
-    let expected = format!("bash: {}: Operation not permitted\n", path.display());
+    let path_display = path.display().to_string();
     assert!(
```
pakrym-oai, line 30: nit: simplify the assert logic?

codex-rs/sandboxing/src/manager.rs
```diff
@@ -46,6 +49,68 @@ pub enum SandboxablePreference {
...
+#[serde(rename_all = "kebab-case")]
+pub enum SandboxLaunchMode {
+    Disabled,
+    Auto,
+    Require,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SandboxLaunchConfig {
```
pakrym-oai, line 56: There are too many types with these same fields.  SandboxExecRequest/SandboxTransformRequest

if we are trying to encapsulate all sandbox related stuff in one type we should reuse it in other places.

codex-rs/sandboxing/src/manager.rs
```diff
@@ -46,6 +49,68 @@ pub enum SandboxablePreference {
     Forbid,
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum SandboxLaunchMode {
+    Disabled,
+    Auto,
```
pakrym-oai, line 49: what does auto mean?

codex-rs/sandboxing/src/manager.rs
```diff
@@ -46,6 +49,68 @@ pub enum SandboxablePreference {
     Forbid,
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum SandboxLaunchMode {
+    Disabled,
```
pakrym-oai: can Disabled be `Optional<SandboxLaunchConfig>`?

codex-rs/core/tests/suite/unified_exec.rs
```diff
@@ -947,13 +947,15 @@ async fn unified_exec_terminal_interaction_captures_delayed_output() -> Result<(
...
-        "cmd": "sleep 3 && echo MARKER1 && sleep 3 && echo MARKER2",
+        "cmd": "sleep 2 && echo MARKER1 && sleep 5 && echo MARKER2",
         "yield_time_ms": 10,
         "tty": true,
     });
 
     // Poll stdin three times: first for no output, second after the first marker,
-    // and a final long poll to capture the second marker.
+    // and a final long poll to capture the second marker. Keep a wider buffer
+    // between MARKER1 and process exit so slower environments still reach the
```
pakrym-oai: unrelated?

codex-rs/exec-server/tests/process.rs
```diff
@@ -63,7 +82,69 @@ async fn exec_server_starts_process_over_websocket() -> anyhow::Result<()> {
...
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn exec_server_starts_sandboxed_process_over_websocket() -> anyhow::Result<()> {
+    let mut server = exec_server().await?;
+    initialize_server(&mut server).await?;
+
+    let cwd = std::env::current_dir()?;
+    let (sandbox_mode, expected_sandbox_type) = sandbox_wire_test_mode();
```
pakrym-oai: nit: we should switch these tests to use the trait and be parameterized

codex-rs/exec-server/tests/exec_process.rs
```diff
@@ -225,6 +229,7 @@ async fn remote_exec_process_reports_transport_disconnect() -> Result<()> {
             env: Default::default(),
             tty: false,
             arg0: None,
+            sandbox: None,
```
pakrym-oai: let's add some integration tests that verify sandbox actually works

codex-rs/exec-server/tests/common/exec_server.rs
```diff
@@ -44,6 +44,11 @@ pub(crate) async fn exec_server() -> anyhow::Result<ExecServerHarness> {
     let binary = cargo_bin("codex-exec-server")?;
     let mut child = Command::new(binary);
     child.args(["--listen", "ws://127.0.0.1:0"]);
+    if cfg!(target_os = "linux")
+        && let Ok(sandbox_binary) = cargo_bin("codex-linux-sandbox")
+    {
+        child.env("CODEX_LINUX_SANDBOX_EXE", sandbox_binary);
```
pakrym-oai: this smells fishy, we should need to set this.

Do we need to make exec-server go throught arg0 dispatch?

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -609,10 +609,12 @@ impl UnifiedExecProcessManager {
                     env: env.env.clone(),
                     tty,
                     arg0: env.arg0.clone(),
+                    sandbox: None,
                 })
                 .await
                 .map_err(|err| UnifiedExecError::create_process(err.to_string()))?;
-            return UnifiedExecProcess::from_remote_started(started, env.sandbox).await;
+            let sandbox_type = started.sandbox_type;
```
pakrym-oai: this is a worrying dependency. is sandbox type used for something important?

codex-rs/core/src/tools/runtimes/unified_exec.rs
```diff
@@ -218,6 +245,37 @@ impl<'a> ToolRuntime<UnifiedExecRequest, UnifiedExecProcess> for UnifiedExecRunt
         if let Some(network) = req.network.as_ref() {
             network.apply_to_env(&mut env);
         }
+        // Remote exec-server now owns sandbox argv construction, so this branch
+        // keeps sending raw command data until we collapse the launch APIs.
+        if ctx.turn.environment.exec_server_url().is_some()
+            && should_remote_exec_server_build_sandbox(
+                req.additional_permissions.is_some(),
```
pakrym-oai: Why can't we carry additional permissions? What are they even used for?

codex-rs/core/src/tools/runtimes/unified_exec.rs
```diff
@@ -218,6 +245,37 @@ impl<'a> ToolRuntime<UnifiedExecRequest, UnifiedExecProcess> for UnifiedExecRunt
         if let Some(network) = req.network.as_ref() {
             network.apply_to_env(&mut env);
         }
+        // Remote exec-server now owns sandbox argv construction, so this branch
+        // keeps sending raw command data until we collapse the launch APIs.
+        if ctx.turn.environment.exec_server_url().is_some()
```
pakrym-oai: I really don't love the fork. Can we keep a single code path (and fail for zsh)?

codex-rs/exec-server/src/local_process.rs
```diff
@@ -91,6 +97,33 @@ struct LocalExecProcess {
...
+                .map(|parent| parent.join(CODEX_LINUX_SANDBOX_ARG0))
+                .filter(|candidate| candidate.exists())
+        });
+        Self {
+            codex_linux_sandbox_exe: env_path.or(sibling_path),
+        }
+    }
+}
+
+struct PreparedExecLaunch {
```
pakrym-oai: isn't there a type for this in codex-sandboxing? ExecRequest?

codex-rs/exec-server/src/local_process.rs
```diff
@@ -91,6 +97,33 @@ struct LocalExecProcess {
     wake_tx: watch::Sender<u64>,
 }
 
+#[derive(Clone, Debug, Default)]
+struct ExecServerRuntimeConfig {
+    codex_linux_sandbox_exe: Option<PathBuf>,
+}
+
+impl ExecServerRuntimeConfig {
```
pakrym-oai, line 106: This needs to go, look into codex_arg0

pakrym-oai (changes requested):
1. Reduce complexity in codex-sandboxing if possible
2. Use codex_arg0 to figure out the linux sandbox path
3. Use single code path in unified exec, fail when using zsh/additional permissions with remote.

codex-rs/exec-server/src/protocol.rs
```diff
@@ -61,12 +63,14 @@ pub struct ExecParams {
     pub env: HashMap<String, String>,
     pub tty: bool,
     pub arg0: Option<String>,
+    pub sandbox: Option<SandboxLaunchConfig>,
```
pakrym-oai: Yeah, this Optional doesn't square with SandboxLaunchMode

codex-rs/exec-server/tests/process.rs
pakrym-oai: I see we have another file full of tests, drop this one?

codex-rs/core/src/unified_exec/process_manager.rs
starr-openai: Updated/current head no longer depends on the sandbox type returned from remote startup. The core remote-start path snapshots the selected sandbox type from the launch params before sending the request, then uses that local intent for UX/reporting state.

codex-rs/sandboxing/src/manager.rs
starr-openai: Latest push collapses the remote payload onto `SandboxLaunchConfig` carrying the actual selected `SandboxType`, which removes the extra launch-mode layer and cuts the `Option<SandboxLaunchConfig>` / launch-mode mismatch from this path.

codex-rs/core/src/tools/runtimes/unified_exec.rs
starr-openai: Latest/current head keeps the single remote unified-exec path when `exec_server_url` is set. It still rejects the unsupported zsh-fork mode up front, and it now carries the resolved sandbox config, including network-policy intent, in the `SandboxLaunchConfig` sent to exec-server.

codex-rs/exec-server/src/protocol.rs
starr-openai: Updated/current shape: the exec-server protocol now carries a required `SandboxLaunchConfig`. Disabled sandboxing is represented inside that config (`SandboxType::None` / `SandboxLaunchConfig::no_sandbox(...)`), and the separate launch-mode enum / optional-protocol mismatch is gone.

codex-rs/sandboxing/src/manager.rs
starr-openai: Addressed by the payload change in the latest push: the remote launch config no longer has an `auto` mode at all. It now carries the actual selected `SandboxType` that this turn resolved to before we serialize across exec-server.

codex-rs/exec-server/src/local_process.rs
starr-openai: Updated/corrected in the latest push: the `CODEX_LINUX_SANDBOX_EXE` handoff is gone. `codex-exec-server` now dispatches the Linux sandbox helper from its own executable when argv0 is the sandbox helper name, and the exec-server launch config uses its current executable as that reexec target.

codex-rs/exec-server/src/local_process.rs
```diff
@@ -458,6 +495,47 @@ impl ExecProcess for LocalExecProcess {
...
+    params: &ExecParams,
+    runtime: &ExecServerRuntimeConfig,
+) -> Result<SandboxExecRequest, JSONRPCErrorError> {
+    let command = build_sandbox_command(
+        &params.argv,
+        params.cwd.as_path(),
+        &params.env,
+        params.sandbox.additional_permissions.clone(),
+    )?;
+    params
```
jif-oai: This server-side transform `ExecRequest::from_sandbox_exec_request`. So remote exec server childs no longer get the env vars like `CODEX_SANDBOX_NETWORK_DISABLED` or `CODEX_SANDBOX`

codex-rs/exec-server/src/local_process.rs
```diff
@@ -91,6 +97,31 @@ struct LocalExecProcess {
...
+struct ExecServerRuntimeConfig {
+    codex_linux_sandbox_exe: Option<PathBuf>,
+}
+
+impl ExecServerRuntimeConfig {
+    fn detect() -> Self {
+        Self {
+            // The Codex CLI and codex-exec-server both dispatch the Linux
+            // sandbox helper from their own executable via argv[0].
+            codex_linux_sandbox_exe: if cfg!(target_os = "linux") {
```
jif-oai: `portable-pty` treats the arg0 value as the executable path. So this will break I think

This should be easy to confirm

codex-rs/core/src/tools/runtimes/unified_exec.rs
```diff
@@ -82,6 +83,28 @@ pub struct UnifiedExecRuntime<'a> {
...
+fn build_remote_exec_sandbox_config(
+    attempt: &SandboxAttempt<'_>,
+    additional_permissions: Option<PermissionProfile>,
+) -> SandboxLaunchConfig {
+    if matches!(attempt.sandbox, codex_sandboxing::SandboxType::None) {
+        return SandboxLaunchConfig::no_sandbox(attempt.sandbox_cwd.to_path_buf());
+    }
+
+    SandboxLaunchConfig {
+        sandbox: attempt.sandbox,
```
jif-oai: this will serialize the sandbox type of the client. But what if the exec-server is on a different OS? It can receive seatbelt for example even if the server is on Linux. How do we plan to handle this?

codex-rs/exec-server/src/local_process.rs
starr-openai: Fixed in c5942d96. I moved this child-process env preparation into the shared sandbox launch result: `SandboxExecRequest::prepare_env_for_spawn()` now injects `CODEX_SANDBOX_NETWORK_DISABLED` and, on macOS/seatbelt, `CODEX_SANDBOX`. Both the core path and exec-server server-side transform call it before spawning, so remote exec children get the same tags again.

codex-rs/exec-server/src/local_process.rs
starr-openai: Fixed in c5942d96. `codex-exec-server` now uses `codex_arg0::arg0_dispatch_or_else`, and the server runtime is constructed from `arg0_paths.codex_linux_sandbox_exe`. That gives sandbox launch a real `codex-linux-sandbox` executable/alias path, so the portable-pty path where arg0 is the executable no longer tries to execute the plain exec-server binary with only an argv0 override. I also removed the `current_exe()` fallback from `ExecServerRuntimeConfig::detect()` so library callers do not silently recreate this mismatch.

codex-rs/core/src/tools/runtimes/unified_exec.rs
starr-openai: Fixed in c5942d96. The remote request no longer serializes the client-selected concrete `SandboxType`. `SandboxLaunchConfig` now carries `sandbox_preference`; for remote unified_exec we derive `Forbid`/`Require` from the client attempt, then exec-server runs `SandboxManager::select_initial()` on the server and returns the actual server-selected sandbox in `ExecResponse`.

codex-rs/core/src/sandboxing/mod.rs
```diff
@@ -88,13 +85,14 @@ impl ExecRequest {
     }
 
     pub(crate) fn from_sandbox_exec_request(
-        request: SandboxExecRequest,
+        mut request: SandboxExecRequest,
```
starr-openai: can we remove mut?

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
```diff
@@ -835,20 +835,24 @@ impl CoreShellCommandExecutor {
             expiration: ExecExpiration::DefaultTimeout,
             capture_policy: ExecCapturePolicy::ShellTool,
         };
-        let exec_request = sandbox_manager.transform(SandboxTransformRequest {
-            command,
-            policy: sandbox_policy,
-            file_system_policy: file_system_sandbox_policy,
+        let sandbox_launch_config = SandboxLaunchConfig {
+            sandbox_preference: SandboxablePreference::from_selected_sandbox(sandbox),
+            policy: sandbox_policy.clone(),
```
starr-openai, line 833: to be sure - this sandbox policy is not materialized / dependent on the host platform (i.e. a macos orchestrator connected to a linux exec-server)?

codex-rs/core/src/tools/runtimes/unified_exec.rs
```diff
@@ -82,6 +83,24 @@ pub struct UnifiedExecRuntime<'a> {
     shell_mode: UnifiedExecShellMode,
 }
 
+fn build_remote_exec_sandbox_config(
+    attempt: &SandboxAttempt<'_>,
+    additional_permissions: Option<PermissionProfile>,
+) -> SandboxLaunchConfig {
```
starr-openai, line 89: should this conversion helper be somewhere else?

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -643,6 +646,20 @@ impl UnifiedExecProcessManager {
...
+        environment: &codex_exec_server::Environment,
+    ) -> Result<UnifiedExecProcess, UnifiedExecError> {
+        let started = environment
+            .get_exec_backend()
+            .start(params)
+            .await
+            .map_err(|err| UnifiedExecError::create_process(err.to_string()))?;
+        let sandbox_type = started.sandbox_type;
+        UnifiedExecProcess::from_remote_started(started, sandbox_type).await
+    }
```
starr-openai, line 661: why not just use `open_session_with_exec_env`?

codex-rs/exec-server/src/bin/codex-exec-server.rs
```diff
@@ -11,8 +17,75 @@ struct ExecServerArgs {
     listen: String,
 }
 
-#[tokio::main]
-async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
-    let args = ExecServerArgs::parse();
-    codex_exec_server::run_main_with_listen_url(&args.listen).await
+fn main() -> anyhow::Result<()> {
+    dispatch_linux_sandbox_arg0();
```
pakrym-oai: arg0_dispatch_or_else

codex-rs/core/src/sandboxing/mod.rs
starr-openai: Done. The request is prepared inside `SandboxManager::transform()` now, so core consumes the `SandboxExecRequest` immutably and exec-server returns the transformed request directly.

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
starr-openai: Correct: this callsite is only the local unix-escalation path. The remote `unified_exec` path sends `SandboxLaunchConfig` (policy + preference) over the exec-server protocol; the exec-server host calls `SandboxLaunchConfig::transform(...)` and selects/materializes its local platform sandbox. So a macOS orchestrator connected to a Linux exec-server sends policy, and the Linux side builds the Linux sandbox command.

codex-rs/core/src/tools/runtimes/unified_exec.rs
starr-openai: I kept it here for now because the helper is translating core-only `SandboxAttempt` into the exec-server wire `SandboxLaunchConfig`. I would not move it into `codex-sandboxing` as-is; that would pull the core runtime concept across the crate boundary. If we get a second caller, I think the right move is a small method/helper next to `SandboxAttempt` / `tools::sandboxing` and then leave this runtime file with just the call.

codex-rs/core/src/unified_exec/process_manager.rs
starr-openai: `open_session_with_exec_env` takes an already client-transformed `ExecRequest`. For the remote sandbox path we need to send raw argv/env/cwd plus `SandboxLaunchConfig` so the exec-server transforms on its own host and reports the effective `SandboxType` back. This keeps macOS-orchestrator -> Linux-exec-server sandboxing server-side. The websocket/session plumbing is still reused underneath.

codex-rs/exec-server/src/bin/codex-exec-server.rs
starr-openai: Updated the standalone entrypoint to the dispatch-or-run shape: `main()` now delegates to `linux_sandbox_arg0_dispatch_or_else(run_main)`. I kept this wrapper local instead of depending on `codex_arg0::arg0_dispatch_or_else`: `codex-arg0` currently owns the CLI aliases/apply_patch path and depends on `codex-exec-server`; using it from this binary would reintroduce that package cycle.

codex-rs/exec-server/src/server.rs
```diff
@@ -16,5 +16,12 @@ pub async fn run_main() -> Result<(), Box<dyn std::error::Error + Send + Sync>>
 pub async fn run_main_with_listen_url(
```
pakrym-oai, line 16: delete this?

codex-rs/exec-server/src/bin/codex-exec-server.rs
pakrym-oai: Stop the babysitting loop. We shouldn't copy the dispatch logic.

codex-rs/core/src/tools/runtimes/unified_exec.rs
```diff
@@ -287,11 +349,6 @@ impl<'a> ToolRuntime<UnifiedExecRequest, UnifiedExecProcess> for UnifiedExecRunt
         let exec_env = attempt
             .env_for(command, options, req.network.as_ref())
             .map_err(|err| ToolError::Codex(err.into()))?;
-        let Some(environment) = ctx.turn.environment.as_ref() else {
```
pakrym-oai, line 290: Can we avoid remote local/split? or do we need network sandbox (or something else) for  that?

pakrym-oai: Can we push the rest of features through the trait even if they don't work remotely?

codex-rs/core/src/sandboxing/mod.rs
```diff
@@ -108,16 +105,6 @@ impl ExecRequest {
             expiration,
             capture_policy,
         } = options;
-        if !network_sandbox_policy.is_enabled() {
```
pakrym-oai, line 111: wouldn't this break stuff for local sandbox code path in unified exec that doesn't go through exec-server?

codex-rs/exec-server/src/local_process.rs
```diff
@@ -102,12 +132,20 @@ impl Default for LocalProcess {
 
 impl LocalProcess {
     pub(crate) fn new(notifications: RpcNotificationSender) -> Self {
+        Self::new_with_runtime(notifications, ExecServerRuntimeConfig::detect())
```
pakrym-oai: shouln't we wire the actual path here from the consuming process?

codex-rs/exec-server/src/local_process.rs
```diff
@@ -458,6 +502,48 @@ impl ExecProcess for LocalExecProcess {
...
+            // TODO: Thread managed-network proxy state across exec-server so
+            // sandbox profile generation preserves proxy-specific allowances.
+            /*network*/
+            None,
+            runtime.codex_linux_sandbox_exe.as_deref(),
+        )
+        .map_err(|err| internal_error(format!("failed to build sandbox launch: {err}")))
+}
+
 impl LocalProcess {
```
pakrym-oai, line 536: why did this have to move out of transform?

codex-rs/exec-server/src/local_process.rs
pakrym-oai: I mean the linux_sandbox_exe so the behavior of local process is exactly the same as with normal sandboxing.

codex-rs/core/src/sandboxing/mod.rs
starr-openai: This is still covered for the local unified-exec path. `SandboxAttempt::env_for()` calls `SandboxManager::transform()`, and `SandboxManager::transform()` now calls `SandboxExecRequest::prepare_env_for_spawn()` before the core `ExecRequest` adapter sees it. So local unified exec still gets `CODEX_SANDBOX_NETWORK_DISABLED` / macOS `CODEX_SANDBOX`; the removed code was the duplicate prep in the core-only adapter. I added a short comment at that adapter in 8cf18dc to make the ownership explicit.

codex-rs/core/src/tools/runtimes/unified_exec.rs
starr-openai: I kept the split in this revision and added the explicit invariant at the branch in 8cf18dc. The remote path must send raw argv + `SandboxLaunchConfig` so the exec-server host materializes its own platform sandbox. The local fallback still transforms in core because it currently owns the session `codex_linux_sandbox_exe` path and the zsh-fork spawn lifecycle/inherited-fd path. I think the clean follow-up is to thread that runtime config into the local `Environment` and then make the default local path go through the `ExecBackend` trait too; doing that here is a broader local-spawn migration than this review cleanup.

codex-rs/exec-server/src/server.rs
starr-openai: Done in 8cf18dc. I removed the unused `run_main_with_listen_url()` wrapper/export, updated `run_main()` to call `run_main_with_runtime(DEFAULT_LISTEN_URL, ...)` directly, and fixed the README export list to point at `run_main_with_runtime()`.

codex-rs/sandboxing/src/seatbelt_tests.rs
```diff
@@ -27,9 +27,9 @@ use tempfile::TempDir;
 
 fn assert_seatbelt_denied(stderr: &[u8], path: &Path) {
     let stderr = String::from_utf8_lossy(stderr);
-    let expected = format!("bash: {}: Operation not permitted\n", path.display());
     assert!(
-        stderr == expected
+        (stderr.contains(&path.display().to_string())
+            && stderr.contains("Operation not permitted"))
```
pakrym-oai: the second contains is handled by the first one

codex-rs/sandboxing/src/manager_tests.rs
```diff
@@ -136,20 +156,18 @@ fn transform_additional_permissions_enable_network_for_external_sandbox() {
...
-            sandbox: SandboxType::None,
-            enforce_managed_network: false,
-            network: None,
-            sandbox_policy_cwd: cwd.as_path(),
-            codex_linux_sandbox_exe: None,
-            use_legacy_landlock: false,
-            windows_sandbox_level: WindowsSandboxLevel::Disabled,
-            windows_sandbox_private_desktop: false,
-        })
+            &test_launch_config(
```
pakrym-oai: nit: i'd keep the original without the extra helper.

codex-rs/exec-server/src/local_process.rs
starr-openai: Done in 66b06cc. I wired the local environment runtime from the TUI consuming process: `run_main()` now builds `EnvironmentManager::from_env_with_runtime(ExecServerRuntimeConfig::new(arg0_paths.codex_linux_sandbox_exe.clone()))`, and `Environment::create_with_runtime()` passes that runtime into the in-process `LocalProcess`. The old `new()` / `from_env()` constructors still exist, but the normal TUI path now carries the actual linux sandbox helper path.

codex-rs/exec-server/src/local_process.rs
starr-openai: Folded it back in in 66b06cc. `prepare_exec_launch()` now performs the small `ExecParams` -> `SandboxCommand` conversion inline immediately before calling `params.sandbox.transform(...)`, so the transform preparation stays in one place.

codex-rs/sandboxing/src/manager.rs
```diff
@@ -39,13 +43,66 @@ impl SandboxType {
...
+        }
+    }
+
+    pub fn transform(
+        &self,
+        command: SandboxCommand,
+        network: Option<&NetworkProxy>,
+        codex_linux_sandbox_exe: Option<&Path>,
+    ) -> Result<SandboxExecRequest, SandboxTransformError> {
+        SandboxManager::new().transform(command, self, network, codex_linux_sandbox_exe)
```
pakrym-oai: nit: is this method even used. most of .transform usages I can find are from the manager.

codex-rs/sandboxing/src/manager.rs
```diff
@@ -39,13 +43,66 @@ impl SandboxType {
...
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
 pub enum SandboxablePreference {
     Auto,
     Require,
     Forbid,
 }
 
+impl SandboxablePreference {
```
pakrym-oai: Isn't it the reverse? we select sandbox type based on preference? https://github.com/openai/codex/blob/7b6486a145ed4c02d878c971bdc8ef92eaa8ab82/codex-rs/sandboxing/src/manager.rs#L138 ?

codex-rs/sandboxing/src/manager.rs
```diff
@@ -39,13 +43,66 @@ impl SandboxType {
...
+    pub sandbox_policy_cwd: PathBuf,
+    pub additional_permissions: Option<PermissionProfile>,
+    pub enforce_managed_network: bool,
+    pub windows_sandbox_level: WindowsSandboxLevel,
+    pub windows_sandbox_private_desktop: bool,
+    pub use_legacy_landlock: bool,
+}
+
+impl SandboxLaunchConfig {
+    pub fn no_sandbox(sandbox_policy_cwd: PathBuf) -> Self {
```
pakrym-oai, line 70: why this over `Optional<SandboxLaunchConfig>`?

codex-rs/sandboxing/src/seatbelt_tests.rs
starr-openai: Done in 61500eed. Dropped the redundant Operation-not-permitted contains check; the path-bearing stderr branch is sufficient, and the sandbox_apply fallback remains.

codex-rs/sandboxing/src/manager_tests.rs
starr-openai: Done in 61500eed. Removed the test_launch_config helper and restored explicit inline SandboxLaunchConfig values at the transform call sites.

codex-rs/sandboxing/src/manager.rs
starr-openai: Done in 61500eed. Removed the SandboxLaunchConfig::transform convenience method; exec-server/local_process now calls SandboxManager::transform directly.

codex-rs/sandboxing/src/manager.rs
starr-openai: Done in 61500eed. Removed SandboxablePreference::from_selected_sandbox, switched ordinary launch paths to pass Auto and let SandboxManager select, and kept preselected retry attempts explicit via SandboxAttempt::sandbox_preference().

codex-rs/sandboxing/src/manager.rs
starr-openai: Kept this as an explicit no-sandbox config for now. The transform path still needs concrete policy/cwd/runtime fields; using Option here would add a second implicit default path rather than just the single SandboxLaunchConfig path. Happy to flip the wire field separately if you want optional/default semantics on the exec-server RPC.

pakrym-oai (commented):
My main q is whether we can maintain a single execution path like we do today (even if some features like network don't work over the remote executor).

Then there are also some non obvious changes to sandboxing logic like moving where env vars are set and how the preference is used.

starr-openai:
Responding to the latest review summary: I pushed 61500eed to make the sandboxing pieces less surprising. Normal core/exec-server launches now pass `SandboxablePreference::Auto` and let `SandboxManager::transform()` select the platform sandbox; the public `SandboxablePreference::from_selected_sandbox()` adapter is gone. The env var prep still has one owner: `SandboxExecRequest::prepare_env_for_spawn()` inside the manager transform result. The only forced preference left is internal to `SandboxAttempt`, where the orchestrator is intentionally retrying a concrete sandboxed/non-sandboxed attempt.

On the single execution path question: this PR now shares the sandbox transform path via `SandboxLaunchConfig` + `SandboxManager::transform()`, but I have not collapsed local unified-exec process spawning through exec-server in this review-fix commit. The local fallback still owns the zsh-fork/StdoutStream/inherited-fd spawn lifecycle. I think the clean next step is to move those local-spawn features behind `ExecBackend`; doing that here would be a broader local-spawn migration than the sandbox-transform cleanup.

codex-rs/exec-server/src/bin/codex-exec-server.rs
starr-openai: Done in 5d6ed706. I removed the copied dispatcher/alias code from the standalone exec-server binary. The Linux sandbox argv0 check now lives with `codex_linux_sandbox::dispatch_if_requested()`, and both `codex_arg0` and `codex-exec-server` call that shared primitive. The standalone exec-server passes its current exe as the sandbox helper path; the sandbox manager already applies the `codex-linux-sandbox` argv0 override when needed.

codex-rs/arg0/src/lib.rs
```diff
@@ -85,10 +86,9 @@ pub fn arg0_dispatch() -> Option<Arg0PathEntryGuard> {
         }
     }
 
-    if exe_name == CODEX_LINUX_SANDBOX_ARG0 {
-        // Safety: [`run_main`] never returns.
-        codex_linux_sandbox::run_main();
-    } else if exe_name == APPLY_PATCH_ARG0 || exe_name == MISSPELLED_APPLY_PATCH_ARG0 {
+    codex_linux_sandbox::dispatch_if_requested();
```
pakrym-oai, line 89: why did this have to change?

starr-openai: this doesn't strictly need to change - the main thing was to pull `codex_linux_sandbox::dispatch_if_requested()` so we can call that from the `exec-server` main, and thi sjust re-uses the same function.

pakrym-oai: why can't we use the entire arg0_dispatch_or_else as every other entrypoint does?

pakrym-oai: I think we need to split exec-server bin and lib to avoid the cicrular dependency. And use the existing pattern.

codex-rs/core/src/tools/runtimes/unified_exec.rs
```diff
@@ -219,6 +238,52 @@ impl<'a> ToolRuntime<UnifiedExecRequest, UnifiedExecProcess> for UnifiedExecRunt
         if let Some(network) = req.network.as_ref() {
             network.apply_to_env(&mut env);
         }
+        if let Some(environment) = ctx
+            .turn
+            .environment
+            .as_ref()
+            .filter(|environment| environment.exec_server_url().is_some())
```
pakrym-oai, line 245: we already have this fork in https://github.com/openai/codex/blob/7b6486a145ed4c02d878c971bdc8ef92eaa8ab82/codex-rs/core/src/unified_exec/process_manager.rs#L596

I don't think we should add another fork and another entry point to remote exec

codex-rs/core/src/tools/sandboxing.rs
```diff
@@ -332,29 +332,41 @@ pub(crate) struct SandboxAttempt<'a> {
 }
 
 impl<'a> SandboxAttempt<'a> {
+    pub(crate) fn sandbox_preference(&self) -> SandboxablePreference {
```
pakrym-oai, line 335: same comment as before, we select sandbox type based on preference, feels a bit sus to do the reverse. Are you sure having htis method makes sense?

codex-rs/exec-server/src/environment.rs
```diff
@@ -291,6 +325,9 @@ mod tests {
                 env: Default::default(),
                 tty: false,
                 arg0: None,
+                sandbox: SandboxLaunchConfig::no_sandbox(
```
pakrym-oai, line 328: do you agree with codex that special SandboxLaunchConfig is better than `Option<SandboxLaunchConfig>` pattern that we used for Fs abstraction?

codex-rs/exec-server/src/server.rs
pakrym-oai: doesn't actually look like it's removed

codex-rs/linux-sandbox/src/lib.rs
```diff
@@ -25,3 +25,19 @@ pub fn run_main() -> ! {
 pub fn run_main() -> ! {
     panic!("codex-linux-sandbox is only supported on Linux");
 }
+
+#[cfg(target_os = "linux")]
+pub fn dispatch_if_requested() {
```
pakrym-oai, line 30: we should break the circular dependency in a separte pr and keep arg0/sandboxing code as is.

codex-rs/sandboxing/src/manager.rs
```diff
@@ -86,24 +123,20 @@ pub struct SandboxExecRequest {
...
-    // TODO(viyatb): Evaluate switching this to Option<Arc<NetworkProxy>>
-    // to make shared ownership explicit across runtime/sandbox plumbing.
-    pub network: Option<&'a NetworkProxy>,
-    pub sandbox_policy_cwd: &'a Path,
-    pub codex_linux_sandbox_exe: Option<&'a Path>,
-    pub use_legacy_landlock: bool,
-    pub windows_sandbox_level: WindowsSandboxLevel,
-    pub windows_sandbox_private_desktop: bool,
+impl SandboxExecRequest {
+    pub fn prepare_env_for_spawn(&mut self) {
```
pakrym-oai, line 127: What is the reason this had to move?

codex-rs/core/src/tools/sandboxing.rs
pakrym-oai: Should SandboxAttempt store preference directly instead?

PR #16751, opened 2026-04-04
Add sandbox support to filesystem APIs

starr-openai:
## Summary
- add optional `sandboxPolicy` support to the app-server filesystem request surface
- thread sandbox-aware filesystem options through app-server and exec-server adapters
- enforce sandboxed read/write access in the filesystem abstraction with focused local and remote coverage

## Validation
- `cargo test -p codex-exec-server file_system`
- `cargo test -p codex-app-server suite::v2::fs`


codex-rs/exec-server/src/local_file_system.rs
```diff
@@ -33,10 +35,29 @@ impl ExecutorFileSystem for LocalFileSystem {
         tokio::fs::read(path.as_path()).await
     }
 
+    async fn read_file_with_options(
```
pakrym-oai: do you like with_options or should we alsways require options: &FileSystemOperationOptions and make it support :default()?

pakrym-oai: actually, I'd keep things simple and just add sandbox as an explicit parameter for now. we can add options when needed

codex-rs/exec-server/tests/file_system.rs
```diff
@@ -212,6 +227,66 @@ async fn file_system_copy_rejects_directory_without_recursive(use_remote: bool)
...
+    {
+        Ok(()) => anyhow::bail!("write should be blocked"),
+        Err(error) => error,
+    };
+    assert_eq!(error.kind(), std::io::ErrorKind::InvalidInput);
+    assert_eq!(
+        error.to_string(),
+        format!(
+            "fs/write is not permitted by sandbox policy for path {}",
+            blocked_path.display()
```
pakrym-oai: will this message ever reach the model? Does it need to be exposing fewer internal details?

codex-rs/exec-server/src/file_system.rs
```diff
@@ -18,6 +19,11 @@ pub struct CopyOptions {
     pub recursive: bool,
 }
 
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct FileSystemOperationOptions {
+    pub sandbox_policy: Option<SandboxPolicy>,
```
pakrym-oai: this doesn't seem to include     pub file_system_policy: FileSystemSandboxPolicy,
from SandboxLaunchConfig. 

I thought that is also used.

codex-rs/core/tests/suite/client.rs
```diff
@@ -165,10 +165,12 @@ impl ProviderAuthCommandFixture {
     fn new(tokens: &[&str]) -> std::io::Result<Self> {
         let tempdir = tempfile::tempdir()?;
         let tokens_file = tempdir.path().join("tokens.txt");
+        // `cmd.exe`'s `set /p` treats LF-only input as one line, so use CRLF on Windows.
```
pakrym-oai: unrelated?

codex-rs/core/tests/suite/rmcp_client.rs
```diff
@@ -496,6 +496,31 @@ async fn stdio_image_responses_are_sanitized_for_text_only_model() -> anyhow::Re
         .await;
     assert_eq!(models_mock.requests().len(), 1);
 
+    let tools_ready_deadline = Instant::now() + Duration::from_secs(30);
+    loop {
```
pakrym-oai: unrelated?

codex-rs/exec/BUILD.bazel
```diff
@@ -4,4 +4,8 @@ codex_rust_crate(
     name = "exec",
     crate_name = "codex_exec",
     test_tags = ["no-sandbox"],
+    unit_test_target_compatible_with = select({
```
pakrym-oai: unrelated?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2168,6 +2168,9 @@ pub struct FeedbackUploadResponse {
 pub struct FsReadFileParams {
     /// Absolute path to read.
     pub path: AbsolutePathBuf,
+    /// Optional sandbox policy for this filesystem read.
+    #[ts(optional = nullable)]
+    pub sandbox_policy: Option<SandboxPolicy>,
```
pakrym-oai: Are we reusing app-server types for exec server? I think we should stop.

codex-rs/app-server/tests/suite/v2/fs.rs
```diff
@@ -335,6 +359,68 @@ async fn fs_write_file_rejects_invalid_base64() -> Result<()> {
     Ok(())
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn fs_read_file_respects_sandbox_policy() -> Result<()> {
```
pakrym-oai: we should only add sandboxing to exec server fs and let app-server be (unsandboxed) for now.

pakrym-oai (changes requested):
let's keep the sandbox parameter only in exec server + remove unrelated changes + simplify overloads maybe (non-blocking, up to you)

codex-rs/exec-server/src/local_file_system.rs
```diff
@@ -178,11 +338,19 @@ fn destination_is_same_or_descendant_of_source(
     destination: &Path,
 ) -> io::Result<bool> {
     let source = std::fs::canonicalize(source)?;
-    let destination = resolve_copy_destination_path(destination)?;
+    let destination = resolve_path_for_access_check(destination, AccessPathMode::ResolveAll)?;
     Ok(destination.starts_with(&source))
 }
 
-fn resolve_copy_destination_path(path: &Path) -> io::Result<PathBuf> {
+fn resolve_path_for_access_check(path: &Path, path_mode: AccessPathMode) -> io::Result<PathBuf> {
```
jif-oai, line 402: You need to resolve symlinks first. Otherwise you'll solve `/allowed/link/../secret.txt` as `/allowed/secret.txt` (and maybe allow it) while the OS will solve it as the symlink resolution (let's say `outside/secret.txt` for example)

codex-rs/exec-server/src/local_file_system.rs
```diff
@@ -152,6 +211,107 @@ impl ExecutorFileSystem for LocalFileSystem {
...
+    path: &AbsolutePathBuf,
+    sandbox_policy: Option<&SandboxPolicy>,
+    is_allowed: fn(&FileSystemSandboxPolicy, &Path, &Path) -> bool,
+    access_kind: &str,
+    path_mode: AccessPathMode,
+) -> FileSystemResult<()> {
+    let Some(sandbox_policy) = sandbox_policy else {
+        return Ok(());
+    };
+    let cwd = std::env::current_dir()
```
jif-oai: This uses the exec-server process cwd as the sandbox cwd, but `WorkspaceWrite`/`Restricted` policies expand cwd and the fs request params do not carry the client/workspace cwd, so a remote fs request can authorize the wrong directory no?

codex-rs/exec-server/src/local_file_system.rs
starr-openai: Addressed at current head: the access checks now resolve through the existing path/symlink parents before evaluating the sandbox policy, and there is a regression test for the `link/../secret.txt` escape case.

codex-rs/exec-server/src/local_file_system.rs
starr-openai: Addressed at current head as well: remote FS params now carry the client-side sandbox cwd, and the exec-server enforces against that explicit cwd instead of the server process cwd.

pakrym-oai (approved):

PR #16753, opened 2026-04-04
[codex] add responses proxy JSON dumps

tibo-openai:
This makes Responses API proxy request/response dumping first-class by adding an optional `--dump-dir` flag that emits paired JSON files with shared sequence/timestamp prefixes, captures full request and response headers and records parsed JSON bodies.


andmis (approved):
Thanks!

PR #16764, opened 2026-04-04
app-server: centralize AuthManager initialization

euroelessar:
Extract a shared helper that builds AuthManager from Config and applies the forced ChatGPT workspace override in one place.

Create the shared AuthManager at MessageProcessor call sites so that upcoming new transport's initialization can reuse the same handle, and keep only external auth refresher wiring inside `MessageProcessor`.

Remove the now-unused `AuthManager::shared_with_external_auth` helper.


codex-rs/app-server/src/auth_manager.rs
```diff
@@ -0,0 +1,17 @@
+use std::sync::Arc;
+
+use codex_core::config::Config;
+use codex_login::AuthManager;
+
+pub(crate) fn auth_manager_from_config(
+    config: &Config,
+    enable_codex_api_key_env: bool,
+) -> Arc<AuthManager> {
```
owenlin0: this doesn't seem specific to app-server, is there a better home for it? could we do something along these lines? 

```
// codex-config or codex-login
pub struct AuthManagerConfig {
    pub codex_home: PathBuf,
    pub credentials_store_mode: AuthCredentialsStoreMode,
    pub forced_chatgpt_workspace_id: Option<String>,
}

// codex-login
impl AuthManager {
    pub fn shared_from_config(
        config: &AuthManagerConfig,
        enable_codex_api_key_env: bool,
    ) -> Arc<Self> {
        let auth_manager = Self::shared(
            config.codex_home.clone(),
            enable_codex_api_key_env,
            config.credentials_store_mode,
        );
        auth_manager.set_forced_chatgpt_workspace_id(
            config.forced_chatgpt_workspace_id.clone(),
        );
        auth_manager
    }
}
```

a shared helper will help since apparently we have this in a bunch of places:
> - core/src/connectors.rs (line 495) constructs an AuthManager but does not set forced_chatgpt_workspace_id.
> - core/src/prompt_debug.rs (line 26) does set it.
> - mcp-server also constructs one inline in mcp-server/src/message_processor.rs.

owenlin0 (approved):
pre-approving

PR #16805, opened 2026-04-04
1/4 Replace realtime websocket transport with WebRTC

aibrahim-oai:
Stack position: 1/4

Stack:
- 1/4 #16805 Replace realtime websocket transport with WebRTC (this PR)
- 2/4 #16806 Route TUI realtime audio through shared echo cancellation
- 3/4 #16807 Remove local realtime playback interruption
- 4/4 #16769 Support ChatGPT realtime call auth and request shape

This PR:
- Move the model-side realtime transport from websocket to WebRTC.
- Preserve the core-facing session/event API so UI changes can stack separately.
- Drop the websocket e2e test that no longer matches the transport implementation.


aibrahim-oai:
@codex review

PR #16810, opened 2026-04-04
(tui): Decode percent-escaped bare local file links

etraut-openai:
Addresses #16622

Problem: bare local file links in TUI markdown render percent-encoded path bytes literally, unlike file:// links.

Solution: decode bare path targets before local-path expansion and add regression coverage for spaces and Unicode.


fcoury-oai (approved):
Tested before: 

Image

And after:

Image

Code looks good as well. Approved!

PR #16813, opened 2026-04-04
Annotate skill doc reads with skill names

etraut-openai:
Addresses #16303

Problem: Skill doc reads render as plain `Read SKILL.md`, so the TUI hides which skill was opened.

Solution: Best-effort annotate exact `SKILL.md` reads with the matching loaded skill name from `skills_all` before rendering exec cells.

Before:
```
• Explored
  └ Read SKILL.md
```

After:
```
• Explored
  └ Read SKILL.md (pr-babysitter skill)
```


fcoury-oai (approved):
This is a great quality of life improvement. Tested the feature and code looks good.

Approved.

PR #16822, opened 2026-04-05
Fix resume picker timestamp labels and stability

etraut-openai:
Problem: The resume picker used awkward "Created at" and "Updated at" headers, and its relative timestamps changed while navigating because they were recomputed on each redraw.

Solution: Rename the headers to "Created" and "Updated", and anchor relative timestamp formatting to the picker load time so the displayed ages stay stable while browsing.


fcoury-oai (approved):
Replicated the before and after behavior. Code looks good.

Approved.

Issue #7646, opened 2025-12-05
Spinner on first run of codex does not let copy link


Komzpa:
### What version of Codex is running?

codex-cli 0.65.0

### What subscription do you have?

Plus

### Which model were you using?

any

### What platform is your computer?

Linux 6.17.0-7-generic x86_64 unknown

### What issue are you seeing?

When I ssh to machine and run `codex` it runs with a nice animation. However browser doesn't auto-open (it's ssh) and link is long and not clickable in terminal (because it spans two lines). Copying the link is impossible because every frame of animation resets selection.

### What steps can reproduce the bug?

When I ssh to machine and run `codex` it runs with a nice animation. However browser doesn't auto-open (it's ssh) and link is long and not clickable in terminal (because it spans two lines). Copying the link is impossible because every frame of animation resets selection.

Bronx83:
Thank you for this!
I created a PR to stop the animation when the authentication URL is present in the cli https://github.com/openai/codex/pull/7872

Bronx83:
Hi @etraut-openai , what do you think will be the best solution for this case? We can consider adding a '--no-animation' flag to the cli.

mdrzn:
+1 to this. Insanely frustrating to waste 30 seconds trying to copy a URL since it keeps "refreshing" with the animation.

Komzpa:
At least a hint that one can run "codex login" to not have the spinner would help.

etraut-openai:
We recently added preview support for [device code authentication](https://github.com/openai/codex/issues/2798#issuecomment-3629482218). It's designed to address headless logins or situations involving ssh. Please give it a try and let us know if this meets your needs.

Neiz-Kap:
The issue still exists. As a sample, the same screen on Claude doesn't have such animation, which re-renders the screen each second, and it doesn't cancel url selection

etraut-openai:
This will be addressed in the next release.

PR #16827, opened 2026-04-05
tui: route device-code auth through app server

etraut-openai:
Addresses #7646
Also enables device code auth for remote TUI sessions

Problem: TUI onboarding handled device-code login directly rather than using the recently-added app server support for device auth. Also, auth screens kept animating while users needed to copy login details.

Solution: Route device-code onboarding through app-server login APIs and make the auth screens static while those copy-oriented flows are visible.


codex-rs/tui/src/onboarding/auth/headless_chatgpt_login.rs
```diff
@@ -1,122 +1,76 @@
...
+            })
+            .await
+        {
+            Ok(LoginAccountResponse::ChatgptDeviceCode {
+                login_id,
+                verification_url,
+                user_code,
+            }) => {
+                *error.write().unwrap() = None;
+                let _updated = set_device_code_state_for_active_attempt(
```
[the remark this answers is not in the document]
etraut-openai: Fixed in `71f5cb6166`. We now clear the auth error only if `set_device_code_state_for_active_attempt(...)` actually updates the active request, so a stale success can’t wipe out the current error state anymore.

codex-rs/tui/src/onboarding/onboarding_screen.rs
```diff
@@ -175,6 +171,13 @@ impl OnboardingScreen {
         out
     }
 
+    fn should_suppress_animations(&self) -> bool {
```
etraut-openai, line 174: It would be good to add a comment in this function to explain why we're suppressing animations in certain cases.

codex-rs/tui/src/onboarding/onboarding_screen.rs
```diff
@@ -527,3 +539,126 @@ pub(crate) async fn run_onboarding_app(
...
+            forced_login_method: None,
+            animations_enabled: true,
+            animations_suppressed: std::cell::Cell::new(false),
+        };
+
+        (widget, codex_home)
+    }
+
+    #[test]
+    fn onboarding_suppresses_welcome_when_copyable_auth_is_visible() {
```
etraut-openai: I don't think this test is necessary. Remove it.

codex-rs/tui/src/onboarding/onboarding_screen.rs
```diff
@@ -527,3 +539,126 @@ pub(crate) async fn run_onboarding_app(
...
+        let area = Rect::new(0, 0, 80, 40);
+        let mut buf = Buffer::empty(area);
+        (&screen).render_ref(area, &mut buf);
+
+        assert_eq!(screen.should_suppress_animations(), true);
+        assert!(welcome_frames.try_recv().is_err());
+    }
+
+    #[test]
+    fn onboarding_keeps_welcome_animation_enabled_in_pick_mode() {
```
etraut-openai: I don't think this test is needed. Remove it.

codex-rs/tui/src/tui/frame_requester.rs
```diff
@@ -65,6 +65,17 @@ impl FrameRequester {
             frame_schedule_tx: tx,
         }
     }
+
+    /// Create a requester that exposes scheduled frame deadlines for assertions.
+    pub(crate) fn test_spy() -> (Self, mpsc::UnboundedReceiver<Instant>) {
```
etraut-openai: With the two tests removed, this test_spy function can also be removed.

codex-rs/tui/src/onboarding/onboarding_screen.rs
etraut-openai: Added a short comment in `64c2dc5f58` explaining that we freeze the onboarding screen once auth is showing copyable login material so redraws don’t fight terminal selection.

codex-rs/tui/src/onboarding/onboarding_screen.rs
etraut-openai: Removed that test in `64c2dc5f58`. The behavior is already covered lower in the stack, so this one was just extra noise.

codex-rs/tui/src/onboarding/onboarding_screen.rs
etraut-openai: Yep, removed in `64c2dc5f58`.

codex-rs/tui/src/tui/frame_requester.rs
etraut-openai: I checked before removing it: `FrameRequester::test_spy()` is still used by the auth rendering tests in `auth.rs` and `headless_chatgpt_login.rs`, so there isn’t dead code to delete here right now.

codex-rs/tui/src/onboarding/auth.rs
```diff
@@ -1085,6 +1147,75 @@ mod tests {
         assert_eq!(found, url, "OSC 8 hyperlink should cover the full URL");
     }
 
+    #[test]
+    fn continue_in_browser_skips_shimmer_when_animations_are_suppressed() {
```
etraut-openai: I think you can remove this test too. Along with test_spy.

etraut-openai: Yep. Removed that test, the device-code counterpart, and `FrameRequester::test_spy()` in `2c895c16c9`.

codex-rs/tui/src/tui/frame_requester.rs
etraut-openai: Followed up in `2c895c16c9`: once the matching auth-side test was removed too, `FrameRequester::test_spy()` was actually dead, so I removed it.

daniel-oai (approved):

daniel-oai:
Tested this locally and it worked great. I also had Codex do a careful review pass, and it found one P2 issue that might be worth adding a small fix for before merging.

Quoting the Codex review note separately:

> I think this should also cancel the app-server login if the start response comes back after the TUI attempt is no longer active.
>
> Right now, if the user presses Esc while `account/login/start` is still in flight, the TUI moves back to `PickMode`. When the app server later returns `ChatgptDeviceCode { login_id, ... }`, `set_device_code_state_for_active_attempt(...)` returns `false`, but we drop the `login_id` without sending `account/login/cancel`. At that point the app server has already started polling and the stale attempt can keep running until timeout or until another login replaces it.
>
> Suggested shape:
>
> ```rust
> let updated = set_device_code_state_for_active_attempt(
>     &sign_in_state,
>     &request_frame,
>     &request_id,
>     ContinueWithDeviceCodeState::ready(
>         request_id.clone(),
>         login_id.clone(),
>         verification_url,
>         user_code,
>     ),
> );
> if updated {
>     *error.write().unwrap() = None;
> } else {
>     cancel_login_attempt(&request_handle, login_id).await;
> }
> ```
>
> That helper could mirror the existing `cancel_active_attempt` request path by sending `ClientRequest::CancelLoginAccount { params: CancelLoginAccountParams { login_id }, ... }`. This keeps ownership aligned with the new app-server device-code flow: once the app server creates a login attempt, the TUI should either adopt it or explicitly cancel it.

PR #16831, opened 2026-04-05
Speed up /mcp inventory listing

etraut-openai:
Addresses #16244

This was a performance regression introduced when we moved the TUI on top of the app server API.

Problem: `/mcp` rebuilt a full MCP inventory through `mcpServerStatus/list`, including resources and resource templates that made the TUI wait on slow inventory probes.

Solution: add a lightweight `detail` mode to `mcpServerStatus/list`, have `/mcp` request tools-and-auth only, and cover the fast path with app-server and TUI tests.

Testing: Confirmed slow (multi-second) response prior to change and immediate response after change.

I considered two options:
1. Change the existing `mcpServerStatus/list` API to accept an optional "details" parameter so callers can request only a subset of the information.
2. Add a separate `mcpServer/list` API that returns only the servers, tools, and auth but omits the resources.

I chose option 1, but option 2 is also a reasonable approach.


fcoury-oai (approved):
Tested the performance difference between main and this branch.

Approved.

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1947,6 +1947,17 @@ pub struct ListMcpServerStatusParams {
     /// Optional page size; defaults to a server-defined value.
     #[ts(optional = nullable)]
     pub limit: Option<u32>,
+    /// Controls how much MCP inventory data to fetch for each server.
```
owenlin0, line 1952: nit: worth documenting the default if not specified?

owenlin0 (approved):
API design lgtm

Issue #16832, opened 2026-04-05
Fast mode toggle doesn’t take effect after switching off


qxcnm:
### What version of Codex CLI is running?

0.1180.

### What subscription do you have?

plus

### Which model were you using?

gpt-5.4 high 

### What platform is your computer?

windows

### What terminal emulator and version are you using (if applicable)?

powershell

### What issue are you seeing?

设置fast模式之后 如果/fast 关闭fast 无效，必须得重启才行 
监测请求发现如果一开始设置了关闭fast 打开codex cli 是正常的，一旦切换到 fast on 之后 在切换回去 请求头中依然带有service_tier: "priority" 参数，导致fast不能热生效必须重启

### What steps can reproduce the bug?

启动 codex cli 之前 设置  # service_tier = "fast" 为不开启
启动之后 /fast 设置  fast on
请求一次后 再次设置 /fast off 
请求依然带有 service_tier: "priority"

### What is the expected behavior?

预期是解决  /fast  on 切换 off 不生效的问题

etraut-openai:
English translation:

After enabling fast mode, running /fast to turn it off does not work.
The change only takes effect after restarting the CLI.

From inspecting the requests:

If fast mode is initially disabled before launching Codex CLI, everything works correctly.
However, once fast mode is turned on, then turned off, the request headers still include
service_tier: "priority".
As a result, fast mode cannot be disabled dynamically and requires a restart to take effect.
Steps to Reproduce

Before starting Codex CLI, ensure:

service_tier = "fast"

is disabled

Launch Codex CLI
Run /fast to enable fast mode
Make a request
Run /fast off to disable fast mode

Observe that subsequent requests still include:

service_tier: "priority"

qxcnm:
好的，非常感谢你的贡献。

PR #16833, opened 2026-04-05
Fix TUI fast mode toggle regression

etraut-openai:
Addresses #16832

Problem: After `/fast on`, the TUI omitted an explicit service-tier clear on later turns, so `/fast off` left app-server sessions stuck on `priority` until restart.

Solution: Always submit the current service tier with user turns, including an explicit clear when Fast mode is off, and add a regression test for the `/fast on` -> `/fast off` flow.


fcoury-oai (approved):
Brought the new test case to main and it failed as expected.

Approved.

PR #16875, opened 2026-04-06
Support anyOf and enum in JsonSchema

vivi:
This brings us into better alignment with the JSON schema subset that is supported in [https://developers.openai.com/api/docs/guides/structured-outputs#supported-schemas], and also allows us to render richer function signatures in code mode (e.g., anyOf{null, OtherObjectType})

### DIFF
--- a/codex-rs/core/src/tools/spec_tests.rs
+++ b/codex-rs/core/src/tools/spec_tests.rs
@@ -832,16 +832,15 @@ fn test_mcp_tool_property_missing_type_defaults_to_string() {
         tool.spec,
         ToolSpec::Function(ResponsesApiTool {
             name: "dash/search".to_string(),
-            parameters: JsonSchema::Object {
-                properties: BTreeMap::from([(
+            parameters: JsonSchema::object(
+                /*properties*/
+                BTreeMap::from([(
                     "query".to_string(),
-                    JsonSchema::String {
-                        description: Some("search query".to_string())
-                    }
+                    JsonSchema::string(Some("search query".to_string())),
                 )]),
-                required: None,
-                additional_properties: None,
-            },
+                /*required*/ None,
+                /*additional_properties*/ None
+            ),
             description: "Search docs".to_string(),
             strict: false,
             output_schema: Some(mcp_call_tool_result_output_schema(serde_json::json!({}))),
@@ -890,14 +889,12 @@ fn test_mcp_tool_integer_normalized_to_number() {
         tool.spec,
         ToolSpec::Function(ResponsesApiTool {
             name: "dash/paginate".to_string(),
-            parameters: JsonSchema::Object {
-                properties: BTreeMap::from([(
-                    "page".to_string(),
-                    JsonSchema::Number { description: None }
-                )]),
-                required: None,
-                additional_properties: None,
-            },
+            parameters: JsonSchema::object(
+                /*properties*/
+                BTreeMap::from([("page".to_string(), JsonSchema::number(/*description*/ None),)]),
+                /*required*/ None,
+                /*additional_properties*/ None
+            ),
             description: "Pagination".to_string(),
             strict: false,
             output_schema: Some(mcp_call_tool_result_output_schema(serde_json::json!({}))),
@@ -947,17 +944,18 @@ fn test_mcp_tool_array_without_items_gets_default_string_items() {
         tool.spec,
         ToolSpec::Function(ResponsesApiTool {
             name: "dash/tags".to_string(),
-            parameters: JsonSchema::Object {
-                properties: BTreeMap::from([(
+            parameters: JsonSchema::object(
+                /*properties*/
+                BTreeMap::from([(
                     "tags".to_string(),
-                    JsonSchema::Array {
-                        items: Box::new(JsonSchema::String { description: None }),
-                        description: None
-                    }
+                    JsonSchema::array(
+                        JsonSchema::string(/*description*/ None),
+                        /*description*/ None,
+                    ),
                 )]),
-                required: None,
-                additional_properties: None,
-            },
+                /*required*/ None,
+                /*additional_properties*/ None
+            ),
             description: "Tags".to_string(),
             strict: false,
             output_schema: Some(mcp_call_tool_result_output_schema(serde_json::json!({}))),
@@ -1008,14 +1006,21 @@ fn test_mcp_tool_anyof_defaults_to_string() {
         tool.spec,
         ToolSpec::Function(ResponsesApiTool {
             name: "dash/value".to_string(),
-            parameters: JsonSchema::Object {
-                properties: BTreeMap::from([(
+            parameters: JsonSchema::object(
+                /*properties*/
+                BTreeMap::from([(
                     "value".to_string(),
-                    JsonSchema::String { description: None }
+                    JsonSchema::any_of(
+                        vec![
+                            JsonSchema::string(/*description*/ None),
+                            JsonSchema::number(/*description*/ None),
+                        ],
+                        /*description*/ None,
+                    ),
                 )]),
-                required: None,
-                additional_properties: None,
-            },
+                /*required*/ None,
+                /*additional_properties*/ None
+            ),
             description: "AnyOf Value".to_string(),
             strict: false,
             output_schema: Some(mcp_call_tool_result_output_schema(serde_json::json!({}))),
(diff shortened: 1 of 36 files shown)


codex-rs/tools/src/json_schema.rs
```diff
@@ -163,14 +183,292 @@ fn sanitize_json_schema(value: &mut JsonValue) {
...
+) -> Result<AdditionalProperties, serde_json::Error> {
+    match value {
+        JsonValue::Bool(flag) => Ok(AdditionalProperties::Boolean(*flag)),
+        _ => Ok(AdditionalProperties::Schema(Box::new(parse_json_schema(
+            value,
+        )?))),
+    }
+}
+
+fn json_schema_to_json(schema: &JsonSchema) -> JsonValue {
```
pakrym-oai: Why do we need custom parse/serialize methods? Can't JSON schema be represented in Rust types for default serializer/deserializer still work?

codex-rs/tools/src/tool_registry_plan_tests.rs
```diff
@@ -1543,6 +1543,93 @@ fn code_mode_augments_mcp_tool_descriptions_with_namespaced_sample() {
...
+        &[],
+    );
+
+    let ToolSpec::Function(ResponsesApiTool { description, .. }) =
+        &find_tool(&tools, "mcp__sample__fn").spec
+    else {
+        panic!("expected function tool");
+    };
+
+    assert!(description.contains("mcp__sample__fn(args: { open?: Array<{"));
```
pakrym-oai: can we use the full multiline string in contains assertion?

codex-rs/tools/src/json_schema_tests.rs
```diff
@@ -96,3 +102,157 @@ fn parse_tool_input_schema_sanitizes_additional_properties_schema() {
         }
     );
 }
+
+#[test]
+fn parse_tool_input_schema_preserves_web_run_shape() {
```
pakrym-oai: Can this test be reduced to only representative parts?

pakrym-oai: As in, specific types or patterns instead of just `web_run`

codex-rs/tools/src/json_schema_tests.rs
```diff
@@ -87,7 +87,13 @@ fn parse_tool_input_schema_sanitizes_additional_properties_schema() {
                 JsonSchema::Object {
                     properties: BTreeMap::from([(
                         "value".to_string(),
-                        JsonSchema::String { description: None },
+                        JsonSchema::AnyOf {
```
pakrym-oai: Does this type anyOf definition work with response api?

pakrym-oai (changes requested):
2 major points:
1. Does the newly extended JSON schema format work with Responses API https://developers.openai.com/api/docs/guides/structured-outputs
2. Can we drop custom serializer/deserializer.

codex-rs/tools/src/json_schema_tests.rs
vivi: yes, anyOf and Enum are supported: https://developers.openai.com/api/docs/guides/structured-outputs#supported-schemas

i will remove allOf, oneOf, and Constant though

vivi:
@codex review

codex-rs/tools/src/json_schema.rs
```diff
@@ -171,6 +274,97 @@ fn sanitize_json_schema(value: &mut JsonValue) {
...
+        JsonValue::String(_) => Some(JsonSchemaPrimitiveType::String),
+        JsonValue::Number(_) => Some(JsonSchemaPrimitiveType::Number),
+        JsonValue::Bool(_) => Some(JsonSchemaPrimitiveType::Boolean),
+        JsonValue::Null => Some(JsonSchemaPrimitiveType::Null),
+        _ => None,
+    }
+}
+
+fn normalize_schema_type_name(schema_type: &str) -> Option<&'static str> {
+    match schema_type {
```
pakrym-oai, line 307: nit: this can be simpler

codex-rs/tools/src/json_schema.rs
```diff
@@ -96,31 +188,47 @@ fn sanitize_json_schema(value: &mut JsonValue) {
...
+                sanitize_json_schema(additional_properties);
+            }
+            if let Some(value) = map.get_mut("prefixItems") {
+                sanitize_json_schema(value);
+            }
+            if let Some(value) = map.get_mut("anyOf") {
+                sanitize_json_schema(value);
+            }
+
+            if let Some(const_value) = map.remove("const") {
```
pakrym-oai, line 203: is this new logic? do we need a test?

codex-rs/tools/src/json_schema.rs
```diff
@@ -96,31 +188,47 @@ fn sanitize_json_schema(value: &mut JsonValue) {
...
-                        && matches!(
-                            candidate_type,
-                            "object" | "array" | "string" | "number" | "integer" | "boolean"
-                        )
-                    {
-                        schema_type = Some(candidate_type.to_string());
-                        break;
-                    }
-                }
+            if matches!(schema_type.as_deref(), Some("enum") | Some("const")) {
```
pakrym-oai: why would the type be enum or const? I don't think the spec supports it.

codex-rs/tools/src/json_schema.rs
```diff
@@ -96,31 +188,47 @@ fn sanitize_json_schema(value: &mut JsonValue) {
...
+                schema_type = None;
+                map.remove("type");
+            }
+
+            if let Some(types) = map.get("type").and_then(JsonValue::as_array).cloned() {
```
pakrym-oai: we are interchangeably using map.get("type")  and schema_type around this method. 

can we normalize and clean up a bit?

codex-rs/tools/src/json_schema.rs
```diff
@@ -131,36 +239,31 @@ fn sanitize_json_schema(value: &mut JsonValue) {
                     schema_type = Some("object".to_string());
                 } else if map.contains_key("items") || map.contains_key("prefixItems") {
                     schema_type = Some("array".to_string());
-                } else if map.contains_key("enum")
-                    || map.contains_key("const")
-                    || map.contains_key("format")
-                {
-                    schema_type = Some("string".to_string());
+                } else if map.contains_key("enum") || map.contains_key("format") {
+                    schema_type = infer_enum_type(map.get("enum").and_then(JsonValue::as_array))
```
pakrym-oai: do we need this extra logic? do we have tests?

codex-rs/tools/src/json_schema.rs
pakrym-oai: is_known_type?

codex-rs/tools/src/json_schema.rs
```diff
@@ -171,6 +274,97 @@ fn sanitize_json_schema(value: &mut JsonValue) {
...
+                map.remove("type");
+            }
+            [single] => {
+                map.insert("type".to_string(), single.clone());
+            }
+            _ => {
+                map.insert("type".to_string(), JsonValue::Array(normalized));
+            }
+        }
+    } else if let Some(schema_type) = map.get("type").and_then(JsonValue::as_str)
```
pakrym-oai: can we always read type into the array (and turn singular type into a single item vec) for processing

That will avoid special casing multiple/singular cases everywhere.

codex-rs/tools/src/json_schema.rs
```diff
@@ -4,40 +4,132 @@ use serde_json::Value as JsonValue;
...
+        Self::typed(JsonSchemaPrimitiveType::Integer, description)
+    }
+
+    pub fn null(description: Option<String>) -> Self {
+        Self::typed(JsonSchemaPrimitiveType::Null, description)
+    }
+
+    pub fn enumeration(values: Vec<JsonValue>, description: Option<String>) -> Self {
+        Self {
+            schema_type: infer_enum_type(Some(&values)).map(JsonSchemaType::Single),
```
pakrym-oai: why can't the caller provide the type?

codex-rs/core/src/tools/spec_tests.rs
```diff
@@ -891,7 +891,10 @@ fn test_mcp_tool_integer_normalized_to_number() {
             name: "dash/paginate".to_string(),
             parameters: JsonSchema::object(
                 /*properties*/
-                BTreeMap::from([("page".to_string(), JsonSchema::number(/*description*/ None),)]),
+                BTreeMap::from([(
+                    "page".to_string(),
+                    JsonSchema::integer(/*description*/ None),
```
vivi, line 896: from https://developers.openai.com/api/docs/guides/structured-outputs#supported-schemas, integer and number are both supported

codex-rs/tools/src/json_schema.rs
vivi: added `parse_tool_input_schema_rewrites_const_to_single_value_enum`

codex-rs/tools/src/json_schema.rs
```diff
@@ -96,81 +188,168 @@ fn sanitize_json_schema(value: &mut JsonValue) {
...
+                && !map.contains_key("properties")
+            {
+                map.insert(
+                    "properties".to_string(),
+                    JsonValue::Object(serde_json::Map::new()),
+                );
             }
 
-            if schema_type == "array" && !map.contains_key("items") {
+            if matches!(schema_types.as_slice(), [JsonSchemaPrimitiveType::Array])
```
pakrym-oai: looks like this is handled in ensure_default_children_for_schema_types

codex-rs/tools/src/json_schema.rs
```diff
@@ -64,16 +149,23 @@ impl From<JsonSchema> for AdditionalProperties {
 pub fn parse_tool_input_schema(input_schema: &JsonValue) -> Result<JsonSchema, serde_json::Error> {
     let mut input_schema = input_schema.clone();
     sanitize_json_schema(&mut input_schema);
-    serde_json::from_value::<JsonSchema>(input_schema)
+    let schema: JsonSchema = serde_json::from_value(input_schema)?;
+    if matches!(
```
pakrym-oai, line 153: nit: this is pretty specific, is it important to have the parser validate? we don't do it for other issues.

pakrym-oai (approved):

codex-rs/tools/src/json_schema.rs
vivi: I added this specific case because the structured output supported subset of the json schema only accepts a null type when it's in a list with another valid type (previous JsonSchema didn't have null as a type so would throw a deserialization error on the {"type": "null"} case as well...)

(and as a result this [existing test](https://github.com/openai/codex/blob/3c1adbabcd0921f848f8c6122e6229689173d063/codex-rs/app-server/src/codex_message_processor.rs#L9117-L9127) expects an error)

PR #16878, opened 2026-04-06
[codex] Add Warp OSC 9 notification support

jgershen-oai:
## Summary
- Detect Warp sessions via `WARP_CLI_AGENT_PROTOCOL_VERSION` and select the OSC 9 notification backend when notification method is `auto`.
- Let Warp receive notification updates even when Codex believes the terminal is focused, since Warp handles notification visibility itself and does not relay focus events to the PTY.
- Port the behavior onto the current `codex-rs/tui` layout without reintroducing the removed `tui_app_server` tree.

## Context
Warp supports OSC 9 notifications and exposes its CLI agent protocol environment variable in supported sessions. Codex already used OSC 9 for terminals like WezTerm/Ghostty; this extends that path to Warp and avoids suppressing Warp notifications based on focus state.

## Testing
- `just fmt`
- `cargo test -p codex-tui`
- `git diff --check`


codex-rs/tui/src/tui.rs
```diff
@@ -364,7 +365,9 @@ impl Tui {
     /// Emit a desktop notification now if the terminal is unfocused.
     /// Returns true if a notification was posted.
     pub fn notify(&mut self, message: impl AsRef<str>) -> bool {
-        if self.terminal_focused.load(Ordering::Relaxed) {
+        // Warp handles notification visibility itself but does not relay focus
+        // events to the PTY, so Codex should not suppress its notifications here.
+        if self.terminal_focused.load(Ordering::Relaxed) && !warp_cli_agent_protocol_detected() {
```
etraut-openai, line 373: I'm not wild about having such specific workarounds for individual terminals. That's not very maintainable. Could we work with the maintainers of warp to get this fixed rather than hacking around this limitation in the TUI?

etraut-openai: If that's not feasible, then we should at least abstract this out to a helper function called `supports_focus()` so we can avoid sprinkling terminal-specific heuristics throughout the code.

etraut-openai (changes requested):

PR #16879, opened 2026-04-06
Render namespace description for tools

vivi:


magus (approved):
Would be interesting to explore rendering the entire MCP schema json directly to avoid gaps in render (we are also losing output typing it's opaque at the moment unfortunately)

PR #16880, opened 2026-04-06
Render function attribute descriptions

vivi:


codex-rs/code-mode/src/description.rs
```diff
@@ -460,6 +487,67 @@ fn render_json_schema_object(map: &serde_json::Map<String, JsonValue>) -> String
     format!("{{ {} }}", lines.join(" "))
 }
 
+fn render_json_schema_object_with_property_descriptions(
```
pakrym-oai: why do we need to fork the object rendering logic instead of adding descriptions to the main one?

codex-rs/code-mode/src/description.rs
pakrym-oai, line 277: Do we not want descriptions in outputs?

pakrym-oai (changes requested):
Can we avoid forking the rendering logic?

codex-rs/code-mode/src/description.rs
vivi: yes, good catch

pakrym-oai (approved):

Issue #15527, opened 2026-03-23
Nested codex exec commands inherit parent CODEX_THREAD_ID instead of nested thread id


aquiles-ai123:
# Nested `codex exec` commands inherit the parent `CODEX_THREAD_ID` instead of the nested thread id

## Summary

When `codex exec` is launched from inside an existing Codex session, the nested session gets its own `thread.started.thread_id`, but commands executed by that nested session still see the **parent** `CODEX_THREAD_ID` in their environment.

This looks like environment leakage from the outer Codex process into nested Codex command executions.

## Why this matters

Any script, hook, or command that relies on `CODEX_THREAD_ID` for correlation, logging, or per-thread state will attribute work to the wrong thread.

## Environment

- Codex CLI: `codex-cli 0.116.0`
- Outer shell environment includes:

```text
CODEX_THREAD_ID=019d1afb-b2c7-72d3-b240-f3ee9d31b2c9
```

## Repro

This repro is intended to be run from inside an existing Codex session.

### Read-only example

```bash
codex exec --ephemeral --skip-git-repo-check --color never --json   -s read-only -C /home/cryptotrading74   'Run env | sort | rg "^CODEX_THREAD_ID=" and stop.'   > /tmp/codex_threadid_mismatch.jsonl
```

Observed transcript:

```json
{"type":"thread.started","thread_id":"019d1be2-5752-7553-80c4-66e6dbddb174"}
{"type":"item.completed","item":{"type":"command_execution","aggregated_output":"CODEX_THREAD_ID=019d1afb-b2c7-72d3-b240-f3ee9d31b2c9\n","exit_code":0,"status":"completed"}}
```

### Workspace-write example

```bash
codex exec --ephemeral --skip-git-repo-check --color never --json   -s workspace-write -C /home/cryptotrading74   'Run env | sort | rg "^CODEX_THREAD_ID=" and stop.'   > /tmp/codex_threadid_mismatch_ws.jsonl
```

Observed transcript:

```json
{"type":"thread.started","thread_id":"019d1be2-fe81-74b3-8809-3cc1f9c8f3a8"}
{"type":"item.completed","item":{"type":"command_execution","aggregated_output":"CODEX_THREAD_ID=019d1afb-b2c7-72d3-b240-f3ee9d31b2c9\n","exit_code":0,"status":"completed"}}
```

## Expected behavior

[...]

PR #16882, opened 2026-04-06
Fix nested exec thread ID restore

etraut-openai:
Addresses #15527

Problem: Nested `codex exec` commands could source a shell snapshot that re-exported the parent `CODEX_THREAD_ID`, so commands inside the nested session were attributed to the wrong thread.

Solution: Reapply the live command env's `CODEX_THREAD_ID` after sourcing the snapshot.


jif-oai (approved):

jif-oai:
@codex review

Issue #16443, opened 2026-04-01
Non-interactive mode doesn't properly close HTTP connections on exit


KeyneZzz:
### What version of Codex CLI is running?

0.118.0

### What subscription do you have?

Free

### What issue are you seeing?

Codex CLI in non-interactive mode calls `std::process::exit()` throughout error paths, terminating the process without properly closing HTTP connections. This causes session leaks on proxy servers that track concurrent connections.

### What steps can reproduce the bug?

# 1. Run a simple non-interactive command
codex "echo hello"

# 2. Check proxy server active sessions immediately
# Expected: Session count decreases
# Actual: Session count remains elevated (5 min TTL)

### Additional information

Found **13 instances** of `std::process::exit(1)` in `codex-rs/exec/src/lib.rs`:

```rust
// Line 524: Git check fails
if get_git_repo_root(&default_cwd).is_none() {
    std::process::exit(1); // ❌ No cleanup
}

// Line 828: Error exit
if error_seen {
    std::process::exit(1); // ❌ Bypasses Drop
}

// ... 11 more instances
```

**Problem**: `exit()` bypasses `Drop` traits → HTTP connections close abruptly → server can't detect session end → session count leaks until TTL expires.

etraut-openai:
Thanks for the bug report and analysis. This will be addressed in the next release.

PR #16890, opened 2026-04-06
Validate exec input before starting app-server

etraut-openai:
Addresses #16443

This was a regression introduced when we moved exec on top of the app server APIs.

Problem: codex exec resolved prompt/stdin and output schema after starting the in-process app-server, so early `process::exit(1)` paths could bypass session shutdown.

Solution: Resolve prompt/stdin and output schema before app-server startup so validation failures happen before any exec session is created.

### DIFF
--- a/codex-rs/exec/tests/suite/prompt_stdin.rs
+++ b/codex-rs/exec/tests/suite/prompt_stdin.rs
@@ -4,6 +4,21 @@
 use core_test_support::responses;
 use core_test_support::test_codex_exec::test_codex_exec;
 use predicates::str::contains;
+use walkdir::WalkDir;
+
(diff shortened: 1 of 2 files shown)


codex-rs/exec/tests/suite/prompt_stdin.rs
```diff
@@ -152,6 +167,8 @@ fn exec_without_prompt_argument_rejects_empty_piped_stdin() {
         .assert()
         .code(1)
         .stderr(contains("No prompt provided via stdin."));
+
+    assert_eq!(session_rollout_count(test.home_path()), 0);
```
etraut-openai: I don't think this test adds much. Remove it along with the other support routines you added above.

fcoury-oai (approved):

PR #16926, opened 2026-04-06
bazel: Always save bazel repository cache

euroelessar:
This should improve the cache hit ratio for external deps and such

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -223,7 +223,7 @@ jobs:
       # Save bazel repository cache explicitly; make non-fatal so cache uploading
       # never fails the overall job. Only save when key wasn't hit.
       - name: Save bazel repository cache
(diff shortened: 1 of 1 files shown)


zbarsky-openai (approved):

.github/workflows/bazel.yml
```diff
@@ -223,7 +223,7 @@ jobs:
       # Save bazel repository cache explicitly; make non-fatal so cache uploading
       # never fails the overall job. Only save when key wasn't hit.
       - name: Save bazel repository cache
-        if: always() && !cancelled() && steps.cache_bazel_repository_restore.outputs.cache-hit != 'true'
+        if: always()
```
zbarsky-openai: probably can snip this line fully

euroelessar: then it will be skipped if prior step failed a test?

euroelessar: probably should return `!cancelled()` back, hm

PR #16944, opened 2026-04-06
[mcp] Expand tool search to custom MCPs.

mzeng-openai:
- [x] Expand tool search to custom MCPs.
- [x] Rename several variables/fields to be more generic.

Updated tool & server name lifecycles:

**Raw Identity**

ToolInfo.server_name is raw MCP server name.
ToolInfo.tool.name is raw MCP tool name.
MCP calls route back to raw via parse_tool_name() returning (tool.server_name, tool.tool.name).
mcpServerStatus/list now groups by raw server and keys tools by Tool.name: mod.rs:599
App-server just forwards that grouped raw snapshot: codex_message_processor.rs:5245

**Callable Names**

On list-tools, we create provisional callable_namespace / callable_name: mcp_connection_manager.rs:1556
For non-app MCP, provisional callable name starts as raw tool name.
For codex-apps, provisional callable name is sanitized and strips connector name/id prefix; namespace includes connector name.
Then qualify_tools() sanitizes callable namespace + name to ASCII alnum / _ only: mcp_tool_names.rs:128
Note: this is stricter than Responses API. Hyphen is currently replaced with _ for code-mode compatibility.

**Collision Handling**

We do initially collapse example-server and example_server to the same base.
Then qualify_tools() detects distinct raw namespace identities behind the same sanitized namespace and appends a hash to the callable namespace: mcp_tool_names.rs:137
Same idea for tool-name collisions: hash suffix goes on callable tool name.
Final list_all_tools() map key is callable_namespace + callable_name: mcp_connection_manager.rs:769

**Direct Model Tools**

Direct MCP tool declarations use the full qualified sanitized key as the Responses function name.
The raw rmcp Tool is converted but renamed for model exposure.

**Tool Search / Deferred**

Tool search result namespace = final ToolInfo.callable_namespace: tool_search.rs:85
Tool search result nested name = final ToolInfo.callable_name: tool_search.rs:86
Deferred tool handler is registered as "{namespace}:{name}": tool_registry_plan.rs:248
When a function call comes back, core recombines namespace + name, looks up the full qualified key, and gets the raw server/tool for MCP execution: codex.rs:4353

**Separate Legacy Snapshot**

collect_mcp_snapshot_from_manager_with_detail() still returns a map keyed by qualified callable name.
mcpServerStatus/list no longer uses that; it uses McpServerStatusSnapshot, which is raw-inventory shaped.


codex-rs/tools/src/tool_discovery.rs
```diff
@@ -247,29 +268,77 @@ pub fn collect_tool_search_output_tools<'a>(
...
+
+            Some(ToolSearchSourceInfo {
+                name: name.to_string(),
+                description: None,
+            })
         })
         .collect()
 }
 
+fn connector_description(description: Option<&str>) -> Option<String> {
```
sayan-oai: only used once. inline?

codex-rs/core/src/codex.rs
```diff
@@ -6510,6 +6510,42 @@ fn connector_inserted_in_messages(
     connector_count == 1 && skill_count == 0 && mention_names_lower.contains(&mention_slug)
 }
 
+struct McpToolExposure {
+    direct_tools: HashMap<String, McpToolInfo>,
+    deferred_tools: Option<HashMap<String, McpToolInfo>>,
+}
+
+fn build_mcp_tool_exposure(
```
sayan-oai: nit: codex.rs is already massive, can we define elsewhere?

sayan-oai (approved):

codex-rs/codex-mcp/src/mcp_connection_manager.rs
```diff
@@ -135,59 +130,20 @@ pub fn codex_apps_tools_cache_key(auth: Option<&CodexAuth>) -> CodexAppsToolsCac
     }
 }
 
-fn qualify_tools<I>(tools: I) -> HashMap<String, ToolInfo>
```
mzeng-openai, line 139: extracting these out to mcp_tool_names.rs

codex-rs/codex-mcp/src/mcp_tool_names.rs
```diff
@@ -0,0 +1,197 @@
...
+            return (namespace, tool_name, qualified_name);
+        }
+        attempt = attempt.saturating_add(1);
+    }
+}
+
+#[derive(Debug)]
+struct CallableToolCandidate {
+    tool: ToolInfo,
+    raw_namespace_identity: String,
```
mzeng-openai, line 100: use raw_* and callable_* to distinguish between raw names and sanitized callable names.

codex-rs/codex-mcp/src/mcp_tool_names.rs
```diff
@@ -0,0 +1,197 @@
...
+    raw_tool_identity: String,
+    callable_namespace: String,
+    callable_name: String,
+}
+
+pub(crate) fn qualify_tools<I>(tools: I) -> HashMap<String, ToolInfo>
```
sayan-oai, line 111: doc string please; hard to grok what this does at first glance

codex-rs/core/src/tools/handlers/tool_search.rs
sayan-oai: maybe worth keeping an eye on: the search space can now be much larger w/ MCPs, is building the BM25 corpus on every tool_search call a concern? we could cache on ToolSearchHandler creation if so

codex-rs/core/src/tools/router.rs
```diff
@@ -42,8 +42,8 @@ pub struct ToolRouter {
 
 pub(crate) struct ToolRouterParams<'a> {
     pub(crate) mcp_tools: Option<HashMap<String, Tool>>,
+    pub(crate) deferred_mcp_tools: Option<HashMap<String, ToolInfo>>,
     pub(crate) tool_namespaces: Option<HashMap<String, ToolNamespace>>,
```
sayan-oai: stylistic nit/question, not urgent and feel free to push back:

can direct MCP tools also be `ToolInfo`, like `deferred_mcp_tools`? and then we derive `mcp_tools` + `tool_namespaces` internally?

splitting the direct case across `mcp_tools` and `tool_namespaces` makes caller have to know the namespace bookkeeping and makes invalid arg combinations possible.

codex-rs/codex-mcp/src/mcp_tool_names.rs
mzeng-openai: Added a doc comment to `qualify_tools` explaining the raw identity vs callable name split and the uniqueness/64-byte constraints.

codex-rs/core/src/tools/handlers/tool_search.rs
mzeng-openai: Good call. Changed `ToolSearchHandler` so it sorts the tool entries and builds the BM25 `SearchEngine` once in `new()`, then reuses that cached index for each `tool_search` call.

codex-rs/core/src/tools/router.rs
mzeng-openai: good point! updated

sayan-oai (approved):

PR #16946, opened 2026-04-06
[codex] Add danger-full-access denylist-only network mode

viyatb-oai:
## Summary

This adds `experimental_network.danger_full_access_denylist_only` for orgs that want yolo / danger-full-access sessions to keep full network access while still enforcing centrally managed deny rules.

When the flag is true and the session sandbox is `danger-full-access`, the network proxy starts with:

- domain allowlist set to `*`
- managed domain `deny` entries enforced
- upstream proxy use allowed
- all Unix sockets allowed
- local/private binding allowed

Caveat: the denylist is best effort only. In yolo / danger-full-access mode, Codex or the model can use an allowed socket or other local/private network path to bypass the proxy denylist, so this should not be treated as a hard security boundary.

The flag is intentionally scoped to `SandboxPolicy::DangerFullAccess`. Read-only and workspace-write modes keep the existing managed/user allowlist, denylist, Unix socket, and local-binding behavior. This does not enable the non-loopback proxy listener setting; that still requires its own explicit config.

This also threads the new field through config requirements parsing, app-server protocol/schema output, config API mapping, and the TUI debug config output.

## How to use

Add the flag under `[experimental_network]` in the network policy config that is delivered to Codex. The setting is not under `[permissions]`.

```toml
[experimental_network]
enabled = true
danger_full_access_denylist_only = true

[experimental_network.domains]
"blocked.example.com" = "deny"
"*.blocked.example.com" = "deny"
```

With that configuration, yolo / danger-full-access sessions get broad network access except for the managed denied domains above. The denylist remains a best-effort proxy policy because the session may still use allowed sockets to bypass it. Other sandbox modes do not get the wildcard domain allowlist or the socket/local-binding relaxations from this flag.

## Verification

- `cargo test -p codex-config network_requirements`
- `cargo test -p codex-core network_proxy_spec`
- `cargo test -p codex-app-server map_requirements_toml_to_api`
- `cargo test -p codex-tui debug_config_output`
- `just fmt`
- `just fix -p codex-config -p codex-core -p codex-app-server-protocol -p codex-app-server -p codex-tui`
- `just fix -p codex-core -p codex-config`
- `git diff --check`
- `cargo clean`

### DIFF
--- a/codex-rs/core/src/config/network_proxy_spec.rs
+++ b/codex-rs/core/src/config/network_proxy_spec.rs
@@ -20,6 +20,8 @@ use codex_protocol::protocol::SandboxPolicy;
 use std::collections::HashSet;
 use std::sync::Arc;
 
+const GLOBAL_ALLOWLIST_PATTERN: &str = "*";
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct NetworkProxySpec {
     config: NetworkProxyConfig,
@@ -195,6 +197,8 @@ impl NetworkProxySpec {
         let allowlist_expansion_enabled =
             Self::allowlist_expansion_enabled(sandbox_policy, hard_deny_allowlist_misses);
         let denylist_expansion_enabled = Self::denylist_expansion_enabled(sandbox_policy);
+        let danger_full_access_denylist_only =
+            Self::danger_full_access_denylist_only_enabled(requirements, sandbox_policy);
 
         if let Some(enabled) = requirements.enabled {
             config.network.enabled = enabled;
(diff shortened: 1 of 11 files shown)


codex-rs/core/src/config/network_proxy_spec.rs
```diff
@@ -289,6 +299,14 @@ impl NetworkProxySpec {
             config.network.allow_local_binding = allow_local_binding;
             constraints.allow_local_binding = Some(allow_local_binding);
         }
+        if danger_full_access_denylist_only {
```
rreichel3-oai, line 302: Does this need to be checking if full-access mode is enabled too?

rreichel3-oai: Ah just saw the function below

rreichel3-oai (approved):

PR #16948, opened 2026-04-06
[codex] Clarify bubblewrap install warning

adrianbravo-oai:
## Summary
- clarify the Linux bubblewrap startup warning to say the required package is the OS-level Bubblewrap package, not an npm package
- add the upstream Bubblewrap GitHub URL so users can verify they are installing the right project
- extend the sandboxing unit test to cover the new clarification text

## Why
Codex already falls back to vendored bubblewrap, but the existing warning did not explicitly disambiguate the system package from the unrelated npm package name.

## Validation
- cargo test -p codex-sandboxing


adrianbravo-oai:

PR #16961, opened 2026-04-06
app-server: Unify config changes handling a bit

euroelessar:
Currently there are 3 separate functions with a mostly identical logic, so fix it


owenlin0 (approved):
nice cleanup!

PR #16969, opened 2026-04-07
Option to Notify Workspace Owner When Usage Limit is Reached

richardopenai:
## Summary
- Replace the manual `/notify-owner` flow with an inline confirmation prompt when a usage-based workspace member hits a credits-depleted limit.
- Fetch the current workspace role from the live ChatGPT `accounts/check/v4-2023-04-27` endpoint so owner/member behavior matches the desktop and web clients.
- Keep owner, member, and spend-cap messaging distinct so we only offer the owner nudge when the workspace is actually out of credits.

## What Changed
- `backend-client`
  - Added a typed fetch for the current account role from `accounts/check`.
  - Mapped backend role values into a Rust workspace-role enum.
- `app-server` and protocol
  - Added `workspaceRole` to `account/read` and `account/updated`.
  - Derived `isWorkspaceOwner` from the live role, with a fallback to the cached token claim when the role fetch is unavailable.
- `tui`
  - Removed the explicit `/notify-owner` slash command.
  - When a member is blocked because the workspace is out of credits, the error now prompts:
    - `Your workspace is out of credits. Request more from your workspace owner? [y/N]`
  - Choosing `y` sends the existing owner-notification request.
  - Choosing `n`, pressing `Esc`, or accepting the default selection dismisses the prompt without sending anything.
  - Selection popups now honor explicit item shortcuts, which is how the `y` / `n` interaction is wired.

## Reviewer Notes
- The main behavior change is scoped to usage-based workspace members whose workspace credits are depleted.
- Spend-cap reached should not show the owner-notification prompt.
- Owners and admins should continue to see `/usage` guidance instead of the member prompt.
- The live role fetch is best-effort; if it fails, we fall back to the existing token-derived ownership signal.

## Testing
- Manual verification
  - Workspace owner does not see the member prompt.
  - Workspace member with depleted credits sees the confirmation prompt and can send the nudge with `y`.
  - Workspace member with spend cap reached does not see the owner-notification prompt.

### Workspace member out of usage
https://github.com/user-attachments/assets/341ac396-eff4-4a7f-bf0c-60660becbea1

### Workspace owner
Screenshot 2026-04-09 at 11 48 22 AM


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -1734,6 +1815,25 @@ impl CodexMessageProcessor {
                 message: format!("failed to fetch codex rate limits: {err}"),
                 data: None,
             })?;
+        tracing::info!(
+            snapshot_count = snapshots.len(),
```
aibrahim-oai: how often this comes? just to not pollute the logs

aibrahim-oai (approved):

PR #16974, opened 2026-04-07
[codex] Preserve zsh PATH in shell snapshots

sayan-oai:
## Summary

Fix zsh shell snapshot export filtering so tied parameters such as `PATH`/`path` are not dropped. zsh emits `PATH` from `export -p` as `export -T PATH path=(...)`, and the previous parser treated `-T PATH path` as the variable name and filtered it out as invalid.

Add a macOS regression test that captures a zsh snapshot from a temporary `.zshrc`, sources the snapshot in a fresh zsh process with a deliberately wrong `PATH`, and verifies the captured login-shell `PATH` is restored.

## Impact

This keeps shell snapshots from losing the login-shell `PATH` on zsh. That matters for extension-host or other thin-parent-environment sessions where inheriting the parent `PATH` does not mask a missing snapshot `PATH` export.

## Validation

- `just fmt`
- `cargo test -p codex-core zsh_snapshot_preserves_tied_path_export`
- `cargo test -p codex-core shell_snapshot::tests`
- `cargo test -p codex-core` was also run locally, but failed in unrelated integration/approval tests after the shell snapshot unit tests passed. Failures included approval/otel cases plus shell snapshot integration tests timing out while waiting for zsh snapshot creation in this local environment.


MSch:
fyi, I ran into this, there's also `typeset -UT` (U = unique) which leads to `export -UT`. See https://github.com/openai/codex/compare/main...MSch:codex:fix-zsh-shell-snapshot-path?expand=1 for how my codex fixed it

viyatb-oai (approved):

codex-rs/core/src/shell_snapshot.rs
```diff
@@ -342,7 +342,14 @@ export_lines=$(export -p | awk '
   line=$0
   name=line
   sub(/^(export|declare -x|typeset -x) /, "", name)
-  sub(/=.*/, "", name)
+  if (name ~ /^-[A-Za-z]*T[A-Za-z]* [A-Za-z_][A-Za-z0-9_]* [A-Za-z_][A-Za-z0-9_]*=/) {
```
jif-oai: Can we parse zsh export -p option groups generically here instead of only special-casing tied params?

codex-rs/core/src/shell_snapshot.rs
```diff
@@ -342,7 +342,10 @@ export_lines=$(export -p | awk '
   line=$0
   name=line
   sub(/^(export|declare -x|typeset -x) /, "", name)
-  sub(/=.*/, "", name)
+  # Zsh prefixes attributed exports with an option group and prints tied
+  # parameters with another name, as in `export -UT PATH path=(...)`.
+  sub(/^-[^ ]+ /, "", name)
```
jif-oai, line 347: I think this broadens the snapshot from preserving tied params to preserving ALL zsh parameter attributes... That can break the wrapper path: if the snapshot contains `export -r HTTPS_PROXY=.../ CODEX_THREAD_ID/PATH`, the later env restore after `.` snapshot errors with read-only variable before we exec the real command

0xdevalias:
fixes https://github.com/openai/codex/issues/20220

PR #16980, opened 2026-04-07
Add full-ci branch trigger

pakrym-oai:
Allow branches to trigger full ci (helpful to run remote tests)


starr-openai (approved):

codex-rs/core/src/plugins/startup_sync_tests.rs
```diff
@@ -34,6 +34,7 @@ fn has_plugins_clone_dirs(codex_home: &Path) -> bool {
     })
 }
 
+#[cfg(unix)]
```
starr-openai, line 37: intentional?

pakrym-oai: yes, it was broken in main

PR #16987, opened 2026-04-07
Fix missing resume hint on zero-token exits

etraut-openai:
Addresses #16421

Problem: Resumed interactive sessions exited before new token usage skipped all footer lines, hiding the `codex resume` continuation command.

It's not clear whether this was an intentional design choice, but I think it's reasonable to expect this message under these circumstances.

Solution: Compose token usage and resume hints independently so resumable sessions still print the continuation command with zero usage.


fcoury-oai (approved):

fcoury-oai (commented):
Replicated the issue on current stable version and confirmed the fix running this branch.

Code looks good, approved.

PR #17025, opened 2026-04-07
Add role-level fork_context defaults

friel-openai:
## Summary
- add role-level `fork_context` defaults for agent roles
- let v1 spawn_agent default `fork_context` from the selected role when omitted
- let MultiAgentV2 default `fork_turns` from the selected role when omitted
- add tests for built-in, discovered-file, and config.toml role overrides

Stacked on top of #16055 (`dev/friel/fork-context-inherits-parent-model`).

## Test
- `cargo +1.93.1 test -p codex-core default_fork_context_for_role --lib`
- `cargo +1.93.1 test -p codex-core spawn_agent_fork_context_ignores_child_model_overrides --lib`
- `cargo +1.93.1 test -p codex-core multi_agent_v2_spawn_fork_turns_ignores_child_model_overrides --lib`


robobryce:
We'd still very much like this feature.

For custom subagent roles, `fork_context` / `fork_turns` is part of the role's intended behavior, not something every spawning manager should have to remember and repeat on each `spawn_agent` call. In particular, we have worker definitions that should consistently receive a bounded amount of parent context. Without a role-level default, that policy cannot live in the subagent definition and is easy for callers to omit.

Please consider reviving this or landing equivalent support for role-level fork defaults in both the v1 and MultiAgentV2 spawn paths.

PR #17036, opened 2026-04-07
feat: enforce allow_git through permissions

viyatb-oai:
## Stack

1. #25664: runtime sandbox support for limited Git metadata writes.
2. This PR: ergonomic `allow_git` permissions and requirements setting built on that runtime support.

## Why

`allow_git` is the ergonomic user-facing switch on top of the runtime limited Git write sandbox support in #25664. It lets requirements and permission profiles ask for Git metadata writes without exposing another `[sandbox_workspace_write]` knob.

## What changed

- Adds `allow_git` under permissions filesystem config, including requirements-level `[permissions.filesystem]`, so it can be enforced across profile selection and runtime profile updates.
- Maps `allow_git` onto the internal limited Git write sandbox policy from #25664.
- Keeps the default restrictive behavior unchanged unless the permission or requirements setting opts in.
- Handles explicit `:workspace_roots` rules for `.git`: managed requirements remove conflicting writes to protected Git config/hooks paths, and disabling the managed setting re-protects all `.git` metadata.
- Does not add `[sandbox_workspace_write].allow_git`; legacy sandbox policy structs keep only the internal plumbing needed by app-server and platform lowerers.

## Security

`allow_git` does not imply network access, so `git push` still depends on the active network policy. It also does not make Git config or hooks writable. Managed requirements win over local profile entries that try to write those protected paths.

## Validation

- `just fmt`
- `just clippy -p codex-core`

Fixes #15505


viyatb-oai:
@codex review

PR #17043, opened 2026-04-07
[mcp] Support server-driven elicitations

mzeng-openai:
- [x] Enables MCP elicitation for custom servers, not just Codex Apps
- [x] Adds an RMCP service wrapper to preserve elicitation _meta
- [x] Round-trips response _meta for persist/approval choices
- [x] Updates TUI empty-schema elicitations into message-only approval prompts


leoshimo-oai:
Tested locally - lgtm 👍 

Curious if it'd be worth upstreaming `_meta` change into `rmcp`'s `CreateElicitationResult`

codex-rs/rmcp-client/src/elicitation_client_service.rs
```diff
@@ -0,0 +1,213 @@
+use std::sync::Arc;
```
mzeng-openai, line 1: Adds a thin RMCP Service wrapper so Codex can round-trip elicitation request/response _meta while keeping the existing logging/notification handler behavior.
This wrapper will be significantly smaller once we add _meta to CreateElicitationResult in rmcp in https://github.com/modelcontextprotocol/rust-sdk/pull/792

codex-rs/codex-mcp/src/mcp_connection_manager.rs
```diff
@@ -1315,19 +1315,15 @@ impl From<anyhow::Error> for StartupOutcomeError {
     }
 }
 
-fn elicitation_capability_for_server(server_name: &str) -> Option<ElicitationCapability> {
-    if server_name == CODEX_APPS_MCP_SERVER_NAME {
```
shijie-oai, line 1319: Is this intended?

mzeng-openai: yes, this is the exact purpose of this PR which is to expand elicitation support to custom MCPs

shijie-oai (approved):

PR #17071, opened 2026-04-08
Configure multi_agent_v2 spawn agent hints

pakrym-oai:
Allow multi_agent_v2 features to have its own temporary configuration under `[features.multi_agent_v2]`

```
[features.multi_agent_v2]
enabled = true
usage_hint_enabled = false
usage_hint_text = "Custom delegation guidance."
hide_spawn_agent_metadata = true
```

Absent `usage_hint_text` means use the default hint.

```
[features]
multi_agent_v2 = true
```

still works as the boolean shorthand.


codex-rs/features/src/lib.rs
```diff
@@ -492,8 +490,76 @@ pub fn is_known_feature_key(key: &str) -> bool {
...
+impl<T: FeatureConfig> FeatureToml<T> {
+    pub fn enabled(&self) -> Option<bool> {
+        match self {
+            Self::Enabled(enabled) => Some(*enabled),
+            Self::Config(config) => config.enabled(),
+        }
+    }
+}
+
+pub trait FeatureConfig {
```
jif-oai, line 548: ultra nit: can you document the pub traits?

codex-rs/features/src/lib.rs
```diff
@@ -492,8 +490,76 @@ pub fn is_known_feature_key(key: &str) -> bool {
...
+    fn enabled(&self) -> Option<bool>;
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Default, PartialEq, Eq, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct MultiAgentV2ConfigToml {
```
jif-oai: Personal opinion but business specific struct + their implementation of public traits should not live in `lib.rs` or `mod.rs`. Those file should be kept as generic as possible. Can we move this to a dedicated file?

jif-oai (approved):

PR #17073, opened 2026-04-08
Support clear SessionStart source

abhinav-oai:
## Motivation

The `SessionStart` hook already receives `startup` and `resume` sources, but sessions created from `/clear` previously looked like normal startup sessions. This makes it impossible for hook authors to distinguish between these with the matcher.

## Summary

- Add `InitialHistory::Cleared` so `/clear`-created sessions can be distinguished from ordinary startup sessions.
- Add `SessionStartSource::Clear` and wire it through core, app-server thread start params, and TUI clear-session flow.
- Update app-server protocol schemas, generated TypeScript, docs, and related tests.

https://github.com/user-attachments/assets/9cae3cb4-41c7-4d06-b34f-966252442e5c


abhinav-oai:

eternal-openai (approved):

PR #17076, opened 2026-04-08
[codex] Show ctrl + t hint on truncated exec output in TUI

mom-oai:
## What

Show an inline `ctrl + t to view transcript` hint when exec output is truncated in the main TUI chat view.

## Why

Today, truncated exec output shows `… +N lines`, but it does not tell users that the full content is already available through the existing transcript overlay. That makes hidden output feel lost instead of discoverable.

This change closes that discoverability gap without introducing a new interaction model.

Fixes: CLI-5740

## How

- added an output-specific truncation hint in `ExecCell` rendering
- applied that hint in both exec-output truncation paths:
  - logical head/tail truncation before wrapping
  - row-budget truncation after wrapping
- preserved the existing row-budget behavior on narrow terminals by reserving space for the longer hint line
- updated the relevant snapshot and added targeted regression coverage

## Intentional design decisions

- **Aligned shortcut styling with the visible footer UI**  
  The inline hint uses `ctrl + t`, not `Ctrl+T`, to match the TUI’s rendered key-hint style.

- **Kept the noun `transcript`**  
  The product already exposes this flow as the transcript overlay, so the hint points at the existing concept instead of inventing a new label.

- **Preserved narrow-terminal behavior**  
  The longer hint text is accounted for in the row-budget truncation path so the visible output still respects the existing viewport cap.

- **Did not add the hint to long command truncation**  
  This PR only changes hidden **output** truncation. Long command truncation still uses the plain ellipsis form because `ctrl + t` is not the same kind of “show hidden output” escape hatch there.

- **Did not widen scope to other truncation surfaces**  
  This does not change MCP/tool-call truncation in `history_cell.rs`, and it does not change transcript-overlay behavior itself.

### Automated
- `just fmt`
- `cargo test -p codex-tui`

### Manual
- ran `just tui-with-exec-server`
- executed `!seq 1 200`
- confirmed the main view showed the new `ctrl + t to view transcript` truncation hint
- pressed `ctrl + t` and confirmed the transcript overlay still exposed the full output
- closed the overlay and returned to the main view

## Visual proof

Screenshot/video attached in the PR UI showing:
- the truncated exec output row with the new hint
- the transcript overlay after `ctrl + t`

### DIFF
--- a/codex-rs/tui/src/exec_cell/render.rs
+++ b/codex-rs/tui/src/exec_cell/render.rs
@@ -29,6 +29,7 @@ use unicode_width::UnicodeWidthStr;
 pub(crate) const TOOL_CALL_MAX_LINES: usize = 5;
 const USER_SHELL_TOOL_CALL_MAX_LINES: usize = 50;
 const MAX_INTERACTION_PREVIEW_CHARS: usize = 80;
+const TRANSCRIPT_HINT: &str = "ctrl + t to view transcript";
 
 pub(crate) struct OutputLinesParams {
     pub(crate) line_limit: usize,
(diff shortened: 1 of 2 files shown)


mom-oai:

canvrno-oai (approved):

PR #17086, opened 2026-04-08
Fix TUI crash when resuming the current thread

etraut-openai:
Problem: Resuming the live TUI thread through `/resume` could unsubscribe and reconnect the same app-server thread, leaving the UI crashed or disconnected.

Solution: No-op `/resume` only when the selected thread is the currently attached active thread; keep the normal resume path for stale/displayed-only threads so recovery and reattach still work.


fcoury-oai (approved):
I was able to replicate the crash and the fix.

Looks good.

PR #17088, opened 2026-04-08
[codex-analytics] ingest server requests and responses

rhan-oai:
## Why

Codex analytics needs a typed seam for app-server-originated request/response traffic so future tool-approval analytics can consume those facts without adding bespoke callsite tracking each time. Server responses arrive as JSON-RPC `id + result` payloads, so analytics has to reconstruct the matching typed response from the original typed request while that request context still exists in app-server.

This also puts analytics on the app-server outbound path, which needs to avoid keeping the runtime alive during shutdown. The final ownership fix keeps the normal strong auth-manager retention in analytics and makes the external-auth refresh bridge hold a weak back-reference to `OutgoingMessageSender`, breaking the runtime cycle at the bridge boundary instead of exposing retention policy through the analytics client API.

## What changed

- Adds typed `ServerRequest` and `ServerResponse` analytics facts, plus `AnalyticsEventsClient::track_server_request` and `track_server_response`.
- Renames the existing client-side facts to `ClientRequest` and `ClientResponse` so reducers can distinguish client-to-server traffic from server-to-client traffic.
- Adds `ServerRequest::response_from_result`, allowing a stored typed request to decode the matching typed server response from a raw JSON-RPC result payload.
- Threads `AnalyticsEventsClient` through `OutgoingMessageSender` and records targeted server requests, replayed targeted requests, and matching targeted responses with the responding connection id needed for correlation.
- Intentionally leaves broadcast server requests/responses out of analytics for now because the current model is per connection, while broadcasts fan one logical request out across multiple connections.
- Breaks the app-server shutdown cycle by storing `Weak<OutgoingMessageSender>` in `ExternalAuthRefreshBridge` and upgrading it only when an external-auth refresh is actually requested.
- Keeps reducer ingestion of the new server-side facts as no-ops for now; this PR is plumbing for later tool-approval analytics work.

## Verification

- `cargo test -p codex-app-server outgoing_message::tests::`
  - Covers typed-response reconstruction plus the targeted, replayed, broadcast-exclusion, and response-attribution analytics paths.

## Follow-up

This PR intentionally stops at ingestion plumbing, so `ServerRequest` and `ServerResponse` facts are still reducer no-ops. Once a follow-up PR adds real downstream analytics output for those facts:

- replace the temporary pre-reducer observation seam with reducer tests for the emitted event shape;
- add end-to-end coverage in `app-server/tests/suite/v2/analytics.rs` for the real app-server workflow and captured analytics payload;
- remove the temporary sender-level observer tests added here in favor of the real-output coverage above.

---

Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/17088).
* #18748
* #18747
* #17090
* #17089
* #20241
* #20239
* __->__ #17088

### DIFF
--- a/codex-rs/analytics/src/client.rs
+++ b/codex-rs/analytics/src/client.rs
@@ -27,6 +27,8 @@ use codex_app_server_protocol::InitializeParams;
 use codex_app_server_protocol::JSONRPCErrorError;
 use codex_app_server_protocol::RequestId;
 use codex_app_server_protocol::ServerNotification;
+use codex_app_server_protocol::ServerRequest;
+use codex_app_server_protocol::ServerResponse;
 use codex_login::AuthManager;
 use codex_login::default_client::create_client;
 use codex_plugin::PluginTelemetryMetadata;
@@ -49,8 +51,7 @@ pub(crate) struct AnalyticsEventsQueue {
 
 #[derive(Clone)]
 pub struct AnalyticsEventsClient {
-    queue: AnalyticsEventsQueue,
-    analytics_enabled: Option<bool>,
+    queue: Option<AnalyticsEventsQueue>,
 }
 
 impl AnalyticsEventsQueue {
@@ -119,9 +120,13 @@ impl AnalyticsEventsClient {
         analytics_enabled: Option<bool>,
     ) -> Self {
         Self {
-            queue: AnalyticsEventsQueue::new(Arc::clone(&auth_manager), base_url),
-            analytics_enabled,
+            queue: (analytics_enabled != Some(false))
+                .then(|| AnalyticsEventsQueue::new(Arc::clone(&auth_manager), base_url)),
         }
+    }
+
+    pub fn disabled() -> Self {
+        Self { queue: None }
     }
 
     pub fn track_skill_invocations(
@@ -182,7 +187,7 @@ impl AnalyticsEventsClient {
     }
 
     pub fn track_request(&self, connection_id: u64, request_id: RequestId, request: ClientRequest) {
-        self.record_fact(AnalyticsFact::Request {
+        self.record_fact(AnalyticsFact::ClientRequest {
             connection_id,
             request_id,
             request: Box::new(request),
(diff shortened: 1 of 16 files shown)


codex-rs/app-server/src/in_process.rs
```diff
@@ -363,7 +364,14 @@ fn start_uninitialized(args: InProcessStartArgs) -> InProcessClientHandle {
 
     let runtime_handle = tokio::spawn(async move {
         let (outgoing_tx, mut outgoing_rx) = mpsc::channel::<OutgoingEnvelope>(channel_capacity);
-        let outgoing_message_sender = Arc::new(OutgoingMessageSender::new(outgoing_tx));
+        let auth_manager =
+            AuthManager::shared_from_config(args.config.as_ref(), args.enable_codex_api_key_env);
```
pakrym-oai: don't we have an auth_manager on InProcessStartArgs?

rhan-oai:
@codex

codex-rs/app-server/src/in_process.rs
rhan-oai: right now is not the case
https://github.com/openai/codex/blob/a9c111da544c976d591343db5493a7da283b72e5/codex-rs/app-server/src/in_process.rs#L111

rhan-oai:
@codex

rhan-oai:
@codex

rhan-oai:
@codex

codex-rs/analytics/src/client.rs
```diff
@@ -49,18 +51,34 @@ pub(crate) struct AnalyticsEventsQueue {
 
 #[derive(Clone)]
 pub struct AnalyticsEventsClient {
-    queue: AnalyticsEventsQueue,
-    analytics_enabled: Option<bool>,
+    queue: Option<AnalyticsEventsQueue>,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum AuthManagerRetention {
```
pakrym-oai: something doesn't feel right here.

owenlin0 (approved):
nit: perhaps call `codex-rs/app-server/src/analytics_events.rs` -> `analytics_utils.py`?

[followup?] also, now we decode the same JSON payload twice in the happy path: once for analytics, then again in the actual handler. as a followup it'd be nice to see if there's a way we can  decode the JSON payload just once (probably not a big deal atm in practice since the client likely doesn't send massive JSON responses - the only use case where that should be possible is dynamic tool calls)

PR #17089, opened 2026-04-08
[codex-analytics] add tool item event schemas

rhan-oai:
## Why

Tool analytics need stable, typed payloads before the later lifecycle reducer starts emitting them. Keeping the event schema definitions isolated in their own PR makes the emitted surface reviewable separately from the reducer logic that produces those events.

## What changed

- Adds the common tool-item analytics event base plus event payload types for command execution, file changes, MCP calls, dynamic tools, collaboration tools, web search, and image generation.
- Extends `TrackEventRequest` with the corresponding tool-item variants.
- Adds serialization coverage for the command-execution event shape.

## Verification

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/17089).
* #18748
* #18747
* #17090
* __->__ #17089
* #20514

### DIFF
--- a/codex-rs/analytics/src/events.rs
+++ b/codex-rs/analytics/src/events.rs
@@ -61,6 +61,20 @@ pub(crate) enum TrackEventRequest {
     Compaction(Box<CodexCompactionEventRequest>),
     TurnEvent(Box<CodexTurnEventRequest>),
     TurnSteer(CodexTurnSteerEventRequest),
+    #[allow(dead_code)]
+    CommandExecution(CodexCommandExecutionEventRequest),
+    #[allow(dead_code)]
+    FileChange(CodexFileChangeEventRequest),
+    #[allow(dead_code)]
+    McpToolCall(CodexMcpToolCallEventRequest),
+    #[allow(dead_code)]
+    DynamicToolCall(CodexDynamicToolCallEventRequest),
+    #[allow(dead_code)]
+    CollabAgentToolCall(CodexCollabAgentToolCallEventRequest),
+    #[allow(dead_code)]
+    WebSearch(CodexWebSearchEventRequest),
+    #[allow(dead_code)]
+    ImageGeneration(CodexImageGenerationEventRequest),
     PluginUsed(CodexPluginUsedEventRequest),
     PluginInstalled(CodexPluginEventRequest),
     PluginUninstalled(CodexPluginEventRequest),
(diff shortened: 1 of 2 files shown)


codex-rs/analytics/src/events.rs
```diff
@@ -386,6 +398,208 @@ pub(crate) struct GuardianReviewEventPayload {
...
+pub(crate) struct CodexCommandExecutionEventParams {
+    #[serde(flatten)]
+    pub(crate) base: CodexToolItemEventBase,
+    pub(crate) command_execution_source: CommandExecutionSource,
+    pub(crate) exit_code: Option<i32>,
+    pub(crate) command_total_action_count: u64,
+    pub(crate) command_read_action_count: u64,
+    pub(crate) command_list_files_action_count: u64,
+    pub(crate) command_search_action_count: u64,
+    pub(crate) command_unknown_action_count: u64,
```
owenlin0, line 495: hmm, how are all these `*_action_count` going to be populated? we're not going to be parsing the shell command, are we?

owenlin0: i mean, technically we can, but would it be brittle and hard to maintain?

owenlin0: or do we already have existing code doing this?

codex-rs/analytics/src/events.rs
```diff
@@ -386,6 +398,208 @@ pub(crate) struct GuardianReviewEventPayload {
...
+    pub(crate) app_server_client: CodexAppServerClientMetadata,
+    pub(crate) runtime: CodexRuntimeMetadata,
+    pub(crate) thread_source: Option<&'static str>,
+    pub(crate) subagent_source: Option<String>,
+    pub(crate) parent_thread_id: Option<String>,
+    pub(crate) tool_name: String,
+    pub(crate) started_at_ms: u64,
+    pub(crate) completed_at_ms: u64,
+    pub(crate) duration_ms: Option<u64>,
+    pub(crate) review_count: u64,
```
owenlin0, line 455: is this just `guardian_review_count` + `user_review_count`?

owenlin0 (approved):
small comment about whether we really want to include `_action_count` on command executions but otherwise lgtm

codex-rs/analytics/src/events.rs
rhan-oai: yes we do, parsed as command actions:
```
pub enum CommandAction {
    Read {
        command: String,
        name: String,
        path: AbsolutePathBuf,
    },
    ListFiles {
        command: String,
        path: Option<String>,
    },
    Search {
        command: String,
        query: Option<String>,
        path: Option<String>,
    },
    Unknown {
        command: String,
    },
}
```

PR #17116, opened 2026-04-08
chore: merge name and title

jif-oai:
Merge title and name concept to leverage the sqlite title column and have more efficient queries


fcoury-oai (changes requested):
We might need to handle backwards compatibility.

If I rename a session on codex stable and try to retrieve it from this branch it doesn't work.

fcoury-oai (approved):
Looks good, approved.

PR #17141, opened 2026-04-08
feat: add layered --profile-v2 config files

jif-oai:
## Why

`--profile-v2 <name>` gives launchers and runtime entry points a named profile config without making each profile duplicate the base user config. The base `$CODEX_HOME/config.toml` still loads first, then `$CODEX_HOME/<name>.config.toml` layers above it and becomes the active writable user config for that session.

That keeps shared defaults, plugin/MCP setup, and managed/user constraints in one place while letting a named profile override only the pieces that need to differ.

## What Changed

- Added the shared `--profile-v2 <name>` runtime option with validated plain names, now represented by `ProfileV2Name`.
- Extended config layer state so the base user config and selected profile config are both `User` layers; APIs expose the active user layer and merged effective user config.
- Threaded profile selection through runtime entry points: `codex`, `codex exec`, `codex review`, `codex resume`, `codex fork`, and `codex debug prompt-input`.
- Made user-facing config writes go to the selected profile file when active, including TUI/settings persistence, app-server config writes, and MCP/app tool approval persistence.
- Made plugin, marketplace, MCP, hooks, and config reload paths read from the merged user config so base and profile layers both participate.
- Updated app-server config layer schemas to mark profile-backed user layers.

## Limits

`--profile-v2` is still rejected for config-management subcommands such as feature, MCP, and marketplace edits. Those paths remain tied to the base `config.toml` until they have explicit profile-selection semantics.

Some adjacent background writes may still update base or global state rather than the selected profile:

- marketplace auto-upgrade metadata
- automatic MCP dependency installs from skills
- remote plugin sync or uninstall config edits
- personality migration marker/default writes

## Verification

Added targeted coverage for profile name validation, layer ordering/merging, selected-profile writes, app-server config writes, session hot reload, plugin config merging, hooks/config fixture updates, and MCP/app approval persistence.


codex-rs/config/src/state.rs
```diff
@@ -17,6 +17,7 @@ use toml::Value as TomlValue;
 /// LoaderOverrides overrides managed configuration inputs (primarily for tests).
 #[derive(Debug, Default, Clone)]
 pub struct LoaderOverrides {
+    pub user_config_path: Option<PathBuf>,
```
bolinfest: I realize the others are `PathBuf`, but this should be `AbsolutePathBuf` and we'll migrate the others later?

Also, should we force this to be required though acknowledge that it's possible no file exists at the specified path?

codex-rs/config/src/state.rs
```diff
@@ -198,14 +206,54 @@ impl ConfigLayerStack {
         self.ignore_user_and_project_exec_policy_rules
     }
 
-    /// Returns the raw user config layer, if any.
+    /// Returns the active raw user config layer, if any.
```
bolinfest, line 282: Rename the function to `get_active_user_layer()` then? Or maybe `get_writable_user_layer()` if you like that better?

codex-rs/config/src/state.rs
```diff
@@ -198,14 +206,54 @@ impl ConfigLayerStack {
...
+
+    /// Returns all user config layers in precedence order.
+    pub fn get_user_layers(
+        &self,
+        ordering: ConfigLayerStackOrdering,
+        include_disabled: bool,
+    ) -> Vec<&ConfigLayerEntry> {
+        self.get_layers(ordering, include_disabled)
+            .into_iter()
+            .filter(|layer| matches!(layer.name, ConfigLayerSource::User { .. }))
```
bolinfest, line 313: Do you think we should add another variant to the `ConfigLayerSource` enum named `UserProfile` instead?

Or change the existing `User` variant

https://github.com/openai/codex/blob/80d8db22bbdb453e3a76dbe5c3475465c26b3a54/codex-rs/app-server-protocol/src/protocol/v2.rs#L592-L596

to have an extra field, like:

```
    User {
        /// This is the path to the user's config.toml file, though it is not
        /// guaranteed to exist.
        file: AbsolutePathBuf,

        /// Name of profile if loaded with --profile.
        profile: Option<String>,
    },
```

In either case `ConfigLayerSource::precedence()` needs to be updated so the profile layer has a higher precedence than the non-profile user layer?

codex-rs/app-server/src/config_manager.rs
```diff
@@ -1,5 +1,6 @@
 use codex_arg0::Arg0DispatchPaths;
 use codex_cloud_requirements::cloud_requirements_loader;
+use codex_config::CONFIG_TOML_FILE;
```
bolinfest: Hmm, to date, only tests have had to import this, which I only call out because it means that we have generally ensured config-related logic goes through our abstractions instead of embedding this config-specific detail.

Though perhaps we should also be asking why `ConfigManager` lives in the `codex-app-server` crate?

codex-rs/app-server/src/config_manager.rs
```diff
@@ -59,6 +60,16 @@ impl ConfigManager {
         self.codex_home.as_path()
     }
 
+    pub(crate) fn user_config_path(&self) -> std::io::Result<AbsolutePathBuf> {
```
bolinfest, line 65: Maybe this method should be on `LoaderOverrides` itself?

codex-rs/app-server/src/config_manager.rs
bolinfest: Oh wait, I lied, we define this 3x:

Image

codex-rs/cli/src/main.rs
```diff
@@ -752,6 +755,9 @@ async fn cli_main(arg0_paths: Arg0DispatchPaths) -> anyhow::Result<()> {
     root_config_overrides.raw_overrides.extend(toggle_overrides);
     let root_remote = remote.remote;
     let root_remote_auth_token_env = remote.remote_auth_token_env;
+    if let Some(subcommand) = subcommand.as_ref() {
+        reject_profile_v2_for_subcommand(&interactive, subcommand)?;
+    }
```
bolinfest, line 861: I think we said we'd just stick with `--profile` instead of `--profile-v2`?

codex-rs/config/src/loader/layer_io.rs
```diff
@@ -42,13 +42,15 @@ pub(super) async fn load_config_layers_internal(
 ) -> io::Result<LoadedConfigLayers> {
     #[cfg(target_os = "macos")]
     let LoaderOverrides {
+        user_config_path: _,
```
bolinfest: This should be used, right?

codex-rs/core/src/config/mod.rs
```diff
@@ -1030,19 +1031,43 @@ impl Config {
     }
 }
 
+pub fn resolve_profile_v2_config_path(
+    codex_home: &Path,
+    profile_name: &str,
```
bolinfest: Instead of excludes, should we just require `profile_name` to match `/[a-zA-Z0-9_-]+/`?

codex-rs/cli/src/main.rs
jif-oai: Yes but I first one to merge and then have a follow-up where we retire the old one

codex-rs/cli/src/main.rs
```diff
@@ -1289,6 +1298,31 @@ async fn cli_main(arg0_paths: Arg0DispatchPaths) -> anyhow::Result<()> {
...
+            "--profile-v2 only applies to runtime commands: `codex`, `codex exec`, `codex review`, `codex resume`, `codex fork`, and `codex debug prompt-input`."
+        );
+    }
+    Ok(())
+}
+
+fn subcommand_supports_profile_v2(subcommand: &Subcommand) -> bool {
+    matches!(
+        subcommand,
+        Subcommand::Exec(_)
```
bolinfest, line 1467: Shouldn't each subcommand that takes `--profile` have it declared on the struct and then I think this method should return the value, if set, so `Option<String>` rather than `bool`.

bolinfest: Maybe we want to do the NewType pattern for the profile value so that, by construction, we assert that the profile name contains only certain characters?

codex-rs/tui/src/app/tests.rs
```diff
@@ -3549,35 +3549,38 @@ async fn discard_side_thread_removes_agent_navigation_entry() -> Result<()> {
...
-    app.active_thread_id = Some(side_thread_id);
-    app.side_threads
-        .insert(side_thread_id, SideThreadState::new(parent_thread_id));
-    app.agent_navigation.upsert(
-        side_thread_id,
-        Some("Side".to_string()),
-        Some("side".to_string()),
-        /*is_closed*/ false,
-    );
+    Box::pin(async {
```
bolinfest, line 3492: This is the first time I can remember us needing to `Box::pin()` within a test?

jif-oai: Yes, I had some stack overflow but they might have been resolved. I can check if main is fine now

codex-rs/utils/cli/src/shared_options.rs
```diff
@@ -2,7 +2,69 @@
...
+            f,
+            "invalid --profile-v2 value `{}`; pass a plain name such as `work`",
+            self.value
+        )
+    }
+}
+
+impl std::error::Error for ProfileV2NameParseError {}
+
+impl FromStr for ProfileV2Name {
```
bolinfest: Not sure if it's worth a test like:

```rust
assert_eq!(
    ProfileV2Name::from_str("../foo"),
    ProfileV2NameParseError {
        value: "../foo".to_string()
    },
    "dots and slashes are disallowed to prevent reading arbitrary files",
)
```

and if you add one test, might as well also do:

```rust
assert_eq!(
    ProfileV2Name::from_str(""),
    ProfileV2NameParseError {
        value: "".to_string()
    },
    "profile name cannot be empty",
)
```

codex-rs/config/src/state.rs
```diff
@@ -11,12 +11,15 @@ use codex_app_server_protocol::ConfigLayerSource;
...
 use std::collections::HashMap;
+use std::path::Path;
 use std::path::PathBuf;
 use toml::Value as TomlValue;
 
 /// LoaderOverrides overrides managed configuration inputs (primarily for tests).
 #[derive(Debug, Default, Clone)]
 pub struct LoaderOverrides {
+    pub user_config_path: Option<AbsolutePathBuf>,
+    pub user_config_profile: Option<String>,
```
bolinfest: Why not `ProfileV2Name`?

```suggestion
    pub user_config_profile: Option<ProfileV2Name>,
```

codex-rs/cli/src/main.rs
```diff
@@ -1422,6 +1454,25 @@ async fn disable_feature_in_config(interactive: &TuiCli, feature: &str) -> anyho
...
+fn loader_overrides_for_profile(
+    profile_v2: Option<&ProfileV2Name>,
+) -> anyhow::Result<LoaderOverrides> {
+    match profile_v2 {
+        Some(profile_v2) => {
+            let codex_home = find_codex_home()?;
+            Ok(LoaderOverrides {
+                user_config_path: Some(resolve_profile_v2_config_path(
+                    &codex_home,
+                    profile_v2.as_str(),
```
bolinfest: Why isn't the second arg of type `&ProfileV2Name`?

codex-rs/config/src/state.rs
```diff
@@ -240,14 +262,54 @@ impl ConfigLayerStack {
...
 
+    pub fn get_user_config_file(&self) -> Option<&AbsolutePathBuf> {
+        let layer = self.get_active_user_layer()?;
+        let ConfigLayerSource::User { file, .. } = &layer.name else {
+            return None;
+        };
+        Some(file)
+    }
+
+    /// Returns all user config layers in precedence order.
```
bolinfest: Clarify if ascending or descending order?

codex-rs/config/src/state.rs
bolinfest: Docstring needs to be updated to explain what happens when there are multiple user layers?

codex-rs/config/src/state.rs
```diff
@@ -261,49 +323,92 @@ impl ConfigLayerStack {
...
+            ConfigLayerSource::User { file, profile } if file == config_toml => profile.clone(),
+            _ => None,
+        });
+        self.with_user_config_profile(config_toml, profile, user_config)
+    }
+
+    pub fn with_user_config_profile(
+        &self,
+        config_toml: &AbsolutePathBuf,
+        profile: Option<String>,
```
bolinfest: Again, prefer `ProfileV2Name`?

codex-rs/config/src/state_tests.rs
```diff
@@ -1,6 +1,11 @@
 use super::*;
 use pretty_assertions::assert_eq;
 
+fn test_user_config_path(file_name: &str) -> AbsolutePathBuf {
+    AbsolutePathBuf::from_absolute_path(std::env::temp_dir().join("codex-test").join(file_name))
```
bolinfest: This feels like it could incur collisions? Should you pass in a `TempDir` instead of using the global `std::env::temp_dir()`?

codex-rs/core-plugins/src/loader_tests.rs
```diff
@@ -1,8 +1,68 @@
...
 use crate::manifest::load_plugin_manifest;
+use codex_config::ConfigLayerEntry;
+use codex_config::ConfigLayerSource;
+use codex_config::ConfigRequirements;
+use codex_config::ConfigRequirementsToml;
 use codex_plugin::PluginId;
 use pretty_assertions::assert_eq;
 
+fn user_config_path(file_name: &str) -> AbsolutePathBuf {
+    AbsolutePathBuf::from_absolute_path(std::env::temp_dir().join("codex-test").join(file_name))
```
bolinfest: Again, I would generally avoid `std::env::temp_dir()` unless you're sure you won't get undefined behavior in the presence of running the test suite from multiple clones of codex concurrently.

I just did `rg std::env::temp_dir` and I didn't realize we were using it so much...

codex-rs/core-plugins/src/loader_tests.rs
```diff
@@ -1,8 +1,68 @@
...
+    )
+}
+
+#[test]
+fn configured_plugins_from_stack_merges_user_layers() {
+    let stack = ConfigLayerStack::new(
+        vec![
+            user_layer(
+                user_config_path("config.toml"),
+                "[plugins.base]\nenabled = true\n",
```
bolinfest, line 33: Consider raw string? no strong preference.

codex-rs/core/src/config/config_tests.rs
```diff
@@ -5505,6 +5511,67 @@ async fn set_model_updates_defaults() -> anyhow::Result<()> {
...
+
+#[test]
+fn profile_v2_config_path_accepts_only_plain_names() -> anyhow::Result<()> {
+    let codex_home = TempDir::new()?;
+    assert_eq!(
+        resolve_profile_v2_config_path(codex_home.path(), "work")?,
+        codex_home.path().join("work.config.toml").abs()
+    );
+
+    for invalid in [
```
bolinfest: Again, I think this happens "naturally" if you prefer `ProfileV2Name` to `str` or `String` in API calls.

codex-rs/core/src/config/edit.rs
```diff
@@ -1093,24 +1101,38 @@ pub async fn apply(
     edits: Vec<ConfigEdit>,
 ) -> anyhow::Result<()> {
     let codex_home = codex_home.to_path_buf();
+    let config_path = codex_home.join(CONFIG_TOML_FILE);
     let profile = profile.map(ToOwned::to_owned);
-    task::spawn_blocking(move || apply_blocking(&codex_home, profile.as_deref(), &edits))
+    task::spawn_blocking(move || apply_blocking_to_path(&config_path, profile.as_deref(), &edits))
```
bolinfest: I'm confused: why is it named `apply_blocking_to_path()` where `config_path` is the `config.toml` but `profile` could also be set...which file are the edits applied to in that case? It feels like the function should get either `config.toml` or `PROFILE.config.toml`, no?

codex-rs/core/src/config/mod.rs
```diff
@@ -1281,19 +1282,43 @@ impl Config {
...
+) -> std::io::Result<AbsolutePathBuf> {
+    if profile_name.is_empty()
+        || !profile_name
+            .bytes()
+            .all(|byte| byte.is_ascii_alphanumeric() || matches!(byte, b'_' | b'-'))
```
bolinfest: Why not convert to `ProfileV2Name`?

codex-rs/core/src/session/mod.rs
```diff
@@ -1436,37 +1438,64 @@ impl Session {
...
+            let user_config = match std::fs::read_to_string(&config_toml_path) {
+                Ok(contents) => match toml::from_str::<toml::Value>(&contents) {
+                    Ok(config) => config,
+                    Err(err) => {
+                        warn!("failed to parse user config while reloading layer: {err}");
+                        return;
+                    }
+                },
+                Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
+                    toml::Value::Table(Default::default())
```
bolinfest, line 1522: could `continue;` instead or do we want the entry in the Vec?

codex-rs/config/src/state.rs
```diff
@@ -240,14 +262,54 @@ impl ConfigLayerStack {
...
+    /// Returns the active raw user config layer, if any.
     ///
     /// This does not merge other config layers or apply any requirements.
-    pub fn get_user_layer(&self) -> Option<&ConfigLayerEntry> {
+    pub fn get_active_user_layer(&self) -> Option<&ConfigLayerEntry> {
         self.user_layer_index
             .and_then(|index| self.layers.get(index))
     }
 
+    pub fn get_user_config_file(&self) -> Option<&AbsolutePathBuf> {
```
bolinfest, line 293: Docstring that this is the one with the profile if the profile is set?

codex-rs/exec/src/lib.rs
```diff
@@ -317,9 +319,19 @@ pub async fn run_main(cli: Cli, arg0_paths: Arg0DispatchPaths) -> anyhow::Result
             std::process::exit(1);
         }
     };
-
-    #[allow(clippy::print_stderr)]
+    let user_config_path = match config_profile_v2.as_deref() {
+        Some(profile_v2) => match resolve_profile_v2_config_path(&codex_home, profile_v2) {
+            Ok(path) => Some(path),
+            Err(err) => {
+                eprintln!("{err}");
```
bolinfest: Hmm, `#[allow(clippy::print_stderr)]` is used above, but not here?

bolinfest (approved):
Mostly small issues which you can take or leave, though one thing I do feel is important is using `ProfileV2Name` instead of `String` wherever appropriate.

I realize now that in some of my comments, `String` may be used because it is still referring to profile v1, so I guess those can't be converted right now. (Or maybe we do and run the risk of breaking folks who decided to create profiles in `config.toml` that have names with characters we don't allow, which I think is acceptable.)

jif-oai:
@codex review

PR #17142, opened 2026-04-08
[codex] Support remote exec cwd in TUI startup

pakrym-oai:
When running with remote executor the cwd is the remote path. Today we check for existence of a local directory on startup and attempt to load config from it.

For remote executors don't do that.


etraut-openai (approved):
Overall looks good. I left some comments for consideration.

fcoury-oai (approved):

PR #17163, opened 2026-04-08
[codex] Defer steering until after sampling the model post-compaction

jgershen-oai:
## Summary
- keep pending steered input buffered until the active user prompt has received a model response
- keep steering pending across auto-compact when there is real model/tool continuation to resume
- allow queued steering to follow compaction immediately when the prior model response was already final
- keep pending-input follow-up owned by `run_turn` instead of folding it into `SamplingRequestResult`
- add regression coverage for mid-turn compaction, final-response compaction, and compaction triggered before the next request after tool output

## Root Cause
Steered input was drained at the top of every `run_turn` loop. After auto-compaction, the loop continued and immediately appended any pending steer after the compact summary, making a queued prompt look like the newest task instead of letting the model first resume interrupted model/tool work.

## Implementation Notes
This patch keeps the follow-up signals separated:

- `SamplingRequestResult.needs_follow_up` means model/tool continuation is needed
- `sess.has_pending_input().await` means queued user steering exists
- `run_turn` computes the combined loop condition from those two signals

In `run_turn`:

```rust
let has_pending_input = sess.has_pending_input().await;
let needs_follow_up = model_needs_follow_up || has_pending_input;
```

After auto-compact we choose whether the next request may drain steering:

```rust
can_drain_pending_input = !model_needs_follow_up;
```

That means:

- model/tool continuation + pending steer: compact -> resume once without draining steer
- completed model answer + pending steer: compact -> drain/send the steer immediately
- fresh user prompt: do not drain steering before the model has answered the prompt once

The drain is still only `sess.get_pending_input().await`; when `can_drain_pending_input` is false, core uses an empty local vec and leaves the steer pending in session state.

## Validation
- PASS `cargo test -p codex-core --test all steered_user_input -- --nocapture`
- PASS `just fmt`
- PASS `git diff --check`
- NOT PASSING HERE `just fix -p codex-core` currently stops before linting this change on an unrelated mainline test-build error: `core/src/tools/spec_tests.rs` initializes `ToolsConfigParams` without `image_generation_tool_auth_allowed`; this PR does not touch that file.


aibrahim-oai (approved):
It is technically correct but it was hard to grasp. Thinking about why I feel that `can_drain_pending_input` was vague to reason about. I suggest we have an enum of the `last_sampling_type` and have `normal` and `compact` (feel free to change the naming). Then, it would be easier to reason about why we don't inject input.

PR #17170, opened 2026-04-08
Render statusline context as a meter

etraut-openai:
Problem: The statusline reported context as an “X% left” value, which could be mistaken for quota, and context usage was included in the default footer.

Solution: Render configured context status items as a filling context meter, preserve `context-used` as a legacy alias while hiding it from the setup menu, and remove context from the default statusline. It will still be available as an opt-in option for users who want to see it.

image


codex-rs/tui/src/bottom_pane/status_line_setup.rs
```diff
@@ -133,6 +130,31 @@ impl StatusLineItem {
...
+    StatusLineItem::CodexVersion,
+    StatusLineItem::ContextWindowSize,
+    StatusLineItem::UsedTokens,
+    StatusLineItem::TotalInputTokens,
+    StatusLineItem::TotalOutputTokens,
+    StatusLineItem::SessionId,
+    StatusLineItem::FastMode,
+];
+
+fn normalize_status_line_item_for_setup(item: StatusLineItem) -> StatusLineItem {
```
etraut-openai: Add a comment here explaining why this function is needed.

fcoury-oai:
BTW I really like this one :-)

fcoury-oai (approved):

PR #17171, opened 2026-04-08
Route Fed API-key logins with existing base-url config

jackz-oai:
Summary:
- Probe public OpenAI /v1/me during API-key login.
- If /v1/me reports the current org is FedRAMP, show a login prompt and write the existing openai_base_url config override to Gov API.
- Leave commercial API-key logins alone; this does not store base URLs in auth.json or change provider request construction.

Depends on:
- openai/openai#804108 for /v1/me current_organization_is_fedramp


viyatb-oai:
Security review note: I think the FedRAMP routing config write needs to happen before the API key is persisted, or the auth write needs to be rolled back if the config write fails.

In the current `run_login_with_api_key` flow, we detect `current_organization_is_fedramp`, then call `login_with_api_key(...)`, and only after that call `write_gov_openai_base_url(&config)`. If `write_gov_openai_base_url` fails, the command exits with an error but the API key has already been stored while `openai_base_url` may still be unset or left at the commercial/default value. A later Codex run could then pick up the saved API-key auth and route requests to the wrong API edge.

Could we either write the required `openai_base_url` before persisting the key, or delete/rollback the just-written auth state on the config-write failure path?

jackz-oai:
Addressed in `029bd9ba1` by removing the separate `openai_base_url` config mutation entirely.

The `/v1/me` result is now stored next to the API key in `auth.json` as `OPENAI_API_KEY_IS_FEDRAMP`, and runtime routing derives the gov base URL from that same auth record. That means there is no longer a split state where the key is persisted but the routing config write failed. A normal/non-Fed API-key login also clears any stale FedRAMP flag with the auth write.

PR #17187, opened 2026-04-09
Add thread title to configurable TUI status line

canvrno-oai:
- Add thread-title as an optional TUI status line item, omitted unless the user has set a custom name (`ChatWidget.thread_name`).
- Refresh the status line when threads are renamded
- Add snapshot coverage for renamed-thread footer behavior.


fcoury-oai (approved):
Tested the feature and it worked perfectly.

Code looks good as well, approved.

PR #17189, opened 2026-04-09
Emit live hook prompts before raw-event filtering

abhinav-oai:
# What
Project raw Stop-hook prompt response items into typed v2 hookPrompt item-completed notifications before applying the raw-response-event filter. Keep ordinary raw response items filtered for normal subscribers; only the existing hookPrompt bridge runs on the filtered raw-item path.

# Why
Blocked Stop hooks record their continuation instruction as a raw model-history user item. Normal v2 desktop subscribers do not opt into raw response events, so the app-server listener filtered that raw item before the existing hookPrompt translator could emit the typed live item/completed notification. As a result, the hook-prompt bubble only appeared after thread history was reloaded.

### DIFF
--- a/codex-rs/app-server/src/bespoke_event_handling.rs
+++ b/codex-rs/app-server/src/bespoke_event_handling.rs
@@ -2059,7 +2059,7 @@ async fn maybe_emit_raw_response_item_completed(
         .await;
 }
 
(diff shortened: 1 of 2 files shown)


abhinav-oai:
@codex review

eternal-openai (approved):

PR #17194, opened 2026-04-09
add parent-id to guardian context

won-openai:
adding parent codex session id to guardian prompt


codex-rs/core/src/guardian/prompt.rs
```diff
@@ -80,6 +80,10 @@ pub(crate) async fn build_guardian_prompt_items(
         });
     };
 
+    push_text(format!(
+        "Parent Codex session id: {}\n\n",
+        session.conversation_id
+    ));
```
won-openai: what would be the best way to push this inside the guardian prompt? wanted to be aligned on this before i make any "explanations" to the guardian in context

owenlin0 (approved):

PR #17210, opened 2026-04-09
Add output_schema to code mode render

vivi:
This updates code-mode tool rendering so MCP tools can surface structured output types from their `outputSchema`.

What changed:
- Detect MCP tool-call result wrappers from the output schema shape instead of relying on tool-name parsing or provenance flags.
- Render shared TypeScript aliases once for MCP tool results (`CallToolResult`, `ContentBlock`, etc.) so multiple MCP tool declarations stay compact.
- Type `structuredContent` from the tool definition's `outputSchema` instead of rendering it as `unknown`.
- Update the shared MCP aliases to match the MCP draft `CallToolResult` schema more closely.

Example:
- Before: `declare const tools: { mcp__rmcp__echo(args: { env_var?: string; message: string; }): Promise<{ _meta?: unknown; content: Array<unknown>; isError?: boolean; structuredContent?: unknown; }>; };`
- After: `declare const tools: { mcp__rmcp__echo(args: { env_var?: string; message: string; }): Promise<CallToolResult<{ echo: string; env: string | null; }>>; };`


codex-rs/code-mode/src/description.rs
```diff
@@ -324,10 +368,39 @@ fn render_code_mode_tool_declaration(
...
 }
 
+fn render_tool_heading(global_name: &str, raw_name: &str) -> String {
+    if global_name == raw_name {
+        format!("### `{global_name}`")
+    } else {
+        format!("### `{global_name}` (`{raw_name}`)")
+    }
+}
+
```
vivi, line 433: drive-by fix

codex-rs/code-mode/src/description.rs
```diff
@@ -185,8 +194,18 @@ pub fn build_exec_tool_description(
     if !enabled_tools.is_empty() {
         let mut current_namespace: Option<&str> = None;
         let mut nested_tool_sections = Vec::with_capacity(enabled_tools.len());
+        let mut shared_mcp_types_rendered = false;
 
-        for (name, nested_description) in enabled_tools {
+        for tool in enabled_tools {
+            let name = tool.name.as_str();
+            if tool.origin == ToolOrigin::Mcp && !shared_mcp_types_rendered {
```
pakrym-oai: nit: I'd move this out of the loop into a separate condition. if (got mcp tool) { prepend types).

codex-rs/code-mode/src/description.rs
```diff
@@ -300,12 +320,34 @@ fn append_code_mode_sample_for_definition(definition: &ToolDefinition) -> String
             .unwrap_or_else(|| "unknown".to_string()),
         CodeModeToolKind::Freeform => "string".to_string(),
     };
+    if definition.origin == ToolOrigin::Mcp {
```
pakrym-oai: nit: make only output_type calculation conditional

codex-rs/code-mode/src/description.rs
```diff
@@ -324,10 +366,65 @@ fn render_code_mode_tool_declaration(
...
+            .unwrap_or(&JsonValue::Bool(true)),
+    )
+}
+
+const MCP_TYPESCRIPT_PREAMBLE: &str = r#"type mcp_annotations = {
+  audience?: Array<"user" | "assistant">;
+  priority?: number;
+  lastModified?: string;
+};
+type mcp_resource =
```
pakrym-oai: nit: ts types are typically PascalCased

codex-rs/code-mode/src/description.rs
```diff
@@ -324,10 +366,65 @@ fn render_code_mode_tool_declaration(
...
+fn extract_mcp_structured_content_schema(output_schema: &JsonValue) -> Option<&JsonValue> {
+    let properties = output_schema.get("properties")?.as_object()?;
+    Some(
+        properties
+            .get("structuredContent")
+            .unwrap_or(&JsonValue::Bool(true)),
+    )
+}
+
+const MCP_TYPESCRIPT_PREAMBLE: &str = r#"type mcp_annotations = {
```
pakrym-oai: I trust you these are correct :)

codex-rs/code-mode/src/description.rs
```diff
@@ -722,11 +862,21 @@ mod tests {
             r#"## mcp__sample
 Shared namespace guidance.
 
-### `mcp__sample__alpha` (`mcp__sample__alpha`)
+### `mcp__sample__alpha`
 First tool
 
-### `mcp__sample__beta` (`mcp__sample__beta`)
-Second tool"#
+exec tool declaration:
```
pakrym-oai: I've heard complaints about `exec tool declaration:` phrasing. Up to you whether and when to change that.

codex-rs/tools/src/utility_tool_tests.rs
```diff
@@ -103,6 +104,7 @@ fn test_sync_tool_matches_expected_spec() {
                     ),
                 ]), /*required*/ None, Some(false.into())),
             output_schema: None,
+            origin: ToolOrigin::Native,
```
pakrym-oai: Is there a non-dirty way to avoid adding origin everywhere?

codex-rs/tools/src/tool_definition.rs
```diff
@@ -1,6 +1,23 @@
 use crate::JsonSchema;
 use serde_json::Value as JsonValue;
 
+/// Where a tool definition originated before it was normalized into the shared
+/// `ToolDefinition` shape.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum ToolOrigin {
```
pakrym-oai: If we keep it let's name it ToolKind

magus (approved):
Thanks for improving this typing!

pakrym-oai (approved):

codex-rs/tools/src/utility_tool_tests.rs
magus: unfortunately no, maybe the best we could do is define a default and update all these call sites to use that so in the future and for this we can omit the field we set by default but arguably being explicit might be best rather than falling back to a default everywhere

codex-rs/code-mode/src/description.rs
magus: I spot checked and found a delta here on the `resource_link` below
https://modelcontextprotocol.io/specification/draft/schema#resourcelink

Maybe we can share this page with codex and ask it to make sure the types are absolutely correct?

codex-rs/code-mode/src/description.rs
magus: I wouldn't change to pascal based on this but I would align to the name of the types as they appear in the spec so it's easy to match them up

So `mcp_output` should become `ContentBlock` for example

```ts
type ContentBlock = 
  | TextContent
  | ImageContent
  | AudioContent
  | ResourceLink
  | EmbeddedResource
```

https://modelcontextprotocol.io/specification/draft/schema#contentblock

codex-rs/tools/src/utility_tool_tests.rs
vivi: the alternative I'm about to push is removing ToolOrigin all rogether and checking outputschema (since its quite specific and has to adhere to the MCP spec)

vivi:
@codex review

PR #17222, opened 2026-04-09
feat: /resume per ID/name

jif-oai:
Support `/resume 00000-0000-0000-00000000` from the TUI (equivalent for the name)

### DIFF
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -47,6 +47,7 @@ use crate::read_session_model;
 use crate::render::highlight::highlight_bash_to_lines;
 use crate::render::renderable::Renderable;
 use crate::resume_picker::SessionSelection;
+use crate::resume_picker::SessionTarget;
 #[cfg(test)]
 use crate::test_support::PathBufExt;
 use crate::tui;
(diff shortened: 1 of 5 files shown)


codex-rs/tui/src/app.rs
```diff
@@ -4199,6 +4218,20 @@ impl App {
...
+            AppEvent::ResumeSessionByIdOrName(id_or_name) => {
+                match crate::lookup_session_target_with_app_server(app_server, &id_or_name).await? {
+                    Some(target_session) => {
+                        return self
+                            .resume_target_session(tui, app_server, target_session)
+                            .await;
+                    }
+                    None => {
+                        self.chat_widget.add_error_message(format!(
+                            "No saved chat found matching '{id_or_name}'."
```
fcoury-oai, line 4230: nit: if we run `/resume` the picker says "Resume a previous session". Should we also say "No saved session found matching '{id_or_name}'" here?

fcoury-oai: Update: I see the description of the `/resume` command says "resume a saved chat". I personally prefer using session in both instances mostly to make it consistent with the wording on the session picker, but open to your opinion here.

fcoury-oai (approved):
Code looks good! Optional and minor nit with the wording.

codex-rs/tui/src/app.rs
jif-oai: I think in the end we should rename everything to "thread"
I will follow-up with this

PR #17223, opened 2026-04-09
fix: MCP leaks in app-server

jif-oai:
The disconnect path now reuses the same teardown flow as explicit unsubscribe, and the thread-state bookkeeping consistently reports only threads that lost their last subscriber

https://github.com/openai/codex/issues/16895


owenlin0 (approved):

jif-oai:
@codex review

PR #17247, opened 2026-04-09
Make forked agent spawns keep parent model config

friel-openai:
## Summary

When a `spawn_agent` call does a full-history fork, keep the parent's effective agent type and model configuration instead of applying child role/model overrides.

This is the minimal config-inheritance slice of #16055. Prompt-cache key inheritance and MCP tool-surface stability are split into follow-up PRs.

## Design

- Reject `agent_type`, `model`, and `reasoning_effort` for v1 `fork_context` spawns.
- Reject `agent_type`, `model`, and `reasoning_effort` for v2 `fork_turns = "all"` spawns.
- Keep v2 partial-history forks (`fork_turns = "N"`) configurable; requested model/reasoning overrides and role config still apply there.
- Keep non-forked spawn behavior unchanged.

## Tests

- `cargo +1.93.1 test -p codex-core spawn_agent_fork_context --lib`
- `cargo +1.93.1 test -p codex-core multi_agent_v2_spawn_fork_turns --lib`
- `cargo +1.93.1 test -p codex-core multi_agent_v2_spawn_partial_fork_turns_allows_agent_type_override --lib`

PR #17248, opened 2026-04-09
Inherit forked agent prompt cache keys

friel-openai:
## Summary

Stacked on #17247.

When a thread is created by an actual forked agent spawn, reuse the parent thread's prompt-cache key instead of keying requests by the child thread id.

This keeps forked-agent requests on the same prompt-cache lineage as the parent while still giving each child its own conversation/session id and rollout.

## Design

- Add an explicit `prompt_cache_key` to `ModelClient` state.
- Thread `InheritedThreadState` through thread creation so forked children can receive parent-owned runtime state.
- Populate inherited cache state only for actual fork creation (`fork_context` / `fork_turns`).
- Do not opportunistically inherit cache state for resume or delegate paths; those are not deterministic fork-creation paths.
- Leave fresh spawns and context-free children keyed to their own thread id.

## Tests

- `cargo +1.93.1 test -p codex-core spawn_agent_can_fork_parent_thread_history_with_sanitized_items --lib`
- `cargo +1.93.1 test -p codex-core resume_thread_subagent_restores_stored_nickname_and_role --lib`
- `cargo +1.93.1 test -p codex-core client::tests::build_ws_client_metadata_includes_window_lineage_and_turn_metadata --lib`


codex-rs/core/src/agent/control.rs
```diff
@@ -1056,6 +1065,32 @@ impl AgentControl {
         ))
     }
 
+    async fn inherited_thread_state_for_source(
+        &self,
```
jif-oai: This method does not depend on self. This shouldn't be part of control

codex-rs/core/src/codex.rs
```diff
@@ -432,10 +432,16 @@ pub(crate) struct CodexSpawnArgs {
     pub(crate) metrics_service_name: Option<String>,
     pub(crate) inherited_shell_snapshot: Option<Arc<ShellSnapshot>>,
     pub(crate) inherited_exec_policy: Option<Arc<ExecPolicyManager>>,
+    pub(crate) inherited_thread_state: InheritedThreadState,
     pub(crate) user_shell_override: Option<shell::Shell>,
     pub(crate) parent_trace: Option<W3cTraceContext>,
 }
 
+#[derive(Clone, Copy, Debug, Default)]
+pub(crate) struct InheritedThreadState {
```
jif-oai: can we put this in a dedicated file and have a builder implementation on it? It will be useful in the future

codex-rs/core/src/agent/control.rs
jif-oai: Btw, would be much cleaner if it was returning an `Option<ThreadId>` and you use the builder defined in the comment below

codex-rs/core/src/codex_delegate.rs
```diff
@@ -94,6 +94,7 @@ pub(crate) async fn run_codex_thread_interactive(
         inherited_shell_snapshot: None,
         user_shell_override: None,
         inherited_exec_policy: Some(Arc::clone(&parent_session.services.exec_policy)),
+        inherited_thread_state: crate::codex::InheritedThreadState::default(),
```
jif-oai: nit: `Default::default()`

codex-rs/core/src/codex_tests_guardian.rs
```diff
@@ -452,6 +452,7 @@ async fn guardian_subagent_does_not_inherit_parent_exec_policy_rules() {
         metrics_service_name: None,
         inherited_shell_snapshot: None,
         inherited_exec_policy: Some(Arc::new(parent_exec_policy)),
+        inherited_thread_state: crate::codex::InheritedThreadState::default(),
```
jif-oai: same

codex-rs/core/src/thread_manager.rs
```diff
@@ -756,6 +757,7 @@ impl ThreadManagerState {
             /*metrics_service_name*/ None,
             /*inherited_shell_snapshot*/ None,
             /*inherited_exec_policy*/ None,
+            InheritedThreadState::default(),
```
jif-oai: same

jif-oai (commented):
I think there is a strange behaviour here. For every non-fork, their cache key become of a random UUID... it means we loose all tracing for cache keys

PR #17249, opened 2026-04-09
adding parent_thread_id in guardian

won-openai:
## Summary

This PR adds the parent conversation/session id to the subagent-start analytics event for Guardian subagents.

Previously, Guardian sessions were emitted as subagent thread-initialized events, but their `parent_thread_id` was serialized as `null`. After this change, the `codex_thread_initialized` analytics event for a Guardian child session includes the parent user conversation id.


codex-rs/core/src/codex_delegate.rs
```diff
@@ -100,10 +101,17 @@ pub(crate) async fn run_codex_thread_interactive(
     if parent_session.enabled(codex_features::Feature::GeneralAnalytics) {
         let thread_config = codex.thread_config_snapshot().await;
         let client_metadata = parent_session.app_server_client_metadata().await;
+        let parent_thread_id = match &subagent_source {
+            SubAgentSource::Other(name) if name == GUARDIAN_REVIEWER_NAME => {
```
owenlin0: let's always populate it if `parent_session.conversation_id` is set? seems generally useful

owenlin0 (approved):

PR #17258, opened 2026-04-09
Omit empty app-server instruction overrides

aibrahim-oai:
## Summary
- omit serialized Responses instructions when an app-server base instruction override is empty
- skip empty developer instruction messages and add v2 coverage for the empty-override request shape

## Validation
- just fmt
- git diff --check


pakrym-oai (approved):

dylan-hurd-oai (commented):
LGTM

PR #17263, opened 2026-04-09
preserve search results order in tool_search_output

sayan-oai:
we used to alpha-sort tool search results because we were using `BTreeMap`, which threw away the actual search result ordering.

Now we use a vec to preserve it.

### Tests
Updated tests

### DIFF
--- a/codex-rs/tools/src/tool_discovery.rs
+++ b/codex-rs/tools/src/tool_discovery.rs
@@ -204,16 +204,17 @@ pub fn create_tool_search_tool(
 pub fn collect_tool_search_output_tools<'a>(
     tool_sources: impl IntoIterator<Item = ToolSearchResultSource<'a>>,
 ) -> Result<Vec<ToolSearchOutputTool>, serde_json::Error> {
-    let grouped = tool_sources.into_iter().fold(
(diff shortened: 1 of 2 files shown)


codex-rs/tools/src/tool_discovery.rs
```diff
@@ -204,16 +204,17 @@ pub fn create_tool_search_tool(
 pub fn collect_tool_search_output_tools<'a>(
     tool_sources: impl IntoIterator<Item = ToolSearchResultSource<'a>>,
 ) -> Result<Vec<ToolSearchOutputTool>, serde_json::Error> {
-    let grouped = tool_sources.into_iter().fold(
-        BTreeMap::<String, Vec<(String, ToolSearchResultSource<'a>)>>::new(),
```
sayan-oai, line 208: @mzeng-openai i saw the owned tuple was added in #16944. do we need it? it seems like the tuple is effectively `(tool.tool_name.to_string(), tool)`, but `tool.tool_name` is already on `ToolSearchResultSource`, so `tool.0 == `tool.1.tool_name`.

aibrahim-oai (approved):

PR #17266, opened 2026-04-09
[codex] Improve hook status rendering

abhinav-oai:
# Motivation

Make hook display less noisy and more useful by keeping transient hook activity out of permanent history unless there is useful output

Also addresses some of the concerns in https://github.com/openai/codex/issues/15497

# Changes

## Hook Lifecycle Rendering
- Hook start events no longer write permanent history rows like `Running PreToolUse hook`.
- Running hooks now render in a dedicated live hook area above the composer. It's similar to the active cell we use for tool calls but its a separate lane.
- Running hook rows use the existing animation setting.

## Hook Reveal Timing
- We wait 300ms before showing running hook rows and linger for up to 600ms once visible.
  - This is so fast hooks don't flash a transient `Running hook` row before user can read it every time.
- If a fast hook completes with meaningful output, only the completed hook result is written to history.
- If a fast hook completes successfully with no output, it leaves no visible trace.

## Completed Hook Output
- Completed hooks with output are sticky, for example `• SessionStart hook (completed)`.
- Hook output entries are rendered under that row with stable prefixes: `warning:`, `stop:`, `feedback:`, `hook context:`, and `error:`.
- Blocked hooks show feedback entries, for example `• PreToolUse hook (blocked)` followed by `feedback: ...`.
- Failed hooks show error entries, for example `• PostToolUse hook (failed)` followed by `error: ...`.
- Stopped hooks show stop entries and remain visually treated as non-success.

## Parallel Hook Behavior
- Multiple simultaneously running hooks can be tracked in one live hook cell.
- Adjacent running hooks with the same hook event name and same status message collapse into a count, for example `• Running 3 PreToolUse hooks: checking command policy`.
- Running hooks with different event names or different status messages remain separate rows.

## Hook Run Identity
- `PreToolUse` and `PostToolUse` hook run IDs now include the tool call ID which prevents concurrent tool-use hooks from sharing a run ID and clobbering each other in the UI.
- This ID scoping applies to tool-use hooks only; other hook event types keep their existing run identity behavior.

## App-Server Hook Notifications
- App-server `HookStarted` and `HookCompleted` notifications use the same live hook rendering path as core hook events.
- `UserPromptSubmit` hook notifications now render through the same completed hook output format, including warning and stop entries.


abhinav-oai:
@codex review

abhinav-oai:
@codex review

abhinav-oai:
@codex review

codex-rs/tui/src/chatwidget.rs
```diff
@@ -10906,8 +10935,16 @@ impl ChatWidget {
     /// should pass the same width the overlay uses; using a different width will cause wrapping
     /// mismatches between the main viewport and the transcript overlay.
     pub(crate) fn active_cell_transcript_lines(&self, width: u16) -> Option<Vec<Line<'static>>> {
-        let cell = self.active_cell.as_ref()?;
-        let lines = cell.transcript_lines(width);
+        let mut lines = Vec::new();
+        if let Some(cell) = self.active_cell.as_ref() {
+            lines.extend(cell.transcript_lines(width));
+        }
+        if let Some(hook_cell) = self.active_hook_cell.as_ref() {
```
fcoury-oai, line 10942: With the help of Codex, I was able to reproduce a case where a hook executes quietly and leaves no history row.

To replicate it, create a `PostToolUse` hook that drains stdin, writes an out-of-band breadcrumb to a file, sleeps for 250ms, and then exits successfully with code 0:

```json
{
  "hooks": {
    "PostToolUse": [
      {
        "matcher": "Bash",
        "hooks": [
          {
            "type": "command",
            "command": "python3 -c 'import pathlib, sys, time; payload = sys.stdin.read(); pathlib.Path(\"/tmp/codex-post-tool-use-
hook.log\").open(\"a\").write(f\"{time.time()} bytes={len(payload)}\\n\"); time.sleep(0.25)'",
            "statusMessage": "checking output policy"
          }
        ]
      }
    ]
  }
}
```

I then started a session and prompted explore this repo. The log showed multiple hook invocations, but the UI had no PostToolUse hook (...) rows, which is the intended
quiet-success behavior.

That means active_hook_cell can exist even when the hook has no visible lines to render. Maybe we could we compute the hook lines first, and only add the separator if those lines
are non-empty?

fcoury-oai (changes requested):
The code looks good, I was able to test it and it works great, aside from one finding below and the one that Codex already pin pointed.

fcoury-oai (approved):
Looks good! Approved.

PR #17271, opened 2026-04-10
fix: fix stale proxy env restoration after shell snapshots

viyatb-oai:
## Summary

This fixes a stale-environment path in shell snapshot restoration. A sandboxed command can source a shell snapshot that was captured while an older proxy process was running. If that proxy has died and come back on a different port, the snapshot can otherwise put old proxy values back into the command environment, which is how tools like `pip` end up talking to a dead proxy.

The wrapper now captures the live process environment before sourcing the snapshot and then restores or clears every proxy env var from the proxy crate's canonical list. That makes proxy state after shell snapshot restoration match the current command environment, rather than whatever proxy values happened to be present in the snapshot. On macOS, the Codex-generated `GIT_SSH_COMMAND` is refreshed when the SOCKS listener changes, while custom SSH wrappers are still left alone.


viyatb-oai:
@codex review

codex-rs/core/src/tools/runtimes/mod.rs
```diff
@@ -126,10 +131,35 @@ pub(crate) fn maybe_wrap_shell_lc_with_snapshot(
...
+        .filter(|key| is_valid_shell_variable_name(key))
+        .collect::<Vec<_>>();
+    keys.sort_unstable();
+    keys.dedup();
+
+    let (captures, restores) =
+        build_override_exports_for_keys("__CODEX_SNAPSHOT_PROXY_OVERRIDE", &keys);
+    let key = ALLOW_LOCAL_BINDING_ENV_KEY;
+    (
+        format!(
```
rreichel3-oai, line 168: Is there any better way to do this? It feels iffy

codex-rs/network-proxy/src/proxy.rs
```diff
@@ -364,6 +364,44 @@ pub const PROXY_URL_ENV_KEYS: &[&str] = &[
 
 pub const ALL_PROXY_ENV_KEYS: &[&str] = &["ALL_PROXY", "all_proxy"];
 pub const ALLOW_LOCAL_BINDING_ENV_KEY: &str = "CODEX_NETWORK_ALLOW_LOCAL_BINDING";
+const ELECTRON_GET_USE_PROXY_ENV_KEY: &str = "ELECTRON_GET_USE_PROXY";
+const GIT_SSH_COMMAND_ENV_KEY: &str = "GIT_SSH_COMMAND";
```
rreichel3-oai, line 370: This might override a user's existing GIT_SSH_COMMAND.. Is there a path to us appending if the command is already set?

rreichel3-oai (approved):

won-openai:
@codex review

codex-rs/core/src/tools/runtimes/mod.rs
```diff
@@ -103,6 +108,9 @@ pub(crate) fn maybe_wrap_shell_lc_with_snapshot(
         override_env.insert(CODEX_THREAD_ID_ENV_VAR.to_string(), thread_id.clone());
     }
     let (override_captures, override_exports) = build_override_exports(&override_env);
+    let (proxy_captures, proxy_exports) = build_proxy_env_exports();
```
jif-oai, line 111: This seems to rewrite proxy env for every shell snapshot, even when managed network is not active. If a user’s shell profile exports `HTTP_PROXY` and Codex itself doesn’t have it, this will unset it after sourcing the snapshot. Can we gate this on the managed proxy being active, or otherwise distinguish Codex proxy values from user proxy config?

codex-rs/core/src/tools/runtimes/mod.rs
```diff
@@ -119,10 +127,51 @@ pub(crate) fn maybe_wrap_shell_lc_with_snapshot(
...
+#[cfg(target_os = "macos")]
+fn build_codex_proxy_git_ssh_command_exports() -> (String, String) {
+    let key = PROXY_GIT_SSH_COMMAND_ENV_KEY;
+    let marker_pattern = format!("{}\\ *", CODEX_PROXY_GIT_SSH_COMMAND_MARKER.trim_end());
+    (
+        format!(
+            "__CODEX_SNAPSHOT_PROXY_GIT_SSH_COMMAND=\"${{{key}-}}\"\ncase \"$__CODEX_SNAPSHOT_PROXY_GIT_SSH_COMMAND\" in\n  {marker_pattern}) __CODEX_SNAPSHOT_PROXY_GIT_SSH_COMMAND_SET=1 ;;\n  *) __CODEX_SNAPSHOT_PROXY_GIT_SSH_COMMAND_SET= ;;\nesac"
+        ),
+        format!(
+            "if [ -n \"$__CODEX_SNAPSHOT_PROXY_GIT_SSH_COMMAND_SET\" ]; then\n  case \"${{{key}-}}\" in\n    {marker_pattern}) export {key}=\"$__CODEX_SNAPSHOT_PROXY_GIT_SSH_COMMAND\" ;;\n    *) if [ -z \"${{{key}+x}}\" ]; then export {key}=\"$__CODEX_SNAPSHOT_PROXY_GIT_SSH_COMMAND\"; fi ;;\n  esac\nfi"
```
jif-oai: This only fixes `GIT_SSH_COMMAND` when the live env already has a marked Codex value. If the snapshot has a marked stale command but the current run has no `SOCKS GIT_SSH_COMMAND`, we leave the stale snapshot value in place. Is it on purpose?

jif-oai (approved):
Ok after my comments

jif-oai:
@codex review

codex-rs/network-proxy/src/proxy.rs
```diff
@@ -475,9 +538,17 @@ fn apply_proxy_env_overrides(
     #[cfg(target_os = "macos")]
     if socks_enabled {
         // Preserve existing SSH wrappers (for example: Secretive/Teleport setups)
-        // and only provide a SOCKS ProxyCommand fallback when one is not present.
-        env.entry("GIT_SSH_COMMAND".to_string())
-            .or_insert_with(|| format!("ssh -o ProxyCommand='nc -X 5 -x {socks_addr} %h %p'"));
+        // but refresh a previously injected Codex fallback so it cannot point
+        // at a stale proxy port after the proxy is restarted.
+        match env.get(GIT_SSH_COMMAND_ENV_KEY) {
+            Some(command) if !is_codex_proxy_git_ssh_command(command) => {}
```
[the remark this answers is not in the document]
viyatb-oai: I am intentionally leaving unmarked proxy-shaped GIT_SSH_COMMAND values alone. They are indistinguishable from user-provided wrappers, and refreshing them would reintroduce the override risk called out above. Codex-owned commands are marked going forward, and those are the only ones refreshed automatically.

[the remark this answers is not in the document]
viyatb-oai: I am intentionally leaving unmarked proxy-shaped GIT_SSH_COMMAND values alone. They are indistinguishable from user-provided wrappers, and refreshing them would reintroduce the override risk called out above. Codex-owned commands are marked going forward, and those are the only ones refreshed automatically.

PR #17272, opened 2026-04-10
Add commandSummary metadata to exec command events

fc-oai:
## Summary
- Add optional commandSummary (present/past short prose) to exec_command tool args.
- Thread it through core unified exec begin/end events.
- Surface it in app-server commandExecution item schema and item notifications.
- Preserve replay/compatibility plumbing in TUI event adapters and adjust schema/README.

Note: this intentionally does not change CLI UX.


fc-oai:

PR #17273, opened 2026-04-10
feat: use scoped remote control server tokens

viyatb-oai:
## Summary

Updates the app-server remote-control client to consume the new short-lived server-scoped token returned by backend enrollment.

- Parse optional `remote_control_token`, `expires_at`, and `scopes` from `/remote/control/server/enroll` for compatibility with old backends.
- Keep the server-scoped token in memory only; persisted enrollment metadata still contains only server/environment identity.
- Prefer `Authorization: Bearer <remote_control_token>` on the remote-control server websocket when the backend provides one.
- Fall back to the legacy account bearer token when talking to an older backend.
- Re-enroll near token expiry and after 401/403 failures when using the scoped token.
- Add remote-control transport tests for token parsing, expiry handling, and Authorization header selection.

## Scope

This is the client half of the server-authority token split. It prevents the app-server websocket from depending on a reusable account bearer token once the backend starts returning scoped server tokens.

### DIFF
--- a/codex-rs/app-server/src/transport/remote_control/protocol.rs
+++ b/codex-rs/app-server/src/transport/remote_control/protocol.rs
@@ -19,6 +19,10 @@ pub(super) struct EnrollRemoteServerRequest {
     pub(super) os: &'static str,
     pub(super) arch: &'static str,
     pub(super) app_server_version: &'static str,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(super) server_id: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(super) environment_id: Option<String>,
 }
 
 #[derive(Debug, Deserialize)]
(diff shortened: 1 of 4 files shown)


codex-rs/app-server/src/transport/remote_control/protocol.rs
```diff
@@ -19,12 +19,22 @@ pub(super) struct EnrollRemoteServerRequest {
...
 #[derive(Debug, Deserialize)]
 pub(super) struct EnrollRemoteServerResponse {
     pub(super) server_id: String,
     pub(super) environment_id: String,
+    #[serde(default)]
+    pub(super) remote_control_token: Option<String>,
+    #[serde(default)]
+    pub(super) expires_at: Option<String>,
+    #[serde(default)]
+    pub(super) scopes: Option<Vec<String>>,
```
euroelessar: any reason not to make them required, assuming we can deploy the backend before merging this change?

PR #17282, opened 2026-04-10
Fix agent-job fanout fill and post-completion hangs

daveaitel-openai:
## Summary

This PR fixes the Codex-side failure modes that were preventing `spawn_agents_on_csv` from behaving like a real high-concurrency fanout primitive and from returning control cleanly to callers.

It now addresses six linked problems:

1. CSV fanout startup underfilled badly because worker creation was serialized.
2. Scheduler slots stayed occupied after rows were already completed in SQLite.
3. Agent-job shutdown could lose track of live worker threads during teardown.
4. `turn/completed` notifications could arrive without the terminal turn items that `exec` needed for clean shutdown.
5. `codex exec` could hang forever after the root turn completed because it performed an unbounded `thread/read` backfill on the same backpressured in-process event pipeline.
6. Under shared `agents.max_threads` pressure, `AgentLimitReached` could turn into a hot retry loop against `spawn_agent`, SQLite, and logging instead of backing off and waiting for capacity.

## User-visible impact

Before this PR:

- `spawn_agents_on_csv(max_concurrency=64)` could sustain only a small fraction of the requested concurrency.
- Rows could be marked completed in state while the scheduler still treated their workers as busy.
- Some agent-job runs finished the underlying work but failed to return control promptly.
- End-to-end wrappers such as `security_audit` could hang after `task_complete`, never producing their final report files.
- Under shared-capacity contention, fanout could churn locally instead of waiting for slots to free.

After this PR:

- worker startup fills toward the requested concurrency instead of serially trickling
- scheduler slots are reclaimed when the DB says rows are no longer running
- agent-job teardown preserves tracking long enough for later cleanup to reap workers correctly
- app-server includes terminal turn items in `turn/completed`
- post-turn `exec` shutdown no longer blocks forever on a backfill `thread/read`
- fanout backs off instead of hot-retrying when global agent slots are full
- the Firefox `security_audit` repros now complete end-to-end and write both markdown and HTML reports

### 1. Serialized worker bring-up in agent jobs

`spawn_agents_on_csv` used to refill by looping pending items and `await`ing a full `spawn_agent(...)` for each one before moving to the next. Because `spawn_agent(...)` waits through child session bring-up until `SessionConfigured`, refill throughput was limited by worker startup latency rather than by the requested concurrency.

### 2. DB completion and scheduler occupancy diverged

`report_agent_job_result` committed row completion immediately, but the in-memory scheduler only freed the corresponding slot later when the worker session itself reached a terminal state. That let rows be completed in SQLite while still consuming fanout capacity.

### 3. Teardown untracked workers too early

The slot-reclamation shutdown path released scheduler capacity by calling a shutdown helper that also removed the worker from `ThreadManager` immediately. That made capacity accounting look better, but it could leave a still-running worker no longer visible to later cleanup.

### 4. Terminal turn payloads were not reliably attached to `turn/completed`

`exec` wants the final turn items when it handles `TurnCompleted`. In large fanout runs, app-server could emit `turn/completed` with an empty terminal snapshot, which forced `exec` into a late `thread/read` backfill during shutdown.

### 5. `exec` did an unbounded backfill request after `TurnCompleted`

When `exec` receives `TurnCompleted`, it backfills missing turn items with a `thread/read`. In large fanout runs, the in-process event stream can still be under heavy backpressure. Because the backfill request shared the same bounded pipeline, `exec` could wait forever on `thread/read` while the worker was blocked trying to forward must-deliver completion events.

### 6. `AgentLimitReached` counted as “progress” instead of “wait”

The parallel startup path requeued blocked items back to `Pending`, but that still looked like scheduler progress to the main loop. Under shared `agents.max_threads` contention, the loop could immediately try the same refill again instead of waiting for a real worker/status change.

### Parallel startup and stronger refill behavior

- Parallelized CSV worker startup in `codex-rs/core/src/tools/handlers/agent_jobs.rs` and `codex-rs/core/src/tools/handlers/agent_jobs_startup.rs`.
- Added tracking for both `active_items` and `starting_items` so refill now targets `active + starting == max_concurrency`.
- Startup now claims rows in state first, then completes assignment when the worker actually comes up.
- Added startup instrumentation so logs show queued launches, startup latency, and effective concurrency.

### Scheduler slot reclamation and terminal-state reconciliation

- Added explicit slot reconciliation in `codex-rs/core/src/tools/handlers/agent_jobs_slots.rs`.
- Refill now trusts terminal DB state instead of waiting only on in-memory worker finalization.
- Workers that report successfully no longer hold scheduler slots hostage while they continue draining.
- Added logging that shows both DB and scheduler views of running work.

### Correctness fixes around shutdown and cancellation

- Preserved thread tracking during shutdown in `codex-rs/core/src/agent/control.rs` so releasing a slot does not also hide a still-live worker from later teardown.
- Added coverage for the preserved-tracking shutdown behavior in `codex-rs/core/src/agent/control_tests.rs`.
- Made result reporting and `stop=true` cancellation state updates atomic in `codex-rs/state/src/runtime/agent_jobs.rs`.
- Kept agent-job semantics intact around state assignment, failure handling, and max-thread limits.

### Turn-completion and exec-shutdown fixes

- App-server now retains the terminal turn snapshot and attaches it directly to `turn/completed` instead of relying on a late best-effort read.
- `exec` still keeps a bounded fallback `thread/read`, but it now times out and continues shutdown instead of hanging forever.
- Added targeted app-server / protocol / exec coverage for the turn-completion payload and bounded backfill behavior.

### Backoff when global agent slots are exhausted

- Startup draining now returns structured state (`progressed` plus `agent_limit_reached`) instead of only a boolean.
- The main agent-job loop now treats `AgentLimitReached` as a blocked refill condition.
- When fanout is locally ready to launch but global `agents.max_threads` is full, the loop stops relaunching immediately and waits for a startup or status change before retrying.
- Added targeted tests for both the wait decision and the `AgentLimitReached -> Pending` requeue path.

### Automated validation

Ran the following locally during development:

- `cargo test -p codex-state report_agent_job_item_result_and_cancel_job_is_atomic`
- `cargo test -p codex-core spawn_agents_on_csv -- --nocapture`
- `cargo test -p codex-core request_live_agent_shutdown_preserving_thread_releases_slot_without_untracking -- --nocapture`
- `cargo test -p codex-core should_wait_when_agent_limit_blocks_refill`
- `cargo test -p codex-core drain_ready_startups_reports_agent_limit_and_requeues_item`
- `cargo test -p codex-exec`
- `cargo test -p codex-app-server test_handle_turn_complete_emits_completed_without_error`
- `just fmt`
- `just fix -p codex-core`
- `just fix -p codex-state`
- `just fix -p codex-exec`

Per repo guidance, I did not run the full workspace `cargo test` without explicit approval.

### Synthetic fanout validation

The synthetic CSV fanout repro reached the requested concurrency instead of serially trickling worker creation. Startup logs showed fanout filling to the target and sustaining materially higher live-worker counts than before.

### End-to-end Firefox validation

Re-ran the real `security_audit` Firefox repros against this branch’s `target/debug/codex` binary.

Successful runs included:

- `netwerk/base`
  - `337` rows processed
  - `0` failures
  - generated `assess_output.csv`, `report.md`, and `report.html`
- `ipc/glue`
  - `162` rows processed
  - `0` failures
  - generated `assess_output.csv`, `report.md`, and `report.html`

These runs returned control to the shell instead of hanging after `task_complete`.

## Remaining caveat

Heavy fanout still produces `in-process app-server event stream lagged; dropped ... events` warnings under load. Those warnings reflect real backpressure, but the critical completion path is now robust enough that the run finishes end-to-end instead of hanging.

## Commit structure

This branch now contains five commits:

- `ab9f9848c7` Parallelize CSV agent job startup
- `dac04fba7e` Reclaim agent-job fanout slots after report
- `1f57fd56fa` Finish agent-job fanout completion and exec shutdown
- `ff98e9ac70` Include terminal items in turn completion events
- `4861584e9d` Back off CSV fanout when agent slots are full


codex-rs/core/src/tools/handlers/agent_jobs_startup.rs
```diff
@@ -0,0 +1,599 @@
...
+        .collect();
+    if stale_task_ids.is_empty() {
+        return Ok(false);
+    }
+
+    for task_id in stale_task_ids {
+        let Some(item) = startup_tasks.launching_items.remove(&task_id) else {
+            continue;
+        };
+        item.abort_handle.abort();
```
jif-oai, line 295: This abort can cancel `spawn_agent` after it has already registered a live thread but before th `ThreadId` is returned here, leaving us with a failed DB item and no way to shut down the worker. Can we avoid aborting the spawn future, or make the spawn path cancellation-safe before marking the item failed?

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -1932,19 +1932,29 @@ struct TurnCompletionMetadata {
...
+        completed_at: None,
+        duration_ms: None,
+    });
     let notification = TurnCompletedNotification {
         thread_id: conversation_id.to_string(),
         turn: Turn {
             id: event_turn_id,
-            items: vec![],
-            error: turn_completion_metadata.error,
+            items: turn.items,
```
jif-oai, line 2043: This attaches all turn.items to turn/completed. That fixes missing terminal data for exec, but it also turns the guaranteed terminal notification into a full replay of potentially large command/tool outputs that were already streamed as item events. Not ideal

jif-oai (approved):
This is a lot of code so I hope I didn't miss anything but globally lgtm after my 2 comments

jif-oai:
@codex review

jif-oai:
@codex review

jif-oai:
For the record, this in hold while we figure out the split

PR #17287, opened 2026-04-10
[code mode] defer mcp tools from exec description

sayan-oai:
## Summary
- hide deferred MCP/app nested tool descriptions from the `exec` prompt in code mode
- add short guidance that omitted nested tools are still available through `ALL_TOOLS`
- cover the code_mode_only path with an integration test that discovers and calls a deferred app tool

## Motivation
`code_mode_only` exposes only top-level `exec`/`wait`, but the `exec` description could still include a large nested-tool reference. This keeps deferred nested tools callable while avoiding that prompt bloat.

## Tests
- `just fmt`
- `just fix -p codex-code-mode`
- `just fix -p codex-tools`
- `cargo test -p codex-code-mode exec_description_mentions_deferred_nested_tools_when_available`
- `cargo test -p codex-tools create_code_mode_tool_matches_expected_spec`
- `cargo test -p codex-core code_mode_only_guides_all_tools_search_and_calls_deferred_app_tools`


codex-rs/code-mode/src/description.rs
```diff
@@ -8,6 +8,9 @@ use crate::PUBLIC_TOOL_NAME;
 const MAX_JS_SAFE_INTEGER: u64 = (1_u64 << 53) - 1;
 const CODE_MODE_ONLY_PREFACE: &str =
     "Use `exec/wait` tool to run all other tools, do not attempt to use any other tools directly";
+const DEFERRED_NESTED_TOOLS_GUIDANCE: &str = r#"Some nested MCP/app tools may be omitted from this description. They are still available on the global `tools` object and listed in `ALL_TOOLS`.
+To find one, filter `ALL_TOOLS` by `name` and `description`; do not print the full `ALL_TOOLS` array. Print only a small set of relevant matches if you need to inspect them.
+Once selected, call it with `await tools[match.name](...)`."#;
```
pakrym-oai: I don't think we need to specify this calling style, just tools.name would work.

codex-rs/core/tests/suite/code_mode.rs
```diff
@@ -314,6 +317,93 @@ async fn code_mode_only_restricts_prompt_tools() -> Result<()> {
     Ok(())
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn code_mode_only_prompt_guides_all_tools_search_for_deferred_app_tools() -> Result<()> {
```
pakrym-oai: can you add a test that actually calls one of these tools? What happens to the namespace how?

codex-rs/core/src/tools/code_mode/mod.rs
```diff
@@ -283,28 +283,33 @@ async fn call_nested_tool(
...
+    let call_id = format!("{PUBLIC_TOOL_NAME}-{}", uuid::Uuid::new_v4());
+    let call = if let Some(mcp_tool_call) = tool_runtime.resolve_mcp_tool_call(&tool_name).await {
+        match serialize_function_tool_arguments(&tool_name, input) {
+            Ok(raw_arguments) => ToolCall {
+                tool_name: mcp_tool_call.dispatch_name,
+                call_id,
+                tool_namespace: mcp_tool_call.dispatch_namespace,
+                payload: ToolPayload::Mcp {
+                    server: mcp_tool_call.tool_info.server_name,
+                    tool: mcp_tool_call.tool_info.tool.name.to_string(),
```
sayan-oai: we have to do this because directly accessible mcp tools are keyed in the registry as `{qualified_tool_name}` (i.e. `mcp__ologs__get_logs`), but deferred mcp tools are keyed as `{namespace}:{name}` (i.e. `mcp__ologs__:get_logs`).

this is because responsesapi's tool_search returns namespaced tools, but that concept is not required for MCPs otherwise. since MCPs can now be discovered via `tool_search`, we have to support both, but right now they are not uniform and this is a headache for callers.

I think it'd make sense to register all MCPs with namespaces to simplify.

codex-rs/code-mode/src/description.rs
```diff
@@ -250,16 +262,19 @@ pub fn is_code_mode_nested_tool(tool_name: &str) -> bool {
 pub fn build_exec_tool_description(
     enabled_tools: &[ToolDefinition],
     namespace_descriptions: &BTreeMap<String, ToolNamespaceDescription>,
-    code_mode_only: bool,
+    options: ExecToolDescriptionOptions,
```
pakrym-oai: nit: use a param

pakrym-oai (approved):

PR #17298, opened 2026-04-10
fix(guardian, app-server): introduce guardian review ids

owenlin0:
## Description

This PR introduces `review_id` as the stable identifier for guardian reviews and exposes it in app-server `item/autoApprovalReview/started` and `item/autoApprovalReview/completed` events.

Internally, guardian rejection state is now keyed by `review_id` instead of the reviewed tool item ID. `target_item_id` is still included when a review maps to a concrete thread item, but it is no longer overloaded as the review lifecycle identifier.

## Motivation

We'd like to give users the ability to preempt a guardian review while it's running (approve or decline).

However, we can't implement the API that allows the user to override a running guardian review because we didn't have a unique `review_id` per guardian review. Using `target_item_id` is not correct since:
- with execve reviews, there can be multiple execve calls (and therefore guardian reviews) per shell command
- with network policy reviews, there is no target item ID

The PR that actually implements user overrides will use `review_id` as the stable identifier.


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -4555,6 +4556,22 @@ pub enum GuardianApprovalReviewStatus {
     Aborted,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+/// [UNSTABLE] Source that produced a terminal guardian approval review decision.
+pub enum AutoReviewDecisionSource {
+    Agent,
```
owenlin0, line 4564: We will be introducing `User` in a followup

dylan-hurd-oai (approved):

PR #17305, opened 2026-04-10
Add sorting/backwardsCursor to thread/list and new thread/turns/list api

ddr-oai:
To improve performance of UI loads from the app, add two main improvements:
1. The `thread/list` api now gets a `sortDirection` request field and a `backwardsCursor` to the response, which lets you paginate forwards and backwards from a window.  This lets you fetch the first few items to display immediately while you paginate to fill in history, then can paginate "backwards" on future loads to catch up with any changes since the last UI load without a full reload of the entire data set.
2. Added a new `thread/turns/list` api which also has sortDirection and backwardsCursor for the same behavior as `thread/list`, allowing you the same small-fetch for immediate display followed by background fill-in and resync catchup.


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3652,7 +3663,14 @@ impl CodexMessageProcessor {
                 thread
             })
             .collect();
-        let response = ThreadListResponse { data, next_cursor };
+        let backwards_cursor = data
```
jif-oai: backwards curose is not a safe high-watermark: thread ordering only has second precision plus a random UUID tiebreaker, so a later create/update in the same second with a lower UUID can sort after this first item and be skipped by the later sortDirection: "asc" catch-up. This will be even more likely with multi-agent

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3869,6 +3887,119 @@ impl CodexMessageProcessor {
...
+        let Some(rollout_path) = rollout_path.as_ref() else {
+            self.send_internal_error(
+                request_id,
+                format!("failed to locate rollout for thread {thread_uuid}"),
+            )
+            .await;
+            return;
+        };
+
+        match read_rollout_items_from_rollout(rollout_path).await {
```
jif-oai, line 4074: I think this paginated endpoint still reads the entire rollout and builds every turn before slicing, so thread/turns/list does not actually avoid the expensive full-history loads

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3869,6 +3887,119 @@ impl CodexMessageProcessor {
...
+            Ok(items) => {
+                let turns = build_turns_from_rollout_items(&items);
```
jif-oai: This path bypasses the normalization used by thread/read. For example, stale inProgress turns from rollout history can leak through thread/turns/list instead of being marked interrupted when the thread is no longer active (same for dead tool calls) 

In general, avoid using `from_rollout_items`

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3869,6 +3887,119 @@ impl CodexMessageProcessor {
...
+            Ok(id) => id,
+            Err(err) => {
+                self.send_invalid_request_error(request_id, format!("invalid thread id: {err}"))
+                    .await;
+                return;
+            }
+        };
+
+        let mut rollout_path =
+            match find_thread_path_by_id_str(&self.config.codex_home, &thread_uuid.to_string())
```
jif-oai: shouldn't we also have something for archived session? Not sure if this is intended

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -9341,6 +9479,95 @@ pub(crate) fn summary_to_thread(summary: ConversationSummary) -> Thread {
...
+            keyed_turns.reverse();
+            if let Some(anchor) = anchor {
+                keyed_turns.retain(|(index, _)| *index < anchor);
+            }
+        }
+    }
+
+    let more_turns_available = keyed_turns.len() > page_size;
+    keyed_turns.truncate(page_size);
+    let backwards_cursor = keyed_turns.first().map(|(index, _)| index.to_string());
```
jif-oai: The turns cursor is just a positional index, which is not stable across rollback/compaction/history rewriting. Use something stable instead

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -9341,6 +9479,95 @@ pub(crate) fn summary_to_thread(summary: ConversationSummary) -> Thread {
...
+    })
+}
+
+fn parse_thread_turns_cursor(cursor: &str, total_turns: usize) -> Result<usize, JSONRPCErrorError> {
+    let index = cursor.parse::<usize>().map_err(|_| JSONRPCErrorError {
+        code: INVALID_REQUEST_ERROR_CODE,
+        message: format!("invalid cursor: {cursor}"),
+        data: None,
+    })?;
+    if index >= total_turns {
```
jif-oai: What if the conversation shrink?

codex-rs/rollout/src/recorder.rs
```diff
@@ -667,6 +684,138 @@ fn truncate_fs_page(
...
+    default_provider: &str,
+    archived: bool,
+    search_term: Option<&str>,
+) -> std::io::Result<ThreadsPage> {
+    let mut all_items = Vec::new();
+    let mut scanned_files = 0usize;
+    let mut reached_scan_cap = false;
+    let mut page_cursor = None;
+    loop {
+        let page = list_threads_from_files_desc(
```
jif-oai, line 931: This looks extremely expensive to loop on the desc like this

codex-rs/rollout/src/recorder.rs
```diff
@@ -667,6 +684,138 @@ fn truncate_fs_page(
...
+        all_items.extend(page.items);
+        page_cursor = page.next_cursor;
+        if page_cursor.is_none() || reached_scan_cap {
+            break;
+        }
+    }
+
+    if let Some(search_term) = search_term {
+        all_items.retain(|item| {
+            item.first_user_message
```
jif-oai: Is it on purpose? We used to search on the title, not on the first user message

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -9341,6 +9479,95 @@ pub(crate) fn summary_to_thread(summary: ConversationSummary) -> Thread {
     }
 }
 
+fn thread_cursor_for_sort_key(thread: &Thread, sort_key: CoreThreadSortKey) -> Option<String> {
+    let timestamp = match sort_key {
+        CoreThreadSortKey::CreatedAt => thread.created_at,
+        CoreThreadSortKey::UpdatedAt => thread.updated_at,
+    };
+    let timestamp = DateTime::<Utc>::from_timestamp(timestamp, /*nsecs*/ 0)?;
+    Some(format!(
```
jif-oai: Ultra nit: it would be cleaner to serialize a nice little struct here

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3217,6 +3231,35 @@ pub struct ThreadReadResponse {
...
+    /// Optional turn pagination direction; defaults to descending.
+    #[ts(optional = nullable)]
+    pub sort_direction: Option<SortDirection>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ThreadTurnsListResponse {
+    pub data: Vec<Turn>,
```
jif-oai, line 3389: Could we document this? Mainly because this is not always populated if I get htis correctly

codex-rs/rollout/src/recorder.rs
```diff
@@ -667,6 +684,138 @@ fn truncate_fs_page(
...
+        next_cursor,
+        num_scanned_files: scanned_files,
+        reached_scan_cap,
+    })
+}
+
+fn thread_item_sort_key(
+    item: &ThreadItem,
+    sort_key: ThreadSortKey,
+) -> Option<(OffsetDateTime, uuid::Uuid)> {
```
jif-oai, line 1016: if this can return None, how do you sort on them? In theory, it will put them first in the sorting but this feels odd... shouldn't we reject those rows?

codex-rs/rollout/src/recorder.rs
```diff
@@ -667,6 +684,138 @@ fn truncate_fs_page(
...
+    search_term: Option<&str>,
+) -> std::io::Result<ThreadsPage> {
+    let mut all_items = Vec::new();
+    let mut scanned_files = 0usize;
+    let mut reached_scan_cap = false;
+    let mut page_cursor = None;
+    loop {
+        let page = list_threads_from_files_desc(
+            codex_home,
+            /*page_size*/ 256,
```
jif-oai: this should be a factor of page_size no?

codex-rs/rollout/src/recorder.rs
```diff
@@ -667,6 +684,138 @@ fn truncate_fs_page(
...
+    }
+
+    all_items.sort_by_key(|item| thread_item_sort_key(item, sort_key));
+    if let Some(cursor) = cursor {
+        let anchor = cursor.parts();
+        all_items
+            .retain(|item| thread_item_sort_key(item, sort_key).is_some_and(|key| key > anchor));
+    }
+
+    let more_matches_available = all_items.len() > page_size;
```
jif-oai: But you filtered on `all_items` 🤔 so you could have `more_matches_available=false` just because of a lot of filtering out while in practice there are still some matches

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3217,6 +3231,35 @@ pub struct ThreadReadResponse {
     pub thread: Thread,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ThreadTurnsListParams {
+    pub thread_id: String,
+    /// Opaque cursor to pass to the next call to continue after the last turn.
```
jif-oai, line 3374: Thanks! I love those opaque cursors

codex-rs/app-server/src/codex_message_processor.rs
ddr-oai: it's designed to optimize over-network communication, not on-disk :(

codex-rs/app-server/src/codex_message_processor.rs
ddr-oai: changed to unique millisecond timestamps and no uuid in cursor

codex-rs/app-server/src/codex_message_processor.rs
ddr-oai: fixed

codex-rs/app-server/src/codex_message_processor.rs
ddr-oai: fixed

codex-rs/app-server/src/codex_message_processor.rs
ddr-oai: changed shape of cursor so it's using id now -- without a useful index in sqlite for the thread turn data, might as well stick with an id since we're reading the whole thread's messages in anyway

codex-rs/app-server/src/codex_message_processor.rs
ddr-oai: since we switched to IDs for the cursor, if the cursor's anchor disappears, it just errors and the caller should start over again.  LMK if you think that doesn't work out

codex-rs/rollout/src/recorder.rs
ddr-oai: without having this data in an indexed thing we can scan through, not sure what else to do.  we're optimizing around the network transmission at the cost of local cpu, which feels like the right tradeoff given the not-that-absurd amount of data for the thread list.  wdyt?  and also we should ideally never hit this anymore.

codex-rs/rollout/src/recorder.rs
ddr-oai: fixed

codex-rs/app-server/src/codex_message_processor.rs
ddr-oai: changed how the cursors work to simplify it all

codex-rs/app-server-protocol/src/protocol/v2.rs
ddr-oai: done

codex-rs/rollout/src/recorder.rs
ddr-oai: fixed -- drops rows with none

codex-rs/rollout/src/recorder.rs
ddr-oai: fixed

codex-rs/rollout/src/recorder.rs
ddr-oai: fixed -- filters then sorts and truncates

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -9623,17 +9820,188 @@ pub(crate) fn summary_to_thread(
...
+    let anchor = cursor.map(parse_thread_turns_cursor).transpose()?;
+    let page_size = limit
+        .map(|value| value as usize)
+        .unwrap_or(THREAD_TURNS_DEFAULT_LIMIT)
+        .clamp(1, THREAD_TURNS_MAX_LIMIT);
+
+    let anchor_index = anchor
+        .as_ref()
+        .and_then(|anchor| turns.iter().position(|turn| turn.id == anchor.turn_id));
+    if anchor.is_some() && anchor_index.is_none() {
```
jif-oai, line 10018: Shouldn't we error?   If rollback/compaction removed the anchor turn, the client needs an invalid/stale cursor error so it can reload.

jif-oai (approved):
This will break previously issued cursor but I don't think we care

The comments mentionned before on stability over same `ms` cursors is still valid but it's out of scope for this PR

codex-rs/app-server/src/codex_message_processor.rs
ddr-oai: ah good catch. fixed.

PR #17346, opened 2026-04-10
feat(tui): generated thread names based on first message

fcoury-oai:
# Problem

New interactive sessions have a stable thread id but often no human-readable name until the user runs `/rename`. That makes terminal tabs and session pickers harder to scan, especially when several Codex sessions are running against the same project.

# Mental model

The user's first accepted prompt is enough context to propose a short default title. Codex treats that title exactly like any other thread name after it is accepted: it is normalized, appended through the session-index path, stored on the session, and broadcast as `ThreadNameUpdated`. Manual `/rename` remains authoritative; generation only fills an empty name.

# Non-goals

This does not rename exec sessions, sub-agent sessions, ephemeral sessions, or sessions that already have a thread name. It does not add a second title field or change the thread-name protocol. It also does not make the terminal title depend on auto-generation specifically; the TUI renders whatever current thread name the core reports.

# Tradeoffs

Title generation runs as a best-effort background model request so the first turn is not blocked on naming. That means the title can arrive after user-visible turn output starts, can fail silently except for debug logs, and can lose a race to `/rename`. The implementation chooses that latency tradeoff over delaying the agent response for a cosmetic name.

# Architecture

`[session_titles]` config resolves into `SessionTitlesConfig`, which can disable generation or select the title model. During a normal user turn, core records and materializes the first user message, checks whether the session source, persistence, config, existing thread name, and one-shot guard allow generation, and spawns a background title task. The task asks a small model for JSON containing `title`, sanitizes the result for display, and commits it with `IfUnset` semantics.

Thread-name commits are centralized in `commit_thread_name`. Both generated titles and `/rename` use the same persistence/event path; the generated path takes the conditional mode, while `/rename` takes the unconditional mode. The session has a dedicated thread-name write lock so the state check, session-index append, in-memory update, and notification stay ordered relative to other title writes.

The TUI default terminal-title selection now includes `thread` between `spinner` and `project`. The existing item renderer omits empty parts, so new sessions still show the project-only title until core emits a thread name.

# Observability

Auto-generation failure is intentionally non-fatal and is logged at debug level as `auto thread name generation skipped or failed`. Successful commits are observable through the existing `ThreadNameUpdated` event and through the appended `sessions_index.jsonl` thread-name record. For UI debugging, inspect the received `ThreadNameUpdated` event and the `last_terminal_title`/terminal OSC update path in the TUI.

# Tests

The branch adds config coverage for `[session_titles]`, core unit coverage for title sanitization and source gating, and TUI coverage for default terminal titles before a name exists, after a thread-name update, and after truncating an overlong name.

### DIFF
--- a/codex-rs/config/src/config_toml.rs
+++ b/codex-rs/config/src/config_toml.rs
@@ -19,6 +19,7 @@ use crate::types::OAuthCredentialsStoreMode;
 use crate::types::OtelConfigToml;
 use crate::types::PluginConfig;
 use crate::types::SandboxWorkspaceWrite;
+use crate::types::SessionTitlesToml;
 use crate::types::ShellEnvironmentPolicyToml;
 use crate::types::SkillsConfig;
 use crate::types::ToolSuggestConfig;
@@ -317,6 +318,9 @@ pub struct ConfigToml {
 
     /// Memories subsystem settings.
     pub memories: Option<MemoriesToml>,
+
+    /// Default generated names for interactive sessions.
+    pub session_titles: Option<SessionTitlesToml>,
 
     /// User-level skill config entries keyed by SKILL.md path.
     pub skills: Option<SkillsConfig>,
--- a/codex-rs/config/src/types.rs
+++ b/codex-rs/config/src/types.rs
@@ -200,6 +200,42 @@ pub struct MemoriesToml {
     pub consolidation_model: Option<String>,
 }
 
+/// Session title settings loaded from config.toml.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Default, JsonSchema)]
+#[schemars(deny_unknown_fields)]
+pub struct SessionTitlesToml {
+    /// When `false`, skip generating a default `/rename` title for new interactive sessions.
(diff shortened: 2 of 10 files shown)


sanogueralorenzo:
@fcoury-oai this PR looks very close to the behavior I’m looking for.

The key parts that seem especially right are:

- generating a default title from the first accepted user prompt
- storing it as the normal `thread_name`
- using the same persistence/event path as `/rename`
- letting manual rename remain authoritative
- aligning the title prompt and constraints with the Codex app

One question: is the title-generation prompt/model here still intended to match the current Codex macOS app behavior? If not, I think it would be valuable for CLI and macOS app to share the same title-generation logic as directly as possible, so session names remain consistent across surfaces.

Given that several thread-name infrastructure PRs have landed since this was closed as stale (`#21260`, `#21264`, `#18633`), is there still interest in reviving this approach or reopening a smaller version of it?

I’m happy to open a PR if that would help.

Keno:
I would also like this feature. It is one of the quality-of-life things I immediately missed coming from Claude Code.

PR #17371, opened 2026-04-10
app-server: add pipelined config rpc regression test

euroelessar:
### Summary
Adds regression coverage for pipelined config RPC reads after writes
### Testing
These are new tests


euroelessar:
@codex review

owenlin0 (approved):

PR #17372, opened 2026-04-10
app-server: prepare to run initialized rpcs concurrently

euroelessar:
## Summary

- Refactors `MessageProcessor` and per-connection session state so initialized service RPC handling can be moved into spawned tasks in a follow-up PR.
- Shares the processor and initialized session data with `Arc`/`OnceLock` instead of mutable borrowed connection state.
- Keeps initialized request handling synchronous in this PR; it does **not** call `tokio::spawn` for service RPCs yet.

## Testing

- `just fmt`
- `cargo test -p codex-app-server` *(fails on existing hardening gaps covered by #17375, #17376, and #17377; the pipelined config regression passed before the unrelated failures)*

### DIFF
--- a/codex-rs/app-server/src/lib.rs
+++ b/codex-rs/app-server/src/lib.rs
@@ -652,7 +652,7 @@ pub async fn run_main_with_transport(
             AuthManager::shared_from_config(&config, /*enable_codex_api_key_env*/ false);
         let cli_overrides: Vec<(String, TomlValue)> = cli_kv_overrides.clone();
         let loader_overrides = loader_overrides_for_config_api;
-        let processor = MessageProcessor::new(MessageProcessorArgs {
+        let processor = Arc::new(MessageProcessor::new(MessageProcessorArgs {
             outgoing: outgoing_message_sender,
             arg0_paths,
             config: Arc::new(config),
@@ -667,7 +667,7 @@ pub async fn run_main_with_transport(
             auth_manager,
             rpc_transport: analytics_rpc_transport(transport),
             remote_control_handle: Some(remote_control_handle),
-        });
+        }));
         let mut thread_created_rx = processor.thread_created_receiver();
         let mut running_turn_count_rx = processor.subscribe_running_assistant_turn_count();
         let mut connections = HashMap::<ConnectionId, ConnectionState>::new();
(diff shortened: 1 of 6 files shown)


codex-rs/app-server/src/message_processor.rs
```diff
@@ -525,149 +565,177 @@ impl MessageProcessor {
...
+                ),
+                None => (false, Vec::new()),
+            };
+            let ClientInfo {
+                name,
+                title: _title,
+                version,
+            } = params.client_info;
+            let originator = name.clone();
+            if let Err(error) = set_default_originator(originator.clone()) {
```
jif-oai: These init side effects still run before the `OnceLock` commit,. So a losing init can mutate process-global client metadata even though it doesn't win session initialization

codex-rs/app-server/src/lib.rs
```diff
@@ -763,23 +763,28 @@ pub async fn run_main_with_transport(
...
                                         processor
                                             .process_request(
                                                 connection_id,
                                                 request,
                                                 transport,
-                                                &mut connection_state.session,
+                                                Arc::clone(&connection_state.session),
                                             )
                                             .await;
+                                        let opted_out_notification_methods_snapshot = connection_state
```
jif-oai, line 782: Is it still necessary now that the session state is immutable after init?

codex-rs/app-server/src/message_processor.rs
```diff
@@ -678,6 +746,29 @@ impl MessageProcessor {
             return;
         }
 
+        let app_server_client_name = session.app_server_client_name().map(str::to_string);
```
jif-oai, line 750: ultra nit: I would defer this clone till the code actually detaches work

codex-rs/app-server/src/message_processor.rs
```diff
@@ -299,11 +339,11 @@ impl MessageProcessor {
     }
 
     pub(crate) async fn process_request(
-        &self,
+        self: &Arc<Self>,
```
jif-oai, line 343: This `Arc` receiver leaks the future concurrency implementation detail into every caller, this is strange. I would keep an `&self` and Arc it only at the spawn site

euroelessar: I can't promote `&self` into `Arc<Self>` without storing weak pointer on itself or similar as a field, but that would be even more strange?
alternatively we can move all the fields into a new `MessageProcessorState` or similar, but that does't provide clean benefits

codex-rs/app-server/src/lib.rs
euroelessar: it's a bit bigger change to change this (e.g. I would need to remove `ConnectionState`, plumb the `Arc<ConnectionSessionState>` throughout the stack, etc)
ok if I do it in a followup?

codex-rs/app-server/src/message_processor.rs
euroelessar: done!

codex-rs/app-server/src/lib.rs
jif-oai: Yes but please really do it as a follow-up :D

codex-rs/app-server/src/message_processor.rs
jif-oai: Ideally we have MessageProcessor as cheap cloneable handle we hide an Arced inner in it

But can be done in the same follow-up

jif-oai (approved):

codex-rs/app-server/src/message_processor.rs
euroelessar: sounds good, will add inner in the followup then

PR #17373, opened 2026-04-10
app-server: run initialized rpcs with keyed serialization

euroelessar:
## Why

Initialized app-server RPCs no longer need to bottleneck behind one request processor path. Running them concurrently improves responsiveness, but several request families still mutate shared state or depend on ordered side effects. Those stateful families need an auditable serialization contract so concurrency does not reorder thread, config, auth, command, watcher, MCP, or similar state transitions.

This PR keeps that boundary explicit: stateful work is serialized by the smallest useful key, while intentionally read-only or externally concurrent work remains unkeyed. In particular, `thread/list` and `thread/turns/list` explicitly have no serialization because they primarily read append-only rollout storage and should continue to be served concurrently.

## What changed

- Adds `ClientRequest::serialization_scope()` in `app-server-protocol` and requires every client request definition to declare its serialization behavior.
- Introduces keyed request scopes for thread, thread path, command exec process, fuzzy search session, fs watch, MCP OAuth, and global state buckets such as config, account auth, memory, and device keys.
- Routes initialized app-server RPCs through per-key FIFO serialization while allowing unkeyed initialized requests to run concurrently.
- Cancels in-flight initialized RPC work when the connection disconnects or the app-server exits so spawned request tasks do not outlive their session.
- Adds focused coverage for representative keyed and unkeyed serialization scopes, including explicitly concurrent `thread/turns/list` behavior.

## Validation

- Added protocol tests for representative keyed serialization scopes and intentionally unkeyed request families.
- Added app-server request serialization tests covering per-key FIFO behavior, concurrent unkeyed execution, disconnect shutdown, and config read-after-write ordering.
- Local focused protocol validation after the latest rebase is currently blocked by packageproxy failing to resolve locked `rustls-webpki 0.103.13`; CI is expected to provide the full validation signal.

### DIFF
--- a/codex-rs/app-server-protocol/src/protocol/common.rs
+++ b/codex-rs/app-server-protocol/src/protocol/common.rs
@@ -97,6 +97,12 @@ macro_rules! serialization_scope_expr {
             thread_id: $actual_params.$field.clone(),
         })
     };
+    ($actual_params:ident, optional_thread_id($params:ident . $field:ident)) => {
+        $actual_params
+            .$field
+            .clone()
+            .map(|thread_id| ClientRequestSerializationScope::Thread { thread_id })
+    };
     ($actual_params:ident, thread_or_path($params:ident . $thread_field:ident, $params2:ident . $path_field:ident)) => {
         if !$actual_params.$thread_field.is_empty() {
             Some(ClientRequestSerializationScope::Thread {
@@ -681,7 +687,7 @@ client_request_definitions! {
 
     McpResourceRead => "mcpServer/resource/read" {
         params: v2::McpResourceReadParams,
-        serialization: thread_id(params.thread_id),
+        serialization: optional_thread_id(params.thread_id),
         response: v2::McpResourceReadResponse,
     },
 
@@ -1468,6 +1474,21 @@ mod tests {
             })
         );
 
+        let mcp_resource_read = ClientRequest::McpResourceRead {
+            request_id: request_id(),
+            params: v2::McpResourceReadParams {
+                thread_id: Some("thread-1".to_string()),
+                server: "server-a".to_string(),
+                uri: "file:///tmp/resource".to_string(),
+            },
+        };
+        assert_eq!(
+            mcp_resource_read.serialization_scope(),
+            Some(ClientRequestSerializationScope::Thread {
+                thread_id: "thread-1".to_string()
+            })
+        );
+
         let config_read = ClientRequest::ConfigRead {
             request_id: request_id(),
             params: v2::ConfigReadParams {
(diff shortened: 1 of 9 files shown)


codex-rs/app-server/src/message_processor.rs
```diff
@@ -746,17 +751,37 @@ impl MessageProcessor {
             return;
         }
 
+        let serialization_scope = codex_request.serialization_scope();
```
jif-oai, line 740: This dispatch is now detached from app-server shutdown/connection teardown.Can we track/drain these initialized RPC tasks so config/plugin/account/... mutations are not left running after disconnect or graceful shutdown?

codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -363,56 +483,68 @@ client_request_definitions! {
...
         response: v2::FsWatchResponse,
     },
     FsUnwatch => "fs/unwatch" {
         params: v2::FsUnwatchParams,
+        serialization: fs_watch_id(params.watch_id),
         response: v2::FsUnwatchResponse,
     },
     SkillsConfigWrite => "skills/config/write" {
         params: v2::SkillsConfigWriteParams,
+        serialization: global("skills-config"),
```
jif-oai: This updates the same things as `config/...` so using a separate queue here will make everything racy

codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -363,56 +483,68 @@ client_request_definitions! {
...
         response: v2::SkillsConfigWriteResponse,
     },
     PluginInstall => "plugin/install" {
         params: v2::PluginInstallParams,
+        serialization: plugin(params.marketplace_path, params.plugin_name),
```
jif-oai: Install and uninstall don't share the same queue. This sounds odd to me. How do you prevent race? This sounds like a way to have some stale plugins or inconsistent states

codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -363,56 +483,68 @@ client_request_definitions! {
...
         response: v2::PluginInstallResponse,
     },
     PluginUninstall => "plugin/uninstall" {
         params: v2::PluginUninstallParams,
+        serialization: plugin_id(params.plugin_id),
```
jif-oai: The key is different from `plugin/install`. This sounds like a bad design

codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -244,19 +350,23 @@ client_request_definitions! {
     ThreadResume => "thread/resume" {
         params: v2::ThreadResumeParams,
         inspect_params: true,
+        serialization: thread_or_path(params.thread_id, params.path),
```
jif-oai, line 339: Keying by path whenever it is present does not match the other behavior for already-loaded threads, which still operates on id first (so what if both are present?)

euroelessar: oh, agree, swapped the order to prioritize `thread_id` if present

codex-rs/app-server/src/message_processor.rs
euroelessar: added a logic to wait for all rpcs for this connection to be either cancelled or waited upon

owenlin0 (approved):
looks very reasonable to me! quite a nice declarative API too 👍

PR #17374, opened 2026-04-10
[codex-analytics] add session source to client metadata

marksteinbrick-oai:
## Summary

Adds `thread_source` field to the existing Codex turn metadata sent to Responses API
- Sends `thread_source: "user"` for user-initiated sessions: CLI, VS Code, and Exec
- Sends `thread_source: "subagent"` for subagent sessions
- Omits `thread_source` for MCP, custom, and unknown session sources
- Uses the existing turn metadata transport:
  - HTTP requests send through the `x-codex-turn-metadata` header
  - WebSocket `response.create` requests send through `client_metadata["x-codex-turn-metadata"]`

## Testing
- `cargo test -p codex-protocol session_source_thread_source_name_classifies_user_and_subagent_sources`
- `cargo test -p codex-core turn_metadata_state`
- `cargo test -p codex-core --test responses_headers responses_stream_includes_turn_metadata_header_for_git_workspace_e2e -- --nocapture`

### DIFF
--- a/codex-rs/core/src/codex.rs
+++ b/codex-rs/core/src/codex.rs
@@ -1570,7 +1570,7 @@ impl Session {
         let per_turn_config = Arc::new(per_turn_config);
         let turn_metadata_state = Arc::new(TurnMetadataState::new(
             conversation_id.to_string(),
-            session_source.to_string(),
+            &session_source,
             sub_id.clone(),
             cwd.to_path_buf(),
             session_configuration.sandbox_policy.get(),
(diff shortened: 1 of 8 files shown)


marksteinbrick-oai:

codex-rs/core/src/codex.rs
```diff
@@ -1558,6 +1558,7 @@ impl Session {
         let per_turn_config = Arc::new(per_turn_config);
         let turn_metadata_state = Arc::new(TurnMetadataState::new(
             conversation_id.to_string(),
+            session_source.to_string(),
```
pash-openai: Can we avoid using `session_source.to_string()` directly here? For thread-spawned subagents, that formats as `subagent_thread_spawn_<parent_thread_id>_d<depth>`, which makes this new metadata field high-cardinality and bakes thread IDs into the value. Since the goal is source classification for analytics, I think we want a normalized label instead, similar to the existing `collab_spawn` / `thread_spawn` mappings elsewhere.

marksteinbrick-oai: Updated, now using a helper function to collapse the session_source into a specific thread_source value (`"user"` or `"subagent"` )

pash-openai (approved):

PR #17381, opened 2026-04-10
representing guardian review timeouts in protocol types

won-openai:
## Summary

- Add `TimedOut` to Guardian/review carrier types:
  - `ReviewDecision::TimedOut`
  - `GuardianAssessmentStatus::TimedOut`
  - app-server v2 `GuardianApprovalReviewStatus::TimedOut`
- Regenerate app-server JSON/TypeScript schemas for the new wire shape.
- Wire the new status through core/app-server/TUI mappings with conservative fail-closed handling.
- Keep `TimedOut` non-user-selectable in the approval UI.

**Does not change runtime behavior yet; emitting `TimeOut` and parent-model timeout messaging will come in followup PRs**

### DIFF
--- a/codex-rs/app-server-protocol/schema/json/ApplyPatchApprovalResponse.json
+++ b/codex-rs/app-server-protocol/schema/json/ApplyPatchApprovalResponse.json
@@ -95,6 +95,13 @@
           "type": "string"
         },
         {
+          "description": "Automatic approval review timed out before reaching a decision.",
+          "enum": [
+            "timed_out"
+          ],
+          "type": "string"
+        },
+        {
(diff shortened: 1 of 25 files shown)


owenlin0 (approved):
codex flagged this:

For GuardianAssessmentStatus::TimedOut, this live app-server path completes the command item as Declined, but ThreadHistoryBuilder::handle_guardian_assessment maps the same timed-out event to CommandExecutionStatus::Failed. Users who reload or replay the thread will see a different command status than live clients for the same timeout. Split the timed-out arm or make both mappings use the same status so live and replay agree.

but otherwise lgtm!

PR #17385, opened 2026-04-10
Add use_agent_identity feature flag

adrian-openai:
## Summary

Stack PR 1 of 4 for feature-gated agent identity support.

This PR adds the under-development `features.use_agent_identity` config flag and schema plumbing only. Runtime behavior remains unchanged with the flag off.

## Stack

- PR1: https://github.com/openai/codex/pull/17385 - this PR
- PR2: https://github.com/openai/codex/pull/17386 - register agent identities when enabled
- PR3: https://github.com/openai/codex/pull/17387 - register agent tasks when enabled
- PR4: https://github.com/openai/codex/pull/17388 - use `AgentAssertion` downstream when enabled

## Validation

Covered as part of the local stack validation pass:

- `just fmt`
- `cargo test -p codex-core --lib agent_identity`
- `cargo test -p codex-core --lib agent_assertion`
- `cargo test -p codex-core --lib websocket_agent_task`
- `cargo test -p codex-api api_bridge`
- `cargo build -p codex-cli --bin codex`

## Notes

The full local app-server E2E path is still being debugged after PR creation. The current branch stack is directionally ready for review while that follow-up continues.

PR #17386, opened 2026-04-10
Register agent identities behind use_agent_identity

adrian-openai:
## Summary

Stack PR 2 of 4 for feature-gated agent identity support.

This PR adds agent identity registration behind `features.use_agent_identity`. It keeps the app-server protocol unchanged and starts registration after ChatGPT auth exists rather than requiring a client restart.

## Stack

- PR2: https://github.com/openai/codex/pull/17386 - this PR
- PR4: https://github.com/openai/codex/pull/17388 - use `AgentAssertion` downstream when enabled

## Validation

- `just fmt`

## Notes


codex-rs/config/src/config_toml.rs
```diff
@@ -271,6 +271,15 @@ pub struct ConfigToml {
...
 
+    /// Base URL for agent identity registration requests.
+    pub agent_identity_base_url: Option<String>,
+
+    /// Base URL for minting human biscuits used by agent identity registration requests.
+    pub agent_identity_biscuit_base_url: Option<String>,
+
+    /// Target URL used when minting the human biscuit for agent identity registration.
+    pub agent_identity_biscuit_target_url: Option<String>,
+
```
shijie-oai: Do we want to consider defining those as env variables instead? For a machine acquiring an agent identity, I am not sure how much an enterprise will be able to inject into the config.toml if they are always starting with a clean copy per node spin-up and tear-down.

shijie-oai: hm but then they need to set the feature to turn on agent identity anyway so...

shijie-oai: Just want to be considerable when adding to the overall config.

codex-rs/core/src/config/mod.rs
```diff
@@ -2079,6 +2088,15 @@ impl Config {
                 .chatgpt_base_url
                 .or(cfg.chatgpt_base_url)
                 .unwrap_or("https://chatgpt.com/backend-api/".to_string()),
+            agent_identity_base_url: cfg
+                .agent_identity_base_url
+                .unwrap_or("https://auth.openai.com/api/accounts".to_string()),
+            agent_identity_biscuit_base_url: cfg
+                .agent_identity_biscuit_base_url
+                .unwrap_or_default(),
```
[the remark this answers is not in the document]
shijie-oai: this sounds reasonable.

codex-rs/config/src/config_toml.rs
shijie-oai: Read more of the code below too - do we expect those to change from our end? Seems like they are just openai backed endpoints?

codex-rs/core/src/agent_identity.rs
```diff
@@ -0,0 +1,755 @@
...
+
+    pub(crate) async fn ensure_registered_identity(&self) -> Result<Option<StoredAgentIdentity>> {
+        if !self.feature_enabled {
+            return Ok(None);
+        }
+
+        let Some(auth) = self.auth_manager.auth().await else {
+            debug!("skipping agent identity registration because no auth is available");
+            return Ok(None);
+        };
```
shijie-oai, line 125: Per flow here we would always require an auth which makes sense but it does not 100% solve the issue with us needing auth on a ci/cd job in the first place. Per thread discussion and hai doc, abom can be sufficiently used to auth an agent in a workflow and sounds like abom is technically the item we would dist to machines to assume human identity and run?

codex-rs/config/src/config_toml.rs
adrian-openai: Ok, so you think that these URLs should be env variables? I think that's actually probably correct. I actually think that they shouldn't be a special snowflake at all - they should just be the same base_url as the rest of the URLs (e.g. what codex-backend uses). Let me update that.

codex-rs/core/src/config/mod.rs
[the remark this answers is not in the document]
efrazer-oai: +1 here, this is important as we'll want to start with a preregistered identity for the #programmatic-codex project and we won't expect user to specify it in the .toml.

codex-rs/core/src/agent_identity.rs
adrian-openai: So the ABOM represents what the agent harness is (i.e. what code is running for this agent), but it doesn't represent the human identity. The human identity would be coming from the access token.

Let me know if I'm misunderstanding your question.

efrazer-oai (changes requested):
I've been looking through the stack and I'm increasingly wondering about the following:


I wonder if this would be better organized by moving this core logic into 

enum CodexAuth {
    ApiKey(ApiKeyAuth),
    Chatgpt(ChatgptAuth),
    ChatgptAuthTokens(ChatgptAuthTokens),
}

It seems like it is a kinda 'different' method of auth and is intended to fully migrate away from Chatgpt auth as it exists right now when enabled (insomuch as it replaces every authorization header).

It's a bit weird to reason about a 'ChatGPT' user also being an agent identity (or using agent identity), then having these kinda 'subtle' overrides and all of this being managed in a separate auth struct. 

We'll also need to add this variant later when we allow starting from a preregistered agent identity, so might be good to do now.

codex-rs/config/src/config_toml.rs
shijie-oai: Agreed - i don't think there is any issue to hardcode this if we expect them to always be the same.

codex-rs/core/src/agent_identity.rs
shijie-oai: It was not super clear to me before with when/where we would create such access token in the flow for a CI/CD flow to consume if we still require auth as the step ahead. Talked to to @efrazer-oai and I have a better picture now which is we would provide an agent identity via an admin UI and such identity would be injected into this workflow via an env variable (not covered in this PR) and we can then parse and operate as expected.

adrian-openai: Yeah, that seems reasonable to me. We need to add the right knob to codex so when it starts up it can find that it is already in agent identity mode, and it already as an identity, then we can materialize that for the remotely running agents.

codex-rs/core/Cargo.toml
```diff
@@ -144,6 +145,7 @@ codex-shell-escalation = { workspace = true }
 assert_cmd = { workspace = true }
 assert_matches = { workspace = true }
 codex-arg0 = { workspace = true }
+codex-keyring-store = { workspace = true }
```
efrazer-oai: nit: i dont think we need this now that we use auth.json

efrazer-oai:
@codex review

efrazer-oai (approved):
Approved to unblock, tests & one nit if you have time!

codex-rs/login/src/auth/manager.rs
```diff
@@ -351,6 +353,52 @@ impl CodexAuth {
         self.get_current_auth_json().and_then(|t| t.tokens)
     }
 
+    pub fn get_agent_identity(&self, workspace_id: &str) -> Option<AgentIdentityAuthRecord> {
+        self.get_current_auth_json()
+            .and_then(|auth| auth.agent_identity)
+            .filter(|identity| identity.workspace_id == workspace_id)
+    }
```
shijie-oai, line 361: The benefit of object would be free uniqueness check. Do we want to throw if more than one match is found?

shijie-oai (approved):
LGTM - I am building this locally to make sure it does function as expected.

shijie-oai:
Built and existing flow works as expected.

codex-rs/login/src/auth/manager.rs
adrian-openai: Sure! We can try that. If we throw, how will they get out of this situation? Will throwing wipe the auth.json file and force people to log in again or something?

codex-rs/core/src/config/mod.rs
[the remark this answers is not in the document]
adrian-openai: Fixed in the current PR shape: there is no separate `agent_identity_biscuit_base_url` config anymore. Agent registration and biscuit minting both derive from the existing `chatgpt_base_url`, so the default ChatGPT/Codex backend base URL is usable without an extra knob.

codex-rs/config/src/config_toml.rs
adrian-openai: Agreed. This no longer adds new agent-identity URL config. The registration and biscuit URLs are derived from the same `chatgpt_base_url` path the ChatGPT-backed Codex flow already uses.

codex-rs/core/Cargo.toml
adrian-openai: Yep, right. Removed the unused `codex-keyring-store` dev-dependency now that agent identity persistence lives in auth state; `Cargo.lock` is updated too.

codex-rs/core/src/agent_identity.rs
```diff
@@ -0,0 +1,714 @@
...
+        target_url: &str,
+    ) -> Result<String> {
+        let url = agent_identity_biscuit_url(&self.chatgpt_base_url);
+        let request_id = agent_identity_request_id()?;
+        let client = create_client();
+        let response = client
+            .get(&url)
+            .bearer_auth(&binding.access_token)
+            .header("X-Request-Id", request_id.clone())
+            .header("X-Original-Method", "GET")
```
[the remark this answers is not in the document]
adrian-openai: Fixed in `46037d77d`: biscuit minting now passes the downstream registration method (`POST`) as `X-Original-Method`, and the registration test expectation was updated.

codex-rs/core/src/agent_identity.rs
```diff
@@ -0,0 +1,714 @@
...
+        })
+    }
+
+    fn to_auth_record(&self) -> AgentIdentityAuthRecord {
+        AgentIdentityAuthRecord {
+            workspace_id: self.chatgpt_account_id.clone(),
+            agent_runtime_id: self.agent_runtime_id.clone(),
+            agent_private_key: self.private_key_pkcs8_base64.clone(),
+            registered_at: self.registered_at.clone(),
+        }
```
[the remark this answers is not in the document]
adrian-openai: Fixed in `616d25cb0`: the auth-backed agent identity record now persists `chatgpt_user_id`, and reuse is checked against that stored value instead of reconstructing it from current auth. I also added coverage for same-workspace/different-user invalidation and re-registration.

codex-rs/login/src/auth/manager.rs
adrian-openai: The current `auth.json` shape is a single optional `agent_identity` object, not a collection, so there is no multiple-match state to detect. `get_agent_identity` only returns that object when its `workspace_id` matches the current workspace; otherwise it behaves as absent. I also pinned the record to `chatgpt_user_id` so a same-workspace user change invalidates and re-registers rather than reusing stale material.

shijie-oai: Ya I think throw is the last line of defense here - we should never really got into that position ideally in the first place.

PR #17387, opened 2026-04-10
Register agent tasks behind use_agent_identity

adrian-openai:
## Summary

Stack PR3 for feature-gated agent identity support.

This PR adds per-thread agent task registration behind `features.use_agent_identity`. Tasks are minted on the first real user turn and cached in thread runtime state for later turns.

## Stack

- PR3: https://github.com/openai/codex/pull/17387 - this PR, original task registration slice
- PR4: https://github.com/openai/codex/pull/17980 - use `AgentAssertion` downstream when enabled

## Validation

- `just fmt`

## Notes


efrazer-oai (commented):
I don't think we need this `auth_binding` concept with our auth.json changes. Can do smth like below:

`match auth {
    CodexAuth::ChatgptAuthTokens(tokens) => {
        if let Some(agent_identity) = tokens.registered_agent_identity() {
            // preregistered path
            register_task_with_agent_identity(agent_identity).await
        } else if let Some(access_token) = tokens.authorization_bearer_token() {
            // human bootstrap path
            bootstrap_and_register_task(access_token, tokens.workspace_id()).await
        } else {
            Ok(None)
        }
    }
    _ => Ok(None),
}
`

We shouldn't exist in a state where the agent `workspace_id` doesn't match up with the user `workspace_id`

efrazer-oai (commented):
### Session lifecycle
In the status quo, we fully persist session details on disk. With this change, we're creating an ephemeral 'task' in memory associated with a session. This is a bit off for a few reasons I think:

1. It means if we turn off codex and start working on the same session later, it shows up as two different 'tasks' which seems like undesirable backend state. 
2. It means we add latency to every initial turn in order to make this HTTP call to create the task (idt it'd be *too* severe, but latency is a huge focus and we try to make most interactions go through websockets; if we can avoid i would).
3. There are some API calls we make that are not session scoped, and some that are session scoped but aren't in the path where we create agent_task. If we're in a regime where there's no user token (i.e. programmatic codex), i believe these will just fail -- would need to make it robust to those.

I wonder if there's a cleaner implementation that loses some fidelity but just creates it on start? 

Or if we can pass the session id itself as the task id?

efrazer-oai:
@codex review

notes/pr17387-agent-task-persistence-assumption.md
```diff
@@ -0,0 +1,15 @@
+# PR 17387 Agent Task Persistence Assumption
```
nicksteele-oai: File should probably be rm'd from commit

codex-rs/core/src/codex.rs
nicksteele-oai, line 1578: Should there be a way to allow users to soft fail after registration fails even when `features.use_agent_identity` is enabled? Calling `shutdown` if registration fails, even by accident, might result in some grotty UX. For instance, `shutdown` calls `abort_all_tasks(TurnAbortReason::Interrupted)`, which iiuc will abort all active session tasks even if they're unrelated to this session.

codex-rs/core/src/codex.rs
```diff
@@ -6391,6 +6373,11 @@ pub(crate) async fn run_turn(
         }))
         .await;
     }
+    if let Err(error) = sess.ensure_agent_task_registered().await {
+        warn!(error = %error, "agent task registration failed");
+        sess.fail_agent_identity_registration(error).await;
```
nicksteele-oai: Could we treat task registration fails as a turn error that can be retried and let the next turn attempt registration again instead of ending the session?

codex-rs/core/src/codex.rs
nicksteele-oai, line 2334: The other thing we might want to consider is have some progress loader/event so if this event hangs the user has some indicator that it's still doing the registration

nicksteele-oai:
Most of my comments (other than the shutdown invocation) are pretty minor, otherwise looking in good shape!

notes/pr17387-agent-task-persistence-assumption.md
adrian-openai: Weird that notes are actually making it into the PR. I'll remove that.

codex-rs/core/src/codex.rs
adrian-openai: This is a fair point. Probably worth a better UX here, where there's an error for _this_ thread, but doesn't necessarily stop all other threads.

codex-rs/core/src/codex.rs
adrian-openai: I'm not sure if I'm following - how will that turn be completed without a task? Do you mean that this 'turn error' just returns control back to the user who can try again in case it is a transient error? I think that seems reasonable!

nicksteele-oai: Yep the latter! The turn shouldn't be completed but also shouldn't cause a hard fail, and the user should be able to attempt a retry

efrazer-oai (approved):
Approved to unblock, make sure to smoke test the path where there's no logged in chat user token locally! (i.e. programmatic agent identity)

PR #17388, opened 2026-04-10
Use AgentAssertion downstream behind use_agent_identity

adrian-openai:
## Summary

Stack PR 4 of 4 for feature-gated agent identity support.

This PR uses the registered agent task to attach `AgentAssertion` authorization on downstream backend paths when `features.use_agent_identity` is enabled.

## Stack

- PR4: https://github.com/openai/codex/pull/17388 - this PR

## Validation

- `just fmt`

## Notes

### DIFF
--- a/codex-rs/core/src/agent_identity/assertion.rs
+++ b/codex-rs/core/src/agent_identity/assertion.rs
@@ -90,8 +90,9 @@ mod tests {
 
     #[tokio::test]
     async fn authorization_header_for_task_skips_when_feature_is_disabled() {
-        let auth_manager =
-            AuthManager::from_auth_for_testing(CodexAuth::create_dummy_chatgpt_auth_for_testing());
+        let codex_home = tempfile::tempdir().expect("tempdir");
+        let auth = make_chatgpt_auth(codex_home.path(), "account-123", Some("user-123"));
+        let auth_manager = AuthManager::from_auth_for_testing(auth);
         let manager = AgentIdentityManager::new_for_tests(
             auth_manager,
             /*feature_enabled*/ false,
@@ -99,9 +100,9 @@ mod tests {
             SessionSource::Cli,
         );
         let agent_task = RegisteredAgentTask {
-            binding_id: "chatgpt-account-account_id".to_string(),
-            chatgpt_account_id: "account_id".to_string(),
-            chatgpt_user_id: None,
+            binding_id: "chatgpt-account-account-123".to_string(),
+            chatgpt_account_id: "account-123".to_string(),
+            chatgpt_user_id: Some("user-123".to_string()),
             agent_runtime_id: "agent-123".to_string(),
             task_id: "task-123".to_string(),
             registered_at: "2026-03-23T12:00:00Z".to_string(),
@@ -118,8 +119,9 @@ mod tests {
 
     #[tokio::test]
     async fn authorization_header_for_task_serializes_signed_agent_assertion() {
-        let auth_manager =
-            AuthManager::from_auth_for_testing(CodexAuth::create_dummy_chatgpt_auth_for_testing());
+        let codex_home = tempfile::tempdir().expect("tempdir");
+        let auth = make_chatgpt_auth(codex_home.path(), "account-123", Some("user-123"));
+        let auth_manager = AuthManager::from_auth_for_testing(auth);
         let manager = AgentIdentityManager::new_for_tests(
             auth_manager,
             /*feature_enabled*/ true,
@@ -131,9 +133,9 @@ mod tests {
             .await
             .expect("seed test identity");
         let agent_task = RegisteredAgentTask {
-            binding_id: "chatgpt-account-account_id".to_string(),
-            chatgpt_account_id: "account_id".to_string(),
-            chatgpt_user_id: None,
+            binding_id: "chatgpt-account-account-123".to_string(),
+            chatgpt_account_id: "account-123".to_string(),
+            chatgpt_user_id: Some("user-123".to_string()),
             agent_runtime_id: "agent-123".to_string(),
             task_id: "task-123".to_string(),
             registered_at: "2026-03-23T12:00:00Z".to_string(),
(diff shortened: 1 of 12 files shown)


codex-rs/core/src/client.rs
```diff
@@ -306,6 +312,33 @@ impl ModelClient {
         enable_request_compression: bool,
         include_timing_metrics: bool,
         beta_features_header: Option<String>,
+    ) -> Self {
+        Self::new_with_agent_identity_manager(
+            auth_manager,
+            /*agent_identity_manager*/ None,
```
shijie-oai, line 318: Just to clarify - it is plumbed but not fully at this point?

codex-rs/core/src/client.rs
```diff
@@ -211,6 +215,8 @@ pub struct ModelClient {
 pub struct ModelClientSession {
     client: ModelClient,
     websocket_session: WebsocketSession,
+    agent_task: Option<RegisteredAgentTask>,
+    cache_websocket_session_on_drop: bool,
```
shijie-oai, line 219: let me play it back a little bit - `cache_websocket_session_on_drop` is a bit that indicates if a websocket session can be cached and reused on drop.

codex-rs/core/src/codex.rs
```diff
@@ -6346,8 +6351,21 @@ pub(crate) async fn run_turn(
 
     // `ModelClientSession` is turn-scoped and caches WebSocket + sticky routing state, so we reuse
     // one instance across retries within this turn.
-    let mut client_session =
-        prewarmed_client_session.unwrap_or_else(|| sess.services.model_client.new_session());
+    let mut prewarmed_client_session = prewarmed_client_session;
+    if agent_task.is_some()
```
shijie-oai, line 6355: For my personal understanding of this code block - why would an agent task associated with a thread not able to use the prewarmed client session of that thread? Would that be a possible scenario?

shijie-oai: Chatted with Codex a bit locally and correct me if I am wrong - in this case an initial prewarmed thread is created without agent_task (none) and if we want to reuse it, it has to have a matching task. The effectively meant that the initial prewarmed thread on thread creation is always discarded and we do not get the benefit of it?

codex-rs/core/src/client.rs
adrian-openai: I'm not sure what happened, but somehow this PR was merged. I'll respond anyway, and we can figure out what's going on, I'll re-open this on another PR, I think.

adrian-openai: Ah ok, my codex fucked up, and github closed this automatically. Will reopen here: https://github.com/openai/codex/pull/17807

shijie-oai: No worries - let me know if you want to continue the conversation here or there.

codex-rs/core/src/agent_identity/assertion.rs
```diff
@@ -0,0 +1,195 @@
...
+
+        let Some(stored_identity) = self.ensure_registered_identity().await? else {
+            return Ok(None);
+        };
+        anyhow::ensure!(
+            stored_identity.agent_runtime_id == agent_task.agent_runtime_id,
+            "agent task runtime {} does not match stored agent identity {}",
+            agent_task.agent_runtime_id,
+            stored_identity.agent_runtime_id
+        );
```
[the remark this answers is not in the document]
adrian-openai: Moved/fixed in the replacement PR: #17807.

We changed this from a hard `ensure!` failure into a recoverable stale-task path. Assertion construction now reports `AgentTaskRuntimeMismatch` when the persisted task points at an old runtime, client setup maps that to `CodexErr::AgentTaskStale`, and the sampling loop clears that cached task, registers a fresh one once, and retries with a new task-scoped session.

Code pointers in #17807:
- mismatch detection: https://github.com/openai/codex/blob/a87e178824ef8ec55f6b2d7a2a6e2518e0c69795/codex-rs/core/src/agent_identity/assertion.rs#L15-L52
- refresh/retry path: https://github.com/openai/codex/blob/a87e178824ef8ec55f6b2d7a2a6e2518e0c69795/codex-rs/core/src/codex.rs#L7072-L7089
- regression coverage: https://github.com/openai/codex/blob/a87e178824ef8ec55f6b2d7a2a6e2518e0c69795/codex-rs/core/src/client_tests.rs#L397-L424

codex-rs/core/src/client.rs
adrian-openai: Yep, that read is right: the path was plumbed, but we are intentionally not making API-key auth work with AgentAssertion in this PR.

In the replacement PR (#17807), the current scope is explicit: AgentAssertion is only for ChatGPT-backed Codex sessions, and API-key sessions continue using the API key until the registration service supports API-key identity binding. The downstream request setup still switches to `AgentAssertion` when a task exists for a ChatGPT-auth session.

Code pointers in #17807:
- ChatGPT-only scope comment: https://github.com/openai/codex/blob/a87e178824ef8ec55f6b2d7a2a6e2518e0c69795/codex-rs/core/src/agent_identity.rs#L454-L459
- downstream AgentAssertion header selection: https://github.com/openai/codex/blob/a87e178824ef8ec55f6b2d7a2a6e2518e0c69795/codex-rs/core/src/client.rs#L751-L775
- API-key task registration remains skipped: https://github.com/openai/codex/blob/a87e178824ef8ec55f6b2d7a2a6e2518e0c69795/codex-rs/core/src/agent_identity/task_registration.rs#L193-L204

codex-rs/core/src/client.rs
adrian-openai: Exactly, and the replacement PR (#17807) makes that bit a little less load-bearing for auth correctness.

The cache now stores the `RegisteredAgentTask` alongside the WebSocket session. `cache_websocket_session_on_drop` only controls whether this session should be put back in the cache at all; actual reuse is allowed only when the cached connection's task matches the next session's task. That means same-thread/same-task WebSockets can be reused, while taskless vs task-scoped or different-task auth contexts are dropped instead of mixed.

Code pointers in #17807:
- task-keyed cache/reuse logic: https://github.com/openai/codex/blob/a87e178824ef8ec55f6b2d7a2a6e2518e0c69795/codex-rs/core/src/client.rs#L383-L459
- cache-on-drop stores the task key: https://github.com/openai/codex/blob/a87e178824ef8ec55f6b2d7a2a6e2518e0c69795/codex-rs/core/src/client.rs#L917-L923
- reuse coverage for the same task: https://github.com/openai/codex/blob/a87e178824ef8ec55f6b2d7a2a6e2518e0c69795/codex-rs/core/src/client_tests.rs#L503-L578

codex-rs/core/src/codex.rs
adrian-openai: Yep, your follow-up was the key point here: a taskless startup-prewarmed WebSocket is not useful once the first real turn needs task-scoped AgentAssertion auth, because we would have to discard that taskless connection.

In the replacement PR (#17807), we stopped scheduling startup prewarm for ChatGPT/AgentIdentity sessions. Instead, task registration starts once a user turn is accepted and overlaps with the rest of turn prep, then we await it before the model request. That preserves the "do not create a task at startup" behavior while still warming up as late/early as core can observe. We did not hook literal keypress/typing events in this PR, since that would require app/TUI input plumbing rather than the core turn path.

Code pointers in #17807:
- skip taskless startup prewarm for AgentIdentity sessions: https://github.com/openai/codex/blob/a87e178824ef8ec55f6b2d7a2a6e2518e0c69795/codex-rs/core/src/session_startup_prewarm.rs#L159-L168
- lazy task registration overlapped with turn prep: https://github.com/openai/codex/blob/a87e178824ef8ec55f6b2d7a2a6e2518e0c69795/codex-rs/core/src/codex.rs#L6357-L6394
- task-aware WebSocket reuse after that: https://github.com/openai/codex/blob/a87e178824ef8ec55f6b2d7a2a6e2518e0c69795/codex-rs/core/src/client.rs#L383-L459

codex-rs/core/src/client.rs
adrian-openai: Let me know if these followups answer your questions @shijie-oai

codex-rs/core/src/codex.rs
shijie-oai: Sorry I am re-reading those and I am not 100% sure no longer leveraging prewarming for chatgpt auth like method is ideal. Correct me if I am wrong but in long term agent identity would be applicable to all user instance and if that is the case we take away all prewarming capability essentially and lose the performance benefit (we are still keeping it for the api auth path)

PR #17398, opened 2026-04-11
app-server: Only unload threads which were unused for some time

euroelessar:
Currently app-server may unload actively running threads once the last connection disconnects, which is not expected.
Instead track when was the last active turn & when there were any subscribers the last time, also add 30 minute idleness/no subscribers timer to reduce the churn.

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -3828,9 +3828,18 @@ impl CodexMessageProcessor {
         self.command_exec_manager
             .connection_closed(connection_id)
             .await;
-        self.thread_state_manager
+        let thread_ids = self
+            .thread_state_manager
             .remove_connection(connection_id)
             .await;
+
+        for thread_id in thread_ids {
+            if self.thread_manager.get_thread(thread_id).await.is_err() {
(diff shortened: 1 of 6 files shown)


euroelessar:
@codex review

euroelessar:
@codex review

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -363,6 +364,7 @@ struct ThreadListFilters {
 const LOGIN_CHATGPT_TIMEOUT: Duration = Duration::from_secs(10 * 60);
 const LOGIN_ISSUER_OVERRIDE_ENV_VAR: &str = "CODEX_APP_SERVER_LOGIN_ISSUER";
 const APP_LIST_LOAD_TIMEOUT: Duration = Duration::from_secs(90);
+const THREAD_UNLOADING_DELAY: Duration = Duration::from_secs(30 * 60);
```
NachoSoto, line 367: Does this mean that the issue would still happen on a task that runs for 30 minutes?

NachoSoto: Never mind just caught up on the Slack thread

maxj-oai (approved):

PR #17402, opened 2026-04-11
chore: refactor name and namespace to single type

sayan-oai:
avoid passing them both around, unify on a type. this now also keys `ToolRegistry`.

tests pass

PR #17404, opened 2026-04-11
register all mcp tools with namespace

sayan-oai:
stacked on #17402.

MCP tools returned by `tool_search` (deferred tools) get registered in our `ToolRegistry` with a different format than directly available tools. this leads to two different ways of accessing MCP tools from our tool catalog, only one of which works for each. fix this by registering all MCP tools with the namespace format, since this info is already available.

also, direct MCP tools are registered to responsesapi without a namespace, while deferred MCP tools have a namespace. this means we can receive MCP `FunctionCall`s in both formats from namespaces. fix this by always registering MCP tools with namespace, regardless of deferral status.

make code mode track `ToolName` provenance of tools so it can map the literal JS function name string to the correct `ToolName` for invocation, rather than supporting both in core.

this lets us unify to a single canonical `ToolName` representation for each MCP tool and force everywhere to use that one, without supporting fallbacks.


codex-rs/codex-mcp/src/mcp_connection_manager.rs
```diff
@@ -1192,13 +1192,25 @@ impl McpConnectionManager {
     }
 
     pub async fn resolve_tool_info(&self, name: &str, namespace: Option<&str>) -> Option<ToolInfo> {
```
pakrym-oai: should this take ToolName?

codex-rs/codex-mcp/src/mcp_connection_manager.rs
```diff
@@ -1192,13 +1192,25 @@ impl McpConnectionManager {
...
-            Some(namespace) => format!("{namespace}{name}"),
-            None => name.to_string(),
-        };
-
-        self.list_all_tools().await.get(&qualified_name).cloned()
+        let all_tools = self.list_all_tools().await;
+        if let Some(namespace) = namespace {
+            let qualified_name = format!("{namespace}{name}");
+            if let Some(tool) = all_tools.get(&qualified_name)
+                && tool.callable_namespace == namespace
```
pakrym-oai: should tool carry tool name?

pakrym-oai: and we index all tools by tool name too?

pakrym-oai: This entire method could be             `all_tools.get(name).cloned()
`

codex-rs/codex-mcp/src/mcp_connection_manager_tests.rs
```diff
@@ -646,6 +646,68 @@ async fn list_all_tools_uses_startup_snapshot_while_client_is_pending() {
     assert_eq!(tool.callable_name, "calendar_create_event");
 }
 
+#[tokio::test]
+async fn resolve_tool_info_accepts_flat_and_split_callable_names() {
```
pakrym-oai: who passes in flat?

codex-rs/tools/src/tool_registry_plan_types.rs
```diff
@@ -73,6 +72,16 @@ pub struct ToolNamespace {
     pub description: Option<String>,
 }
 
+/// Direct MCP tool metadata needed to expose the flat Responses API tool while
+/// registering its runtime handler with the canonical namespace/name split.
+#[derive(Debug, Clone)]
+pub struct ToolRegistryPlanMcpTool<'a> {
```
pakrym-oai, line 79: should deffered be moved here?

pakrym-oai: Should this use ToolName?

codex-rs/tools/src/tool_registry_plan.rs
```diff
@@ -471,25 +470,26 @@ pub fn build_tool_registry_plan(
...
-            .collect();
-        entries.sort_by(|left, right| left.0.cmp(&right.0));
+        let mut entries = mcp_tools.to_vec();
+        entries.sort_by(|left, right| left.qualified_name.cmp(&right.qualified_name));
 
-        for (name, tool) in entries {
-            match mcp_tool_to_responses_api_tool(name.clone(), tool) {
+        for tool in entries {
+            let qualified_name = tool.qualified_name.clone();
+            match mcp_tool_to_responses_api_tool(qualified_name.clone(), tool.tool) {
```
pakrym-oai: why are we using qualified_name here?

pakrym-oai: isn't it alwasy .name?

codex-rs/core/src/tools/spec.rs
```diff
@@ -29,16 +30,21 @@ pub(crate) fn tool_user_shell_type(user_shell: &Shell) -> ToolUserShellType {
...
 
-fn map_mcp_tools_for_plan(mcp_tools: &HashMap<String, ToolInfo>) -> McpToolPlanInputs {
+fn map_mcp_tools_for_plan(mcp_tools: &HashMap<String, ToolInfo>) -> McpToolPlanInputs<'_> {
     McpToolPlanInputs {
         mcp_tools: mcp_tools
             .iter()
-            .map(|(name, tool)| (name.clone(), tool.tool.clone()))
+            .map(|(qualified_name, tool)| ToolRegistryPlanMcpTool {
+                qualified_name: qualified_name.clone(),
+                callable_name: tool.callable_name.clone(),
```
pakrym-oai: Why aren't we reusing ToolInfo here?

codex-rs/core/src/tools/router.rs
```diff
@@ -144,6 +143,10 @@ impl ToolRouter {
                         },
                     }))
                 } else {
+                    let tool_name = match namespace {
```
pakrym-oai: nit: add ToolName::new(ns, name)

codex-rs/core/src/tools/code_mode/mod.rs
```diff
@@ -284,25 +284,29 @@ async fn call_nested_tool(
...
+                ToolPayload::Mcp {
                     server: tool_info.server_name,
                     tool: tool_info.tool.name.to_string(),
                     raw_arguments,
                 },
-                Err(error) => return Err(FunctionCallError::RespondToModel(error)),
-            }
+            )
         } else {
             match build_nested_tool_payload(tool_runtime.find_spec(&tool_name), &tool_name, input) {
```
pakrym-oai, line 316: surprised runtime still uses name:str

codex-rs/core/src/tools/code_mode/mod.rs
```diff
@@ -284,25 +284,29 @@ async fn call_nested_tool(
         )));
     }
 
-    let payload =
+    let (tool_call_name, payload) =
```
pakrym-oai, line 301: Should this method take ToolName and we handle JS method -> ToolName conversion internally in code mode?

codex-rs/codex-mcp/src/mcp_connection_manager.rs
sayan-oai: fair. will do this in a follow up since it'd bloat this diff

codex-rs/codex-mcp/src/mcp_connection_manager_tests.rs
sayan-oai: updated the test to use ToolName; flat is only passed in at the boundary of code mode (function name is a string).

codex-rs/tools/src/tool_registry_plan_types.rs
sayan-oai: made it use ToolName; i want to clean up deferred vs direct in another PR (you pointed out some questions about `ToolRegistryPlanDeferredTool` in DMs).

codex-rs/core/src/tools/spec.rs
sayan-oai: we'd have to pull in codex-mcp as a dep, not sure that's worth it for the single type. we could make a small type crate, but again the only thing that it'd meaningfully contain is `ToolInfo`. thoughts?

codex-rs/codex-mcp/src/mcp_connection_manager.rs
sayan-oai: #17504, still debating whether its worth doing

codex-rs/codex-mcp/src/mcp_connection_manager.rs
```diff
@@ -155,6 +156,12 @@ pub struct ToolInfo {
     pub connector_description: Option<String>,
 }
 
+impl ToolInfo {
+    pub fn callable_tool_name(&self) -> ToolName {
+        ToolName::namespaced(self.callable_namespace.clone(), self.callable_name.clone())
```
pakrym-oai, line 161: Do we always namespace mcps now (even when we don't defer?)

codex-rs/codex-mcp/src/mcp_connection_manager_tests.rs
pakrym-oai: long term code mode should pass the correct value.

codex-rs/core/src/tools/spec.rs
pakrym-oai: core already depends on codex-mcp, no?

codex-rs/codex-mcp/src/mcp_connection_manager.rs
```diff
@@ -1191,14 +1198,20 @@ impl McpConnectionManager {
...
+            .into_values()
+            .map(|tool| (tool.callable_tool_name(), tool))
+            .collect::<HashMap<_, _>>();
 
-        self.list_all_tools().await.get(&qualified_name).cloned()
+        if let Some(tool) = tools_by_name.get(tool_name) {
+            return Some(tool.clone());
+        }
+
+        tools_by_name
```
pakrym-oai: still a bit sus of this fallback.

codex-rs/core/src/tools/code_mode/mod.rs
```diff
@@ -274,38 +274,40 @@ async fn build_nested_router(exec: &ExecContext) -> ToolRouter {
...
-    };
+    let display_name = tool_name.display();
+    let (tool_call_name, payload) =
+        if let Some(tool_info) = exec.session.resolve_mcp_tool_info(&tool_name).await {
+            let raw_arguments = match serialize_function_tool_arguments(&display_name, input) {
+                Ok(raw_arguments) => raw_arguments,
+                Err(error) => return Err(FunctionCallError::RespondToModel(error)),
+            };
+            (
+                tool_info.callable_tool_name(),
```
pakrym-oai: non blocking: tool_runtime.find_spec should have all this logic.

codex-rs/core/src/tools/code_mode/mod.rs
```diff
@@ -349,10 +352,11 @@ fn build_nested_tool_payload(
 }
 
 fn build_function_tool_payload(
-    tool_name: &str,
+    tool_name: &ToolName,
     input: Option<JsonValue>,
 ) -> Result<ToolPayload, String> {
-    let arguments = serialize_function_tool_arguments(tool_name, input)?;
+    let display_name = tool_name.display();
```
pakrym-oai: nit: i'd pass entire tool name into serialize_function_tool_arguments so we don't accidentally use displan name for real

codex-rs/core/src/tools/code_mode/mod.rs
```diff
@@ -371,11 +375,12 @@ fn serialize_function_tool_arguments(
...
 fn build_freeform_tool_payload(
-    tool_name: &str,
+    tool_name: &ToolName,
     input: Option<JsonValue>,
 ) -> Result<ToolPayload, String> {
+    let display_name = tool_name.display();
     match input {
         Some(JsonValue::String(input)) => Ok(ToolPayload::Custom { input }),
-        _ => Err(format!("tool `{tool_name}` expects a string input")),
+        _ => Err(format!("tool `{display_name}` expects a string input")),
```
pakrym-oai: can't you use tool_name in formatting directly, considering it implements display?

codex-rs/codex-mcp/src/mcp_connection_manager.rs
pakrym-oai: Is this the right name? The callable name seems to be `mcp__ns__name`?

codex-rs/tools/src/tool_registry_plan.rs
```diff
@@ -76,9 +74,9 @@ pub fn build_tool_registry_plan(
             .tool_namespaces
             .into_iter()
             .flatten()
-            .map(|(name, detail)| {
+            .map(|(tool_name, detail)| {
                 (
-                    name.clone(),
+                    tool_name.display(),
```
pakrym-oai: this is a bit strange, shouldn't namespace descriptions be map<ns, descr>?

codex-rs/tools/src/tool_registry_plan.rs
```diff
@@ -471,25 +466,23 @@ pub fn build_tool_registry_plan(
...
-            .map(|(name, tool)| (name.clone(), tool))
-            .collect();
-        entries.sort_by(|left, right| left.0.cmp(&right.0));
+        let mut entries = mcp_tools.to_vec();
+        entries.sort_by_key(|tool| tool.name.display());
 
-        for (name, tool) in entries {
-            match mcp_tool_to_responses_api_tool(name.clone(), tool) {
+        for tool in entries {
+            let display_name = tool.name.display();
```
pakrym-oai: we shouldn't use display name here.

pakrym-oai: for mcp_tool_to_responses_api_tool I mean.

codex-rs/tools/src/tool_registry_plan_tests.rs
```diff
@@ -1886,14 +1888,31 @@ fn build_specs_with_optional_tool_namespaces<'a>(
     config: &ToolsConfig,
     mcp_tools: Option<HashMap<String, rmcp::model::Tool>>,
     deferred_mcp_tools: Option<Vec<ToolRegistryPlanDeferredTool<'a>>>,
-    tool_namespaces: Option<HashMap<String, ToolNamespace>>,
+    tool_namespaces: Option<HashMap<ToolName, ToolNamespace>>,
     discoverable_tools: Option<Vec<DiscoverableTool>>,
     dynamic_tools: &[DynamicToolSpec],
 ) -> (Vec<ConfiguredToolSpec>, Vec<ToolHandlerSpec>) {
+    let mcp_tool_inputs = mcp_tools.as_ref().map(|mcp_tools| {
+        mcp_tools
```
pakrym-oai, line 1887: mcp_tools in this test should probably change to ToolName?

pakrym-oai: So we dont' reparse

codex-rs/core/src/tools/spec.rs
sayan-oai: this is in the codex-tools subcrate

codex-rs/codex-mcp/src/mcp_connection_manager.rs
sayan-oai: renamed to canonical_tool_name()

sayan-oai: and yes we always namespace mcps with the latest changes

codex-rs/tools/src/tool_registry_plan.rs
sayan-oai: makes sense, changed to use `ToolName`

codex-rs/codex-mcp/src/mcp_connection_manager.rs
sayan-oai: gone with latest changes

codex-rs/code-mode/src/runtime/globals.rs
```diff
@@ -53,10 +53,10 @@ fn build_tools_object<'s>(
         .map(|state| state.enabled_tools.clone())
         .unwrap_or_default();
 
-    for tool in enabled_tools {
+    for (tool_index, tool) in enabled_tools.iter().enumerate() {
         let name = v8::String::new(scope, &tool.global_name)
             .ok_or_else(|| "failed to allocate tool name".to_string())?;
-        let function = tool_function(scope, &tool.tool_name)?;
+        let function = tool_function(scope, tool_index)?;
```
pakrym-oai, line 59: can we embed `ToolName`?

pakrym-oai: or name + namespace?

codex-rs/code-mode/src/description.rs
```diff
@@ -269,11 +271,15 @@ pub fn build_exec_tool_description(
         for tool in enabled_tools {
             let name = tool.name.as_str();
             let nested_description = render_code_mode_sample_for_definition(tool);
-            let next_namespace = namespace_descriptions
-                .get(name)
+            let namespace_description = tool
```
pakrym-oai, line 274: later: we should have a higher level concept for namespaces. this is painful.

codex-rs/core/src/tools/router.rs
```diff
@@ -102,20 +103,48 @@ impl ToolRouter {
...
-                .any(|config| config.name() == tool_name.name.as_str())
+        if tool_name.namespace.is_some() {
+            return false;
+        }
+
+        self.specs
+            .iter()
+            .filter(|config| config.supports_parallel_tool_calls)
+            .any(|config| match &config.spec {
+                ToolSpec::Function(tool) => tool.name == tool_name.name.as_str(),
```
pakrym-oai, line 140: consider tool_name() method on ToolSpec enum

codex-rs/core/tests/suite/openai_file_mcp.rs
```diff
@@ -124,7 +115,13 @@ async fn codex_apps_file_params_upload_local_paths_before_mcp_tool_call() -> Res
 
     let requests = mock.requests();
     let body = requests[0].body_json();
-    let extract_tool = tool_by_name(&body, DOCUMENT_EXTRACT_TOOL);
+    let Some(extract_tool) =
+        namespace_child_tool(&body, DOCUMENT_EXTRACT_NAMESPACE, DOCUMENT_EXTRACT_TOOL)
```
pakrym-oai: nit: mock.requests()[0].tool_by_name(..)?

codex-rs/tools/src/tool_registry_plan.rs
```diff
@@ -471,28 +467,56 @@ pub fn build_tool_registry_plan(
...
+                        let tool_name = &tool.name;
+                        tracing::error!(
+                            "Failed to convert `{tool_name}` MCP tool to OpenAI tool: {error:?}"
+                        );
+                    }
                 }
             }
+
+            if !tools.is_empty() {
+                plan.push_spec(
```
pakrym-oai, line 510: This is good but will it conflict with TST deferred tools?

pakrym-oai (approved):

codex-rs/code-mode/src/runtime/globals.rs
sayan-oai: the canonical `ToolName` is available in the `EnabledToolMetadata` the index points to. this way we dont have to add v8 parsing for `ToolName`. the list we have indexes over is immutable, so i think this is fine.

codex-rs/core/src/tools/router.rs
sayan-oai: the `ToolSpec::Namespace` variant can have multiple tool names though (its children). we could abstract that logic into a helper, but its only used once right now.

codex-rs/tools/src/tool_registry_plan.rs
sayan-oai: added check at deferral time to make sure tool is not already present in direct tools (sets must be disjoint).

PR #17417, opened 2026-04-11
Emit plan-mode prompt notifications for questionnaires

etraut-openai:
Addresses #17252

Problem: Plan-mode clarification questionnaires used the generic user-input notification type, so configs listening for plan-mode-prompt did not fire when request_user_input waited for an answer.

Solution: Map request_user_input prompts to the plan-mode-prompt notification and remove the obsolete user-input TUI notification variant.


fcoury-oai (approved):
Tested and the proper notification was triggered when asking a question on plan mode.

Code also looks good.

PR #17425, opened 2026-04-11
Auto-upgrade configured marketplaces

xli-oai:
## Summary
- Add best-effort auto-upgrade for user-configured Git marketplaces recorded in `config.toml`.
- Track the last activated Git revision with `last_revision` so unchanged marketplace sources skip clone work.
- Trigger the upgrade from plugin startup and `plugin/list`, while preserving existing fail-open plugin behavior with warning logs rather than new user-visible errors.

## Details
- Remote configured marketplaces use `git ls-remote` to compare the source/ref against the recorded revision.
- Upgrades clone into a staging directory, validate that `.agents/plugins/marketplace.json` exists and that the manifest name matches the configured marketplace key, then atomically activate the new root.
- Local `.agents/plugins/marketplace.json` marketplaces remain live filesystem state and are not auto-pulled.
- Existing non-curated plugin cache refresh is kicked after successful marketplace root upgrades.

## Validation
- `just write-config-schema`
- `cargo test -p codex-core marketplace_upgrade`
- `cargo check -p codex-cli -p codex-app-server`
- `just fix -p codex-core`

Did not run the complete `cargo test` suite because the repo instructions require asking before a full core workspace run.


codex-rs/core/src/plugins/manager.rs
```diff
@@ -1032,6 +1054,7 @@ impl PluginsManager {
     ) {
         if config.features.enabled(Feature::Plugins) {
             self.start_curated_repo_sync();
+            self.maybe_start_configured_marketplace_upgrade_for_config(config);
```
xl-openai: You can move this into maybe_start_plugin_startup_tasks_for_config

codex-rs/core/src/plugins/manager.rs
```diff
@@ -1056,30 +1079,116 @@ impl PluginsManager {
...
+            let mut state = match self.configured_marketplace_upgrade_state.write() {
+                Ok(state) => state,
+                Err(err) => err.into_inner(),
+            };
+            state.in_flight = false;
+            warn!("failed to start configured marketplace auto-upgrade task: {err}");
+        }
+    }
+
     pub fn maybe_start_non_curated_plugin_cache_refresh_for_roots(
```
xl-openai: I’d prefer not to introduce a thin wrapper like this

codex-rs/core/src/plugins/manager.rs
xli-oai: Addressed in `72b4885e4` by folding the configured marketplace auto-upgrade startup directly into `maybe_start_plugin_startup_tasks_for_config`, so that extra single-use helper is gone.

codex-rs/core/src/plugins/manager.rs
xli-oai: Addressed in `72b4885e4` by removing the thin `force_reinstall` wrapper and calling the mode-bearing refresh helper directly from the upgrade path.

xl-openai (approved):

xli-oai:
@codex review

xli-oai:
@codex review

PR #17470, opened 2026-04-11
Fix custom tool output cleanup on stream failure

etraut-openai:
Addresses #16255

Problem: Incomplete Responses streams could leave completed custom tool outputs out of cleanup and retry prompts, making persisted history inconsistent and retries stale.

Solution: Route stream and output-item errors through shared cleanup, and rebuild retry prompts from fresh session history after the first attempt.


jif-oai:
@codex review

jif-oai (approved):

Issue #17453, opened 2026-04-11
Incorrect truncation of reset times in /status (CLI, macOS app, VS Code extension)


matii-dev:
### What issue are you seeing?

The `/status` (or `/stan`) command incorrectly displays reset times for usage limits.

- 5h limit reset shows only partial hour (e.g. "14" instead of full time like "14:32")
- 7 day limit reset shows only first digit of the day (e.g. "2" instead of "21")
- this makes the information effectively unusable

### Where does it occur?

- Codex CLI
- Codex macOS app
- Codex VS Code extension

### Expected behavior

Full timestamp should be displayed (date + time), not truncated.

### Why this matters

The `/status` command becomes partially useless because users cannot determine when limits reset.

### Additional notes

This appears to be a shared UI/formatting issue rather than CLI-specific, since it occurs across multiple Codex interfaces.

Possibly related to string truncation, layout width, or formatting logic.

<img width="631" height="304" alt="Image" src="https://github.com/user-attachments/assets/46d0d2c5-025a-489e-9224-85a8919f193d" />

aosen-xiong:
Hi @etraut-openai, thanks for looking into this. Just wondering: was this issue fixed in #17481 and already released? I am still experiencing it on version 26.506.31421 (2620).

etraut-openai:
Yes, this was fixed. Closing

@aosen-xiong, it sounds like you're using the app? This bug was specifically about the CLI (TUI). If you're seeing a problem in the app, please file a new bug report.

aosen-xiong:
Okay, opened #22617. Thanks a lot.

PR #17481, opened 2026-04-11
Wrap status reset timestamps in narrow layouts

etraut-openai:
Addresses #17453

Problem: /status rate-limit reset timestamps can be truncated in narrow layouts, leaving users with partial times or dates.

Solution: Let narrow rate-limit rows drop the fixed progress bar to preserve the percent summary, and wrap reset timestamps onto continuation lines instead of truncating them.


fcoury-oai (commented):
Image

Works great. Approved.

fcoury-oai (approved):

PR #17486, opened 2026-04-11
changing decision semantics after guardian timeout

won-openai:
**Summary**

This PR treats Guardian timeouts as distinct from explicit denials in the core approval paths.
Timeouts now return timeout-specific guidance instead of Guardian policy-rejection messaging.
It updates the command, shell, network, and MCP approval flows and adds focused test coverage.

### DIFF
--- a/codex-rs/core/src/guardian/review.rs
+++ b/codex-rs/core/src/guardian/review.rs
@@ -38,6 +38,12 @@ const GUARDIAN_REJECTION_INSTRUCTIONS: &str = concat!(
     "Otherwise, stop and request user input.",
 );
 
+const GUARDIAN_TIMEOUT_INSTRUCTIONS: &str = concat!(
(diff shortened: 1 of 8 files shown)


codex-rs/core/src/guardian/review.rs
```diff
@@ -38,6 +38,12 @@ const GUARDIAN_REJECTION_INSTRUCTIONS: &str = concat!(
     "Otherwise, stop and request user input.",
 );
 
+const GUARDIAN_TIMEOUT_INSTRUCTIONS: &str = concat!(
+    "The automatic approval review did not finish before its deadline. ",
+    "Do not assume the action is unsafe based on the timeout alone. ",
```
won-openai, line 43: thoughts on what the instruction shuold be?

dylan-hurd-oai (approved):

PR #17489, opened 2026-04-11
Moving updated-at timestamps to unique millisecond times

ddr-oai:
To allow the ability to have guaranteed-unique cursors, we make two important updates:
* Add new updated_at_ms and created_at_ms columns that are in millisecond precision
* Guarantee uniqueness -- if multiple items are inserted at the same millisecond, bump the new one by one millisecond until it becomes unique

This lets us use single-number cursors for forwards and backwards paging through resultsets and guarantee that the cursor is a fixed point to do (timestamp > cursor) and get new items only.

This updated implementation is backwards-compatible since multiple appservers can be running and won't handle the previous method well.

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -9437,7 +9437,7 @@ fn summary_from_thread_metadata(metadata: &ThreadMetadata) -> ConversationSummar
             .to_rfc3339_opts(SecondsFormat::Secs, true),
         metadata
             .updated_at
-            .to_rfc3339_opts(SecondsFormat::Secs, true),
+            .to_rfc3339_opts(SecondsFormat::Millis, true),
         metadata.model_provider.clone(),
         metadata.cwd.clone(),
         metadata.cli_version.clone(),
@@ -9703,7 +9703,7 @@ async fn read_updated_at(path: &Path, created_at: Option<&str>) -> Option<String
         .and_then(|meta| meta.modified().ok())
         .map(|modified| {
             let updated_at: DateTime<Utc> = modified.into();
-            updated_at.to_rfc3339_opts(SecondsFormat::Secs, true)
+            updated_at.to_rfc3339_opts(SecondsFormat::Millis, true)
         });
     updated_at.or_else(|| created_at.map(str::to_string))
 }
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -9956,7 +9956,8 @@ mod tests {
 
     #[test]
     fn summary_from_thread_metadata_formats_protocol_timestamps_as_seconds() -> Result<()> {
-        let mut metadata = test_thread_metadata(None, None)?;
+        let mut metadata =
+            test_thread_metadata(/*model*/ None, /*reasoning_effort*/ None)?;
         metadata.created_at =
             DateTime::parse_from_rfc3339("2025-09-05T16:53:11.123Z")?.with_timezone(&Utc);
         metadata.updated_at =
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -9437,7 +9437,7 @@ fn summary_from_thread_metadata(metadata: &ThreadMetadata) -> ConversationSummar
             .to_rfc3339_opts(SecondsFormat::Secs, true),
         metadata
             .updated_at
-            .to_rfc3339_opts(SecondsFormat::Millis, true),
+            .to_rfc3339_opts(SecondsFormat::Secs, true),
         metadata.model_provider.clone(),
         metadata.cwd.clone(),
         metadata.cli_version.clone(),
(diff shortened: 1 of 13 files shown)


codex-rs/state/src/runtime/threads.rs
```diff
@@ -599,14 +601,61 @@ ON CONFLICT(id) DO NOTHING
...
+        let candidate = datetime_to_epoch_millis(updated_at);
+        let allocated = loop {
+            let current = self.thread_updated_at_millis.load(Ordering::Relaxed);
+
+            // New wall-clock time: advance the process-local high-water mark and use it as-is.
+            if candidate > current {
+                if self
+                    .thread_updated_at_millis
+                    .compare_exchange(current, candidate, Ordering::Relaxed, Ordering::Relaxed)
+                    .is_ok()
```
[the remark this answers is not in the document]
ddr-oai: @jif-oai this is the crux question -- is it an invariant that only one agent process will ever write to a single thread at the same time?  my understanding is that subagents would be on their own threads.  if that's not the case then i need to go back to my original implementation that always checked the database in a transaction for a matching timestamp.  just trying to avoid extra database operations that would (very) slightly slow down conversations.

ddr-oai:

jif-oai (approved):
Going to ms feels a bit like a hack to me (as it's still not a perfect guarantee) but I guess in this case this is more than enough
Also, this won't fix the backward compatibility case

codex-rs/state/src/runtime/threads.rs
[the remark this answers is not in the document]
jif-oai: This is a good point actually... each agent have their own thread indeed

codex-rs/state/migrations/0025_thread_timestamps_millis.sql
```diff
@@ -0,0 +1,80 @@
...
+    SET created_at_ms = NEW.created_at * 1000
+    WHERE id = NEW.id;
+END;
+
+CREATE TRIGGER threads_updated_at_ms_after_insert
+AFTER INSERT ON threads
+WHEN NEW.updated_at_ms IS NULL
+BEGIN
+    UPDATE threads
+    SET updated_at_ms = NEW.updated_at * 1000
```
jif-oai, line 87: This can still induce collisions on older rollouts but I guess we don't have a choice here

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -9436,7 +9436,7 @@ fn summary_from_thread_metadata(metadata: &ThreadMetadata) -> ConversationSummar
             .to_rfc3339_opts(SecondsFormat::Secs, true),
         metadata
             .updated_at
-            .to_rfc3339_opts(SecondsFormat::Secs, true),
+            .to_rfc3339_opts(SecondsFormat::Millis, true),
```
jif-oai: This preserves milliseconds in the summary string, but thread/list still emits updatedAt as seconds. So any reason for this?

codex-rs/app-server/BUILD.bazel
```diff
@@ -3,6 +3,7 @@ load("//:defs.bzl", "codex_rust_crate")
 codex_rust_crate(
     name = "app-server",
     crate_name = "codex_app_server",
+    integration_test_args = ["--test-threads=1"],
```
jif-oai: huh?

codex-rs/core/src/plugins/discoverable_tests.rs
```diff
@@ -200,6 +200,7 @@ async fn list_tool_suggest_discoverable_plugins_does_not_reload_marketplace_per_
     assert_eq!(discoverable_plugins[0].id, "slack@openai-curated");
 
     let logs = String::from_utf8(buffer.lock().expect("buffer lock").clone()).expect("utf8 logs");
+    let logs = logs.replace('\\', "/");
```
jif-oai: huh?

codex-rs/exec-server/tests/common/exec_server.rs
```diff
@@ -22,7 +22,7 @@ use tokio::time::timeout;
 use tokio_tungstenite::connect_async;
 use tokio_tungstenite::tungstenite::Message;
 
-const CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
+const CONNECT_TIMEOUT: Duration = Duration::from_secs(10);
```
jif-oai: This should have been fixed by another PR

codex-rs/exec-server/tests/exec_process.rs
```diff
@@ -153,9 +153,9 @@ async fn assert_exec_process_write_then_read(use_remote: bool) -> Result<()> {
         .start(ExecParams {
             process_id: process_id.clone().into(),
             argv: vec![
-                "/usr/bin/python3".to_string(),
+                "/bin/sh".to_string(),
```
jif-oai: unrelated

jif-oai (approved):
Can you revert all the unrelated changes or move them to a dedicated PR?

codex-rs/core/src/plugins/discoverable_tests.rs
ddr-oai: that test broke on windows

codex-rs/exec-server/tests/exec_process.rs
ddr-oai: was a broken test

codex-rs/app-server/BUILD.bazel
ddr-oai: the test suit was doing things in parallel with a single appserver so it had a race condition

codex-rs/app-server/src/codex_message_processor.rs
ddr-oai: ah oops good catch

codex-rs/state/migrations/0025_thread_timestamps_millis.sql
ddr-oai: fixed

PR #17556, opened 2026-04-12
[codex] Support flattened deferred MCP tool calls

fc-oai:
## Summary
- register flattened handler aliases for deferred MCP tools
- cover the node_repl-shaped deferred MCP call path in tool registry tests

## Root Cause
Deferred MCP tools were registered only under their namespaced handler key, e.g. `mcp__node_repl__:js`. If the model/bridge emitted the flattened qualified name `mcp__node_repl__js`, core parsed it as an MCP payload but dispatch looked up the flattened handler key and returned `unsupported call` before reaching the MCP handler.

## Validation
- `just fmt`
- `cargo test -p codex-tools search_tool_registers_deferred_mcp_flattened_handlers`
- `cargo test -p codex-core search_tool_registers_namespaced_mcp_tool_aliases`
- `git diff --check`

### DIFF
--- a/codex-rs/core/src/tools/spec.rs
+++ b/codex-rs/core/src/tools/spec.rs
@@ -263,6 +263,15 @@ pub(crate) fn build_specs_with_discoverable_tools(
             }
         }
     }
+    if let Some(deferred_mcp_tools) = deferred_mcp_tools.as_ref() {
+        for (name, _) in deferred_mcp_tools.iter().filter(|(name, _)| {
+            !mcp_tools
+                .as_ref()
+                .is_some_and(|tools| tools.contains_key(*name))
+        }) {
+            builder.register_handler(name.clone(), mcp_handler.clone());
+        }
+    }
     builder
 }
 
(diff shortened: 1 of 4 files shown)


dylan-hurd-oai (approved):

raye-deng:
Good catch on the flattened handler alias gap. This is a classic "namespacing convention mismatch" problem — the tool registry expects one naming convention (`mcp__node_repl__:js`) while the model/bridge emits another (`mcp__node_repl__js`), and neither side validates that the other convention exists.

What makes these bugs particularly tricky in AI agent systems is that the failure mode is a silent "unsupported call" rather than an explicit error. The agent thinks it called a tool, gets back a non-descript failure, and either retries (burning tokens) or silently falls back to a different approach (producing subtly wrong output).

We have seen similar issues where AI-generated code references MCP tools using slightly different namespacing conventions than what the runtime expects. The code looks correct (the tool name is "almost" right), and unit tests pass (because mocks use the same convention as the test), but production fails because the actual MCP server uses the other convention.

A practical mitigation is to add a tool-name normalization step that canonicalizes namespaced identifiers before dispatch, plus a registration-time check that warns about potential alias collisions. We include this kind of MCP protocol validation in our CI scans.

npx @opencodereview/cli scan . --sla L1

PR #17570, opened 2026-04-12
fix: protect active arg0 helper dirs

### LINKED ISSUE #16970: Unified exec caches stale ~/.codex/tmp/arg0 session path and fails with ENOENT until app restart

codex 26.325.31654 app
codex-cli 0.118.0-alpha.2

plus

gpt-5.3 Codex (Codex desktop default coding model)

Windows 11 + WSL2  
Linux kernel: `Linux 6.6.87.2-microsoft-standard-WSL2 x86_64`

In an active Codex Desktop thread, non-escalated tool execution suddenly fails with ENOENT from unified exec.  
The failing spawn target points to a temp runtime path under `~/.codex/tmp/arg0/<session>/codex-linux-sandbox` that no longer exists.

After restarting the app, a new `codex-arg0*` runtime folder is created and tool execution works again.

Intermittent, but this sequence reproduces it for me:

1. Start Codex Desktop in a WSL-backed workspace.
2. Use normal tool calls for a while (`exec_command`, `apply_patch`, etc.).
3. At some point, all non-escalated `exec_command` calls fail immediately.
4. Error references a specific stale temp path:
   `.../tmp/arg0/codex-arg06KNIXk/codex-linux-sandbox ... ENOENT`
5. Confirm that folder no longer exists.
6. Restart Codex Desktop.
7. A new folder appears (example: `codex-arg0nHm5Gj`) containing `codex-linux-sandbox`, and commands succeed again.

If an ephemeral `codex-arg0*` runtime folder is missing/rotated, unified exec should automatically rebind/recreate sandbox shims instead of hard-failing until full app restart.

### What do you see instead?

Exact error captured in session log:

`Failed to create unified exec process: Unable to spawn /mnt/c/Users/<redacted>/.codex/tmp/arg0/codex-arg06KNIXk/codex-linux-sandbox because it doesn't exist on the filesystem (ENOENT: No such file or directory)`

### Additional information

I checked related issues before filing:
- #3947 (`codex-linux-sandbox ... not provided` on resumed sessions)
- #4754 (generic `os error 2`)
- #2252 (`os error 2` after env/path changes)

This report is specifically about a stale `~/.codex/tmp/arg0/codex-arg*` runtime path being referenced by unified exec during an active desktop session.

Local evidence:
- Failing log:
  `/mnt/c/Users/<redacted>/.codex/sessions/2026/04/07/rollout-2026-04-07T10-37-51-019d655f-d19c-75e0-9fea-dc5638991fd9.jsonl`  
  (entry timestamp `2026-04-07T01:06:04.320Z`)
- Missing stale folder after failure:
  `~/.codex/tmp/arg0/codex-arg06KNIXk` (not found)
- Healthy after restart:
  `~/.codex/tmp/arg0/codex-arg0nHm5Gj/codex-linux-sandbox -> ~/.codex/bin/wsl/codex`
```
### Sanitized diagnostic excerpt

[2026-04-07T01:06:04.320Z]
exec_command failed for `/bin/bash -lc ''`: CreateProcess {
  message: "Rejected(\"Failed to create unified exec process: Unable to spawn /mnt/c/Users/<redacted>/.codex/tmp/arg0/codex-arg06KNIXk/codex-linux-sandbox because it doesn't exist on the filesystem (ENOENT: No such file or directory)\")"
}

[2026-04-07T01:09:36.879Z]
Command: /bin/bash -lc 'ls -la /mnt/c/Users/<redacted>/.codex/tmp/arg0/codex-arg06KNIXk 2>/dev/null || true'
Output: <empty>  (folder missing)

[2026-04-07T01:17:00.619Z]
Command: /bin/bash -lc 'ls -la /mnt/c/Users/<redacted>/.codex/tmp/arg0'
Output:
  ... codex-arg0nHm5Gj

[2026-04-07T01:17:21.860Z]
Command: /bin/bash -lc 'ls -la /mnt/c/Users/<redacted>/.codex/tmp/arg0/codex-arg0nHm5Gj'
Output:
  ... codex-linux-sandbox -> /mnt/c/Users/<redacted>/.codex/bin/wsl/codex

[2026-04-07T01:18:04Z]
`pwd` and `node -v` succeed again in the same workspace after restart.

Environment:
- codex-cli 0.118.0-alpha.2
- Windows 11 + WSL2

``` 

Root-cause hypothesis:
- Unified exec/harness caches a per-session sandbox shim path under `tmp/arg0`.
- If that ephemeral folder is removed/rotated, the cached path is reused and all non-escalated spawns fail with ENOENT.

High-level fix suggestion:
- On ENOENT for the sandbox shim path, force a one-time runtime rehydrate/rebind:
  - recreate the per-session symlink in a fresh `codex-arg0*` folder, or
  - fall back to `~/.codex/bin/wsl/codex` and rebuild session shims.

viyatb-oai:
## Summary

This keeps live arg0 helper directories from being removed during startup cleanup. Each new helper dir now writes a `.pid` owner sentinel, and the janitor skips directories whose owner process is still running before falling back to the existing lock-based stale cleanup.

## Issues Fixed

- Fixes #16970: prevents Codex Desktop on Windows + WSL2 from caching a now-deleted `~/.codex/tmp/arg0/codex-arg0*/codex-linux-sandbox` helper path and failing non-escalated unified exec / `apply_patch` calls with ENOENT until app restart.
- Expected to address the same stale arg0-wrapper failure mode discussed in #16791 when it is caused by active helper directories being cleaned up during workspace/session switches. I am not marking #16791 as closing here because the reports may not be identical.

## Root Cause

Codex Desktop sessions can cache helper paths such as `~/.codex/tmp/arg0/codex-arg0*/codex-linux-sandbox`. The arg0 janitor previously trusted the helper directory lock as the only signal that another process still owned the directory. On WSL-backed workspaces with `CODEX_HOME` on the Windows-mounted home, those advisory locks can be unreliable, so a later startup could remove another live session's helper directory. Non-escalated unified exec and `apply_patch` then fail with ENOENT until the app restarts.


eocodn:
Additional confirmation from Windows Desktop + WSL mode.

Environment observed locally:
- Codex Desktop: `OpenAI.Codex_26.527.3686.0_x64__2p2nqsd0c76g0`
- WSL distro: Ubuntu 24.04
- Codex WSL binary: `codex-cli 0.135.0-alpha.1`
- `CODEX_HOME` is on the Windows-mounted path under `/mnt/c/Users/<user>/.codex`

Observed the same stale arg0 helper-dir failure:

- Non-escalated tool calls failed before command execution with `Failed to create unified exec process: No such file or directory (os error 2)`.
- The live app-server process still held an fd to a deleted lock path like:
  `/proc/<app-server-pid>/fd/3 -> /mnt/c/Users/<user>/.codex/tmp/arg0/codex-arg0VHZzae/.lock`
- `fdinfo` showed that fd still had a POSIX advisory write lock owned by the app-server PID, but the `.lock` path was not stat-able from WSL.
- The app-server appeared to retain the in-memory helper path to the missing `codex-linux-sandbox` under that same `codex-arg0*` directory.
- Recreating that exact cached helper directory from the Windows side with `cmd.exe mkdir ...`, then restoring the Linux helper symlinks (`apply_patch`, `applypatch`, `codex-execve-wrapper`, `codex-linux-sandbox` -> current WSL Codex binary), immediately restored non-escalated `pwd` in the same active session.

This strongly matches the root cause described in this PR: a live session can cache an arg0 helper path whose directory has been removed even though the process still believes its `.lock` is held. In this WSL/DrvFS state, relying only on advisory locks was not enough. The `.pid` owner sentinel proposed here looks like the right class of fix for this failure mode.

MisterRound:
Additional confirmation from a current Windows Desktop + WSL failure on `OpenAI.Codex_26.527.3686.0_x64__2p2nqsd0c76g0` / WSL app-server `codex-cli 0.135.0-alpha.1`.

This PR matches the core failure class: the active app-server can retain a cached `tmp/arg0` helper path whose directory has been removed.

Current live evidence:

```text
PATH contains:
/mnt/c/Users/<user>/.codex/tmp/arg0/codex-arg0UXHjoV

stat:
No such file or directory

/proc/<app-server-pid>/fd/3:
/mnt/c/Users/<user>/.codex/tmp/arg0/codex-arg0UXHjoV/.lock

fdinfo:
POSIX ADVISORY WRITE lock owned by the live app-server pid
```

The active thread remains broken even if a newer `codex-arg0*` directory is created later, because the running app-server/thread state is still pinned to the deleted helper path.

One extra concern from source review: this branch's `.pid` sentinel is the right class of fix for Linux-side cleanup, but `owner_pid_is_running()` returns `false` on non-Linux. In Windows Desktop + WSL, `CODEX_HOME` is shared under `%USERPROFILE%\.codex` / `/mnt/c/Users/<user>/.codex`, and Windows-native Codex processes can also touch the same `tmp/arg0` tree. If a Windows janitor sees a WSL-owned helper dir, a non-Linux liveness result of `false` may still allow deletion of an active WSL helper dir.

So I think the durable fix needs either:

- platform/runtime namespacing under `tmp/arg0`, or
- a helper manifest with platform/runtime/WSL distro identity, with janitors restricted to their own runtime family, or
- a conservative cross-runtime rule where Windows does not delete helper dirs it cannot prove are Windows-owned/stale.

I opened a focused regression issue with the current evidence here: https://github.com/openai/codex/issues/25317

MisterRound:
Follow-up with a sharper mixed-runtime repro from #25317.

I tested the proposed recovery behavior in a live Windows Desktop + WSL session:

- Recreating the exact stale `CODEX_HOME/tmp/arg0/codex-arg0*` path with Linux symlinks to the active WSL Codex binary immediately restored non-escalated shell calls in the same active thread.
- `pwd`, `uname -a`, `command -v codex-linux-sandbox`, and `apply_patch` all worked after rehydration.
- Running the Windows-native bundled `codex.exe` afterward deleted that repaired WSL helper dir again and broke normal shell calls again.
- A new Windows `.bat` arg0 helper dir appeared under the same shared `CODEX_HOME/tmp/arg0` root.
- Holding a WSL `flock` on the repaired `.lock` did not protect the helper dir from the Windows-native cleanup path.

This is the important implication for this PR: a `.pid` sentinel is the right class of fix for Linux-side cleanup, but it likely needs cross-runtime semantics. In Windows Desktop + WSL, the Windows-native Codex process and WSL app-server are both using `%USERPROFILE%/.codex/tmp/arg0`. If the Windows janitor cannot verify that a WSL PID is dead, it should not delete that helper dir. Better still, Windows and WSL helpers should be placed in separate runtime/platform namespaces so their janitors do not sweep each other.

I also observed a DrvFS/open-delete oddity around the stale helper path: WSL `stat` reports missing and Windows PowerShell `Get-Item` reports missing, but WSL `mkdir()` on the same path can return `EEXIST`. Creating the directory from Windows PowerShell works around that recreate state, after which WSL symlink population can proceed.

MisterRound:
Follow-up after comparing this PR's diff with the newer Windows Desktop + WSL repros in #25317, #25280, and #25370.

I think this PR is still the right direction for the stale-helper class, but I do not think it is sufficient as written for the deeper mixed-runtime regression in #25317.

The current diff writes a `.pid` sentinel and skips cleanup when `owner_pid_is_running(pid)` is true. That check only proves liveness on Linux via `/proc`; on non-Linux it returns `false`. In Windows Desktop + WSL, both runtimes can share the same `%USERPROFILE%/.codex/tmp/arg0` root:

- WSL app-server creates Linux helper dirs with symlinks such as `codex-linux-sandbox`, `apply_patch`, etc.
- Windows-native Codex/plugin paths can create `.bat` helper dirs under the same root.
- In #25317, rehydrating the exact stale WSL helper dir restored the live WSL thread, but invoking the Windows-native bundled `codex.exe` deleted that repaired WSL helper dir again.
- Holding a WSL `flock` on `.lock`, and separately holding a WSL process cwd inside the helper dir, did not protect it from Windows-native cleanup.

So the important missing condition is cross-runtime ownership. If a Windows janitor sees a WSL-owned helper dir, it should either be unable to select it for cleanup, or it should conservatively skip it when it cannot prove that the owning runtime is dead. A `.pid` sentinel helps Linux-side cleanup, but without runtime/platform namespacing or a manifest that records `windows` vs `wsl/linux` ownership, the Windows path can still treat an active WSL helper dir as stale.

Suggested durable shape:

- namespace helper dirs by runtime family, for example `tmp/arg0/windows/...` vs `tmp/arg0/wsl/<distro-or-runtime-id>/...`; or
- write a helper manifest with platform/runtime identity and require janitors to sweep only their own family; and
- treat unknown or foreign-runtime helper dirs as live unless there is a runtime-specific way to prove staleness.

This distinction matters for triage: landing this PR may improve the original stale-lock case, but it should not close #25317 unless the cross-runtime Windows-native-vs-WSL deletion path is also covered.

PR #17573, opened 2026-04-12
feat: add sandbox violation monitoring in codex-sandboxing

viyatb-oai:
## Why

Codex already has one load-bearing classifier for sandbox denials: `is_likely_sandbox_denied`. Filesystem denials and managed-network blocks did not share a structured event shape, so every persistence, telemetry, or product consumer would otherwise need to rediscover the same enforcement paths and parse backend-specific output.

This PR adds that shared seam without changing sandbox selection or denial behavior. `SandboxType::LinuxSeccomp` and `use_legacy_landlock` stay as-is. Linux violations use `bubblewrap` as the audit backend while execution continues to use the existing sandbox type.

## What changed

- Added a normalized `SandboxViolationEvent` with filesystem and network variants in `codex-sandboxing`.
- Classified filesystem denials into stable backend and reason fields, with an optional denied path and a bounded output snippet.
- Converted managed-network proxy blocks into the same event model while preserving host, port, protocol, mode, decision, and source context.
- Recorded filesystem events only after the existing `is_likely_sandbox_denied` gate in exec, apply-patch, shell-escalation, and unified-exec paths.
- Preserved the classifier's comments, keyword ordering, quick rejects, and `SIGSYS` behavior.
- Added a single tracing seam for downstream persistence and export. This PR does not add durable storage.

## Stack

1. #17573 (this PR): normalize filesystem and network sandbox violations in `codex-sandboxing`
2. #22866: persist the normalized events in bounded SQLite and emit OTel audit events

## Testing

- `just test -p codex-sandboxing`
- `just test -p codex-core sandbox_detection --lib`

### DIFF
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -919,8 +919,7 @@ pub struct Config {
     pub codex_self_exe: Option<PathBuf>,
 
     /// Path to the `codex-linux-sandbox` executable. This must be set if
-    /// [`codex_sandboxing::SandboxType::LinuxBubblewrap`] or
-    /// [`codex_sandboxing::SandboxType::LinuxLegacyLandlock`] is used. Note that this
+    /// [`codex_sandboxing::SandboxType::LinuxSeccomp`] is used. Note that this
     /// cannot be set in the config file: it must be set in code via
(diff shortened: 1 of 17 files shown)


rreichel3-oai (approved):
lgtm!

PR #17578, opened 2026-04-12
Add injected message transcript events (2 of 4)

etraut-openai:
Stack position: This is PR 2 of 4 in a stack.

Adds the generated-message representation layer: model-visible XML envelopes for timer and external messages, transcript-safe injected-message events, and app-server/TUI rendering support. This PR does not add timer scheduling or the queue CLI.


codex-rs/core/src/pending_input.rs
```diff
@@ -0,0 +1,47 @@
+//! Pending input queued for delivery into a future or active regular turn.
+//!
+//! Most pending input is an ordinary model input item. Generated input carries
```
etraut-openai: This comment refers to "generated input", but the code uses the term "injected event". I think this is a stale comment. Update it.

PR #17581, opened 2026-04-12
Add timer and queued message state storage (1 of 4)

etraut-openai:
Stack position: This is PR 1 of 4 in a stack.

Adds durable SQLite storage and claim APIs for thread timers and queued thread messages. This PR is storage-only: it does not deliver timers/messages into sessions, expose timer tools, or add `codex queue`.


codex-rs/state/src/model/thread_message.rs
```diff
@@ -0,0 +1,85 @@
+use sqlx::FromRow;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ThreadMessageCreateParams {
```
etraut-openai: I'm not crazy about the term "thread message" here. Can we come up with a better term like "external message"? This should then propagate through the name of the module and the symbols.

Issue #17514, opened 2026-04-12
Multiple context compactions happening at the same time


guidedways:
### What version of Codex CLI is running?

0.120.0

### What subscription do you have?

Pro

### Which model were you using?

gpt-5.4

### What platform is your computer?

macOS

### What terminal emulator and version are you using (if applicable)?

iTerm

### What issue are you seeing?

```
• Context compacted

• Context compacted
```

I see this every time the context auto compacts. Two compactions oddly.

### What steps can reproduce the bug?

Just let it auto compact.

etraut-openai:
I'm guessing that this is a display issue rather than two compactions. Could you confirm by looking at the rollout file? Or use `/feedback` to upload it so I can take a look?

guidedways:
Sure: `019d852c-c0d9-7ba2-bd55-b0e8243514e6`

Easy to reproduce:

1) Launch Codex
2) Type `/compact`

You'll see two messages. I also see the same for other messages like

```
• Waited for background terminal

• Waited for background terminal
```

etraut-openai:
Thanks, I'm able to repro the problem.

etraut-openai:
This will be addressed in the next release.

PR #17601, opened 2026-04-13
Suppress duplicate compaction and terminal wait events

etraut-openai:
Addresses #17514

Problem: PR #16966 made the TUI render the deprecated context-compaction notification, while v2 could also receive legacy unified-exec interaction items alongside terminal-interaction notifications, causing duplicate "Context compacted" and "Waited for background terminal" messages.

Solution: Suppress deprecated context-compaction notifications and legacy unified-exec interaction command items from the app-server v2 projection, and render canonical context-compaction items through the existing TUI info-event path.


fcoury-oai (approved):
Looks good.

PR #17606, opened 2026-04-13
1/4 python-runtime: prepare CLI bin package

shaqayeq-oai:
TL;DR: Prepare the Python runtime wheel package for production publishing.

- Rename the runtime distribution while keeping the import module stable.
- Stage complete platform runtime bundles, including Windows helper executables.
- Add runtime path helpers and staging/version tests.

PR #17607, opened 2026-04-13
2/4 python-sdk: prepare Python package

shaqayeq-oai:
TL;DR: Prepare the Python SDK distribution for the two-package publish flow.

- Rename the SDK distribution metadata while keeping imports stable.
- Generate Python protocol types from the pinned runtime schema.
- Update docs for the final install and versioning shape.


shaqayeq-oai:
This change is part of the following stack:

- #17606
    - #17607 ◀
        - #17608
            - #17609

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

PR #17611, opened 2026-04-13
Fix Windows CI flake: shell timeout race

etraut-openai:
I've been seeing a lot of Windows test flakes lately. This appears to be one of the main causes.

Problem: Windows CI could classify a shell command as timed out when expiration won the executor race after the child had already exited and produced output.

Solution: Check for an already-exited child before killing on expiration, add regression coverage for that race, and use a lightweight cmd.exe marker command in the realtime test. Switching to cmd.exe should reduce a lot of startup time variance for tests.


codex-rs/core/src/exec.rs
```diff
@@ -1241,15 +1241,20 @@ async fn consume_output(
     };
     tokio::pin!(expiration_wait);
     let (exit_status, timed_out) = tokio::select! {
+        biased;
+        _ = &mut expiration_wait => {
```
etraut-openai, line 1246: Add a comment here explaining what this code is doing.

PR #17626, opened 2026-04-13
feat: disable memory endpoint

jif-oai:


jif-oai:
@codex review

dylan-hurd-oai (approved):
Should this also be an option on thread/start? Or would that just be done via config?

jif-oai:
> Should this also be an option on thread/start? Or would that just be done via config?

Already via config

PR #17632, opened 2026-04-13
feat: memories menu

jif-oai:
Add menu that:
1. If memories feature is not enabled, propose to enable it
2. Let you choose if you want to generate memories and to use memories


codex-rs/tui/src/app.rs
```diff
@@ -1466,6 +1466,48 @@ impl App {
         }
     }
 
+    async fn update_memory_settings(&mut self, use_memories: bool, generate_memories: bool) {
+        let active_profile = self.active_profile.clone();
+        let scoped_memory_segments = |key: &str| {
+            if let Some(profile) = active_profile.as_deref() {
+                vec![
```
fcoury-oai, line 1489: I may have found a potential issue. 

It seems like we are writing the memory settings to `profile.<profile>.memories` and `ConfigProfile` doesn't define a memories field.

So from a user's perspective it looks like they were saved, but they would be ignored on the next restart.

Should this be saved globally for now, or should profile-level memories support be added end-to-end in the config model/schema first?

fcoury-oai (changes requested):

fcoury-oai (approved):
Profiles will be nuked and the rest looks good!

jif-oai:
@codex review

jif-oai:
@codex review

PR #17654, opened 2026-04-13
Support Unix socket allowlists in macOS sandbox

aaronl-openai:
## Changes

Allows sandboxes to restrict overall network access while granting access to specific unix sockets on mac.

## Details

- `codex sandbox macos`: adds a repeatable `--allow-unix-socket` option.
- `codex-sandboxing`: threads explicit Unix socket roots into the macOS Seatbelt profile generation.
- Preserves restricted network behavior when only Unix socket IPC is requested, and preserves full network behavior when full network is already enabled.

## Verification

- `cargo test -p codex-cli -p codex-sandboxing`
- verified that `codex sandbox macos --allow-unix-socket /tmp/test.sock -- test-client` grants access as expected


codex-rs/sandboxing/src/seatbelt.rs
```diff
@@ -465,7 +517,7 @@ pub fn create_seatbelt_command_args_for_policies(
             }
         };
 
-    let proxy = proxy_policy_inputs(network);
+    let proxy = proxy_policy_inputs(network).with_extra_unix_sockets(extra_allow_unix_sockets);
```
bolinfest: `proxy_policy_inputs()` has only one callsite. Why not update it to take `extra_allow_unix_sockets` so that we don't have to introduce `with_extra_unix_sockets()` which takes `mut self`?

codex-rs/sandboxing/src/seatbelt.rs
```diff
@@ -381,6 +413,26 @@ pub fn create_seatbelt_command_args_for_policies(
...
+        file_system_sandbox_policy,
+        network_sandbox_policy,
+        sandbox_policy_cwd,
+        enforce_managed_network,
+        network,
+        &[],
+    )
+}
+
+pub fn create_seatbelt_command_args_for_policies_with_extra_unix_sockets(
```
bolinfest: Maybe this also merits using a builder pattern, but we have these three things calling each other right now, each adding one param:

`create_seatbelt_command_args()` -> 
`create_seatbelt_command_args_for_policies()` -> `create_seatbelt_command_args_for_policies_with_extra_unix_sockets()`

This feels a bit much: can we just have one function that takes a struct with the full set of arguments?

If we feel strongly, we could make the struct impl `Default` so it's easier to elide the arguments with trivial default values when constructing it.

codex-rs/cli/src/lib.rs
```diff
@@ -34,6 +39,11 @@ pub struct SeatbeltCommand {
     pub command: Vec<String>,
 }
 
+fn parse_absolute_path(raw: &str) -> Result<AbsolutePathBuf, String> {
+    AbsolutePathBuf::from_absolute_path_checked(raw)
```
bolinfest:
```suggestion
    AbsolutePathBuf::relative_to_current_dir(raw)
```

codex-rs/cli/src/lib.rs
```diff
@@ -34,6 +39,11 @@ pub struct SeatbeltCommand {
     pub command: Vec<String>,
 }
 
+fn parse_allow_unix_socket_path(raw: &str) -> Result<AbsolutePathBuf, String> {
+    AbsolutePathBuf::relative_to_current_dir(raw)
+        .map_err(|err| format!("invalid path {raw:?}: {err}"))
```
bolinfest: Since `raw` is `&str`, you don't need `:?` to print the `Debug` version of the value.

```suggestion
        .map_err(|err| format!("invalid path {raw}: {err}"))
```

codex-rs/sandboxing/src/manager.rs
```diff
@@ -12,7 +12,9 @@ use crate::policy_transforms::should_require_platform_sandbox;
 #[cfg(target_os = "macos")]
```
bolinfest: Instead of these top-level imports with `#[cfg(target_os = "macos")]`, you might want to just do the imports inside the `SandboxType::MacosSeatbelt` case and then you can drop the `cfg` on each import.

codex-rs/sandboxing/src/seatbelt.rs
```diff
@@ -244,8 +257,17 @@ fn dynamic_network_policy_for_network(
...
 ) -> String {
-    let should_use_restricted_network_policy =
-        !proxy.ports.is_empty() || proxy.has_proxy_config || enforce_managed_network;
+    let has_unix_socket_access = matches!(
+        proxy.unix_domain_socket_policy,
+        UnixDomainSocketPolicy::AllowAll
+    ) || matches!(
+        &proxy.unix_domain_socket_policy,
+        UnixDomainSocketPolicy::Restricted { allowed } if !allowed.is_empty()
+    );
```
bolinfest: This is better because if a new variant is added to the `UnixDomainSocketPolicy` enum, this forces the `match` to be updated:

```suggestion
    let has_unix_socket_access = match proxy.unix_domain_socket_policy {
        UnixDomainSocketPolicy::AllowAll => true,
        UnixDomainSocketPolicy::Restricted { allowed } => !allowed.is_empty()
    };    
```

bolinfest: Maybe `has_some_unix_socket_access` is more accurate since it is `true` when `Restricted` is non-empty, but that is not quite a strong a statement as `AllowAll`?

codex-rs/sandboxing/src/seatbelt.rs
```diff
@@ -285,9 +304,13 @@ fn dynamic_network_policy_for_network(
 
     if network_policy.is_enabled() {
         // No proxy env is configured: retain the existing full-network behavior.
-        format!(
-            "(allow network-outbound)\n(allow network-inbound)\n{MACOS_SEATBELT_NETWORK_POLICY}"
-        )
+        let mut policy = String::from("(allow network-outbound)\n(allow network-inbound)\n");
+        let unix_socket_policy = unix_socket_policy(proxy);
```
bolinfest, line 308: So was it an oversight that the result of `unix_socket_policy()` was not included previously?

codex-rs/sandboxing/src/seatbelt.rs
```diff
@@ -357,31 +380,48 @@ fn build_seatbelt_access_policy(
...
-pub fn create_seatbelt_command_args_for_policies(
-    command: Vec<String>,
-    file_system_sandbox_policy: &FileSystemSandboxPolicy,
-    network_sandbox_policy: NetworkSandboxPolicy,
-    sandbox_policy_cwd: &Path,
-    enforce_managed_network: bool,
-    network: Option<&NetworkProxy>,
-) -> Vec<String> {
+#[derive(Debug)]
+pub struct SeatbeltCommandArgs<'a> {
```
bolinfest: nit: I would name this `SeatbeltCommandParams` since `create_seatbelt_command_args()` takes this struct as an argument, it doesn't create it.

bolinfest: Or even `CreateSeatbeltCommandArgsParams` because it's the "params" for the `create_seatbelt_command_args()` function?

codex-rs/sandboxing/src/seatbelt_tests.rs
```diff
@@ -48,6 +58,23 @@ fn seatbelt_policy_arg(args: &[String]) -> &str {
         .expect("seatbelt args should include policy text")
 }
 
+struct TestConfigReloader;
+
+#[async_trait::async_trait]
```
bolinfest, line 63: We got burned by `async_trait` recently: https://github.com/openai/codex/pull/16630. Prefer native `async` traits, which effectively makes `#[async_trait::async_trait]` unnecessary.

codex-rs/sandboxing/src/seatbelt.rs
aaronl-openai: yeah, I think so. I think it mostly wasn’t an issue before because the unix socket policy was exercised through the proxy/restricted-network path.

codex-rs/sandboxing/src/seatbelt_tests.rs
bolinfest: Oh, I see, we already have:

```rust
#[async_trait]
pub trait ConfigReloader: Send + Sync {
    /// Human-readable description of where config is loaded from, for logs.
    fn source_label(&self) -> String;

    /// Return a freshly loaded state if a reload is needed; otherwise, return `None`.
    async fn maybe_reload(&self) -> Result<Option<ConfigState>>;

    /// Force a reload, regardless of whether a change was detected.
    async fn reload_now(&self) -> Result<ConfigState>;
}
```

OK, we can let this be for now, but I'll go after it in a follow-up...

bolinfest (approved):

PR #17665, opened 2026-04-13
Always enable original image detail on supported models

fjord-oai:
## Summary

This PR removes `image_detail_original` as a runtime experiment and makes original image detail available whenever the selected model supports it.

Concretely, this change:
- drops the `image_detail_original` feature flag from the feature registry and generated config schema
- makes tool-emitted image detail depend only on `ModelInfo.supports_image_detail_original`
- updates `view_image` and `code_mode`/`js_repl` image emission to use that capability check directly
- removes now-redundant experiment-specific tests and instruction coverage
- keeps backward compatibility for existing configs by silently ignoring a stale `features.image_detail_original` entry

The net effect is that `detail: "original"` is always available on supported models, without requiring an experiment toggle.


fjord-oai:
@codex review

pakrym-oai (approved):

PR #17667, opened 2026-04-13
Add `supports_parallel_tool_calls` flag to included mcps

josiah-openai:
## Why

For more advanced MCP usage, we want the model to be able to emit parallel MCP tool calls and have Codex execute eligible ones concurrently, instead of forcing all MCP calls through the serial block.

The main design choice was where to thread the config. I made this server-level because parallel safety depends on the MCP server implementation. Codex reads the flag from `mcp_servers`, threads the opted-in server names into `ToolRouter`, and checks the parsed `ToolPayload::Mcp { server, .. }` at execution time. That avoids relying on model-visible tool names, which can be incomplete in deferred/search-tool paths or ambiguous for similarly named servers/tools.

## What was added

Added `supports_parallel_tool_calls` for MCP servers.

Before:

```toml
[mcp_servers.docs]
command = "docs-server"
```

After:

```toml
[mcp_servers.docs]
command = "docs-server"
supports_parallel_tool_calls = true
```

MCP calls remain serial by default. Only tools from opted-in servers are eligible to run in parallel. Docs also now warn to enable this only when the server’s tools are safe to run concurrently, especially around shared state or read/write races.

## Testing

Tested with a local stdio MCP server exposing real delay tools. The model/Responses side was mocked only to deterministically emit two MCP calls in the same turn.

Each test called `query_with_delay` and `query_with_delay_2` with `{ "seconds": 25 }`.

| Build/config | Observed | Wall time |
| --- | --- | --- |
| main with flag enabled | serial | `58.79s` |
| PR with flag enabled | parallel | `31.73s` |
| PR without flag | serial | `56.70s` |

PR with flag enabled showed both tools start before either completed; main and PR-without-flag completed the first delay before starting the second.

Also added an integration test.

Additional checks:

- `cargo test -p codex-tools` passed
- `cargo test -p codex-core mcp_parallel_support_uses_exact_payload_server` passed
- `git diff --check` passed


josiah-openai:

codex-rs/core/src/tools/js_repl/mod.rs
```diff
@@ -1561,12 +1562,25 @@ impl JsReplManager {
             .await
             .list_all_tools()
             .await;
+        let parallel_mcp_server_names = exec
```
jif-oai: this looks unused here...

codex-rs/rmcp-client/src/bin/test_stdio_server.rs
```diff
@@ -469,6 +559,90 @@ impl TestToolServer {
...
+                entry.insert(SyncBarrierState {
+                    barrier: barrier.clone(),
+                    participants: args.participants,
+                });
+                barrier
+            }
+        }
+    };
+
+    let wait_result = tokio::time::timeout(Duration::from_millis(args.timeout_ms), barrier.wait())
```
jif-oai: Can we clean up SYNC_BARRIERS on timeout/error too?

codex-rs/core/tests/suite/rmcp_client.rs
```diff
@@ -230,6 +231,121 @@ async fn stdio_server_round_trip() -> anyhow::Result<()> {
     Ok(())
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn stdio_mcp_parallel_tool_calls_opt_in_runs_concurrently() -> anyhow::Result<()> {
```
jif-oai, line 383: Please add the default/false e2e case too. Right now we only prove the opt-in path

codex-rs/config/src/mcp_types_tests.rs
```diff
@@ -245,6 +245,19 @@ fn deserialize_server_config_with_tool_filters() {
     assert_eq!(cfg.disabled_tools, Some(vec!["blocked".to_string()]));
 }
 
+#[test]
+fn deserialize_server_config_with_parallel_tool_calls() {
```
jif-oai, line 249: This only covers deserialization. Since the PR also changes both config writers, can we add a round-trip/serialization test for supports_parallel_tool_calls = true?

jif-oai (approved):
Ok once my comments are processed

codex-rs/core/src/tools/js_repl/mod.rs
josiah-openai: Done - this one is slightly weird because it looks like js_repl uses its own tool call router/dispatcher, but I didn't want to mess around with the parameters too much to allow you to leave out the server names.

codex-rs/core/tests/suite/rmcp_client.rs
josiah-openai: added

codex-rs/config/src/mcp_types_tests.rs
josiah-openai: added

PR #17671, opened 2026-04-13
Stabilize exec-server filesystem tests in CI

starr-openai:
## Summary\n- add an exec-server package-local test helper binary that can run exec-server and fs-helper flows\n- route exec-server filesystem tests through that helper instead of cross-crate codex helper binaries\n- stop relying on Bazel-only extra binary wiring for these tests\n\n## Testing\n- not run (per repo guidance for codex changes)

### DIFF
--- a/codex-rs/Cargo.lock
+++ b/codex-rs/Cargo.lock
@@ -1902,7 +1902,6 @@ dependencies = [
  "codex-api",
  "codex-app-server-protocol",
  "codex-apply-patch",
- "codex-arg0",
  "codex-async-utils",
  "codex-code-mode",
  "codex-config",
@@ -1932,6 +1931,7 @@ dependencies = [
  "codex-shell-escalation",
  "codex-state",
  "codex-terminal-detection",
+ "codex-test-binary-support",
  "codex-tools",
  "codex-utils-absolute-path",
  "codex-utils-cache",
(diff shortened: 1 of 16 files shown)


codex-rs/exec-server/src/bin/test_helper.rs
```diff
@@ -0,0 +1,73 @@
...
+use codex_exec_server::CODEX_FS_HELPER_ARG1;
+use codex_exec_server::ExecServerRuntimePaths;
+
+#[cfg(target_os = "linux")]
+use codex_sandboxing::landlock::CODEX_LINUX_SANDBOX_ARG0;
+#[cfg(target_os = "linux")]
+use std::path::Path;
+
+fn main() -> Result<(), Box<dyn Error + Send + Sync>> {
+    #[cfg(target_os = "linux")]
```
pakrym-oai: would be nice to break circular dep and use arg0 helper directly. not blocking.

pakrym-oai (approved):

codex-rs/exec-server/src/bin/test_helper.rs
starr-openai: going to try to refactor and share the same test setup core is using for the same purposes

PR #17673, opened 2026-04-13
[codex] Add marketplace upgrade command

xli-oai:
## Summary
Add a `codex marketplace upgrade` CLI command that reuses the shared configured-marketplace upgrade pipeline.

## What Changed
- added `marketplace upgrade [name]` CLI parsing and user-facing output
- reused `PluginsManager::upgrade_configured_marketplaces_for_config(...)` instead of duplicating upgrade logic
- exposed the minimal core upgrade result surface the CLI needs for reporting and single-marketplace selection
- added CLI integration coverage for upgrading all configured Git marketplaces, upgrading one marketplace, and rejecting unknown names

## Why
The auto-upgrade branch already contains the upgrade implementation, but there was no explicit CLI command to invoke the same flow on demand. This PR adds that command while keeping the actual upgrade logic shared.

## Validation
- `cargo test -p codex-core marketplace_upgrade`
- `cargo test -p codex-cli marketplace_upgrade`
- `just fmt`
- `just fix -p codex-core`
- `just fix -p codex-cli`

### DIFF
--- a/codex-rs/app-server/src/message_processor.rs
+++ b/codex-rs/app-server/src/message_processor.rs
@@ -725,6 +725,7 @@ impl MessageProcessor {
         session: Arc<ConnectionSessionState>,
         request_context: RequestContext,
     ) {
+        let connection_id = connection_request_id.connection_id;
         if !session.initialized() {
             let error = JSONRPCErrorError {
                 code: INVALID_REQUEST_ERROR_CODE,
--- a/codex-rs/cli/src/marketplace_cmd.rs
+++ b/codex-rs/cli/src/marketplace_cmd.rs
@@ -85,12 +85,12 @@ async fn run_add(args: AddMarketplaceArgs) -> Result<()> {
 
     if outcome.already_added {
         println!(
-            "Marketplace '{}' is already added from {}.",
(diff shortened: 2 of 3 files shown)


codex-rs/core/src/plugins/manager.rs
```diff
@@ -1131,23 +1142,44 @@ impl PluginsManager {
...
     }
 
-    pub fn maybe_start_non_curated_plugin_cache_refresh_for_roots(
-        self: &Arc<Self>,
-        roots: &[AbsolutePathBuf],
-    ) {
-        self.maybe_start_non_curated_plugin_cache_refresh_for_roots_with_mode(
-            roots,
-            NonCuratedCacheRefreshMode::IfVersionChanged,
+    pub fn upgrade_configured_marketplaces_for_config(
```
xl-openai, line 1186: It’s getting quite difficult to follow the upgrade logic here. It seems like we now have multiple paths for plugin updates (e.g., upgrade-triggered vs background refresh), which makes the flow harder to reason about.
It might be worth simplifying and unifying these paths—reducing thin wrappers and making the refresh logic more general and consistent.

PR #17674, opened 2026-04-13
Route apply_patch through the environment filesystem

starr-openai:
## Summary
- route apply_patch runtime execution through the selected Environment filesystem instead of the local self-exec path
- keep the standalone apply_patch command surface intact while restoring its launcher/test/docs contract
- add focused apply_patch filesystem sandbox regression coverage

## Validation
- remote devbox Bazel run in progress
- passed: //codex-rs/apply-patch:apply-patch-unit-tests --test_filter=test_read_file_utf8_with_context_reports_invalid_utf8
- in progress / follow-up: focused core and exec Bazel test slices on dev

## Follow-up under review
- remote pre-verification and approval/retry behavior still need explicit scrutiny for delete/update flows
- runtime sandbox-denial classification may need a tighter assertion path than rendered stderr matching


codex-rs/apply-patch/tests/fs_sandbox.rs
```diff
@@ -0,0 +1,353 @@
...
+                path: path_abs.to_path_buf(),
+                sandbox: Some(sandbox),
+            },
+        ]
+    );
+}
+
+#[tokio::test]
+async fn apply_patch_passes_sandbox_to_delete_file_operations() {
+    let dir = tempdir().unwrap_or_else(|err| panic!("tempdir failed: {err}"));
```
pakrym-oai: what's the point of this coverage, we should have existing coverage that uses two real impls we care about.

codex-rs/arg0/src/lib.rs
```diff
@@ -518,4 +526,58 @@ mod tests {
...
+    #[cfg(unix)]
+    #[test]
+    fn prepend_path_entry_creates_apply_patch_aliases() -> std::io::Result<()> {
+        let _guard = ENV_LOCK.lock().expect("env lock should not be poisoned");
+        let codex_home = tempfile::tempdir()?;
+        let previous_codex_home = std::env::var_os("CODEX_HOME");
+        let previous_path = std::env::var_os("PATH");
+
+        unsafe {
+            std::env::set_var("CODEX_HOME", codex_home.path());
```
pakrym-oai: we can't have a test that changes env like this.  we should have this scenario covered via integration tests instead.

starr-openai: good point, updating with an actual e2e test

starr-openai: Removed the in-process env-mutation unit test. This path is already covered by the existing integration command surface in `core/tests/suite/mod.rs` plus `shell_snapshot_still_intercepts_apply_patch`, which exercises `apply_patch <<'EOF' ... EOF` through the arg0-dispatched alias setup.

codex-rs/core/tests/suite/mod.rs
```diff
@@ -1,25 +1,75 @@
...
+    _arg0: Arg0PathEntryGuard,
+    _previous_codex_home: Option<OsString>,
+}
+
+const CODEX_HOME_ENV_VAR: &str = "CODEX_HOME";
 
 // This code runs before any other tests are run.
-// It allows the test binary to behave like codex and dispatch to apply_patch and codex-linux-sandbox
-// based on the arg0.
+// It allows the test binary to behave like codex and dispatch to apply_patch
```
pakrym-oai: revert the entire file?

codex-rs/apply-patch/src/invocation.rs
```diff
@@ -171,15 +170,17 @@ pub async fn maybe_parse_apply_patch_verified(
                         );
                     }
                     Hunk::DeleteFile { .. } => {
-                        let content = match fs.read_file_text(&path, sandbox).await {
+                        let content = match crate::read_file_utf8_with_context(
```
pakrym-oai: was this an existing requirement?

pakrym-oai: the need to report utf8 parsing error with context ?

codex-rs/apply-patch/src/lib.rs
```diff
@@ -379,6 +379,40 @@ async fn apply_hunks_to_files(
...
+    fs: &dyn ExecutorFileSystem,
+    path_abs: &AbsolutePathBuf,
+    contents: Vec<u8>,
+    sandbox: Option<&FileSystemSandboxContext>,
+) -> anyhow::Result<()> {
+    match fs.write_file(path_abs, contents.clone(), sandbox).await {
+        Ok(()) => Ok(()),
+        Err(err) if err.kind() == io::ErrorKind::NotFound => {
+            if let Some(parent_abs) = path_abs.parent() {
+                fs.create_directory(
```
pakrym-oai, line 372: can we not just call create_directory optimistically?

pakrym-oai: like we did before?

codex-rs/core/src/tools/runtimes/apply_patch.rs
```diff
@@ -215,51 +196,43 @@ impl ToolRuntime<ApplyPatchRequest, ExecToolCallOutput> for ApplyPatchRuntime {
...
-        }
-
-        #[cfg(target_os = "windows")]
-        let command = Self::build_sandbox_command(req, &ctx.turn.config.codex_home)?;
-        #[cfg(not(target_os = "windows"))]
-        let command = Self::build_sandbox_command(req, ctx.turn.codex_self_exe.as_ref())?;
-        let options = ExecOptions {
-            expiration: req.timeout_ms.into(),
-            capture_policy: ExecCapturePolicy::ShellTool,
+        let environment = ctx.turn.environment.as_ref().ok_or_else(|| {
```
pakrym-oai, line 199: ideally we'd stop pretending we are doing an exec tool call but we can do it later.

codex-rs/core/src/tools/runtimes/apply_patch_tests.rs
```diff
@@ -69,70 +75,94 @@ fn guardian_review_request_includes_patch_context() {
     );
 }
 
-#[cfg(not(target_os = "windows"))]
 #[test]
-fn build_sandbox_command_prefers_configured_codex_self_exe_for_apply_patch() {
+fn file_system_sandbox_context_uses_active_attempt() {
```
pakrym-oai, line 79: Feels like these tests are very marginally useful. Do we have enough integration test to validate the actual local behavior of apply patch + sandbox?

codex-rs/apply-patch/src/lib.rs
```diff
@@ -379,6 +379,40 @@ async fn apply_hunks_to_files(
     })
 }
 
+async fn write_file_with_missing_parent_retry(
```
pakrym-oai, line 362: can we avoid these overly specific fs helpers?

codex-rs/core/tests/suite/mod.rs
starr-openai: Yes. I reverted the stray formatting-only diff in `6c7edb90f`, so `core/tests/suite/mod.rs` drops out of the PR again.

codex-rs/apply-patch/src/invocation.rs
starr-openai: No, that was not a new requirement. In `6c7edb90f` I removed the extra UTF-8 helper/tests and went back to the existing `read_file_text(...).map_err(...)` path so `InvalidData` still flows through with the same contextual error we already used for patch verification.

starr-openai: Right, there was not a separate requirement for a new UTF-8 wrapper here. `6c7edb90f` drops that helper and keeps the normal `read_file_text` behavior, which already reports `InvalidData`; I only keep the existing read-context string on top.

codex-rs/apply-patch/src/lib.rs
starr-openai: Yep, restored in `6c7edb90f`.

codex-rs/core/src/tools/runtimes/apply_patch_tests.rs
starr-openai: The e2e coverage here is in the existing integration suite rather than in these runtime unit tests. In particular `core/tests/suite/apply_patch_cli.rs` already exercises the real local tool path in `apply_patch_cli_multiple_operations_integration`, `apply_patch_cli_moves_file_to_new_directory`, `apply_patch_cli_rejects_path_traversal_outside_workspace`, and `apply_patch_cli_rejects_move_path_traversal_outside_workspace`, and `core/tests/suite/shell_snapshot.rs` still covers the heredoc interception path in `shell_command_snapshot_still_intercepts_apply_patch`. I kept the runtime tests only for the new `SandboxAttempt -> FileSystemSandboxContext` mapping, since that translation is internal and not directly asserted by the harness.

codex-rs/apply-patch/src/lib.rs
starr-openai: Agreed. `6c7edb90f` goes back to the optimistic `create_directory(..., recursive: true)` before the write, which is what we did before and keeps the flow simpler across both local and remote filesystem implementations.

codex-rs/core/src/tools/runtimes/apply_patch.rs
starr-openai: Agreed. This PR keeps the existing `ExecToolCallOutput` / output-delta contract so the surrounding tool plumbing does not change, but the actual patch application now goes straight through the environment filesystem. I would treat “stop modeling this as exec” as a follow-up cleanup once we want to split that protocol surface.

codex-rs/apply-patch/src/lib.rs
starr-openai: Agreed. `6c7edb90f` removes the write-retry helper and goes back to the direct mkdir-plus-write path.

pakrym-oai (approved):

PR #17678, opened 2026-04-13
fix: use stale cloud requirements cache fallback

viyatb-oai:
## Summary
- allow eligible ChatGPT Business/Enterprise accounts to use a same-identity signed stale cloud requirements cache when remote loading fails or times out
- improve the startup error with retry count, last backend status when available, and clearer managed-config/cache guidance
- add focused cloud-requirements regression coverage for stale-cache fallback and clearer failure messages

## Root Cause
Cloud requirements cache reads rejected expired entries before remote fetch. That meant a short backend/config-store outage could fail closed at startup even when the local signed cache was only slightly stale.

## Validation
- cargo test -p codex-cloud-requirements --quiet
- cargo clippy -p codex-cloud-requirements --tests --quiet


xl-openai:
I think there is a risk here that we may silently keep using stale cache for an extended period of time.

If the backend continues to fail, this effectively bypasses the expiration and we might never refresh to a newer version, which makes the expiration less meaningful in practice.

PR #17684, opened 2026-04-13
Adjust default tool search result caps

malone-oai:
## Summary

- Allows selected MCP results to return a larger default result set.
- Keeps the existing default cap for other MCP results.
- Applies the cap consistently when higher explicit limits are requested.

## Testing

- `cargo test -p codex-core tool_search`
- Ran a local CLI smoke test with two stdio MCP servers exposing 100 tools each; the selected-server query returned 20 tools and the regular-server query returned 8.

### DIFF
--- a/codex-rs/core/src/tools/handlers/tool_search.rs
+++ b/codex-rs/core/src/tools/handlers/tool_search.rs
@@ -14,6 +14,9 @@ use codex_tools::TOOL_SEARCH_TOOL_NAME;
 use codex_tools::ToolSearchResultSource;
 use codex_tools::collect_tool_search_output_tools;
 
+const COMPUTER_USE_MCP_SERVER_NAME: &str = "computer-use";
+const COMPUTER_USE_TOOL_SEARCH_LIMIT: usize = 20;
+
 pub struct ToolSearchHandler {
     entries: Vec<(String, ToolInfo)>,
     search_engine: SearchEngine<usize>,
@@ -67,7 +70,8 @@ impl ToolHandler for ToolSearchHandler {
                 "query must not be empty".to_string(),
             ));
         }
-        let limit = args.limit.unwrap_or(TOOL_SEARCH_DEFAULT_LIMIT);
+        let requested_limit = args.limit;
+        let limit = requested_limit.unwrap_or(TOOL_SEARCH_DEFAULT_LIMIT);
 
         if limit == 0 {
             return Err(FunctionCallError::RespondToModel(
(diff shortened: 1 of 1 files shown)


malone-oai:

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -79,21 +83,18 @@ impl ToolHandler for ToolSearchHandler {
             return Ok(ToolSearchOutput { tools: Vec::new() });
         }
 
-        let results = self.search_engine.search(query, limit);
+        let results = self.search_result_entries(query, requested_limit);
```
shijie-oai: wouldn't you want to use limit here?

malone-oai: yep, agreed. i updated this so the validated `limit` is passed into `search_result_entries` directly, and kept a separate boolean for whether the model omitted `limit`.

shijie-oai (approved):

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -104,6 +105,63 @@ impl ToolHandler for ToolSearchHandler {
...
+                .any(|(_, tool)| tool.server_name == COMPUTER_USE_MCP_SERVER_NAME)
+        {
+            results = self
+                .search_engine
+                .search(query, COMPUTER_USE_TOOL_SEARCH_LIMIT)
+                .into_iter()
+                .filter_map(|result| self.entries.get(result.document.id))
+                .collect();
+        }
+        limit_results_per_server(results)
```
shijie-oai, line 136: Just want to make sure this is expected - limit_results_per_server does not respect the limit provided by the user but always enforce the default value. It seems like the user provided limit is only used for query above but not the find result.

malone-oai: ah good catch, this was not the intention. 

i updated this so the per-server cap only applies on the default/no-limit path. explicit `limit` now keeps the existing behavior: it controls the raw search result count and is not post-capped by server. the intended change is just that the default cap can be higher for one server, not that user-provided limits become per-server capped.

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -104,6 +105,66 @@ impl ToolHandler for ToolSearchHandler {
...
+            return results;
+        }
+
+        if results
+            .iter()
+            .any(|(_, tool)| tool.server_name == COMPUTER_USE_MCP_SERVER_NAME)
+        {
+            results = self
+                .search_engine
+                .search(query, COMPUTER_USE_TOOL_SEARCH_LIMIT)
```
shijie-oai, line 131: I see so if there is computer use, we would like to refetch with a higher limit for tools.

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -104,6 +105,66 @@ impl ToolHandler for ToolSearchHandler {
     }
 }
 
+impl ToolSearchHandler {
+    fn search_result_entries(
+        &self,
+        query: &str,
+        limit: usize,
+        use_default_limit: bool,
```
shijie-oai, line 113: very small nit and we can address as followup to unblock - I don't see the point of unpacking and unwrap with fallback at line 73 to be fair.

shijie-oai (approved):

PR #17691, opened 2026-04-13
Fix tui compilation

davidhao3300:
The recent release broke, codex suggested this as the fix

Source failure: https://github.com/openai/codex/actions/runs/24362949066/job/71147202092

Probably from https://github.com/openai/codex/commit/ac82443d073f7f9a2248bad51bae2fa424ef4946

For why it got in:
```
The relevant setup:

.github/workflows/rust-ci.yml (line 1) runs on PRs, but for codex-rs it only does:

cargo fmt --check
cargo shear
argument-comment lint via Bazel
no cargo check, no cargo clippy over the workspace, no cargo test over codex-tui
.github/workflows/rust-ci-full.yml (line 1) runs on pushes to main and branches matching **full-ci**. That one does compile TUI because:

codex-rs/Cargo.toml includes "tui" as a workspace member
lint_build runs cargo clippy --target ... --tests --profile ...
the matrix includes both dev and release profiles
tests runs cargo nextest run ..., but only dev-profile tests
Release CI also compiles it indirectly. .github/workflows/rust-release.yml (line 235) builds --bin codex, and cli/Cargo.toml (line 46) depends on codex-tui.
```

Codex tested locally with `cargo check -p codex-tui --release` and was able to repro, and verified that this fixed it

### DIFF
--- a/codex-rs/tui/src/updates.rs
+++ b/codex-rs/tui/src/updates.rs
@@ -70,7 +70,7 @@ struct HomebrewCaskInfo {
 }
 
 fn version_filepath(config: &Config) -> PathBuf {
-    config.codex_home.join(VERSION_FILENAME)
+    config.codex_home.join(VERSION_FILENAME).into_path_buf()
 }
 
 fn read_version_info(version_file: &Path) -> anyhow::Result<VersionInfo> {


euroelessar (approved):

jif-oai (approved):

davidhao3300:
just to follow up, it seems like main CI SHOULD have failed with this error, but it currently doesn't for some reason. PR CI indeed does not set the flags in such a way that would catch this kind of error

bolinfest:
I believe that this code is only reachable behind:

```
#[cfg(not(debug_assertions))]
```

and the Bazel builds in CI only build in debug mode?

PR #17702, opened 2026-04-13
Fix remote skill popup loading

starr-openai:
## Summary

Fix the TUI `$` skill popup so personal skills appear reliably when Codex is connected to a remote app-server.

## What changed

- load skills on TUI startup with an explicit forced refresh
- refresh skills using the actual current cwd instead of an empty `cwds` list
- resync an already-open `$` popup when skill mentions are updated
- add a regression test for refreshing an open mention popup

## Root cause

The TUI was sometimes sending `list_skills` with `cwds: []` after `SessionConfigured`.

For the launchd app-server flow, the server resolved that empty cwd list to its own process cwd, which was `/`. The response therefore came back tagged with `cwd: "/"`, and the TUI later filtered skills by exact cwd match against the actual project cwd such as `/Users/starr/code/dream`. That dropped all personal skills from the mention list, so `$` only showed plugins/apps.

## Verification

Built successfully with remote cache disabled:

```bash
cd /Users/starr/code/codex-worktrees/starr-skill-popup-20260413130509
bazel --output_base=/tmp/codex-bazel-verify-starr-skill-popup build //codex-rs/cli:codex --noremote_accept_cached --noremote_upload_local_results --disk_cache=
```

Also verified interactively in a PTY against the live app-server at `ws://127.0.0.1:4511`:
- launched the built TUI
- typed `$`
- confirmed personal skills appeared in the popup, including entries such as `Applied Devbox`, `CI Debug`, `Channel Summarization`, `Codex PR Review`, and `Daily Digest`

## Files changed

- `codex-rs/tui/src/app.rs`
- `codex-rs/tui/src/bottom_pane/chat_composer.rs`

### DIFF
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -3847,6 +3847,17 @@ impl App {
             app.enqueue_primary_thread_session(started.session, started.turns)
                 .await?;
         }
+        match app_server
+            .skills_list(codex_app_server_protocol::SkillsListParams {
+                cwds: vec![app.config.cwd.to_path_buf()],
+                force_reload: true,
+                per_cwd_extra_user_roots: None,
+            })
+            .await
+        {
+            Ok(response) => app.handle_skills_list_response(response),
+            Err(err) => tracing::warn!("failed to load skills on startup: {err:#}"),
+        }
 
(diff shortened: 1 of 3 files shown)


etraut-openai (approved):
Nice! Thanks for fixing this.

PR #17703, opened 2026-04-13
Add turn item injection API

pakrym-oai:
## Summary
- Add `turn/inject_items` app-server v2 request support for appending raw Responses API items to a loaded thread history without starting a turn.
- Generate JSON schema and TypeScript protocol artifacts for the new params and empty response.
- Document the new endpoint and include a request/response example.
- Preserve compatibility with the typo alias `turn/injet_items` while returning the canonical method name.

## Testing
- Not run (not requested)


andmis (approved):
TY!

PR #17705, opened 2026-04-13
Add Bazel verify-release-build job

bolinfest:
## Why

`main` recently needed [#17691](https://github.com/openai/codex/pull/17691) because code behind `cfg(not(debug_assertions))` was not being compiled by the Bazel PR workflow. Our existing CI only built the fast/debug configuration, so PRs could stay green while release-only Rust code still failed to compile. This PR adds a release-style compile check that is cheap enough to run on every PR.

## What Changed

- Added a `verify-release-build` job to `.github/workflows/bazel.yml`.
- Represented each supported OS once in that job's matrix: x64 Linux, arm64 macOS, and x64 Windows.
- Kept the build close to fastbuild cost by using `--compilation_mode=fastbuild` while forcing Rust to compile with `-Cdebug-assertions=no`, which makes `cfg(not(debug_assertions))` true without also turning on release optimizations or debug-info generation.
- Added comments in `.github/workflows/bazel.yml` and `scripts/list-bazel-release-targets.sh` to make the job's intent and target scope explicit.
- Restored the Bazel repository cache save behavior to run after every non-cancelled job, matching [#16926](https://github.com/openai/codex/pull/16926), and removed the now-unused `repository-cache-hit` output from `prepare-bazel-ci`.
- Reused the shared `prepare-bazel-ci` action from the parent PR so the new job does not duplicate Bazel setup boilerplate.

## Verification

- Used `bazel aquery` on `//codex-rs/tui:codex-tui` to confirm the Rust compile still uses `opt-level=0` and `debuginfo=0` while passing `-Cdebug-assertions=no`.
- Parsed `.github/workflows/bazel.yml` as YAML locally.
- Ran `bash -n scripts/list-bazel-release-targets.sh`.

### DIFF
--- a/.github/actions/prepare-bazel-ci/action.yml
+++ b/.github/actions/prepare-bazel-ci/action.yml
@@ -9,9 +9,6 @@ inputs:
     required: false
     default: "false"
 outputs:
-  repository-cache-hit:
-    description: Whether the Bazel repository cache restore hit.
-    value: ${{ steps.cache_bazel_repository_restore.outputs.cache-hit }}
   repository-cache-path:
     description: Filesystem path used for the Bazel repository cache.
     value: ${{ steps.setup_bazel.outputs.repository-cache-path }}
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -109,10 +109,10 @@ jobs:
           path: ${{ runner.temp }}/bazel-execution-logs
           if-no-files-found: ignore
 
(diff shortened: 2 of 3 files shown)


pakrym-oai (approved):

euroelessar:
we may need to separate bazel cache snapshots for release build vs regular build as they produce different artifacts (and there is only one snapshot per key, so first one wins or smth)

PR #17709, opened 2026-04-13
[codex] Initialize ICU data for code mode V8

pakrym-oai:
Link ICU data into code mode, otherwise locale-dependent methods cause a panic and a crash.


codex-rs/code-mode/src/service.rs
```diff
@@ -561,6 +561,69 @@ mod tests {
...
+    async fn date_locale_time_string_formats_with_icu_data() {
+        let service = CodeModeService::new();
+
+        let response = service
+            .execute(ExecuteRequest {
+                source: r#"
+const value = new Date("2025-01-02T03:04:05Z")
+  .toLocaleTimeString("en-US", { hour12: false, timeZone: "UTC" });
+text(value);
+"#
```
bolinfest, line 583: Consider:

```javascript
new Date("2025-01-02T03:04:05Z").toLocaleString("fr-FR", {
  weekday: "long",   // Thursday
  month: "long",     // January
  day: "numeric",    // 2
  hour: "2-digit",
  minute: "2-digit",
  second: "2-digit",
  hour12: false,
  timeZone: "UTC",
})
```

which should return `jeudi 2 janvier à 03:04:05`, I believe, which is maybe a stronger verification that ICU data is loaded correctly?

codex-rs/code-mode/src/service.rs
```diff
@@ -561,6 +561,69 @@ mod tests {
...
+    async fn intl_date_time_format_constructs_with_icu_data() {
+        let service = CodeModeService::new();
+
+        let response = service
+            .execute(ExecuteRequest {
+                source: r#"
+new Intl.DateTimeFormat(void 0, {
+  timeZone: "America/Los_Angeles",
+});
+text("ok");
```
bolinfest: Is this just proving that `Intl.DateTimeFormat` does not throw an exception? Why not use it with something to be sure it works for real?

bolinfest (approved):

PR #17711, opened 2026-04-13
[codex] Add marketplace/add app-server RPC

xli-oai:
## Summary

Add a new app-server `marketplace/add` RPC so clients can add remote plugin marketplaces through the backend instead of shelling out.

This change:
- adds `MarketplaceAddParams` / `MarketplaceAddResponse` to the app-server protocol
- implements marketplace add orchestration in core plugin code, including source validation, clone/install, and user config persistence
- wires the new request through `codex_message_processor`
- updates generated schema files and adds focused tests for protocol, core, and app-server coverage

## Why

The TUI can already browse, inspect, install, and uninstall plugins, but it did not have a backend API for adding a marketplace source. This fills that backend gap so a future `/plugins` management UI can create marketplaces through the same RPC layer as the rest of plugin management.

## Validation

- `cargo test -p codex-app-server-protocol marketplace_add`
- `cargo test -p codex-core marketplace_add`
- `cargo test -p codex-app-server marketplace_add`
- `just fix -p codex-core`
- `just fmt`


xli-oai:
Superseded by the stacked setup for this work:
- shared refactor: #17717
- app-server RPC on top: #17718

This branch was restacked on top of the shared refactor, so the new review path is split across those two PRs.

PR #17713, opened 2026-04-13
feat: add opt-in provider runtime abstraction

celia-oai:
## Summary

- Add `codex-model-provider` as the runtime home for model-provider behavior that does not belong in `codex-core`, `codex-login`, or `codex-api`.
- The new crate wraps configured `ModelProviderInfo` in a `ModelProvider` trait object that can resolve the API provider config, provider-scoped auth manager, and request auth provider for each call.
- This centralizes provider auth behavior in one place today, and gives us an extension point for future provider-specific auth, model listing, request setup, and related runtime behavior.

## Tests
Ran tests manually to make sure that provider auth under different configs still work as expected.


celia-oai:
@codex review

celia-oai:
> ### 💡 Codex Review
> https://github.com/openai/codex/blob/4d9d09db3a877ac8d8f07e4ed5961e9d7d1b5ed0/codex-rs/core/src/codex.rs#L1744-L1746
> 
> 
> **P2 Badge Use runtime auth source when building session telemetry**
> `Session::new` still builds auth-env telemetry from the legacy provider and base auth manager. When `ProviderRuntime::Resolved` is active, request auth can come from resolved auth state (env/static/command) and a different auth manager, so startup telemetry reports incorrect auth-env context and can mislead diagnostics.
> 
> ℹ️ About Codex in GitHub

this seems to be a small tech debt. Cleaning it up as part of this PR.

celia-oai:
@codex review

codex-rs/core/src/config/config_tests.rs
```diff
@@ -4561,6 +4561,7 @@ fn test_precedence_fixture_with_o3_profile() -> std::io::Result<()> {
             service_tier: None,
             model_provider_id: "openai".to_string(),
             model_provider: fixture.openai_provider.clone(),
+            provider_runtime: ProviderRuntime::Legacy,
```
pakrym-oai: shoul the runtime be on config? I think config should be static data and provider will encode behaviors.

codex-rs/model-provider/src/lib.rs
```diff
@@ -0,0 +1,380 @@
...
+
+/// Provider IDs opted in to the runtime provider framework in production.
+///
+/// This is intentionally code-owned rather than user-configurable. Leaving it
+/// empty preserves legacy behavior for every provider.
+pub const PROVIDER_FRAMEWORK_ENABLED_PROVIDER_IDS: &[&str] = &[];
+
+/// Runtime strategy selected for the active model provider.
+#[derive(Debug, Clone, PartialEq, Default)]
+pub enum ProviderRuntime {
```
pakrym-oai: I don't think we should have this enum. Otherwise we'll forever keep forks all over our code.

Before we add any new features we should extract our existing feature behind the provider.

codex-rs/model-provider/src/lib.rs
```diff
@@ -0,0 +1,380 @@
...
+    /// Command-backed bearer token.
+    CommandBearer { config: ModelProviderAuthInfo },
+    /// Bearer token supplied by the session auth manager.
+    SessionAuth,
+}
+
+/// Policy controlling which providers may use the new runtime framework.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct ProviderResolutionPolicy {
+    enabled_provider_ids: HashSet<String>,
```
pakrym-oai: This is not the shape we need though. We need name -> CustomImpl. Where for each name the provider implementation code is going to be different.

celia-oai:
@codex review

celia-oai:
@codex review

codex-rs/model-provider/src/provider.rs
```diff
@@ -0,0 +1,91 @@
...
+    /// Resolves the auth and API-provider configuration for a request.
+    fn resolve_auth(&self) -> ModelProviderAuthFuture<'_>;
+}
+
+/// Shared runtime model provider handle.
+pub type SharedModelProvider = Arc<dyn ModelProvider>;
+
+/// Future returned while resolving model-provider auth.
+pub type ModelProviderAuthFuture<'a> =
+    Pin<Box<dyn Future<Output = codex_protocol::error::Result<ResolvedProviderAuth>> + Send + 'a>>;
```
pakrym-oai: async_trait?

codex-rs/model-provider/src/provider.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// Auth and provider configuration resolved for a model-provider request.
+pub struct ResolvedProviderAuth {
+    /// The current Codex auth session, when one is available.
+    pub auth: Option<CodexAuth>,
```
pakrym-oai: this value is also available via auth_manager, why both?

codex-rs/codex-api/src/auth.rs
```diff
@@ -7,4 +8,26 @@ use http::HeaderMap;
...
+    fn auth_header_attached(&self) -> bool {
+        false
+    }
+
+    fn auth_header_name(&self) -> Option<&'static str> {
+        None
+    }
+}
+
+impl<T: AuthProvider + ?Sized> AuthProvider for Arc<T> {
```
pakrym-oai: do we need this proxy impl?

codex-rs/codex-api/src/auth.rs
```diff
@@ -7,4 +8,26 @@ use http::HeaderMap;
 /// reach this interface.
 pub trait AuthProvider: Send + Sync {
     fn add_auth_headers(&self, headers: &mut HeaderMap);
+
+    fn auth_header_attached(&self) -> bool {
```
pakrym-oai: can we have a default impl based on add_auth_headers? or just consider dropping these methods. Are they very useful?

codex-rs/model-provider/src/provider.rs
celia-oai: yes, this is refactored from here: https://github.com/openai/codex/pull/17713/changes#diff-c39945d3c42f29b506ff54b7fa2be0795b06d7ad97f1bf33956f60e3c6f19c19L165. This is just so that we have one returned shape for resolve_auth, as these values should be resolved together. otherwise client has to make a separate call to auth_manager.auth() to fetch this value everytime it calls resolve_auth.

codex-rs/codex-api/src/files.rs
```diff
@@ -291,8 +291,21 @@ mod tests {
     use wiremock::matchers::method;
     use wiremock::matchers::path;
 
-    fn chatgpt_auth() -> CoreAuthProvider {
-        CoreAuthProvider::for_test(Some("token"), Some("account_id"))
+    #[derive(Clone, Copy)]
+    struct ChatGptTestAuth;
+
+    impl AuthProvider for ChatGptTestAuth {
```
pakrym-oai, line 297: we may still want some test provider so we don't ahve to reimplement it everywhere.

pakrym-oai: maybe it's just this file though

codex-rs/core/src/client.rs
```diff
@@ -275,7 +264,7 @@ pub(crate) struct RealtimeWebrtcCallStart {
 /// API-key sessions send that API bearer. ChatGPT-auth sessions send their bearer plus account id;
 /// transceiver is responsible for accepting that same call-create identity on the direct
 /// `api.openai.com` sideband path.
-fn sideband_websocket_auth_headers(api_auth: &CoreAuthProvider) -> ApiHeaderMap {
+fn sideband_websocket_auth_headers(api_auth: &dyn AuthProvider) -> ApiHeaderMap {
```
pakrym-oai, line 279: consider using one exchange type dyn AuthProvider or SharedAuthProvider as a primitive you pass around.

codex-rs/core/src/client.rs
```diff
@@ -298,14 +287,15 @@ impl ModelClient {
         include_timing_metrics: bool,
         beta_features_header: Option<String>,
     ) -> Self {
-        let auth_manager = auth_manager_for_provider(auth_manager, &provider);
-        let codex_api_key_env_enabled = auth_manager
+        let provider = create_model_provider(provider, auth_manager);
```
pakrym-oai: should we pass in the existing provider?

codex-rs/model-provider/src/provider.rs
celia-oai: good call!

celia-oai:
@codex review

codex-rs/codex-api/src/files.rs
celia-oai: yeah I think other tests directly do `impl from AuthProvider for DummyAuth`.  this is also a special case of avoiding circular dependencies (model-provider crate depends on codex-api)

codex-rs/core/src/client.rs
celia-oai: this is created from model_provider_info. maybe I can rename to make it less confusing.

codex-rs/models-manager/src/manager.rs
```diff
@@ -213,7 +212,7 @@ impl ModelsManager {
         collaboration_modes_config: CollaborationModesConfig,
         provider: ModelProviderInfo,
```
pakrym-oai: Consider just passing the full model provider in. 

This relashionship feels backwards though. Model provider will have to return/use the models manager, not the reverse.

OK to keep as is for now.

codex-rs/model-provider/src/auth.rs
```diff
@@ -0,0 +1,61 @@
...
+            account_id: None,
+        })
+    }
+}
+
+pub(crate) fn resolve_provider_auth(
+    auth: Option<CodexAuth>,
+    provider: &ModelProviderInfo,
+) -> codex_protocol::error::Result<Arc<dyn AuthProvider>> {
+    let api_auth = bearer_auth_provider_from_auth(auth, provider)?;
```
pakrym-oai: inline?

codex-rs/model-provider/src/provider.rs
pakrym-oai: I don't think we should group all of these values together tbh. `Provider` has nothing to do with auth, it just describes where to send the request.

codex-rs/model-provider/src/provider.rs
```diff
@@ -0,0 +1,86 @@
...
+    info: ModelProviderInfo,
+    auth_manager: Option<Arc<AuthManager>>,
+) -> SharedModelProvider {
+    let auth_manager =
+        auth_manager.map(|auth_manager| auth_manager_for_provider(auth_manager, &info));
+    Arc::new(ConfiguredModelProvider { info, auth_manager })
+}
+
+/// Runtime model provider backed by configured `ModelProviderInfo`.
+#[derive(Clone, Debug)]
```
pakrym-oai, line 64: I don't understand the relashionship here. create_model_provider both receives and creates an auth manager.

pakrym-oai: Feels very unreliable, other parts of the system might not use the correct auth manager.

codex-rs/codex-api/src/auth.rs
celia-oai: nope, removed

codex-rs/core/src/compact.rs
```diff
@@ -57,8 +58,8 @@ pub(crate) enum InitialContextInjection {
     DoNotInject,
 }
 
-pub(crate) fn should_use_remote_compact_task(provider: &ModelProviderInfo) -> bool {
-    provider.supports_remote_compaction()
+pub(crate) fn should_use_remote_compact_task(provider: &dyn ModelProvider) -> bool {
+    provider.info().supports_remote_compaction()
```
pakrym-oai: followup move onto ModelProvider

codex-rs/codex-api/src/auth.rs
celia-oai: put these in a separate AuthHeaderTelemetry so it doesn't interfere with the main trait

codex-rs/core/src/client.rs
celia-oai: sg. will use SharedAuthProvider as the default primitive unless the function doesn't need ownership. will define it in codex-api crate to avoid circular dependencies.

codex-rs/models-manager/src/manager.rs
celia-oai: left TODO

codex-rs/core/src/compact.rs
celia-oai: left to-do

codex-rs/model-provider/src/provider.rs
celia-oai: i added a docstring above that hopefully leaves enough context for follow-up:

>     /// TODO(celia-oai): Make auth manager access internal to this crate so callers
>     /// resolve provider-specific auth only through `ModelProvider`. We first need
>     /// to think through whether Codex should have a unified provider-specific auth
>     /// manager throughout the codebase; that is a larger refactor than this change.

codex-rs/model-provider/src/provider.rs
celia-oai: broke this into separate functions

celia-oai: I think it's still good to have a `fn auth` though, and eventually we should not expose the auth manager either

pakrym-oai (approved):

PR #17714, opened 2026-04-13
Support original-detail metadata on MCP image outputs

fjord-oai:
## Summary
- honor `_meta["codex/imageDetail"] == "original"` on MCP image content and map it to `detail: "original"` where supported
- strip that detail back out when the active model does not support original-detail image inputs
- update code-mode `image(...)` to accept individual MCP image blocks
- teach `js_repl` / `codex.emitImage(...)` to preserve the same hint from raw MCP image outputs
- document the new `_meta` contract and add generic RMCP-backed coverage across protocol, core, code-mode, and js_repl paths


fjord-oai:
@codex review

codex-rs/code-mode/src/runtime/value.rs
```diff
@@ -39,40 +42,76 @@ pub(super) fn normalize_output_image(
         let (image_url, detail) = if value.is_string() {
             (value.to_rust_string_lossy(scope), None)
         } else if value.is_object() && !value.is_array() {
```
pakrym-oai, line 45: I wouldn't push this into code mode and keep the feature only at raw MCP level. In code mode the model has ability to request whatever detail it wants.

codex-rs/protocol/src/models.rs
```diff
@@ -1372,6 +1374,8 @@ fn convert_mcp_content_to_items(
             data: String,
             #[serde(rename = "mimeType", alias = "mime_type")]
             mime_type: Option<String>,
+            #[serde(rename = "_meta", default)]
```
pakrym-oai, line 1377: shouldn't _meta be on CallToolResult level?

codex-rs/code-mode/src/runtime/value.rs
fjord-oai: The intent here isn’t to add a new code_mode-specific way to choose detail. It’s to preserve existing metadata when the caller forwards a raw MCP image item through `image()`.

Without this, the caller has to manually unpack the MCP image block and re-thread detail, which is awkward and loses parity with the existing`view_image` passthrough flow.

codex-rs/protocol/src/models.rs
fjord-oai: Related to the other comment, having it per-image makes it easier to pass individual images to the `image()` code_mode helper.

codex-rs/code-mode/src/runtime/value.rs
```diff
@@ -39,40 +42,76 @@ pub(super) fn normalize_output_image(
...
-                        .to_string()
-                })?;
-            let detail = match object.get(scope, detail_key.into()) {
-                Some(value) if value.is_string() => Some(value.to_rust_string_lossy(scope)),
-                Some(value) if value.is_null() || value.is_undefined() => None,
-                Some(_) => return Err("image detail must be a string when provided".to_string()),
-                None => None,
-            };
-            (image_url, detail)
+            if let Some(image_url) = object.get(scope, image_url_key.into())
```
pakrym-oai: let's split this method into MCP and non-MCP parsers.

codex-rs/core/src/tools/context.rs
```diff
@@ -122,6 +123,7 @@ impl ToolOutput for CallToolResult {
 pub struct McpToolOutput {
     pub result: CallToolResult,
     pub wall_time: Duration,
+    pub can_request_original_image_detail: bool,
```
pakrym-oai: original_image_detail_supported

codex-rs/core/src/tools/context.rs
```diff
@@ -155,6 +157,10 @@ impl ToolOutput for McpToolOutput {
 impl McpToolOutput {
     fn response_payload(&self) -> FunctionCallOutputPayload {
         let mut payload = self.result.as_function_call_output_payload();
+        if let Some(items) = payload.content_items_mut() {
+            sanitize_original_image_detail(self.can_request_original_image_detail, items);
```
pakrym-oai: nit: inline sanitize_original_image_detail into McpToolOutput

codex-rs/core/src/original_image_detail.rs
```diff
@@ -1,2 +1,22 @@
+use codex_protocol::models::FunctionCallOutputContentItem;
+use codex_protocol::models::ImageDetail;
+
 pub(crate) use codex_tools::can_request_original_image_detail;
 pub(crate) use codex_tools::normalize_output_image_detail;
+
+pub(crate) fn sanitize_original_image_detail(
```
pakrym-oai: we already have some helpers in codex-rs/tools/src/image_detail.rs

pakrym-oai (approved):

codex-rs/core/src/original_image_detail.rs
fjord-oai: good point, moved

codex-rs/code-mode/src/runtime/value.rs
fjord-oai: done

codex-rs/core/src/tools/context.rs
fjord-oai: done

PR #17718, opened 2026-04-13
[codex] Add marketplace/add app-server RPC

xli-oai:
## Summary

Add a new app-server `marketplace/add` RPC on top of the shared marketplace-add implementation.

This change:
- adds `MarketplaceAddParams` / `MarketplaceAddResponse` to the app-server protocol
- wires the new request through `codex_message_processor`
- reuses the shared core marketplace-add flow from the lower refactor PR
- updates generated schema files and adds focused protocol and app-server coverage

## Why

The TUI can already browse, inspect, install, and uninstall plugins, but it did not have a backend API for adding a marketplace source. With the shared CLI/core refactor below this PR, the API now reuses the same underlying implementation instead of carrying its own duplicate flow.

## Validation

- `cargo test -p codex-app-server-protocol marketplace_add`
- `cargo test -p codex-app-server marketplace_add`


xl-openai:
Also need to update the README.

PR #17720, opened 2026-04-13
Make skill loading filesystem-aware

pakrym-oai:
Migrates skill loading to support reading repo skills from the remote environment.


pakrym-oai:
@codex review

codex-rs/core/src/plugins/manager.rs
```diff
@@ -1734,15 +1725,7 @@ fn load_plugin(
         .or_else(|| Some(manifest.name.clone()));
     loaded_plugin.manifest_description = manifest.description.clone();
     loaded_plugin.skill_roots = plugin_skill_roots(&plugin_root, manifest_paths);
-    let resolved_skills = load_plugin_skills(
```
pakrym-oai, line 1749: codex, what is this change about?

codex-rs/core/src/codex.rs
```diff
@@ -496,10 +496,19 @@ impl Codex {
         let (tx_sub, rx_sub) = async_channel::bounded(SUBMISSION_CHANNEL_CAPACITY);
         let (tx_event, rx_event) = async_channel::unbounded();
 
+        let environment = environment_manager
+            .current()
+            .await
+            .map_err(|err| CodexErr::Fatal(format!("failed to create environment: {err}")))?;
+        let skill_fs = environment
```
pakrym-oai: just fs

codex-rs/core/src/thread_manager.rs
```diff
@@ -903,11 +907,23 @@ impl ThreadManagerState {
...
-        );
+        let environment = self
+            .environment_manager
+            .current()
+            .await
+            .map_err(|err| CodexErr::Fatal(format!("failed to create environment: {err}")))?;
+        let skill_fs = environment
+            .as_ref()
+            .map(|environment| environment.get_filesystem());
+        let watch_registration = self
```
pakrym-oai: only watch for local environment

codex-rs/core/tests/suite/skills.rs
```diff
@@ -146,14 +326,14 @@ async fn skill_load_errors_surface_in_session_configured() -> Result<()> {
             !skill
                 .path
                 .to_string_lossy()
-                .ends_with("skills/broken/SKILL.md")
+                .ends_with(".agents/skills/broken/SKILL.md")
```
pakrym-oai: why did this change?

codex-rs/core/src/plugins/manager.rs
pakrym-oai: Updated: addressed in c7a20296f2 by stacking this PR on `pakrym/make-loadplugin-async`. `load_plugins_from_layer_stack`/`load_plugin` are async now, so the `load_plugin_skills(...).await` call is back in config-layer plugin loading without adding a blocking sync bridge. The merge keeps this PR's filesystem-aware `SkillRoot` loading and local-only watcher behavior.

codex-rs/core/src/codex.rs
pakrym-oai: Addressed in 6924771106: renamed this to `fs`.

codex-rs/core/src/thread_manager.rs
pakrym-oai: Addressed in 6924771106: skill watcher registration now only uses the filesystem for a local environment; remote/no-env sessions get an empty watch registration.

codex-rs/core/tests/suite/skills.rs
pakrym-oai: That change was unnecessary. I restored this test to the home-skill `skills/broken/SKILL.md` path in 6924771106; repo skill coverage is handled by the separate remote-aware repo+home skills test.

starr-openai (approved):
let's add an ExecutorAbsolutePath or something like that sooner rather than later - this was we can just pass in the path object and many of these changes are drop in replacements.

As is we have to thread in the filesystem and remember to use that for all fs operations on the path (eg calling cwd.ancestors() right now is host local, not the executor side ancestors)

PR #17721, opened 2026-04-13
Stabilize marketplace/remove installedRoot test

xli-oai:
## Why

This addresses the review comment from #17751 about `marketplace/remove` app-server test portability: https://github.com/openai/codex/pull/17751#discussion_r3104378613

The API returns the removed installed root using the app-server's effective `CODEX_HOME`. On macOS, temporary directory paths can appear as either `/var/...` or `/private/var/...`, so comparing one raw path against another can fail even when `marketplace/remove` behaves correctly.

## What changed

- Removed the direct whole-response equality assertion for the installed root path.
- Asserted the stable response field, `marketplace_name`, directly.
- Compared the expected and returned installed-root paths after canonicalizing their existing parent directories, which avoids requiring the removed leaf directory to still exist.

## Verification

- `cargo test -p codex-app-server marketplace_remove_deletes_config_and_installed_root`
- `cargo test -p codex-app-server marketplace_remove`

### DIFF
--- a/codex-rs/app-server/tests/suite/v2/marketplace_remove.rs
+++ b/codex-rs/app-server/tests/suite/v2/marketplace_remove.rs
@@ -1,5 +1,6 @@
 use std::time::Duration;
 
+use anyhow::Context;
 use anyhow::Result;
 use app_test_support::McpProcess;
 use app_test_support::to_response;
@@ -10,7 +11,6 @@ use codex_app_server_protocol::RequestId;
 use codex_config::MarketplaceConfigUpdate;
 use codex_config::record_user_marketplace;
 use codex_core::plugins::marketplace_install_root;
-use codex_utils_absolute_path::AbsolutePathBuf;
 use pretty_assertions::assert_eq;
 use tempfile::TempDir;
 use tokio::time::timeout;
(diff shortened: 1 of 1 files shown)


codex-rs/cli/src/marketplace_cmd.rs
```diff
@@ -67,41 +95,674 @@ async fn run_add(args: AddMarketplaceArgs) -> Result<()> {
...
+
+    if !metadata.is_file() {
+        bail!(
+            "local marketplace source must be a file or directory: {}",
+            source_path.display()
+        );
+    }
+
+    if let Some(marketplace_root) = marketplace_root_for_manifest_path(source_path) {
+        copy_dir_recursive(marketplace_root, staged_root)?;
```
xl-openai: Do we really need to do this copy?

codex-rs/cli/src/marketplace_cmd.rs
```diff
@@ -67,41 +95,674 @@ async fn run_add(args: AddMarketplaceArgs) -> Result<()> {
...
+                    target_path.display()
+                )
+            })?;
+        }
+    }
+
+    Ok(())
+}
+
+async fn download_manifest_url_source(url: &str, staged_root: &Path) -> Result<()> {
```
xl-openai: What's this source used for? So we only download the marketplace.json? is it good enough?

0xsarawut: base.app/profile/0xsarawut

codex-rs/cli/src/marketplace_cmd.rs
xli-oai: I was debating with myself, so i asked codex and take its advice, but yea i think no copy makes perfect sense as well. Let me make the change.

> For the current marketplace add semantics, copying is the safer default.
> What happens today is that add behaves like an install/import step, not a live reference. Git sources already get cloned into the marketplace install area, and local roots are being treated the same way. Why that matters: if we don’t copy a local marketplace root, the installed marketplace becomes tied to whatever happens in that working directory afterward. Example: you add ~/code/my-marketplace, then switch branches, edit files, or delete the folder; without a copy, Codex would silently change or break the “installed” marketplace.

xl-openai (approved):

PR #17727, opened 2026-04-14
Show organization and groups in TUI /status menu

canvrno-oai:
Adds a best-effort ChatGPT account metadata read path for /status, showing organization and group names in the TUI when the active session is eligible. We perform an eligibility-only probe first, and only perform live metadata fetches for first-party ChatGPT workspace accounts.

  - Adds account/metadata/read with `metadataFetchEligible`, `accountDisplayName`, and `accountGroupNames`.
  - Surfaces organization and group names in the `/status` card, with an async loading state and no live metadata calls for ineligible sessions.
  - Updates app-server docs, generated schemas/types, and snapshot/test coverage.

image


canvrno-oai:
May revisit later, held for now in favor of #17877 to address a specific need.

PR #17743, opened 2026-04-14
Keep Local Bazel Cache Uploads Read Only By Default

evawong-oai:
## Problem

We are debugging a cache correctness issue in Codex where a local Bazel run appears to have reused a machine specific artifact from another developer. The immediate concern is that local machines can populate the shared remote cache by default.

Example Buildbudy link - https://app.buildbuddy.io/invocation/3d930d29-16ad-4d05-8eee-4c8198fcb78a

Locally building codex failed with this error for another user other than neil@:

`bazel   run //codex-rs/cli:codex
cxxbridge error: Failed to create directory /Users/neil/Library/Caches/bazel/_bazel_neil/ab4e3ea67871de3405481bb5c0f62f4f/sandbox/darwin-sandbox/1921/execroot/_main/bazel-out/darwin_arm64-opt-exec/bin/external/rules_rs++crate+crates__scra`

## Root Cause

1. `.bazelrc` points local builds at the shared BuildBuddy cache.
2. Before this change, local builds also uploaded local action results by default.
3. BuildBuddy side permissions may still matter, but the checked in repo config currently permits the local write path.

## Fix

1. Add `try-import %workspace%/user.bazelrc` so per user auth can live outside the repo.
2. Disable local cache uploads by default in the checked in `.bazelrc`.
3. Re enable uploads only for the `remote` config so CI and remote exec remain cache writers.
4. Ignore `user.bazelrc` in git.

## Validation

1. Inspected Bazel flag behavior for `remote_upload_local_results`.
2. Verified the staged diff is scoped to `.bazelrc` and `.gitignore`.
3. No full Bazel build was run locally because this is a checked in config only change.

## Follow Up

1. BuildBuddy side enforcement should still restrict writer permissions by key, role, or partition across machines.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,3 +1,5 @@
+try-import %workspace%/user.bazelrc
+
 common --repo_env=BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
 common --repo_env=BAZEL_NO_APPLE_CPP_TOOLCHAIN=1
 # Dummy xcode config so we don't need to build xcode_locator in repo rule.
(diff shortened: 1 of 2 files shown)


evawong-oai:
Closing this for now because BuildBuddy can enforce the split we want with developer owned keys for humans and an org read write key for CI, so the checked in bazelrc change is not needed for the main fix.

PR #17746, opened 2026-04-14
Support PermissionRequest updatedPermissions (addRules type)

abhinav-oai:
## What
- allow `PermissionRequest` hooks to return `decision.updatedPermissions` when `behavior` is `allow`
- validate that returned updates are an exact subset of the offered `permission_suggestions`
- apply selected `addRules` exec-policy suggestions to session memory or persisted rules state
- add schema/parser/runtime coverage and an end-to-end test for the persisted exec-rule path

## Why
- the previous PR only exposed permission suggestions to hooks
- this follow-up lets hooks actually accept and apply those suggestions so they can auto-approve exec requests by selecting the offered rule updates
- this keeps the first implementation narrow to `addRules`, which matches the current rules-file support

## Follow-up
- replaceRules and removeRules

### DIFF
--- a/codex-rs/core/src/project_doc.rs
+++ b/codex-rs/core/src/project_doc.rs
@@ -16,18 +16,13 @@
 //! 3.  We do **not** walk past the project root.
 
 use crate::config::Config;
-use crate::config_loader::ConfigLayerStackOrdering;
-use crate::config_loader::default_project_root_markers;
-use crate::config_loader::merge_toml_values;
-use crate::config_loader::project_root_markers_from_config;
-use codex_app_server_protocol::ConfigLayerSource;
+use crate::config_loader::resolve_project_root;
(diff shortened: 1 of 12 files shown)


codex-rs/core/src/project_doc.rs
```diff
@@ -226,51 +221,14 @@ pub async fn discover_project_doc_paths(
...
-                    project_root = Some(AbsolutePathBuf::try_from(ancestor.to_path_buf())?);
-                    break;
-                }
-            }
-            if project_root.is_some() {
-                break;
-            }
-        }
-    }
+    let project_root = resolve_project_root(&config.config_layer_stack, &dir).await?;
```
abhinav-oai, line 224: pull this into a helper in config_loader since we were doing similar work in hook_runtime

PR #17757, opened 2026-04-14
[codex] Implement turn-scoped hook interrupts

abhinav-oai:
## Summary
- treat `PermissionRequest` `deny { interrupt: true }` as a turn-scoped abort request instead of a detached session interrupt
- let the running task own `TurnAborted` finalization after its future unwinds, while keeping external aborts synchronous and deterministic
- thread the new interrupt-capable deny shape through hook parsing, schema, and the shell/tool/network approval paths

## Why
The first implementation detached abort work from the hook path. That created two races: the interrupted turn could still emit `TurnComplete`, and the delayed abort could later cancel a different active turn. A follow-up refactor that called `abort_turn()` from inside the running task fixed the first race but could still start replacement work before the interrupted task had actually exited.

This version keeps the interrupt request scoped to the current turn and moves self-abort finalization into the task runner, so successor work only starts after the interrupted task has really unwound.

## Validation
- `cargo test -p codex-hooks`
- `cargo test -p codex-core tools::handlers::multi_agents::tests::multi_agent_v2_followup_task_interrupts_busy_child_without_losing_message -- --exact --nocapture`
- `cargo test -p codex-core` *(still shows two unrelated existing failures: `plugins::manager::tests::list_marketplaces_ignores_installed_roots_missing_from_config` and `tools::js_repl::tests::js_repl_imported_local_files_can_access_repl_globals` in the sandboxed environment)*
- `just fmt`
- `just fix -p codex-hooks`
- `just fix -p codex-core`

### DIFF
--- a/codex-rs/core/src/function_tool.rs
+++ b/codex-rs/core/src/function_tool.rs
@@ -2,6 +2,8 @@ use thiserror::Error;
 
 #[derive(Debug, Error, PartialEq)]
 pub enum FunctionCallError {
+    #[error("turn interrupted")]
+    Interrupted,
     #[error("{0}")]
     RespondToModel(String),
     #[error("LocalShellCall without call_id or id")]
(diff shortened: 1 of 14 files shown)


abhinav-oai:
@codex review

PR #17763, opened 2026-04-14
Send sandbox state through MCP tool metadata

aaronl-openai:
## Changes

Allows MCPs to opt in to receiving sandbox config info through `_meta` on model-initiated tool calls. This lets MCPs adhere to the thread's sandbox if they choose to.

## Details

- Adds the `codex/sandbox-state-meta` experimental MCP capability.
- Tracks whether each MCP server advertises that capability.
- When a server opts in, `codex-core` injects the current `SandboxState` into model-initiated MCP tool-call request `_meta`.

## Verification

- added an integration test for the capability

### DIFF
--- a/codex-rs/codex-mcp/src/mcp_connection_manager.rs
+++ b/codex-rs/codex-mcp/src/mcp_connection_manager.rs
@@ -659,6 +659,27 @@ pub struct SandboxState {
     pub use_legacy_landlock: bool,
 }
 
+pub fn augment_request_meta_with_sandbox_state(
+    mut meta: Option<JsonValue>,
+    sandbox_state: SandboxState,
+) -> serde_json::Result<Option<JsonValue>> {
+    let sandbox_state = serde_json::to_value(sandbox_state)?;
+
+    match meta.as_mut() {
+        Some(JsonValue::Object(map)) => {
+            map.insert(MCP_SANDBOX_STATE_META_CAPABILITY.to_string(), sandbox_state);
+        }
+        Some(_) => {}
+        None => {
+            let mut map = Map::new();
+            map.insert(MCP_SANDBOX_STATE_META_CAPABILITY.to_string(), sandbox_state);
+            meta = Some(JsonValue::Object(map));
+        }
+    }
+
+    Ok(meta)
+}
+
 /// A thin wrapper around a set of running [`RmcpClient`] instances.
 pub struct McpConnectionManager {
     clients: HashMap<String, AsyncManagedClient>,
--- a/codex-rs/codex-mcp/src/mcp_connection_manager.rs
+++ b/codex-rs/codex-mcp/src/mcp_connection_manager.rs
@@ -659,27 +659,6 @@ pub struct SandboxState {
     pub use_legacy_landlock: bool,
 }
 
-pub fn augment_request_meta_with_sandbox_state(
-    mut meta: Option<JsonValue>,
-    sandbox_state: SandboxState,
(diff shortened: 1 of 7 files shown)


codex-rs/codex-mcp/src/mcp_connection_manager.rs
bolinfest, line 645: I should get rid of this, but I'll do that in a follow-up PR.

codex-rs/codex-mcp/src/mcp_connection_manager.rs
```diff
@@ -646,6 +647,7 @@ pub const MCP_SANDBOX_STATE_CAPABILITY: &str = "codex/sandbox-state";
 /// Custom MCP request to push sandbox state updates.
 /// When used, the `params` field of the notification is [`SandboxState`].
 pub const MCP_SANDBOX_STATE_METHOD: &str = "codex/sandbox-state/update";
+pub const MCP_SANDBOX_STATE_META_CAPABILITY: &str = "codex/sandbox-state-meta";
```
bolinfest, line 653: Can you add a docstring to explain what this capability does?

codex-rs/codex-mcp/src/mcp_connection_manager.rs
```diff
@@ -646,6 +647,7 @@ pub const MCP_SANDBOX_STATE_CAPABILITY: &str = "codex/sandbox-state";
 /// Custom MCP request to push sandbox state updates.
 /// When used, the `params` field of the notification is [`SandboxState`].
 pub const MCP_SANDBOX_STATE_METHOD: &str = "codex/sandbox-state/update";
```
bolinfest, line 649: I should be able to remove this, as well.

codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -480,6 +485,52 @@ async fn execute_mcp_tool_call(
...
+        .read()
+        .await
+        .server_supports_sandbox_state_meta_capability(server)
+        .await
+        .unwrap_or(false);
+    if !supports_sandbox_state_meta {
+        return Ok(meta);
+    }
+
+    let sandbox_state = serde_json::to_value(SandboxState {
```
bolinfest, line 506: `SandboxState` was introduced before we introduced `PermissionProfile`:

https://github.com/openai/codex/blob/d7f9d5cbce8b858614997e7d782feac8243909d8/codex-rs/protocol/src/models.rs#L106-L110

I think we should be passing that instead (and update `codex sandbox` to take it).

Also, I don't think you need `codex_linux_sandbox_exe`, though you likely want the path to `codex` itself?

bolinfest (commented):
It would be good to have an integration test for this.

codex-rs/core/src/mcp_tool_call.rs
aaronl-openai: one thing I'm concerned about is that it looks like `FileSystemSandboxPolicy`/`NetworkSandboxPolicy` are more expressive than `PermissionProfile`. for example, `FileSystemSandboxPolicy` can give readwrite access to `/some/folder` except readonly access to `/some/folder/readonly/`, which I don't think we can express with `PermissionProfile`.

bolinfest: Sorry, yes, I'll fix this up once I merge https://github.com/openai/codex/pull/15914

bolinfest (approved):

PR #17783, opened 2026-04-14
Preserve Cloudfare HTTP cookies in codex

shijie-oai:
## Summary
- Adds a process-local, in-memory cookie store for ChatGPT HTTP clients.
- Limits cookie storage and replay to a shared ChatGPT host allowlist.
- Wires the shared store into the default Codex reqwest client and backend client.
- Shares the ChatGPT host allowlist with remote-control URL validation to avoid drift.
- Enables reqwest cookie support and updates lockfiles.


codex-rs/codex-client/src/chatgpt_cookies.rs
```diff
@@ -0,0 +1,129 @@
...
+            self.jar.cookies(url)
+        } else {
+            None
+        }
+    }
+}
+
+/// Adds the process-local ChatGPT cookie jar used by Codex HTTP clients.
+///
+/// The jar is intentionally not persisted to disk. It only preserves cookies for ChatGPT backend
```
shijie-oai: This is the key - we do it in memory and no disk persistence so that new CLI/app-server instance get the latest cookies from initial request.

codex-rs/codex-client/src/chatgpt_hosts.rs
```diff
@@ -0,0 +1,36 @@
+/// Returns whether `host` is one of the ChatGPT hosts Codex is allowed to treat
+/// as first-party ChatGPT traffic.
+pub fn is_allowed_chatgpt_host(host: &str) -> bool {
+    host == "chatgpt.com"
+        || host.ends_with(".chatgpt.com")
+        || host == "chat.openai.com"
```
shijie-oai: I am not 100% sure if this is needed? Can we confirm?

codex-rs/codex-client/src/chatgpt_hosts.rs
```diff
@@ -0,0 +1,36 @@
+/// Returns whether `host` is one of the ChatGPT hosts Codex is allowed to treat
+/// as first-party ChatGPT traffic.
+pub fn is_allowed_chatgpt_host(host: &str) -> bool {
+    host == "chatgpt.com"
+        || host.ends_with(".chatgpt.com")
```
ccy-oai: these two overlap?
and what about chatgpt-preview-*?
also rust noob question: does rust have a set presence operator?

codex-rs/codex-client/src/chatgpt_cookies.rs
ccy-oai: qq what's the story between app-server requests and the client layer requests today? Would client layer have a different set of cookies and if they did would it pass down to app-server?

shijie-oai: Nope they would use the same one cause most if not all request goes through app server. There are definitely individual requests in the app (i.e. codex web tasks etc) but those are not concerning for now.

codex-rs/codex-client/src/chatgpt_hosts.rs
shijie-oai: > these two overlap?

one is strict check and one is looking at the ending so `staging.chatgpt.com` like url would qualify too. 

> chatgpt-preview-*

I am not familiar with and I am not sure if we can even point CLI traffic that way? 

> set presence operator

It most and we can convert to use that.

apanasenko-oai (approved):

shijie-oai:
@codex review

codex-rs/codex-client/src/chatgpt_hosts.rs
ccy-oai: > one is strict check and one is looking at the ending so staging.chatgpt.com like url would qualify too.

1st comment more around `if strict || loose` can we omit the `strict`

> preview-*

that's the personal sa-server staging deployment

codex-rs/codex-client/src/chatgpt_cookies.rs
ccy-oai: sg. just wasn't sure if that story behind the `vs-code` requests and `app-server` requests had evolved

codex-rs/app-server/src/transport/remote_control/protocol.rs
```diff
@@ -201,6 +199,16 @@ mod tests {
                         .to_string(),
             }
         );
+        assert_eq!(
+            normalize_remote_control_url("https://chat.openai.com/backend-api")
+                .expect("chat.openai.com URL should normalize"),
+            RemoteControlTarget {
+                websocket_url: "wss://chat.openai.com/backend-api/wham/remote/control/server"
```
ccy-oai: did we touch websocket?

ccy-oai (approved):

codex-rs/codex-client/src/chatgpt_cookies.rs
```diff
@@ -0,0 +1,231 @@
...
+/// those cookies, the store must be scoped to the auth/session owner instead of shared globally.
+pub fn with_chatgpt_cloudflare_cookie_store(
+    builder: reqwest::ClientBuilder,
+) -> reqwest::ClientBuilder {
+    builder.cookie_provider(Arc::clone(&SHARED_CHATGPT_CLOUDFLARE_COOKIE_STORE))
+}
+
+fn is_chatgpt_cookie_url(url: &reqwest::Url) -> bool {
+    match url.scheme() {
+        "http" | "https" => {}
```
jif-oai: I think this should reject http for non-local ChatGPT hosts for multiple security reasons

codex-rs/app-server/src/transport/remote_control/protocol.rs
```diff
@@ -1,5 +1,6 @@
 use crate::outgoing_message::OutgoingMessage;
```
jif-oai, line 1: This couples remote-control endpoint validation to the cookie/client host allowlist. Those are different security boundaries: a host that is safe for Cloudflare cookie replay is not automatically a valid remote-control websocket/enrollment endpoint. Can we keep a remote-control-specific predicate here, or move the shared data into a small policy module that forces each caller to opt into the right use case?

jif-oai (approved):
We need to make sure everyone uses the codex-client. After quickly checking, I could find `cloud-tasks/src/env_detect.rs` for example that does not rely on it (which is not good though)

Ok after processing my other comments

codex-rs/app-server/src/transport/remote_control/protocol.rs
shijie-oai: Sorry did not get to reply earlier - agree with the assessment and I will break it up again.

codex-rs/app-server/src/transport/remote_control/protocol.rs
shijie-oai: This is for remote control only and I think I will the chagpt_host check consolidation here.

pakrym-oai (approved):

PR #17807, opened 2026-04-14
Use AgentAssertion downstream behind use_agent_identity

adrian-openai:
Recreates the old PR4 on top of the restored PR3 branch after the accidental stack collapse.\n\nThis branch contains the original AgentAssertion changes plus the fixture/task-shape adjustments needed after PR3's task persistence changes.


adrian-openai:
This PR also had a lot of challenges with its ancestry drift, so we're moving here instead: https://github.com/openai/codex/pull/17980

Sorry for all of the moves!

PR #17820, opened 2026-04-14
feat: add AWS SigV4 auth for OpenAI-compatible model providers

celia-oai:
## Summary

Add first-class Amazon Bedrock Mantle provider support so Codex can keep using its existing Responses API transport with OpenAI-compatible AWS-hosted endpoints such as AOA/Mantle.

This is needed for the AWS launch path, where provider traffic should authenticate with AWS credentials instead of OpenAI bearer credentials. Requests are authenticated immediately before transport send, so SigV4 signs the final method, URL, headers, and body bytes that `reqwest` will send.

## What Changed

- Added a new `codex-aws-auth` crate for loading AWS SDK config, resolving credentials, and signing finalized HTTP requests with AWS SigV4.
- Added a built-in `amazon-bedrock` provider that targets Bedrock Mantle Responses endpoints, defaults to `us-east-1`, supports region/profile overrides, disables WebSockets, and does not require OpenAI auth.
- Added Amazon Bedrock auth resolution in `codex-model-provider`: prefer `AWS_BEARER_TOKEN_BEDROCK` when set, otherwise use AWS SDK credentials and SigV4 signing.
- Added `AuthProvider::apply_auth` and `Request::prepare_body_for_send` so request-signing providers can sign the exact outbound request after JSON serialization/compression.
- Determine the region by taking the `aws.region` config first (required for bearer token codepath), and fallback to SDK default region.

## Testing
Amazon Bedrock Mantle Responses paths:

- Built the local Codex binary with `cargo build`.
- Verified the custom proxy-backed `aws` provider using `env_key = "AWS_BEARER_TOKEN_BEDROCK"` streamed raw `responses` output with `response.output_text.delta`, `response.completed`, and `mantle-env-ok`.
- Verified a full `codex exec --profile aws` turn returned `mantle-env-ok`.
- Confirmed the custom provider used the bearer env var, not AWS profile auth: bogus `AWS_PROFILE` still passed, empty env var failed locally, and malformed env var reached Mantle and failed with `401 invalid_api_key`.
- Verified built-in `amazon-bedrock` with `AWS_BEARER_TOKEN_BEDROCK` set passed despite bogus AWS profiles, returning `amazon-bedrock-env-ok`.
- Verified built-in `amazon-bedrock` SDK/SigV4 auth passed with `AWS_BEARER_TOKEN_BEDROCK` unset and temporary AWS session env credentials, returning `amazon-bedrock-sdk-env-ok`.


celia-oai:
@codex review

celia-oai:
@codex review

codex-rs/model-provider-info/src/lib.rs
```diff
@@ -90,6 +90,8 @@ pub struct ModelProviderInfo {
     pub experimental_bearer_token: Option<String>,
     /// Command-backed bearer-token configuration for this provider.
     pub auth: Option<ModelProviderAuthInfo>,
+    /// AWS SigV4 auth configuration for this provider.
+    pub aws: Option<ModelProviderAwsAuthInfo>,
```
pakrym-oai, line 98: I'm not a huge fan of having this exposed on the model info struct itself. Ideally the rest of the code shouldn't care about details of aws. Still reviwing though.

codex-rs/model-provider/src/aws_auth_provider.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+#[async_trait::async_trait]
+impl AuthProvider for AwsSigV4AuthProvider {
+    fn add_auth_headers(&self, _headers: &mut HeaderMap) {}
+
+    /// AWS SigV4 requests are sent to OpenAI-compatible provider endpoints, not
+    /// Codex's legacy Responses backend, so avoid signing and forwarding the
+    /// Codex-private `session_id` compatibility header.
+    fn should_send_legacy_conversation_header(&self) -> bool {
```
pakrym-oai: Can we still send it? Session id is imporant header for server impl.

codex-rs/model-provider/src/auth.rs
```diff
@@ -60,5 +66,13 @@ pub(crate) fn resolve_provider_auth(
     auth: Option<&CodexAuth>,
     provider: &ModelProviderInfo,
 ) -> codex_protocol::error::Result<SharedAuthProvider> {
+    if let Some(aws) = provider.aws.as_ref() {
```
pakrym-oai: Why is this logic not in the ModelProvider implemenation?

pakrym-oai: We should encapsulate provider-specific logic in ModelProvider implemenation and not spread it as ifs though the code.

codex-rs/codex-client/src/request.rs
```diff
@@ -63,6 +64,131 @@ impl Request {
         self.compression = compression;
         self
     }
+
+    /// Convert the request body into the exact bytes that will be sent.
+    ///
+    /// Auth schemes such as AWS SigV4 need to sign the final body bytes, including
+    /// compression and content headers. Calling this method is idempotent for
+    /// already-finalized raw request bodies.
+    pub fn prepare_body_for_send(&mut self) -> Result<Bytes, String> {
```
pakrym-oai: Can we avoid mutating the request itself in this method? Or at least just cache the computed value without rewriting main fields. 

Feels wrong to both mutate and return new values from the same method.

codex-rs/codex-api/src/auth.rs
```diff
@@ -1,13 +1,53 @@
...
+#[async_trait]
 pub trait AuthProvider: Send + Sync {
+    /// Adds any auth headers that are available without request body access.
+    ///
+    /// Implementations should be cheap and non-blocking. This method is also
+    /// used by telemetry and non-HTTP request paths.
     fn add_auth_headers(&self, headers: &mut HeaderMap);
+
+    /// Whether Responses requests should include Codex's legacy `session_id` header.
+    fn should_send_legacy_conversation_header(&self) -> bool {
```
pakrym-oai: let's rem

codex-rs/codex-api/src/endpoint/session.rs
```diff
@@ -55,7 +60,6 @@ impl<T: HttpTransport> EndpointSession<T> {
         if let Some(body) = body {
             req.body = Some(RequestBody::Json(body.clone()));
         }
-        self.auth.add_auth_headers(&mut req.headers);
```
pakrym-oai, line 58: why did we have to move auth from here?

codex-rs/codex-api/src/auth.rs
```diff
@@ -1,13 +1,53 @@
...
 use std::sync::Arc;
 
-/// Adds authentication headers to API requests.
+/// Error returned while applying authentication to an outbound request.
+#[derive(Debug, thiserror::Error)]
+pub enum AuthError {
+    #[error("request auth build error: {0}")]
+    Build(String),
+    #[error("transient auth error: {0}")]
+    Transient(String),
```
pakrym-oai, line 13: What scenario is this used for?

codex-rs/aws-auth/src/lib.rs
```diff
@@ -0,0 +1,259 @@
...
+        let credentials = self.credentials_provider.provide_credentials().await?;
+        signing::sign_request(&credentials, &self.region, &self.service, request, time)
+    }
+}
+
+impl AwsAuthError {
+    /// Returns whether retrying the outbound request can reasonably recover from this auth error.
+    pub fn is_retryable(&self) -> bool {
+        match self {
+            AwsAuthError::Credentials(error) => matches!(
```
pakrym-oai, line 118: are these actually retryable? Can we recover from CredentialsNotLoaded?

codex-rs/aws-auth/src/lib.rs
```diff
@@ -0,0 +1,259 @@
...
+                session_token.map(str::to_string),
+                /*expires_after*/ None,
+                "unit-test",
+            )),
+            region: "us-east-1".to_string(),
+            service: "bedrock".to_string(),
+        }
+    }
+
+    fn test_request() -> AwsRequestToSign {
```
pakrym-oai, line 161: is it possible to add integration tests for AWS model provider? Set up some basic fake auth that AWS SDK will pick up and assert that the request is being built correctly?

codex-rs/aws-auth/src/signing.rs
```diff
@@ -0,0 +1,76 @@
...
+        .time(time)
+        .settings(SigningSettings::default())
+        .build()
+        .map_err(|err| AwsAuthError::SigningParams(err.to_string()))?;
+    let (instructions, _signature) = sign(signable_request, &signing_params.into())
+        .map_err(AwsAuthError::SigningFailure)?
+        .into_parts();
+
+    let uri = Uri::from_str(&request.url).map_err(AwsAuthError::InvalidUri)?;
+    let mut http_request = Request::builder()
```
pakrym-oai, line 56: Do we have to go through a temp request? Can we grab parts we need from instructions?

pakrym-oai (changes requested):
The main issue is that we are not using the ModelProvider infrastructure we've added that allows us to replace feature based on the provider.

codex-rs/aws-auth/src/lib.rs
celia-oai: I can do it as a follow-up - seems doable. for now, I have manual tests locally that do exactly this.

codex-rs/model-provider/src/aws_auth_provider.rs
celia-oai: actually, the mantle path strips the header with underscore, so this is not possible.

celia-oai: I made the logic a bit more straightforward so we just remove this header for mantle path, not as a function for generic auth_provider.

codex-rs/aws-auth/src/lib.rs
celia-oai: it's not really documented which ones are retryable but agreed that this doesn't seem retryable. kept the timeout case and removed this one

codex-rs/codex-client/src/request.rs
celia-oai: good catch

codex-rs/aws-auth/src/signing.rs
celia-oai: according to codex: instructions can also contain query params for presigned/query-param signing, and apply_to_request_http1x handles those correctly.

codex-rs/codex-api/src/auth.rs
celia-oai: given this mapping:
```
 match error {
            AuthError::Build(message) => TransportError::Build(message),
            AuthError::Transient(message) => TransportError::Network(message),
        }
```
Transient errors are retryable via `TransportError::Network`. this is for retrying e.g. provider timeout error.

codex-rs/codex-api/src/endpoint/session.rs
celia-oai: this is moved later in run_with_request_telemetry to be part of apply_auth. this allows the aws path to sign the final request sent with query params/ compressions/etc.

celia-oai:
not a huge fan of this code that reads region from api key, feels very error prone. so I changed the logic back to reading from config. I can add this back as a fallback if needed.
```
pub fn region_from_bedrock_bearer_token(token: &str) -> Result<String, AwsAuthError> {
    const PREFIX: &str = "bedrock-api-key-";

    let token_body = token
        .trim()
        .strip_prefix(PREFIX)
        .ok_or_else(|| invalid_bedrock_bearer_token("missing bedrock-api-key prefix"))?;
    let encoded_token = token_body
        .split_once("&Version=")
        .map_or(token_body, |(encoded, _)| encoded);
    let decoded = general_purpose::STANDARD
        .decode(encoded_token)
        .map_err(|_| invalid_bedrock_bearer_token("base64 payload could not be decoded"))?;
    let decoded = String::from_utf8(decoded)
        .map_err(|_| invalid_bedrock_bearer_token("decoded payload is not UTF-8"))?;
    let decoded_url = if decoded.starts_with("http://") || decoded.starts_with("https://") {
        decoded
    } else {
        format!("https://{decoded}")
    };
    let url = Url::parse(&decoded_url)
        .map_err(|_| invalid_bedrock_bearer_token("decoded payload is not a URL"))?;
    let credential = url
        .query_pairs()
        .find_map(|(key, value)| (key == "X-Amz-Credential").then_some(value.into_owned()))
        .ok_or_else(|| invalid_bedrock_bearer_token("missing X-Amz-Credential"))?;
    let mut parts = credential.split('/');
    let _access_key = parts.next();
    let _date = parts.next();
    let region = parts
        .next()
        .filter(|region| !region.trim().is_empty())
        .ok_or_else(|| invalid_bedrock_bearer_token("credential scope is missing region"))?;

    Ok(region.to_string())
}
```

codex-rs/model-provider/src/provider.rs
```diff
@@ -53,6 +55,20 @@ pub fn create_model_provider(
...
+        let aws = provider_info
+            .aws
+            .clone()
+            .unwrap_or(ModelProviderAwsAuthInfo {
+                profile: None,
+                region: None,
+            });
+        return Arc::new(AmazonBedrockModelProvider {
+            info: provider_info,
+            aws,
```
pakrym-oai, line 68: a bit funky to pull out aws from provider_info that already carries it. non blocking

codex-rs/model-provider/src/provider.rs
```diff
@@ -123,4 +140,39 @@ mod tests {
...
+                }),
+                requires_openai_auth: false,
+                ..ModelProviderInfo::create_openai_provider(/*base_url*/ None)
+            },
+            Some(AuthManager::from_auth_for_testing(CodexAuth::from_api_key(
+                "custom-api-key",
+            ))),
+        );
+
+        assert!(provider.auth_manager().is_some());
```
pakrym-oai: This is wrong, righ?

codex-rs/model-provider/src/amazon_bedrock/auth.rs
```diff
@@ -0,0 +1,231 @@
...
+        region: String,
+    },
+    AwsSdkAuth {
+        config: AwsAuthConfig,
+        context: AwsAuthContext,
+    },
+}
+
+async fn resolve_auth_method(aws: &ModelProviderAwsAuthInfo) -> Result<BedrockAuthMethod> {
+    if let Some(token) = bearer_token_from_env() {
```
pakrym-oai, line 39: does amazon sdk not resolve this?

codex-rs/model-provider/src/amazon_bedrock/auth.rs
```diff
@@ -0,0 +1,231 @@
...
+        let _ = cell.set(context);
+        Self {
+            config,
+            context: cell,
+        }
+    }
+
+    async fn context(&self) -> std::result::Result<&AwsAuthContext, AuthError> {
+        self.context
+            .get_or_try_init(|| AwsAuthContext::load(self.config.clone()))
```
pakrym-oai, line 128: do we ever create without context?

pakrym-oai: Seems like with_context always sets a value.

codex-rs/codex-api/src/auth.rs
```diff
@@ -1,13 +1,48 @@
...
+    /// used by telemetry and non-HTTP request paths.
     fn add_auth_headers(&self, headers: &mut HeaderMap);
+
+    /// Applies auth to a complete outbound request.
+    ///
+    /// Header-only auth providers can rely on the default implementation.
+    /// Request-signing providers can override this to inspect the final URL,
+    /// headers, and body bytes before the transport sends the request.
+    async fn apply_auth(&self, mut request: Request) -> Result<Request, AuthError> {
+        self.add_auth_headers(&mut request.headers);
```
pakrym-oai, line 50: Is the contract of this method that the request is mutated or that a request is returned?

pakrym-oai (approved):

codex-rs/model-provider/src/provider.rs
celia-oai: I think the goal is to eventually pull out provider info so I left this shape, but happy to collapse if you think it's better

codex-rs/model-provider/src/provider.rs
celia-oai: sorry this is a horrible test, removing

codex-rs/model-provider/src/amazon_bedrock/auth.rs
celia-oai: nope, this is the only one outside of sdk path because it's bearer token based

codex-rs/model-provider/src/amazon_bedrock/auth.rs
celia-oai: I think there's some benefit of lazy-loading, so I kept this logic. lmk what you think

codex-rs/codex-api/src/auth.rs
celia-oai: ah good catch! I'll keep the returned path and update the docstring to make it clear

codex-rs/model-provider/src/amazon_bedrock/auth.rs
```diff
@@ -0,0 +1,232 @@
...
+async fn resolve_auth_method(aws: &ModelProviderAwsAuthInfo) -> Result<BedrockAuthMethod> {
+    if let Some(token) = bearer_token_from_env() {
+        let region = bearer_token_region_from_config(aws)?;
+        return Ok(BedrockAuthMethod::EnvBearerToken { token, region });
+    }
+
+    let config = aws_auth_config(aws);
+    let context = AwsAuthContext::load(config.clone())
+        .await
+        .map_err(aws_auth_error_to_codex_error)?;
```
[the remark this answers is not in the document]
celia-oai: this is expected behavior

PR #17659, opened 2026-04-13
thread store interface

wiltzius-openai:
Introduce a ThreadStore interface for mediating access to the filesystem (rollout jsonl files + sqlite db) based thread storage.

In later PRs we'll move the existing fs code behind a "local" implementation of this ThreadStore interface.

This PR should be a no-op behaviorally, it only introduces the interface.

PR #17824, opened 2026-04-14
[codex] Add local thread store listing

wiltzius-openai:
Builds on top of #17659 

Move the filesystem + sqlite thread listing-related operations inside of a local ThreadStore implementation and call ThreadStore from the places that used to perform these filesystem/sqlite operations.

This is the first of a series of PRs that will implement the rest of the local ThreadStore.

Testing:
- added unit tests for the thread store implementation
- adjusted some unit tests in the realtime + personality packages whose callsites changed. Specifically I'm trying to hide ThreadMetadata inside of the local implementation and make ThreadMetadata a sqlite implementation detail concern rather than a public interface, preferring the more generate StoredThread interface instead
- added a corner case test for the personality migration package that wasn't covered by the existing test suite
- adjust the behavior of searched thread listing to run the existing local rollout repair/backfill pass _before_ querying SQLite results, so callers using ThreadStore::list_threads do not miss matches after a partial metadata warm-up

### DIFF
--- a/codex-rs/core/src/realtime_context.rs
+++ b/codex-rs/core/src/realtime_context.rs
@@ -4,8 +4,10 @@ use crate::event_mapping::is_contextual_user_message_content;
 use chrono::Utc;
 use codex_git_utils::resolve_root_git_project_for_trust;
 use codex_protocol::models::ResponseItem;
-use codex_state::SortKey;
-use codex_state::ThreadMetadata;
+use codex_thread_store::ListThreadsParams;
+use codex_thread_store::StoredThread;
+use codex_thread_store::ThreadSortKey;
+use codex_thread_store::ThreadStore;
 use codex_utils_output_truncation::TruncationPolicy;
 use codex_utils_output_truncation::truncate_text;
 use dirs::home_dir;
@@ -98,7 +100,7 @@ pub(crate) async fn build_realtime_startup_context(
     }
     if let Some(section) = format_section(
         "Notes",
-        Some("Built at realtime startup from the current thread history, persisted thread metadata in the state DB, and a bounded local workspace scan. This excludes repo memory instructions, AGENTS files, project-doc prompt blends, and memory summaries.".to_string()),
+        Some("Built at realtime startup from the current thread history, local thread metadata, and a bounded local workspace scan. This excludes repo memory instructions, AGENTS files, project-doc prompt blends, and memory summaries.".to_string()),
         NOTES_SECTION_TOKEN_BUDGET,
     ) {
         parts.push(section);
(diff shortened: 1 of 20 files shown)


codex-rs/rollout/src/recorder.rs
```diff
@@ -279,7 +279,7 @@ impl RolloutRecorder {
         let codex_home = config.codex_home();
         let state_db_ctx = state_db::get_state_db(config).await;
 
-        if search_term.is_some()
+        if search_term.is_none()
```
wiltzius-openai: this early return was only safe for unsearched listing; for searched listing it could return a partially populated SQLite result before the existing rollout repair pass had a chance to backfill older matching threads

codex-rs/thread-store/src/types.rs
```diff
@@ -135,6 +135,8 @@ pub struct ThreadPage {
 pub struct StoredThread {
     /// Thread id.
     pub thread_id: ThreadId,
+    /// Local rollout path when the backing store is filesystem-based.
+    pub rollout_path: Option<PathBuf>,
```
pakrym-oai, line 139: is this temporary?

codex-rs/core/src/realtime_context.rs
```diff
@@ -117,33 +120,31 @@ pub(crate) async fn build_realtime_startup_context(
...
-            MAX_RECENT_THREADS,
-            /*anchor*/ None,
-            SortKey::UpdatedAt,
-            &[],
-            /*model_providers*/ None,
-            /*archived_only*/ false,
-            /*search_term*/ None,
-        )
+async fn load_recent_threads(sess: &Session) -> Vec<StoredThread> {
+    let config = sess.get_config().await;
```
pakrym-oai: any reason not to toss LocalThreadStore onto sess.services?

codex-rs/thread-store/src/local/mod.rs
```diff
@@ -0,0 +1,565 @@
...
+}
+
+async fn list_rollout_threads(
+    config: &RolloutConfig,
+    params: &ListThreadsParams,
+    cursor: Option<&codex_rollout::Cursor>,
+    sort_key: codex_rollout::ThreadSortKey,
+) -> ThreadStoreResult<codex_rollout::ThreadsPage> {
+    let page = if params.archived {
+        RolloutRecorder::list_archived_threads(
```
pakrym-oai, line 148: maybe worth inlining later separately

codex-rs/thread-store/src/types.rs
wiltzius-openai: It's temporary in the sense that an optional path is necessary in the ThreadStore until we make the path non-required in the AppServer protocol. The app server currently has a non-nullable path field when returning summaries of listed threads, so the ThreadStore sort of has to provide it.

I'm trying to do as much of this migration as possible underneath the current app server shape, but we will eventually need to change that layer too.

Do you have a preferred approach for how to handle that?

codex-rs/core/src/realtime_context.rs
wiltzius-openai: don't think so, can do

codex-rs/thread-store/src/local/mod.rs
wiltzius-openai: by inline do you mean moving the archived/unarchived distinction into a param on the rollout recorder methods?

pakrym-oai: I mean getting rid of RolloutRecorder and having methods be native on LocalThreadStorage.

RolloutRecorder was the previous place where we were abstracting away rollout management.

codex-rs/thread-store/src/types.rs
pakrym-oai: I think keeping it here for a bit is fine especially considering rollout paths are all over app-server.

codex-rs/thread-store/src/local/mod.rs
wiltzius-openai: ah I see. Yes we can do that over time!

pakrym-oai (approved):

PR #17825, opened 2026-04-14
[codex] Revoke ChatGPT tokens on logout

sashank-oai:
## Summary

This changes Codex logout so managed ChatGPT auth is revoked against AuthAPI before local auth state is removed. CLI logout, TUI `/logout`, and the app-server account logout path now use the token-revoking logout flow instead of only deleting `auth.json` / credential store state.

## Root Cause

Logout previously cleared only local auth storage. That removed Codex's local credentials but did not ask the backend to invalidate the refresh/access token state associated with a managed ChatGPT login.

## Behavior

For managed ChatGPT auth, logout sends the stored refresh token to `https://auth.openai.com/oauth/revoke` with `token_type_hint: refresh_token` and the Codex OAuth client id, then deletes all local auth stores after revocation succeeds. If only an access token is available, it falls back to revoking that access token. API key auth and externally supplied `chatgptAuthTokens` are still only cleared locally because Codex does not own a refresh token for those modes.

Revocation failures are fail-closed: if Codex cannot load stored auth or the backend revoke call fails, logout returns an error and leaves local auth in place so the user can retry instead of silently clearing local state while backend tokens remain valid.

## Validation
ran local version of `codex-cli` with staging overrides/harness for auth

ran `codex login` then `codex logout`:

saw auth.json clear and  backend revocation endpoints were called

```
POST /oauth/revoke
status: 200

revoking access token
should clear auth session
clearing auth session due to token revocation
successfully revoked session and access token
CANONICAL-API-LINE Response: status='200' method='POST' path='/oauth/revoke
```


codex-rs/login/src/auth/revoke.rs
```diff
@@ -0,0 +1,170 @@
...
+use codex_client::CodexHttpClient;
+
+use super::manager::CLIENT_ID;
+use super::manager::REFRESH_TOKEN_URL_OVERRIDE_ENV_VAR;
+use super::storage::AuthDotJson;
+use super::util::try_parse_error_message;
+use crate::default_client::create_client;
+use crate::token_data::TokenData;
+
+const REVOKE_TOKEN_URL: &str = "https://auth.openai.com/oauth/revoke";
```
cooper-oai: i'm not super sure where we store OAI-specific config but it may be worth poking around to see if there is an existing place for that in the code

codex-rs/login/src/auth/revoke.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum RevokeTokenKind {
+    Access,
+    Refresh,
+}
+
+impl RevokeTokenKind {
+    fn hint(self) -> &'static str {
```
cooper-oai: any reason we need both this and `label`? maybe we can use `impl Into<&str>`?

codex-rs/login/src/auth/manager.rs
```diff
@@ -428,6 +430,32 @@ pub fn logout(
...
+        Ok(auth_dot_json) => auth_dot_json,
+        Err(err) => {
+            let removed = logout_all_stores(codex_home, auth_credentials_store_mode)?;
+            return Ok(LogoutResult {
+                removed,
+                revoke_error: Some(format!("failed to load stored auth for revocation: {err}")),
+            });
+        }
+    };
+    let revoke_error = revoke_auth_tokens(auth_dot_json.as_ref())
```
cooper-oai: do we really want fail open semantics?

sashank-oai:

codex-rs/tui/src/chatwidget/slash_dispatch.rs
```diff
@@ -235,13 +235,24 @@ impl ChatWidget {
...
-                    &self.config.codex_home,
-                    self.config.cli_auth_credentials_store_mode,
-                ) {
-                    tracing::error!("failed to logout: {e}");
-                }
-                self.request_quit_without_confirmation();
+                let codex_home = self.config.codex_home.clone();
+                let auth_credentials_store_mode = self.config.cli_auth_credentials_store_mode;
+                let app_event_tx = self.app_event_tx.clone();
+                tokio::spawn(async move {
```
etraut-openai: We should change this to call the app server API for logging out rather than calling the local version. Otherwise this won't work for scenarios where the TUI is connected to a remote host.

This was an existing bug (that I wasn't aware of). Now that we know about it, we should probably fix it here. If you decide not to fix it in this PR, let me know and I can do a follow-up fix.

codex-rs/login/src/auth/revoke.rs
```diff
@@ -0,0 +1,203 @@
+use serde::Serialize;
```
etraut-openai, line 7: Nit: Add header comment explaining what this module does.

codex-rs/login/src/auth/manager.rs
```diff
@@ -483,6 +486,24 @@ pub fn logout(
     storage.delete()
 }
 
+pub async fn logout_with_revoke(
+    codex_home: &Path,
+    auth_credentials_store_mode: AuthCredentialsStoreMode,
+) -> std::io::Result<bool> {
+    let storage = create_auth_storage(codex_home.to_path_buf(), auth_credentials_store_mode);
```
etraut-openai: Why are we reloading the auth information from disk here rather than using the AuthManager? It is meant to be the source of truth for the current auth and any tokens associated with it. By going around the AuthManager's back and loading the auth information directly from disk, you're breaking the internal contract.

etraut-openai (changes requested):

etraut-openai (approved):
Looks good!

PR #17826, opened 2026-04-14
[codex] Add remote thread store implementation

wiltzius-openai:
- Add a "remote" thread store implementation
- Implement the remote thread store as a thin wrapper that makes grpc calls to a configurable service endpoint
- Implement only the thread/list method to start
- Encode the grpc method/param shape as protobufs in the remote implementation

A wart: the proto generation script is an "example" binary target. This is an example target only because Cargo lets examples use dev-dependencies, which keeps tonic-prost-build out of the normal codex-thread-store dependency surface. A regular bin would either need to add proto generation deps as normal runtime deps, or use a feature-gated optional dep, which this repo’s manifest checks explicitly reject.


codex-rs/thread-store/src/remote/mod.rs
```diff
@@ -0,0 +1,496 @@
...
+}
+
+fn proto_source(kind: proto::SessionSourceKind) -> proto::SessionSource {
+    proto::SessionSource {
+        kind: kind.into(),
+        ..Default::default()
+    }
+}
+
+fn stored_thread_from_proto(thread: proto::StoredThread) -> ThreadStoreResult<StoredThread> {
```
rasmusrygaard: Would it make sense to write a test that asserts that you can roundtrip from `proto::StoredThread` to `StoredThread` and back? Not sure what's the best way to make sure that the mapping stays in sync

wiltzius-openai: good idea, added

codex-rs/thread-store/scripts/generate-proto.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
```
starr-openai, line 1: for a follow-up -- if we anticipate any additional grpc/protobuf apis it might be worth having a consolidated place for those so we can re-use the gen code

codex-rs/thread-store/src/remote/proto/codex.thread_store.v1.proto
```diff
@@ -0,0 +1,86 @@
...
+  THREAD_SORT_KEY_CREATED_AT = 0;
+  THREAD_SORT_KEY_UPDATED_AT = 1;
+}
+
+message ListThreadsResponse {
+  repeated StoredThread threads = 1;
+  optional string next_cursor = 2;
+}
+
+message StoredThread {
```
starr-openai, line 33: is there a good way to keep these in sync if we add new fields to StoredThread?

starr-openai: I suppose we could use the proto ones as source of truth and derive from that in the rest of the code, but that's more committing.  keeping separate+sync seems ok for now

wiltzius-openai: Currently requires manual updates. I did explore some autogen approaches, but ultimately decided that in the Codex era its almost easier to have the bot mechanically trace and update the fields than it is to get it to reliably understand + run the autogen script. Don't feel strongly about this so can explore autogen further, but given how awkward even the protobuf -> rust proto impl autogen script is it'll probably be a little goofy.

However, as a guardrail, I implemented the test Rasmus mentioned above. Specifically make sure we can round trip a stored thread from the native rust type to proto wire format and back. Doesn't totally guarantee we never miss fields (since after all you have to update the test) but it's at least some protection against drift.

codex-rs/thread-store/scripts/generate-proto.sh
wiltzius-openai: Yeah -- before generalizing I would love to come up with a more elegant way to do the generation. This "example" target is pretty goofy. It would be better I think as a dedicated part of the build step, but that requires some project-level build changes I didn't want to embark on until this has more momentum.

starr-openai (approved):

PR #17831, opened 2026-04-14
[codex][mcp] Add resource uri meta to tool call item.

mzeng-openai:
- [x] Add resource uri meta to tool call item so that the app-server client can start prefetching resources immediately without loading mcp server status.


codex-rs/app-server-protocol/src/protocol/thread_history.rs
```diff
@@ -518,11 +519,11 @@ impl ThreadHistoryBuilder {
         let duration_ms = i64::try_from(payload.duration.as_millis()).ok();
         let (result, error) = match &payload.result {
             Ok(value) => (
-                Some(McpToolCallResult {
+                Some(Box::new(McpToolCallResult {
```
shijie-oai, line 522: why the change?

codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -1063,6 +1077,25 @@ pub(crate) async fn lookup_mcp_tool_metadata(
     })
 }
 
+fn get_mcp_app_resource_uri(
+    meta: Option<&serde_json::Map<String, serde_json::Value>>,
+) -> Option<String> {
+    let meta = meta?;
```
shijie-oai: nit: is this reassignment needed?

shijie-oai (approved):

codex-rs/app-server-protocol/src/protocol/thread_history.rs
mzeng-openai: This is intentional. Adding `mcp_app_resource_uri` grows the `ThreadItem::McpToolCall` variant; boxing the result keeps every `ThreadItem` from carrying the full result payload inline. It does not change the serialized JSON or generated TypeScript shape.

codex-rs/core/src/mcp_tool_call.rs
mzeng-openai: Done in e57039243f. Switched this to `meta.and_then(...)` so we avoid shadowing/reassigning the parameter while keeping the same nested/flat/outputTemplate fallback behavior.

PR #17836, opened 2026-04-14
[codex] Add tmux-aware OSC 9 notifications

caseychow-oai:
## Summary
- wrap OSC 9 notifications in tmux's DCS passthrough so terminal notifications make it through tmux
- use codex-terminal-detection for OSC 9 auto-selection so tmux sessions inherit the underlying client terminal support
- add focused notification backend tests for plain OSC 9 and tmux-wrapped output

## Stack
- base PR: #18479
- review order: #18479, then this PR

## Why
Tmux does not forward OSC 9 notifications directly; the sequence has to be wrapped in tmux's DCS passthrough envelope. Codex also had local notification heuristics that could miss supported terminals when running under tmux, even though codex-terminal-detection already knows how to attribute tmux sessions to the client terminal.

## Validation
- `just fmt`
- `cargo test -p codex-tui` *(currently blocked by an unrelated existing compile error in `app-server/src/message_processor.rs:754` referencing `connection_id` out of scope; not caused by this change)*

### DIFF
--- a/codex-rs/tui/src/notifications/mod.rs
+++ b/codex-rs/tui/src/notifications/mod.rs
@@ -1,11 +1,13 @@
 mod bel;
 mod osc9;
 
-use std::env;
 use std::io;
 
 use bel::BelBackend;
 use codex_config::types::NotificationMethod;
+use codex_terminal_detection::TerminalInfo;
+use codex_terminal_detection::TerminalName;
+use codex_terminal_detection::terminal_info;
 use osc9::Osc9Backend;
 
 #[derive(Debug)]
@@ -18,13 +20,13 @@ impl DesktopNotificationBackend {
     pub fn for_method(method: NotificationMethod) -> Self {
         match method {
             NotificationMethod::Auto => {
-                if supports_osc9() {
-                    Self::Osc9(Osc9Backend)
+                if supports_osc9(&terminal_info()) {
+                    Self::Osc9(Osc9Backend::new())
                 } else {
                     Self::Bel(BelBackend)
                 }
             }
-            NotificationMethod::Osc9 => Self::Osc9(Osc9Backend),
+            NotificationMethod::Osc9 => Self::Osc9(Osc9Backend::new()),
             NotificationMethod::Bel => Self::Bel(BelBackend),
         }
     }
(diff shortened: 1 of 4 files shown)


codex-rs/tui/src/notifications/mod.rs
```diff
@@ -48,67 +50,33 @@ pub fn detect_backend(method: NotificationMethod) -> DesktopNotificationBackend
...
-    // iTerm still provides a strong session signal even when TERM_PROGRAM is missing.
-    if env::var_os("ITERM_SESSION_ID").is_some() {
-        return true;
-    }
-    // TERM-based hints cover kitty/wezterm setups without TERM_PROGRAM.
+fn supports_osc9(terminal: &TerminalInfo) -> bool {
     matches!(
-        env::var("TERM").ok().as_deref(),
-        Some("xterm-kitty" | "wezterm" | "wezterm-mux")
+        terminal.name,
```
fcoury-oai, line 55: This drops the previous `TERM=wezterm` / `TERM=wezterm-mux` fallback. Today `codex_terminal_detection` does not classify those `TERM` values as `TerminalName::WezTerm`, so `notification_method = "auto"` now falls back to BEL in setups where OSC 9 used to be selected.

Could we either preserve the `terminal.term` fallback here, or preferably teach `codex-terminal-detection` to classify `TERM=wezterm` and `TERM=wezterm-mux` as `TerminalName::WezTerm` with tests?

fcoury-oai (changes requested):
Code overall looks good. The only thing is we might have a regression for WezTerm where I pointed it out. Can you add a test to check and fix it if found to be the case?

fcoury-oai:
I also tested it with Ghostty and tmux and it worked even without the `set -g allow-passthrough on` and `focus-events on`. I just had to set focus elsewhere :-)

codex-rs/tui/src/notifications/mod.rs
caseychow-oai: Done — I took the preferred route and taught codex-terminal-detection to classify both TERM=wezterm and TERM=wezterm-mux as TerminalName::WezTerm, with tests in terminal_tests.rs.

I also validated the behavior manually:
- confirmed notifications still work in WezTerm
- forced TERM=wezterm-mux TERM_PROGRAM=WezTerm in a live tmux Codex session and confirmed the notification fired there too

So notification_method = "auto" should keep selecting OSC 9 for those WezTerm setups without needing the fallback in notifications/mod.rs.

caseychow-oai: Fixed

codex-rs/tui/src/app.rs
```diff
@@ -8365,6 +8328,7 @@ mod tests {
         let (mut app, _app_event_rx, _op_rx) = make_test_app_with_channels().await;
         let codex_home = tempdir()?;
         app.config.codex_home = codex_home.path().to_path_buf().abs();
+        app.config.sqlite_home = codex_home.path().to_path_buf();
```
caseychow-oai: straggler to fix tests

caseychow-oai: Pulled this out into the downstack test-only PR instead: #18479. PR #17836 now bases on that branch, so this `app.rs` hunk no longer belongs in the notification diff.

fcoury-oai:
Question: did you changes introduce the error you are fixing on the base PR?

The code looks good and I was able to replicate the behavior. I don't think you should submit this PR stacked on the other test fix. I would like to approve this PR with the test error and handle the other one separately. Would that be possible?

caseychow-oai:
> Question: did you changes introduce the error you are fixing on the base PR?
> 
> The code looks good and I was able to replicate the behavior. I don't think you should submit this PR stacked on the other test fix. I would like to approve this PR with the test error and handle the other one separately. Would that be possible?

I'll restack and we can see!

caseychow-oai:
> Question: did you changes introduce the error you are fixing on the base PR?
> 
> The code looks good and I was able to replicate the behavior. I don't think you should submit this PR stacked on the other test fix. I would like to approve this PR with the test error and handle the other one separately. Would that be possible?

@fcoury-oai looks like this is all green now.

fcoury-oai (approved):
Great work, can't wait for this to land on an alpha version.

Code looks good, tested and still works.

PR #17846, opened 2026-04-14
Fix for Guardian CI Tests stack overflow, applying Box to reduce stack pressure

won-openai:
### **Issue**
guardian_parallel_reviews_fork_from_last_committed_trunk_history was failing on Windows/Bazel with a stack overflow:

`thread 'guardian::tests::guardian_parallel_reviews_fork_from_last_committed_trunk_history' has overflowed its stack`

- This problem was a stack-headroom problem

### **Solution**

Reduced stack pressure in the guardian async path by boxing thin wrapper futures, and run the affected test on a dedicated 2 MiB thread stack.

Concretely:
- added Box::pin(...) around thin async wrapper hops in the guardian review/delegate path
- changed guardian_parallel_reviews_fork_from_last_committed_trunk_history to run inside an explicitly sized thread stack so it has enough headroom in low-stack environments

### DIFF
--- a/codex-rs/core/src/guardian/tests.rs
+++ b/codex-rs/core/src/guardian/tests.rs
@@ -1332,8 +1332,19 @@ async fn guardian_review_surfaces_responses_api_errors_in_rejection_reason() -> 
     Ok(())
 }
 
-#[tokio::test(flavor = "current_thread")]
-async fn guardian_parallel_reviews_fork_from_last_committed_trunk_history() -> anyhow::Result<()> {
+#[test]
+fn guardian_parallel_reviews_fork_from_last_committed_trunk_history() -> anyhow::Result<()> {
+    const TEST_STACK_SIZE_BYTES: usize = 2 * 1024 * 1024;
+
+    let handle =
+        std::thread::Builder::new()
+            .name("guardian_parallel_reviews_fork_from_last_committed_trunk_history".to_string())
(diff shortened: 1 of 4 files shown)


codex-rs/core/src/guardian/tests.rs
```diff
@@ -1332,8 +1317,19 @@ async fn guardian_review_surfaces_responses_api_errors_in_rejection_reason() ->
...
+
+    let handle =
+        std::thread::Builder::new()
+            .name("guardian_parallel_reviews_fork_from_last_committed_trunk_history".to_string())
+            .stack_size(TEST_STACK_SIZE_BYTES)
+            .spawn(|| -> anyhow::Result<()> {
+                let runtime = tokio::runtime::Builder::new_current_thread()
+                    .enable_all()
+                    .build()?;
+                runtime.block_on(Box::pin(async {
```
bolinfest, line 1347: Can we get rid of this now that you have boxed things elsewhere?

won-openai: tried removing and ci regressed, bazel local-build fails without the harness due to the same issue

bolinfest:
Let's land as-is to get the build green, but this is a bad code smell to me, so unless you have good evidence that it is _not_ a bad code smell, then please eliminate the stack carveout post-merge.

bolinfest (approved):

PR #17849, opened 2026-04-14
Make deferred dynamic tools searchable

pash-openai:
Deferred dynamic tools were being registered as callable handlers, but `tool_search` only indexed deferred MCP tools. That meant a session could include something like `automation_update` with `deferLoading: true`, the model could search for the exact tool name, and the search result would still come back empty.

This teaches `tool_search` about deferred dynamic tools alongside deferred MCP tools. The registry now advertises `tool_search` when deferred dynamic tools are present, the handler indexes dynamic tool names, descriptions, and schema fields, and search returns the matching dynamic tool as a deferred function. The ordinary turn tool list still stays small; the tool is only exposed after search.

I also updated the tool-search wording and app-server docs so this path is described as deferred tool discovery rather than MCP-only discovery.

### DIFF
--- a/codex-rs/tools/src/tool_registry_plan.rs
+++ b/codex-rs/tools/src/tool_registry_plan.rs
@@ -11,6 +11,7 @@ use crate::ToolHandlerKind;
 use crate::ToolRegistryPlan;
 use crate::ToolRegistryPlanParams;
 use crate::ToolSearchSource;
+use crate::ToolSearchSourceInfo;
 use crate::ToolSpec;
 use crate::ToolsConfig;
 use crate::ViewImageToolOptions;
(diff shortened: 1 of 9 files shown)


codex-rs/tools/src/tool_registry_plan.rs
```diff
@@ -251,26 +252,46 @@ pub fn build_tool_registry_plan(
...
 
+    let deferred_dynamic_tools = params
+        .dynamic_tools
+        .iter()
+        .filter(|tool| tool.defer_loading)
+        .collect::<Vec<_>>();
+
     if config.search_tool
-        && let Some(deferred_mcp_tools) = params.deferred_mcp_tools
+        && (params.deferred_mcp_tools.is_some() || !deferred_dynamic_tools.is_empty())
```
pakrym-oai, line 262: you're killing me. I request consistency :D

pakrym-oai: is_some || is_some!

pakrym-oai (approved):

PR #17853, opened 2026-04-14
[mcp] Add dummy tools for previously called but currently missing tools.

mzeng-openai:
- [x] Add dummy tools for previously called but currently missing tools. Currently supporting MCP tools only.


codex-rs/core/src/mcp_tool_exposure.rs
```diff
@@ -34,16 +45,70 @@ pub(crate) fn build_mcp_tool_exposure(
...
         filter_codex_apps_mcp_tools(all_mcp_tools, explicitly_enabled_connectors, config);
     McpToolExposure {
-        direct_tools,
+        direct_tools: has_mcp_servers.then_some(direct_tools),
         deferred_tools: Some(deferred_tools),
+        unavailable_called_tools: Vec::new(),
+    }
+}
+
+pub(crate) fn collect_unavailable_called_mcp_tools(
```
pakrym-oai: should this even be mcp specific?

mzeng-openai: updated to support generic tools, but filtering only MCP tools.

codex-rs/core/src/tools/spec_tests.rs
```diff
@@ -914,6 +935,59 @@ fn search_tool_registers_namespaced_mcp_tool_aliases() {
...
+        features: &features,
+        image_generation_tool_auth_allowed: true,
+        web_search_mode: Some(WebSearchMode::Cached),
+        session_source: SessionSource::Cli,
+        sandbox_policy: &SandboxPolicy::DangerFullAccess,
+        windows_sandbox_level: WindowsSandboxLevel::Disabled,
+    });
+
+    let unavailable_tool = UnavailableTool {
+        qualified_name: "mcp__codex_apps__calendar_create_event".to_string(),
```
pakrym-oai: I think we should use ToolName here and elsewhere.

codex-rs/core/src/unavailable_tool.rs
```diff
@@ -0,0 +1,121 @@
+use std::collections::BTreeMap;
+use std::collections::HashSet;
+
+use codex_protocol::models::ResponseItem;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct UnavailableTool {
```
pakrym-oai: Can this entire struct be replaced by ToolName?

codex-rs/core/src/unavailable_tool.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    unavailable_tools.into_values().collect()
+}
+
+fn should_collect_unavailable_tool(name: &str, namespace: Option<&str>) -> bool {
+    namespace.is_some_and(|namespace| namespace.starts_with("mcp__")) || name.starts_with("mcp__")
+}
+
+fn qualified_tool_name(name: &str, namespace: Option<&str>) -> String {
```
pakrym-oai: let's not add yet another place where we format names, + this is incorrect because MCPs now use namespaced tool names

codex-rs/core/tests/suite/tools.rs
```diff
@@ -91,6 +100,99 @@ async fn custom_tool_unknown_returns_custom_output_error() -> Result<()> {
     Ok(())
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn historical_unavailable_mcp_call_is_exposed_as_placeholder_tool() -> Result<()> {
+    skip_if_no_network!(Ok(()));
+
+    let historical_call_id = "historical-mcp-call";
```
pakrym-oai, line 124: can we do a more complete test? run a turn with MCP, resume, run a turn without MCP.

codex-rs/core/src/tools/spec.rs
```diff
@@ -272,6 +288,42 @@ pub(crate) fn build_specs_with_discoverable_tools(
...
+                name: unavailable_tool.qualified_name.clone(),
+                description: unavailable_tool_message(
+                    &unavailable_tool.qualified_name,
+                    "Calling this placeholder returns an error explaining that the tool is unavailable.",
+                ),
+                strict: false,
+                parameters: JsonSchema::object(
+                    Default::default(),
+                    /*required*/ None,
+                    Some(AdditionalProperties::Boolean(true)),
```
pakrym-oai: ho do we reconsile this with additionalProperties: false must always be set in objects https://developers.openai.com/api/docs/guides/structured-outputs

pakrym-oai: Is it because we are not strict?

codex-rs/core/src/tools/spec.rs
```diff
@@ -272,6 +288,42 @@ pub(crate) fn build_specs_with_discoverable_tools(
...
+            } else {
+                spec
+            };
+            builder.push_spec(spec);
+        }
+        builder.register_handler(
+            ToolName::plain(unavailable_tool.qualified_name.clone()),
+            unavailable_tool_handler.clone(),
+        );
+        if let Some(namespace) = unavailable_tool.namespace {
```
pakrym-oai: we know for sure which mode the tool should be registered in we don't need the fallback.

pakrym-oai (commented):
Good direction, few details

codex-rs/core/src/tools/spec_tests.rs
mzeng-openai: Done. The unavailable-tool plumbing and tests now use codex_tools::ToolName directly instead of constructing a separate placeholder shape.

codex-rs/core/src/unavailable_tool.rs
mzeng-openai: Done. I removed the custom UnavailableTool struct; collect_unavailable_called_tools now returns Vec<ToolName>.

codex-rs/core/src/unavailable_tool.rs
mzeng-openai: Done. I removed the extra formatter and preserve the split ToolName for namespaced calls, using ToolName::display() only when we need the function spec name.

codex-rs/core/tests/suite/tools.rs
mzeng-openai: Done. The test now runs a real first turn with a configured stdio MCP server, resumes the same rollout without that MCP config, retries the historical namespaced MCP call, and verifies the placeholder tool/handler path.

codex-rs/core/src/tools/spec.rs
mzeng-openai: Good catch. I set the placeholder parameter object to additionalProperties: false; it still uses strict: false because the placeholder accepts no meaningful args and only returns the unavailable-tool error.

codex-rs/core/src/tools/spec.rs
mzeng-openai: Done. The placeholder handler is now registered only for the exact ToolName we collected, so there is no plain/namespaced fallback registration.

pakrym-oai (approved):

PR #17854, opened 2026-04-14
Update ToolSearch to be enabled by default

mzeng-openai:
## Summary
- Promote `Feature::ToolSearch` to `Stable` and enable it in the default feature set
- Update feature tests and tool registry coverage to match the new default
- Adjust the search-tool integration test to assert the default-on path and explicit disable fallback

## Testing
- `just fmt`
- `cargo test -p codex-features`
- `cargo test -p codex-core --test all search_tool`
- `cargo test -p codex-tools`


codex-rs/core/tests/suite/search_tool.rs
```diff
@@ -110,22 +110,26 @@ fn configure_apps_without_tool_search(config: &mut Config, apps_base_url: &str)
     config.model_catalog = Some(model_catalog);
 }
 
-fn configure_apps(config: &mut Config, apps_base_url: &str) {
-    configure_apps_without_tool_search(config, apps_base_url);
+fn configure_apps_without_tool_search(config: &mut Config, apps_base_url: &str) {
```
pakrym-oai, line 115: does this method make sense?

pakrym-oai: I guess it does while we still keep the Feature wiring.

pakrym-oai (approved):

sayan-oai (approved):

PR #17862, opened 2026-04-15
Stream apply_patch changes

akshaynathan:
Adds new events for streaming apply_patch changes from responses api. This is to enable clients to show progress during file writes.

Caveat: This does not work with apply_patch in function call mode, since that required adding streaming json parsing.

### DIFF
--- a/codex-rs/apply-patch/src/parser.rs
+++ b/codex-rs/apply-patch/src/parser.rs
@@ -372,15 +372,7 @@ fn parse_one_hunk(
             );
             let (chunk, chunk_lines) = match parsed_chunk {
                 Ok(parsed) => parsed,
-                Err(err)
-                    if allow_incomplete
-                        && !chunks.is_empty()
-                        && matches!(
-                            &err,
-                            InvalidHunkError { message, .. }
-                                if message == "Update hunk does not contain any lines"
-                        ) =>
-                {
+                Err(InvalidHunkError { .. }) if allow_incomplete && !chunks.is_empty() => {
                     break;
                 }
                 Err(err) => return Err(err),
--- a/codex-rs/apply-patch/src/parser.rs
+++ b/codex-rs/apply-patch/src/parser.rs
@@ -196,26 +196,9 @@ fn parse_patch_text(patch: &str, mode: ParseMode) -> Result<ApplyPatchArgs, Pars
     let mut hunks: Vec<Hunk> = Vec::new();
     let mut remaining_lines = hunk_lines;
     let mut line_number = 2;
+    let allow_incomplete = matches!(mode, ParseMode::Streaming);
     while !remaining_lines.is_empty() {
-        let parsed_hunk = parse_one_hunk(
-            remaining_lines,
-            line_number,
-            match mode {
-                ParseMode::Streaming => IncompleteFinalChunk::Allow,
-                ParseMode::Strict | ParseMode::Lenient => IncompleteFinalChunk::Deny,
-            },
-        );
-        let (hunk, hunk_lines) = match parsed_hunk {
-            Ok(parsed) => parsed,
-            // Return accumulated hunks even if parsing the last one failed in streaming mode.
-            Err(err) if matches!(mode, ParseMode::Streaming) => {
-                if hunks.is_empty() {
-                    return Err(err);
-                }
-                break;
-            }
-            Err(err) => return Err(err),
-        };
+        let (hunk, hunk_lines) = parse_one_hunk(remaining_lines, line_number, allow_incomplete)?;
         hunks.push(hunk);
         line_number += hunk_lines;
         remaining_lines = &remaining_lines[hunk_lines..]
@@ -312,18 +295,12 @@ fn check_start_and_end_lines_strict(
     }
 }
 
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
-enum IncompleteFinalChunk {
-    Allow,
-    Deny,
-}
-
 /// Attempts to parse a single hunk from the start of lines.
 /// Returns the parsed hunk and the number of lines parsed (or a ParseError).
 fn parse_one_hunk(
     lines: &[&str],
     line_number: usize,
-    incomplete_final_chunk: IncompleteFinalChunk,
+    allow_incomplete: bool,
 ) -> Result<(Hunk, usize), ParseError> {
     // Be tolerant of case mismatches and extra padding around marker strings.
     let first_line = lines[0].trim();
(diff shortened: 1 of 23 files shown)


akshaynathan:

akshaynathan:
recheck

pakrym-oai:
Needs an integration test. Search `suite` or `test_codex`.

codex-rs/apply-patch/src/parser.rs
```diff
@@ -180,16 +194,37 @@ fn parse_patch_text(patch: &str, mode: ParseMode) -> Result<ApplyPatchArgs, Pars
                 return Err(e);
             }
             ParseMode::Lenient => check_patch_boundaries_lenient(&lines, e)?,
+            ParseMode::Streaming => check_patch_boundaries_streaming(&lines, e)?,
```
pakrym-oai: can we return the "actual patch scope" from this match expression? or individual check_ methods? that way we won't need the `matches!(mode, ParseMode::Streaming` below again.

codex-rs/apply-patch/src/parser.rs
```diff
@@ -180,16 +194,37 @@ fn parse_patch_text(patch: &str, mode: ParseMode) -> Result<ApplyPatchArgs, Pars
...
+        lines.len().saturating_sub(1)
+    };
+    let mut remaining_lines = &lines[1..body_end_index];
     let mut line_number = 2;
     while !remaining_lines.is_empty() {
-        let (hunk, hunk_lines) = parse_one_hunk(remaining_lines, line_number)?;
+        let parsed_hunk = parse_one_hunk(remaining_lines, line_number);
+        let (hunk, hunk_lines) = match parsed_hunk {
+            Ok(parsed) => parsed,
+            Err(err) if matches!(mode, ParseMode::Streaming) => {
```
pakrym-oai:
```suggestion
            // return accumulated hunks even if parsing the last one failed in streaming mode
            Err(err) if matches!(mode, ParseMode::Streaming) => {
```

codex-rs/apply-patch/src/parser.rs
```diff
@@ -453,6 +498,76 @@ fn parse_update_file_chunk(
     Ok((chunk, parsed_lines + start_index))
 }
 
+#[test]
+fn test_parse_patch_streaming() {
+    assert_eq!(
```
pakrym-oai, line 505: can you please add a stress test that takes in one large patch that has types of operations, slices it in 1 character increments and feeds to the streaming parser. We should see the ever increasing hunk count. (feel free to add stricter assertions if you'd like)

codex-rs/core/src/tools/handlers/apply_patch.rs
```diff
@@ -16,26 +20,141 @@ use crate::tools::events::ToolEventCtx;
...
+            return None;
+        }
+        if self.last_progress.as_ref() == Some(&hunks) {
+            return None;
+        }
+
+        let active_path = hunks.last().map(Hunk::path).map(Path::to_path_buf);
+        let changes = convert_apply_patch_hunks_to_protocol(&hunks);
+        self.last_progress = Some(hunks);
+        Some(PatchApplyDeltaEvent {
```
pakrym-oai: this is not a delta event right? is the latest state. 

PatchApplyUpdatedEvent?

codex-rs/core/src/codex.rs
```diff
@@ -7959,6 +7965,19 @@ async fn try_run_sampling_request(
         match event {
             ResponseEvent::Created => {}
             ResponseEvent::OutputItemDone(item) => {
+                if active_tool_argument_diff_consumer
```
pakrym-oai: I think we can clean this out unconditionally. you don't support parallel diff consumers anyway.

codex-rs/core/src/file_watcher_tests.rs
```diff
@@ -37,9 +37,13 @@ async fn throttled_receiver_coalesces_within_interval() {
     );
 
     tx.add_changed_paths(&[path("b"), path("c")]).await;
+    // Force a fresh throttle window so slow CI scheduling after the first
+    // receive cannot consume it before the blocked receive starts.
+    throttled.next_allowed = Some(Instant::now() + Duration::from_secs(1));
```
pakrym-oai: revert

codex-rs/protocol/src/protocol.rs
```diff
@@ -1536,6 +1536,9 @@ pub enum EventMsg {
     /// `ExecCommandBegin` so front‑ends can show progress indicators.
     PatchApplyBegin(PatchApplyBeginEvent),
 
+    /// Incremental model-generated structured changes for an `apply_patch` call.
+    PatchApplyDelta(PatchApplyDeltaEvent),
```
pakrym-oai: you also need to expose this event through app-server for App to be able to consume

codex-rs/protocol/src/protocol.rs
```diff
@@ -3214,6 +3217,16 @@ pub struct PatchApplyBeginEvent {
...
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize, JsonSchema, TS)]
+pub struct PatchApplyDeltaEvent {
+    /// Identifier for the originating `apply_patch` tool call.
+    pub call_id: String,
+    /// Structured file changes parsed from the model-generated patch input so far.
+    pub changes: HashMap<PathBuf, FileChange>,
+    /// File path currently being written, when known.
+    pub active_path: Option<PathBuf>,
```
pakrym-oai: can this be calculated by the client based on changes[-1]?

pakrym-oai (approved):
Very clean!! Nice work!

Let's add a core integration test and expose through app-server API. And maybe rename the event because it's not truly a delta

codex-rs/apply-patch/src/parser.rs
akshaynathan: done

codex-rs/apply-patch/src/parser.rs
akshaynathan: done

codex-rs/core/src/tools/handlers/apply_patch.rs
akshaynathan: done

codex-rs/core/src/codex.rs
akshaynathan: done

codex-rs/core/src/file_watcher_tests.rs
akshaynathan: removed

codex-rs/protocol/src/protocol.rs
akshaynathan: coming in the next PR

codex-rs/protocol/src/protocol.rs
akshaynathan: removed, will calculate on the client!

PR #17873, opened 2026-04-15
Add runtime model provider abstraction

pakrym-oai:
## Summary
- Add a `ModelProvider` runtime trait with a private generic implementation and factory constructor.
- Store `Arc<dyn ModelProvider>` in turn and model-client state instead of raw provider metadata.
- Route provider metadata access through `provider.info()` while keeping existing `ModelProviderInfo` helpers intact.
- Update affected tests and subagent config construction for the new provider wrapper.

## Testing
- `just fmt`
- `cargo check -p codex-core`
- `cargo test -p codex-core --lib build_agent_`
- `just fix -p codex-core`


celia-oai:
closing this in favor of #17713.

PR #17877, opened 2026-04-15
Display YOLO mode permissions if set when launching TUI

canvrno-oai:
- When launching the TUI client, if YOLO mode is enabled, display this in the header.
- Eligibility is determined by `approval_policy = "never"` and `sandbox_mode = "danger-full-access"`

image


etraut-openai (approved):
Implementation looks good!

A few thoughts. I don't feel strongly about either,  so feel free to ignore.
1. I wonder if we should use lowercase for "yolo". It doesn't appear as uppercase anywhere else in our docs or code, and most people talk about it in lowercase.
2. I wonder if we should add "/permissions to change" at the end of the line, consistent with the "/model to change" two lines above it?

canvrno-oai:
> Implementation looks good!
> 
> A few thoughts. I don't feel strongly about either, so feel free to ignore.
> 
> 1. I wonder if we should use lowercase for "yolo". It doesn't appear as uppercase anywhere else in our docs or code, and most people talk about it in lowercase.
> 2. I wonder if we should add "/permissions to change" at the end of the line, consistent with the "/model to change" two lines above it?

Both are good points- I'll do a followup for these but would like to get this in for now to satisfy the request and prevent more CI retry cycles.

PR #17883, opened 2026-04-15
Remove exec-server fs sandbox request preflight

pakrym-oai:
## Summary
- Remove the exec-server-side manual filesystem request path preflight before invoking the sandbox helper.
- Keep sandbox helper policy construction and platform sandbox enforcement as the access boundary.
- Add a portable local+remote regression for writing through an explicitly configured alias root.
- Remove the metadata symlink-escape assertion that depended on the deleted manual preflight; no replacement metadata-specific access probe is added.

## Tests
- `cargo test -p codex-exec-server --lib`
- `cargo test -p codex-exec-server --test file_system`
- `git diff --check`

### DIFF
--- a/codex-rs/exec-server/tests/file_system.rs
+++ b/codex-rs/exec-server/tests/file_system.rs
@@ -139,18 +139,6 @@ fn assert_normalized_path_rejected(error: &std::io::Error) {
     }
 }
 
-fn sandbox_temp_dir() -> Result<TempDir> {
-    let slash_tmp = Path::new("/tmp");
-    if cfg!(target_os = "linux") && slash_tmp.is_dir() {
-        tempfile::Builder::new()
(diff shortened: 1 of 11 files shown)


codex-rs/exec-server/tests/file_system.rs
```diff
@@ -528,35 +561,6 @@ async fn file_system_create_directory_rejects_symlink_escape(use_remote: bool) -
     Ok(())
 }
 
-#[test_case(false ; "local")]
-#[test_case(true ; "remote")]
-#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-async fn file_system_get_metadata_rejects_symlink_escape(use_remote: bool) -> Result<()> {
```
pakrym-oai, line 534: aligns with shell

starr-openai (approved):

PR #17885, opened 2026-04-15
feat: Support alternate marketplace manifests and local string

xl-openai:
- Discover marketplace manifests from different supported layout paths instead of only .agents/plugins/marketplace.json.
- Accept local plugin sources written either as { source: "local", path: ... } or as a direct string path.
- Skip unsupported or invalid plugin source entries without failing the entire marketplace, and keep valid local plugins loadable.


xli-oai (approved):

xl-openai:
@codex review

PR #17892, opened 2026-04-15
Migrate archive/unarchive to local ThreadStore

wiltzius-openai:
# Summary
- implement local ThreadStore archive/unarchive operations
- implement local ThreadStore read_thread operation
- break up the various ThreadStore local method implementations into separate files
- migrate app-server archive/unarchive and core archive fixture to use ThreadStore (but not all read operations yet!)
- use the ThreadStore's read operation as a proxy check for thread persistence/existence in the app server code
- move all other filesystem operations related to archive (path validation etc) into the local thread store. 

# Tests
- add dedicated local store archive/unarchive tests

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -244,7 +244,6 @@ use codex_core::plugins::load_plugin_apps;
 use codex_core::plugins::load_plugin_mcp_servers;
 use codex_core::read_head_for_summary;
 use codex_core::read_session_meta_line;
-use codex_core::rollout_date_parts;
 use codex_core::sandboxing::SandboxPermissions;
 use codex_core::windows_sandbox::WindowsSandboxLevelExt;
 use codex_core::windows_sandbox::WindowsSandboxSetupMode as CoreWindowsSandboxSetupMode;
@@ -319,8 +318,10 @@ use codex_state::StateRuntime;
 use codex_state::ThreadMetadata;
 use codex_state::ThreadMetadataBuilder;
 use codex_state::log_db::LogDbLayer;
+use codex_thread_store::ArchiveThreadParams as StoreArchiveThreadParams;
 use codex_thread_store::ListThreadsParams as StoreListThreadsParams;
 use codex_thread_store::LocalThreadStore;
+use codex_thread_store::ReadThreadParams as StoreReadThreadParams;
 use codex_thread_store::StoredThread;
 use codex_thread_store::ThreadSortKey as StoreThreadSortKey;
 use codex_thread_store::ThreadStore;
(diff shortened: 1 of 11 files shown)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2729,32 +2724,23 @@ impl CodexMessageProcessor {
...
-                        data: None,
-                    };
-                    self.outgoing.send_error(request_id, error).await;
-                    return;
-                }
-            };
-
         let thread_id_str = thread_id.to_string();
-        match self.archive_thread_common(thread_id, &rollout_path).await {
+        let store = LocalThreadStore::new(codex_rollout::RolloutConfig::from_view(&self.config));
```
pakrym-oai: let's stash the store on self so we don't have to re-create it every time.

codex-rs/thread-store/src/local/mod.rs
```diff
@@ -348,6 +573,223 @@ mod tests {
         Ok(path)
     }
 
+    #[tokio::test]
+    async fn archive_thread_moves_rollout_to_archived_collection() {
```
pakrym-oai: It's very-very good to have tests for these but let's move them out into their own file  (or files, per op?) otherwise this mod will get huge.

pakrym-oai (approved):

codex-rs/thread-store/src/local/mod.rs
```diff
@@ -35,11 +47,81 @@ pub struct LocalThreadStore {
...
+            thread_id,
+            canonical_rollout_path,
+            file_name,
+        })
+    }
+}
+
+impl PreparedLocalThreadArchive {
+    /// Move the prepared rollout into the archived collection.
+    pub async fn complete(self) -> ThreadStoreResult<()> {
```
pakrym-oai: this isn't abstractable.

codex-rs/thread-store/src/local/mod.rs
```diff
@@ -35,11 +47,81 @@ pub struct LocalThreadStore {
...
+}
+
 impl LocalThreadStore {
     /// Create a local store from the rollout configuration used by existing local persistence.
     pub fn new(config: RolloutConfig) -> Self {
         Self { config }
     }
+
+    /// Prepare to archive a local thread by resolving and validating its active rollout.
+    pub async fn prepare_archive_thread(
```
pakrym-oai: it's a bit strange to have the two stage archive. Can we shutdown the thread first and have a single step archive call?

codex-rs/app-server/src/codex_message_processor.rs
wiltzius-openai: done

codex-rs/thread-store/src/local/mod.rs
wiltzius-openai: split into multiple files, one per ThreadStore method plus some helpers

codex-rs/thread-store/src/local/mod.rs
wiltzius-openai: now replaced by the reworked thread lookup followed by app-server-internal prepare followed by ThreadStore archive call

codex-rs/thread-store/src/local/mod.rs
wiltzius-openai: now replaced by the reworked thread lookup followed by app-server-internal prepare followed by ThreadStore archive call

pakrym-oai (approved):

PR #17896, opened 2026-04-15
Clarify realtime v2 context and handoff messages

bxie-openai:
## Summary
- wrap realtime startup context in `<startup_context>...</startup_context>` tags
- prefix V2 mirrored user text and relayed backend text with `[USER]` / `[BACKEND]`
- remove the V2 progress suffix and replace the final V2 handoff output with a short completion acknowledgement while preserving the existing V1 wrapper

## Testing
- cargo test -p codex-api realtime_v2_session_update_includes_background_agent_tool_and_handoff_output_item -- --exact
- cargo test -p codex-app-server webrtc_v2_background_agent_
- cargo test -p codex-app-server webrtc_v2_text_input_is_
- cargo test -p codex-core conversation_user_text_turn_is_


codex-rs/core/src/realtime_conversation.rs
```diff
@@ -1074,18 +1091,17 @@ async fn handle_handoff_output(
...
+                writer.send_conversation_item_create(output_text).await
             }
             HandoffOutput::FinalUpdate {
                 handoff_id,
-                output_text,
+                output_text: _,
             } => {
                 if let Err(err) = writer
-                    .send_conversation_handoff_append(handoff_id, output_text)
+                    .send_conversation_handoff_append(
```
aibrahim-oai, line 1101: I think we will miss the final agent message here because we aren't using output text

bxie-openai: the backend output text already shows up in earlier [BACKEND] messages. this is intentional to dedup. with the current behavior, the model may say the same final message twice. but with the change, it will behave correctly.

aibrahim-oai (approved):

PR #17931, opened 2026-04-15
fix: support encrypted local secrets for keyring auth

### LINKED ISSUE #10353: OAuth tokens too long for Windows keyring

0.93.0

Plus

_No response_

Windows 11

Windows Terminal

The primary login mechanism via OAuth doesn't work on Windows when using the keyring storage option.

```
Unable to persist auth file: failed to write OAuth tokens to keyring: Attribute 'password encoded as UTF-16' is longer than platform limit of 2560 chars
```

Image

Log in with the following configuration:

```toml
cli_auth_credentials_store = "keyring"
mcp_oauth_credentials_store = "keyring"
```

_No response_

### Additional information

_No response_

mikhail-oai:
## Why

Windows Credential Manager limits a generic credential blob to 2,560 bytes. Large ChatGPT auth payloads, and MCP OAuth token payloads in the same keyring-backed storage path, can exceed that limit and make keyring persistence fail on Windows.

This PR keeps keyring-backed auth secure without requiring those large serialized payloads to fit directly in the OS credential entry. Instead, Windows can store the payload in Codex's encrypted local secrets backend while keeping the local secrets key in the OS keyring. Non-Windows platforms continue to use direct keyring storage by default. Goal is to move everything to the secrets keyring path and deprecate the old path after it's fully validated.

## What Changed

- Added `AuthKeyringBackendKind` and a `secret_auth_storage` feature that selects between direct keyring storage and encrypted local-secrets-backed keyring storage. The feature is enabled by default on Windows.
- Updated CLI login, logout, auth loading, auth refresh, device-code login, login status, and `codex doctor` paths to use the configured keyring backend.
- Reworked keyring-mode CLI auth storage so the secrets backend stores serialized `AuthDotJson`, removes stale fallback `auth.json` data after successful saves, and deletes both encrypted and fallback auth on logout.
- Extended MCP OAuth persistence to use the same keyring backend selection for token load/save/delete, including encrypted local secrets storage and the existing file fallback behavior when keyring access fails.
- Threaded the backend selection through MCP auth status checks, MCP client startup, app-server processors, cloud task/config adapters, and related test fixtures.

## Validation

Manual macOS/Windows validation with `cli_auth_credentials_store = "auto"`, `mcp_oauth_credentials_store = "auto"`, and `features.secret_auth_storage = true`:

- Ran CLI login and confirmed Keychain Access created the shared secrets key entry with service `codex` and account `secrets|...`.
- Ran MCP OAuth login for configured MCP servers and confirmed the same Keychain entry is used for the local secrets encryption key rather than a per-MCP plaintext token entry.
- Listed decrypted secret names from `$CODEX_HOME/secrets/local.age` without printing secret values, confirming one CLI auth entry and MCP OAuth entries are present:
  - `global/CLI_AUTH`
  - `global/MCP_OAUTH_...`

Also validated that the feature flag is default on for windows and default off for macOS.

#closes #10353


celia-oai:
@codex review

codex-rs/login/src/auth/storage.rs
```diff
@@ -247,16 +255,89 @@ impl AuthStorageBackend for KeyringAuthStorage {
...
+impl AuthStorageBackend for SecretsKeyringAuthStorage {
+    fn load(&self) -> std::io::Result<Option<AuthDotJson>> {
+        match self
+            .secrets_backend
+            .get(&SecretScope::Global, &CLI_AUTH_SECRET_NAME)
+            .map_err(|err| {
+                std::io::Error::other(format!(
+                    "failed to load CLI auth from encrypted auth storage: {err}"
+                ))
+            })? {
```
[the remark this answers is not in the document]
celia-oai: i think this is fine as a one-time migration cost

codex-rs/rmcp-client/src/oauth.rs
```diff
@@ -134,7 +145,23 @@ fn load_oauth_tokens_from_keyring_with_fallback_to_file<K: KeyringStore>(
...
+    server_name: &str,
+    url: &str,
+) -> Result<Option<StoredOAuthTokens>> {
+    match keyring_backend_kind {
+        AuthKeyringBackendKind::Direct => {
+            load_oauth_tokens_from_direct_keyring(keyring_store, server_name, url)
+        }
+        AuthKeyringBackendKind::Secrets => {
+            load_oauth_tokens_from_secrets_keyring(keyring_store, server_name, url)
+        }
```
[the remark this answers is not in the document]
celia-oai: same as above

celia-oai:
@codex review

codex-rs/core/src/config/mod.rs
```diff
@@ -1557,6 +1569,34 @@ pub async fn load_config_as_toml_with_cli_and_load_options(
...
+pub fn auth_keyring_backend_kind_from_config_toml(
+    config_toml: &ConfigToml,
+) -> AuthKeyringBackendKind {
+    let features = Features::from_sources(
+        FeatureConfigSource {
+            features: config_toml.features.as_ref(),
+            experimental_use_unified_exec_tool: config_toml.experimental_use_unified_exec_tool,
+        },
+        FeatureConfigSource::default(),
+        FeatureOverrides::default(),
```
viyatb-oai, line 1688: **[P1] Apply managed feature requirements before bootstrap auth selection**

This resolver reconstructs `Features` from `ConfigToml` only. The bootstrap load immediately above has already loaded local/system `requirements.toml`, but its `feature_requirements` are discarded when this helper receives only the merged TOML. If an admin pins `secret_auth_storage` opposite the user config or platform default, TUI/exec open auth from the wrong backend; the cloud-requirements fetch then sees no auth and returns `Ok(None)`, silently skipping the remaining workspace constraints.

Please retain the bootstrap `ConfigLayerStack` (or its requirements) and resolve through `ManagedFeatures` before creating the `AuthManager`. Suggested patch shape:

```rust
let configured = Features::from_sources(...);
let managed = ManagedFeatures::from_configured(
    configured,
    config_layer_stack.requirements().feature_requirements.clone(),
)?;
auth_keyring_backend_kind_from_secret_auth_storage(
    managed.enabled(Feature::SecretAuthStorage),
)
```

codex-rs/rmcp-client/src/oauth.rs
```diff
@@ -249,6 +350,48 @@ fn delete_oauth_tokens_from_keyring_and_file<K: KeyringStore>(
...
+    server_name: &str,
+    url: &str,
+) -> Result<bool> {
+    match keyring_backend_kind {
+        AuthKeyringBackendKind::Direct => {
+            delete_oauth_tokens_from_direct_keyring(keyring_store, server_name, url)
+        }
+        AuthKeyringBackendKind::Secrets => {
+            delete_oauth_tokens_from_secrets_keyring(keyring_store, server_name, url)
+        }
```
viyatb-oai, line 413: **[P1] Delete the legacy direct MCP entry when logging out**

The secrets branch removes the `local.age` entry but leaves the previous direct-keyring entry intact. After logout, rollback or disabling `secret_auth_storage` makes the stale OAuth token active again. Remove both representations while this compatibility boundary exists.

```suggestion
        AuthKeyringBackendKind::Secrets => {
            let direct_removed =
                delete_oauth_tokens_from_direct_keyring(keyring_store, server_name, url)?;
            let secrets_removed =
                delete_oauth_tokens_from_secrets_keyring(keyring_store, server_name, url)?;
            Ok(direct_removed || secrets_removed)
        }
```

codex-rs/rmcp-client/src/oauth.rs
```diff
@@ -197,12 +268,34 @@ fn save_oauth_tokens_with_keyring<K: KeyringStore>(
...
+    server_name: &str,
+    tokens: &StoredOAuthTokens,
+) -> Result<()> {
+    let serialized = serde_json::to_string(tokens).context("failed to serialize OAuth tokens")?;
+    let codex_home = find_codex_home()?;
+    let backend =
+        LocalSecretsBackend::new(codex_home.to_path_buf(), Arc::new(keyring_store.clone()));
+    let secret_name = compute_secret_name(server_name, &tokens.url)?;
+    backend
+        .set(&SecretScope::Global, &secret_name, &serialized)
```
viyatb-oai, line 326: **[P1] Serialize updates to the shared encrypted secrets file**

This adds a runtime MCP-token writer to the shared `local.age` file, and the CLI auth path now writes the same file. `LocalSecretsBackend::set` currently does an unlocked `load_file()` -> mutate -> `save_file()` transaction. Atomic rename prevents a torn file, but two MCP refreshes or two Codex processes can both read the same generation and the last writer drops the other secret.

Please put coordination inside `LocalSecretsBackend` so every caller is covered: acquire an exclusive lockfile before `load_file()` and hold it through `save_file()` for both `set` and `delete`. This needs a cross-process file lock; an instance-local mutex would not protect parallel Codex processes.

codex-rs/login/src/auth/storage.rs
```diff
@@ -247,16 +255,89 @@ impl AuthStorageBackend for KeyringAuthStorage {
     }
 }
 
+#[derive(Clone, Debug)]
+struct SecretsKeyringAuthStorage {
+    codex_home: PathBuf,
+    secrets_backend: LocalSecretsBackend,
```
viyatb-oai: **[P1] Delete the legacy direct CLI entry when logging out**

`SecretsKeyringAuthStorage::delete` clears the encrypted entry and `auth.json`, but it leaves the previous direct-keyring entry intact. After logout, rollback or disabling `secret_auth_storage` resurrects the stale CLI login.

Please retain the direct backend on this wrapper and delete both stores. Suggested patch shape:

```rust
struct SecretsKeyringAuthStorage {
    codex_home: PathBuf,
    direct_storage: DirectKeyringAuthStorage,
    secrets_backend: LocalSecretsBackend,
}

// In delete():
let direct_removed = self.direct_storage.delete()?;
Ok(secrets_removed || direct_removed)
```

Construct `direct_storage` with `Arc::clone(&keyring_store)` before moving the store into `LocalSecretsBackend`.

codex-rs/rmcp-client/src/oauth.rs
```diff
@@ -152,26 +179,70 @@ fn load_oauth_tokens_from_keyring<K: KeyringStore>(
     }
 }
 
+fn load_oauth_tokens_from_secrets_keyring<K: KeyringStore + Clone + 'static>(
+    keyring_store: &K,
+    server_name: &str,
+    url: &str,
+) -> Result<Option<StoredOAuthTokens>> {
+    let codex_home = find_codex_home()?;
```
fcoury-oai, line 223: Here we could pass `codex_home` or at least an initialized secrets store to avoid rediscovering it from process state?

`McpConnectionManager::new` already receives `codex_home`, we could forward it to `AsyncManagedClient` and into `RmcpClient`.

This would avoid making credential storage depend on a global environment lookup and forces tests to mutate and serialize around CODEX_HOME.

codex-rs/login/src/auth/storage.rs
```diff
@@ -247,16 +255,89 @@ impl AuthStorageBackend for KeyringAuthStorage {
...
+}
+
+impl SecretsKeyringAuthStorage {
+    fn new(codex_home: PathBuf, keyring_store: Arc<dyn KeyringStore>) -> Self {
+        let secrets_backend = LocalSecretsBackend::new(codex_home.clone(), keyring_store);
```
fcoury-oai: Did we mean to bypass `SecretsManager` here? MCP OAuth also constructs `LocalSecretsBackend` directly. 

If auth is intentionally pinned to local encrypted storage, could we document that invariant? Otherwise, using `SecretsManager::new_with_keyring_store` would keep callers from depending on the concrete backend.

fcoury-oai (approved):
I was able to test on Windows and it worked as expected. The code and architecture in general looks sane, but I have a few comments:

Non blocking architectural nit: `store_mode` and `keyring_backend_kind` are being passed together across many layers. Would be nice to bundle them into a `AuthStoragePolicy` value object to minimize plumbing and avoid inconsistencies by combining them together as the code evolves.

Another thing I had to do in Windows to exercise this PR was change the package version to something other than `0.0.0`. Local source builds intentionally resolve `Auto` and `Keyring` back to file storage.

Should we add a release-like Windows CI case that verifies encrypted CLI auth save/reload, MCP OAuth save/reload, and the explicit feature override?

I also left 2 inline comments for your consideration.

The unresolved `viyatb-oai` comments look like the remaining blockers, particularly the shared `local.age` update race and logout cleanup behavior. Once those are addressed, I am comfortable approving this from my side. The notes above and my inline comments are non-blocking follow-ups.

celia-oai:
@fcoury-oai thanks for the review!

> store_mode and keyring_backend_kind are being passed together across many layers. Would be nice to bundle them into a AuthStoragePolicy value object to minimize plumbing and avoid inconsistencies by combining them together as the code evolves.

i'll prob not do this, because `keyring_backend_kind` is a temp feature flag that I aim to remove in 1-2 releases after we roll out to all users, but `store_mode` will stay as a valid choice.

> Another thing I had to do in Windows to exercise this PR was change the package version to something other than 0.0.0. Local source builds intentionally resolve Auto and Keyring back to file storage.

yes, this is currently hardcoded unfortunately. I can get a follow up to allow config overwrite locally. reason for it is that local builds prompt for user passwords without signing so it slows down the dev process

> Should we add a release-like Windows CI case that verifies encrypted CLI auth save/reload, MCP OAuth save/reload, and the explicit feature override?

this would be good to have, but we need to think through how to do this correctly carefully, as we don't have a smoke test pattern covering this auth path we could just reuse. I'm happy to do this as a follow-up if needed.

codex-rs/rmcp-client/src/oauth.rs
celia-oai: seems that we currently already do this for default file storage path, so it's not really a regression: https://github.com/openai/codex/blob/main/codex-rs/rmcp-client/src/oauth.rs#L537. happy to do it as a follow-up though

codex-rs/login/src/auth/storage.rs
celia-oai: ah good call! fixing

codex-rs/rmcp-client/src/oauth.rs
celia-oai: I'll focus on keeping the contract the same as the current `auth.json` and `.credentials.json`: there're separate file for cli auth and mcp auth, but no locking behavior. this way we ensure there's no regression in concurrent write/read contract. happy to get out a follow up with more locking if needed.

celia-oai:
@codex review

codex-rs/tui/src/lib.rs
```diff
@@ -1025,6 +1028,7 @@ pub async fn run_main(
         bootstrap_config_toml
             .cli_auth_credentials_store
             .unwrap_or_default(),
+        resolve_bootstrap_auth_keyring_backend_kind(&bootstrap_config)?,
```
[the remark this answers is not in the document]
celia-oai: I think we don't have to support this. this will be a temp feature flag removed in 2-3 releases, and we prob won't need to flip it much.

celia-oai:
this has been merged with the PR stack ending in #27541.

PR #17946, opened 2026-04-15
Fix empty tool descriptions

shijie-oai:
## Summary
- Ensure direct namespaced MCP tool groups are emitted with a non-empty namespace description even when namespace metadata is missing or blank.
- Add regression coverage for missing MCP namespace descriptions.

## Cause
Latest `main` can serialize a direct namespaced MCP tool group with an empty top-level `description`. The namespace description path used `unwrap_or_default()` when `tool_namespaces` did not include metadata for that namespace, so the outbound Responses API payload could contain a tool like `{"type":"namespace","description":""}`. The Responses API rejects that because namespace tool descriptions must be a non-empty string.

## Fix
- Add a fallback namespace description: `Tools in the <namespace> namespace.`
- Preserve provided namespace descriptions after trimming, but treat blank descriptions as missing.

### Issue I am seeing
This is what I am seeing on the local build.
Screenshot 2026-04-15 at 10 55 55 AM

### DIFF
--- a/codex-rs/tools/src/responses_api.rs
+++ b/codex-rs/tools/src/responses_api.rs
@@ -91,13 +91,28 @@ pub fn mcp_tool_to_deferred_responses_api_tool(
 }
 
 pub fn tool_definition_to_responses_api_tool(tool_definition: ToolDefinition) -> ResponsesApiTool {
+    let ToolDefinition {
+        name,
+        description,
+        input_schema,
+        output_schema,
+        defer_loading,
+    } = tool_definition;
+
+    let description = description.trim();
+    let description = if description.is_empty() {
+        format!("Call the {name} tool.")
+    } else {
+        description.to_string()
+    };
+
     ResponsesApiTool {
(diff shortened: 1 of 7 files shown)


codex-rs/tools/src/responses_api.rs
```diff
@@ -91,13 +91,28 @@ pub fn mcp_tool_to_deferred_responses_api_tool(
...
+        name,
+        description,
+        input_schema,
+        output_schema,
+        defer_loading,
+    } = tool_definition;
+
+    let description = description.trim();
+    let description = if description.is_empty() {
+        format!("Call the {name} tool.")
```
shijie-oai: I am not really sure what we should do if there is a missing description but it does not like empty string here for sure. Not opposed just to put the name of the tool in here instead but I think the current approach is fine.

pakrym-oai (approved):

sayan-oai (approved):

PR #17952, opened 2026-04-15
fix: preserve plugin cache symlinks

conrad-oai:


dylan-hurd-oai (approved):
cc @xl-openai for review as well but this feels reasonable to me

xl-openai (approved):

PR #17958, opened 2026-04-15
Support remote compaction for Azure responses providers

### LINKED ISSUE #17773: Azure provider silently uses local compaction instead of `POST /responses/compact`
## Problem

When using an Azure OpenAI provider (`name = "Azure"` or an Azure base URL), Codex silently falls back to local model-generated compaction instead of calling `POST /responses/compact`, even though [Azure's Responses API fully supports it](https://learn.microsoft.com/en-us/azure/foundry/openai/how-to/responses?tabs=python-key#compact-a-response). No error or warning is shown.

## Root Cause

The compaction routing gate in `core/src/compact.rs` checks only `is_openai()`:

```rust
pub(crate) fn should_use_remote_compact_task(provider: &ModelProviderInfo) -> bool {
    provider.is_openai()   // true only when name == "OpenAI" (exact, case-sensitive)
}
```

The standard Azure configuration requires `name = "Azure"` to activate `attach_item_ids` via `is_azure_responses_endpoint()` (case-insensitive check in `codex-api/src/provider.rs`). But that same name makes `is_openai()` return `false`, silently disabling remote compaction. The two checks are mutually exclusive — no single `name` value satisfies both.

| `name` | `is_openai()` | `is_azure_responses_endpoint()` | Result |
|---|---|---|---|
| `"OpenAI"` | ✓ | ✗ | Remote compaction on, `attach_item_ids` off (breaks Azure) |
| `"Azure"` | ✗ | ✓ | Remote compaction off, `attach_item_ids` on |

ivanmurashko:
Azure Responses providers were still falling back to local compaction because the compaction gate only checked `ModelProviderInfo::is_openai()`.

Move the capability check onto `ModelProviderInfo` with `supports_remote_compaction()`, backed by the existing Azure Responses endpoint detection used in `codex-api`, and have `core::compact` delegate to that helper.

Add regression coverage for:
- OpenAI providers using remote compaction
- Azure providers using remote compaction
- non-OpenAI/non-Azure providers staying on the local path

resolves #17773

### DIFF
--- a/codex-rs/model-provider-info/src/model_provider_info_tests.rs
+++ b/codex-rs/model-provider-info/src/model_provider_info_tests.rs
@@ -131,7 +131,7 @@ supports_websockets = true
 
 #[test]
 fn test_supports_remote_compaction_for_openai() {
-    let provider = ModelProviderInfo::create_openai_provider(/* base_url */ None);
+    let provider = ModelProviderInfo::create_openai_provider(/*base_url*/ None);
 
     assert!(provider.supports_remote_compaction());
 }
--- a/codex-rs/model-provider-info/src/model_provider_info_tests.rs
+++ b/codex-rs/model-provider-info/src/model_provider_info_tests.rs
@@ -130,6 +130,61 @@ supports_websockets = true
 }
 
 #[test]
+fn test_supports_remote_compaction_for_openai() {
+    let provider = ModelProviderInfo::create_openai_provider(/* base_url */ None);
+
+    assert!(provider.supports_remote_compaction());
+}
+
+#[test]
+fn test_supports_remote_compaction_for_azure_name() {
+    let provider = ModelProviderInfo {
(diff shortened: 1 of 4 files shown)


codex-rs/model-provider-info/src/model_provider_info_tests.rs
```diff
@@ -129,6 +129,61 @@ supports_websockets = true
     assert_eq!(provider.websocket_connect_timeout_ms, Some(15_000));
 }
 
+#[test]
+fn test_supports_remote_compaction_for_openai() {
+    let provider = ModelProviderInfo::create_openai_provider(/* base_url */ None);
```
bolinfest: The linter I created requires this format:

```suggestion
    let provider = ModelProviderInfo::create_openai_provider(/*base_url*/ None);
```

PR #17965, opened 2026-04-15
fix: rename is_azure_responses_wire_base_url to is_azure_responses_provider

bolinfest:
## Why

While reviewing https://github.com/openai/codex/pull/17958, the helper name `is_azure_responses_wire_base_url` looked misleading because the helper returns true for either the `azure` provider name or an Azure Responses `base_url`. The new name makes both inputs part of the contract.

## What

- Rename `is_azure_responses_wire_base_url` to `is_azure_responses_provider`.
- Move the `openai.azure.` marker into `matches_azure_responses_base_url` so all base URL marker matching is centralized.
- Keep `Provider::is_azure_responses_endpoint()` behavior unchanged.

## Verification

- Compared the parent and current implementations. `name.eq_ignore_ascii_case("azure")` still returns true before consulting `base_url`, `None` still returns false, base URLs are still lowercased before marker matching, and the same Azure marker set is checked.
- Ran `cargo test -p codex-api`.

### DIFF
--- a/codex-rs/codex-api/src/provider.rs
+++ b/codex-rs/codex-api/src/provider.rs
@@ -86,7 +86,7 @@ impl Provider {
     }
 
     pub fn is_azure_responses_endpoint(&self) -> bool {
-        is_azure_responses_wire_base_url(&self.name, Some(&self.base_url))
+        is_azure_responses_provider(&self.name, Some(&self.base_url))
     }
 
     pub fn websocket_url_for_path(&self, path: &str) -> Result<Url, url::ParseError> {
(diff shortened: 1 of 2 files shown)


codex-rs/codex-api/src/provider.rs
pakrym-oai: can you also unify base.contains("openai.azure and matches_azure_responses_base_url

pakrym-oai (approved):

codex-rs/codex-api/src/provider.rs
bolinfest: good call: fixed!

PR #17968, opened 2026-04-15
Add a ConfigStore interface

rasmusrygaard:
Add a ConfigStore interface for reading config files. By default we'll back this with a local-only implementation, which I'll add in a follow-up

### DIFF
--- a/codex-rs/config-store/src/store.rs
+++ b/codex-rs/config-store/src/store.rs
@@ -1,0 +1,20 @@
+use async_trait::async_trait;
+
+use crate::ConfigDocumentRead;
+use crate::ConfigStoreResult;
+use crate::ReadConfigDocumentParams;
+
+/// Storage-neutral reader for path-addressed config documents.
+///
+/// Implementations should only read and parse the requested document. Codex config loading remains
+/// responsible for deciding what the document represents, how missing documents are handled, how
+/// parse errors interact with project trust, how relative paths are resolved, and how layers are
+/// ordered and merged.
+#[async_trait]
+pub trait ConfigDocumentStore: Send + Sync {
+    /// Reads one config document addressed by path.
+    async fn read_config_document(
+        &self,
+        params: ReadConfigDocumentParams,
+    ) -> ConfigStoreResult<ConfigDocumentRead>;
+}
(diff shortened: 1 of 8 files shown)


codex-rs/config-store/src/store.rs
```diff
@@ -0,0 +1,20 @@
+use async_trait::async_trait;
+
+use crate::ConfigDocumentRead;
+use crate::ConfigStoreResult;
+use crate::ReadConfigDocumentParams;
+
+/// Storage-neutral reader for path-addressed config documents.
```
wiltzius-openai, line 7: are all implementations going to be path-addressed? are we going to invent some special path namespacing to make them generic (rather than fs paths)?

codex-rs/config-store/src/types.rs
```diff
@@ -0,0 +1,119 @@
...
+/// Request to read one path-addressed config document.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ReadConfigDocumentParams {
+    /// Absolute path to the config document to read.
+    pub path: AbsolutePathBuf,
+}
+
+/// Byte span for a config document parse error.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ConfigDocumentErrorSpan {
```
wiltzius-openai, line 16: interesting that we need structured byte span ranges for errors, I'd have guessed that could be represented simply in an error string

wiltzius-openai (commented):
How are we going to handle typing of the config documents? Will there be specific config file loaders that have schemas knowing what fields they expect etc, and we'll hand them a ConfigStore to do the actual byte fetching?

I had originally imagined this interface would have more structured config knowledge in it rather than a generic loader, but if those exist elsewhere and this is only meant to abstract away whether its loaded from filesystem or somewhere else it seems reasonable.

rasmusrygaard:
The code parses these values into `ConfigToml`. I guess we could also parse into that type instead of the raw document? It would just require updating more code that reads these values to use the parsed type when merging, right now we can just merge toml values together

wiltzius-openai:
Huh -- I had assumed we want to decide value precedence on more of a case by case basis rather than using a general toml merge, but I haven't examined how this works today. Fine w/ me if this is simpler and works.

rasmusrygaard:
the other thing that's helpful about the toml merges is that presence can be 1/ not set 2/ set to none/default 3/ set to some value. We mostly have `Option` types in `ConfigToml` which means we can only track 1 and 3 without having `Option<Option<...>>`

PR #17971, opened 2026-04-15
fix: deprecate use_legacy_landlock feature flag

viyatb-oai:
## Summary
- mark `features.use_legacy_landlock` as a deprecated feature flag
- emit a startup deprecation notice when the flag is configured
- add feature- and core-level regression coverage for the notice

Screenshot 2026-04-15 at 11 14 00 PM


etraut-openai (approved):
The tests seem a bit excessive for a deprecation warning, but they will presumably get stripped once the feature is removed entirely. The remainder looks good.

PR #17975, opened 2026-04-15
Fix flaky app list tests

mzeng-openai:
## Summary
- Abort unfinished app-list refresh tasks on error or timeout so a failed request cannot leave background work mutating caches later.
- Make the disabled-connectors test explicitly write `connectors = false` into the temp config.
- Strengthen the force-refetch cache test to exercise the race window that was causing intermittent failures.

## Testing
- `just fmt`
- `cargo test -p codex-app-server suite::v2::app_list`
- stress-tested `suite::v2::app_list::list_apps_force_refetch_preserves_previous_cache_on_failure` 30x
- `env CODEX_HOME=/tmp/codex-app-server-test-home cargo test -p codex-app-server`

### DIFF
--- a/codex-rs/app-server/tests/suite/v2/app_list.rs
+++ b/codex-rs/app-server/tests/suite/v2/app_list.rs
@@ -926,9 +926,12 @@ async fn list_apps_force_refetch_preserves_previous_cache_on_failure() -> Result
     assert!(initial_next_cursor.is_none());
     assert_eq!(initial_data.len(), 1);
     assert!(initial_data.iter().all(|app| app.is_accessible));
+    timeout(
+        DEFAULT_TIMEOUT,
+        server_control.wait_for_tools_list_finished(),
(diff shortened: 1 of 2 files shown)


codex-rs/app-server/tests/suite/v2/app_list.rs
```diff
@@ -939,6 +951,8 @@ async fn list_apps_force_refetch_preserves_previous_cache_on_failure() -> Result
     .await??;
     assert!(refetch_error.error.message.contains("failed to"));
 
+    tokio::time::sleep(Duration::from_millis(250)).await;
```
pakrym-oai: this doesn't sounds deterministic.

mzeng-openai: it's not, did some rewrite to make it much cleaner.

PR #17978, opened 2026-04-15
Persist and prewarm agent tasks per thread

adrian-openai:
## Summary
- persist registered agent tasks in the session state update stream so the thread can reuse them
- prewarm task registration once identity registration succeeds, while keeping startup failures best-effort
- isolate the session-side task lifecycle into a dedicated module so AgentIdentityManager and RegisteredAgentTask do not leak across as many core layers

## Testing
- cargo test -p codex-core startup_agent_task_prewarm
- cargo test -p codex-core cached_agent_task_for_current_identity_clears_stale_task
- cargo test -p codex-core record_initial_history_

### DIFF
--- a/codex-rs/app-server/tests/suite/v2/initialize.rs
+++ b/codex-rs/app-server/tests/suite/v2/initialize.rs
@@ -24,7 +24,8 @@ use std::time::Duration;
 use tempfile::TempDir;
 use tokio::time::timeout;
 
-const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(20);
+// These tests start full app-server processes; keep headroom for concurrent debug startup.
+const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(30);
 
 #[tokio::test]
 async fn initialize_uses_client_info_name_as_originator() -> Result<()> {
--- a/codex-rs/app-server/tests/suite/v2/initialize.rs
+++ b/codex-rs/app-server/tests/suite/v2/initialize.rs
@@ -24,8 +24,9 @@ use std::time::Duration;
 use tempfile::TempDir;
 use tokio::time::timeout;
 
-// These tests start full app-server processes; keep headroom for concurrent debug startup.
-const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(30);
+// This covers debug app-server process startup before the first JSON-RPC response,
+// not expected steady-state request latency.
+const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(20);
 
 #[tokio::test]
 async fn initialize_uses_client_info_name_as_originator() -> Result<()> {
--- a/codex-rs/app-server/tests/suite/v2/initialize.rs
+++ b/codex-rs/app-server/tests/suite/v2/initialize.rs
@@ -24,7 +24,7 @@ use std::time::Duration;
 use tempfile::TempDir;
 use tokio::time::timeout;
 
-const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(20);
 
 #[tokio::test]
 async fn initialize_uses_client_info_name_as_originator() -> Result<()> {
--- a/codex-rs/core/src/agent_identity.rs
+++ b/codex-rs/core/src/agent_identity.rs
@@ -153,14 +153,16 @@ impl AgentIdentityManager {
         Ok(stored_identity)
     }
 
-    pub(crate) async fn task_matches_current_binding(&self, task: &RegisteredAgentTask) -> bool {
+    pub(crate) async fn task_matches_current_identity(&self, task: &RegisteredAgentTask) -> bool {
         if !self.feature_enabled {
             return false;
         }
 
-        self.current_auth_binding()
+        self.current_stored_identity()
             .await
-            .is_some_and(|(_, binding)| task.matches_binding(&binding))
+            .is_some_and(|stored_identity| {
+                stored_identity.agent_runtime_id == task.agent_runtime_id
+            })
     }
 
     async fn current_auth_binding(&self) -> Option<(CodexAuth, AgentIdentityBinding)> {
@@ -175,6 +177,11 @@ impl AgentIdentityManager {
             debug!("skipping agent identity flow because ChatGPT auth is unavailable");
         }
         binding.map(|binding| (auth, binding))
+    }
+
+    async fn current_stored_identity(&self) -> Option<StoredAgentIdentity> {
+        let (auth, binding) = self.current_auth_binding().await?;
+        self.load_stored_identity(&auth, &binding).ok().flatten()
     }
 
     async fn register_agent_identity(
--- a/codex-rs/core/src/session/agent_task_lifecycle.rs
+++ b/codex-rs/core/src/session/agent_task_lifecycle.rs
@@ -1,0 +1,182 @@
+use crate::agent_identity::RegisteredAgentTask;
+use crate::session::session::Session;
+use codex_protocol::protocol::RolloutItem;
+use codex_protocol::protocol::SessionAgentTask;
+use codex_protocol::protocol::SessionStateUpdate;
+use tracing::debug;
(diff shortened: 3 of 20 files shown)


codex-rs/app-server/tests/suite/v2/client_metadata.rs
```diff
@@ -18,7 +18,7 @@ use std::path::Path;
 use tempfile::TempDir;
 use tokio::time::timeout;
 
-const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(20);
```
efrazer-oai: why do we need to change all the test timings for this? was it flaky?

efrazer-oai: why do we need to change all the test timings for this? was it flaky?

adrian-openai: I believe this was a codex artifact. Rolling this back after making sure that things aren't flaky.

adrian-openai: Same as above. Nice catch!

codex-rs/protocol/src/protocol.rs
```diff
@@ -2713,6 +2713,21 @@ impl fmt::Display for SubAgentSource {
     }
 }
 
+/// Persisted agent-task details that let a resumed thread keep using the same backend task.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema, TS)]
+pub struct SessionAgentTask {
```
efrazer-oai, line 2775: Do we want to allow the session to have a different agent runtime id than the globally registered one? 

Seems like it could potentially result in weird state/cause issues (like when building auth headers, which one do we use?)

Or maybe this is this tracking the runtime id that was created on thread start, which we might want a comment to clarify.

adrian-openai:
Stack navigation for this slice:

- PR1: #17385 - Add use_agent_identity feature flag
- PR2: #17386 - Register agent identities behind use_agent_identity
- PR3: #17387 - Register agent tasks behind use_agent_identity
- PR3.1: #17978 - Persist and prewarm agent tasks per thread (this PR)
- PR4: #17980 - Use AgentAssertion downstream behind use_agent_identity
- PR4.1: #18094 - Use background agent task auth for backend calls

codex-rs/app-server/tests/suite/v2/initialize.rs
```diff
@@ -24,7 +24,8 @@ use std::time::Duration;
 use tempfile::TempDir;
 use tokio::time::timeout;
 
-const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+// These tests start full app-server processes; keep headroom for concurrent debug startup.
+const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(30);
```
shijie-oai: Are we doing this only for debug? Would like to understand better what/if the code change here is causing longer run loop or something.

nicksteele-oai:
@codex review

codex-rs/app-server/tests/suite/v2/initialize.rs
adrian-openai: Good catch! This has been updated.

codex-rs/core/src/agent_identity.rs
```diff
@@ -153,14 +153,16 @@ impl AgentIdentityManager {
         Ok(stored_identity)
     }
 
-    pub(crate) async fn task_matches_current_binding(&self, task: &RegisteredAgentTask) -> bool {
+    pub(crate) async fn task_matches_current_identity(&self, task: &RegisteredAgentTask) -> bool {
```
shijie-oai, line 156: This is going to be a catch 22 situation - probably will be answered in a later PR and I will come back to comment on this. What happens if there is a mismatch on the agent runtime id between the session vs the task? Agent identity has ttl that can expire mid session and be reacquired?

codex-rs/core/src/session/agent_task_lifecycle.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    fn latest_persisted_agent_task(
+        rollout_items: &[RolloutItem],
+    ) -> Option<Option<SessionAgentTask>> {
+        rollout_items.iter().rev().find_map(|item| match item {
+            RolloutItem::SessionState(update) => Some(update.agent_task.clone()),
+            _ => None,
+        })
+    }
```
shijie-oai, line 29: I am not 100% sure if we want to actually persist agent task as a part of the rollout? Any downside for this to be in memory and limited the usage to the current session and on future resume we generate a new one or conceptually it is incorrect?

codex-rs/core/src/session/rollout_reconstruction.rs
```diff
@@ -207,7 +207,9 @@ impl Session {
                         active_segment.get_or_insert_with(ActiveReplaySegment::default);
                     active_segment.counts_as_user_turn |= is_user_turn_boundary(response_item);
                 }
-                RolloutItem::EventMsg(_) | RolloutItem::SessionMeta(_) => {}
+                RolloutItem::EventMsg(_)
+                | RolloutItem::SessionMeta(_)
+                | RolloutItem::SessionState(_) => {}
```
shijie-oai, line 212: Ties to if we want to store sessionState at all.

codex-rs/core/src/session/session.rs
```diff
@@ -828,6 +829,7 @@ impl Session {
 
         // record_initial_history can emit events. We record only after the SessionConfiguredEvent is emitted.
         sess.record_initial_history(initial_history).await;
+        sess.start_agent_identity_registration();
```
shijie-oai, line 832: DO we want to await on this?

shijie-oai (commented):
Main question is that do we need task agent stored within the rollout? I do not see the downside (and to be fair in most situation) to recreate the task identity when we resume (i.e. huge window span).

efrazer-oai:
> Main question is that do we need task agent stored within the rollout? I do not see the downside (and to be fair in most situation) we would need to recreate the task identity when we resume (i.e. huge window span).

I think I pushed for this originally, but I think you'd have better intuition here @shijie-oai. Core idea was to keep the task lifecycle very close to a session lifecycle (e.g. we can have multiple tasks per rollout now), but i'm g either way.

codex-rs/core/src/agent_identity.rs
shijie-oai: Okay I am back - my concern still stands and will leave a PR level comment.

shijie-oai (commented):
I am not seeing a clear failure handling if an agent or task validation is outdated when checking against backend. 

If you are agent identity only (for enterprise use case), it is okay to hard fail as we need to strictly enforce defined TTL. But if you are chatgpt auth and we assign agent identity on the fly, I wonder what is the default TTL for that agent identity is and how we should best recover. Maybe that agent identity has the same life cycle as chatgpt refresh token?

codex-rs/core/src/agent_identity.rs
adrian-openai: Discussed directly - generally this should not happen in the current implementation.

codex-rs/core/src/session/agent_task_lifecycle.rs
adrian-openai: Discussed in person - we want the task to be persisted so that we can keep permissions the same across the entirety of the rollout, and not start from scratch every so often.

codex-rs/core/src/session/rollout_reconstruction.rs
adrian-openai: Still think we do :)

codex-rs/core/src/session/session.rs
adrian-openai: [codex] We intentionally do not await this at startup. `record_initial_history(...)` needs to run first so resumed `SessionState` can restore/clear the persisted task before prewarm can cache a new one; after that, startup registration is best-effort prewarm. The actual user-turn path awaits `ensure_agent_task_registered()` in `session/turn.rs`, emits an error for that thread if registration fails, and retries on the next turn rather than blocking session startup or shutting down other threads.

shijie-oai (approved):

PR #17980, opened 2026-04-15
[codex] Use AgentAssertion downstream behind use_agent_identity

adrian-openai:
## Summary

This is the AgentAssertion downstream slice for feature-gated agent identity support, replacing the oversized AgentAssertion slice from PR #17807.

It isolates task-scoped downstream AgentAssertion wiring on top of the merged PR3.1 work without re-carrying the earlier agent registration, task registration, or task-state history.

This PR includes the task-scoped bug-fix call sites from the review: generic file upload auth, MCP OpenAI file upload auth, and ARC monitor auth. Broader user/control-plane calls move to PR4.1 and PR4.2.

## Stack

- PR4: this PR - use task-scoped `AgentAssertion` downstream when enabled
- PR4.1: https://github.com/openai/codex/pull/18094 - introduce AuthManager-owned background/control-plane `AgentAssertion` auth
- PR4.2: https://github.com/openai/codex/pull/18260 - use background task auth for additional backend/control-plane calls

## What Changed

- add AgentAssertion envelope generation in `codex-core`
- route downstream HTTP and websocket auth through AgentAssertion when an agent task is present
- extend the model-provider auth provider so non-bearer authorization schemes can be passed through cleanly
- make generic file uploads attach the full authorization header value
- make MCP OpenAI file uploads use the cached thread agent task assertion when present
- make ARC monitor calls use the cached thread agent task assertion when present

## Why

The original PR had drifted ancestry and showed a much larger diff than the semantic change actually required. Restacking it onto PR3.1 keeps the reviewable surface down to the downstream assertion slice.

## Validation

- `just fmt`
- `cargo check -p codex-core -p codex-login -p codex-analytics -p codex-app-server -p codex-cloud-requirements -p codex-cloud-tasks -p codex-models-manager -p codex-chatgpt -p codex-model-provider -p codex-mcp -p codex-core-skills`
- `cargo test -p codex-model-provider bearer_auth_provider`
- `cargo test -p codex-core agent_assertion`
- `cargo test -p codex-app-server remote_control`
- `cargo test -p codex-cloud-requirements fetch_cloud_requirements`
- `cargo test -p codex-models-manager manager::tests`
- `cargo test -p codex-chatgpt`
- `cargo test -p codex-cloud-tasks`
- `cargo test -p codex-login agent_identity`
- `just fix -p codex-core -p codex-login -p codex-analytics -p codex-app-server -p codex-cloud-requirements -p codex-cloud-tasks -p codex-models-manager -p codex-chatgpt -p codex-model-provider -p codex-mcp -p codex-core-skills`
- `git diff --check`


efrazer-oai (commented):
I had Codex go find all the places that are still hardcoded to use `Bearer` as discussed on Slack, here are the results:



As it stands, we have successfully migrated the shared Responses HTTP, Responses websocket, compact, and ChatGPT-backend realtime-call paths.

But there are still several other backend surfaces that build `Authorization: Bearer ...` directly or go through helpers that only understand bearer tokens. Those still need to migrate if we want this change to be exhaustive.

Shared backend client paths that still need the new auth shape:

- `codex-rs/backend-client/src/client.rs`
  - `/wham/usage`
  - `/wham/tasks/list`
  - `/wham/tasks/{id}`
  - `/wham/tasks/{id}/turns/{turn_id}/sibling_turns`
  - `/wham/config/requirements`
  - `POST /wham/tasks`

Current callers of that shared client include:

- `codex-rs/cloud-requirements/src/lib.rs`
- `codex-rs/app-server/src/codex_message_processor.rs`
- `codex-rs/cloud-tasks-client/src/http.rs`

Direct backend callers that still attach bearer auth themselves:

- `codex-rs/codex-api/src/files.rs`
  - `/files`
  - `/files/{id}/uploaded`
- `codex-rs/core/src/mcp_openai_file.rs`
- `codex-rs/models-manager/src/manager.rs`
  - `/codex/models`
- `codex-rs/codex-mcp/src/mcp/mod.rs`
  - `.../wham/apps`
- `codex-rs/chatgpt/src/chatgpt_client.rs`
  - shared GET helper used by:
    - `codex-rs/chatgpt/src/get_task.rs`
    - connector directory loading paths
- `codex-rs/core/src/connectors.rs`
  - connector directory requests
- `codex-rs/core/src/plugins/remote.rs`
  - `/plugins/list`
  - `/plugins/featured`
  - `/plugins/{id}/enable`
  - `/plugins/{id}/uninstall`
- `codex-rs/core-skills/src/remote.rs`
  - `/hazelnuts`
  - `/hazelnuts/{skill_id}/export`
- `codex-rs/analytics/src/client.rs`
  - `/codex/analytics-events/events`
- `codex-rs/core/src/arc_monitor.rs`
  - `/codex/safety/arc`
- `codex-rs/app-server/src/transport/remote_control/enroll.rs`
  - `/wham/remote/control/server/enroll`
- `codex-rs/app-server/src/transport/remote_control/websocket.rs`
  - websocket handshake for `/wham/remote/control/server`
- `codex-rs/cloud-tasks/src/util.rs`
- `codex-rs/cloud-tasks/src/env_detect.rs`
  - `/wham/environments`
  - `/wham/environments/by-repo/...`

One backend hit I found that does not look like part of this auth migration is:

- `codex-rs/core/src/plugins/startup_sync.rs`
  - backup archive fetch at `/backend-api/plugins/export/curated`

So the main point is: this PR fixes the shared model path, but it does not yet migrate all of the places where we talk to `codex-backend`.

nicksteele-oai (approved):
Looks clean to me

codex-rs/core/src/mcp_openai_file.rs
```diff
@@ -109,12 +118,20 @@ async fn build_uploaded_local_argument_value(
...
+            Some(authorization_header_value),
+            /*account_id*/ None,
+        )
+    } else {
+        let token_data = auth
+            .get_token_data()
+            .map_err(|error| format!("failed to read ChatGPT auth for file upload: {error}"))?;
+        CoreAuthProvider::from_bearer_token(Some(token_data.access_token), token_data.account_id)
     };
     let uploaded = upload_local_file(
```
nicksteele-oai, line 146: Will file upload with ChatGPT auth be affected if the above code block fails?

codex-rs/app-server/tests/suite/v2/app_list.rs
```diff
@@ -56,7 +56,7 @@ use tokio::net::TcpListener;
 use tokio::task::JoinHandle;
 use tokio::time::timeout;
 
-const DEFAULT_TIMEOUT: Duration = Duration::from_secs(10);
+const DEFAULT_TIMEOUT: Duration = Duration::from_secs(20);
```
shijie-oai: hm maybe we should drop this? I assume it has to do with windows test flake?

codex-rs/codex-api/src/files.rs
```diff
@@ -260,8 +261,8 @@ fn authorized_request(
     let mut request = client
         .request(method, url)
         .timeout(OPENAI_FILE_REQUEST_TIMEOUT);
-    if let Some(token) = auth.bearer_token() {
-        request = request.bearer_auth(token);
+    if let Some(authorization) = auth.authorization_header_value() {
+        request = request.header(AUTHORIZATION, authorization);
```
shijie-oai: This would enable us to use non bearer token (agent identity) for as the auth header vs `bearer_auth` is for bearer token only I assume.

codex-rs/app-server/tests/suite/v2/thread_unsubscribe.rs
```diff
@@ -130,9 +130,13 @@ async fn thread_unsubscribe_keeps_thread_loaded_until_idle_timeout() -> Result<(
...
-        "-Command".to_string(),
-        "Start-Sleep -Seconds 1".to_string(),
+        "cmd.exe".to_string(),
+        "/d".to_string(),
+        "/c".to_string(),
+        "ping".to_string(),
+        "-n".to_string(),
+        "2".to_string(),
+        "127.0.0.1".to_string(),
     ];
```
shijie-oai: No action: I think this is updated/different on latest main already.

codex-rs/core/src/agent_identity/assertion.rs
```diff
@@ -0,0 +1,176 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub(crate) struct AgentAssertionEnvelope {
+    pub(crate) agent_runtime_id: String,
+    pub(crate) task_id: String,
+    pub(crate) timestamp: String,
+    pub(crate) signature: String,
+}
+
+impl AgentIdentityManager {
+    pub(crate) async fn authorization_header_for_task(
```
shijie-oai: nit: better name maybe? `authorization_header_for_task` sounds like it would return the auth but it is a validation technically?

shijie-oai: actually my bad - it does return the actua header value.

codex-rs/core/src/client.rs
```diff
@@ -211,6 +215,8 @@ pub struct ModelClient {
 pub struct ModelClientSession {
     client: ModelClient,
     websocket_session: WebsocketSession,
+    agent_task: Option<RegisteredAgentTask>,
```
shijie-oai, line 218: Do we need this? agent_task is scoped to a session (thread) and here we are on the top level?

codex-rs/core/src/mcp_openai_file.rs
```diff
@@ -109,12 +118,20 @@ async fn build_uploaded_local_argument_value(
             "ChatGPT auth is required to upload local files for Codex Apps tools".to_string(),
         );
     };
-    let token_data = auth
-        .get_token_data()
-        .map_err(|error| format!("failed to read ChatGPT auth for file upload: {error}"))?;
-    let upload_auth = CoreAuthProvider {
-        token: Some(token_data.access_token),
-        account_id: token_data.account_id,
```
shijie-oai, line 117: Did not realize that we have so many locally built headers....

shijie-oai (approved):
Makes sense to me overall.

codex-rs/app-server/tests/suite/v2/app_list.rs
adrian-openai: Yes! The test is flaky locally as a result of the 10s.

codex-rs/core/src/client.rs
adrian-openai: The field is in ModelClientSession,
- The registered agent task is owned by the Codex thread/session state.
- But each model turn needs to carry that task into all Responses requests made during that turn: HTTP stream, websocket prewarm, retries, continuation requests.
- Putting it on ModelClient would be wrong because ModelClient is longer-lived and shared for the Codex session.
- Passing it into every lower-level request method would be more error-prone.
- Storing it on the per-turn ModelClientSession also lets the code avoid reusing a cached bearer-auth websocket when an AgentAssertion task is present.
Let me know if you disagree with that ^

PR #17986, opened 2026-04-15
[4/8] Abstract MCP stdio server launching

aibrahim-oai:
## Summary
- Move local MCP stdio process startup behind a launcher trait.
- Preserve existing local stdio behavior while making transport creation explicit.

## Stack
```text
o  #18027 [8/8] Fail exec client operations after disconnect
│
o  #18025 [7/8] Cover MCP stdio tests with executor placement
│
o  #17974 [6/8] Wire remote MCP stdio through executor
│
o  #17987 [5/8] Add executor process transport for MCP stdio
│
@  #17986 [4/8] Abstract MCP stdio server launching
│
o  #18020 [3/8] Add pushed exec process events
│
o  #17985 [2/8] Support piped stdin in exec process API
│
o  #17984 [1/8] Add MCP server environment config
│
o  main
```


codex-rs/rmcp-client/src/rmcp_client.rs
```diff
@@ -307,9 +304,8 @@ impl StreamableHttpClient for StreamableHttpResponseClient {
 }
 
 enum PendingTransport {
-    ChildProcess {
-        transport: TokioChildProcess,
-        process_group_guard: Option<ProcessGroupGuard>,
+    Stdio {
```
pakrym-oai, line 307: instead of implementing yet another layer of abstraction can we reuse existing RMCP Transport abstraction?

pakrym-oai (changes requested):
Can we reuse RMCP Transport abstraction?

aibrahim-oai:
Recreated as #18087 from an openai/codex head branch so internal CI uses the normal credentials path.

PR #17988, opened 2026-04-15
fix(tui): restore context usage when resuming sessions

fcoury-oai:
# Problem

Resumed TUI sessions could initially show `Context 0% used` even when the previous session had a much higher context usage value. The correct value appeared only after the user submitted another turn, because the live runtime token-count path refreshed the TUI state then. That made resume look like it had lost context accounting, even though the historical token-count events were already present in the rollout.

# Mental model

Thread history is reconstructed from the rollout stream. The visible transcript is one projection of that stream, and token usage is another projection. A resume/read response that restores turns but omits the latest token-count projection leaves the TUI with a partially restored session: history is current, but context-window state is still at the default.

The fix treats latest token usage as thread snapshot state. App-server reads the newest persisted `TokenCount` event while rebuilding turns and exposes it as `Thread.tokenUsage`. The TUI carries that value through `ThreadSessionState` and applies it during session bootstrap before any new live token-count update can arrive.

# Non-goals

This does not change the token-count calculation itself, the baseline adjustment used for context-window percentages, or the shape of live `thread/tokenUsage/updated` notifications. It also does not backfill token usage for rollouts that never persisted token-count events.

# Tradeoffs

Adding `tokenUsage` to `Thread` expands every response schema that embeds a thread. That is noisy in generated schema diffs, but it keeps resume/read/fork/rollback clients on a single snapshot object instead of requiring a second RPC or a parallel cache lookup. The field is nullable and defaults during deserialization so older persisted or test fixtures remain compatible.

The helper scans rollout items from the end to find the latest token-count event. This is simple and matches the desired semantics, but it means the response derives token usage only from events already persisted in the rollout. In-flight token usage remains the responsibility of live notifications.

# Architecture

`Thread.tokenUsage` is the app-server protocol boundary for latest known usage. When app-server builds a thread from a rollout-backed source, it now updates both `thread.turns` and `thread.token_usage` from the same ordered item list. The helper also preserves the existing active-turn merge behavior: active turns affect the rendered turns list, while token usage still comes from persisted rollout state.

On the TUI side, app-server lifecycle responses are normalized into `ThreadSessionState`. The normalized state includes the optional token usage from the thread snapshot. `ChatWidget::handle_thread_session` applies the legacy session configuration first, then applies restored token usage through `set_token_info`, which refreshes context-window status surfaces and requests a redraw.

# Observability

There are no new logs. The fastest debug path is to inspect a resumed thread/read response and check whether `thread.tokenUsage` is present. If the field is present but the TUI status line is wrong, the issue is likely in `ThreadSessionState` normalization or `ChatWidget::set_token_info`. If the field is absent, inspect the rollout for persisted `TokenCount` events and check app-server rollout reconstruction.

# Tests

Protocol serialization tests now include `tokenUsage`. App-server has coverage that the latest token-count event wins when deriving usage from rollout items. TUI session tests verify that resume responses preserve token usage and that applying a restored thread session seeds the status line before a new turn arrives.

### DIFF
--- a/codex-rs/analytics/src/analytics_client_tests.rs
+++ b/codex-rs/analytics/src/analytics_client_tests.rs
@@ -121,6 +121,7 @@ fn sample_thread_with_source(
         agent_role: None,
         git_info: None,
         name: None,
+        token_usage: None,
         turns: Vec::new(),
     }
 }
--- a/codex-rs/app-server-protocol/schema/json/ServerNotification.json
+++ b/codex-rs/app-server-protocol/schema/json/ServerNotification.json
@@ -2531,6 +2531,18 @@
           ],
           "description": "Current runtime status for the thread."
         },
+        "tokenUsage": {
+          "anyOf": [
+            {
+              "$ref": "#/definitions/ThreadTokenUsage"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "default": null,
+          "description": "Latest known token usage for the thread, when available from runtime or rollout history."
+        },
         "turns": {
           "description": "Only populated on `thread/resume`, `thread/rollback`, `thread/fork`, and `thread/read` (when `includeTurns` is true) responses. For all other responses and notifications returning a Thread, the turns field will be an empty list.",
           "items": {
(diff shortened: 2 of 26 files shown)


fcoury-oai:
@codex review

PR #17998, opened 2026-04-15
bazel: add explicit rust test shard labels

bolinfest:
## Why

The larger Rust test targets are expensive as single Bazel test actions. Native Bazel sharding helps split execution, but it still reports through the same test target label. That is not enough for the Codex workflow: we want BuildBuddy to show timing and flakiness per shard label, and we want a failed or flaky shard to be rerunnable as a concrete Bazel target instead of treating the whole aggregate test as one opaque target.

This PR therefore moves the selected large Rust tests to explicit generated shard targets. It also incorporates the rules_rust behavior from https://github.com/hermeticbuild/rules_rust/pull/14 so each test name maps to a stable shard bucket by hash, rather than by list position.

## What

- Extends `codex_rust_crate` with `test_shard_counts`.
- For configured targets, keeps the original test target name as a `test_suite` aggregate and generates one concrete test rule per shard.
- For unit tests, keeps the existing `workspace_root_test` wrapper shape around one `*-unit-tests-bin` `rust_test`.
- For integration tests, compiles one manual `*-all-test-bin` `rust_test` and makes each shard label a lightweight `test_binary_test` wrapper around that binary. This preserves distinct labels without compiling the Rust test crate once per shard.
- Sets `RULES_RUST_TEST_TOTAL_SHARDS` / `RULES_RUST_TEST_SHARD_INDEX` on each generated shard target so the rules_rust wrapper can run the correct subset without using Bazel-reserved `TEST_*` env vars.
- Adds `patches/rules_rust_stable_explicit_test_shards.patch`, which mirrors hermeticbuild/rules_rust#14 until Codex can bump to a merged `rules_rust` commit containing that support.
- Adds a Windows manifest fallback for nested test launchers so the rules_rust sharding wrapper can find the real test binary when it is run from another test rule's runfiles tree.
- Uses explicit decimal `UInt64` constants in the Windows PowerShell FNV hash expression so the 32-bit mask cannot be interpreted as `-1`.
- Uses `TEST_TMPDIR` plus a per-wrapper temp directory in the Windows sharding wrapper so parallel shards do not collide on shared `%TEMP%\rust_test_list_*.txt` files.
- Configures 8 shards for:
  - `//codex-rs/core:core-all-test`
  - `//codex-rs/core:core-unit-tests`
  - `//codex-rs/app-server:app-server-all-test`
  - `//codex-rs/app-server:app-server-unit-tests`
  - `//codex-rs/tui:tui-unit-tests`

## Example Labels

The aggregate label remains available:

```text
//codex-rs/core:core-all-test
```

but it now expands to explicit shard labels:

```text
//codex-rs/core:core-all-test-shard-1-of-8
//codex-rs/core:core-all-test-shard-2-of-8
...
//codex-rs/core:core-all-test-shard-8-of-8
```

For integration tests, those shard labels point at one compiled test binary:

```starlark
rust_test(
    name = "core-all-test-bin",
    experimental_enable_sharding = True,
    tags = ["manual", "no-sandbox"],
    ...
)

test_binary_test(
    name = "core-all-test-shard-1-of-8",
    test_bin = ":core-all-test-bin",
    env = {
        "RULES_RUST_TEST_TOTAL_SHARDS": "8",
        "RULES_RUST_TEST_SHARD_INDEX": "0",
    },
    tags = ["no-sandbox"],
)
```

That means BuildBuddy sees labels such as `//codex-rs/core:core-all-test-shard-8-of-8`, but Bazel still has only one underlying `rust_test` rule for the integration test binary: `//codex-rs/core:core-all-test-bin`.

Unit tests use the same explicit shard label pattern while still running through the workspace-root launcher:

```text
//codex-rs/core:core-unit-tests
//codex-rs/core:core-unit-tests-bin
//codex-rs/core:core-unit-tests-shard-1-of-8
//codex-rs/core:core-unit-tests-shard-8-of-8
```

The label is one-indexed for readability (`shard-1-of-8`), while the env value is the zero-indexed shard index consumed by the wrapper.

## Verification

- `just bazel-lock-check`
- `bazel query 'kind("rust_test rule", //codex-rs/core:*)'`
- `bazel query 'kind(".* rule", //codex-rs/core:core-all-test-shard-8-of-8 + //codex-rs/core:core-all-test-bin + //codex-rs/app-server:app-server-all-test-shard-8-of-8 + //codex-rs/tui:tui-unit-tests-bin + //codex-rs/tui:tui-unit-tests-shard-1-of-8)' --output=build`
- `bazel test --test_output=errors //codex-rs/core:core-all-test-shard-1-of-8 //codex-rs/core:core-all-test-shard-8-of-8 //codex-rs/tui:tui-unit-tests-shard-1-of-8`
- `bazel test --test_output=errors //codex-rs/core:core-all-test-shard-7-of-8`
- `bazel test --test_output=errors //codex-rs/app-server:app-server-unit-tests-shard-3-of-8 //codex-rs/core:core-all-test-shard-7-of-8`


starr-openai (approved):

bolinfest:
#18082 took care of this

PR #17999, opened 2026-04-15
[codex] Make command exec delta tests chunk tolerant

euroelessar:
## Summary
- Make command/exec output-delta tests accumulate streamed chunks instead of assuming complete logical output in a single notification.
- Collect stdout and stderr independently so stream interleaving does not fail the pipe streaming test.

## Why
The command/exec protocol exposes output as deltas, so tests should not rely on chunk boundaries being stable. A line like `out-start\n` may arrive split across multiple notifications, and stdout/stderr notifications may interleave.

## Validation
- `just fmt`
- `git diff --check`
- `cargo test -p codex-app-server suite::v2::command_exec`

### DIFF
--- a/codex-rs/app-server/tests/suite/v2/command_exec.rs
+++ b/codex-rs/app-server/tests/suite/v2/command_exec.rs
@@ -411,11 +411,29 @@ async fn command_exec_streaming_does_not_buffer_output() -> Result<()> {
         })
         .await?;
 
-    let delta = read_command_exec_delta(&mut mcp).await?;
-    assert_eq!(delta.process_id, process_id.as_str());
-    assert_eq!(delta.stream, CommandExecOutputStream::Stdout);
-    assert_eq!(STANDARD.decode(&delta.delta_base64)?, b"abcde");
-    assert!(delta.cap_reached);
+    let deadline = Instant::now() + DEFAULT_READ_TIMEOUT;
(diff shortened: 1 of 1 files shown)


codex-rs/app-server/tests/suite/v2/command_exec.rs
```diff
@@ -411,11 +411,29 @@ async fn command_exec_streaming_does_not_buffer_output() -> Result<()> {
...
         .await?;
 
-    let delta = read_command_exec_delta(&mut mcp).await?;
-    assert_eq!(delta.process_id, process_id.as_str());
-    assert_eq!(delta.stream, CommandExecOutputStream::Stdout);
-    assert_eq!(STANDARD.decode(&delta.delta_base64)?, b"abcde");
-    assert!(delta.cap_reached);
+    let deadline = Instant::now() + DEFAULT_READ_TIMEOUT;
+    let mut stdout = String::new();
+    loop {
```
pakrym-oai: didn't we add a mehtod below do to just this?

pakrym-oai (approved):

codex-rs/app-server/tests/suite/v2/command_exec.rs
euroelessar: they are slightly different, but let me think a bit more

PR #18003, opened 2026-04-15
[docs] Revert extra changes from PR 17848

evawong-oai:
## Summary

1. Revert https://github.com/openai/codex/pull/17848 so the Bazel and `BUILD` file changes leave `main`.
2. Prepare for a narrower follow up that restores only `SECURITY.md`.

## Validation

1. Reviewed the revert diff against `main`.
2. Ran a clean diff check before push.

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -94,7 +94,7 @@ jobs:
             "${bazel_wrapper_args[@]}" \
             -- \
             test \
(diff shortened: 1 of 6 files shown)


bolinfest (approved):

bolinfest:
Verified that Windows failures were different between Attempt #1 and Attempt #2, so appear to be flaky tests and are not caused by this change.

PR #18006, opened 2026-04-15
fix: more flake

jif-oai:
Stabilizes the Responses API proxy header test by splitting the coverage at the right boundary:
  - Core integration test now verifies parent/subagent identity headers directly from captured `/responses` requests.
  - Proxy dump unit test now verifies those identity headers are preserved in dumped request JSON.
  - Removes the flaky real proxy process + temp-file dump polling path from the core test.


jif-oai:
@codex review

PR #18009, opened 2026-04-15
[codex] route thread reads through thread store

wiltzius-openai:
High level: move app server's thread_read and get_thread_summary to use the ThreadStore interface

Specifically:
- make the implementation of thread reading in the local ThreadStore implementation more complete.
- introduce a "read_thread_view" in App Server that is responsible for merging "live" thread state with "persisted" thread state coming from the ThreadStore. Currently this merge is conceptually present in a number of places in the app server code but not clearly identified or centralized.

Caveats:
- in a couple places where the app server interface explicitly accepts a rollout path continue to operate directly on that path; focus on getting the read-by-thread-id case functional first
- does not yet attempt to migrate other call sites that read threads in app server methods that are about forking/resuming/creating threads -- these require more extensive cleanup to remove path assumptions first

Testing:
- add additional unit test coverage for the thread store local implementation of read_thread

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -3788,7 +3788,7 @@ impl CodexMessageProcessor {
         };
 
         let thread = match self
-            .read_thread_view(thread_uuid, include_turns, /*include_archived*/ false)
+            .read_thread_view(thread_uuid, include_turns, /*include_archived*/ true)
             .await
         {
             Ok(thread) => thread,
@@ -3916,15 +3916,26 @@ impl CodexMessageProcessor {
         // ThreadStore, but loaded threads that are not yet readable from the store may
         // still have local metadata. Keep this compatibility path centralized here so
         // future remote threadstore implementations have one app-server boundary to replace.
-        if let Some(summary) =
-            read_summary_from_state_db_by_thread_id(&self.config, thread_id).await
-        {
+        let loaded_state_db = loaded_thread.state_db();
+        let summary = match loaded_state_db.as_ref() {
+            Some(state_db) => {
+                read_summary_from_state_db_context_by_thread_id(Some(state_db), thread_id).await
+            }
+            None => read_summary_from_state_db_by_thread_id(&self.config, thread_id).await,
+        };
+        if let Some(summary) = summary {
             merge_mutable_thread_metadata(
                 &mut thread,
                 summary_to_thread(summary, &self.config.cwd),
             );
         }
         self.attach_thread_name(thread_id, &mut thread).await;
+        if thread.forked_from_id.is_none()
+            && let Some(path) = loaded_rollout_path.as_deref()
+            && path.exists()
+        {
+            thread.forked_from_id = forked_from_id_from_rollout(path).await;
+        }
 
         if include_turns {
             let path = loaded_materialized_rollout_path(
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -434,6 +434,11 @@ enum ThreadShutdownResult {
     TimedOut,
 }
 
+enum ThreadReadViewError {
+    InvalidRequest(String),
+    Internal(String),
+}
+
 impl Drop for ActiveLogin {
     fn drop(&mut self) {
         self.cancel();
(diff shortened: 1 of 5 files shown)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3822,65 +3823,59 @@ impl CodexMessageProcessor {
...
-            self.send_internal_error(
-                request_id,
-                format!("failed to locate rollout for thread {thread_uuid}"),
-            )
-            .await;
-            return;
-        }
 
-        let mut thread = if let Some(summary) = db_summary {
-            summary_to_thread(summary, &self.config.cwd)
```
starr-openai, line 3826: are we losing the ability to return thread summary directly from sqlite?  read_thread.rs seems to use sqlite to extend but not as a fallback

wiltzius-openai: Yeah I think this does change behavior too much, I am going to rework this to move this sqlite read into the local thread store implementation and then introduce a more generalized read helper in the app server code that owns merging "live threads" + "stored threads" since this comes up in multiple app server methods.

wiltzius-openai:
@codex review

codex-rs/thread-store/src/local/read_thread.rs
```diff
@@ -59,20 +100,204 @@ pub(super) async fn read_thread(
...
+        .await
+        .ok()
+        .and_then(|meta_line| meta_line.meta.forked_from_id);
+    let mut found_sqlite_title = false;
+    if let Ok(runtime) = codex_state::StateRuntime::init(
+        store.config.sqlite_home.clone(),
+        store.config.model_provider_id.clone(),
+    )
+    .await
+        && let Ok(Some(metadata)) = runtime.get_thread(thread.thread_id).await
```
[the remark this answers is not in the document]
wiltzius-openai: Leaving this out of this PR: caching StateRuntime is a performance improvement, but this PR keeps LocalThreadStore as a lightweight config-backed value and focuses on behavior regressions in read semantics. A cached runtime/shared state handle is a reasonable follow-up once the store owns more SQLite write paths too.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -9199,7 +9337,13 @@ fn summary_from_stored_thread(
         path,
         preview: thread.first_user_message.unwrap_or(thread.preview),
         timestamp: Some(thread.created_at.to_rfc3339_opts(SecondsFormat::Secs, true)),
-        updated_at: Some(thread.updated_at.to_rfc3339_opts(SecondsFormat::Secs, true)),
+        // Match the legacy rollout-backed summary path, which reports updated_at from the
+        // rollout file mtime with millisecond precision.
+        updated_at: Some(
+            thread
+                .updated_at
+                .to_rfc3339_opts(SecondsFormat::Millis, true),
```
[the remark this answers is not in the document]
wiltzius-openai: I am leaving this unchanged. For rollout-backed reads, StoredThread.updated_at is already derived from the rollout list/read summary path, which uses file mtime; for SQLite-only fallback rows, using metadata.updated_at is intentional because there is no discoverable rollout path to stat.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3230,6 +3239,18 @@ impl CodexMessageProcessor {
             .await;
             return;
         }
+        if clears_git_info
```
aibrahim-oai, line 3242: is this change relevant?

codex-rs/thread-store/src/local/read_thread.rs
```diff
@@ -38,41 +60,324 @@ pub(super) async fn read_thread(
...
+    // summary was readable, so SQLite only supplies mutable metadata such as
+    // user-edited title, git fields, or explicit git-field clears.
+    if let Ok(Some(metadata)) = read_sqlite_metadata(store, thread.thread_id).await {
+        if let Some(title) = distinct_title(&metadata) {
+            found_sqlite_title = true;
+            set_thread_name_from_title(&mut thread, title);
+        }
+        if let Some(git_info) = git_info_from_metadata(&metadata) {
+            thread.git_info = Some(git_info);
+        } else if metadata.updated_at > thread.updated_at {
```
starr-openai, line 121: is this a sufficient check - or should we be sure git info was explicitly cleared here?

wiltzius-openai:
Closing this in favor of splitting up into two PRs:
#18231 
#18232

PR #18013, opened 2026-04-15
Add environment registry to exec-server

starr-openai:
## Summary
- turn `EnvironmentManager` into an ID-keyed registry keyed by environment ID
- add `EnvironmentConfig`, registration, lookup, and listing support
- keep default-environment fallback semantics in exec-server

## Notes
- this is the storage and caching layer only
- follow-up PRs thread the selected environment through core and app-server


starr-openai:
Closing in favor of recombined single PR for this work.

PR #18020, opened 2026-04-15
[3/6] Add pushed exec process events

aibrahim-oai:
## Summary
- Add a pushed `ExecProcessEvent` stream alongside retained `process/read` output.
- Publish local and remote output, exit, close, and failure events.
- Cover the event stream with shared local/remote exec process tests.

## Testing
- `cargo check -p codex-exec-server`
- `cargo check -p codex-rmcp-client`
- Not run: `cargo test` per repo instruction; CI will cover.

## Stack
```text
o  #18027 [6/6] Fail exec client operations after disconnect
│
o  #18212 [5/6] Wire executor-backed MCP stdio
│
o  #18087 [4/6] Abstract MCP stdio server launching
│
@  #18020 [3/6] Add pushed exec process events
│
o  #18086 [2/6] Support piped stdin in exec process API
│
o  #18085 [1/6] Add MCP server environment config
│
o  main
```


codex-rs/exec-server/src/client.rs
```diff
@@ -645,6 +677,7 @@ async fn handle_server_notification(
             let session = inner.remove_session(&params.process_id).await;
```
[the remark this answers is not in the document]
jif-oai: Isn't it a fair comment?

codex-rs/exec-server/src/process.rs
```diff
@@ -1,24 +1,125 @@
...
+        }
+    }
+
+    pub(crate) fn publish(&self, event: ExecProcessEvent) {
+        let mut history = self
+            .inner
+            .history
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        history.push_back(event.clone());
```
jif-oai: This event history is capped by event count, not retained bytes. So with large chunks we could potentially be unbounded (and above the existing 1MiNB cap)

codex-rs/exec-server/src/process.rs
```diff
@@ -1,24 +1,125 @@
...
+impl ExecProcessEventReceiver {
+    pub fn empty() -> Self {
+        let (_live_tx, live_rx) = broadcast::channel(1);
+        Self {
+            replay: VecDeque::new(),
+            live_rx,
+        }
+    }
+
+    pub async fn recv(&mut self) -> Result<ExecProcessEvent, broadcast::error::RecvError> {
```
jif-oai, line 149: OOC what's the plan to support `Lagged`? I think this function should be documented as it will be pretty important

codex-rs/exec-server/tests/exec_process.rs
```diff
@@ -148,6 +229,96 @@ async fn assert_exec_process_streams_output(use_remote: bool) -> Result<()> {
     Ok(())
 }
 
+async fn assert_exec_process_pushes_events(use_remote: bool) -> Result<()> {
```
jif-oai, line 232: We should also test the hard case where output/exited/closed/.. notifications arrive out of seq order

codex-rs/exec-server/src/process.rs
```diff
@@ -1,24 +1,125 @@
...
+/// stdout, stderr, or pty bytes. `Exited` reports the process exit status, while
+/// `Closed` means all output streams have ended and no more output events will
+/// arrive. `Failed` is used when the process session cannot continue, for
+/// example because the remote executor connection disconnected.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ExecProcessEvent {
+    Output(ProcessOutputChunk),
+    Exited { seq: u64, exit_code: i32 },
+    Closed { seq: u64 },
+    Failed(String),
```
jif-oai, line 33: Should we add a seq number? Would make things cleaner

jif-oai (commented):
The fact notifications can arrive out of orders was expected (which is why I love having seq numbers) but we will have to be very careful on the client side

Approving to unblock but please process my comments

jif-oai (approved):

aibrahim-oai:
@codex review

PR #18022, opened 2026-04-15
Async config loading

pakrym-oai:
Parts of config will come from executor. Prepare for that by making config loading methods async.


codex-rs/config/src/config_toml.rs
```diff
@@ -607,14 +607,19 @@ impl ConfigToml {
         let sandbox_mode_was_explicit = sandbox_mode_override.is_some()
             || profile_sandbox_mode.is_some()
             || self.sandbox_mode.is_some();
+        let active_project = if sandbox_mode_was_explicit {
+            None
+        } else {
+            self.get_active_project(resolved_cwd).await
+        };
```
starr-openai: why does this move up?

codex-rs/core/src/config/mod.rs
```diff
@@ -1455,7 +1482,7 @@ impl Config {
         let ConfigOverrides {
             model,
             review_model: override_review_model,
-            cwd,
+            cwd: _,
```
starr-openai: ?

starr-openai (approved):

PR #18028, opened 2026-04-16
analytics: add responses api call schema

rhan-oai:
## Why

We need a Responses API call analytics event that can support context-window-health analysis without logging raw prompt, tool, or model content. This PR adds the analytics-side event schema and item-metadata extraction primitives that describe the shape and approximate size of the `ResponseItem`s involved in a Responses API call.

## What changed

This adds the `codex_responses_api_call_event` request shape, including thread and turn identifiers, thread lifecycle metadata, client/runtime metadata, Responses API identifiers, model metadata, terminal status, timing, token counts, and per-item metadata.

It introduces `CodexResponsesApiCallInput`, `CodexResponsesApiCallFact`, `CodexResponsesApiCallStatus`, `CodexResponsesApiItemMetadata`, `CodexResponsesApiItemPhase`, and `CodexResponseItemType` in `codex-rs/analytics/src/facts.rs`. The public API intentionally exposes only the core-facing input/status types from `codex_analytics`; the event fact and item metadata remain analytics-internal.

It also adds `codex-rs/analytics/src/response_items.rs`, which maps protocol `ResponseItem`s into metadata-only analytics records. The mapper captures fields such as item phase, item index, response item type, role, status, message phase, call id, tool name, payload byte size, text part count, and image part count. It avoids raw content while preserving enough structure to analyze context-window composition.

Finally, this adds small serialization/byte helpers in `codex-rs/analytics/src/lib.rs` that are shared by the response item metadata mapper.

## Verification

Verified with `cargo test -p codex-analytics`, which covers the new response item metadata mapping tests.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/18028).
* #18030
* #18029
* __->__ #18028


codex-rs/analytics/src/events.rs
```diff
@@ -55,6 +57,8 @@ pub(crate) enum TrackEventRequest {
     AppUsed(CodexAppUsedEventRequest),
     HookRun(CodexHookRunEventRequest),
     Compaction(Box<CodexCompactionEventRequest>),
+    #[allow(dead_code)]
```
rhan-oai, line 60: to pass lint tests, removed in next commit

PR #18030, opened 2026-04-16
core: emit responses api call analytics

rhan-oai:
## Why

The analytics crate can now represent and reduce Responses API call events, but core still needs to observe actual sampling request lifecycles and submit those observations. This PR wires the event emission into the turn execution path without changing model behavior.

## What changed

`run_turn` now owns a monotonically increasing `turn_responses_call_index` counter for the turn. `run_sampling_request` allocates an index for each actual Responses API attempt, including retry attempts, so emitted analytics can be ordered within the turn.

`run_sampling_request` creates a `ResponsesApiCallAttempt` after prompt construction. That attempt captures the prompt input items that are actually sent to the model, the call start time, completed output items observed from the stream, the Responses API `response_id` when available, and token usage when the stream completes.

To avoid cloning and retaining the full prompt for analytics, the attempt stores precomputed item metadata rather than raw `ResponseItem`s. Input metadata is computed from borrowed prompt items before the API call, and output metadata is appended as each completed output item arrives.

`try_run_sampling_request` now receives the attempt state. It records completed output metadata on `ResponseEvent::OutputItemDone`, records `responses_id` and token usage on `ResponseEvent::Completed`, and otherwise leaves the existing stream handling behavior intact.

Terminal paths in `run_sampling_request` now submit the attempt through `emit_responses_api_call_attempt`. Calls that observe `response.completed` emit `completed`; context-window, usage-limit, exhausted-retry, non-retryable failures, and retryable failures before a retry/fallback emit `failed`; aborted turns and locally preempted streams emit `interrupted`. If cancellation wins before stream setup completes, the attempt is not emitted, avoiding phantom API-call analytics for requests that were never established. The emission remains gated by `Feature::GeneralAnalytics`.

The analytics client also truncates Responses API call errors to 1024 bytes before building the final fact, keeping analytics-specific payload constraints in the analytics crate instead of core.

## Verification

Verified with `cargo check -p codex-core`. The stack was also exercised with `cargo test -p codex-analytics` for the analytics-side event handling.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/18030).
* __->__ #18030
* #18029
* #18028


rhan-oai:
@codex

rhan-oai:
@codex

PR #18033, opened 2026-04-16
feat: introduce codex-pr-body skill

bolinfest:
## Motivation

Codex needs a repeatable workflow for updating PR metadata after a pull request already exists. This is more specific than generic GitHub handling: the assistant needs to preserve author-provided body content, explain why the PR exists before listing implementation details, and describe only the net change under review, including when Sapling stacks put a PR on top of another PR instead of `main`.

## Changes

- Adds `.codex/skills/codex-pr-body/SKILL.md`.
- Documents how to infer the target PR from the current branch or commit, including Sapling-specific PR metadata and `sl sl` output.
- Defines the expected PR body update behavior: inspect the existing body, preserve key content such as images, avoid local absolute paths, use Markdown formatting, include relevant issue/PR references, and call out developer docs follow-up only when applicable.
- Captures stacked-PR handling so generated PR text describes the change between the PR's base and head, rather than unrelated ancestor changes.

## Verification

Not run; this is a Codex skill documentation addition.

### DIFF
--- a/.codex/skills/codex-pr-body/SKILL.md
+++ b/.codex/skills/codex-pr-body/SKILL.md
@@ -1,0 +1,59 @@
+---
+name: codex-pr-body
+description: Update the title and body of one or more pull requests.
+---
+
+## Determining the PR(s)
+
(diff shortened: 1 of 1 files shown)


pakrym-oai (approved):

bolinfest:
I should mention that `sl sl` is equivalent to `sl`, but this:

```
prefix_rule(
  pattern = ["sl", "sl"],
)
```

is safer than this:

```
prefix_rule(
  pattern = ["sl"],
)
```

since the latter allows all Sapling commands, not just the smartlog.

PR #18042, opened 2026-04-16
Make yolo skip managed-network tool enforcement

won-openai:
## Summary

This makes `DangerFullAccess` / yolo tool execution fully opt out of managed-network enforcement.

Previously, yolo turns could have `turn.network` stripped while tool orchestration still derived `enforce_managed_network=true` from `requirements.toml.network`. That created an inconsistent state where the turn had no managed proxy attached, but tool execution still behaved like managed networking was active.

This updates the tool orchestration and JS REPL paths to treat managed networking as active only when the current turn actually has `turn.network`.

## Behavior

- Yolo / `DangerFullAccess`: no managed proxy, no managed-network enforcement.
- Guardian / workspace-write with managed proxy: managed-network enforcement still applies.
- Avoids the half-state where yolo has no proxy but still gets managed-network sandbox behavior.

## Tests

- `just fmt`
- `cargo test -p codex-core danger_full_access_tool_attempts_do_not_enforce_managed_network -- --nocapture`
- `cargo test -p codex-core danger_full_access -- --nocapture`
- `just fix -p codex-core`


won-openai:
@codex review

dylan-hurd-oai (approved):
cc @bolinfest for visibility as well

PR #18046, opened 2026-04-16
Add built-in local and remote environment selection

starr-openai:
## Summary
- simplify EnvironmentManager to support built-in `local` and `remote` selections instead of a dynamic registry
- preserve existing default environment behavior while allowing explicit local and remote overrides
- add exec-server unit coverage for default, local, remote, and disabled selection cases

## Notes
- `environment_id` values are now constrained to `local` and `remote` when provided explicitly
- this PR intentionally removes the dynamic registration/listing direction in favor of the simpler built-in model discussed in Slack

## Validation
- not run (not requested)


starr-openai:
Closing this stack to reset and rethink the minimal environment-id design.

Issue #12178, opened 2026-02-18
Defaulting to original name in session renaming


xiaowu0162:
### What variant of Codex are you using?

CLI

### What feature would you like to see?

Thank you for the great work! I wonder if user runs `/rename` in a thread that already has custom names, is it possible to populate the name field with the original name? This would make renaming much easier since new names in this case are often a few edits from the old names. 

019c7227-e41e-70e3-bf67-08363ba3eda2

PremPrakashCodes:
I have the implementation on a branch if it would be helpful: https://github.com/PremPrakashCodes/codex/tree/fix/rename-prompt-prefill

- Added a with_initial_text(text: &str) -> Self builder method to CustomPromptView in custom_prompt_view.rs. It calls the existing set_text_clearing_elements API on the  
  internal TextArea and then sets the cursor to text.len(), so the prompt opens with the caret at the end of the pre-filled text rather than at position 0.
- Updated show_rename_prompt() in chatwidget.rs to read the existing thread_name as Option<&str> via as_deref().filter(|n| !n.is_empty()), avoiding an extra String       
  allocation. When the thread already has a non-empty name, that borrowed slice is passed into the prompt via with_initial_text; otherwise the prompt opens blank as before.
- The prompt title is still chosen from the same condition: "Rename thread" when a name exists, "Name thread" when it doesn't.
- No change to the "Name thread" flow — unnamed threads still open with an empty field and the existing title.                                                            
- No change to the submit path — the entered name still flows through normalize_thread_name, so trimming and empty-name rejection behave exactly as before.               
- No new dependencies; the with_initial_text builder follows the same consuming-self builder pattern already used elsewhere in CustomPromptView.

etraut-openai:
This will be included in the next release.

PR #18057, opened 2026-04-16
Prefill rename prompt with current thread name

etraut-openai:
Addresses #12178

Problem: The TUI /rename prompt opened blank even when the current thread already had a custom name, making small edits awkward.

Solution: Let custom prompts receive initial text and prefill /rename with the existing thread name while preserving the empty prompt for unnamed threads.

Testing: Manually verified that the feature works by using `/rename` with unnamed and already-named threads.


fcoury-oai (approved):
Tested the feature, and it works great.

Code looks good, approved.

Issue #18011, opened 2026-04-15
"codex resume" command hints print invalid <session-guid>


mirekjany:
### What version of Codex CLI is running?

codex-cli 0.121.0

### What subscription do you have?

Pro

### Which model were you using?

gpt-5.3-codex, gpt-5.4-codex

### What platform is your computer?

Linux 6.6.87.2-microsoft-standard-WSL2 x86_64 x86_64

### What terminal emulator and version are you using (if applicable)?

wsl2

### What issue are you seeing?

First observed when i upgraded from codex-cli 0.118.0 to 0.120.0 and now to 0.121.0 hoping this would already be fixed, codex-cli simply started errorneously issuing "To continue this session, run codex resume session-guid" command hints with invalid session-guids: 1) hitting ctrl+d on a new session without starting any conversation yet - this hint is issued 1x upon exiting. 2) resuming existing session, hitting ctrl+d -> two (!!) of these hints are appended to the session history with different (!!) guids. reverting to 0.118.0 doesn't do any of this. 

### What steps can reproduce the bug?

Empty session:
1) start codex-cli (either 0.120.0 or 0.121.0) on a new project, accept/trust the folder
2) hit ctrl+d -> a "To continue this session, run codex resume session-guid" hint is printed before exiting codex-cli
3) try to run "codex resume" from CLI with that session guid and you're greeted with "ERROR: No saved session found with ..."

Existing session:
1) start codex-cli (either 0.120 or 0.121.0) on an existing project
2) type /resume and choose an existing session
3) wait for session to resume
4) a "To continue this session, run codex resume session-guid" hint is leaked to the session history
5) hit ctrl+d -> another "To continue this session, run codex resume session-guid" is printed before fully exiting codex-cli
6) 1st session guid is invalid (the same error as with empty session), however the second session guid is the valid one.

### What is the expected behavior?

Expected behaviour is not to print "codex resume" command hint with invalid session guid which can't be resumed (on both empty + existing session).

etraut-openai:
Thanks for the bug report. This will be addressed in the next release.

PR #18059, opened 2026-04-16
Fix invalid TUI resume hints

etraut-openai:
Addresses #18011

Problem: #16987 allowed zero-token TUI exits to print resume hints, which exposed precomputed thread ids before their rollout files were persisted; #17222 made the same invalid hint visible when switching sessions via `/resume`.

Solution: Only include resume commands for TUI sessions backed by a materialized non-empty rollout, and cover both missing-rollout and persisted-rollout summary behavior. 

Testing: Manually verified by pressing Ctrl+D before the first prompt and confirming that no "to continue this session" message was generated.


fcoury-oai (approved):
Replicated the issue and the fix. Approved.

Issue #17951, opened 2026-04-15
TUI crashes when skills/list fails behind corporate proxy


jinyule:
## Bug description

When running Codex CLI behind a corporate proxy/firewall that intercepts HTTPS requests to `chatgpt.com`, the TUI crashes with:

```
Error: skills/list failed in TUI

Caused by:
    0: request to https://chatgpt.com/backend-api/plugins/featured?platform=codex failed
    1: error decoding response body
    2: expected value at line 1 column 1
```

The proxy redirects the request via 302 to an HTML warning page, which is then returned with a 200 status code. The reqwest client follows the redirect, receives HTML instead of JSON, and fails to parse it.

## Root cause

In `codex-rs/tui/src/app.rs`, the `AppCommandView::ListSkills` handler propagates errors with `?`:

```rust
AppCommandView::ListSkills { cwds, force_reload } => {
    let response = app_server
        .skills_list(...)
        .await?;  // <-- crashes TUI on error
    self.handle_skills_list_response(response);
    Ok(true)
}
```

This error bubbles up through the TUI event loop and causes a fatal exit. However, the startup code path for the same operation already handles this gracefully with `match + warn`.

## Expected behavior

Plugin discovery failures (skills/list, featured plugins, etc.) should degrade gracefully — log a warning and continue operating. The main TUI functionality should not depend on plugin discovery succeeding.

## Steps to reproduce

1. Run Codex CLI behind a corporate proxy/firewall that intercepts requests to `chatgpt.com`
2. The proxy returns HTML instead of JSON for `chatgpt.com/backend-api/plugins/featured`
3. TUI crashes on startup or when skills are refreshed

## Environment

- OS: Windows 11
- Network: Behind corporate proxy/firewall
- Auth: ChatGPT auth mode

etraut-openai:
Thanks for the bug report. This will be fixed in the next release.

PR #18061, opened 2026-04-16
Avoid fatal TUI errors on skills list failure

etraut-openai:
Addresses #17951

Problem: The TUI treated skills/list failures as fatal during refresh, so proxy/firewall responses that break plugin discovery could crash the session.

Solution: Route startup and refresh skills/list responses through shared graceful handling that logs a warning and keeps the TUI running.

### DIFF
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -2486,14 +2486,16 @@ impl App {
                 Ok(true)
             }
             AppCommandView::ListSkills { cwds, force_reload } => {
-                let response = app_server
-                    .skills_list(codex_app_server_protocol::SkillsListParams {
-                        cwds: cwds.to_vec(),
-                        force_reload,
-                        per_cwd_extra_user_roots: None,
-                    })
-                    .await?;
-                self.handle_skills_list_response(response);
+                self.handle_skills_list_result(
+                    app_server
+                        .skills_list(codex_app_server_protocol::SkillsListParams {
+                            cwds: cwds.to_vec(),
+                            force_reload,
+                            per_cwd_extra_user_roots: None,
+                        })
+                        .await,
+                    "failed to refresh skills",
+                );
                 Ok(true)
             }
             AppCommandView::Compact => {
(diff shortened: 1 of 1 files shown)


codex-rs/tui/src/app.rs
```diff
@@ -2486,14 +2486,16 @@ impl App {
...
             AppCommandView::ListSkills { cwds, force_reload } => {
-                let response = app_server
-                    .skills_list(codex_app_server_protocol::SkillsListParams {
-                        cwds: cwds.to_vec(),
-                        force_reload,
-                        per_cwd_extra_user_roots: None,
-                    })
-                    .await?;
-                self.handle_skills_list_response(response);
+                self.handle_skills_list_result(
```
bolinfest, line 2489: Shouldn't this take the return value and return it with `Ok` instead of hardcoding `Ok(true)` below?

codex-rs/tui/src/app.rs
bolinfest, line 2378: Should this be forced to return `bool` rather than `Result`?

codex-rs/tui/src/app.rs
etraut-openai: [codex] I agree the ignored return value is awkward, but I don't think we should propagate `false` here. In this method, `Ok(false)` means "this op was not handled" and would fall through to a user-visible "Not available in TUI yet" message. For a failed skills refresh, we want the op to count as handled while the refresh itself degrades gracefully. I'll remove the helper's boolean return instead so this is clearer.

codex-rs/tui/src/app.rs
etraut-openai: [codex] I think `try_submit_active_thread_op_via_app_server` should stay `Result<bool>` because most branches still represent important app-server failures that should bubble up. The narrower fix is to make only `skills/list` fail-open, since plugin/skills discovery is noncritical. I'll change the skills helper to return `()` so the intent is explicit without broadening error swallowing.

codex-rs/tui/src/app.rs
```diff
@@ -2567,6 +2569,19 @@ impl App {
         }
     }
 
+    fn handle_skills_list_result(
```
bolinfest, line 2572: To make this a more meaningful helper, maybe this should take `app_server`, `cwds`, and `force_reload` and make the `skills_list()` call and then handle the `Result` internally?

Up to you: I don't feel strongly either way.

bolinfest (approved):

PR #18063, opened 2026-04-16
Fix: /review child flows leaving the menu open

canvrno-oai:
This fixes a /review regression introduced by #15134 , where child review flows could submit successfully but leave the parent review preset menu visible. The regression came from the bottom-pane stack moving to more precise pop-self behavior for completed views, which was correct some dismissals but doesn't work for wizard-style `/review` submenus.

- Added a small `pop_all_views()` helper in the bottom pane that drains the stack by repeatedly popping the active view.
- Added a targeted app event to dismiss all active bottom-pane views.
- Updated `/review` child submit flows to use that dismiss-all behavior after dispatching the review request:
  - base branch picker
  - commit picker
  - custom review prompt
- Added a regression test covering branch-picker submit dismissing the full `/review` popup stack.

Steps to reproduce before the fix:
  1. Start Codex TUI in a git repo.
  2. Run /review.
  3. Select Review against a base branch.
  4. Select any branch and press Enter.
  5. Observe: review starts, but the /review preset menu appears again.

Steps to verify the fix:
  1. Run /review.
  2. Select Review against a base branch.
  3. Select any branch and press Enter.
  4. Expected: review starts and the whole /review menu stack disappears.
  
  Repeat with commit selection and custom instruction options for a complete dupliation/test,


fcoury-oai (approved):
Looks good. I replicated the issue on main and confirmed this branch fixes it.

Approved.

canvrno-oai:
Closing as this was fixed by #18154

PR #18070, opened 2026-04-16
Extract plugin loading and marketplace logic into codex-core-plugins

xl-openai:
Split plugin loading, marketplace, and related infrastructure out of core into codex-core-plugins, while keeping the core-facing configuration and orchestration flow in codex-core.


xl-openai:
@codex can you check if there is anything missing in this refactor? reply concisely.

codex-rs/core/src/plugins/remote.rs
```diff
@@ -1,313 +1,63 @@
...
-        let mut segments = url
-            .path_segments_mut()
-            .map_err(|()| RemotePluginMutationError::InvalidBaseUrlPath)?;
-        segments.pop_if_empty();
-        segments.push("plugins");
-        segments.push(plugin_id);
-        segments.push(action);
-    }
-    Ok(url.to_string())
+    codex_core_plugins::remote::uninstall_remote_plugin(
```
pakrym-oai: can you remove these proxy methods?

codex-rs/core-plugins/src/loader.rs
```diff
@@ -0,0 +1,807 @@
+use crate::manifest::PluginManifestPaths;
```
pakrym-oai, line 1: can this file move out as well?

pakrym-oai (approved):

PR #18073, opened 2026-04-16
Add goal persistence foundation (1 / 5)

etraut-openai:
Adds the persisted goal foundation for the rest of the stack. This PR is intentionally limited to feature flag and state-layer behavior; app-server APIs, model tools, runtime continuation, and TUI UX are layered in later PRs.

## Why

Goal mode needs durable thread-level state before clients or model tools can safely build on it. The state layer needs to know whether a goal exists, what objective it tracks, whether it is active, paused, budget-limited, or complete, and how much time/token usage has already been accounted.

## What changed

- Added the `goals` feature flag and generated config schema entry.
- Added the `thread_goals` state table and Rust model for persisted thread goals.
- Added state runtime APIs for creating, replacing, updating, deleting, and accounting goal usage.
- Added `goal_id`-based stale update protection so an old goal update cannot overwrite a replacement.
- Kept this PR scoped to persistence and state runtime behavior, with no app-server, model-facing, continuation, or TUI behavior yet.

## Verification

- Added state runtime coverage for goal creation, replacement, stale update protection, status transitions, token-budget behavior, and usage accounting.


codex-rs/features/src/tests.rs
```diff
@@ -138,6 +138,13 @@ fn request_permissions_tool_is_under_development() {
     assert_eq!(Feature::RequestPermissionsTool.default_enabled(), false);
 }
 
+#[test]
+fn goal_mode_is_under_development() {
```
etraut-openai: This test isn't valuable. Remove it.

codex-rs/features/src/lib.rs
```diff
@@ -171,6 +171,8 @@ pub enum Feature {
     DefaultModeRequestUserInput,
     /// Enable automatic review for approval prompts.
     GuardianApproval,
+    /// Enable persisted thread goals and automatic goal continuation.
+    GoalMode,
```
etraut-openai: The name of the feature should be "Goals", not "GoalMode".

codex-rs/state/src/runtime/goals.rs
```diff
@@ -0,0 +1,748 @@
...
+    created_at_ms,
+    updated_at_ms
+) VALUES (?, ?, ?, ?, 0, 0, ?, ?)
+ON CONFLICT(thread_id) DO UPDATE SET
+    objective = excluded.objective,
+    status = excluded.status,
+    token_budget = excluded.token_budget,
+    tokens_used = 0,
+    time_used_seconds = 0,
+    created_at_ms = excluded.created_at_ms,
```
jif-oai, line 81: we could have collisions on `ms`. This is low probability but still interesting to note

codex-rs/state/src/runtime/goals.rs
```diff
@@ -0,0 +1,748 @@
...
+            "#,
+        )
+        .bind(thread_id.to_string())
+        .execute(self.pool.as_ref())
+        .await?;
+
+        Ok(result.rows_affected() > 0)
+    }
+
+    pub async fn account_thread_goal_usage(
```
jif-oai, line 316: Here you're `UPDATE` and return `SELECT` (through `get_thread_goal()`) are in different transaction. This is an entry point for race conditions. Ideally we should switch to a `UPDATE ... RETURNING ...` to conserve single transaction

codex-rs/state/migrations/0029_thread_goals.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE thread_goals (
```
jif-oai, line 1: because of this, we need to sync the release between the app and the CLI

codex-rs/features/src/lib.rs
```diff
@@ -896,6 +898,16 @@ pub const FEATURES: &[FeatureSpec] = &[
         },
         default_enabled: false,
     },
+    FeatureSpec {
+        id: Feature::Goals,
+        key: "goals",
+        stage: Stage::Experimental {
```
jif-oai: Let's first make sure every is merged before going to experimental

jif-oai (approved):
Ok after my comments

codex-rs/state/src/runtime/goals.rs
etraut-openai: Good point. I've changed it to use a goal_id (a GUID) to eliminate this hole.

PR #18074, opened 2026-04-16
Add goal app-server API (2 / 5)

etraut-openai:
Adds the app-server v2 goal API on top of the persisted goal state from PR 1.

## Why

Clients need a stable app-server surface for reading and controlling materialized thread goals before the model tools and TUI can use them. Goal changes also need to be observable by app-server clients, including clients that resume an existing thread.

## What changed

- Added v2 `thread/goal/get`, `thread/goal/set`, and `thread/goal/clear` RPCs for materialized threads.
- Added `thread/goal/updated` and `thread/goal/cleared` notifications so clients can keep local goal state in sync.
- Added resume/snapshot wiring so reconnecting clients see the current goal state for a thread.
- Added app-server handlers that reconcile persisted rollout state before direct goal mutations.
- Updated the app-server README plus generated JSON and TypeScript schema fixtures for the new API surface.

## Verification

- Added app-server v2 coverage for goal get/set/clear behavior, notification emission, resume snapshots, and non-local thread-store interactions.


codex-rs/protocol/src/protocol.rs
```diff
@@ -3512,6 +3515,40 @@ pub struct ThreadNameUpdatedEvent {
     pub thread_name: Option<String>,
 }
 
+#[derive(Debug, Clone, Copy, Deserialize, Serialize, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "protocol/")]
+pub enum ThreadGoalStatus {
+    Active,
+    Paused,
+    #[serde(alias = "budgetExceeded")]
```
etraut-openai: There shouldn't be an alias here.

codex-rs/protocol/src/protocol.rs
```diff
@@ -3512,6 +3515,40 @@ pub struct ThreadNameUpdatedEvent {
...
+    BudgetStopped,
```
etraut-openai: I don't like the name "BudgetStopped". Rename to "BudgetLimited".

codex-rs/protocol/src/protocol.rs
```diff
@@ -3614,6 +3651,7 @@ pub enum TurnAbortReason {
     Interrupted,
     Replaced,
     ReviewEnded,
+    BudgetExceeded,
```
etraut-openai: This shouldn't be BudgetExceeded. Change to BudgetLimited.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3035,6 +3057,458 @@ impl CodexMessageProcessor {
             .await;
     }
 
+    async fn thread_goal_set(&self, request_id: ConnectionRequestId, params: ThreadGoalSetParams) {
```
etraut-openai: This file is already really large. Move the goal-related handlers out of this file into a dedicated module.

codex-rs/core/src/agent/control_tests.rs
```diff
@@ -308,6 +308,19 @@ async fn on_event_updates_status_from_turn_aborted() {
     assert_eq!(status, Some(expected));
 }
 
+#[tokio::test]
+async fn on_event_updates_status_from_budget_limited_abort() {
```
etraut-openai: I don't think this test adds much value. Remove it.

codex-rs/app-server/src/codex_message_processor/thread_goal_handlers.rs
```diff
@@ -0,0 +1,470 @@
...
+                );
+                return;
+            }
+        };
+        let listener_command_tx = {
+            let thread_state = self.thread_state_manager.thread_state(thread_id).await;
+            let thread_state = thread_state.lock().await;
+            thread_state.listener_command_tx()
+        };
+        match state_db.get_thread_goal(thread_id).await {
```
fcoury-oai: I asked Codex to do another review pass, and it found a client sync issue where the client can display stale objective, status, token budget and usage until a subsequent fresh read or goal update. 

It seems here we are now routing the final notification through `emit_thread_goal_updated_ordered`, but `emit_thread_goal_snapshot` is still reading the goal from sql before entering this ordered path.

A concurrent `thread/goal/set` can commit a newer goal after this read and before the resume snapshot is enqueued, so clients can still receive `new` and then the stale resume snapshot `old`.

The read plus notification enqueue needs to happen inside the same serialized listener command path as goal mutations, not just the notification send.

It seems to be more of a algorithm correctness concern than a likely bug, so make of it what you will.

I did attempt to trigger it with Codex's help by creating a script:

1. Thread has goal old.
2. Client A sends thread/resume on a cold/unloaded thread.
3. Resume path reads old for its goal snapshot.
4. Client B sends thread/goal/set to new.
5. Client receives thread/goal/updated new.
6. Resume snapshot arrives afterward as thread/goal/updated old.

It didn't trigger after 20 iterations.

fcoury-oai (approved):
I have reviewed the code and it looks good. I also manually tested the solution by running the app server and sending jsonl messages that were crafted with Codex's help and the happy path worked perfectly.

Codex found a follow-up potential correctness issue, related to another one you fixed before. 

The solution design is very solid. One feature I missed as a user is the ability to have a budget by iterations instead of by tokens, something that is very common on similar solutions and would feel more natural to me as a user of the other solutions. This can easily be added as a follow-up later if we want.

None of those findings are blockers, so I am approving the code.

codex-rs/app-server/src/codex_message_processor/thread_goal_handlers.rs
etraut-openai: Addressed in ef6d66c7f5. Cold resume now enqueues an `EmitThreadGoalSnapshot` listener command, and that command reads the current goal state inside the serialized listener path before emitting the notification. Running-thread resume now uses the same snapshot helper, so resume snapshots are ordered with concurrent goal mutations.

PR #18075, opened 2026-04-16
Add goal model tools (3 / 5)

etraut-openai:
Adds the model-facing goal tools on top of the app-server API from PR 2.

## Why

Once goals are persisted and exposed to clients, the model needs a small, constrained tool surface for goal workflows. The tool contract should let the model inspect goals, create them only when explicitly requested, and mark them complete without giving it broad control over user/runtime-owned state.

## What changed

- Added `get_goal`, `create_goal`, and `update_goal` tool specs behind the `goals` feature flag.
- Added core goal tool handlers that validate objectives and token budgets before mutating persisted state.
- Constrained `create_goal` to create only when no goal exists, with optional `token_budget` only when a budget is explicitly provided.
- Tightened the `create_goal` instructions so the model does not infer goals from ordinary task requests.
- Constrained `update_goal` to expose only goal completion; pause, resume, clear, and budget-limited transitions remain user- or runtime-controlled.
- Registered the goal tools in the tool registry and kept them out of review contexts where they should not appear.

## Verification

- Added tool-registry coverage for feature gating and tool availability.
- Added core session tests for create/get/update behavior, duplicate goal rejection, budget validation, and completion-only updates.


codex-rs/core/src/tools/spec_tests.rs
```diff
@@ -429,6 +429,34 @@ async fn test_build_specs_gpt5_codex_default() {
     .await;
 }
 
+#[tokio::test]
+async fn test_goal_mode_adds_goal_tools() {
```
etraut-openai: I don't think this test is valuable. Remove it.

codex-rs/tools/src/goal_tool.rs
```diff
@@ -0,0 +1,111 @@
...
+            JsonSchema::integer(Some(
+                "Optional positive token budget for the new active goal.".to_string(),
+            )),
+        ),
+    ]);
+
+    ToolSpec::Function(ResponsesApiTool {
+        name: SET_GOAL_TOOL_NAME.to_string(),
+        description: r#"Start a new long-running goal for this thread.
+This tool always creates or replaces an active goal. It resets time/token usage accounting to zero when the objective is new.
```
etraut-openai: The text says that it replaces an active goal. Actually, it replaces any existing goal regardless of whether it's active.

codex-rs/tools/src/goal_tool.rs
```diff
@@ -0,0 +1,111 @@
...
+If the existing non-terminal goal already has the same objective, the call is treated as an idempotent update that keeps the goal active and preserves usage accounting.
```
etraut-openai: We added a new update_goal tool to eliminate the need for the set_goal tool to have this strange dual-nature behavior where it changes its behavior based on whether the objective changes. This text should be updated to match the current actual semantics of set_goal. You should confirm, but I think that set_goal no longer preserves usage accounting; update_goal does.

codex-rs/tools/src/goal_tool.rs
```diff
@@ -0,0 +1,111 @@
...
+            "status".to_string(),
+            JsonSchema::string_enum(
+                vec![
+                    json!("active"),
+                    json!("paused"),
+                    json!("budgetLimited"),
+                    json!("complete"),
+                ],
+                Some(
+                    "Optional. Set to active, paused, budgetLimited, or complete. Use complete only when the objective is achieved and no required work remains. Use budgetLimited when the objective has not been achieved and cannot be achieved within the remaining budget, or when the remaining budget is too small for productive continuation."
```
etraut-openai: Remove the language about budgetLimited. We don't expect the model to ever set budgetLImited. This is the responsibility of the harness runtime when it discovers that the token budget is exceeded. The current language was left over from a previous attempt to allow the agent to control this, but those experiments were not successful.

codex-rs/tools/src/goal_tool.rs
```diff
@@ -0,0 +1,111 @@
...
+                        .to_string(),
+                ),
+            ),
+        ),
+        (
+            "token_budget".to_string(),
```
etraut-openai: update_goal should not accept an updated token budget. That should be settable only through set_goal. Remove this parameter.

codex-rs/tools/src/goal_tool.rs
```diff
@@ -0,0 +1,111 @@
...
+            nullable_integer("Optional positive token budget. Use null to clear the existing token budget while preserving usage accounting."),
+        ),
+    ]);
+
+    ToolSpec::Function(ResponsesApiTool {
+        name: UPDATE_GOAL_TOOL_NAME.to_string(),
+        description: r#"Update the existing long-running goal while preserving time/token usage accounting.
+Use this tool to pause, resume, mark the goal achieved, mark it budget-limited, or change the token budget.
```
etraut-openai: Remove "mark it budget-limited" from this text. We don't want the agent to mark the goal as budget-limited. That's done by the harness runtime. Also, remove the "change the token budget". This should be setable only through set_goal.

codex-rs/tools/src/goal_tool.rs
```diff
@@ -0,0 +1,111 @@
...
+Set status to `complete` only when the objective has actually been achieved and no required work remains.
+Set status to `budgetLimited` when a budgeted goal has not been achieved and cannot be achieved within the remaining budget, or when the budget is exhausted or nearly exhausted.
+Do not mark a goal complete merely because its budget is nearly exhausted or because you are stopping work.
+When marking a budgeted goal achieved with status `complete`, report the final token usage from the tool result to the user.
+The system owns usage fields, so this tool cannot set them directly."#
```
etraut-openai: I think it's OK to remove the statement "The system owns usage fields, so this tool cannot set them directly.". That's implicit in the parameters.

codex-rs/core/src/tools/handlers/goal.rs
```diff
@@ -0,0 +1,275 @@
...
+    session: &Session,
+    turn_context: &TurnContext,
+    arguments: &str,
+) -> Result<FunctionToolOutput, FunctionCallError> {
+    let args: SetGoalArgs = parse_arguments(arguments)?;
+    if session
+        .get_thread_goal()
+        .await
+        .map_err(|err| FunctionCallError::RespondToModel(format_goal_error(err)))?
+        .is_some()
```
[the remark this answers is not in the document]
etraut-openai: [codex] This has been addressed on the current branch. `create_goal` now routes through `Session::create_thread_goal`, which calls `StateRuntime::insert_thread_goal`; that path uses a single `INSERT ... ON CONFLICT(thread_id) DO NOTHING` operation and returns the existing-goal error when no row is inserted. There is also coverage in `insert_thread_goal_does_not_replace_existing_goal` for the duplicate insert behavior.

codex-rs/core/src/tools/handlers/goal.rs
```diff
@@ -0,0 +1,271 @@
+//! Built-in model tool handlers for persisted thread goals.
+//!
+//! The public tool contract intentionally splits goal creation from completion:
+//! `create_goal` starts an active objective, while `update_goal` can only mark
+//! the existing goal complete and preserve usage accounting.
```
etraut-openai: The statement "and preserve usage accounting" is a bit confusing here. Maybe just delete this?

codex-rs/core/src/tools/handlers/goal.rs
```diff
@@ -0,0 +1,271 @@
...
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "snake_case")]
+struct UpdateGoalArgs {
+    status: ToolGoalStatus,
+}
+
+#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+enum ToolGoalStatus {
```
etraut-openai: Would it make sense to reuse the existing enumeration for ThreadGoalStatus rather than create another one here? Leave it as is if you have a good reason for creating a new enum here.

codex-rs/tools/src/goal_tool.rs
```diff
@@ -0,0 +1,114 @@
...
+            "token_budget".to_string(),
+            JsonSchema::integer(Some(
+                "Optional positive token budget for the new active goal.".to_string(),
+            )),
+        ),
+    ]);
+
+    ToolSpec::Function(ResponsesApiTool {
+        name: CREATE_GOAL_TOOL_NAME.to_string(),
+        description: r#"Create a new long-running goal for this thread when no goal is already defined.
```
etraut-openai: Do we need to mention "long-running" here? Maybe remove that phrase?

codex-rs/tools/src/goal_tool.rs
```diff
@@ -0,0 +1,114 @@
...
+This tool creates a fresh active goal and resets time/token usage accounting to zero.
```
etraut-openai: "fresh" is an odd word choice here. How about "new"?

codex-rs/tools/src/goal_tool.rs
```diff
@@ -0,0 +1,114 @@
...
+It fails if any goal already exists. Use update_goal, not create_goal, only to mark an existing goal achieved while preserving usage accounting.
```
etraut-openai: I don't think we need to say "while preserving usage accounting" here. The model doesn't need to know this.

codex-rs/tools/src/goal_tool.rs
```diff
@@ -0,0 +1,114 @@
...
+To replace the objective, the user must clear or replace the goal through goal-management UI or API.
```
etraut-openai: Do we really need to say anything about replacing the objective? Maybe remove this sentence?

codex-rs/tools/src/goal_tool.rs
```diff
@@ -0,0 +1,114 @@
...
+            Some(
+                "Required. Set to complete only when the objective is achieved and no required work remains."
+                    .to_string(),
+            ),
+        ),
+    )]);
+
+    ToolSpec::Function(ResponsesApiTool {
+        name: UPDATE_GOAL_TOOL_NAME.to_string(),
+        description: r#"Update the existing long-running goal while preserving time/token usage accounting.
```
etraut-openai: See my previous comments about "long-running" terminology, etc.

codex-rs/core/src/goals.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+}
+
+pub(crate) fn protocol_goal_from_state(goal: codex_state::ThreadGoal) -> ThreadGoal {
+    ThreadGoal {
+        thread_id: goal.thread_id,
+        objective: goal.objective,
+        status: protocol_goal_status_from_state(goal.status),
+        token_budget: goal.token_budget,
+        tokens_used: goal.tokens_used,
```
jif-oai, line 223: Maybe I'm blind but I can't find where this is updated

codex-rs/tools/src/goal_tool.rs
```diff
@@ -0,0 +1,113 @@
...
+                "Optional positive token budget for the new active goal.".to_string(),
+            )),
+        ),
+    ]);
+
+    ToolSpec::Function(ResponsesApiTool {
+        name: CREATE_GOAL_TOOL_NAME.to_string(),
+        description: r#"Create a new goal for this thread when no goal is already defined.
+This tool creates a new active goal and resets time/token usage accounting to zero.
+It fails if any goal already exists. Use update_goal, not create_goal, only to mark an existing goal achieved.
```
jif-oai: For `update_goal` us UPDATE_GOAL_TOOL_NAME

codex-rs/tools/src/goal_tool.rs
```diff
@@ -0,0 +1,113 @@
...
+        ),
+        (
+            "token_budget".to_string(),
+            JsonSchema::integer(Some(
+                "Optional positive token budget for the new active goal.".to_string(),
+            )),
+        ),
+    ]);
+
+    ToolSpec::Function(ResponsesApiTool {
```
jif-oai, line 45: What if the goal it tries to create has a collision with an already existing goal? Then the model is prompted to use `update_goal` but this will fail right?

jif-oai (commented):
Arguably `update` and `create` could be the same tool

codex-rs/core/src/goals.rs
etraut-openai: This is updated in the next PR in the stack (#18076), where goal runtime/accounting paths update the persisted goal row and refresh updated_at_ms. This PR only exposes the field through the protocol conversion, so it isn't obvious from this diff alone.

codex-rs/tools/src/goal_tool.rs
etraut-openai: I'm not 100% sure I know what you meant by this comment, but I took a crack at updating the create_goal description.

codex-rs/core/src/goals.rs
```diff
@@ -0,0 +1,245 @@
...
+                .await?
+        } else {
+            let status = request.status.map(state_goal_status_from_protocol);
+            state_db
+                .update_thread_goal(
+                    self.conversation_id,
+                    codex_state::ThreadGoalUpdate {
+                        status,
+                        token_budget: request.token_budget,
+                        expected_goal_id: None,
```
jif-oai, line 78: I guess this will get populated at some point?

codex-rs/tools/src/goal_tool.rs
```diff
@@ -0,0 +1,114 @@
...
+    )]);
+
+    ToolSpec::Function(ResponsesApiTool {
+        name: UPDATE_GOAL_TOOL_NAME.to_string(),
+        description: r#"Update the existing goal.
+Use this tool only to mark the goal achieved.
+Set status to `complete` only when the objective has actually been achieved and no required work remains.
+Do not mark a goal complete merely because its budget is nearly exhausted or because you are stopping work.
+You cannot use this tool to pause, resume, or budget-limit a goal; those status changes are controlled by the user or system.
+When marking a budgeted goal achieved with status `complete`, report the final token usage from the tool result to the user."#
```
jif-oai, line 81: This does not seem to be implemented as the current code return immediatly after the goal is marked as complete (i.e. before the completing turn's usage is accounted)

codex-rs/state/src/runtime/goals.rs
```diff
@@ -105,6 +105,49 @@ RETURNING
...
+        .bind(now_ms)
+        .bind(now_ms)
+        .execute(self.pool.as_ref())
+        .await?;
+
+        if result.rows_affected() == 0 {
+            return Ok(None);
+        }
+
+        self.get_thread_goal(thread_id).await
```
jif-oai: I think I already posted on this but could we just do `INSERT ... ON CONFLICT DO NOTHING RETURNING ...`? Having everything in one transaction is more efficient and will prevent future races

codex-rs/core/src/goals.rs
etraut-openai: Yes, this gets populated in the next PR (4 / 5).

codex-rs/tools/src/goal_tool.rs
etraut-openai: The behavior described here is implemented in the next PR (4 / 5).

codex-rs/tools/src/goal_tool.rs
etraut-openai: I updated the prompt so it's clearer.

codex-rs/state/src/runtime/goals.rs
etraut-openai: Good catch. 

PR1 now adds the atomic state helper here, using `INSERT ... ON CONFLICT(thread_id) DO NOTHING RETURNING ...`.

PR3 then wires `create_thread_goal` to call that helper here, so goal creation now succeeds only when the insert wins and returns an error when a goal already exists.

codex-rs/tools/src/tool_config.rs
```diff
@@ -144,6 +145,7 @@ impl ToolsConfig {
         let include_js_repl_tools_only =
             include_js_repl && features.enabled(Feature::JsReplToolsOnly);
         let include_collab_tools = features.enabled(Feature::Collab);
+        let include_goal_tools = features.enabled(Feature::Goals);
```
jif-oai, line 144: Should we also gate somewhere about goal + ephemeral?

codex-rs/core/src/goals.rs
```diff
@@ -0,0 +1,247 @@
...
+
+pub(crate) struct CreateGoalRequest {
+    pub(crate) objective: String,
+    pub(crate) token_budget: Option<i64>,
+}
+
+impl Session {
+    pub(crate) async fn get_thread_goal(&self) -> anyhow::Result<Option<ThreadGoal>> {
+        if !self.enabled(Feature::Goals) {
+            return Ok(None);
```
jif-oai: The other functions error when the feature is disabled, shouldn't we also error? This would feel strange to go through this code without the feature flag

codex-rs/core/src/tools/handlers/goal.rs
```diff
@@ -0,0 +1,263 @@
...
+
+impl GoalToolResponse {
+    fn new(goal: Option<ThreadGoal>) -> Self {
+        let remaining_tokens = goal.as_ref().and_then(|goal| {
+            goal.token_budget
+                .map(|budget| (budget - goal.tokens_used).max(0))
+        });
+        let completion_budget_report = goal
+            .as_ref()
+            .filter(|goal| goal.status == ThreadGoalStatus::Complete)
```
jif-oai: Since this response helper is also used by `get_goal`, any later read of a completed goal can keep telling the model to report final budget usage again. I’d make `completionBudgetReport` update-goal-only.

jif-oai (commented):
Minor comments but lgtm

jif-oai (approved):

codex-rs/tools/src/tool_config.rs
etraut-openai: Yeah, I didn't think of that. I agree we should disallow these tools for ephemeral threads. Changed.

codex-rs/core/src/goals.rs
etraut-openai: Yep. Done.

codex-rs/core/src/tools/handlers/goal.rs
etraut-openai: Yep, good point. Done.

codex-rs/core/src/session/turn_context.rs
```diff
@@ -442,6 +443,7 @@ impl Session {
         .with_spawn_agent_usage_hint(per_turn_config.multi_agent_v2.usage_hint_enabled)
         .with_spawn_agent_usage_hint_text(per_turn_config.multi_agent_v2.usage_hint_text.clone())
         .with_hide_spawn_agent_metadata(per_turn_config.multi_agent_v2.hide_spawn_agent_metadata)
+        .with_goal_tools_enabled(!per_turn_config.ephemeral)
```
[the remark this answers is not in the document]
etraut-openai: [codex] Fixed in d85daa4572 by changing the goal-tools builder to preserve the Feature::Goals gate and only further restrict goal tools for ephemeral sessions.

codex-rs/core/src/goals.rs
```diff
@@ -0,0 +1,247 @@
...
+        let state_db = if let Some(state_db) = self.state_db() {
+            state_db
+        } else {
+            codex_state::StateRuntime::init(
+                config.sqlite_home.clone(),
+                config.model_provider_id.clone(),
+            )
+            .await
+            .context("failed to initialize sqlite state db for thread goals")?
+        };
```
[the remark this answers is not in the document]
etraut-openai: [codex] Fixed in d85daa4572 by removing the disconnected StateRuntime::init fallback. Goal operations now use the session state DB or the session LocalThreadStore DB, and non-local stores return an error instead of reading or writing an unrelated local DB.

codex-rs/core/src/goals.rs
```diff
@@ -0,0 +1,246 @@
...
+            anyhow::bail!("thread goals require a local persisted thread with a state database");
+        };
+
+        let thread_metadata_present = state_db
+            .get_thread(self.conversation_id)
+            .await
+            .context("failed to read thread metadata before reconciling thread goals")?
+            .is_some();
+        if !thread_metadata_present
+            && let Ok(Some(rollout_path)) = self.current_rollout_path().await
```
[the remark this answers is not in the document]
etraut-openai: [codex] Fixed in cbc61c8a4e by propagating current_rollout_path() errors when thread metadata is missing, so goal operations surface the persistence failure instead of continuing with an unreconciled DB.

codex-rs/core/src/goals.rs
```diff
@@ -0,0 +1,246 @@
...
+                config.model_provider_id.as_str(),
+                /*builder*/ None,
+                &[],
+                /*archived_only*/ None,
+                /*new_thread_memory_mode*/ None,
+            )
+            .await;
+        }
+
+        Ok(state_db)
```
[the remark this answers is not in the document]
etraut-openai: [codex] Fixed in cbc61c8a4e by re-reading thread metadata after reconcile_rollout and returning a clear error if reconciliation still did not materialize the thread row.

codex-rs/core/src/session/turn_context.rs
```diff
@@ -442,6 +443,7 @@ impl Session {
         .with_spawn_agent_usage_hint(per_turn_config.multi_agent_v2.usage_hint_enabled)
         .with_spawn_agent_usage_hint_text(per_turn_config.multi_agent_v2.usage_hint_text.clone())
         .with_hide_spawn_agent_metadata(per_turn_config.multi_agent_v2.hide_spawn_agent_metadata)
+        .with_goal_tools_allowed(!per_turn_config.ephemeral)
```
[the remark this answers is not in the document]
etraut-openai: [codex] Fixed in cbc61c8a4e by threading goal-tool availability into turn-context construction. The tools now stay disabled unless the session is non-ephemeral and has a state DB handle, and derived review/model contexts preserve that gate.

PR #18076, opened 2026-04-16
Add goal core runtime (4 / 5)

etraut-openai:
Adds the core runtime behavior for active goals on top of the model tools from PR 3.

## Why

A long-running goal should be a core runtime concern, not something every client has to implement. Core owns the turn lifecycle, tool completion boundaries, interruptions, resume behavior, and token usage, so it is the right place to account progress, enforce budgets, and decide when to continue work.

## What changed

- Centralized goal lifecycle side effects behind `Session::goal_runtime_apply(GoalRuntimeEvent::...)`.
- Starts goal continuation turns only when the session is idle; pending user input and mailbox work take priority.
- Accounts token and wall-clock usage at turn, tool, mutation, interrupt, and resume boundaries; `get_thread_goal` remains read-only.
- Preserves sub-second wall-clock remainder across accounting boundaries so long-running goals do not drift downward over time.
- Treats token budget exhaustion as a soft stop by marking the goal `budget_limited` and injecting wrap-up steering instead of aborting the active turn.
- Suppresses budget steering when `update_goal` marks a goal complete.
- Pauses active goals on interrupt and auto-reactivates paused goals when a thread resumes outside plan mode.
- Suppresses repeated automatic continuation when a continuation turn makes no tool calls.
- Added continuation and budget-limit prompt templates.

## Verification

- Added focused core coverage for continuation scheduling, accounting boundaries, budget-limit steering, completion accounting, interrupt pause behavior, resume auto-activation, and wall-clock remainder accounting.


codex-rs/core/src/tools/handlers/goal.rs
```diff
@@ -1,10 +1,11 @@
 //! Built-in model tool handlers for persisted thread goals.
 //!
-//! The public tool contract mirrors the app-server `thread/goal/set` overload:
```
etraut-openai: The changes in this file should be made in the previous PR in the stack, not in this PR. PR number 3 in the stack introduced the goal-related tools, so these changes belong there.

codex-rs/core/src/goals.rs
```diff
@@ -1,27 +1,190 @@
-//! Explicit get/set support for persisted thread goals.
```
etraut-openai: These comment changes should not be done in this PR. The comment should be correct in the stacked PR that introduced this file (I think it's PR 3 in the stack).

codex-rs/core/src/goals.rs
```diff
@@ -128,6 +872,68 @@ impl Session {
     }
 }
 
+fn should_ignore_goal_for_mode(mode: ModeKind) -> bool {
```
etraut-openai, line 1389: Add a comment for this function to explain its beahvior.

codex-rs/core/src/goals.rs
```diff
@@ -128,6 +872,68 @@ impl Session {
...
+Objective:
+{objective}
+
+Budget:
+- Time spent pursuing goal: {time_used_seconds} seconds
+- Tokens used: {tokens_used}
+- Token budget: {token_budget}
+- Tokens remaining: {remaining_tokens}
+
+Avoid repeating work that is already done. Choose the next concrete action toward the objective.
```
etraut-openai: I think that other core modules that use prompts place their prompts in markdown files within the "prompts" directory. In other words, they don't encode the prompt as a string literal in the code. Look at other examples in the code base and follow suit.

codex-rs/core/src/goals.rs
```diff
@@ -128,6 +872,68 @@ impl Session {
...
+- Treat uncertainty as not achieved; do more verification or continue the work.
+
+Do not rely on intent, partial progress, elapsed effort, memory of earlier work, or a plausible final answer as proof of completion. Only mark the goal achieved when the audit shows that the objective has actually been achieved and no required work remains. If any requirement is missing, incomplete, or unverified, keep working instead of marking the goal complete. If the objective is achieved, call update_goal with status "complete" so usage accounting is preserved. Report the final elapsed time, and if the achieved goal has a token budget, report the final consumed token budget to the user after update_goal succeeds.
+
+If the goal has not been achieved and cannot be achieved within the remaining budget, or the remaining budget is too small for productive continuation, call update_goal with status "budgetLimited". Do not mark a goal complete merely because the budget is nearly exhausted or because you are stopping work. If the goal is otherwise blocked and cannot continue productively for a non-budget reason, call update_goal with status "paused"."#,
```
etraut-openai: Remove the guidance about budgetLimited. We've changed our mind and decided that the model should not ever set budgetLImited. That will be something that only the runtime does when the token budget is exceeded.

codex-rs/app-server/src/codex_message_processor/thread_goal_handlers.rs
```diff
@@ -29,8 +29,9 @@ impl CodexMessageProcessor {
                 return;
             }
         };
-        let rollout_path = match self.thread_manager.get_thread(thread_id).await {
```
etraut-openai: I'm surprised to see changes in this file in this stacked PR. Shouldn't these changes have been made in PR 2 within the stack?

etraut-openai: [codex] I checked whether this hunk could move down to PR 2. The app-server API surface is introduced there, but these particular changes are runtime wiring: they call the goal continuation/abort/accounting helpers that are introduced in PR 4. Moving this patch to PR 2 would either fail to compile or require pulling the runtime behavior earlier in the stack. So I left the app-server API shape in PR 2 and kept this cross-layer runtime hook in PR 4, where the core methods exist.

codex-rs/core/src/tasks/mod.rs
```diff
@@ -242,47 +257,120 @@ impl Session {
...
         let done = Arc::new(Notify::new());
 
-        let queued_response_items = self.take_queued_response_items_for_next_turn().await;
-        let mailbox_items = self.get_pending_input().await;
-        let turn_state = {
-            let mut active = self.active_turn.lock().await;
-            let turn = active.get_or_insert_with(ActiveTurn::default);
-            debug_assert!(turn.tasks.is_empty());
-            Arc::clone(&turn.turn_state)
+        let include_next_turn_queue = matches!(
```
etraut-openai: This method has gotten pretty long and messy. See if you can come up with a way to simplify it and/or refactor it without changing its semantics.

codex-rs/core/src/tasks/mod.rs
```diff
@@ -353,6 +463,64 @@ impl Session {
...
     }
 
+    pub(crate) async fn maybe_start_turn_for_pending_work_or_goal_continuation(self: &Arc<Self>) {
+        self.clear_stale_goal_continuations_for_next_turn().await;
+        self.maybe_start_turn_for_pending_work().await;
+        let _continuation_guard = self.goal_continuation_lock.lock().await;
+        self.maybe_start_goal_continuation_if_idle().await;
+    }
+
+    pub(crate) async fn maybe_start_turn_for_active_goal_continuation(self: &Arc<Self>) {
```
etraut-openai: Add function comment.

codex-rs/core/src/tools/handlers/goal.rs
```diff
@@ -5,6 +5,7 @@
 //! on the existing goal and preserves usage accounting.
 
 use crate::function_tool::FunctionCallError;
+use crate::goals::GoalAccountingBoundary;
```
etraut-openai: I'm surprised to see changes to this file (or any files under the tools directory) in this stacked PR. I would have expected to see these in PR 3 in the stack. Review these changes and move those that belong in PR 3.

codex-rs/core/src/codex_thread.rs
```diff
@@ -80,6 +81,48 @@ impl CodexThread {
         self.codex.shutdown_and_wait().await
     }
 
+    pub async fn continue_active_goal_if_idle(&self) {
```
etraut-openai: This file involves a bunch of added complexity that I'd like to try to simplify if possible.

codex-rs/core/src/goals.rs
```diff
@@ -9,17 +9,191 @@ use crate::session::session::Session;
 use crate::session::turn_context::TurnContext;
 use anyhow::Context;
 use codex_features::Feature;
+use codex_protocol::config_types::ModeKind;
```
etraut-openai, line 15: Review the code changes in this module and look for opportunities to simplify.

codex-rs/core/src/tasks/mod.rs
```diff
@@ -384,16 +409,113 @@ impl Session {
...
-            for task in active_turn.drain_tasks() {
-                self.handle_task_abort(task, reason.clone()).await;
+        self.abort_all_tasks_without_restart(reason.clone()).await;
+        if reason == TurnAbortReason::Interrupted {
+            self.clear_queued_goal_continuations_for_next_turn().await;
+            self.maybe_start_turn_for_pending_work().await;
+        }
+    }
+
+    pub(crate) async fn abort_all_tasks_without_restart(self: &Arc<Self>, reason: TurnAbortReason) {
```
etraut-openai: Add function comment

codex-rs/core/src/tasks/mod.rs
```diff
@@ -384,16 +409,113 @@ impl Session {
...
+        }
+
+        if reason == TurnAbortReason::Interrupted
+            && let Err(err) = self.pause_active_thread_goal_for_interrupt().await
+        {
+            warn!("failed to pause active thread goal after interrupt: {err}");
+        }
+    }
+
+    pub(crate) async fn abort_all_tasks_without_goal_accounting_from_current_turn(
```
etraut-openai: Add function comment

codex-rs/core/src/tasks/mod.rs
```diff
@@ -384,16 +409,113 @@ impl Session {
...
+    ) {
+        self.abort_all_tasks_without_goal_accounting_inner(
+            reason,
+            Some(current_turn_id.to_string()),
+            GoalPauseOnInterrupt::Pause,
+        )
+        .await;
+    }
+
+    async fn abort_all_tasks_without_goal_accounting_inner(
```
etraut-openai: Add function comment

codex-rs/core/src/tasks/mod.rs
```diff
@@ -629,6 +748,61 @@ impl Session {
         });
         self.send_event(task.turn_context.as_ref(), event).await;
     }
+
+    async fn handle_current_task_abort(
```
etraut-openai: Add function comment

codex-rs/core/src/tasks/mod.rs
```diff
@@ -629,6 +748,61 @@ impl Session {
...
+            reason,
+            completed_at,
+            duration_ms,
+        });
+        self.send_event(task.turn_context.as_ref(), event).await;
+        task.handle.abort();
+    }
+}
+
+fn idle_pending_work_scheduler(session: Arc<Session>) -> BoxFuture<'static, ()> {
```
etraut-openai: Add function comment

codex-rs/core/src/tools/handlers/goal.rs
etraut-openai: [codex] I reviewed this diff against PR 3. The PR 4 changes in `core/src/tools/handlers/goal.rs` are not introducing the tools themselves; they wire the existing `update_goal` handler into PR 4 runtime accounting (`GoalAccountingBoundary::Tool`), budget-limited preservation, and stale-continuation clearing. Those helpers are introduced in PR 4, so moving this hunk down to PR 3 would either fail to compile or pull the runtime behavior earlier in the stack. I left the tool surface in PR 3 and kept only the runtime accounting behavior here.

codex-rs/core/src/codex_thread.rs
```diff
@@ -80,6 +80,64 @@ impl CodexThread {
...
+                    .maybe_start_turn_for_pending_work_or_goal_continuation()
+                    .await;
+            }
+            codex_state::ThreadGoalStatus::Paused
+            | codex_state::ThreadGoalStatus::BudgetLimited
+            | codex_state::ThreadGoalStatus::Complete => {
+                self.codex
+                    .session
+                    .clear_stopped_thread_goal_runtime_state()
+                    .await;
```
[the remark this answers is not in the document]
etraut-openai: You can ignore this.

codex-rs/core/src/goals.rs
```diff
@@ -148,6 +449,622 @@ impl Session {
...
+    pub(crate) async fn clear_cached_thread_goal_after_delete(&self) {
+        self.clear_stopped_thread_goal_runtime_state().await;
+    }
+
+    pub(crate) async fn apply_external_thread_goal_status(
+        self: &Arc<Self>,
+        status: codex_state::ThreadGoalStatus,
+    ) {
+        match status {
+            codex_state::ThreadGoalStatus::Active => {
```
jif-oai, line 563: If thread/goal/set makes a goal active while a turn is already running, this only tries to start work if we’re idle. It doesn’t attach the goal to the current TurnContext, so the rest of that turn can skip accounting/budget checks. Could we mark the activ turn contexts here too?

codex-rs/app-server/src/codex_message_processor/thread_goal_handlers.rs
```diff
@@ -73,6 +74,10 @@ impl CodexMessageProcessor {
         )
         .await;
 
+        if let Some(thread) = running_thread.as_ref() {
+            thread.prepare_external_goal_mutation().await;
```
jif-oai: This runs accounting before we validate the requested objective/budget. That means even a rejected thread/goal/set can still charge usage or flip the goal to budget_limited

codex-rs/core/src/tasks/goal_continuation.rs
```diff
@@ -0,0 +1,47 @@
...
+        .new_default_turn_with_sub_id(uuid::Uuid::new_v4().to_string())
+        .await;
+    session
+        .maybe_emit_unknown_model_warning_for_turn(turn_context.as_ref())
+        .await;
+    session
+        .mark_thread_goal_continuation_turn_started(turn_context.sub_id.clone())
+        .await;
+    session
+        .start_task(turn_context, Vec::new(), RegularTask::new())
```
jif-oai: I think there’s still a race here: after we queue the continuation prompt, the goal can be cleared/paused before start_task runs, but the stale queued input will still launch

codex-rs/core/templates/goals/continuation.md
```diff
@@ -0,0 +1,25 @@
+Continue working toward the active thread goal.
+
+Objective:
+{{ objective }}
```
jif-oai, line 6: I'm a bit concerned about this being sent as a developer message. Could we add some "UNTRUSTED" boundaries or so? Otherwise the objective can quite easily become a prompt injection source (same thing in `budget_limit.md` though

jif-oai (commented):
I feel the code is becoming more and more spread around... this makes it hard to reason about
The lifecycle side effects are spread between app-server, tasks, tool handling, resume/interrupt path etc etc. 
I'm not sure this is possible to avoid but it would be much easier to have one single place that hold all the runtime behaviour. Then we _could_ have something like this

```
session.goal_runtime_apply(GoalRuntimeEvent::ExternalSet { status, ... }).await;
session.goal_runtime_apply(GoalRuntimeEvent::ToolUpdate { status, ... }).await;
session.goal_runtime_apply(GoalRuntimeEvent::Interrupt).await;
session.goal_runtime_apply(GoalRuntimeEvent::TurnFinished { ... }).await;
...
```

and we have the actual strong implementation in there with good documentation
(I don't have strong opinion between one entry point with enum or multiple entry point but I think it's generally better to keep `Session` """"small""")

codex-rs/core/src/goals.rs
etraut-openai: [codex] Addressed in 548b072b9b. ExternalSet { Active } now reloads the persisted active goal, marks any current active turn contexts for that goal, and resets their token baseline to current usage so the rest of the in-flight turn is accounted normally without charging pre-set tokens.

codex-rs/app-server/src/codex_message_processor/thread_goal_handlers.rs
etraut-openai: [codex] Addressed in 7c84167c9e. thread/goal/set now validates the objective and token budget before calling the external goal mutation accounting hook, so rejected set requests do not charge usage or move the goal to budget_limited.

codex-rs/core/src/tasks/goal_continuation.rs
etraut-openai: [codex] Addressed in 6f01f08229. Continuation launch now builds a candidate tied to the persisted goal_id and re-reads the goal immediately before injecting the developer prompt. If the goal was cleared, paused, completed, budget-limited, or replaced, the reserved continuation is dropped instead of launching stale work.

codex-rs/core/templates/goals/continuation.md
etraut-openai: [codex] Addressed in 03295d3846. Both continuation.md and budget_limit.md now wrap the interpolated objective in <untrusted_objective> boundaries and explicitly tell the model to treat it as user-provided data, not higher-priority instructions. The rendering tests assert those boundaries are present.

etraut-openai:
[codex] For the top-level concern about goal runtime behavior being spread across app-server, tasks, tools, and resume/interrupt paths: addressed in a932e6ed59 by centralizing lifecycle side effects behind Session::goal_runtime_apply(GoalRuntimeEvent::...). The surrounding callsites now report runtime events while goals.rs owns accounting, continuation, pause/resume, budget steering, and notification side effects. The follow-up commits in this push tighten the external-set, validation-order, stale-continuation, and prompt-boundary cases.

etraut-openai:
[codex] Addressed in 52e0f2579d: turn-finish goal accounting now runs while the turn state is still reserved and before TurnComplete, and continuation launch is deferred until after the active turn is released.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -8766,6 +8771,12 @@ async fn handle_pending_thread_resume_request(
         }
     }
 
+    if pending.emit_thread_goal_update
+        && let Err(err) = conversation.apply_goal_resume_runtime_effects().await
+    {
```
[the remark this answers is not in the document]
etraut-openai: I think this is fine. We'll leave as is.

codex-rs/core/src/goals.rs
```diff
@@ -202,6 +1343,75 @@ impl Session {
...
+        .token_budget
+        .map(|budget| budget.to_string())
+        .unwrap_or_else(|| "none".to_string());
+    let remaining_tokens = goal
+        .token_budget
+        .map(|budget| (budget - goal.tokens_used).max(0).to_string())
+        .unwrap_or_else(|| "unbounded".to_string());
+    let tokens_used = goal.tokens_used.to_string();
+    let time_used_seconds = goal.time_used_seconds.to_string();
+    let objective = escape_xml_text(&goal.objective);
```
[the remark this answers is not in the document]
jif-oai: Ooc why not? Isn't it relevant?

codex-rs/core/src/goals.rs
```diff
@@ -148,6 +547,745 @@ impl Session {
...
+
+        let turn_context = self
+            .new_default_turn_with_sub_id(uuid::Uuid::new_v4().to_string())
+            .await;
+        self.mark_thread_goal_continuation_turn_started(turn_context.sub_id.clone())
+            .await;
+        self.maybe_emit_unknown_model_warning_for_turn(turn_context.as_ref())
+            .await;
+        self.start_task(turn_context, Vec::new(), RegularTask::new())
+            .await;
```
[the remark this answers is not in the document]
jif-oai: OOC why ignoring?

jif-oai (approved):

codex-rs/core/src/goals.rs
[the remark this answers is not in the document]
etraut-openai: My thinking was that we shouldn't truncate user-defined goals. If the user wants to define a long goal, we shouldn't prevent it arbitrarily trim it here.

But now that I think about it more, maybe we should enforce a max character length at the tool and api layers so an error is returned earlier. I'll make that change.

PR #18077, opened 2026-04-16
Add goal TUI UX (5 / 5)

etraut-openai:
Adds the TUI user experience for goals on top of the core runtime from PR 4.

## Why

Users need a direct TUI control surface for long-running goals. The UI should make the current goal visible, support common goal actions without waiting for a model turn, and avoid confusing end-of-turn notifications while an active goal is immediately continuing.

## What changed

- Added `/goal` summary rendering for the current goal, including active, paused, budget-limited, and complete states.
- Added `/goal <objective>` creation/replacement through the app-server goal API rather than a model prompt.
- Added `/goal clear`, `/goal pause`, and `/goal unpause` command variants.
- Added a confirmation menu when the user enters a new goal while another goal already exists.
- Updated `/goal` help and summary tip text so it reflects the supported command variants without advertising slash-command token budgets.
- Added footer/statusline goal indicators, including elapsed time and token budget display when a budget exists from API/tool-created goals.
- Consumes goal updated/cleared notifications so the TUI stays in sync with external app-server changes.
- Suppresses end-of-turn desktop notifications only when a goal is still active and follow-up work is expected.
- Preserves slash-command history behavior and avoids leaking queued `/goal` state into unrelated submissions.

## Verification

- Added TUI unit and snapshot coverage for goal command availability, summary rendering, control commands, replacement menu behavior, status/footer display, notification handling, and command history.


codex-rs/tui/src/app/app_server_adapter.rs
```diff
@@ -347,15 +347,15 @@ fn server_notification_thread_target(
         ServerNotification::ThreadNameUpdated(notification) => {
             Some(notification.thread_id.as_str())
         }
-        ServerNotification::ThreadTokenUsageUpdated(notification) => {
```
etraut-openai, line 350: The change in this file seems unnecessary. It looks like it's simply moving this code block to a new location for no good reason. If you agree, undo the change in this file.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -1043,6 +1063,18 @@ pub(crate) struct UserMessage {
     mention_bindings: Vec<MentionBinding>,
 }
 
+#[derive(Clone, Debug, PartialEq)]
```
etraut-openai, line 1116: There are a bunch of changes in this file that look like they may be unrelated to the goal TUI support. I wonder if they crept in during a merge from main? Please review all of the changes in this file and make sure that they're all in service of the TUI goal support. If not, remove them.

codex-rs/tui/src/app/app_server_adapter.rs
etraut-openai: [codex] Agreed. Reverted this ordering-only hunk in adf52598e4 so the adapter diff now only contains the goal notification handling that this PR actually needs.

codex-rs/tui/src/chatwidget.rs
etraut-openai: [codex] Audited the remaining `chatwidget.rs` diff against the #18076 base and the branch commits. I did not find merge-from-main drift. The large non-obvious part is the queued/rejected message history plumbing: `/goal ...` can submit an internal parser prompt while the user-facing history/composer text must remain the original slash command, including mentions and attachments, across queueing, rejection, restore, and interruption paths. I added comments in adf52598e4 to make that relationship explicit.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -1301,28 +1335,100 @@ fn remap_placeholders_for_message(message: UserMessage, next_label: &mut usize)
...
-        local_images: remapped_images,
+    (rebuilt, rebuilt_elements)
+}
+
+// When merging multiple queued drafts (e.g., after interrupt), each draft starts numbering
+// its attachments at [Image #1]. Reassign placeholder labels based on the attachment list so
+// the combined local_image_paths order matches the labels, even if placeholders were moved
+// in the text (e.g., [Image #2] appearing before [Image #1]). Apply the same remapping to
+// history overrides so restored drafts and rendered transcript entries agree.
+fn remap_placeholders_for_message_and_history_record(
```
fcoury-oai, line 1361: I saw that you created new focused modules for the goal behavior on ChatWidget. I wonder if there is opportunity for extracting a new one or moving these new methods to the existing ones?

codex-rs/tui/src/chatwidget.rs
```diff
@@ -7461,38 +7903,45 @@ impl ChatWidget {
...
+                    submitted_follow_up = self.submit_user_message_with_history_record(
+                        queued_message.into_user_message(),
+                        history_record,
+                    );
                     break;
                 }
                 QueuedInputAction::ParseSlash => {
                     let drain = self.submit_queued_slash_prompt(queued_message.into_user_message());
                     if drain == QueueDrain::Stop {
+                        submitted_follow_up = true;
```
fcoury-oai: Codex helped me identify a possible issue here. 

This treats “dequeued something after a turn” as equivalent to “started a follow-up model turn”, and suppresses the normal completion notification, which makes sense since we don't want to call the user attention if there are more work to be done.

This is correct for plain messages (as follow-up tasks), but I learned that you can enqueue slash commands and some of them do not start work. They may just open a menu, restore a draft, or stop for another local reason. In this case a notification makes sense.

I didn’t realize slash commands could be queued, but I verified they can be queued via Tab while a task is running. In that case, a queued `/model` can open the picker after the turn completes, while the genuine “turn complete” notification is omitted.

codex-rs/tui/src/app/thread_goal_actions.rs
```diff
@@ -0,0 +1,86 @@
...
+    ) {
+        match app_server
+            .thread_goal_set(
+                thread_id,
+                /*objective*/ None,
+                Some(status),
+                /*token_budget*/ None,
+            )
+            .await
+        {
```
fcoury-oai: Codex found an issue here as well: this action path is missing the stale-thread guard used by `open_thread_goal_menu`. If the user chooses Pause/Clear and switches threads before the app-server request returns, this success/error message is added to whichever chat is active at that point, even though the operation targeted the old `thread_id`.

When opening the menu we have protection against a stale `thread_id`:

```rust
let result = app_server.thread_goal_get(thread_id).await;
if self.current_displayed_thread_id() != Some(thread_id) {
    return;
}
```

So if the user typed bare `/goal`, then switched threads before the app-server responded, don’t open the old thread’s goal menu in the new thread. 

The same protection seems needed after `thread_goal_set(...).await` and `thread_goal_clear(thread_id).await`, before adding success/error messages to `self.chat_widget`.

fcoury-oai (changes requested):
I have tested the feature both with the happy path, where I set a open-ended goal and waited for it to be successfully achieved.

Then I also tested giving it a token budget, which clearly stated the token usage progress and also indicated the unmet goal with the expected and actual used tokens.

**EDIT:** during test I attempted to do a new `/goal` with a unmet goal being present it it was a bit hard to realize I had a goal still present. I wanted if we should add some type of notification if we attempt this.

The code looks very good, feature is a pleasure to use. 

My only concern is - as usual - about discoverability. In this case, not for the command itself, but to the available limits.

I wonder if we should have a follow-up PR in the future where we introduce a "wizard-like" picker where we hand hold the user's hand by asking the goal, asking if it wants to set a budget by token and the same by time. These are not showstoppers, just food for thought.

I am just requesting changes so you can review the new findings but I am ready to approve after your pass.

fcoury-oai:
I found another issue while testing: I resumed a thread with an ongoing goal on another terminal: the original session finished the goal and updated the statusline but the resumed session didn't:

image

Even thought the goal seems to be completed there as well:

image

If I attempt to fork the sessions instead, the goal is apparently interrupted:

image

Which is good

codex-rs/tui/src/chatwidget.rs
etraut-openai: Let's wait on this until we do the big ChatWidget cleanup (hopefully soon).

codex-rs/tui/src/chatwidget.rs
etraut-openai: Good catch. Fixed.

codex-rs/tui/src/app/thread_goal_actions.rs
etraut-openai: Yep, this is an issue too. Fixed.

etraut-openai:
Since the last review, I've made the following improvements based on feedback:
* If the user attempts `/goal <objective>` when there's an existing goal, I now display a menu: 1) Replace, 2) Cancel
* Added support for `/goal clear`, `/goal pause` and `/goal unpause`
* Added tip at bottom of `/goal` summary indicating which additional `/goal` subcommands are available
* Added suppression logic for tui.agent_turn_complete notifications when a goal is active
* Removed confusing (left over) mention of time budgets

fcoury-oai:
Sorry for the huge message, I just wanted to be very detailed on the findings I got. 

One note: when I started testing, I got an error saying a migration on my database was modified and I figured it was because we migrated it for the previous version of this PR and we modified it. Thankfully Codex fixed it and I was able to start testing. I just wanted to mention it in case you think some of the behavior below has anything to do with it, which I don't.

Below is a list of odd/unexpected behavior I noticed throughout my testing. I did `/feedback` for all of them but if you need the rollout jsonl files as well just let me know.

**1. Sub-optimal error message**

I tried to start a goal with `/goal` and right after do `/goal pause` and I got this unexpected error:

```
■ Failed to update thread goal: thread/goal/set failed in TUI
```

I was able to capture the session here:

```
• Feedback uploaded. Please report this in #codex-feedback:

  http://go/codex-feedback-internal

  Share this and add some info about your problem:
    https://go/codex-feedback/019dbf95-6421-70c1-b13d-70518d0092a6
```

I asked Codex to investigate the issue by giving it the session rollout file and I included the raw explanation below[^1] for your reference but long story short, I think if you attempt to do a `/goal pause` right after you send the goal, it's obviously not yet set, so we get the error message.

I am not worried about the issue itself, because I think it's very reasonable that we have to wait until the goal is set to pause it. What I think is suboptimal is the error message itself: it's vague and is maybe overly technical. 

I think something more user-driven would say something along the lines of: "Can't set the goal right now: either the goal was not yet set or there was an error setting it. Please try again.".

<details>
<summary>Codex analysis</summary>

• Findings

  I found the rollout file at /Users/felipe.coury/.codex/sessions/2026/04/24/rollout-2026-04-24T10-02-12-019dbf95-6421-70c1-b13d-70518d0092a6.jsonl:1.

  The main thing: the recorded session does not show /goal pause before your research prompt. history.jsonl records /goal do a barrel roll, /goal stop, /goal find top 10 largest file in this repo, /goal stop, and /goal find top 10 largest file in this repo, but no /goal pause around those turns: /Users/felipe.coury/.codex/history.jsonl:1239.

  What did fail above was the first attempt to set the “find top 10 largest file” goal. The model called create_goal, and the tool rejected it with: cannot create a new goal because this thread already has a goal; use update_goal only when the existing goal is complete

  That’s in the rollout at /Users/felipe.coury/.codex/sessions/2026/04/24/rollout-2026-04-24T10-02-12-019dbf95-6421-70c1-b13d-70518d0092a6.jsonl:57. The stored goal was already complete, so the error wording is contradictory with the behavior.

  Code-wise, /goal pause is supposed to be handled by the TUI directly, not sent to the model: tui/src/chatwidget/slash_dispatch.rs:611 recognizes only clear, pause, and unpause as goal
  control commands. If you typed /goal stop, though, that is not a control command, so it falls through to the normal “set goal objective” prompt path at tui/src/chatwidget/slash_dispatch.rs:642. That matches the rollout: /goal stop became a model-visible “Set the current thread goal...” message, and the model called update_goal {"status":"complete"}.

  One more relevant split: the model tool cannot pause goals at all. update_goal only exposes complete in its schema: tools/src/goal_tool.rs:64, and the handler rejects any non-complete status: core/src/tools/handlers/goal.rs:147. Pausing is controlled through the app-server/TUI path, not the model tool path.

  So my read is: there is no evidence in the rollout that /goal pause itself failed. The visible failure in the transcript was create_goal refusing to create a new goal after a completed goal existed, and /goal stop was treated as a new goal objective/control-by-model path rather than a TUI control command. If you saw a local /goal pause UI error, it likely happened on the app-server control path and was not captured as a model turn in this rollout file.
</details>

**2. Inconsistent state**

On a fresh session I started one goal as "find top 10 largest file in this repo". I am not sure at which moment it set the message "Goal achieved (5s)" when the goal was still going. I noticed it only when I issued `/goal do a barrel roll` while it was running to test the new menu.

Session: [go/codex-feedback/019dbf9b-9653-7a20-a3a7-3286fc9684be](https://www.golinks.io/codex-feedback/019dbf9b-9653-7a20-a3a7-3286fc9684be?trackSource=github)

**3. Menu to change goal never appeared**

Session: [go/codex-feedback/019dbfa6-e22c-72a3-bd57-335a76a1a1b8](https://www.golinks.io/codex-feedback/019dbfa6-e22c-72a3-bd57-335a76a1a1b8?trackSource=github)

In this session I started the `/goal find the 10 largest files in this repo` and sent a `/goal` right after and it properly displayed the message when you send it with no goal. Right after I sent `/goal do a barrel roll` and it started the new goal without asking me anything. Maybe all the tool calls for the previous goal were finished but its results were not yet output? Not sure.

I tried it once again, same idea but made the initial request a bit longer:

```
/goal find top 10 largest file in this repo accros all folders and then the 10 smaller files as well
```

During the execution I did:

```
/goal do a barrel roll
```

And it told me this:

• Previous file-scan goal was completed; final budget usage was 34 seconds. I tried to set the new goal, but this thread’s goal tool won’t create a second goal after one already exists.

  Barrel roll: ↻

I did one more attempt:

Session: [go/codex-feedback/019dbfb1-7d9f-7c62-9d0b-8a88d5cef0e5](https://www.golinks.io/codex-feedback/019dbfb1-7d9f-7c62-9d0b-8a88d5cef0e5?trackSource=github)

In this session I decided to provide more complex goals to make sure that I was sending a new goal before the initial one finished.

I sent was the `/goal find the rust file with the greatest cyclomatic complexity in this project` command and after I made sure it started issuing some commands for this goal, I changed it to `/goal find the 10 largest files in this repo`, which apparently it did correctly:

-----

• The objective has no explicit token budget. I could not replace the active thread goal because the goal tool only allows creating a goal when none is active, and the previous goal is still active; I continued with the latest requested objective anyway.

  Top 10 largest tracked files by git ls-files size:

  1. MODULE.bazel.lock - 1.4 MiB
  2. .github/codex-cli-splash.png - 818.5 KiB
  3. codex-rs/tui/tests/fixtures/oss-story.jsonl - 818.5 KiB
  4. codex-rs/tui/src/chatwidget.rs - 477.3 KiB
  5. codex-rs/app-server-protocol/schema/json/codex_app_server_protocol.schemas.json - 465.7 KiB
  6. codex-rs/app-server/src/codex_message_processor.rs - 421.2 KiB
  7. codex-rs/app-server-protocol/schema/json/codex_app_server_protocol.v2.schemas.json - 386.4 KiB
  8. codex-rs/app-server-protocol/src/protocol/v2.rs - 364.3 KiB
  9. codex-rs/tui/src/bottom_pane/chat_composer.rs - 352.1 KiB
  10. codex-rs/Cargo.lock - 330.4 KiB

  This excludes untracked and ignored files such as build artifacts.

---

<details>
<summary>But after this output it went back to the previous goal</summary>
• I need to focus on the original Rust file with the highest cyclomatic complexity. It’s not clear if I should repeat previous steps, but I’ve checked the tools and they seem missing. I
  wonder if I should use memory for this task since it might not be necessary. The user mentioned the project, and it's likely I need to reference related memory and document my findings. I
  could run a quick search for cyclomatic complexity and then compute the necessary details without edits.

  Considering command options

  I’m thinking about using commands with redirection or heredoc for execution. It seems I won't need escalation, but I note that the approved prefix for Python commands isn’t listed, so
  there’s some uncertainty. My goal is to compute cyclomatic complexity, ideally with rust-code-analysis if it's installed—maybe I can install it using Cargo if the network allows. I wonder
  if using lizard through pip might work too, or if I should consider cargo install rust-code-analysis-cli, even though it feels heavy. I’ll need to identify robust methods for finding the
  file with the greatest cyclomatic complexity.

  Computing cyclomatic complexity

  I'm considering how to compute "rudimentary cyclomatic complexity" for each file by counting decision points like if, else if, loops, match arms, and logical operators. I might choose to
  ignore comments and strings for accuracy. Using tree-sitter with Rust could be a good option, and I should check if the necessary packages are in node_modules. If rust-code-analysis isn’t
  installed, I could use cargo metadata and possibly the syn crate to assist with the calculations.

• I’m resuming the active goal, which is specifically the Rust cyclomatic-complexity question, not the later largest-files query. I’ll avoid changing the repo and will compute a file-level
  ranking with enough detail to identify the top candidate and audit the result.

• Explored
  └ Search cyclomatic|complexity in MEMORY.md
    Search node_modules in .
    Search Cargo.toml in .

• Ran cargo search rust-code-analysis --limit 5
  └     Blocking waiting for file lock on package cache
    rust-code-analysis = "0.0.25"        # Tool to compute and export code metrics
    … +4 lines (ctrl + t to view transcript)
    ... and 1 crates more (use --limit N to see more)
    note: to learn more about a package, run `cargo info <name>`

• Ran cargo search lizard --limit 5
  └ curve25519-dalek = "5.0.0-pre.6"    # A pure-Rust implementation of group operations on ristretto255 and Curve25519
    wikibase_rest_api = "0.1.14"        # A Rust client for the Wikibase REST API.
    … +3 lines (ctrl + t to view transcript)
    ... and 7 crates more (use --limit N to see more)
    note: to learn more about a package, run `cargo info <name>`

• I’m considering whether I can install rust-code-analysis-cli using the existing cargo cache. It seems that the installation tries to write to /.cargo, but that’s not writable. I’m thinking
  about using /tmp/rustca as an alternative root and whether I can get approval for that. However, cargo usually accesses the cache in the home directory. I could try running the command, but
  I may need extra permissions if it encounters issues. Trying with a longer timeout could help too.

───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

• There is a published rust-code-analysis-cli crate, so I’m going to try the purpose-built analyzer first and only fall back to an approximate local counter if installing or running it is
  impractical in this sandbox.

• Ran cargo install --root /tmp/codex-rust-code-analysis rust-code-analysis-cli --version 0.0.25
  └  Downloading crates ...
      Downloaded rust-code-analysis-cli v0.0.25
    … +148 lines (ctrl + t to view transcript)
    error: failed to compile `rust-code-analysis-cli v0.0.25`, intermediate artifacts can be found at `/var/folders/_h/r4j8gjq13x96wcp0cgsj1y640000gn/T/cargo-installeYvnBJ`.
    To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.

• I need to figure out if I should install something that’s locked. It sounds like there could be a command I can use: “cargo install --locked.” This makes me think about the importance of
  proper installation practices to avoid issues. I wonder if I should check my current setup or if there are specific dependencies I need to be cautious about. Let's ensure everything's in
  place before proceeding!

───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

• The current rust-code-analysis-cli install failed because Cargo resolved incompatible tree-sitter versions. I’m retrying with the crate’s lockfile constraints, which is the right way to use
  older CLI crates when their dependency graph has drifted.

• Ran cargo install --root /tmp/codex-rust-code-analysis rust-code-analysis-cli --version 0.0.25 --locked
  └   Installing rust-code-analysis-cli v0.0.25
        Updating crates.io index
    … +143 lines (ctrl + t to view transcript)
       Installed package `rust-code-analysis-cli v0.0.25` (executable `rust-code-analysis-cli`)
    warning: be sure to add `/tmp/codex-rust-code-analysis/bin` to your PATH to be able to run the installed binaries
</details>

I am not sure this is the desired behavior? Should it replace the goal or just queue the current one and resume it later? Maybe we need to make our prompt clearer in this aspect?

Finally at the end of this I ended up pressing Esc to cancel the execution and I got the message "Goal paused (/goal to continue)" in the footer. Then I thought of running `/goal do a barrel roll` again here just to see if the menu to choose from replace or cancel but it never did, it just gave me this output:

---
• The new objective is do a barrel roll, with no explicit token budget. I’ll try to set it as the thread goal; if the goal API refuses because the previous goal is still active, I’ll still
  satisfy the requested objective directly.

• I need to determine if I can satisfy the user's request directly. There's a possibility of using "↻", but the user set a specific goal. It appears that I might not be able to achieve this
  with the tool I have. I could try to do a "barrel roll," but that isn't the current goal. If I can't meet the objective due to API constraints, I'll need to mention that. I want to keep
  everything clear and concise.

• Couldn’t replace the active thread goal because the goal tool only allows creating a goal when none is active.

↻ barrel roll complete.
---

At this point, since the goal was being displayed as paused, I did a `/goal unpause` and that worked perfectly.

fcoury-oai:
OK I played with it a lot more and it was way more controlled and I found almost no oddities. Attempts to change the goal were all presented with the menu.

Our users will be delighted with this feature.

Some fun attempts during my testing:

_start a goal of finding the 10 files with the greatest cyclomatic complexity in this project, give it a small amount of tokens_ → 2,500 tokens (couldn't finish, obviously)
_medium amount_ → 30,000 tokens (near but didn't finish either)
_large amount_ -> 250,000 tokens 

With the large budget it completed the task with 53.2K tokens. You'll be surprised with the #1 on the list:

---
Goal complete. Budget usage: 53,151 of 200,000 tokens; time used: 122 seconds.

I ranked tracked Rust files by rust-code-analysis-cli’s file-level metrics.cyclomatic.sum. The analyzer produced metrics for 1,637 files, matching git ls-files "*.rs" exactly.

| Rank | File | Cyclomatic Sum | Max | Avg | PLOC |
|---:|---|---:|---:|---:|---:|
| 1 | tui/src/chatwidget.rs:1 | 2032 | 118 | 2.47 | 10722 |
| 2 | app-server/src/codex_message_processor.rs:1 | 1487 | 79 | 3.74 | 10710 |
| 3 | tui/src/bottom_pane/chat_composer.rs:1 | 1052 | 47 | 2.29 | 7646 |
| 4 | app-server-protocol/src/export.rs:1 | 736 | 58 | 3.35 | 2523 |
| 5 | app-server/tests/suite/v2/turn_start.rs:1 | 735 | 47 | 12.67 | 3225 |
| 6 | core/src/config/config_tests.rs:1 | 709 | 13 | 3.44 | 6361 |
| 7 | app-server-protocol/src/protocol/v2.rs:1 | 658 | 16 | 2.01 | 9073 |
| 8 | app-server/tests/suite/v2/thread_resume.rs:1 | 624 | 39 | 16.00 | 2552 |
| 9 | tui/src/history_cell.rs:1 | 593 | 25 | 1.82 | 4281 |
| 10 | protocol/src/protocol.rs:1 | 534 | 37 | 2.35 | 4270 |

Metric output is in `/tmp/codex-complexity.VOeYpa`. Working tree is clean; no code changes or tests were run.   

---

:shipit:

fcoury-oai (approved):

codex-rs/tui/src/app/thread_goal_actions.rs
```diff
@@ -0,0 +1,184 @@
...
+            }
+        }
+
+        let result = app_server
+            .thread_goal_set(
+                thread_id,
+                Some(objective),
+                Some(ThreadGoalStatus::Active),
+                Some(None),
+            )
```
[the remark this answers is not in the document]
etraut-openai: Fixed.

codex-rs/tui/src/chatwidget.rs
```diff
@@ -9851,6 +10316,15 @@ impl ChatWidget {
             self.sync_plugins_command_enabled();
             self.refresh_plugin_mentions();
         }
+        if feature == Feature::Goals {
+            self.sync_goal_command_enabled();
+            if !enabled {
+                self.current_goal_status_indicator = None;
+                self.current_goal_status = None;
+                self.goal_status_active_turn_started_at = None;
+                self.update_collaboration_mode_indicator();
```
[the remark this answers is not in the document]
etraut-openai: [codex] Fixed in e9dfe42579 by clearing budget_limited_turn_ids when Feature::Goals is disabled, alongside the visible goal status state.

PR #18096, opened 2026-04-16
feat(sandbox): add glob deny-read platform enforcement

viyatb-oai:
## Summary
- adds macOS Seatbelt deny rules for unreadable glob patterns
- expands unreadable glob matches on Linux and masks them in bwrap, including canonical symlink targets
- keeps Linux glob expansion robust when `rg` is unavailable in minimal or Bazel test environments
- adds sandbox integration coverage that runs `shell` and `exec_command` with a `**/*.env = none` policy and verifies the secret contents do not reach the model

## Linux glob expansion

```text
Prefer:   rg --files --hidden --no-ignore --glob <pattern> -- <search-root>
Fallback: internal globset walker when rg is not installed
Failure:  any other rg failure aborts sandbox construction
```

```
[permissions.workspace.filesystem]
glob_scan_max_depth = 2

[permissions.workspace.filesystem.":project_roots"]
"**/*.env" = "none"
```

This keeps the common path fast without making sandbox construction depend on an ambient `rg` binary. If `rg` is present but fails for another reason, the sandbox setup fails closed instead of silently omitting deny-read masks.

## Platform support
- macOS: subprocess sandbox enforcement is handled by Seatbelt regex deny rules
- Linux: subprocess sandbox enforcement is handled by expanding existing glob matches and masking them in bwrap
- Windows: policy/config/direct-tool glob support is already on `main` from #15979; Windows subprocess sandbox paths continue to fail closed when unreadable split filesystem carveouts require runtime enforcement, rather than silently running unsandboxed

## Stack
1. #15979 - merged: cross-platform glob deny-read policy/config/direct-tool support for macOS, Linux, and Windows
2. This PR - macOS/Linux subprocess sandbox enforcement plus Windows fail-closed clarification
3. #17740 - managed deny-read requirements

## Verification
- Added integration coverage for `shell` and `exec_command` glob deny-read enforcement
- `cargo check -p codex-sandboxing -p codex-linux-sandbox --tests`
- `cargo check -p codex-core --test all`
- `cargo clippy -p codex-linux-sandbox -p codex-sandboxing --tests`
- `just bazel-lock-check`


codex-rs/core/src/tools/handlers/unified_exec.rs
```diff
@@ -330,11 +332,25 @@ impl ToolHandler for UnifiedExecHandler {
                         &context,
                     )
                     .await
-                    .map_err(|err| {
-                        FunctionCallError::RespondToModel(format!(
+                {
+                    Ok(response) => response,
+                    Err(UnifiedExecError::SandboxDenied { output, .. }) => ExecCommandToolOutput {
+                        event_call_id: context.call_id.clone(),
+                        chunk_id: String::new(),
```
bolinfest: Is empty string OK here?

codex-rs/core/src/tools/handlers/unified_exec.rs
```diff
@@ -330,11 +332,25 @@ impl ToolHandler for UnifiedExecHandler {
...
+                        wall_time: output.duration,
+                        raw_output: output.aggregated_output.text.into_bytes(),
+                        max_output_tokens,
+                        process_id: None,
```
bolinfest: Is `None` OK here?

codex-rs/core/src/tools/handlers/unified_exec.rs
```diff
@@ -330,11 +332,25 @@ impl ToolHandler for UnifiedExecHandler {
...
+                        exit_code: Some(output.exit_code),
+                        original_token_count: None,
```
bolinfest: Perhaps this `None` doesn't really matter?

codex-rs/core/tests/suite/tools.rs
```diff
@@ -448,6 +454,109 @@ async fn sandbox_denied_shell_returns_original_output() -> Result<()> {
...
+        "/bin/sh".to_string(),
+        "-c".to_string(),
+        "status=0; cat \"$1\" || status=$?; cat \"$2\"; exit \"$status\"".to_string(),
+        "sh".to_string(),
+        denied_path.to_string_lossy().into_owned(),
+        allowed_path.to_string_lossy().into_owned(),
+    ];
+    let args = json!({
+        "command": command,
+        "timeout_ms": 1_000,
```
bolinfest, line 503: Might as well make this `5_000`?

codex-rs/core/tests/suite/unified_exec.rs
```diff
@@ -2341,6 +2346,127 @@ async fn unified_exec_runs_under_sandbox() -> Result<()> {
     Ok(())
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn unified_exec_enforces_glob_deny_read_policy() -> Result<()> {
+    skip_if_no_network!(Ok(()));
+    skip_if_sandbox!(Ok(()));
+    skip_if_windows!(Ok(()));
```
bolinfest: Should we `#[cfg(unix)]` on the test instead?

codex-rs/core/tests/suite/unified_exec.rs
```diff
@@ -2341,6 +2346,127 @@ async fn unified_exec_runs_under_sandbox() -> Result<()> {
...
+    fs::write(&denied_path, format!("{secret}\n")).context("write denied fixture")?;
+    fs::write(&allowed_path, format!("{allowed}\n")).context("write allowed fixture")?;
+
+    let call_id = "uexec-glob-deny-read";
+    let cmd = format!(
+        "read_status=0; cat {denied_path:?} || read_status=$?; cat {allowed_path:?}; exit $read_status"
+    );
+    let args = serde_json::json!({
+        "cmd": cmd,
+        "yield_time_ms": 500,
```
bolinfest: Again, I would just do `5_000`: I don't want this test to flake because the timeout is too small.

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -99,7 +117,9 @@ pub(crate) fn create_bwrap_command_args(
     command_cwd: &Path,
     options: BwrapOptions,
 ) -> Result<BwrapArgs> {
-    if file_system_sandbox_policy.has_full_disk_write_access() {
+    let unreadable_globs =
+        file_system_sandbox_policy.get_unreadable_globs_with_cwd(sandbox_policy_cwd);
+    if file_system_sandbox_policy.has_full_disk_write_access() && unreadable_globs.is_empty() {
```
bolinfest, line 124: Should it be possible for `has_full_disk_write_access()` to be true if `unreadable_globs` is non-empty? I guess that sort of makes sense?

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -414,6 +460,243 @@ fn create_filesystem_args(
     })
 }
 
+fn expand_unreadable_globs_with_ripgrep(
+    patterns: &[String],
+    cwd: &Path,
+    max_depth: Option<usize>,
+) -> Result<Vec<PathBuf>> {
```
bolinfest: Why work with `PathBuf` instead of `AbsolutePathBuf`?

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -1540,4 +1919,97 @@ mod tests {
...
+            args.windows(5).any(|window| {
+                window[0] == "--perms"
+                    && window[1] == "000"
+                    && window[2] == "--ro-bind-data"
+                    && window[4] == path
+            }),
+            "expected file mask for {path}: {args:#?}"
+        );
+    }
 }
```
bolinfest, line 2021: Is `max_depth` configurable by the user in this PR? I think the answer is "no"?

I think we need to give folks with smaller repos some way to address this. Or maybe if they want a max depth of 2, it is possible (albeit tedious) to do the following:

```
":project_roots" = {
  "*.env": "none",
  "*/*.env": "none",
  "*/*/*.env": "none",
}
```

we should probably make the depth of `**` on non-MacOS configurable in `config.toml` and if it is set, we shouldn't display that diagnostic warning?

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -414,6 +460,243 @@ fn create_filesystem_args(
...
+        let relative = path.strip_prefix(search_root).unwrap_or(path.as_path());
+
+        if (file_type.is_file() || file_type.is_symlink()) && glob_set.is_match(relative) {
+            paths.push(path.clone());
+        }
+
+        if !file_type.is_dir() {
+            continue;
+        }
+        if max_depth.is_some_and(|max_depth| relative.components().count() >= max_depth) {
```
bolinfest: I'm a little surprise we don't use `max_depth - 1` to recurse so we don't have to count the components each time?

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -625,6 +911,9 @@ fn canonical_target_for_symlink_in_path(
     target_path: &Path,
     allowed_write_paths: &[PathBuf],
 ) -> Option<PathBuf> {
+    // Only follow symlinks that are themselves under writable roots. Those are
```
bolinfest: Does this open up a TOCTTOU issue where the target of the symlink could change after this point?

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -1540,4 +1919,97 @@ mod tests {
...
+    }
+
+    fn ripgrep_available() -> bool {
+        Command::new("rg")
+            .arg("--version")
+            .output()
+            .is_ok_and(|output| output.status.success())
+    }
+
+    fn assert_file_masked(args: &[String], path: &Path) {
```
bolinfest, line 2009: Docstring like:

```rust
/// Assert the `path` is masked due to the presence of the following pattern
/// in the args to bwrap:
///
/// bwrap ... --perms 000 --ro-bind-data PATH
```

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -755,6 +1066,39 @@ mod tests {
...
+        let args = create_bwrap_command_args(
+            command.clone(),
+            &policy,
+            temp_dir.path(),
+            temp_dir.path(),
+            BwrapOptions::default(),
+        )
+        .expect("create bwrap args");
+
+        assert_ne!(args.args, command);
```
bolinfest: Assertion message?

codex-rs/linux-sandbox/src/bwrap.rs
bolinfest, line 997: In a follow-up, we should also migrate this crate to use the `_tests.rs` pattern.

codex-rs/sandboxing/src/seatbelt_tests.rs
```diff
@@ -257,6 +259,76 @@ fn explicit_unreadable_paths_are_excluded_from_readable_roots() {
     );
 }
 
+#[test]
+fn unreadable_globstar_slash_matches_zero_or_more_directories() {
+    let regex =
```
bolinfest: I feel like we should `assert_eq!()` on the `Option<String>` returned by each `seatbelt_regex_for_unreadable_glob()` here in the tests.

In practice, it doesn't matter what `regex_lite` thinks of the pattern but whether Seatbelt honors it as intended, right?

I suppose that is verified in things like `tests/suite/unified_exec.rs`.

bolinfest (approved):

codex-rs/linux-sandbox/src/bwrap.rs
viyatb-oai: yep, fixed it. thanks for catching this!

PR #18097, opened 2026-04-16
defer all tools behind feature flag

sayan-oai:
add new `tool_search_always_defer_mcp_tools` feature flag that always defers all mcp tools rather than deferring once > 100 deferrable tools.

add new tests, also move `mcp_exposure` tests into dedicated file rather than polluting `codex_tests`.

### DIFF
--- a/codex-rs/core/src/mcp_tool_exposure.rs
+++ b/codex-rs/core/src/mcp_tool_exposure.rs
@@ -33,21 +33,11 @@ pub(crate) fn build_mcp_tool_exposure(
         ));
     }
 
-    if !tools_config.search_tool {
-        return McpToolExposure {
-            direct_tools: deferred_tools,
-            deferred_tools: None,
-        };
-    }
-
-    let should_defer = if config
-        .features
-        .enabled(Feature::ToolSearchAlwaysDeferMcpTools)
-    {
-        true
-    } else {
-        deferred_tools.len() >= DIRECT_MCP_TOOL_EXPOSURE_THRESHOLD
-    };
+    let should_defer = tools_config.search_tool
+        && (config
+            .features
+            .enabled(Feature::ToolSearchAlwaysDeferMcpTools)
+            || deferred_tools.len() >= DIRECT_MCP_TOOL_EXPOSURE_THRESHOLD);
 
     if !should_defer {
         return McpToolExposure {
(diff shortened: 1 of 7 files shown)


codex-rs/core/tests/suite/search_tool.rs
```diff
@@ -189,6 +189,56 @@ async fn search_tool_enabled_by_default_adds_tool_search() -> Result<()> {
...
+    test.submit_turn_with_policies(
+        "list tools",
+        AskForApproval::Never,
+        SandboxPolicy::DangerFullAccess,
+    )
+    .await?;
+
+    let body = mock.single_request().body_json();
+    let tools = tool_names(&body);
+    assert!(
```
pakrym-oai, line 226: assert no tools start with mcp?

codex-rs/core/tests/suite/search_tool.rs
```diff
@@ -189,6 +189,56 @@ async fn search_tool_enabled_by_default_adds_tool_search() -> Result<()> {
...
+            ev_completed("resp-1"),
+        ]),
+    )
+    .await;
+
+    let mut builder =
+        configured_builder(apps_server.chatgpt_base_url.clone()).with_config(|config| {
+            config
+                .features
+                .enable(Feature::ToolSearchAlwaysDeferMcpTools)
```
pakrym-oai, line 212: enable apps and connectors and plugins?

codex-rs/core/src/mcp_tool_exposure.rs
```diff
@@ -32,7 +33,23 @@ pub(crate) fn build_mcp_tool_exposure(
         ));
     }
 
-    if !tools_config.search_tool || deferred_tools.len() < DIRECT_MCP_TOOL_EXPOSURE_THRESHOLD {
+    if !tools_config.search_tool {
```
pakrym-oai: can you collapse this return into condition below?

pakrym-oai (approved):

codex-rs/core/src/mcp_tool_exposure.rs
sayan-oai: good catch, ty

codex-rs/core/tests/suite/search_tool.rs
sayan-oai: apps already enabled by `configured_builder`. plugins arent needed if apps are enabled, and theyre enabled by default now anyways

PR #18102, opened 2026-04-16
Stabilize realtime startup context retention under truncation

ddr-oai:
Disclaimer: Codex-generated, human-reviewed.  The principle makes sense, but I don't know if it's missing an important aspect of the context truncation here.

- Stop asserting on a markdown subheading marker that is not part of the behavior contract
- Preserve realtime startup context sections in order when the final prompt has to be truncated
- Keep the startup context blob within its token budget with a targeted unit test


alexsong-oai (approved):

ddr-oai:
fixed with another method

PR #18132, opened 2026-04-16
key mcp tool listings by toolname

sayan-oai:
follow-up to #17404, now rebased directly on `main`.

key MCP tool maps by `ToolName` instead of flat strings.

this keeps MCP callable-name uniqueness inside the MCP layer, so callers can use typed tool identities directly.

### DIFF
--- a/codex-rs/core/src/tools/spec.rs
+++ b/codex-rs/core/src/tools/spec.rs
@@ -41,12 +41,12 @@ struct McpToolPlanInputs<'a> {
     tool_namespaces: HashMap<String, ToolNamespace>,
 }
 
-fn map_mcp_tools_for_plan(mcp_tools: &HashMap<String, ToolInfo>) -> McpToolPlanInputs<'_> {
+fn map_mcp_tools_for_plan(mcp_tools: &HashMap<ToolName, ToolInfo>) -> McpToolPlanInputs<'_> {
     McpToolPlanInputs {
         mcp_tools: mcp_tools
-            .values()
-            .map(|tool| ToolRegistryPlanMcpTool {
-                name: tool.canonical_tool_name(),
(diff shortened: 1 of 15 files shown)


codex-rs/core/src/unavailable_tool.rs
```diff
@@ -6,7 +6,7 @@ use codex_tools::ToolName;
 
 pub(crate) fn collect_unavailable_called_tools(
     input: &[ResponseItem],
-    exposed_tool_names: &HashSet<&str>,
+    exposed_tool_names: &HashSet<String>,
```
pakrym-oai, line 9: shouldn't this be taking ToolName?

codex-rs/core/src/tools/spec.rs
```diff
@@ -113,9 +113,9 @@ pub(crate) fn build_specs_with_discoverable_tools(
     let mcp_tool_plan_inputs = mcp_tools.as_ref().map(map_mcp_tools_for_plan);
     let deferred_mcp_tool_sources = deferred_mcp_tools.as_ref().map(|tools| {
         tools
-            .values()
-            .map(|tool| ToolRegistryPlanDeferredTool {
-                name: tool.canonical_tool_name(),
+            .iter()
+            .map(|(name, tool)| ToolRegistryPlanDeferredTool {
+                name: name.clone(),
```
pakrym-oai, line 118: canonical_tool_name still exists and is being used, so should we just use the ToolInfo without the hash/

pakrym-oai (approved):

pakrym-oai (approved):
This is step forward but we are carrying duplicated information. ToolName as the key + canonical_tool_name on ToolInfo which should be exactly the same thing.

PR #18153, opened 2026-04-16
[codex] Add OpenAI file support for builtin codex_apps tools

lt-oai:
## Summary
https://github.com/openai/openai/pull/833843

Add support for builtin `codex_apps` tools (e.g., ChatGPT Library tool) that use OpenAI file upload and download semantics.

- Uploads path follows the same approach as https://github.com/openai/codex/pull/15197 -- local file-path arguments for Library tools declared with `openai/fileParams` metadata, then rewrites them into OpenAI file references before the MCP tool call.
  - Skips the deprecated `/files/{file_id}/uploaded finalize` call for library (`store_in_library=true`) uploads.
- metadata-driven "builtin" `codex_apps` support
- Download path materializes (metadata-driven) Library download results back to local files so downstream tools can reuse them.
- Adds `openai_file_api_base_url` for local/dev, since `chatgpt_base_url` points at local codex-backend while the file API lives on sa-server. (Note: in staging/prod, both `codex-backend` and `sa-server` has the same base URL)

## High-level approach

- Treat some `codex_apps` tools as first-party builtin tools, identified by backend-provided `_codex_apps` metadata.
- Use that metadata to:
  - expose builtin tools directly without connector selection
  - optionally materialize file-download results locally
- Reuse the existing OpenAI file upload path for tool arguments declared through `openai/fileParams`.
- Honor backend-declared `openai/fileUploadConfig` for those uploads.
- Use `openai_file_api_base_url` in local/dev, since the file API can live separately from `chatgpt_base_url`.
- For builtin tools marked for download materialization:
  - fetch the file after the tool call with ChatGPT auth
  - write a local artifact under `CODEX_HOME/.tmp/codex_apps_downloads/...`
  - return `local_path` in the tool result
- For `store_in_library=true` uploads, skip the deprecated `/files/{file_id}/uploaded` finalize call.

## Why this shape

- Keeps everything inside the existing `codex_apps` MCP surface instead of introducing a separate MCP server.
- Makes the behavior explicit and opt-in through backend metadata rather than hardcoded tool-name checks in core exposure logic.
- Keeps upload support generic on the input side, while keeping download materialization narrowly scoped to marked builtin `codex_apps` tools.
- Uses a separate file API base in local/dev because `chatgpt_base_url` can point at local `codex-backend` while the file API lives elsewhere.
- The initial backend use case is ChatGPT Library.

## Validation
- local test


lt-oai:

PR #18159, opened 2026-04-16
[codex] preserve nested mcp object schemas

soheil-oai:
## Issue
This fixes a schema rendering mismatch for MCP tools in Codex. In the Outlook Calendar `create_event` tool, nested object parameters such as `start` and `end` could be shown to the model as `string` even though the live runtime validator requires objects shaped like `{ dateTime: string, timeZone: string }`.

## Root cause
The breakage was in Codex schema normalization rather than the connector or Responses API layers. `parse_tool_input_schema` sanitizes MCP input schemas into Codex's reduced JSON Schema representation before those schemas are rendered into tool definitions. During that sanitization, object-shaped nodes that arrived through local `$ref` indirections or single-variant `allOf`/`oneOf`/`anyOf` wrappers were not preserved. Once those wrappers were ignored, the fallback inference path collapsed the node to a primitive type, which is how nested object params ended up displayed as `string`.

## Fix
The sanitizer now resolves local `$ref` pointers and unwraps single-variant `allOf`/`oneOf`/`anyOf` nodes before inferring fallback types. That preserves nested object properties for the model-visible tool schema while keeping the existing normalization logic for other schema shapes.

## Validation
I added regression tests covering both object shapes that come through local `$ref` and object shapes wrapped in a single-entry `allOf`. I also reran the crate tests after rebasing onto the latest `origin/main`.

`cargo +1.93.1-aarch64-apple-darwin test -p codex-tools`

### DIFF
--- a/codex-rs/tools/src/json_schema.rs
+++ b/codex-rs/tools/src/json_schema.rs
@@ -147,8 +147,9 @@ impl From<JsonSchema> for AdditionalProperties {
 
 /// Parse the tool `input_schema` or return an error for invalid schema.
 pub fn parse_tool_input_schema(input_schema: &JsonValue) -> Result<JsonSchema, serde_json::Error> {
-    let mut input_schema = input_schema.clone();
-    sanitize_json_schema(&mut input_schema);
+    let root_schema = input_schema.clone();
+    let mut input_schema = root_schema.clone();
+    sanitize_json_schema(&mut input_schema, &root_schema);
     let schema: JsonSchema = serde_json::from_value(input_schema)?;
     if matches!(
         schema.schema_type,
@@ -162,11 +163,13 @@ pub fn parse_tool_input_schema(input_schema: &JsonValue) -> Result<JsonSchema, s
 /// Sanitize a JSON Schema (as serde_json::Value) so it can fit our limited
 /// schema representation. This function:
 /// - Ensures every typed schema object has a `"type"` when required.
+/// - Resolves local `$ref` indirections and unwraps single-variant
+///   `oneOf`/`anyOf`/`allOf` wrappers before inferring a fallback type.
 /// - Preserves explicit `anyOf`.
 /// - Collapses `const` into single-value `enum`.
 /// - Fills required child fields for object/array schema types, including
 ///   nullable unions, with permissive defaults when absent.
-fn sanitize_json_schema(value: &mut JsonValue) {
+fn sanitize_json_schema(value: &mut JsonValue, root_schema: &JsonValue) {
     match value {
         JsonValue::Bool(_) => {
             // JSON Schema boolean form: true/false. Coerce to an accept-all string.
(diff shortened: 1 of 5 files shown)


soheil-oai:
@codex pls review

soheil-oai:
@codex pls review

soheil-oai:
@codex pls review

codex-rs/tools/src/json_schema.rs
```diff
@@ -162,42 +163,76 @@ pub fn parse_tool_input_schema(input_schema: &JsonValue) -> Result<JsonSchema, s
...
+/// - Resolves local `$ref` indirections and unwraps single-variant
+///   `oneOf`/`anyOf`/`allOf` wrappers before inferring a fallback type.
 /// - Preserves explicit `anyOf`.
 /// - Collapses `const` into single-value `enum`.
 /// - Fills required child fields for object/array schema types, including
 ///   nullable unions, with permissive defaults when absent.
-fn sanitize_json_schema(value: &mut JsonValue) {
+fn sanitize_json_schema(
+    value: &mut JsonValue,
+    root_schema: &JsonValue,
```
bolinfest, line 182: Can you update the docstring to explain these params? Is `value` (which should maybe be renamed?) the thing being sanitized whereas `root_schema` is provided "for reference"?

codex-rs/tools/src/json_schema.rs
```diff
@@ -234,11 +277,271 @@ fn sanitize_json_schema(value: &mut JsonValue) {
 
             write_schema_types(map, &schema_types);
             ensure_default_children_for_schema_types(map, &schema_types);
+
+            active_refs.truncate(active_ref_depth);
```
bolinfest, line 281: What is this doing?

codex-rs/tools/src/json_schema.rs
```diff
@@ -159,45 +160,87 @@ pub fn parse_tool_input_schema(input_schema: &JsonValue) -> Result<JsonSchema, s
...
+            for array_value in values {
+                sanitize_json_schema(array_value, root_schema, active_refs);
             }
         }
         JsonValue::Object(map) => {
+            let active_ref_depth = active_refs.len();
+            if let Some(pointer) = map
+                .get("$ref")
+                .and_then(JsonValue::as_str)
+                .and_then(|reference| reference.strip_prefix('#'))
```
bolinfest, line 200: This is repeated in `resolve_json_schema_reference()`, so it would be nice if we only did it once and then `resolve_json_schema_reference()` would have fewer cases to deal with?

codex-rs/tools/src/json_schema.rs
```diff
@@ -234,11 +277,271 @@ fn sanitize_json_schema(value: &mut JsonValue) {
...
+    if matches!(replacement, JsonValue::Bool(true)) {
+        replacement = JsonValue::Object(serde_json::Map::new());
+    }
+    if let JsonValue::Object(replacement_map) = &mut replacement {
+        merge_json_schema_objects(replacement_map, map);
+    }
+    Some(replacement)
+}
+
+fn cyclic_reference_fallback(
```
bolinfest, line 300: Docstring would be helpful here.

PR #18164, opened 2026-04-16
More Bazel build timing mitigations

ddr-oai:
Address test-side flakes caused by slower CI runners racing background persistence or app-server startup.
- Increase the SQLite persistence poll timeout in `resume_thread_subagent_restores_stored_nickname_and_role`
- Increase the app-server initialize/read timeout for the conversation summary test on slower macOS and Windows Bazel runners
- More and more similar timing fixes


ddr-oai:
working with @bolinfest to address actual test issues instead

PR #18173, opened 2026-04-16
Add explicit thread environment selection

starr-openai:
## Summary
- add ordered thread/turn environment selection for app-server callers, with required absolute cwd per environment
- keep `EnvironmentManager` owned by `ThreadManagerState` and resolve protocol params into `SessionEnvironment` before session state sees them
- collapse the prior dual environment carriers so core passes one resolved `SessionEnvironment { environment_id, cwd, environment }` list through spawn/session/turn state
- remove spurious `codex.rs` rebase churn and keep the modular main layout

## Validation
- `git diff --check`
- `rustfmt` run directly on touched Rust files because local `cargo` is guarded in this shell
- Not run: tests/builds for the latest push, per request to update the PR first while review continues


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2538,6 +2548,8 @@ impl CodexMessageProcessor {
                 persist_extended_history,
                 service_name,
                 request_trace,
+                environment_id,
+                requested_cwd,
```
starr-openai: why do we pass this here now?

codex-rs/app-server/tests/suite/v2/thread_start.rs
```diff
@@ -45,6 +51,9 @@ use super::analytics::thread_initialized_event;
 use super::analytics::wait_for_analytics_payload;
 
 const DEFAULT_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(10);
+const CODEX_EXEC_SERVER_URL_ENV_VAR: &str = "CODEX_EXEC_SERVER_URL";
+const CODEX_EXEC_SERVER_CWD_ENV_VAR: &str = "CODEX_EXEC_SERVER_CWD";
+const EXEC_SERVER_CONNECT_RETRY_INTERVAL: Duration = Duration::from_millis(25);
```
starr-openai: can we write the test to avoid needing this?

codex-rs/app-server/tests/suite/v2/thread_start.rs
```diff
@@ -222,6 +422,64 @@ fn normalize_path_for_comparison(path: impl AsRef<Path>) -> PathBuf {
...
 }
 
+struct TestExecServer {
+    websocket_url: String,
+    task: JoinHandle<()>,
+}
+
+impl TestExecServer {
+    async fn spawn() -> Result<Self> {
+        let bind_addr = TcpListener::bind("127.0.0.1:0")?.local_addr()?;
```
starr-openai: can we just use an in-memory transport so we dont need to bind/connect over tcp for unit tests?

codex-rs/app-server/tests/suite/v2/thread_start.rs
```diff
@@ -222,6 +422,64 @@ fn normalize_path_for_comparison(path: impl AsRef<Path>) -> PathBuf {
...
+}
+
+impl Drop for TestExecServer {
+    fn drop(&mut self) {
+        self.task.abort();
+    }
+}
+
+async fn wait_for_exec_server(websocket_url: &str) -> Result<()> {
+    let deadline = Instant::now() + DEFAULT_READ_TIMEOUT;
```
starr-openai: come up with a better solution that is more deterministic, e.g. use an in-memory exec server / signal with events

codex-rs/core/src/codex.rs
```diff
@@ -430,6 +430,8 @@ pub(crate) struct CodexSpawnArgs {
     pub(crate) auth_manager: Arc<AuthManager>,
     pub(crate) models_manager: Arc<ModelsManager>,
     pub(crate) environment_manager: Arc<EnvironmentManager>,
+    pub(crate) environment_id: Option<String>,
+    pub(crate) requested_cwd: Option<PathBuf>,
```
starr-openai: this should just be a property of the environment for now, pulled from the env var and only set in the environment.rs code

codex-rs/core/src/codex.rs
```diff
@@ -503,9 +507,23 @@ impl Codex {
...
+                .and_then(|environment| environment.default_cwd())
+        {
+            config.cwd =
+                codex_utils_absolute_path::AbsolutePathBuf::from_absolute_path(default_cwd)
+                    .map_err(|err| {
+                        CodexErr::Fatal(format!(
+                            "failed to resolve environment default cwd {}: {err}",
+                            default_cwd.display()
+                        ))
+                    })?;
```
starr-openai: remove

codex-rs/exec-server/src/environment.rs
```diff
@@ -14,17 +16,37 @@ use crate::remote_file_system::RemoteFileSystem;
...
+    exec_server_url: Option<String>,
+    default_cwd: Option<PathBuf>,
+}
 
 /// Lazily creates and caches the active environment for a session.
 ///
 /// The manager keeps the session's environment selection stable so subagents
 /// and follow-up turns preserve an explicit disabled state.
 #[derive(Debug)]
 pub struct EnvironmentManager {
```
starr-openai, line 47: please make this interface a trait so the public interface is more clear, and then this is the concrete implementation we use for now

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2390,8 +2399,9 @@ impl CodexMessageProcessor {
         service_name: Option<String>,
         experimental_raw_events: bool,
         request_trace: Option<W3cTraceContext>,
+        environment_id: Option<String>,
     ) {
-        let requested_cwd = typesafe_overrides.cwd.clone();
+        let explicit_cwd = typesafe_overrides.cwd.clone();
```
starr-openai: revert these cwd changes too. leave it as requested_cwd, and requested_cwd takes precedence exactly as it did before.

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2722,6 +2722,11 @@ pub struct ThreadStartParams {
     pub ephemeral: Option<bool>,
     #[ts(optional = nullable)]
     pub session_start_source: Option<ThreadStartSource>,
+    /// Optional ordered environment ids to use for this thread. For now only
+    /// the first id is used. When omitted or empty, the default environment is
+    /// used.
+    #[ts(optional = nullable)]
+    pub environment_ids: Option<Vec<String>>,
```
pakrym-oai: + experimental

codex-rs/exec-server/src/environment.rs
```diff
@@ -14,17 +16,37 @@ use crate::remote_file_system::RemoteFileSystem;
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum ExplicitEnvironmentId {
+    Local,
+    Remote,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct EnvironmentConfig {
+    exec_server_url: Option<String>,
+    default_cwd: Option<PathBuf>,
```
pakrym-oai: I don't think we should have default_cwd on env registration. EnvironmentManager outlives threads and can provide many different cwds

codex-rs/exec-server/src/environment.rs
```diff
@@ -14,17 +16,37 @@ use crate::remote_file_system::RemoteFileSystem;
...
+}
 
 /// Lazily creates and caches the active environment for a session.
 ///
 /// The manager keeps the session's environment selection stable so subagents
 /// and follow-up turns preserve an explicit disabled state.
 #[derive(Debug)]
 pub struct EnvironmentManager {
-    exec_server_url: Option<String>,
+    current_environment_config: Option<EnvironmentConfig>,
```
pakrym-oai: Why are we not storing an array of envs?

codex-rs/app-server-protocol/src/protocol/v2.rs
pakrym-oai: Should we start with Vec<Struct>? So we have somewhere to put cwd later?

pakrym-oai (changes requested):
Let's not have a global per-env cwd and make env_manager less hardcoded - imagine we do have a dynamic list of environments but for now populate it from env var/local.

codex-rs/exec-server/src/environment.rs
starr-openai: my thinking here was more for convenience. cwd of the app-server host will almost never be valid for the exec-server, so this could pick up the cwd where exec-server is started / configured.  otherwise exec-server is not likely to work without passing a thread-level cwd, but that seems fragile.

codex-rs/exec-server/src/environment.rs
starr-openai: I can change this to a Map from the env id => env.  For this PR the only real thing that mattered was the public signature, which allows selecting an env by id. internally this was simpler / slightly more explicit.  but sure, lemme swap.

starr-openai:
Superseded by stacked PRs #18225 and #18226.

codex-rs/core/src/thread_manager.rs
```diff
@@ -906,10 +964,44 @@ impl ThreadManagerState {
         inherited_exec_policy: Option<Arc<crate::exec_policy::ExecPolicyManager>>,
         parent_trace: Option<W3cTraceContext>,
         user_shell_override: Option<crate::shell::Shell>,
+        environments: Option<Vec<codex_protocol::protocol::TurnEnvironment>>,
     ) -> CodexResult<NewThread> {
-        let environment = self
-            .environment_manager
-            .current()
+        let environment_manager = if environments
```
pakrym-oai: This looks surprisingly complicated.

codex-rs/core/src/codex_tests_guardian.rs
```diff
@@ -452,6 +452,10 @@ async fn guardian_subagent_does_not_inherit_parent_exec_policy_rules() {
         user_shell_override: None,
         parent_trace: None,
         analytics_events_client: None,
+        environments: vec![codex_protocol::protocol::TurnEnvironment {
```
pakrym-oai: Can we keep all tests to None for now?

codex-rs/core/src/codex_tests.rs
```diff
@@ -3145,6 +3154,9 @@ pub(crate) async fn make_session_and_context() -> (Session, TurnContext) {
             config.js_repl_node_path.clone(),
         ),
         environment: Some(Arc::clone(&environment)),
+        environment_manager: Arc::new(codex_exec_server::EnvironmentManager::new(
```
pakrym-oai: Why do we need environment manager on session?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2742,6 +2747,24 @@ pub struct ThreadStartParams {
     pub persist_extended_history: bool,
 }
 
+#[derive(Serialize, Deserialize, Debug, Default, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ThreadEnvironmentParams {
+    pub environment_id: String,
+    #[ts(optional = nullable)]
+    pub cwd: Option<PathBuf>,
```
pakrym-oai: AbsolutePathBuf

pakrym-oai: Not optional

codex-rs/protocol/src/protocol.rs
```diff
@@ -495,6 +502,11 @@ pub enum Op {
     /// turns that rely on persistent session-level context (for example,
     /// [`Op::UserInput`]).
     OverrideTurnContext {
```
pakrym-oai, line 504: Why have environment on OverrideTurnContext but not UserTurn?

codex-rs/exec-server/src/environment.rs
```diff
@@ -14,17 +15,40 @@ use crate::remote_file_system::RemoteFileSystem;
...
+}
 
 /// Lazily creates and caches the active environment for a session.
 ///
 /// The manager keeps the session's environment selection stable so subagents
 /// and follow-up turns preserve an explicit disabled state.
 #[derive(Debug)]
 pub struct EnvironmentManager {
-    exec_server_url: Option<String>,
+    current_environment_id: Option<EnvironmentId>,
```
pakrym-oai: I think environment manager should stay shared and not have a current id. Just the list of available environments. Then each thread get it's own list of Environment structs

codex-rs/exec-server/src/environment.rs
```diff
@@ -165,6 +236,29 @@ impl Environment {
         local_runtime_paths: Option<ExecServerRuntimePaths>,
     ) -> Result<Self, ExecServerError> {
         let (exec_server_url, disabled) = normalize_exec_server_url(exec_server_url);
+        let id = if disabled {
```
pakrym-oai, line 228: This doesn't look right. Env manager should have either:
1 env when REMOTE_ENV_ENV_VAR=none or not set. 
2 envs local and remote when REMOTE_ENV_ENV_VAR is set.

codex-rs/exec-server/src/environment.rs
```diff
@@ -236,11 +330,73 @@ fn normalize_exec_server_url(exec_server_url: Option<String>) -> (Option<String>
         Some(url) => (Some(url.to_string()), false),
     }
 }
+
+fn normalize_environment_id(environment_id: Option<&str>) -> Option<EnvironmentId> {
```
pakrym-oai, line 323: why?

pakrym-oai (changes requested):
I don't think we should mutate and clone EnvironmentManager per turn. There is a global environment manager that contains a list of all environments registered with app server when a turn is started we create/attach a set of environments to a specific turn.

PR #18176, opened 2026-04-16
Accept JWT agent identity auth input

efrazer-oai:
## Stack

1. #18176 - Accept JWT agent identity auth input
2. #18177 - Enable agent identity for tokenless startup auth
3. #18180 - Load agent identity from `CODEX_AGENT_IDENTITY`

## What changed

`auth.json.agent_identity` can now be either the existing object or a JWT string.

The JWT payload requires:

- `workspace_id`
- `agent_runtime_id`
- `agent_private_key`

It can also include:

- `chatgpt_user_id`
- `registered_at`

`background_task_id` is not part of the JWT contract. Codex can cache that later in the stored object after task registration.

## Why

Launchers can pass one predictable string for an existing registered agent identity. Codex decodes the JWT payload into the same `AgentIdentityAuthRecord` it already stores.

## Validation

- `cd codex-rs && just fmt`
- `cd codex-rs && cargo check -p codex-login -p codex-core`
- `cd codex-rs && cargo test -p codex-login agent_identity`
- `cd codex-rs && cargo test -p codex-core agent_identity`
- `cd codex-rs && just fix -p codex-login`

PR #18177, opened 2026-04-16
Enable agent identity for tokenless startup auth

efrazer-oai:
## Stack

1. #18176 - Accept JWT agent identity auth input
2. #18177 - Enable agent identity for tokenless startup auth
3. #18180 - Load agent identity from `CODEX_AGENT_IDENTITY`

## What changed

Codex now treats agent identity as enabled when ChatGPT auth has `agent_identity` and no user OAuth tokens.

The small shape change is that `AgentIdentityBinding` can carry `access_token: None`. Token-backed auth still mints the human biscuit before registration. Tokenless auth reuses the stored agent identity, signs the task registration payload with the agent private key, and then uses the task id to build `AgentAssertion` auth.

This also handles two startup checks for tokenless auth:

- request auth setup no longer fails while looking for a bearer token
- forced workspace checks compare against `agent_identity.workspace_id`

## Validation

- `cd codex-rs && just fmt`
- `cd codex-rs && cargo check -p codex-login -p codex-core`
- `cd codex-rs && cargo test -p codex-login agent_identity`
- `cd codex-rs && cargo test -p codex-core agent_identity`
- `cd codex-rs && cargo test -p codex-login enforce_login_restrictions`
- `cd codex-rs && just fix -p codex-login`
- `cd codex-rs && just fix -p codex-core`

### DIFF
--- a/codex-rs/core/src/agent_identity.rs
+++ b/codex-rs/core/src/agent_identity.rs
@@ -91,7 +91,7 @@ struct AgentIdentityBinding {
     binding_id: String,
     chatgpt_account_id: String,
     chatgpt_user_id: Option<String>,
-    access_token: String,
+    access_token: Option<String>,
 }
 
 struct GeneratedAgentKeyMaterial {
@@ -116,10 +116,15 @@ impl AgentIdentityManager {
 
     pub(crate) fn is_enabled(&self) -> bool {
         self.feature_enabled
+            || self
+                .auth_manager
+                .auth_cached()
+                .as_ref()
+                .is_some_and(CodexAuth::is_agent_identity_only)
     }
 
     pub(crate) async fn ensure_registered_identity(&self) -> Result<Option<StoredAgentIdentity>> {
-        if !self.feature_enabled {
+        if !self.is_enabled() {
             return Ok(None);
         }
 
@@ -154,7 +159,7 @@ impl AgentIdentityManager {
     }
 
     pub(crate) async fn task_matches_current_identity(&self, task: &RegisteredAgentTask) -> bool {
-        if !self.feature_enabled {
+        if !self.is_enabled() {
             return false;
         }
 
@@ -244,11 +249,15 @@ impl AgentIdentityManager {
         target_url: &str,
     ) -> Result<String> {
         let url = agent_identity_biscuit_url(&self.chatgpt_base_url);
+        let access_token = binding
+            .access_token
+            .as_deref()
+            .context("ChatGPT access token is unavailable")?;
         let request_id = agent_identity_request_id()?;
         let client = create_client();
         let response = client
             .get(&url)
-            .bearer_auth(&binding.access_token)
+            .bearer_auth(access_token)
             .header("X-Request-Id", request_id.clone())
             .header("X-Original-Method", target_method)
             .header("X-Original-Url", target_url)
@@ -455,6 +464,16 @@ impl AgentIdentityBinding {
             return None;
         }
 
+        if auth.is_agent_identity_only() {
+            let record = auth.agent_identity_record()?;
+            return Some(Self {
+                binding_id: format!("chatgpt-account-{}", record.workspace_id),
+                chatgpt_account_id: record.workspace_id,
+                chatgpt_user_id: record.chatgpt_user_id,
+                access_token: None,
+            });
+        }
+
         let token_data = auth.get_token_data().ok()?;
         let resolved_account_id =
             forced_workspace_id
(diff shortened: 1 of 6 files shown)


adrian-openai (commented):
I got these reviews from Codex:

```
[High] Identity-only auth still tries to build Bearer-token request auth before AgentAssertion can happen

Location: [codex-rs/login/src/api_bridge.rs#L26-L30](https://github.com/openai/codex/blob/4f088d5c7391fe467a75e9743dacacbff4d6d65e/codex-rs/login/src/api_bridge.rs#L26-L30)
Additional context: [codex-rs/core/src/client.rs#L653-L663](https://github.com/openai/codex/blob/4f088d5c7391fe467a75e9743dacacbff4d6d65e/codex-rs/core/src/client.rs#L653-L663), [codex-rs/codex-api/src/api_bridge.rs#L205-L211](https://github.com/openai/codex/blob/4f088d5c7391fe467a75e9743dacacbff4d6d65e/codex-rs/codex-api/src/api_bridge.rs#L205-L211)
Why: With tokenless ChatGPT auth, CodexAuth::is_agent_identity_only() is true, but auth_provider_from_auth() still calls auth.get_token()?. Since tokens is None, this fails with token data unavailable while constructing CoreAuthProvider. In the exact PR head, CoreAuthProvider only emits Authorization: Bearer ...; there is no Authorization: AgentAssertion ... request-auth path.
Suggested fix: Add a narrow request-auth path for identity-only auth that can emit AgentAssertion without first requiring auth.get_token(). Ideally test this by capturing a /responses request from tokenless auth and asserting the auth scheme is AgentAssertion.
```

```
[High] Tokenless task registration/reuse still depends on token-backed ChatGPT binding

Location: [codex-rs/core/src/agent_identity/task_registration.rs#L39-L48](https://github.com/openai/codex/blob/4f088d5c7391fe467a75e9743dacacbff4d6d65e/codex-rs/core/src/agent_identity/task_registration.rs#L39-L48)
Additional context: [codex-rs/core/src/agent_identity.rs#L184-L192](https://github.com/openai/codex/blob/4f088d5c7391fe467a75e9743dacacbff4d6d65e/codex-rs/core/src/agent_identity.rs#L184-L192), [codex-rs/core/src/agent_identity.rs#L466-L488](https://github.com/openai/codex/blob/4f088d5c7391fe467a75e9743dacacbff4d6d65e/codex-rs/core/src/agent_identity.rs#L466-L488), [codex-rs/core/src/codex.rs#L1537-L1565](https://github.com/openai/codex/blob/4f088d5c7391fe467a75e9743dacacbff4d6d65e/codex-rs/core/src/codex.rs#L1537-L1565)
Why: PR 18177 updates startup identity loading, but the task path still uses the old assumptions. register_task() returns Ok(None) when feature_enabled is false. Even when the feature is true, current_auth_binding() goes through AgentIdentityBinding::from_auth(), which calls auth.get_token_data().ok()?; tokenless auth cannot produce that. task_matches_current_binding() has the same problem and may treat a valid cached/persisted task as stale.
Suggested fix: Add an identity-only binding path derived from AgentIdentityAuthRecord instead of token data. If tokenless auth is expected to carry a pre-registered task id, make task matching/reuse validate against that stored identity. If tokenless auth cannot register a task, fail explicitly instead of silently returning None.
```


```
Medium] Forced workspace enforcement logs out matching identity-only auth

Location: [codex-rs/login/src/auth/manager.rs#L596-L610](https://github.com/openai/codex/blob/4f088d5c7391fe467a75e9743dacacbff4d6d65e/codex-rs/login/src/auth/manager.rs#L596-L610)
Additional context: [codex-rs/exec/src/lib.rs#L420-L425](https://github.com/openai/codex/blob/4f088d5c7391fe467a75e9743dacacbff4d6d65e/codex-rs/exec/src/lib.rs#L420-L425)
Why: codex exec runs login restriction enforcement on startup. If forced_chatgpt_workspace_id is set, identity-only ChatGPT auth enters the workspace restriction branch, then calls auth.get_token_data(). With no tokens, that errors and logs out, even if agent_identity.workspace_id matches the forced workspace. Our smoke config used a forced workspace, so this matters for the devbox scenario.
Suggested fix: Before calling get_token_data(), handle auth.is_agent_identity_only() by comparing auth.agent_identity_record().workspace_id against the forced workspace. Add matching and mismatching tests.
```

I think that's what is left to get agent assertion only auth to work (my smoke test, in getting working got through all of these changes too). Doesn't need to be changed here, but I think needs to happen first.

PR #18188, opened 2026-04-16
Add tabbed lists, single line rendering, col width changes

canvrno-oai:
This PR adds shared bottom-pane selection-list for future `/plugins` menu work and wires the existing `/plugins` menu into the new list-rendering path without changing it to tabs yet. The main user-visible effect is that the current plugin list now renders as a denser single-line list with shared name-column sizing, while the tabbed selection support remains available for follow-up PRs but is currently unused in production menus.

- Add generic tabbed selection-list support to the bottom pane, including per-tab headers/items and tab-aware list state
- Add single-line row rendering with ellipsis truncation for dense list UIs
- Add shared name-column width support so descriptions align consistently across rows
- Wire the current /plugins menu to the new single-line and shared column-width behavior only
- Keep tabbed menu adoption deferred; no existing menu is switched to tabs in this PR

### DIFF
--- a/codex-rs/tui/src/bottom_pane/bottom_pane_view.rs
+++ b/codex-rs/tui/src/bottom_pane/bottom_pane_view.rs
@@ -49,6 +49,11 @@ pub(crate) trait BottomPaneView: Renderable {
         None
     }
 
+    /// Active tab id for tabbed list-based views.
+    fn active_tab_id(&self) -> Option<&str> {
+        None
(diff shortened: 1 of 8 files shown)


canvrno-oai:
@codex review

fcoury-oai (approved):
I was able to replicate the initial improvement on the denser list. 

Code looks good as well, approved.

PR #18196, opened 2026-04-16
Use in-process app-server for unknown-thread MCP read test

mzeng-openai:
## Summary
- Switch the unknown-thread MCP resource read test from the stdio subprocess to the in-process app-server path.
- Keep the assertion focused on the returned error message while avoiding child-process teardown timing issues in nextest.

## Testing
- Not run (not requested)

### DIFF
--- a/codex-rs/app-server/tests/suite/v2/mcp_resource.rs
+++ b/codex-rs/app-server/tests/suite/v2/mcp_resource.rs
@@ -209,12 +209,9 @@ async fn mcp_resource_read_returns_error_for_unknown_thread() -> Result<()> {
         Ok(result) => anyhow::bail!("expected thread-not-found error, got response: {result:?}"),
         Err(error) => error,
     };
-    let error_debug = format!("{error:?}");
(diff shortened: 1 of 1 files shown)


codex-rs/app-server/tests/suite/v2/mcp_resource.rs
```diff
@@ -149,25 +160,61 @@ stream_max_retries = 0
...
+    };
+    let error_debug = format!("{error:?}");
+    let message = error.message;
 
     assert!(
-        error.error.message.contains("thread not found"),
-        "expected thread-not-found error, got: {error:?}"
+        message.contains("thread not found"),
+        "expected thread-not-found error, got: {error_debug}"
     );
```
bolinfest: I think this can be simplified?

```suggestion
assert!(
        error.message.contains("thread not found"),
        "expected thread-not-found error, got: {error:?}"
);
```

bolinfest (approved):
Thanks for tightening up this test!

PR #18197, opened 2026-04-16
feat: add guardian network approval trigger context

viyatb-oai:
## Summary

Give guardian network-access reviews the command context that triggered a managed-network approval. The prompt JSON now includes the originating tool call id, tool name, command argv, cwd, sandbox permissions, additional permissions, justification, and tty state when a single active tool call can be attributed.

The implementation keeps the trigger shape canonical by serializing `GuardianNetworkAccessTrigger` directly and lets each runtime build that trigger from its `ToolCtx`. Non-guardian approval prompts avoid cloning the full trigger payload.

## UX changes

Guardian network-access reviews now include a `trigger` object that explains what command caused the network approval. Instead of seeing only the requested host, the guardian reviewer can also see the originating tool call, argv, working directory, sandbox mode, justification, and tty state.

Example payload the guardian reviewer can see:

```json
{
  "tool": "network_access",
  "target": "https://api.github.com:443",
  "host": "api.github.com",
  "protocol": "https",
  "port": 443,
  "trigger": {
    "callId": "call_abc123",
    "toolName": "shell",
    "command": ["gh", "api", "/repos/openai/codex/pulls/18197"],
    "cwd": "/workspace/codex",
    "sandboxPermissions": "require_escalated",
    "justification": "Fetch PR metadata from GitHub.",
    "tty": false
  }
}
```

The network review itself remains scoped to the network decision: `target_item_id` stays `null`. `trigger.callId` is attribution context only, so clients can still distinguish network reviews from item-targeted command reviews.

## Verification

- Added coverage for serializing network trigger context in guardian approval JSON.
- Added regression coverage that network guardian reviews do not reuse `trigger.callId` as `target_item_id`.


codex-rs/core/src/guardian/approval_request.rs
```diff
@@ -332,6 +363,7 @@ pub(crate) fn guardian_assessment_action(
             host,
             protocol,
             port,
+            trigger: _,
```
dylan-hurd-oai: For my understanding, why can this be empty here?

codex-rs/core/src/tools/network_approval_tests.rs
```diff
@@ -207,17 +210,63 @@ fn denied_blocked_request(host: &str) -> BlockedRequest {
     })
 }
 
+async fn register_test_call(service: &NetworkApprovalService, registration_id: &str) {
```
dylan-hurd-oai: nit, can we make this name slight more specific

codex-rs/core/src/guardian/approval_request.rs
```diff
@@ -123,6 +140,18 @@ struct McpToolCallApprovalAction<'a> {
     annotations: Option<&'a GuardianMcpAnnotations>,
 }
 
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct NetworkAccessApprovalAction<'a> {
```
dylan-hurd-oai, line 153: do we need a custom lifetime annotation here? what benefit does it bring?

dylan-hurd-oai (commented):
couple small comments but overall looks reasonable

codex-rs/core/src/guardian/approval_request.rs
viyatb-oai: Good question. The trigger is only Guardian reviewer prompt context, while GuardianAssessmentAction keeps the stable action identity scoped to the network target. I also switched this match arm to use `..` so we do not spell out ignored fields here.

codex-rs/core/src/tools/network_approval_tests.rs
viyatb-oai: Renamed the helper to `register_call_with_default_shell_trigger` in fcbc8f7d03 so the test setup is more specific at each call site.

codex-rs/core/src/guardian/approval_request.rs
viyatb-oai: The lifetime is needed because this private serializer struct borrows `target`, `host`, and `trigger` instead of cloning them into owned fields. It matches the surrounding borrowed action structs; removing it would mostly trade the annotation for extra owned data just for serialization.

codex-rs/core/src/guardian/approval_request.rs
dylan-hurd-oai: We shouldn't use `..` here we should be explicit about what we're excluding

codex-rs/core/src/tools/network_approval.rs
```diff
@@ -415,6 +425,11 @@ impl NetworkApprovalService {
             }
         }
         let use_guardian = routes_approval_to_guardian(&turn_context);
+        let trigger = if use_guardian {
+            owner_call.as_ref().map(|call| call.trigger.clone())
+        } else {
+            None
+        };
```
dylan-hurd-oai: feels unnecessarily complicated to make presence here conditional on guardian mode. Is there a disadvantage beside a bit of extra struct contents?

dylan-hurd-oai (approved):

PR #18202, opened 2026-04-16
feat(sandbox): add Windows deny-read parity

viyatb-oai:
## Why

The split filesystem policy stack already supports exact and glob `access = none` read restrictions on macOS and Linux. Windows still needed subprocess handling for those deny-read policies without claiming enforcement from a backend that cannot provide it.

## Key finding

The unelevated restricted-token backend cannot safely enforce deny-read overlays. Its `WRITE_RESTRICTED` token model is authoritative for write checks, not read denials, so this PR intentionally fails that backend closed when deny-read overrides are present instead of claiming unsupported enforcement.

## What changed

This PR adds the Windows deny-read enforcement layer and makes the backend split explicit:

- Resolves Windows deny-read filesystem policy entries into concrete ACL targets.
- Preserves exact missing paths so they can be materialized and denied before an enforceable sandboxed process starts.
- Snapshot-expands existing glob matches into ACL targets for Windows subprocess enforcement.
- Honors `glob_scan_max_depth` when expanding Windows deny-read globs.
- Plans both the configured lexical path and the canonical target for existing paths so reparse-point aliases are covered.
- Threads deny-read overrides through the elevated/logon-user Windows sandbox backend and unified exec.
- Applies elevated deny-read ACLs synchronously before command launch rather than delegating them to the background read-grant helper.
- Reconciles persistent deny-read ACEs per sandbox principal so policy changes do not leave stale deny-read ACLs behind.
- Fails closed on the unelevated restricted-token backend when deny-read overrides are present, because its `WRITE_RESTRICTED` token model is not authoritative for read denials.

## Landed prerequisites

These prerequisite PRs are already on `main`:

1. #15979 `feat(permissions): add glob deny-read policy support`
2. #18096 `feat(sandbox): add glob deny-read platform enforcement`
3. #17740 `feat(config): support managed deny-read requirements`

This PR targets `main` directly and contains only the Windows deny-read enforcement layer.

## Implementation notes

- Exact deny-read paths remain enforceable on the elevated path even when they do not exist yet: Windows materializes the missing path before applying the deny ACE, so the sandboxed command cannot create and read it during the same run.
- Existing exact deny paths are preserved lexically until the ACL planner, which then adds the canonical target as a second ACL target when needed. That keeps both the configured alias and the resolved object covered.
- Windows ACLs do not consume Codex glob syntax directly, so glob deny-read entries are expanded to the concrete matches that exist before process launch.
- Glob traversal deduplicates directory visits within each pattern walk to avoid cycles, without collapsing distinct lexical roots that happen to resolve to the same target.
- Persistent deny-read ACL state is keyed by sandbox principal SID, so cleanup only removes ACEs owned by the same backend principal.
- Deny-read ACEs are fail-closed on the elevated path: setup aborts if mandatory deny-read ACL application fails.
- Unelevated restricted-token sessions reject deny-read overrides early instead of running with a silently unenforceable read policy.

## Verification

- `cargo test -p codex-core windows_restricted_token_rejects_unreadable_split_carveouts`
- `just fmt`
- `just fix -p codex-core`
- GitHub Actions rerun is in progress on the pushed head.


codex-rs/windows-sandbox-rs/src/lib.rs
```diff
@@ -96,6 +112,17 @@ pub use conpty::ConptyInstance;
 #[cfg(target_os = "windows")]
 pub use conpty::spawn_conpty_process_as_user;
 #[cfg(target_os = "windows")]
+pub use deny_read_acl::DenyReadAclRecordKind;
+#[cfg(target_os = "windows")]
+pub use deny_read_acl::apply_deny_read_acls;
+#[cfg(target_os = "windows")]
+pub use deny_read_acl::cleanup_stale_persistent_deny_read_acls;
```
iceweasel-oai: I'd consider not doing the stale cleanup as part of this first pass, to keep it (somewhat) simple and more tightly scoped. yes it's incorrect if someone changes their mind about one of these paths, but the result is that the sandbox has less access than it should, and could generally escalate in this situation.

That would allow

codex-rs/windows-sandbox-rs/src/acl.rs
```diff
@@ -507,6 +549,85 @@ pub unsafe fn add_deny_write_ace(path: &Path, psid: *mut c_void) -> Result<bool>
...
+/// Adds a deny ACE to prevent reads for the given SID on the target path.
+///
+/// `SetEntriesInAclW` places newly-created deny ACEs before allow ACEs, which
+/// keeps the resulting DACL in the order Windows expects for denies to win.
+/// The ACE is inheritable so a deny applied to a materialized directory also
+/// covers files and directories later created underneath it.
+///
+/// # Safety
+/// Caller must ensure `psid` points to a valid SID and `path` refers to an existing file or directory.
+pub unsafe fn add_deny_read_ace(path: &Path, psid: *mut c_void) -> Result<bool> {
```
iceweasel-oai, line 589: I think this overlaps a fair bit with https://github.com/openai/codex/blob/main/codex-rs/windows-sandbox-rs/src/acl.rs#L447

I would probably merge them into one function where you can pass in which ACL mask you want to apply (read vs write) 

We do something similar for applying "allow" ACEs too

codex-rs/windows-sandbox-rs/src/deny_read_acl.rs
```diff
@@ -0,0 +1,226 @@
...
+/// # Safety
+/// Caller must pass a valid SID pointer for the sandbox principal being denied.
+pub unsafe fn apply_deny_read_acls(paths: &[PathBuf], psid: *mut c_void) -> Result<Vec<PathBuf>> {
+    let planned = plan_deny_read_acl_paths(paths);
+    let mut applied = Vec::new();
+    let mut seen = HashSet::new();
+    let mut added_in_this_call: Vec<PathBuf> = Vec::new();
+    for path in planned {
+        let result = (|| -> Result<bool> {
+            if !path.exists() {
```
iceweasel-oai, line 58: I think codex mentioned this before, but it looks like for these newly materialized paths, they get a deny read ACE but not a deny write ACE, ie they don't go through `add_deny_write_ace`

double check this because this seems like a probably blocking bug

codex-rs/windows-sandbox-rs/src/deny_read_resolver.rs
```diff
@@ -0,0 +1,298 @@
...
+    seen_paths: &mut HashSet<PathBuf>,
+    seen_scan_dirs: &mut HashSet<PathBuf>,
+    max_depth: Option<usize>,
+    depth: usize,
+) -> Result<(), String> {
+    if !path.exists() {
+        return Ok(());
+    }
+
+    if matcher.is_read_denied(path) {
```
iceweasel-oai, line 84: I think this is a blocking bug. `ReadDenyMatcher` intentionally fails closed when any deny-read glob fails to compile: `is_read_denied()` returns true for every path. That behavior may be OK for direct runtime read checks, but here it is being used to decide which host paths receive persistent DENY-read ACEs.

So if a malformed glob reaches this resolver, the walker will treat every path under the glob scan root as denied and `apply_deny_read_acls` will materialize DENY-read ACEs for the whole visited tree. A config typo like an invalid character range/brace should fail sandbox setup, not persist broad ACL denies.

Can we validate/compile the glob patterns for ACL expansion separately and return an error before walking, instead of reusing the fail-closed matcher behavior here?

codex-rs/windows-sandbox-rs/src/lib.rs
viyatb-oai: Done in 4a886a0: I removed the persistent deny-read record/cleanup path so this first pass only adds enforcement and does not try to manage stale ACL lifecycle yet.

codex-rs/windows-sandbox-rs/src/acl.rs
viyatb-oai: Done in 4a886a0: `add_deny_read_ace` and `add_deny_write_ace` now share the same `add_deny_ace` implementation, parameterized by deny kind/mask.

codex-rs/windows-sandbox-rs/src/deny_read_acl.rs
viyatb-oai: Confirmed, and fixed in 4a886a0: the restricted-token path now materializes explicit deny-write carveouts before applying ACLs, so a missing exact `None` path gets the write deny as well as the read deny.

codex-rs/windows-sandbox-rs/src/deny_read_resolver.rs
viyatb-oai: Done in 4a886a0: ACL expansion now uses a fallible `ReadDenyMatcher::try_new` path, so malformed deny-read globs fail setup before walking the tree. I also added a regression test for the invalid-range case.

iceweasel-oai (approved):

codex-rs/cli/src/debug_sandbox.rs
bolinfest, line 350: Should we be collapsing things into the legacy sandbox policy here?

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -886,6 +886,34 @@ impl UnifiedExecProcessManager {
                     "windows sandbox: failed to resolve codex_home: {err}"
                 ))
             })?;
+            let additional_deny_write_paths = request
+                .windows_sandbox_filesystem_overrides
+                .as_ref()
+                .map(|overrides| overrides.additional_deny_write_paths.clone())
```
bolinfest, line 892: I feel like we should be going in the other direction and changing functions like `spawn_windows_sandbox_session_elevated` to take `AbsolutePathBuf` instead of `PathBuf`.

codex-rs/core/src/exec.rs
```diff
@@ -1152,6 +1153,16 @@ fn normalize_windows_override_path(path: &Path) -> std::result::Result<PathBuf,
...
 }
 
+fn windows_policy_has_root_read_access(
+    file_system_sandbox_policy: &FileSystemSandboxPolicy,
+    cwd: &AbsolutePathBuf,
+) -> bool {
+    let Some(root) = cwd.as_path().ancestors().last() else {
+        return false;
+    };
+    file_system_sandbox_policy.can_read_path_with_cwd(root, cwd.as_path())
```
bolinfest, line 1157:
```suggestion
    match cwd.as_path().ancestors().last() {
        Some(root) =>  file_system_sandbox_policy.can_read_path_with_cwd(root, cwd.as_path()),
        None => false,
    }
```

codex-rs/core/src/exec_tests.rs
```diff
@@ -662,11 +662,65 @@ fn windows_restricted_token_supports_full_read_split_write_read_carveouts() {
...
+        .abs();
+    let blocked = cwd.join("blocked");
+    std::fs::create_dir_all(blocked.as_path()).expect("create blocked");
+    let policy = SandboxPolicy::WorkspaceWrite {
+        writable_roots: vec![],
+        network_access: false,
+        exclude_tmpdir_env_var: true,
+        exclude_slash_tmp: true,
+    };
+    let file_system_policy = FileSystemSandboxPolicy::restricted(vec![
```
bolinfest, line 685: I feel like we should create a macro or function that takes the TOML for `config.toml` because I find these test setups really hard to read...

bolinfest (commented):
Still reviewing, but I wanted to send what I found so far...

codex-rs/cli/src/debug_sandbox.rs
viyatb-oai: This debug path still intentionally serializes `config.permissions.legacy_sandbox_policy(...)`: the Windows debug runner entrypoints consume the compatibility policy JSON, and this PR only threads empty filesystem-override slices through the updated signatures. I’m leaving that conversion as-is here.

codex-rs/core/src/exec.rs
viyatb-oai: Already addressed on the current branch: `windows_policy_has_root_read_access` now uses `cwd.as_path().ancestors().last()` and passes that root into `can_read_path_with_cwd`.

codex-rs/core/src/unified_exec/process_manager.rs
viyatb-oai: Done in `8bccc16368`: the newly added deny-read / deny-write override slices now stay `AbsolutePathBuf` through the cross-crate Windows sandbox entrypoints and core callers, and only convert to `PathBuf` inside the Windows implementation edge that feeds the existing ACL/setup plumbing.

codex-rs/core/tests/suite/windows_sandbox.rs
```diff
@@ -0,0 +1,224 @@
...
+            path: FileSystemPath::Path { path: exact_secret },
+            access: FileSystemAccessMode::None,
+        },
+    ]);
+    let permission_profile = PermissionProfile::from_runtime_permissions(
+        &file_system_sandbox_policy,
+        NetworkSandboxPolicy::Restricted,
+    );
+
+    let output = process_exec_tool_call(
```
bolinfest:
```suggestion
    let ExecToolCallOutput { stdout, stderr, .. } = process_exec_tool_call(
```

and then below you can inline these values into the format strings you use with `assert!`

bolinfest (approved):

PR #18208, opened 2026-04-16
Add session config loader interface

rasmusrygaard:
## Why

Cloud-hosted sessions need a way for the service that starts or manages a thread to provide session-owned config without treating all config as if it came from the same user/project/workspace TOML stack.

The important boundary is ownership: some values should be controlled by the session/orchestrator, some by the authenticated user, and later some may come from the executor. The earlier broad config-store shape made that boundary too fuzzy and overlapped heavily with the existing filesystem-backed config loader. This PR starts with the smaller piece we need now: a typed session config loader that can feed the existing config layer stack while preserving the normal precedence and merge behavior.

## What Changed

- Added `ThreadConfigLoader` and related typed payloads in `codex-config`.
  - `SessionThreadConfig` currently supports `model_provider`, `model_providers`, and feature flags.
  - `UserThreadConfig` is present as an ownership boundary, but does not yet add TOML-backed fields.
  - `NoopThreadConfigLoader` preserves existing behavior when no external loader is configured.
  - `StaticThreadConfigLoader` supports tests and simple callers.

- Taught thread config sources to produce ordinary `ConfigLayerEntry` values so the existing `ConfigLayerStack` remains the place where precedence and merging happen.

- Wired the loader through `ConfigBuilder`, the config loader, and app-server startup paths so app-server can provide session-owned config before deriving a thread config.

- Added coverage for:
  - translating typed thread config into config layers,
  - inserting thread config layers into the stack at the right precedence,
  - applying session-provided model provider and feature settings when app-server derives config from thread params.

## Follow-Ups

This intentionally stops short of adding the remote/service transport. The next pieces are expected to be:

1. Define the proto/API shape for this interface.
2. Add a client implementation that can source session config from the service side.

## Verification

- Added unit coverage in `codex-config` for the loader and layer conversion.
- Added `codex-core` config loader coverage for thread config layer precedence.
- Added app-server coverage that verifies session thread config wins over request-provided config for model provider and feature settings.

### DIFF
--- a/codex-rs/app-server-client/src/lib.rs
+++ b/codex-rs/app-server-client/src/lib.rs
@@ -41,6 +41,7 @@ use codex_app_server_protocol::Result as JsonRpcResult;
 use codex_app_server_protocol::ServerNotification;
 use codex_app_server_protocol::ServerRequest;
 use codex_arg0::Arg0DispatchPaths;
+use codex_config::NoopThreadConfigLoader;
 use codex_core::config::Config;
 use codex_core::config_loader::CloudRequirementsLoader;
 use codex_core::config_loader::LoaderOverrides;
@@ -385,6 +386,7 @@ impl InProcessClientStartArgs {
             cli_overrides: self.cli_overrides,
             loader_overrides: self.loader_overrides,
             cloud_requirements: self.cloud_requirements,
+            thread_config_loader: Arc::new(NoopThreadConfigLoader),
             feedback: self.feedback,
             log_db: self.log_db,
             environment_manager: self.environment_manager,
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -213,9 +213,7 @@ use codex_backend_client::AddCreditsNudgeCreditType as BackendAddCreditsNudgeCre
 use codex_backend_client::Client as BackendClient;
 use codex_chatgpt::connectors;
 use codex_cloud_requirements::cloud_requirements_loader;
-use codex_config::ThreadConfigContext;
 use codex_config::ThreadConfigLoader;
-use codex_config::ThreadConfigSource;
 use codex_config::types::McpServerTransportConfig;
 use codex_core::CodexThread;
 use codex_core::ForkSnapshot;
@@ -354,7 +352,6 @@ use std::collections::BTreeMap;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::io::Error as IoError;
-use std::io::ErrorKind;
 use std::path::Path;
 use std::path::PathBuf;
 use std::sync::Arc;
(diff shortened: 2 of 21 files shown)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -9065,6 +9238,35 @@ async fn derive_config_from_params(
...
+                    merged_cli_overrides.push((format!("model_providers.{provider_id}"), value));
+                }
+                if let Some(model_provider) = &config.model_provider {
+                    typesafe_overrides.model_provider = Some(model_provider.clone());
+                }
+            }
+            ThreadConfigSource::User(_) => {}
+        }
+    }
+    Ok(())
```
rasmusrygaard: not sure if this is the best way to actually inject the values, feels a bit dissatisfying that we're merging into two different locations

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2397,10 +2401,32 @@ impl CodexMessageProcessor {
...
+                let error = JSONRPCErrorError {
+                    code: INTERNAL_ERROR_CODE,
+                    message: format!("failed to load thread config: {err}"),
+                    data: None,
+                };
+                listener_task_context
+                    .outgoing
+                    .send_error(request_id, error)
+                    .await;
+                return;
```
rasmusrygaard: unclear to me what the correct error semantics are here

codex-rs/app-server/src/codex_message_processor.rs
wiltzius-openai, line 9350: think this "three" is up to "five" by now

codex-rs/config/src/thread_config.rs
```diff
@@ -0,0 +1,175 @@
...
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct ThreadConfigContext {
+    pub thread_id: Option<String>,
+    pub cwd: Option<PathBuf>,
+}
+
+/// Config values owned by the service that starts or manages the session.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct SessionThreadConfig {
+    pub model_provider: Option<String>,
```
wiltzius-openai, line 23: should this also be a ModelProviderInfo?

wiltzius-openai (commented):
seems reasonable to me!

codex-rs/app-server-client/src/lib.rs
```diff
@@ -406,8 +405,8 @@ impl InProcessClientStartArgs {
             cli_overrides: self.cli_overrides,
             loader_overrides: self.loader_overrides,
             cloud_requirements: self.cloud_requirements,
```
wiltzius-openai, line 388: this "cloud requirements" value being passed through the config stack is confusingly named -- I think this is like enterprise config requirements sent down from the cloud? But it seems likely we'll trip on it.

codex-rs/config/src/thread_config.rs
rasmusrygaard: this is the pointer to a default model provider id

codex-rs/core/src/config_loader/mod.rs
```diff
@@ -158,6 +161,15 @@ pub async fn load_config_layers_state(
     )
     .await?;
 
+    let thread_config_context = ThreadConfigContext {
+        thread_id: None,
+        cwd: cwd.as_ref().map(AbsolutePathBuf::to_path_buf),
```
pakrym-oai: use absolute path everywhere

pakrym-oai (approved):

PR #18209, opened 2026-04-16
Refactor config loading to use filesystem abstraction

pakrym-oai:
Initial pass propagating FileSystem through config loading.


codex-rs/git-utils/src/lib.rs
```diff
@@ -45,72 +44,3 @@ pub use info::local_git_branches;
...
-
-impl GitSha {
-    pub fn new(sha: &str) -> Self {
-        Self(sha.to_string())
-    }
-}
-
-/// Details of a ghost commit created from a repository state.
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema, TS)]
-pub struct GhostCommit {
```
starr-openai, line 68: why does this move?

starr-openai (approved):

codex-rs/git-utils/src/lib.rs
pakrym-oai: Untangles a circular dependency introduced by git-util depending on exec server (and protocol transitively)

PR #18231, opened 2026-04-16
codex: split thread/read view loading

wiltzius-openai:
Summary
- refactor thread/read into explicit persisted-load, live-load, and merge steps
- preserve existing SQLite/filesystem/live-thread behavior exactly
- keep ThreadStore migration out of this PR so the next PR is easier to review

Validation
- this one's a pure reorganization that relies on existing test coverage

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -438,6 +438,16 @@ enum ThreadShutdownResult {
     TimedOut,
 }
 
+struct ThreadReadViewSource {
+    thread: Thread,
+    rollout_path: Option<PathBuf>,
+}
+
+enum ThreadReadViewError {
+    InvalidRequest(String),
+    Internal(String),
+}
+
 impl Drop for ActiveLogin {
     fn drop(&mut self) {
         self.cancel();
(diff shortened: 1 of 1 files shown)


starr-openai (approved):

aibrahim-oai (approved):

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -3786,17 +3791,83 @@ impl CodexMessageProcessor {
...
+                self.send_internal_error(request_id, message).await;
+                return;
+            }
+        };
+        let response = ThreadReadResponse { thread };
+        self.outgoing.send_response(request_id, response).await;
+    }
+
+    /// Builds the API view for `thread/read` from persisted metadata plus optional live state.
+    async fn read_thread_view(
```
aibrahim-oai, line 3810: nit: can we rename to make it clear that this is for the live thread? and maybe move it to its own file?

wiltzius-openai: this isn't just for the live thread, it is for the combined state derived from the live thread + the persisted thread. The persisted thread state takes precedence if available, falling back to live thread state, but even in the persisted case it adds in some state from the in-memory thread manager (like "status").

The two individual underlying methods are named `load_live_thread_view` and `load_persisted_thread_for_read` which are pretty literal already but if you have suggestions for other names lmk.

I can definitely move this to another file if you think that's better though?

PR #18246, opened 2026-04-17
Sync local plugin imports, async remote imports, refresh caches after…

alexsong-oai:
… import

## Why

`externalAgentConfig/import` used to spawn plugin imports in the background and return immediately. That meant local marketplace imports could still be in flight when the caller refreshed plugin state, so newly imported plugins would not show up right away.

This change makes local marketplace imports complete before the RPC returns, while keeping remote marketplace imports asynchronous so we do not block on remote fetches.

## What changed

- split plugin migration details into local and remote marketplace imports based on the external config source
- import local marketplaces synchronously during `externalAgentConfig/import`
- return pending remote plugin imports to the app-server so it can finish them in the background
- clear the plugin and skills caches before responding to plugin imports, and again after background remote imports complete, so the next `plugin/list` reloads fresh state
- keep marketplace source parsing encapsulated behind `is_local_marketplace_source(...)` instead of re-exporting the internal enum
- add core and app-server coverage for the synchronous local import path and the pending remote import path

## Verification

- `cargo test -p codex-core` (currently fails an existing unrelated test: `config_loader::tests::cli_override_can_update_project_local_mcp_server_when_project_is_trusted`)
- `cargo test` (currently fails existing `codex-app-server` integration tests in MCP/skills/thread-start areas, plus the unrelated `codex-core` failure above)


xl-openai (approved):

alexsong-oai:
@codex review

codex-rs/tui/src/app/app_server_adapter.rs
```diff
@@ -188,6 +188,16 @@ impl App {
                 );
                 return;
             }
+            ServerNotification::ExternalAgentConfigImportCompleted(_) => {
```
alexsong-oai, line 192: FYI @xl-openai, adding a notification when import completed and refresh the plugins state on TUI

alexsong-oai:
@codex review

PR #18254, opened 2026-04-17
uds: add async Unix socket crate

euroelessar:
## Summary
- add a codex-uds crate with async UnixListener and UnixStream wrappers
- expose helpers for private socket directory setup and stale socket path checks
- migrate codex-stdio-to-uds onto codex-uds and Tokio-based stdio/socket relaying
- update the CLI stdio-to-uds command path for the async runner

## Tests
- cargo test -p codex-uds -p codex-stdio-to-uds
- cargo test -p codex-cli
- just fmt
- just fix -p codex-uds
- just fix -p codex-stdio-to-uds
- just fix -p codex-cli
- just bazel-lock-check
- git diff --check

### DIFF
--- a/codex-rs/stdio-to-uds/tests/stdio_to_uds.rs
+++ b/codex-rs/stdio-to-uds/tests/stdio_to_uds.rs
@@ -1,6 +1,5 @@
 use std::io::ErrorKind;
 use std::io::Read;
-use std::io::Write;
 use std::process::Command;
 use std::process::Stdio;
 use std::sync::mpsc;
@@ -10,13 +9,10 @@ use std::time::Instant;
 
 use anyhow::Context;
 use anyhow::anyhow;
+use codex_uds::UnixListener;
 use pretty_assertions::assert_eq;
-
-#[cfg(unix)]
-use std::os::unix::net::UnixListener;
-
-#[cfg(windows)]
-use uds_windows::UnixListener;
+use tokio::io::AsyncReadExt;
+use tokio::io::AsyncWriteExt;
 
 #[test]
 fn pipes_stdin_and_stdout_through_socket() -> anyhow::Result<()> {
@@ -32,7 +28,8 @@ fn pipes_stdin_and_stdout_through_socket() -> anyhow::Result<()> {
     let request = b"request";
     let request_path = dir.path().join("request.txt");
     std::fs::write(&request_path, request).context("failed to write child stdin fixture")?;
-    let listener = match UnixListener::bind(&socket_path) {
+    let runtime = tokio::runtime::Runtime::new().context("failed to create tokio runtime")?;
+    let listener = match runtime.block_on(UnixListener::bind(&socket_path)) {
         Ok(listener) => listener,
         Err(err) if err.kind() == ErrorKind::PermissionDenied => {
             eprintln!("skipping test: failed to bind unix socket: {err}");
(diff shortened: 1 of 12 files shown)


codex-rs/uds/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+            return Err(io::Error::new(
+                ErrorKind::AlreadyExists,
+                format!(
+                    "socket directory path exists and is not a directory: {}",
+                    socket_dir.display()
+                ),
+            ));
+        }
+
+        let permissions = metadata.permissions();
```
viyatb-oai, line 127: In the SSH-over-UDS remote-control model, the directory contract should be exactly `0700`, not only "no group/other bits". As written, an existing `$CODEX_HOME/app-server-control` with mode `0600` passes this check even though the owner cannot traverse it. I would compare `(mode & 0o777) != SOCKET_DIR_MODE` and set it to `0700`, maybe also add a regression test for an existing `0600` directory.

viyatb-oai (commented):
Reviewing in the context of SSH remote control over Unix sockets.

codex-rs/stdio-to-uds/tests/stdio_to_uds.rs
```diff
@@ -46,20 +43,21 @@ fn pipes_stdin_and_stdout_through_socket() -> anyhow::Result<()> {
     let (tx, rx) = mpsc::channel();
     let (event_tx, event_rx) = mpsc::channel();
     let server_thread = thread::spawn(move || -> anyhow::Result<()> {
```
bolinfest: Why don't we make this an async task now?

codex-rs/stdio-to-uds/tests/stdio_to_uds.rs
```diff
@@ -10,13 +9,10 @@ use std::time::Instant;
...
-#[cfg(unix)]
-use std::os::unix::net::UnixListener;
-
-#[cfg(windows)]
-use uds_windows::UnixListener;
+use tokio::io::AsyncReadExt;
+use tokio::io::AsyncWriteExt;
 
 #[test]
 fn pipes_stdin_and_stdout_through_socket() -> anyhow::Result<()> {
```
bolinfest: make this a tokio::test?

codex-rs/uds/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+use std::task::Context;
+use std::task::Poll;
+
+use tokio::io::AsyncRead;
+use tokio::io::AsyncWrite;
+use tokio::io::ReadBuf;
+
+/// Creates `socket_dir` if needed and restricts it to the current user where
+/// the platform exposes Unix permissions.
+pub fn prepare_private_socket_directory(socket_dir: impl AsRef<Path>) -> IoResult<()> {
```
bolinfest: Since it seems like we're doing a bunch of work to provide a `UnixListener` abstraction that is `async`, should this and the next function be `async`, as well?

codex-rs/uds/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+    use std::io::Result as IoResult;
+    use std::os::unix::fs::DirBuilderExt;
+    use std::os::unix::fs::FileTypeExt;
+    use std::os::unix::fs::PermissionsExt;
+    use std::path::Path;
+
+    use tokio::net::UnixListener;
+    use tokio::net::UnixStream;
+
+    const SOCKET_DIR_MODE: u32 = 0o700;
```
bolinfest, line 100: Docstring to clarify why this is the desired mode.

codex-rs/uds/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+                ErrorKind::AlreadyExists,
+                format!(
+                    "socket directory path exists and is not a directory: {}",
+                    socket_dir.display()
+                ),
+            ));
+        }
+
+        let permissions = metadata.permissions();
+        if permissions.mode() & 0o077 != 0 {
```
bolinfest: Is there a cleaner way to express this as the complement of `SOCKET_DIR_MODE`?

codex-rs/uds/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+            io::Write::flush(&mut self.0)
+        }
+    }
+
+    unsafe impl async_io::IoSafe for WindowsUnixListener {}
+    unsafe impl async_io::IoSafe for WindowsUnixStream {}
+}
+
+#[cfg(test)]
+mod tests {
```
bolinfest: we've started putting tests in a sibling `_tests.rs` file. this is helpful when looking at grep output and other things.

bolinfest (commented):
Can you check on the P1 reported by Codex?

codex-rs/uds/src/lib.rs
bolinfest: @euroelessar seems worth capturing this in a comment with this code?

bolinfest (approved):

PR #18255, opened 2026-04-17
app-server: add Unix socket transport

euroelessar:
## Summary
- add unix:// app-server transport backed by the shared codex-uds crate
- reuse the websocket connection loop for axum and tungstenite-backed streams
- add codex app-server proxy to bridge stdio clients to the control socket
- tolerate Windows UDS backends that report a missing rendezvous path as connection refused before binding

## Tests
- cargo test -p codex-app-server control_socket_acceptor_forwards_websocket_text_messages_and_pings
- just fmt
- just fix -p codex-app-server
- git -c core.fsmonitor=false diff --check


codex-rs/app-server/src/transport/unix_socket.rs
```diff
@@ -0,0 +1,260 @@
...
+use tracing::info;
+use tracing::warn;
+
+pub(crate) async fn start_control_socket_acceptor(
+    socket_path: PathBuf,
+    transport_event_tx: mpsc::Sender<TransportEvent>,
+    shutdown_token: CancellationToken,
+) -> IoResult<JoinHandle<()>> {
+    prepare_control_socket_path(&socket_path).await?;
+    let listener = UnixListener::bind(&socket_path).await?;
```
viyatb-oai: For the SSH-over-UDS remote-control boundary, the socket file itself should be chmodded to `0600` after bind. The private `0700` parent directory is the main protection, but this listener currently leaves the socket node permissions to process `umask`, which can produce a broader mode. I would set the socket path permissions immediately after a successful bind and add a Unix-only test that asserts `mode & 0o777 == 0o600`.

viyatb-oai (commented):
Reviewing in the context of SSH remote control over Unix sockets.

codex-rs/app-server/src/transport/mod.rs
```diff
@@ -31,23 +33,42 @@ pub(crate) const CHANNEL_CAPACITY: usize = 128;
...
+            .join(APP_SERVER_CONTROL_SOCKET_FILE_NAME),
+    )
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub enum AppServerTransport {
     Stdio,
-    WebSocket { bind_address: SocketAddr },
+    UnixSocket {
+        socket_path: Option<AbsolutePathBuf>,
```
bolinfest: I would say this should be defined at construction so that the user of `AppServerTransport` never needs a reference to `$CODEX_HOME`.

```suggestion
        socket_path: AbsolutePathBuf,
```

codex-rs/cli/src/main.rs
```diff
@@ -2198,6 +2225,32 @@ mod tests {
         );
     }
 
+    #[test]
+    fn app_server_listen_unix_socket_url_parses() {
+        let app_server =
+            app_server_from_args(["codex", "app-server", "--listen", "unix://"].as_ref());
+        assert_eq!(
+            app_server.listen,
+            codex_app_server::AppServerTransport::UnixSocket { socket_path: None }
```
bolinfest: The only time you are constructing with `None` is in a test, anyway?

codex-rs/app-server/src/transport/unix_socket.rs
```diff
@@ -0,0 +1,314 @@
...
+            warn!(
+                socket_path = %self.socket_path.display(),
+                %err,
+                "failed to remove app-server control socket file"
+            );
+        }
+    }
+}
+
+#[cfg(test)]
```
bolinfest: Prefer `unix_socket_tests.rs` with the `path=` thing. I need to codemod the rest of the repo so it looks like `codex-core`.

bolinfest (approved):
Just some small things: nice job with the tests!

codex-rs/cli/src/main.rs
euroelessar: we're also constructing it while parsing `--listen` flag, at which point we don't have the config yet, so can't materialize the path

PR #18094, opened 2026-04-16
[codex] Use background agent task auth for backend calls

adrian-openai:
## Summary

Introduces a single background/control-plane agent task for ChatGPT backend requests that do not have a thread-scoped task, with `AuthManager` owning the default ChatGPT backend authorization decision.

Callers now ask `AuthManager` for the default ChatGPT backend authorization header. `AuthManager` decides whether that is bearer or background AgentAssertion based on config/internal state, while low-level bootstrap paths can explicitly request bearer-only auth.

This PR is stacked on PR4 and focuses on the shared background task auth plumbing plus the first tranche of backend/control-plane consumers. The remaining callsite wiring is split into PR4.2 to keep review size down.

## Stack

- PR1: https://github.com/openai/codex/pull/17385 - add `features.use_agent_identity`
- PR2: https://github.com/openai/codex/pull/17386 - register agent identities when enabled
- PR3: https://github.com/openai/codex/pull/17387 - register agent tasks when enabled
- PR3.1: https://github.com/openai/codex/pull/17978 - persist and prewarm registered tasks per thread
- PR4: https://github.com/openai/codex/pull/17980 - use task-scoped `AgentAssertion` for downstream calls
- PR4.1: this PR - introduce AuthManager-owned background/control-plane `AgentAssertion` auth
- PR4.2: https://github.com/openai/codex/pull/18260 - use background task auth for additional backend/control-plane calls

## What Changed

- add background task registration and assertion minting inside `codex-login`
- persist `agent_identity.background_task_id` separately from per-session task state
- make `BackgroundAgentTaskManager` private to `codex-login`; call sites do not instantiate or pass it around
- teach `AuthManager` the ChatGPT backend base URL and feature-derived background auth mode from resolved config
- expose bearer-only helpers for bootstrap/registration/refresh-style paths that must not use AgentAssertion
- wire `AuthManager` default ChatGPT authorization through app listing, connector directory listing, remote plugins, MCP status/listing, analytics, and core-skills remote calls
- preserve bearer fallback when the feature is disabled, the backend host is unsupported, or background task registration is not available

## Validation

- `just fmt`
- `cargo check -p codex-core -p codex-login -p codex-analytics -p codex-app-server -p codex-cloud-requirements -p codex-cloud-tasks -p codex-models-manager -p codex-chatgpt -p codex-model-provider -p codex-mcp -p codex-core-skills`
- `cargo test -p codex-login agent_identity`
- `cargo test -p codex-model-provider bearer_auth_provider`
[...]

PR #18260, opened 2026-04-17
[codex] Use background task auth for additional backend calls

adrian-openai:
## Summary

Splits the larger PR4.1 background task auth rollout by moving additional backend/control-plane call sites into this downstream PR.

This PR keeps callers on the same design as PR4.1: most code asks `AuthManager` for the default ChatGPT backend authorization header, and `AuthManager` decides bearer vs background AgentAssertion internally. Task-pinned inference auth remains separate because it needs the thread's registered task id.

## Stack

- PR4: https://github.com/openai/codex/pull/17980 - use task-scoped `AgentAssertion` for downstream calls
- PR4.1: https://github.com/openai/codex/pull/18094 - introduce AuthManager-owned background/control-plane `AgentAssertion` auth
- PR4.2: this PR - use background task auth for additional backend/control-plane calls

## What Changed

- pass full authorization header values through backend-client and cloud-tasks-client call paths where needed
- move ChatGPT client, cloud requirements, cloud tasks, thread-manager, and models-manager background auth usage into this downstream slice
- make app-server remote control enrollment/websocket auth ask `AuthManager` for the local backend authorization header instead of threading a background auth mode through transport options
- keep the same feature-gated bearer fallback behavior from PR4.1

## Validation

- `just fmt`
- `cargo check -p codex-core -p codex-login -p codex-analytics -p codex-app-server -p codex-cloud-requirements -p codex-cloud-tasks -p codex-models-manager -p codex-chatgpt -p codex-model-provider -p codex-mcp -p codex-core-skills`
- `cargo test -p codex-login agent_identity`
- `cargo test -p codex-model-provider bearer_auth_provider`
- `cargo test -p codex-core agent_assertion`
- `cargo test -p codex-app-server remote_control`
- `cargo test -p codex-cloud-requirements fetch_cloud_requirements`
- `cargo test -p codex-models-manager manager::tests`
- `cargo test -p codex-chatgpt`
- `cargo test -p codex-cloud-tasks`
- `just fix -p codex-core -p codex-login -p codex-analytics -p codex-app-server -p codex-cloud-requirements -p codex-cloud-tasks -p codex-models-manager -p codex-chatgpt -p codex-model-provider -p codex-mcp -p codex-core-skills`
- `git diff --check`

### DIFF
--- a/codex-rs/app-server/src/lib.rs
+++ b/codex-rs/app-server/src/lib.rs
@@ -27,10 +27,11 @@ use crate::outgoing_message::QueuedOutgoingMessage;
 use crate::transport::CHANNEL_CAPACITY;
 use crate::transport::ConnectionState;
 use crate::transport::OutboundConnectionState;
+use crate::transport::RemoteControlStartOptions;
 use crate::transport::TransportEvent;
 use crate::transport::auth::policy_from_settings;
 use crate::transport::route_outgoing_envelope;
-use crate::transport::start_remote_control;
+use crate::transport::start_remote_control_with_options;
 use crate::transport::start_stdio_connection;
 use crate::transport::start_websocket_acceptor;
 use codex_analytics::AppServerRpcTransport;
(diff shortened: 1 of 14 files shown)


codex-rs/app-server/src/lib.rs
```diff
@@ -578,16 +583,20 @@ pub async fn run_main_with_transport(
...
+            remote_control_url: config.chatgpt_base_url.clone(),
+            state_db: state_db.clone(),
+            auth_manager: auth_manager.clone(),
+            transport_event_tx: transport_event_tx.clone(),
+            shutdown_token: transport_shutdown_token.clone(),
+            app_server_client_name_rx,
+            initial_enabled: remote_control_enabled,
+            background_agent_task_auth_mode: BackgroundAgentTaskAuthMode::from_feature_enabled(
+                config.features.enabled(Feature::UseAgentIdentity),
+            ),
```
shijie-oai: I don't think we should be worried about if the remote control should be using anget identity or not? The app server on the remote device should have its own feature enabled?

adrian-openai: The added mode applies to the local app-server process opening the remote-control websocket to the ChatGPT/control-plane backend. It is not a setting sent across the websocket to tell the other end what to do.

The flow is:
- run_main_with_transport has the app-server Config.
- That config knows whether Feature::UseAgentIdentity is enabled.
- The remote-control transport needs to build the Authorization header for enrollment/websocket connection.
- The low-level websocket connector does not have Config, so the feature-derived BackgroundAgentTaskAuthMode gets threaded down.
- If enabled, it may use a background AgentAssertion authorization header.
- If disabled or unsupported, it falls back to bearer.
- So the remote side cannot “respect its own config” for this decision because this decision happens before the request is sent. The remote side can validate the auth it receives, but it does not choose whether this client process sends bearer vs AgentAssertion.

Let me know if that makes sense. I was wrong about this being part of slingshot!

shijie-oai (approved):

PR #18261, opened 2026-04-17
fix(tui): use BEL for terminal title updates

### LINKED ISSUE #18160: Terminal/tab suffix shows codex" (stray quote in process/app name)

codex-cli 0.120.0

ChatGPT Enterprise

gpt-5.4

Darwin 25.3.0 arm64 arm

iTerm2

In iTerm2, my tab titles append the current foreground process in parentheses (e.g. `(zsh)`, `(less)`, etc.). When running Codex, it shows `(codex")` with a trailing quote.

This looks like Codex is setting/reporting an incorrect process name (e.g., argv[0]/process title) that includes a `"`.

**Expected:** `(codex)`  
**Actual:** `(codex")`

Run codex-cli and send any prompt. The terminal will show the name of the current directory:

```bash
❯ mkdir -p temp && cd temp && codex -- hi
```

doing ☝️ will produce `temp (codex")` in the iterm tab title

The expected title for the same command should be `temp (codex)`

### Additional information

_No response_

etraut-openai:
## Summary

Fixes #18160.

iTerm2 can append the current foreground process to tab titles, and Codex's terminal-title updates were causing that decoration to appear as `(codex")` with a stray trailing quote. Codex was writing OSC 0 title sequences terminated with ST (`ESC \`). Some terminal title integrations appear to accept that title update but still expose the ST terminator in their own process/title decoration.

## Changes

- Update `codex-rs/tui/src/terminal_title.rs` to terminate OSC 0 title updates with BEL instead of ST.
- Update the focused terminal-title encoding test to assert the BEL-terminated sequence.

## Compatibility

This should be low risk: the title payload and update timing are unchanged, and BEL is the form already emitted by `crossterm::terminal::SetTitle` in the crossterm version used by this repository. BEL is also the widely supported xterm-family title terminator used by common terminals and multiplexers. The main theoretical risk would be a very old or unusual terminal that accepted only ST and not BEL for OSC title termination, but that is unlikely compared with the observed iTerm2 issue.

## Verification

- `cargo test -p codex-tui terminal_title`
- `cargo test -p codex-tui`

### DIFF
--- a/codex-rs/tui/src/terminal_title.rs
+++ b/codex-rs/tui/src/terminal_title.rs
@@ -84,9 +84,10 @@ struct SetWindowTitle(String);
 
 impl Command for SetWindowTitle {
     fn write_ansi(&self, f: &mut impl fmt::Write) -> fmt::Result {
-        // xterm/ctlseqs documents OSC 0/2 title sequences with ST (ESC \) termination.
-        // Most terminals also accept BEL for compatibility, but ST is the canonical form.
-        write!(f, "\x1b]0;{}\x1b\\", self.0)
+        // Match crossterm's SetTitle command and terminate OSC 0 with BEL.
+        // Some terminal title integrations expose the ST terminator in process
+        // decorations even though they otherwise accept the title update.
+        write!(f, "\x1b]0;{}\x07", self.0)
     }
 
     #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


fcoury-oai (approved):
Replicated the issue by unselecting all items on the title picker and was able to also confirm the fix.

PR #18263, opened 2026-04-17
enable tool search over dynamic tools

sayan-oai:
## Summary

- Normalize deferred MCP and dynamic tools into `ToolSearchEntry` values before constructing `ToolSearchHandler`.
- Move the tool-search entry adapter out of `tools/handlers` and into `tools/tool_search_entry.rs` so the handlers directory stays focused on handlers.
- Keep `ToolSearchHandler` operating over one generic entry list for BM25 search, namespace grouping, and per-bucket default limits.

## Why

Follow-up cleanup for #17849. The dynamic tool-search support made the handler juggle source-specific MCP and dynamic tool lists, index arithmetic, output conversion, and namespace emission. This keeps source adaptation outside the handler so the search loop itself is smaller and source-agnostic.

## Validation

- `just fmt`
- `cargo test -p codex-core tools::handlers::tool_search::tests`
- `git diff --check`
- `cargo test -p codex-core` currently fails in unrelated `plugins::manager::tests::list_marketplaces_ignores_installed_roots_missing_from_config`; rerunning that single test fails the same way at `core/src/plugins/manager_tests.rs:1692`.


codex-rs/core/src/tools/tool_search_entry.rs
```diff
@@ -0,0 +1,169 @@
...
+
+#[derive(Clone)]
+pub(crate) struct ToolSearchEntry {
+    pub(crate) search_text: String,
+    pub(crate) output: ToolSearchEntryOutput,
+    pub(crate) limit_bucket: Option<String>,
+}
+
+#[derive(Clone)]
+pub(crate) enum ToolSearchEntryOutput {
```
pakrym-oai: ToolSearchOutputTool?

codex-rs/core/src/tools/tool_search_entry.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    entries
+}
+
+fn mcp_tool_search_entry(info: &ToolInfo) -> Result<ToolSearchEntry, serde_json::Error> {
+    let tool_name = info.canonical_tool_name();
+    let tool = mcp_tool_to_deferred_responses_api_tool(&tool_name, &info.tool)
+        .map(ResponsesApiNamespaceTool::Function)?;
+    let namespace_description = info
```
pakrym-oai: can we reuse as much logic as possible from the main tool building path?

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -4,33 +4,31 @@ use crate::tools::context::ToolPayload;
...
-
+    pub(crate) fn new(entries: Vec<ToolSearchEntry>) -> Self {
         let documents: Vec<Document<usize>> = entries
             .iter()
+            .map(|entry| entry.search_text.clone())
             .enumerate()
-            .map(|(idx, (name, info))| Document::new(idx, build_search_text(name, info)))
+            .map(|(idx, search_text)| Document::new(idx, search_text))
             .collect();
         let search_engine =
```
pakrym-oai, line 33: random thought. we rebuild this per request. is it slow? cpu intensive?

pakrym-oai: are we benefiting from the index?

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -83,157 +81,226 @@ impl ToolHandler for ToolSearchHandler {
...
 impl ToolSearchHandler {
-    fn search_result_entries(
+    fn search(
         &self,
         query: &str,
         limit: usize,
         use_default_limit: bool,
-    ) -> Vec<&(String, ToolInfo)> {
+    ) -> Result<Vec<ToolSearchOutputTool>, FunctionCallError> {
+        let result_ids = self.search_result_ids(query, limit, use_default_limit);
```
pakrym-oai: another random thought, the index should be ToolName
non blocking

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -83,157 +81,226 @@ impl ToolHandler for ToolSearchHandler {
...
+        // Preserve search order: group namespace children, emit standalone tools directly.
+        for result_id in result_ids {
+            let Some(entry) = self.entries.get(result_id) else {
+                continue;
+            };
+            match &entry.output {
+                ToolSearchOutputTool::Function(tool) => {
+                    tools.push(ToolSearchOutputTool::Function(tool.clone()));
+                }
+                ToolSearchOutputTool::Namespace(namespace) => {
```
pakrym-oai, line 146: why do we need this complicated logic? are we creating duplicated namespaces somewhere?

pakrym-oai: do we do the same namespace merging on normal path?

codex-rs/core/src/tools/handlers/tool_search.rs
```diff
@@ -83,157 +81,226 @@ impl ToolHandler for ToolSearchHandler {
...
             .collect::<Vec<_>>();
         if !use_default_limit {
             return results;
         }
 
-        if results
-            .iter()
-            .any(|(_, tool)| tool.server_name == COMPUTER_USE_MCP_SERVER_NAME)
-        {
+        if results.iter().any(|&id| {
```
pakrym-oai: we already convert ids into entries here
and then in limit_results_by_bucket
and then in search_output_tools

should we carry around entries?

pakrym-oai (approved):

codex-rs/tools/src/tool_discovery.rs
```diff
@@ -224,39 +224,55 @@ pub fn collect_tool_search_output_tools<'a>(
...
+            description: tool_search_result_source_namespace_description(*first_tool),
             tools,
         }));
     }
 
     Ok(results)
 }
 
+pub fn tool_search_result_source_to_output_tool(
+    source: ToolSearchResultSource<'_>,
```
pakrym-oai, line 207: why do we have ToolSearchResultSource too?

codex-rs/core/src/tools/tool_search_entry.rs
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+
+    entries
+}
+
+fn mcp_tool_search_entry(info: &ToolInfo) -> Result<ToolSearchEntry, serde_json::Error> {
+    Ok(ToolSearchEntry {
+        search_text: build_mcp_search_text(info),
+        output: tool_search_result_source_to_output_tool(ToolSearchResultSource {
```
pakrym-oai, line 58: Don't we have existing logic to turn tool info into  responsesapi tool?

codex-rs/tools/src/tool_discovery.rs
```diff
@@ -224,39 +224,55 @@ pub fn collect_tool_search_output_tools<'a>(
...
-            .or_else(|| {
-                first_tool
-                    .connector_name
-                    .map(str::trim)
-                    .filter(|connector_name| !connector_name.is_empty())
-                    .map(|connector_name| format!("Tools for working with {connector_name}."))
-            });
-
         let tools = tools
             .iter()
```
pakrym-oai: Is there a prod usage of this method?

codex-rs/core/src/tools/handlers/tool_search.rs
sayan-oai: its to avoid duplicate namespace entries while preserving search result order. dynamic tools can interleave results from the same namespace, so we merge later hits into the first namespace entry. the ideas being:

1. we want to preserve search ordering
2. we dont want to show the model repeated entries for the same namespace

codex-rs/core/src/tools/handlers/tool_search.rs
sayan-oai: good point, right now it's rebuilt per sampling request. will look into lazy loading, or maybe caching

codex-rs/core/src/tools/tool_search_entry.rs
sayan-oai: not directly, no. this is the most direct existing conversion path

codex-rs/tools/src/tool_discovery.rs
sayan-oai: no, will rm

sayan-oai: thanks for catching

codex-rs/core/src/tools/handlers/tool_search.rs
sayan-oai: maybe, but then you have a ToolName -> ToolSearchEntry map. and now that we pass around entries rather than ids as you suggested, the numeric index is contained to the entry list

codex-rs/core/src/tools/handlers/tool_search.rs
sayan-oai: good point

codex-rs/tools/src/tool_discovery.rs
sayan-oai: adding this to the list to cleanup, it seems like it was introduced for mcp-specific tool search. now its the most direct conversion to responses type

sayan-oai:
follow ups i noted but dont want to block this on:
- lazy load bm25 index?
- clean up ToolSearchResultSource
- is it worth indexing tool search entries by ToolName rather than usize?

PR #18265, opened 2026-04-17
Avoid instant remote disconnects under websocket queue pressure

### LINKED ISSUE #18203: App-server disconnects remote TUI mid-turn when outbound queue fills (128 messages)

v0.121.0

None, API Keys

gpt-5.4

Darwin 25.4.0 arm64 arm

Warp

The app-server's WebSocket transport disconnects remote TUI clients instantly when its 128-message outbound queue fills, instead of applying backpressure. This makes `--remote` mode unusable for any turn that produces moderate output volume.

The TUI exits with:

```
ERROR: remote app server at `ws://127.0.0.1:<port>/` transport failed:
  WebSocket protocol error: Connection reset without closing handshake
```

The app-server logs:

```
disconnecting slow connection after outbound queue filled: ConnectionId(0)
```

This reproduces on stock upstream Codex with a single remote TUI client and no third-party tooling.

Terminal 1 — start the app-server:

```bash
codex app-server --listen ws://127.0.0.1:0
# note the printed ws:// URL
```

Terminal 2 — attach the remote TUI:

```bash
codex --remote ws://127.0.0.1:<port>
```

Then send a prompt that produces moderate streaming output:

```
Run `rg -n '120|timeout' ~/.codex` and answer briefly.
```

The TUI disconnects within seconds. Smaller prompts sometimes survive, but any turn with realistic tool output triggers it.

The remote TUI should remain connected during normal turns. A client that falls 128 messages behind during streaming is momentarily slower than the producer, not stuck. it should not be terminated instantly.

Stdio clients already handle this correctly: they use blocking `.send().await` (`stdio.rs:35` sets `disconnect_sender: None`), so they never hit the disconnect path. WebSocket clients should get equivalent resilience, either through a larger queue, a send timeout, or per-connection send tasks.

### Additional information

root cause is in `send_message_to_connection()` at `codex-rs/app-server/src/transport/mod.rs` line 308-316. WebSocket clients are marked disconnectable (`websocket.rs:176` sets `disconnect_sender: Some(...)`), so when `try_send()` returns `TrySendError::Full` on the 128-slot bounded channel, the connection is terminated immediately.

The outbound router is a single `select!` loop (`lib.rs:643`), so blocking on a slow WebSocket is not safe. But the `remote_control` transport already solves this differently, `remote_control/websocket.rs` uses per-stream `BoundedOutboundBuffer` with backpressure instead of instant disconnect.

Design options that preserve the non-blocking router:

1. **Per-connection send tasks.** Structurally similar to `remote_control`'s existing approach.
2. **Larger queue + send_timeout.** Simplest fix, small blast radius.
3. **Priority-based notification dropping under queue pressure.** `opt_out_notification_methods` already has the right shape.

The existing test `broadcast_does_not_block_on_slow_connection` (line 878) uses channel capacity 1 and validates instant disconnect as expected behavior. That test encodes the wrong contract — it should validate a grace period, not instant termination.

Related: #13949 (static analysis of same path), #15355 (local ingress feature request).

Reproduced on `codex-cli 0.120.0` and `0.121.0`, both stock upstream and with a third-party wrapper using `--remote` mode.

etraut-openai:
Fixes #18203.

## Why

Remote TUI clients connected through `codex app-server --listen ws://...` can receive bursts of outbound turn/progress notifications. Before this change, a full per-connection writer queue was treated as an immediate slow-client signal, so a healthy WebSocket client that was only momentarily behind could be disconnected mid-turn.

## What Changed

Disconnectable WebSocket connections now get a bounded grace window when their outbound writer queue is full. If the queue drains within that window, the message is delivered and the connection stays open. If it remains full, the connection is cancelled so the outbound router cannot be blocked indefinitely.

The implementation also adds a per-connection bounded overflow channel for WebSocket sends. This keeps the shared outbound router from waiting on one slow connection, preserves message order for that connection while the overflow drains, and avoids creating an unbounded backlog of detached send waiters.

When a connection times out or its writer closes, the overflow worker now notifies the outbound router so the connection is removed from routing promptly instead of waiting for the later `ConnectionClosed` event. The overflow-channel-full path also uses the same grace window before disconnecting, so short bursts that saturate both bounded queues do not immediately drop the client.

The stdio and in-process transports keep their existing reliable-delivery behavior; the grace/overflow path applies to disconnectable WebSocket connections.

## Testing

Added transport tests covering:

- temporary WebSocket queue pressure draining without disconnecting
- per-connection message ordering while overflow messages drain
- disconnect after the queue remains full past the grace window
- applying the grace window before disconnecting when both the writer queue and overflow queue are saturated

I also manually confirmed that the repro methodology from #18203 disconnects before the fix and stays connected with this change.


maxj-oai (approved):
thank you @etraut-openai

PR #18267, opened 2026-04-17
Support Ctrl+P/Ctrl+N in resume picker

### LINKED ISSUE #18179: Support Ctrl+P / Ctrl+N navigation in /resume picker

codex-cli 0.121.0

OpenAI API key, Enterprise

_No response_

Darwin 25.4.0 arm64 arm

Ghostty

When I trigger /resume and the session picker opens, I can navigate with Up/Down arrows, but Ctrl+P and Ctrl+N do not move the selection. Those shortcuts already work in other Codex TUI flows like ListSelectionView-backed pickers and the textarea/composer, so /resume feels inconsistent.

1. Start Codex CLI.
2. Trigger /resume to open the saved-session picker.
3. Press Ctrl+P or Ctrl+N.
4. Notice that the selection does not move, while Up/Down arrows do.

Ctrl+P should move to the previous session and Ctrl+N should move to the next session, matching the existing keyboard behavior in other Codex TUI pickers and the textarea.

### Additional information

This appears to be isolated to the fullscreen resume picker path rather than the shared popup/list selection components. Up/Down already work there; only the Ctrl+P/Ctrl+N aliases are missing.

etraut-openai:
Fixes #18179.

## Why
The fullscreen `/resume` picker accepted Up/Down navigation but ignored Ctrl+P/Ctrl+N, which made it inconsistent with other TUI selection flows such as `ListSelectionView`-backed pickers and composer navigation.

## What Changed
Updated `codex-rs/tui/src/resume_picker.rs` so the resume picker treats Ctrl+P/Ctrl+N as aliases for Up/Down, including the raw `^P`/`^N` control-character events some terminals emit without a CONTROL modifier.


codex-rs/tui/src/resume_picker.rs
```diff
@@ -2738,6 +2782,56 @@ mod tests {
         assert_eq!(state.selected, state.filtered_rows.len().saturating_sub(2));
     }
 
+    #[tokio::test]
+    async fn ctrl_p_and_ctrl_n_browse_rows() {
```
etraut-openai: I don't think we need regression tests for this. It's not worth it. Remove the new test(s).

fcoury-oai (approved):
Tested and works. Code looks good. Approved.

PR #18271, opened 2026-04-17
show bash mode in the TUI

abhinav-oai:
## What

- Explicitly show our "bash mode" by changing the color and adding a callout similar to how we do for `Plan mode (shift + tab to cycle)`
- Also replace our `›` composer prefix with a bang `!`

## Why

- It was unclear that we had a Bash mode
- This feels more responsive
- It looks cool!


codex-rs/tui/src/bottom_pane/chat_composer.rs
```diff
@@ -2701,7 +2772,13 @@ impl ChatComposer {
...
 
     fn is_bang_shell_command(&self) -> bool {
-        self.textarea.text().trim_start().starts_with('!')
+        self.current_text().trim_start().starts_with('!')
+    }
+
+    fn shell_mode_footer_line(&self) -> Option<Line<'static>> {
+        self.is_bang_shell_command()
+            .then_some(())
+            .map(|_| Line::from(vec![Span::from("Bash mode").light_red()]))
```
abhinav-oai, line 2913: btw the copy is just this now, no more `(! to toggle)` since it's not really a toggle but actually a prefix

fcoury-oai (changes requested):
I found one bug: when you open an external editor and leaves without saving it returns the command with double `!`.

fcoury-oai (approved):
Looks good! 👍

PR #18274, opened 2026-04-17
protocol: canonicalize file system permissions

bolinfest:
## Why

`PermissionProfile` needs stable, canonical file-system semantics before it can become the primary runtime permissions abstraction. Without a canonical form, callers have to keep re-deriving legacy sandbox maps and profile comparisons remain lossy or order-dependent.

## What changed

This adds canonicalization helpers for `FileSystemPermissions` and `PermissionProfile`, expands special paths into explicit sandbox entries, and updates permission request/conversion paths to consume those canonical entries. It also tightens the legacy bridge so root-wide write profiles with narrower carveouts are not silently projected as full-disk legacy access.

## Verification

- `cargo test -p codex-protocol root_write_with_read_only_child_is_not_full_disk_write -- --nocapture`
- `cargo test -p codex-sandboxing permission -- --nocapture`
- `cargo test -p codex-tui permissions -- --nocapture`


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1155,22 +1155,46 @@ impl From<CoreNetworkApprovalContext> for NetworkApprovalContext {
...
 impl From<AdditionalFileSystemPermissions> for CoreFileSystemPermissions {
     fn from(value: AdditionalFileSystemPermissions) -> Self {
-        Self {
-            read: value.read,
-            write: value.write,
+        if let Some(entries) = value.entries {
+            Self {
+                entries: entries
+                    .into_iter()
+                    .map(CoreFileSystemSandboxEntry::from)
```
viyatb-oai, line 1198: [P0] Define canonical filesystem entry types before using them

This new `entries` field references `FileSystemSandboxEntry` and the conversion references `CoreFileSystemSandboxEntry`, but neither type is defined/imported in this module. `cargo check -p codex-app-server-protocol` fails with E0425/E0433 at these lines, so the PR cannot build. The generated TypeScript also imports `./FileSystemSandboxEntry`, but no such generated file exists, so app-server protocol clients would be broken too. Please add the v2/core type aliases or wrapper types, wire the conversions, and regenerate the schema/TS fixtures.

codex-rs/protocol/src/models.rs
```diff
@@ -135,15 +143,126 @@ impl SandboxPermissions {
     }
 }
 
-#[derive(Debug, Clone, Default, Eq, Hash, PartialEq, Serialize, Deserialize, JsonSchema, TS)]
+#[derive(Debug, Clone, Default, Eq, Hash, PartialEq, JsonSchema, TS)]
 pub struct FileSystemPermissions {
-    pub read: Option<Vec<AbsolutePathBuf>>,
-    pub write: Option<Vec<AbsolutePathBuf>>,
+    pub entries: Vec<FileSystemSandboxEntry>,
 }
```
viyatb-oai, line 149: [P1] Finish migrating FileSystemPermissions constructors

`FileSystemPermissions` now only has `entries`, but this PR leaves many direct `FileSystemPermissions { read, write }` literals in test code across `codex-core` and `app-server`. Once the app-server-protocol compile blocker is fixed, those test targets will fail to compile because `read` and `write` are no longer fields. Convert the remaining call sites to `FileSystemPermissions::from_read_write_roots(...)` or explicit canonical entries in this same PR.

codex-rs/sandboxing/src/policy_transforms.rs
```diff
@@ -45,13 +44,29 @@ pub fn normalize_additional_permissions(
...
+                let path = match entry.path {
+                    FileSystemPath::Path { path } => FileSystemPath::Path {
+                        path: canonicalize_preserving_symlinks(path.as_path())
+                            .ok()
+                            .and_then(|path| AbsolutePathBuf::from_absolute_path(path).ok())
+                            .unwrap_or(path),
+                    },
+                    FileSystemPath::GlobPattern { pattern } => {
+                        FileSystemPath::GlobPattern { pattern }
+                    }
```
viyatb-oai, line 65: [P2] Reject unsupported glob grants in permission profiles

The new canonical permission path accepts `FileSystemPath::GlobPattern` entries regardless of access mode, but the sandbox stack only implements glob entries as deny-read rules: config parsing rejects non-`none` globs, unreadable glob expansion filters to `access == None`, and the legacy permission projection ignores non-path entries entirely. A client or model can now request and display an approved read/write glob permission through the new `entries` field, but the runtime will not grant matching read/write access, so the approval is misleading and the command can still fail. Please reject non-`None` glob entries during permission normalization, or keep them out of this API until read/write glob enforcement exists.

viyatb-oai (approved):
one remaining finding but otherwise looks in a good shape

PR #18275, opened 2026-04-17
sandboxing: intersect permission profiles semantically

bolinfest:
## Why

Permission approval responses must not be able to grant more access than the tool requested. Moving this flow to `PermissionProfile` means the comparison must be profile-shaped instead of `SandboxPolicy`-shaped, and cwd-relative special paths such as `:cwd` and `:project_roots` must stay anchored to the turn that produced the request.

## What changed

This implements semantic `PermissionProfile` intersection in `codex-sandboxing` for file-system and network permissions. The intersection accepts narrower path grants, rejects broader grants, preserves deny-read carve-outs and glob scan depth, and materializes cwd-dependent special-path grants to absolute paths before they can be recorded for reuse.

The request-permissions response paths now use that intersection consistently. App-server captures the request turn cwd before waiting for the client response, includes that cwd in the v2 approval params, and core stores the requested profile plus cwd for direct TUI/client responses and Guardian decisions before recording turn- or session-scoped grants. The TUI app-server bridge now preserves the app-server request cwd when converting permission approval params into core events.

## Verification

- `cargo test -p codex-sandboxing intersect_permission_profiles -- --nocapture`
- `cargo test -p codex-app-server request_permissions_response -- --nocapture`
- `cargo test -p codex-core request_permissions_response_materializes_session_cwd_grants_before_recording -- --nocapture`
- `cargo check --tests`
- `cargo test -p codex-tui app_server_request_permissions_preserves_file_system_permissions`


codex-rs/sandboxing/src/policy_transforms.rs
```diff
@@ -177,6 +191,74 @@ pub fn intersect_permission_profiles(
     }
 }
 
+fn granted_file_system_entry_within_request(
+    requested: &FileSystemPermissions,
+    requested_policy: &FileSystemSandboxPolicy,
+    granted_entry: &FileSystemSandboxEntry,
+    cwd: &Path,
+) -> bool {
+    if !granted_entry.access.can_read() {
```
viyatb-oai, line 270: [P1] Preserve deny entries when intersecting grants

The intersection still exits before considering granted `FileSystemAccessMode::None` entries because `can_read()` is false. Those entries are the carve-outs that keep a broad grant safe; for example, a client can approve `:root` write with `/secret` denied, but the stored grant becomes plain root write after this filter. Deny entries that are part of the granted/requested policy need to survive the intersection, or the effective permissions become broader than the reviewer approved.

codex-rs/sandboxing/src/policy_transforms.rs
```diff
@@ -146,21 +149,43 @@ pub fn merge_permission_profiles(
...
                 .collect();
+            for entry in requested_file_system
+                .entries
+                .iter()
+                .chain(granted_file_system.entries.iter())
+                .filter(|entry| entry.access == FileSystemAccessMode::None)
+            {
+                let entry = materialize_cwd_dependent_entry(entry, cwd);
+                if !entries.contains(&entry) {
+                    entries.push(entry);
```
viyatb-oai: [P2] Do not persist standalone deny entries from partial grants

The intersection now appends every `FileSystemAccessMode::None` entry from both the request and the response after filtering accepted read/write grants. That preserves carveouts for broad filesystem grants, but it also records denies when no filesystem grant survived. For example, if the model requests network plus `:cwd` write with `/repo/secrets` denied, and the client approves only network, the returned session grant still includes the `/repo/secrets` deny. Later commands then run with network enabled but unexpectedly lose read access to `/repo/secrets`. Deny entries should only be retained when attached to an accepted filesystem grant they constrain.

codex-rs/sandboxing/src/policy_transforms.rs
```diff
@@ -230,6 +255,95 @@ enum GlobScanDepth {
...
+    requested_policy: &FileSystemSandboxPolicy,
+    granted_entry: &FileSystemSandboxEntry,
+    cwd: &Path,
+) -> bool {
+    if !granted_entry.access.can_read() {
+        return false;
+    }
+
+    if let Some(path) = resolve_permission_path(&granted_entry.path, cwd) {
+        return access_covers(
```
viyatb-oai, line 279: [P2] Honor deny globs when accepting concrete grants

For concrete granted paths, this check asks `resolve_access_with_cwd()` whether the requested policy covers the path, but that resolver only considers resolved path/special entries and ignores `GlobPattern` deny entries. A request like `:cwd` write plus `**/*.env` deny can therefore accept a client response granting `/repo/.env` write because the broad `:cwd` entry covers it during intersection. The returned profile also carries the deny glob, so enforcement either later rejects the approved path or, on paths that consult `can_write_path_with_cwd()`, treats it as writable despite the requested carveout. The semantic subset check should reject concrete grants matched by requested deny globs before accepting the broad grant.

codex-rs/app-server/src/bespoke_event_handling.rs
```diff
@@ -908,6 +909,10 @@ pub(crate) async fn apply_bespoke_event_handling(
                     .note_permission_requested(&conversation_id.to_string())
                     .await;
                 let requested_permissions = request.permissions.clone();
+                let request_cwd = match request.cwd.clone() {
+                    Some(cwd) => cwd,
+                    None => conversation.config_snapshot().await.cwd,
+                };
```
viyatb-oai, line 915: [P2] Send request cwd to permission reviewers

The server computes `request_cwd` and uses it later to normalize the response, but the approval request sent to app-server clients still omits that cwd. Permission profiles can contain `:cwd`, `:project_roots`, and cwd-relative deny globs; in delegated flows the request cwd can differ from the visible thread cwd. A client can therefore approve `write :cwd` without knowing which concrete directory the server will apply. Include the request cwd in the approval params, or materialize cwd-dependent entries before sending the request to clients.

viyatb-oai (approved):
non blocking - there are some edge cases codex found that are not covered by subsequent prs in the stack

PR #18278, opened 2026-04-17
app-server: expose thread permission profiles

bolinfest:
## Why

The `PermissionProfile` migration needs app-server clients to see the same constrained permission model that core is using at runtime. Before this PR, thread lifecycle responses only exposed the legacy `SandboxPolicy` shape, so clients still had to infer active permissions from sandbox fields. That makes downstream resume, fork, and override flows harder to make `PermissionProfile`-first.

External sandbox policies are intentionally excluded from this canonical view. External enforcement cannot be round-tripped as a `PermissionProfile`, and exposing a lossy root-write profile would let clients accidentally change sandbox semantics if they echo the profile back later.

## What changed

- Adds the app-server v2 `PermissionProfile` wire shape, including filesystem permissions and glob scan depth metadata.
- Adds `PermissionProfileNetworkPermissions` so the profile response does not expose active network state through the older additional-permissions naming.
- Returns `permissionProfile` from thread start, resume, and fork responses when the active sandbox can be represented as a `PermissionProfile`.
- Keeps legacy `sandbox` in those responses for compatibility and documents `permissionProfile` as canonical when present.
- Makes lifecycle `permissionProfile` nullable and returns `null` for `ExternalSandbox` to avoid exposing a lossy profile.
- Regenerates the app-server JSON schema and TypeScript fixtures.

## Verification

- `cargo test -p codex-app-server thread_response_permission_profile_omits_external_sandbox -- --nocapture`
- `cargo check --tests -p codex-analytics -p codex-exec -p codex-tui`
- `just fix -p codex-app-server-protocol -p codex-app-server -p codex-analytics -p codex-exec -p codex-tui`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/18278).
* #18279
* __->__ #18278


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2697,6 +2697,7 @@ impl CodexMessageProcessor {
                     approval_policy: config_snapshot.approval_policy.into(),
                     approvals_reviewer: config_snapshot.approvals_reviewer.into(),
                     sandbox: config_snapshot.sandbox_policy.into(),
+                    permission_profile: config_snapshot.permission_profile.into(),
```
viyatb-oai: [P2] Do not expose lossy profiles for ExternalSandbox

This now publishes `permissionProfile` for every thread response, including threads whose `sandbox` is `externalSandbox`. `PermissionProfile::from_runtime_permissions()` cannot represent `FileSystemSandboxKind::ExternalSandbox`: it serializes that filesystem policy as root write, and converting the profile back yields a restricted root-write filesystem policy instead of external enforcement. This is not fixed above; later stack PRs start accepting and carrying `permissionProfile`, so non-TUI clients can round-trip the advertised value and change sandbox semantics. Omit or mark the profile for ExternalSandbox, or extend the profile model to carry the enforcement kind before exposing it as canonical thread state.

viyatb-oai (approved):

PR #18279, opened 2026-04-17
app-server: accept permission profile overrides

bolinfest:
## Why

`PermissionProfile` is becoming the canonical permissions shape shared by core and app-server. After app-server responses expose the active profile, clients need to be able to send that same shape back when starting, resuming, forking, or overriding a turn instead of translating through the legacy `sandbox`/`sandboxPolicy` shorthands.

This still needs to preserve the existing requirements/platform enforcement model. A profile-shaped request can be downgraded or rejected by constraints, but the server should keep the user's elevated-access intent for project trust decisions. Turn-level profile overrides also need to retain existing read protections, including deny-read entries and bounded glob-scan metadata, so a permission override cannot accidentally drop configured protections such as `**/*.env = deny`.

## What changed

- Adds optional `permissionProfile` request fields to `thread/start`, `thread/resume`, `thread/fork`, and `turn/start`.
- Rejects ambiguous requests that specify both `permissionProfile` and the legacy `sandbox`/`sandboxPolicy` fields, including running-thread resume requests.
- Converts profile-shaped overrides into core runtime filesystem/network permissions while continuing to derive the constrained legacy sandbox projection used by existing execution paths.
- Preserves project-trust intent for profile overrides that are equivalent to workspace-write or full-access sandbox requests.
- Preserves existing deny-read entries and `globScanMaxDepth` when applying turn-level `permissionProfile` overrides.
- Updates app-server docs plus generated JSON/TypeScript schema fixtures and regression coverage.

## Verification

- `cargo test -p codex-app-server-protocol schema_fixtures`
- `cargo test -p codex-core session_configuration_apply_permission_profile_preserves_existing_deny_read_entries`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/18279).
* #18288
* #18287
* #18286
* #18285
* #18284
* #18283
* #18282
* #18281
* #18280
* __->__ #18279


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -6757,6 +6796,7 @@ impl CodexMessageProcessor {
                             .approvals_reviewer
                             .map(codex_app_server_protocol::ApprovalsReviewer::to_core),
                         sandbox_policy: params.sandbox_policy.map(|p| p.to_core()),
+                        permission_profile: params.permission_profile.map(Into::into),
```
viyatb-oai: [P2] Do not start a turn after a rejected permissionProfile override

`submit_core_op` only enqueues the `OverrideTurnContext`; if `SessionConfiguration::apply` later rejects the new `permissionProfile`, the failure is emitted as a core error event by the submission loop, but this handler has already queued the `UserInput` and will still return a `turn/start` response. That is especially easy to hit with the new full-profile field because profiles can fail conversion when they request unsupported write roots or violate sandbox constraints. For example, a client trying to narrow a currently broad thread with an invalid `permissionProfile` can still get the turn started under the previous broader permissions. The override needs to be validated/applied synchronously before accepting the user input, or the request should fail instead of continuing with stale permissions.

codex-rs/core/src/config/mod.rs
```diff
@@ -1736,7 +1745,34 @@ impl Config {
             sandbox_policy,
             file_system_sandbox_policy,
             network_sandbox_policy,
-        ) = if profiles_are_active {
+        ) = if let Some(permission_profile) = permission_profile {
+            let configured_network_proxy_config = NetworkProxyConfig::default();
```
viyatb-oai: [P1] Preserve network proxy config for permissionProfile overrides

The permission-profile override path builds filesystem and network sandbox permissions from the supplied profile, but it resets the network proxy policy to `NetworkProxyConfig::default()`. The named-profile path just below preserves the profile network settings with `network_proxy_config_from_profile_network(profile.network.as_ref())`, including proxy URLs, SOCKS settings, domain and socket rules, and local-binding policy. In #18280 the clients send `config.permissions.permission_profile()` as this override, so a config that locally runs with a managed proxy or allowlist can be restarted through app-server with the same apparent `permissionProfile` but no proxy policy, broadening or de-auditing network access. This path needs to carry the configured profile network proxy state, or the API should not treat `permissionProfile` as a full config override until that policy is representable.

viyatb-oai (approved):

PR #18280, opened 2026-04-17
clients: send permission profiles to app-server

bolinfest:
## Why

After app-server can accept `PermissionProfile`, first-party clients should stop preferring legacy sandbox fields when canonical permission information is available. This keeps the migration moving without removing legacy compatibility yet.

The client side still has mixed surfaces during the stack: embedded thread start/resume/fork and exec initial turns can derive a profile directly from local config, while TUI remote sessions and some turn-start paths only have a legacy/server-context-safe sandbox projection. Those paths keep sending legacy sandbox fields rather than synthesizing or sending lossy/local-only profiles.

## What changed

- Sends `permissionProfile` from exec and embedded TUI thread start/resume/fork requests when config has a representable profile.
- Keeps legacy sandbox fallback for external sandbox policies, TUI remote thread lifecycle requests, and TUI turn-start requests that do not yet carry the active profile.
- Sends the actual config-derived `permissionProfile` for exec initial turns instead of rebuilding one from the legacy sandbox projection.
- Stores response `permissionProfile` as optional in TUI session state so external sandbox responses and compatibility payloads preserve `null`.
- Updates tests for request construction and response mapping.

## Verification

- `cargo check --tests -p codex-tui -p codex-exec`
- `cargo test -p codex-tui app_server_session -- --nocapture`
- `cargo test -p codex-exec thread_start_params -- --nocapture`
- `cargo test -p codex-tui app_server_session::tests::thread_lifecycle_params -- --nocapture`
- `just fix -p codex-tui -p codex-exec`
- `just fix -p codex-tui`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/18280).
* #18288
* #18287
* #18286
* #18285
* #18284
* #18283
* #18282
* #18281
* __->__ #18280

### DIFF
--- a/codex-rs/tui/src/app_server_session.rs
+++ b/codex-rs/tui/src/app_server_session.rs
@@ -85,6 +85,7 @@ use codex_app_server_protocol::TurnSteerParams;
 use codex_app_server_protocol::TurnSteerResponse;
 use codex_otel::TelemetryAuthMode;
 use codex_protocol::ThreadId;
+use codex_protocol::models::PermissionProfile;
 use codex_protocol::models::ResponseItem;
 use codex_protocol::openai_models::ModelAvailabilityNux;
 use codex_protocol::openai_models::ModelPreset;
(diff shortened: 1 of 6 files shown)


codex-rs/tui/src/app_server_session.rs
```diff
@@ -539,6 +541,9 @@ impl AppServerSession {
                     cwd: Some(cwd),
                     approval_policy: Some(approval_policy.into()),
                     approvals_reviewer: Some(approvals_reviewer.into()),
+                    // This path only receives the legacy projection today; do
+                    // not synthesize a lossy profile until the active profile
+                    // is plumbed through user turns.
                     sandbox_policy: Some(sandbox_policy.into()),
                     permission_profile: None,
```
viyatb-oai, line 560: This TUI turn/start path still sends `sandboxPolicy` with `permissionProfile: None`. After this PR, embedded thread start/resume/fork installs the full `permissionProfile`, but every ordinary prompt is an override, so app-server applies this legacy sandbox update before running the turn. `SessionConfiguration::apply` rebuilds the filesystem policy from `SandboxPolicy` and only preserves deny entries/depth, which discards any active profile state that is not represented by the legacy projection. Plumb the active `ThreadSessionState.permission_profile` into `turn_start` like the exec path does, or omit the sandbox override when there is no per-turn permission change.

viyatb-oai (approved):

PR #18282, opened 2026-04-17
protocol: report session permission profiles

bolinfest:
## Why

Clients that observe `SessionConfigured` need the same canonical permission view that app-server thread responses provide. Reporting the profile in protocol events lets clients keep their local state synchronized without reinterpreting legacy sandbox fields.

## What changed

This adds `permission_profile` to `SessionConfigured` and propagates it through core, exec JSON output, MCP server messages, and TUI history/widget handling.

## Verification

- `cargo test -p codex-tui permissions -- --nocapture`
- `cargo test -p codex-core --test all permissions_messages -- --nocapture`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/18282).
* #18288
* #18287
* #18286
* #18285
* #18284
* #18283
* __->__ #18282


codex-rs/core/src/session/session.rs
```diff
@@ -806,6 +806,7 @@ impl Session {
                 approval_policy: session_configuration.approval_policy.value(),
                 approvals_reviewer: session_configuration.approvals_reviewer,
                 sandbox_policy: session_configuration.sandbox_policy.get().clone(),
+                permission_profile: Some(session_configuration.permission_profile()),
```
viyatb-oai: [P2] Do not snapshot ExternalSandbox as a permission profile

This now always reports `SessionConfiguredEvent.permission_profile`, but `PermissionProfile::from_runtime_permissions()` cannot represent `FileSystemSandboxKind::ExternalSandbox`: it serializes that filesystem policy as root write, and converting the profile back yields a restricted root-write policy instead of external enforcement. The later stack head still has this exact assignment, while app-server response code explicitly omits profiles for ExternalSandbox, so this event becomes the inconsistent/lossy source. Please leave `permission_profile` as `None` when the active filesystem sandbox kind is ExternalSandbox, or extend the profile model before exposing it here.

viyatb-oai (approved):

viyatb-oai (approved):

PR #18284, opened 2026-04-17
tui: sync session permission profiles

bolinfest:
## Why

Once `SessionConfigured` carries the active `PermissionProfile`, the TUI must treat that as authoritative session state. Otherwise the widget can keep stale local permission details after a session is configured or resumed.

The TUI also keeps a local `Config` copy used for later operations, so session-sourced profiles and subsequent local sandbox changes need to keep the derived split runtime permissions in sync. Because this PR may land before the follow-up user-turn profile plumbing, embedded app-server turns also need a standalone path for carrying local runtime sandbox overrides.

## What changed

- Sync the chat widget runtime filesystem/network permissions from `SessionConfigured.permission_profile`, with the legacy `sandbox_policy` as the fallback.
- Recompute split runtime permissions whenever the TUI applies or carries forward a local sandbox-policy override.
- Mark feature-driven Auto-review sandbox changes as runtime sandbox overrides so the standalone embedded turn-start profile path is used even without the follow-up user-turn profile PR.
- Send a turn-start `permissionProfile` for embedded, non-ExternalSandbox turns when the TUI has a runtime sandbox override; remote and ExternalSandbox turns keep using the legacy sandbox field.
- Extend coverage for profile sync, local sandbox changes, ExternalSandbox fallback, feature-driven sandbox overrides, and turn-start permission override selection.

## Verification

- `cargo test -p codex-tui update_feature_flags_enabling_guardian_selects_auto_review`
- `cargo test -p codex-tui turn_start_permission_overrides_send_profiles_only_for_embedded_runtime_overrides`
- `cargo test -p codex-tui permission_settings_sync`
- `cargo test -p codex-tui session_configured_external_sandbox_keeps_external_runtime_policy`
- `cargo test -p codex-tui session_configured_syncs_widget_config_permissions_and_cwd`
- `just fix -p codex-tui`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/18284).
* #18288
* #18287
* #18286
* #18285
* __->__ #18284

### DIFF
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -2119,6 +2119,20 @@ impl ChatWidget {
         display: SessionConfiguredDisplay,
         fork_parent_title: Option<String>,
     ) {
+        let (file_system_sandbox_policy, network_sandbox_policy) = match event
+            .permission_profile
+            .as_ref()
+        {
+            Some(permission_profile) => permission_profile.to_runtime_permissions(),
+            None => (
+                codex_protocol::permissions::FileSystemSandboxPolicy::from_legacy_sandbox_policy(
+                    &event.sandbox_policy,
+                    &event.cwd,
+                ),
+                codex_protocol::permissions::NetworkSandboxPolicy::from(&event.sandbox_policy),
+            ),
+        };
+
         self.last_agent_markdown = None;
         self.agent_turn_markdowns.clear();
         self.visible_user_turn_count = 0;
@@ -2156,6 +2170,8 @@ impl ChatWidget {
             self.config.permissions.sandbox_policy =
                 Constrained::allow_only(event.sandbox_policy.clone());
         }
+        self.config.permissions.file_system_sandbox_policy = file_system_sandbox_policy;
+        self.config.permissions.network_sandbox_policy = network_sandbox_policy;
         self.config.approvals_reviewer = event.approvals_reviewer;
         self.status_line_project_root_name_cache = None;
         let forked_from_id = event.forked_from_id;
@@ -9764,6 +9780,14 @@ impl ChatWidget {
     #[cfg_attr(not(target_os = "windows"), allow(dead_code))]
     pub(crate) fn set_sandbox_policy(&mut self, policy: SandboxPolicy) -> ConstraintResult<()> {
         self.config.permissions.sandbox_policy.set(policy)?;
+        let sandbox_policy = self.config.permissions.sandbox_policy.get();
+        self.config.permissions.file_system_sandbox_policy =
+            codex_protocol::permissions::FileSystemSandboxPolicy::from_legacy_sandbox_policy(
+                sandbox_policy,
+                &self.config.cwd,
+            );
+        self.config.permissions.network_sandbox_policy =
+            codex_protocol::permissions::NetworkSandboxPolicy::from(sandbox_policy);
         Ok(())
     }
 
--- a/codex-rs/tui/src/app/config_persistence.rs
+++ b/codex-rs/tui/src/app/config_persistence.rs
@@ -72,13 +72,15 @@ impl App {
                 "Failed to carry forward approval policy override: {err}"
             ));
         }
-        if let Some(policy) = self.runtime_sandbox_policy_override.as_ref()
-            && let Err(err) = config.permissions.sandbox_policy.set(policy.clone())
-        {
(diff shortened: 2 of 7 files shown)


codex-rs/tui/src/chatwidget.rs
```diff
@@ -2119,6 +2120,12 @@ impl ChatWidget {
         display: SessionConfiguredDisplay,
         fork_parent_title: Option<String>,
     ) {
+        let permission_profile = event.permission_profile.clone().unwrap_or_else(|| {
```
viyatb-oai: [P2] Preserve ExternalSandbox when profile is absent

When `SessionConfigured` intentionally omits `permission_profile` for an `ExternalSandbox` session, this fallback converts the legacy sandbox through `PermissionProfile`. That model cannot carry the ExternalSandbox kind: it serializes the filesystem side as root write, and `to_runtime_permissions()` rebuilds a restricted root-write filesystem policy. The TUI config is then left with lossy split permissions, so later non-external thread start/resume/fork params can advertise root write from `config.permissions.permission_profile()` instead of the actual active policy. For the `None` case, build `FileSystemSandboxPolicy::from_legacy_sandbox_policy(...)` and `NetworkSandboxPolicy::from(...)` directly; only call `to_runtime_permissions()` when the event actually contains a profile.

viyatb-oai:
[P1] Preserve runtime sandbox changes on embedded turns

This issue still appears on the rebased stack, but `codex-rs/tui/src/app_server_session.rs` is no longer in this PR diff, so I am leaving it as a PR-level note instead of an inline comment. Embedded `turn/start` still omits both `sandboxPolicy` and `permissionProfile` for every non-ExternalSandbox turn, and `OverrideTurnContext` is still swallowed on the app-server path. That means TUI runtime sandbox changes such as `/sandbox` or approval-preset updates only change local/chat config; the next embedded turn does not carry the updated policy to app-server core, so a thread can keep running under the previous sandbox after switching read-only/workspace-write. Please distinguish “no permission change” from “new runtime sandbox override”, or send the corresponding permission profile when the sandbox policy has changed.

Suggested small fix: when `/sandbox` or an approval preset updates the legacy sandbox, also rebuild the split permissions in the affected config copies, then send a profile only while the TUI has a runtime sandbox override.

```rust
// config/chat sandbox setters, after the legacy sandbox_policy.set(...)
config.permissions.file_system_sandbox_policy =
    FileSystemSandboxPolicy::from_legacy_sandbox_policy(
        config.permissions.sandbox_policy.get(),
        &config.cwd,
    );
config.permissions.network_sandbox_policy =
    NetworkSandboxPolicy::from(config.permissions.sandbox_policy.get());
```

Then thread `permission_profile: Option<PermissionProfile>` through `AppServerSession::turn_start` and pass it from `try_submit_active_thread_op_via_app_server` only for embedded, non-ExternalSandbox runtime overrides:

```rust
let permission_profile = if !app_server.is_remote()
    && self.runtime_sandbox_policy_override.is_some()
    && !matches!(sandbox_policy, SandboxPolicy::ExternalSandbox { .. })
{
    Some(self.config.permissions.permission_profile())
} else {
    None
};
```

`turn_start` can keep sending the legacy `sandboxPolicy` for remote/ExternalSandbox, but for this embedded runtime-change case it should set `permissionProfile: permission_profile.map(Into::into)` instead of `None`. That keeps ordinary turns from re-projecting permissions, while still applying real runtime sandbox changes to the core session.

codex-rs/tui/src/chatwidget.rs
bolinfest: [codex] Addressed on the current head (`9afc4e9eec80`): the `None` fallback now builds split runtime permissions directly with `FileSystemSandboxPolicy::from_legacy_sandbox_policy(...)` and `NetworkSandboxPolicy::from(...)`; it only calls `to_runtime_permissions()` when `SessionConfigured` actually includes a `permission_profile`. I also added `session_configured_external_sandbox_keeps_external_runtime_policy` coverage.

bolinfest:
[codex] Addressed on the current head (`9afc4e9eec80`): `/sandbox` and Auto-review sandbox changes now rebuild split runtime permissions, mark `runtime_sandbox_policy_override`, and embedded non-ExternalSandbox `turn/start` sends a `permissionProfile` only for that runtime-override case. Remote and ExternalSandbox turns still use the legacy `sandboxPolicy` fallback.

viyatb-oai (approved):

PR #18289, opened 2026-04-17
Wire the PatchUpdated events through app_server

akshaynathan:
Wires patch_updated events through app_server. These events are parsed and streamed while apply_patch is being written by the model. The eventual goal is to use this to display better progress indicators in the codex app.


codex-rs/exec/src/exec_events.rs
```diff
@@ -28,6 +28,12 @@ pub enum ThreadEvent {
     /// Emitted when an item is updated.
     #[serde(rename = "item.updated")]
     ItemUpdated(ItemUpdatedEvent),
+    /// Emitted when the model starts streaming input for an item.
```
pakrym-oai: I would revert all exec changes. We don't actively add APIs to it.

pakrym-oai (approved):
Let's not touch exec

codex-rs/exec/src/exec_events.rs
akshaynathan: done

akshaynathan:
Removed exec changes and added 500ms buffering of events.

PR #18292, opened 2026-04-17
Make MCP resource read threadless

mzeng-openai:
## Summary

Making thread id optional so that we can better cache resources for MCPs for connectors since their resource templates is universal and not particular to projects.

- Make `mcpServer/resource/read` accept an optional `threadId`
- Read resources from the current MCP config when no thread is supplied
- Keep the existing thread-scoped path when `threadId` is present
- Update the generated schemas, README, and integration coverage

## Testing
- `just fmt`
- `cargo test -p codex-mcp`
- `cargo test -p codex-app-server --test all mcp_resource`
- `just fix -p codex-mcp`


shijie-oai (commented):
What was resource/read used for? Just want to make sure I understand.

mzeng-openai:
> What was resource/read used for? Just want to make sure I understand.

The app-server clients (Codex App) need to read UI widgets from MCP servers.

shijie-oai (approved):

PR #18304, opened 2026-04-17
Fix Windows exec policy test flake

etraut-openai:
## Summary

This fixes a Windows-only failure in the exec policy multi-segment shell test. The test was meant to verify that a compound shell command only bypasses sandboxing when every parsed segment has an explicit exec policy allow rule.

On Windows, the read-only sandbox setup is intentionally treated as lacking sandbox protection, so the old fixture could take the approval path before reaching the intended bypass assertion. The test now uses the workspace-write sandbox policy, keeping the focus on the per-segment bypass rule while preserving the expected bypass_sandbox false result when only cat is explicitly allowed.


dylan-hurd-oai:
LGTM

PR #18347, opened 2026-04-17
[codex-login] Clear refresh token after terminal refresh failures

cooper-oai:
## Summary
- Treat non-429 4xx ChatGPT token refresh responses as terminal failures: clear the persisted managed ChatGPT refresh token and reload auth state.
- Treat managed ChatGPT auth with a missing or cleared refresh token as unauthenticated on load, while keeping externally managed ChatGPT auth tokens valid without a refresh token.
- Preserve 429/5xx/network refresh failures as transient.

## CLI UX
- The first terminal refresh failure keeps the existing CLI-style copy: "Your access token could not be refreshed ... Please sign in again."
- After the refresh token is cleared, Codex does not surface a follow-up "refresh token missing" state. `codex login status` reports `Not logged in`, `account/read` and `getAuthStatus` report no auth, and TUI startup shows the normal login flow.

No Statsig gate; this applies the ChatGPT web clearing pattern directly for Codex.

## Testing
- `just fmt`
- `cargo test -p codex-login`
- `cargo test -p codex-app-server permanent_refresh_failure`
- `cargo test -p codex-app-server get_auth_status_omits_token_after_proactive_refresh_failure`
- `cargo test -p codex-app-server surfaces_cloud_requirements_load_errors`
- `just fix -p codex-login -p codex-app-server`

### DIFF
--- a/codex-rs/app-server/tests/suite/auth.rs
+++ b/codex-rs/app-server/tests/suite/auth.rs
@@ -310,7 +310,7 @@ async fn get_auth_status_omits_token_after_permanent_refresh_failure() -> Result
     assert_eq!(
         status,
         GetAuthStatusResponse {
-            auth_method: Some(AuthMode::Chatgpt),
+            auth_method: None,
             auth_token: None,
             requires_openai_auth: Some(true),
         }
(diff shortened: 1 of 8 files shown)


cooper-oai:
closing for now until we align on UX

PR #18232, opened 2026-04-16
codex: route thread/read persistence through thread store

wiltzius-openai:
Summary
- replace the thread/read persisted-load helper with ThreadStore::read_thread
- move SQLite/rollout summary, name, fork metadata, and history loading for persisted reads into LocalThreadStore
- leave getConversationSummary unchanged for a later PR

Stack
- Base: #18231

PR #18352, opened 2026-04-17
codex: route thread/read persistence through thread store

wiltzius-openai:
Summary
- replace the thread/read persisted-load helper with ThreadStore::read_thread
- move SQLite/rollout summary, name, fork metadata, and history loading for persisted reads into LocalThreadStore
- leave getConversationSummary unchanged for a later PR

Context
- Replaces closed stacked PR #18232 after PR #18231 merged and its base branch was deleted.

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -3817,7 +3817,7 @@ impl CodexMessageProcessor {
     ) -> Result<Thread, ThreadReadViewError> {
         let loaded_thread = self.load_live_thread_for_read(thread_id).await;
         let mut thread = if let Some(thread) = self
-            .load_persisted_thread_for_read(thread_id, include_turns, loaded_thread.as_ref())
+            .load_persisted_thread_for_read(thread_id, include_turns)
             .await?
         {
             thread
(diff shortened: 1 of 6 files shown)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -9291,6 +9255,56 @@ fn thread_store_list_error(err: ThreadStoreError) -> JSONRPCErrorError {
     }
 }
 
+fn thread_from_stored_thread(
```
aibrahim-oai, line 9258: if we decided to go with 2 types, let's do `From<StoredThread> for Thread`

wiltzius-openai: I agree with the direction, but a literal `impl From<StoredThread> for Thread` cannot live in `codex-app-server`: `From` is a std trait and both `Thread` (`codex-app-server-protocol`) and `StoredThread` (`codex-thread-store`) are foreign types to this crate, so the orphan rules reject it. Putting the impl in either defining crate would force a dependency between the protocol and persistence crates that we have been trying to avoid. This helper is intentionally the app-server-local adapter boundary; once it has a second caller, I think the better cleanup is to move it into a small private thread-view module rather than making the protocol or store crate depend on the other.

aibrahim-oai (approved):

codex-rs/app-server/src/codex_message_processor.rs
wiltzius-openai: lol sorry @aibrahim-oai I had the pr babysitter skill running and codex wrote this itself to respond to comments (I just wanted it to retry the failing CI build)

...but fwiw I think its true

PR #18359, opened 2026-04-17
Show default reasoning in /status

aibrahim-oai:
- Shows the model catalog default reasoning effort when no reasoning override is configured.\n- Adds /status coverage for the empty-config fallback.


aibrahim-oai:
Recreated from an openai/codex branch so required CI can run with normal repository credentials: #18373.

PR #18382, opened 2026-04-17
Add max context window model metadata

aibrahim-oai:
Adds max_context_window to model metadata and routes core context-window reads through resolved model info. Config model_context_window overrides are clamped to max_context_window when present; without an override, the model context_window is used.


aibrahim-oai:
Addressed in e9a7bd9f3b: model_context_window overrides are now clamped to max_context_window when model metadata provides a max, and the integration tests assert the oversized override path uses the max.

codex-rs/protocol/src/openai_models.rs
```diff
@@ -300,9 +303,13 @@ pub struct ModelInfo {
 }
 
 impl ModelInfo {
+    pub fn resolved_context_window(&self) -> Option<i64> {
```
pakrym-oai, line 306: Feels like we shouldn't need this.

pakrym-oai: But it's fine.

pakrym-oai (approved):
Can we add an integration test that sets very large context size? Or is the side effect to hard to observe?

aibrahim-oai:
Yep, added the very-large-context integration coverage in `remote_models_config_context_window_override_clamps_to_max_context_window`: it sets `model_context_window = 1_000_000` with remote model metadata `context_window = 273_000` and `max_context_window = 400_000`, then asserts the emitted runtime context window is clamped to `400_000`.

PR #18391, opened 2026-04-17
fix(core): emit hooks for apply_patch edits

### LINKED ISSUE #16732: ApplyPatchHandler doesn't emit PreToolUse/PostToolUse hook event. Hooks only fire for Bash tool.

0.118.0

ChatGPT Plus

_No response_

Darwin 24.6.0 arm64 arm

_No response_

PreToolUse and PostToolUse hooks registered via hooks.json never fire for file writes made through apply_patch. They only fire for Bash/shell tool calls.

Two root causes in the CLI source:

ApplyPatchHandler in registry.rs uses the default None implementation for pre_tool_use_payload() / post_tool_use_payload() (lines 63-65 of the trait), so the hook runtime skips it entirely.

hook_runtime.rs hardcodes tool_name: "Bash" (line 131) in both run_pre_tool_use_hooks and run_post_tool_use_hooks, so even if payloads were emitted, they'd be misidentified as Bash calls.

This means hook-based file access control, audit logging, or write coordination cannot cover the primary file write path.

1. Create a hooks.json with a PreToolUse hook command that logs all invocations to a file
2. Ask Codex to edit a source file (triggers apply_patch)
3. Ask Codex to run a shell command (triggers Bash)
4. Check the log, observe only the Bash call appears and the file edit is missing

Tested with Codex CLI v0.118.0. Hook log captured 4 entries (2 PreToolUse + 2 PostToolUse), all with tool_name: "Bash" for sed and rg calls. The apply_patch call that actually modified the file produced zero hook events.

PreToolUse and PostToolUse hooks should fire for apply_patch with tool_name: "ApplyPatch" (or similar) and a payload containing the file paths being modified, consistent with how Bash tool calls emit hook events.

### Additional information

We're building multi-agent orchestration that coordinates file writes across Codex, Claude Code, and OpenCode. Claude Code's SDK fires PreToolUse hooks on Write/Edit/MultiEdit tools, which lets us implement file-level locking between agents. This is foundational for implementing checkpoints in an environment where multiple agents edit the same worktree concurrently. Agents acquire a lock on a file before editing, providing a clear record of what changed and how to undo it.

Codex's hook infrastructure supports the same pattern architecturally, but apply_patch not opting in blocks the integration. The fix is scoped, implement pre_tool_use_payload() on ApplyPatchHandler and parameterize tool_name in hook_runtime.rs. Happy to elaborate on the use case or provide more analysis.

fcoury-oai:
Fixes https://github.com/openai/codex/issues/16732.

## Why

`apply_patch` is Codex's primary file edit path, but it was not emitting `PreToolUse` or `PostToolUse` hook events. That meant hook-based policy, auditing, and write coordination could observe shell commands while missing the actual file mutation performed by `apply_patch`.

The issue also exposed that the hook runtime serialized command hook payloads with `tool_name: "Bash"` unconditionally. Even if `apply_patch` supplied hook payloads, hooks would either fail to match it directly or receive misleading stdin that identified the edit as a Bash tool call.

## What Changed

- Added `PreToolUse` and `PostToolUse` payload support to `ApplyPatchHandler`.
- Exposed the raw patch body as `tool_input.command` for both JSON/function and freeform `apply_patch` calls.
- Taught tool hook payloads to carry a handler-supplied hook-facing `tool_name`.
- Preserved existing shell compatibility by continuing to emit `Bash` for shell-like tools.
- Serialized the selected hook `tool_name` into hook stdin instead of hardcoding `Bash`.
- Relaxed the generated hook command input schema so `tool_name` can represent tools other than `Bash`.

## Verification

Added focused handler coverage for:

- JSON/function `apply_patch` calls producing a `PreToolUse` payload.
- Freeform `apply_patch` calls producing a `PreToolUse` payload.
- Successful `apply_patch` output producing a `PostToolUse` payload.
- Shell and `exec_command` handlers continuing to expose `Bash`.

Added end-to-end hook coverage for:

- A `PreToolUse` hook matching `^apply_patch$` blocking the patch before the target file is created.
- A `PostToolUse` hook matching `^apply_patch$` receiving the patch input and tool response, then adding context to the follow-up model request.
- Non-participating tools such as the plan tool continuing not to emit `PreToolUse`/`PostToolUse` hook events.

Also validated manually with a live `codex exec` smoke test using an isolated temp workspace and temp `CODEX_HOME`. The smoke test confirmed that a real `apply_patch` edit emits `PreToolUse`/`PostToolUse` with `tool_name: "apply_patch"`, a shell command still emits `tool_name: "Bash"`, and a denying `PreToolUse` hook prevents the blocked patch file from being created.


fcoury-oai:
@codex review

canvrno-oai (approved):

e-fu:
This PR would unblock an important real workflow for us.

We’ve been porting an Elixir/Phoenix hook setup from another agent into Codex.

What we confirmed locally:
- `PreToolUse` for `Bash` is working in our environment.
- We can visibly block `mix test` and `mix dialyzer` and force `mix test.json` / `mix dialyzer.json` instead.
- Plugin skills are loading correctly.
- The missing piece is edit-time validation after Codex file edits.

Our workflow depends on post-edit checks like:
- `mix format` on the edited file
- `mix compile --warnings-as-errors`
- targeted `mix test.json` for the corresponding test file
- framework-specific checks such as `mix ash.codegen --check`

Right now that workflow cannot be implemented cleanly because the file edit path does not emit hook events, so we had to fall back to a weaker `Stop`-hook design. That loses precision, runs later than we want, and is much less ergonomic for iterative coding.

So from a user/workflow perspective: this is not just a nice-to-have hook expansion. It’s the difference between:
- immediate post-edit validation that keeps the loop tight
- versus delayed whole-turn validation that is noisier and less targeted

If this lands and `apply_patch` emits `PreToolUse` / `PostToolUse` with a non-Bash `tool_name`, that would materially improve Codex for edit-centric code-quality workflows.

codex-rs/core/tests/suite/hooks.rs
```diff
@@ -2267,7 +2345,92 @@ async fn post_tool_use_exit_two_replaces_one_shot_exec_command_output_with_feedb
...
+    let tool_response = hook_inputs[0]["tool_response"]
+        .as_str()
+        .context("apply_patch tool_response should be a string")?;
+    assert_eq!(
+        serde_json::from_str::<Value>(tool_response)?,
+        serde_json::json!({
+            "output": "Success. Updated the following files:\nA post_tool_use_apply_patch.txt\n",
+            "metadata": {
+                "exit_code": 0,
+                "duration_seconds": 0.0,
```
abhinav-oai: nit: codex is saying this test could flake on a slow CI worker if the tool takes >50ms

looks like other tests like in `shell_serialization.rs` will drop the `duration_seconds` before asserting equality if we wanna follow that pattern

abhinav-oai (changes requested):
can we also wire up the `PermissionRequest` hook to be able to match the `apply_patch` tool?

I think we just need an impl of `permission_request_payload()` and minor schema change so the logic in `tools/orchestrator.rs` kicks in

fcoury-oai:
I implemented the compatibility pieces for apply_patch hooks.

The behavior now is:
- hook stdin stays canonical: tool_name remains apply_patch
- matcher selection recognizes apply_patch, Write, and Edit for apply_patch calls
- a handler only runs once if its regex matches multiple aliases, like apply_patch|Write|Edit
- PermissionRequest supports apply_patch, with the patch body in tool_input.command

I also exercised this locally and the hook config matched PreToolUse and PermissionRequest via `^Write$`, and PostToolUse via `^Edit$`, while all hook payloads still logged tool_name: apply_patch.

abhinav-oai (approved):
🚢

etraut-openai (approved):

PR #18393, opened 2026-04-17
feat(auto-review) Handle request_permissions calls

dylan-hurd-oai:
## Summary
When auto-review is enabled, it should handle request_permissions tool. We'll need to clean up the UX but I'm planning to do that in a separate pass

## Testing
- [x] Ran locally
Screenshot 2026-04-17 at 1 16 13 PM

### DIFF
--- a/codex-rs/core/src/session/mod.rs
+++ b/codex-rs/core/src/session/mod.rs
@@ -812,28 +812,6 @@ async fn thread_title_from_state_db(
         .flatten()
 }
 
-struct CancelTokenOnDrop {
-    token: Option<CancellationToken>,
-}
-
-impl CancelTokenOnDrop {
-    fn new(token: CancellationToken) -> Self {
-        Self { token: Some(token) }
-    }
-
-    fn disarm(&mut self) {
-        self.token = None;
-    }
-}
-
-impl Drop for CancelTokenOnDrop {
-    fn drop(&mut self) {
-        if let Some(token) = self.token.take() {
-            token.cancel();
-        }
-    }
-}
-
 impl Session {
     pub(crate) async fn app_server_client_metadata(&self) -> AppServerClientMetadata {
         let state = self.state.lock().await;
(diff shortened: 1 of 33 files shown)


won-openai (approved):

codex-rs/core/src/session/mod.rs
```diff
@@ -1913,6 +1994,72 @@ impl Session {
             | AskForApproval::Granular(_) => {}
         }
 
+        if crate::guardian::routes_approval_to_guardian(turn_context.as_ref()) {
```
jif-oai, line 1924: This still keys off `routes_approval_to_guardian()`, which excludes `AskForApproval::Granular. So `request_permissions` keeps surfacing the manual prompt in granular+guardian configs instead of being auto-reviewed.

codex-rs/core/src/session/mod.rs
```diff
@@ -1913,6 +1994,72 @@ impl Session {
...
+                request,
+                /*retry_reason*/ None,
+                cancellation_token.clone(),
+            );
+            let decision = tokio::select! {
+                biased;
+                _ = cancellation_token.cancelled() => return None,
+                decision = review_rx => decision.unwrap_or(ReviewDecision::Denied),
+            };
+            let response = match decision {
```
jif-oai, line 1952: If I read this correctly, approved for session won't work anymore right? Is this expected?

jif-oai (approved):
After those 2 comments, lgtm

codex-rs/core/src/session/mod.rs
dylan-hurd-oai: I think it's just that Guardian doesn't approve for session, which is fine?

codex-rs/core/src/session/mod.rs
dylan-hurd-oai: good catch, addressed

PR #18395, opened 2026-04-17
/plugins: Add inline enablement toggles

canvrno-oai:
This PR adds inline enable/disable controls to the new /plugins browse menu. Installed plugins can now be toggled directly from the list with keyboard interaction, and the associated config-write plumbing is included so the UI and persisted plugin state stay in sync. This also includes the queued-write handling needed to avoid stale toggle completions overwriting newer intent.

- Add toggleable plugin rows for installed plugins in /plugins
- Support Space to enable or disable without leaving the list
- Persist plugin enablement through the existing app/config write path
- Preserve the current selection while the list refreshes after a toggle
- Add tests and snapshot updates for toggling behavior


canvrno-oai:
@codex review

fcoury-oai (approved):
Tested and it looks good. One minor detail is pressing Esc while on a plugin detail page exits to the composer instead of falling back to the previous menu, but this can be adjusted on a follow up PR.

PR #18397, opened 2026-04-17
[codex] Upgrade rules_rs and llvm to latest BCR versions

zbarsky-openai:
## Why
This branch brings the Bazel module pins for `rules_rs` and `llvm` up to the latest BCR releases and aligns the root direct dependencies with the versions the module graph already resolves to.

That gives us a few concrete wins:
- picks up newer upstream fixes in the `rules_rs` / `rules_rust` stack, including work around repo-rule nondeterminism and default Cargo binary target generation
- picks up test sharding support from the newer `rules_rust` stack ([hermeticbuild/rules_rust#13](https://github.com/hermeticbuild/rules_rust/pull/13))
- picks up newer built-in knowledge for common system crates like `gio-sys`, `glib-sys`, `gobject-sys`, `libgit2-sys`, and `libssh2-sys`, which gives us a future path to reduce custom build-script handling
- reduces local patch maintenance by dropping fixes that are now upstream and rebasing the remaining Windows patch stack onto a newer upstream base
- removes the direct-dependency warnings from `bazel-lock-check` by making the root pins match the resolved graph

## What Changed
- bump `rules_rs` from `0.0.43` to `0.0.58`
- bump `llvm` from `0.6.8` to `0.7.1`
- bump `bazel_skylib` from `1.8.2` to `1.9.0` so the root direct dep matches the resolved graph
- regenerate `MODULE.bazel.lock` for the updated module graph
- refresh the remaining Windows-specific patch stack against the newer upstream sources:
  - `patches/rules_rs_windows_gnullvm_exec.patch`
  - `patches/rules_rs_windows_exec_linker.patch`
  - `patches/rules_rust_windows_exec_std.patch`
  - `patches/rules_rust_windows_msvc_direct_link_args.patch`
- remove patches that are no longer needed because the underlying fixes are upstream now:
  - `patches/rules_rs_delete_git_worktree_pointer.patch`
  - `patches/rules_rust_repository_set_exec_constraints.patch`

## Validation
- `just bazel-lock-update`
- `just bazel-lock-check`


bolinfest (approved):
If CI is green, then this is good to go!

PR #18401, opened 2026-04-17
Support multiple managed environments

starr-openai:
## Summary
- refactor EnvironmentManager to own keyed environments with default/local lookup helpers
- keep remote exec-server client creation lazy until exec/fs use
- preserve disabled agent environment access separately from internal local environment access

## Validation
- not run (per Codex worktree instruction to avoid tests/builds unless requested)


codex-rs/exec-server/src/environment.rs
```diff
@@ -15,109 +13,108 @@ use crate::remote_process::RemoteProcess;
...
-    ) -> Self {
-        Self::new_with_runtime_paths(
-            std::env::var(CODEX_EXEC_SERVER_URL_ENV_VAR).ok(),
-            local_runtime_paths,
-        )
-    }
+        } = args;
+        let (exec_server_url, environment_disabled) = normalize_exec_server_url(exec_server_url);
+        let mut environments = HashMap::new();
+        let default_environment = if environment_disabled {
```
pakrym-oai, line 88: nit: I think we should move disabling logic from here and to the point thread start.

EnvironmentManager always has either local or local+remote
If thread/start has environment set to:
- [] - use none
- ["local"] or ["remote"] - noneselected environment
- None and CODEX_EXEC_SERVER_URL_ENV_VAR is set - interpret CODEX_EXEC_SERVER_URL_ENV_VAR into either `remote` or []
- None and no CODEX_EXEC_SERVER_URL_ENV_VAR - `local`

pakrym-oai: Confusingly CODEX_EXEC_SERVER_URL_ENV_VAR is both a signal for EnvManager as to what remote is and a singnal for default thread init behavior but that's temporary

codex-rs/exec-server/src/remote_file_system.rs
```diff
@@ -46,8 +50,8 @@ impl ExecutorFileSystem for RemoteFileSystem {
         sandbox: Option<&FileSystemSandboxContext>,
     ) -> FileSystemResult<Vec<u8>> {
         trace!("remote fs read_file");
-        let response = self
-            .client
+        let client = self.client().await?;
```
pakrym-oai: might as well push initi into individual operations like fs_read_file

codex-rs/exec-server/src/environment.rs
starr-openai: +1 - in an earlier version of this PR I had an explicit signal on EM to indicate whether tools were disabled for the agent itself, distinct from the environments available to codex.

Should this be a property of the thread / turn itself then?

codex-rs/exec/src/lib.rs
```diff
@@ -492,9 +494,10 @@ pub async fn run_main(cli: Cli, arg0_paths: Arg0DispatchPaths) -> anyhow::Result
         cloud_requirements: run_cloud_requirements,
         feedback: CodexFeedback::new(),
         log_db: None,
-        environment_manager: std::sync::Arc::new(EnvironmentManager::from_env_with_runtime_paths(
-            Some(local_runtime_paths),
-        )),
+        environment_manager: std::sync::Arc::new(EnvironmentManager::new(EnvironmentManagerArgs {
+            exec_server_url: std::env::var(CODEX_EXEC_SERVER_URL_ENV_VAR).ok(),
```
pakrym-oai: nit: Should we hide the env-var reading?

codex-rs/app-server/src/message_processor/tracing_tests.rs
```diff
@@ -241,7 +241,9 @@ fn build_test_processor(
         outgoing,
         arg0_paths: Arg0DispatchPaths::default(),
         config,
-        environment_manager: Arc::new(EnvironmentManager::new(/*exec_server_url*/ None)),
```
pakrym-oai, line 282: nit: EnvironmentManager::default_for_tests?

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -5673,27 +5673,20 @@ impl CodexMessageProcessor {
             .await;
         let auth_manager = Arc::clone(&self.auth_manager);
         let auth = auth_manager.auth().await;
-        let runtime_environment = match self.thread_manager.environment_manager().current().await {
-            Ok(Some(environment)) => {
+        let runtime_environment = match self
+            .thread_manager
+            .environment_manager()
+            .default_environment()
```
pakrym-oai: Ideally this uses the selected environmet. Non blocking.

codex-rs/core/src/session/session.rs
```diff
@@ -786,8 +786,9 @@ impl Session {
             tx_event.clone(),
             session_configuration.sandbox_policy.get().clone(),
             McpRuntimeEnvironment::new(
-                environment
-                    .clone()
+                sess.services
```
pakrym-oai, line 773: yeah, in the future we should pass the list of active environments to this.

codex-rs/core/tests/suite/tools.rs
```diff
@@ -75,6 +75,66 @@ fn ev_namespaced_function_call(
     })
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn exec_server_none_omits_environment_backed_tools() -> Result<()> {
```
pakrym-oai: though we had this one

codex-rs/exec-server/src/client.rs
```diff
@@ -155,6 +156,37 @@ pub struct ExecServerClient {
     inner: Arc<Inner>,
 }
 
+#[derive(Clone)]
+pub(crate) struct LazyRemoteExecServerClient {
+    websocket_url: String,
+    client: Arc<OnceCell<ExecServerClient>>,
+}
+
+impl LazyRemoteExecServerClient {
```
pakrym-oai, line 184: is there a reason not to inline into ExecServerClient?

codex-rs/exec-server/src/environment.rs
```diff
@@ -242,57 +244,102 @@ mod tests {
...
+                .get_environment(LOCAL_ENVIRONMENT_ID)
+                .expect("local environment")
+                .is_remote()
+        );
     }
 
     #[test]
-    fn environment_manager_reports_remote_url() {
-        let manager = EnvironmentManager::new(Some("ws://127.0.0.1:8765".to_string()));
+    fn create_remote_environment_does_not_connect() {
```
pakrym-oai: rem, not very useful

codex-rs/exec-server/src/client.rs
starr-openai: mmm, i'd still lean towards keeping them separate. this makes it clear that we're holding a ref to a client that is connect-on-first-use, vs ExecServerClient which is definitively in a connected state.

codex-rs/exec-server/src/environment.rs
```diff
@@ -171,34 +185,42 @@ impl Environment {
...
+                Self::remote_with_runtime_paths(exec_server_url, local_runtime_paths)
+            }
+            None => Self::local_with_runtime_paths(local_runtime_paths),
+        })
+    }
+
+    fn local_with_runtime_paths(local_runtime_paths: Option<ExecServerRuntimePaths>) -> Self {
+        let filesystem: Arc<dyn ExecutorFileSystem> = match local_runtime_paths.clone() {
+            Some(runtime_paths) => Arc::new(LocalFileSystem::with_runtime_paths(runtime_paths)),
+            None => Arc::new(LocalFileSystem::unsandboxed()),
```
pakrym-oai: Is this a security incident waiting to happen?

codex-rs/exec-server/src/environment.rs
```diff
@@ -15,130 +13,146 @@ use crate::remote_process::RemoteProcess;
...
     local_runtime_paths: Option<ExecServerRuntimePaths>,
 }
 
 impl Default for Environment {
     fn default() -> Self {
         Self {
             exec_server_url: None,
-            remote_exec_server_client: None,
             exec_backend: Arc::new(LocalProcess::default()),
+            filesystem: Arc::new(LocalFileSystem::unsandboxed()),
```
pakrym-oai, line 148: I'm a bit worried how freely we are creationg unsandboxed fs impls.
Not blocking. Longer term we need to reduce how many ways to create environment there are.

pakrym-oai (approved):

codex-rs/exec-server/src/environment.rs
starr-openai: good point - let me try to make ExecServerRuntimePaths required so we don't silently drop

starr-openai: that said in `codex-rs/exec-server/src/local_file_system.rs` we do throw if we try to make a sandboxed fs operation but have no configured runtime operations

codex-rs/exec-server/src/environment.rs
starr-openai: made this `default_for_tests` only, remved Default constructor to avoid change it's created accidentally

PR #18414, opened 2026-04-17
Skip .tsh in Windows sandbox profile read roots

efrazer-oai:
## Summary
- Add .tsh to the Windows sandbox user-profile read-root exclusions.
- Extend the existing profile-read-root test so Teleport state is covered beside .ssh.

## Why
Teleport stores local client state under the human user's .tsh directory. The Windows sandbox setup should not grant CodexSandboxUsers inherited read ACLs on that directory while scanning top-level profile entries.

## Testing
- just fmt
- cargo test -p codex-windows-sandbox --lib

## Note
- just bazel-lock-update and just bazel-lock-check currently fail locally before evaluating this change because Bazel cannot load @@rules_rs++toolchains+cargo_macos_aarch64_1_93_0//:bin/cargo.


efrazer-oai:
Closing in favor of a smaller Windows sandbox fix that prevents USERPROFILE itself from becoming an ACL root.

PR #18415, opened 2026-04-17
Skip SSH config dependency roots in Windows sandbox

efrazer-oai:
## Summary
- Add a small OpenSSH-config dependency scanner for the Windows sandbox profile scan.
- Follow Include directives recursively, including globbed includes, and record top-level profile folders named by file-bearing SSH directives.
- Exclude those folders from profile read roots before ACLs are granted to CodexSandboxUsers.

## Why
Adding .tsh covers the known Teleport failure. This follow-up handles the same class of bug when .ssh/config points at other profile folders through IdentityFile, CertificateFile, UserKnownHostsFile, ControlPath, IdentityAgent, PKCS11Provider, SecurityKeyProvider, XAuthLocation, or included config files.

The scanner is intentionally narrower than a general SSH client config parser. It answers one question: which top-level folders under the human user profile are named by the .ssh/config dependency chain?

## Testing
- just fmt
- cargo test -p codex-windows-sandbox --lib

### DIFF
--- a/codex-rs/Cargo.lock
+++ b/codex-rs/Cargo.lock
@@ -3275,6 +3275,7 @@ dependencies = [
  "codex-utils-string",
  "dirs-next",
  "dunce",
+ "glob",
  "pretty_assertions",
  "rand 0.8.5",
  "serde",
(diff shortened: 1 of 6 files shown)


efrazer-oai:
Closing in favor of a smaller Windows sandbox fix that prevents USERPROFILE itself from becoming an ACL root.

PR #18416, opened 2026-04-17
Add turn-scoped environment selections

starr-openai:
## Summary
- add experimental turn/start.environments params for per-turn environment id + cwd selections
- pass selections through core protocol ops and resolve them with EnvironmentManager before TurnContext creation
- treat omitted selections as default behavior, empty selections as no environment, and non-empty selections as first environment/cwd as the turn primary

## Testing
- ran `just fmt`
- ran `just write-app-server-schema`
- not run: unit tests for this stacked PR


codex-rs/codex-experimental-api-macros/src/lib.rs
```diff
@@ -261,11 +261,8 @@ fn presence_expr_for_access(
     access: proc_macro2::TokenStream,
     ty: &Type,
 ) -> proc_macro2::TokenStream {
-    if let Some(inner) = option_inner(ty) {
```
pakrym-oai, line 264: ooc why did this have to change?

codex-rs/core/src/session/tests.rs
```diff
@@ -3759,6 +3766,132 @@ async fn user_turn_updates_approvals_reviewer() {
     );
 }
 
+#[tokio::test]
+async fn turn_environment_selection_sets_primary_environment() {
```
pakrym-oai, line 3763: up to you but I don't find these unit tests that assert internal implementation details particularly useful long-term.

codex-rs/core/src/session/tests.rs
```diff
@@ -3759,6 +3766,132 @@ async fn user_turn_updates_approvals_reviewer() {
...
+    let (session, _turn_context, _rx) = make_session_and_context_with_rx().await;
+    let selected_cwd =
```
pakrym-oai, line 3765: Are we turning AbsolutePath into pathbuf just to parse it back?

codex-rs/core/src/session/turn_context.rs
```diff
@@ -346,6 +357,8 @@ impl Session {
         models_manager: &ModelsManager,
         network: Option<NetworkProxy>,
         environment: Option<Arc<Environment>>,
+        environments: Option<Vec<TurnEnvironment>>,
```
pakrym-oai, line 363: should this be optional?

pakrym-oai: Feels like it should be materialized either into an empty list or non-empty list but never None.

codex-rs/core/src/session/turn_context.rs
```diff
@@ -511,17 +538,50 @@ impl Session {
...
+            return Ok(None);
+        };
+
+        let mut turn_environments = Vec::with_capacity(environment_selections.len());
+        for environment_selection in environment_selections {
+            let environment = self
+                .services
+                .environment_manager
+                .get_environment(&environment_selection.environment_id)
+                .ok_or_else(|| codex_config::ConstraintError::InvalidValue {
```
pakrym-oai: is this the correct error type?

pakrym-oai: I think ConstraintError::InvalidValue is for requirements.toml

codex-rs/core/src/session/turn_context.rs
```diff
@@ -619,6 +695,7 @@ impl Session {
             sub_id,
             session_configuration,
             /*final_output_json_schema*/ None,
+            /*turn_environments*/ None,
```
pakrym-oai, line 700: Is this correct? WebSocket prewarm uses this. 
We establish a websocket connection and send an initial empty responses api request to warm up inference. That request includes tools. If we pass None environemnt here the first request will be sent without tools even for local environment which will break prewarming.

We might want to add environments onto thread/start to help the prewarm send the right request.

codex-rs/core/src/session/turn_context.rs
```diff
@@ -537,14 +597,28 @@ impl Session {
                 &per_turn_config.to_models_manager_config(),
             )
             .await;
+        let environment = match turn_environments.as_ref() {
```
pakrym-oai: nit: maybe grab `turn_environments.first()` first and then set both `environment` and `cwd`

codex-rs/core/src/session/turn_context.rs
```diff
@@ -537,14 +597,28 @@ impl Session {
...
+        let cwd = turn_environments
+            .as_ref()
+            .and_then(|turn_environments| {
+                turn_environments
+                    .first()
+                    .map(|turn_environment| turn_environment.cwd.clone())
+            })
+            .unwrap_or_else(|| session_configuration.cwd.clone());
+        per_turn_config.cwd = cwd.clone();
         let plugin_outcome = self
```
pakrym-oai: nit: can we pass cwd into build_per_turn_config and have it build the correct value from the getgo?

pakrym-oai (approved):
Should we make environments non-optional?

codex-rs/codex-experimental-api-macros/src/lib.rs
starr-openai: weird - maybe a rebase conflict or other change i had.- let me remove

starr-openai: ah this was for the experimental config stuff. removing that, I don't think we really need to gate it

codex-rs/core/src/session/turn_context.rs
starr-openai: we're still using the Option state to determine if env is enabled or not for tool access.  will make this explicit in the later pr once this is attached to persistent thread state

PR #18419, opened 2026-04-17
[Linux] Reserve missing top level .git at runtime

evawong-oai:
## Summary

A writable workspace with no `.git` at startup could run `git init` inside Linux bubblewrap and create git metadata outside the intended protected path set. This change closes that path without adding path specific mount cleanup.

Bugcrowd submission: linked from BUGB 15632

## Root Cause

The shared policy only protects `.git` after it exists. Linux bubblewrap needs to reserve the missing top level `.git` path for the active workspace root before the sandboxed command runs. Bubblewrap can materialize a host side mount target for that reservation, so the helper now removes only the synthetic empty target after bubblewrap exits.

## Scope

This PR touches four files.

1. `bwrap.rs` adds the missing top level `.git` reservation only for the active workspace root. Missing protected mount targets now share one read only tmpfs mask and cleanup helper, so `.git`, `.codex`, and future missing protected paths use the same handling.
2. `linux_run_main.rs` keeps the normal exec path when no cleanup is needed, runs bubblewrap through a child helper only when synthetic mount target cleanup is required, and treats a failed proc probe as a reason to retry without proc.
3. `linux_run_main_tests.rs` proves cleanup removes an empty synthetic target and preserves real git metadata.
4. `landlock.rs` adds a Linux regression test that proves `git init` is blocked while ordinary writable files still work.

## Validation

1. Local project format passed.
2. Local cargo format check passed.
3. Local Linux sandbox clippy passed.
4. Local Linux sandbox crate test passed on macOS, with Linux tests compiled out on this host.
5. Local whitespace diff check passed.
6. GitHub Actions is running for commit `88effe34556fb6fd065a158332bb3a29fd0a28dd`.

### DIFF
--- a/codex-rs/linux-sandbox/src/bwrap.rs
+++ b/codex-rs/linux-sandbox/src/bwrap.rs
@@ -24,6 +24,8 @@ use std::process::Command;
 
 use codex_protocol::error::CodexErr;
 use codex_protocol::error::Result;
+use codex_protocol::protocol::FileSystemAccessMode;
+use codex_protocol::protocol::FileSystemPath;
 use codex_protocol::protocol::FileSystemSandboxPolicy;
 use codex_protocol::protocol::WritableRoot;
 use codex_utils_absolute_path::AbsolutePathBuf;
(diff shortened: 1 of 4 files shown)


evawong-oai:
@codex review

evawong-oai:
@codex review

PR #18428, opened 2026-04-18
app-server: define device key v2 protocol

euroelessar:
## Why

Clients need a stable app-server protocol surface for enrolling a local device key, retrieving its public key, and producing a device-bound proof.

The protocol reports `protectionClass` explicitly so clients can distinguish hardware-backed keys from an explicitly allowed OS-protected fallback. Signing uses a tagged `DeviceKeySignPayload` enum rather than arbitrary bytes so each signed statement is auditable at the API boundary.

## What changed

- Added v2 JSON-RPC methods for `device/key/create`, `device/key/public`, and `device/key/sign`.
- Added request/response types for device-key metadata, SPKI public keys, protection classes, and ECDSA signatures.
- Added `DeviceKeyProtectionPolicy` with hardware-only default behavior and an explicit `allow_os_protected_nonextractable` option.
- Added the initial `remoteControlClientConnection` signing payload variant.
- Regenerated JSON Schema and TypeScript fixtures for app-server clients.

## Stack

This is PR 1 of 4 in the device-key app-server stack.

## Validation


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2306,6 +2306,126 @@ pub struct FeedbackUploadResponse {
...
+
+/// Structured payloads accepted by `device/key/sign`.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, JsonSchema, TS)]
+#[serde(tag = "type", rename_all = "camelCase")]
+#[ts(tag = "type", export_to = "v2/")]
+pub enum DeviceKeySignPayload {
+    /// Payload bound to one remote-control controller websocket `/client` connection challenge.
+    #[serde(rename_all = "camelCase")]
+    #[ts(rename_all = "camelCase")]
+    RemoteControlClientConnection {
```
viyatb-oai, line 2592: This signed payload is not explicitly bound to the target server/environment/session/role/audience. It currently signs nonce/account/client/tokenHash/token expiry/scopes, but the verifier needs replay protection across the target it is authorizing. Either add the target binding fields directly, or make tokenHash a hash of a canonical backend challenge envelope that contains those bindings and is recomputed by the backend.

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2306,6 +2306,126 @@ pub struct FeedbackUploadResponse {
...
+pub struct DeviceKeySignParams {
+    pub key_id: String,
+    pub payload: DeviceKeySignPayload,
+}
+
+/// ASN.1 DER signature returned by `device/key/sign`.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct DeviceKeySignResponse {
```
viyatb-oai, line 2646: The response returns only the DER signature, but backend verification needs a stable cross-service byte contract for what was signed. Returning only signatureDerBase64 means the verifier has to exactly reproduce Rust serde output. Please either return signedPayloadBase64 alongside the signature, or define a canonical encoding and add test vectors for backend verification.

viyatb-oai:
[P1] Target binding is still implicit. The signed payload only carries targetOrigin/targetPath + token hash, but no explicit server_id/environment_id/session_id. If the backend controller token isn’t already bound to the target server/session, PoP doesn’t prevent cross‑server lateral movement. Either enforce target binding via token claims on the backend, or add explicit target identifiers to the signed payload.

euroelessar:
[codex] Addressed in `a38e6c1`: the remote-control client connection signed payload now includes an explicit `sessionId` binding. That puts the backend-issued websocket session identifier inside the bytes signed by the device key, alongside nonce/audience/target/token binding.

viyatb-oai (approved):

etraut-openai (approved):

PR #18429, opened 2026-04-18
app-server: add codex-device-key crate

euroelessar:
## Why

Device-key storage and signing are local security-sensitive operations with platform-specific behavior. Keeping the core API in `codex-device-key` keeps app-server focused on routing and business logic instead of owning key-management details.

The crate keeps the signing surface intentionally narrow: callers can create a bound key, fetch its public key, or sign one of the structured payloads accepted by the crate. It does not expose a generic arbitrary-byte signing API.

Key IDs cross into platform-specific labels, tags, and metadata paths, so externally supplied IDs are constrained to the same auditable namespace created by the crate: `dk_` followed by unpadded base64url for 32 bytes. Remote-control target paths are also tied to each signed payload shape so connection proofs cannot be reused for enrollment endpoints, or vice versa.

## What changed

- Added the `codex-device-key` workspace crate.
- Added account/client-bound key creation with stable `dk_` key IDs.
- Added strict `key_id` validation before public-key lookup or signing reaches a provider.
- Added public-key lookup and structured signing APIs.
- Split remote-control client endpoint allowlists by connection vs enrollment payload shape.
- Added validation for key bindings, accepted payload fields, token expiration, and payload/key binding mismatches.
- Added flow-oriented docs on the validation helpers that gate provider signing.
- Added protection policy and protection-class types without wiring a platform provider yet.
- Added an unsupported default provider so platforms without an implementation fail explicitly instead of silently falling back to software-backed keys.
- Updated Cargo and Bazel lock metadata for the new crate and non-platform-specific dependencies.

## Stack

This is stacked on #18428.

## Validation

- Added unit coverage for strict `key_id` validation before provider use.
- Added unit coverage that rejects remote-control paths from the wrong signed payload shape.
- `just bazel-lock-update`
- `just bazel-lock-check`


codex-rs/device-key/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+) -> Result<(), DeviceKeyError> {
+    if payload.nonce.is_empty() {
+        return Err(DeviceKeyError::InvalidPayload("nonce must not be empty"));
+    }
+    let binding = KeyBinding {
+        account_user_id: &payload.account_user_id,
+        client_id: &payload.client_id,
+    };
+    validate_binding(&binding)?;
+    if payload.token_hash.is_empty() {
```
viyatb-oai: tokenHash is accepted as any non-empty string, so the API boundary does not enforce what token material is being bound. If this value binds the proof to a controller credential, please define the exact hash input and encoding, e.g. tokenSha256Base64url, validate the shape here, and require the backend to recompute it from the actual credential/challenge.

codex-rs/device-key/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct SignedPayload<'a> {
+    domain: &'static str,
+    payload: &'a DeviceKeySignPayload,
+}
+
+fn signed_payload(payload: &DeviceKeySignPayload) -> Result<Vec<u8>, DeviceKeyError> {
```
viyatb-oai: This makes Rust serde JSON the implicit security boundary for signatures. Since another service will verify these signatures, the exact signed bytes should not be private crate behavior. Please expose the signed payload bytes to the caller or use an explicitly specified canonical representation with test vectors.

codex-rs/device-key/src/lib.rs
```diff
@@ -0,0 +1,574 @@
...
+        account_user_id: &payload.account_user_id,
+        client_id: &payload.client_id,
+    };
+    validate_binding(&binding)?;
+    if payload.token_hash.is_empty() {
+        return Err(DeviceKeyError::InvalidPayload(
+            "tokenHash must not be empty",
+        ));
+    }
+    if payload.scopes.is_empty() || payload.scopes.iter().any(String::is_empty) {
```
viyatb-oai: Scopes are accepted as arbitrary non-empty caller input. This is safe only if they are treated as echo data and compared exactly against authenticated token/challenge claims. If there is only one supported launch use case, consider constraining this to the expected controller websocket scope or documenting/enforcing the exact comparison contract.

viyatb-oai:
just make sure backend must verify signature against signed_payload_base64 bytes exactly. The crate now exposes exact bytes, but the security property is only real if the backend does NOT reserialize payloads. Please ensure backend verification uses the provided bytes and the token hash/nonce from the same challenge.

euroelessar:
[codex] Addressed in `4f7294e`: `signedPayloadBase64` now documents that it is the exact byte string covered by `signatureDer`, and that verifiers must verify those bytes directly instead of reserializing the structured `payload`.

viyatb-oai:
Non-blocking hardening note: validate key_id shape more strictly before using it downstream. Created ids are safe random dk_... values, but sign/get_public currently only reject empty key ids; the macOS provider later uses the caller-provided key_id in Keychain tags and binding metadata paths. Requiring exactly dk_ + unpadded base64url(32 bytes) would make this namespace auditable and remove path/tag weirdness.

viyatb-oai (approved):

etraut-openai (approved):

PR #18430, opened 2026-04-18
app-server: implement device key v2 methods

euroelessar:
## Why

The device-key protocol needs an app-server implementation that keeps local key operations behind the same request-processing boundary as other v2 APIs.

app-server owns request dispatch, transport policy, documentation, and JSON-RPC error shaping. `codex-device-key` owns key binding, validation, platform provider selection, and signing mechanics. Keeping the adapter thin makes the boundary easier to review and avoids moving local key-management details into thread orchestration code.

## What changed

- Added `DeviceKeyApi` as the app-server adapter around `DeviceKeyStore`.
- Converted protocol protection policies, payload variants, algorithms, and protection classes to and from the device-key crate types.
- Encoded SPKI public keys and DER signatures as base64 protocol fields.
- Routed `device/key/create`, `device/key/public`, and `device/key/sign` through `MessageProcessor`.
- Rejected remote transports before provider access while allowing local `stdio` and in-process callers to reach the device-key API.
- Added stdio, in-process, and websocket tests for device-key validation and transport policy.
- Documented the device-key methods in the app-server v2 method list.

## Test coverage

- `device_key_create_rejects_empty_account_user_id`
- `in_process_allows_device_key_requests_to_reach_device_key_api`
- `device_key_methods_are_rejected_over_websocket`

## Stack

This is PR 3 of 4 in the device-key app-server stack. It is stacked on #18429.

## Validation

- `cargo test -p codex-app-server device_key`


codex-rs/app-server/src/message_processor.rs
```diff
@@ -1120,6 +1157,91 @@ impl MessageProcessor {
...
+            Err(error) => self.outgoing.send_error(request_id, error).await,
+        }
+    }
+
+    async fn reject_device_key_request_over_remote_transport(
+        &self,
+        request_id: ConnectionRequestId,
+        method: &str,
+    ) -> bool {
+        if matches!(
```
viyatb-oai: This gate uses the app-server process transport, not the origin of the specific connection. A stdio-launched app-server can also have backend remote-control virtual clients in the same process, so those clients would be allowed through as Stdio and could call device/key/sign. Device-key operations need to be allowed only for the local controller surface that owns the key; please track per-connection origin and reject these methods for remote-control/remote origins, with a regression test that exercises the remote-control path.

viyatb-oai:
[P1] Add regression coverage that device/key/* is rejected over remote‑control origin (not just raw websocket). Code blocks it via ConnectionOrigin::RemoteControl, but no test asserts the slingshot path can’t call device/key/sign.

euroelessar:
[codex] Addressed in `f664fa1`: added regression coverage for `ConnectionOrigin::RemoteControl` attempting `device/key/sign` through the app-server message processor. The test asserts the request is rejected with `INVALID_REQUEST` and the remote-transport rejection message.

codex-rs/app-server/src/transport/mod.rs
```diff
@@ -117,6 +118,20 @@ pub(crate) enum TransportEvent {
...
+pub(crate) enum ConnectionOrigin {
+    Stdio,
+    InProcess,
+    WebSocket,
+    RemoteControl,
+}
+
+impl ConnectionOrigin {
+    pub(crate) fn allows_device_key_requests(self) -> bool {
+        matches!(self, Self::Stdio | Self::InProcess)
```
etraut-openai, line 134: I recommend adding a big comment here so it's clear to future reviewers (and more importantly to Codex!) that it's important to restrict this to local connection origins. That way, if a new remote connection origin is added, it will be properly excluded from this check.

etraut-openai (commented):
One recommended addition for future maintainability, but otherwise looks good.

etraut-openai (approved):

PR #18431, opened 2026-04-18
app-server: add macOS device key provider

euroelessar:
## Why

The device-key crate needs a platform provider that can keep private keys non-extractable on macOS while preserving the crate-level protection policy and structured signing boundary.

macOS has two useful local protection classes for this API: hardware-backed Secure Enclave keys when available, and OS-protected non-extractable keys as an explicit fallback. Reporting which class was selected keeps that distinction visible to callers.

The Security.framework access-control flags used here provide device-local non-exportability and continuity while the user session is unlocked. They do not require UserPresence or Biometry for each signature, so the key itself is not treated as a complete same-controller compromise boundary. The API still relies on local-transport restriction, app-server authorization, and structured payload validation.

## What changed

- Added the macOS `DeviceKeyProvider` implementation backed by Security.framework.
- Created and loaded P-256 private keys by stable device-key ID.
- Preferred Secure Enclave keys by default and allowed OS-protected non-extractable fallback only when requested by policy.
- Returned SPKI DER public keys and DER-encoded ECDSA signatures through the existing crate API.
- Documented the macOS provider threat-model boundary around this-device-only keys without UserPresence or Biometry flags.
- Added macOS-only Security.framework and CoreFoundation dependencies.

## Stack

This is stacked on #18430.

## Validation

- `just bazel-lock-update`
- `just bazel-lock-check`


codex-rs/device-key/src/platform/macos.rs
```diff
@@ -0,0 +1,274 @@
...
+        MacKeyClass::OsProtectedNonextractable,
+    ] {
+        if let Some(key) = load_private_key(key_id, class)? {
+            return Ok(Some((class, key)));
+        }
+    }
+    Ok(None)
+}
+
+fn load_private_key(key_id: &str, class: MacKeyClass) -> Result<Option<SecKey>, DeviceKeyError> {
```
viyatb-oai: Loaded keys are classified by the lookup path rather than by verified key attributes. For the Secure Enclave path, the query should constrain or inspect the returned key attributes, especially kSecAttrTokenIDSecureEnclave, before reporting HardwareSecureEnclave. Otherwise a wrong/preexisting keychain item with the same tag can be reported as hardware-backed.

codex-rs/device-key/src/platform/macos.rs
```diff
@@ -0,0 +1,274 @@
...
+        if let Some((class, key)) = load_existing_private_key(request.key_id)? {
+            if !request.protection_policy.allows(class.protection_class()) {
+                return Err(DeviceKeyError::DegradedProtectionNotAllowed {
+                    available: class.protection_class(),
+                });
+            }
+            return key_info(request.key_id, class, &key);
+        }
+
+        match create_or_load_private_key(request.key_id, MacKeyClass::SecureEnclave) {
```
viyatb-oai: The fallback to OS-protected keys currently happens for any Secure Enclave creation/load failure when fallback is allowed. That can silently downgrade on unexpected keychain, access-control, duplicate-item, or load errors. Please limit fallback to known unavailable/unsupported Secure Enclave cases and surface unexpected Secure Enclave errors instead.

viyatb-oai:
[P2] OS‑protected fallback load path doesn’t validate non‑extractable attrs on existing keys; it is tag‑based. If we want the degraded path to still enforce non‑exportable, consider validating key attributes when loading, not just on create.

euroelessar:
[codex] Addressed in `30d2da0`: the macOS OS-protected fallback load query now includes `kSecAttrIsExtractable = false`, matching the create-time non-extractability constraint before an existing key is accepted.

viyatb-oai:
Non-blocking security note: this key setup gives us device-local non-exportability/continuity, but I would avoid describing it as malware-on-controller protection by itself. The macOS key is created with AccessibleWhenUnlockedThisDeviceOnly + kSecAccessControlPrivateKeyUsage; I don’t see a UserPresence/Biometry-style access-control flag here. That may be exactly fine for this rollout if sandboxing/backend enforcement are the real boundary, but worth spelling out in the threat model so we don’t overclaim what the platform key proves.

viyatb-oai (approved):

etraut-openai (approved):
Codex left a few additional code review comments that need your attention, but otherwise looks good.

euroelessar:
Update pushed in two commits:

- `21ab239bc5` adds `kSecUseDataProtectionKeychain` to the macOS private-key query/create attributes.
- `eed4fccd03` adds `codex-rs/scripts/check-device-key-app-server.py`, a repeatable local harness that spawns `codex app-server --listen stdio://` and exercises `device/key/create`, `device/key/sign`, and `device/key/public`.

Local check against the OpenAI-signed debug CLI currently reproduces the remaining blocker at `device/key/create`:

```json
{
  "failedMethod": "device/key/create",
  "error": {
    "code": -32603,
    "message": "device key platform error: macOS Keychain entitlements are required to create persistent device keys"
  }
}
```

Command used:

```bash
codex-rs/scripts/check-device-key-app-server.py --binary codex-rs/target/debug/codex
```

PR #18435, opened 2026-04-18
/statusline & /title - Shared preview values

canvrno-oai:
This PR makes the `/statusline` and `/title` setup UIs share one preview-value source instead of each surface using its own examples. Both pickers now render consistent live values when available, and stable placeholders when they are not. It also resolves live preview values at the shared preview-item layer, so `/title` preview can use real runtime values for title-specific cases like status text, task progress, and project-name fallback behavior.

- Adds a shared preview data model for status surfaces
- Maps status-line items and terminal-title items onto that shared preview list
- Feeds both setup views from the same chatwidget-derived preview data, with terminal-title-specific formatting applied before `/title` preview renders
- Keeps project-root preview aligned with status-line behavior while project in /title keeps its title fallback/truncation behavior
- Adds snapshot coverage for live-only, hardcoded-only, and mixed cases

Test Steps
- Open Codex TUI and launch `/statusline`.
- Toggle and reorder items, then verify the preview uses current session values when possible, and placeholder values for missing values (ex: no thread ID).
- Open `/title` and verify it shows the same normalized values, including live status/task-progress values when available.


canvrno-oai:
@codex review

canvrno-oai:
@codex review

canvrno-oai:
@codex  review

etraut-openai (approved):
Code looks good.

I also built locally and gave it a try and didn't see any visible or behavioral regressions.

PR #18446, opened 2026-04-18
Reserve missing preserved paths in Linux sandbox policy

evawong-oai:
## Summary

This PR protects preserved workspace metadata paths under workspace write sandboxes without breaking normal Git discovery. The preserved names are `.git`, `.codex`, and `.agents`.

BUGB 15632 tracks the Bugcrowd submission.

## Root Cause

The previous policy only treated `.git` and `.agents` as protected after they already existed. That left a first run gap where sandboxed code in a new workspace could create `.git/config`.

The report used `git init` followed by a Git config value that could persist on the host. A later host side Git command could then load attacker controlled repo metadata.

The hard part is that missing `.git` cannot always be materialized as a read only placeholder. If Codex runs inside a subdirectory of an existing Git repo or worktree, materializing a child `.git` changes Git discovery and hides the valid parent repo. In that case the child `.git` must stay absent while creation is still denied.

## Design

1. The protocol layer now carries a real preserved path policy primitive on writable roots through `preserved_path_names`.
2. Writable root policy rejects writes whose path components include `.git`, `.codex`, or `.agents` unless an explicit user write policy covers that preserved path.
3. Existing `.git`, `.codex`, and `.agents` paths stay read only under default writable roots.
4. Missing `.codex` and `.agents` are reserved under writable roots so first creation is blocked.
5. Missing child `.git` under a valid parent repo or worktree stays absent so normal `git status` and `git rev parse` keep discovering the parent.
6. Seatbelt enforces preserved names directly in the generated sandbox profile for writable roots.
7. Linux bwrap setup masks representable missing preserved paths with `/dev/null`.
8. Linux tracks missing preserved targets that cannot be safely materialized as `protected_create_targets`, runs bwrap with a parent helper only when cleanup or protected target checks are needed, and removes any forbidden preserved target that a sandboxed process managed to create before returning failure.
9. The command preflight no longer contains a hard coded table for `touch`, `mkdir`, `rm`, `rmdir`, `ln`, `mv`, `cp`, `install`, or `git init`. It only catches literal shell redirection targets early for user feedback. The sandbox policy is the security boundary.
10. The Linux launcher forwards termination signals to the bwrap child process group, clears the child pid after wait, and then runs cleanup.

## Validation

Current PR head: `3ae2e24e72eb879d66de9e28b6f32226f52d5eae`.

Current base main: `bb83eec825b74aaf06f74650d2c004b0629dd19a`.

1. Local macOS build passed for the CLI crate and Linux sandbox crate.
2. Formatting passed with `just fmt`.
3. Diff whitespace check passed.
4. Conflict marker scan passed across the checkout outside vendored code.
5. GitHub Actions passed on the current head, including macOS, Ubuntu, Windows, Bazel local builds, Bazel clippy, release builds, argument comment lint, format, spellcheck, SDK checks, cargo shear, cargo deny, blob policy, CLA, and required CI results.
6. Linux devbox validation passed all 46 preserved path smoke cases on the current head using literal `just c` for each case.
7. The devbox smoke run reported all 46 cases passed on the current head.
8. Case 46 covered Viyat's parent repo subdirectory scenario: parent Git discovery still works, `git init` and `mkdir .codex` are denied, normal coding can create the requested JSONL viewer, and no child `.git`, `.codex`, or `.agents` directory is left behind.
9. The detailed devbox log is available in the validation workspace.

### DIFF
--- a/codex-rs/linux-sandbox/src/bwrap.rs
+++ b/codex-rs/linux-sandbox/src/bwrap.rs
@@ -15,15 +15,18 @@ use std::collections::HashSet;
 use std::ffi::OsString;
 use std::fs;
 use std::fs::File;
+use std::fs::Metadata;
 use std::io;
 use std::os::fd::AsRawFd;
 use std::os::unix::ffi::OsStringExt;
+use std::os::unix::fs::MetadataExt;
 use std::path::Path;
 use std::path::PathBuf;
 use std::process::Command;
 
 use codex_protocol::error::CodexErr;
 use codex_protocol::error::Result;
+use codex_protocol::permissions::is_preserved_path_name;
 use codex_protocol::protocol::FileSystemSandboxPolicy;
 use codex_protocol::protocol::WritableRoot;
 use codex_utils_absolute_path::AbsolutePathBuf;
@@ -101,6 +104,121 @@ impl BwrapNetworkMode {
 pub(crate) struct BwrapArgs {
     pub args: Vec<String>,
     pub preserved_files: Vec<File>,
+    pub synthetic_mount_targets: Vec<SyntheticMountTarget>,
+    pub protected_create_targets: Vec<ProtectedCreateTarget>,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct FileIdentity {
+    dev: u64,
+    ino: u64,
+}
+
+impl FileIdentity {
+    fn from_metadata(metadata: &Metadata) -> Self {
+        Self {
+            dev: metadata.dev(),
+            ino: metadata.ino(),
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum SyntheticMountTargetKind {
+    EmptyFile,
+    EmptyDirectory,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct SyntheticMountTarget {
+    path: PathBuf,
+    kind: SyntheticMountTargetKind,
+    // If an empty preserved path was already present, remember its inode so
+    // cleanup does not delete a real pre-existing file or directory.
+    pre_existing_path: Option<FileIdentity>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct ProtectedCreateTarget {
+    path: PathBuf,
+}
+
+impl ProtectedCreateTarget {
+    pub(crate) fn missing(path: &Path) -> Self {
+        Self {
+            path: path.to_path_buf(),
+        }
+    }
+
+    pub(crate) fn path(&self) -> &Path {
+        &self.path
+    }
+}
+
+impl SyntheticMountTarget {
+    pub(crate) fn missing(path: &Path) -> Self {
+        Self {
+            path: path.to_path_buf(),
+            kind: SyntheticMountTargetKind::EmptyFile,
+            pre_existing_path: None,
+        }
+    }
+
+    pub(crate) fn missing_empty_directory(path: &Path) -> Self {
+        Self {
+            path: path.to_path_buf(),
+            kind: SyntheticMountTargetKind::EmptyDirectory,
+            pre_existing_path: None,
+        }
+    }
+
+    pub(crate) fn existing_empty_file(path: &Path, metadata: &Metadata) -> Self {
+        Self {
+            path: path.to_path_buf(),
+            kind: SyntheticMountTargetKind::EmptyFile,
+            pre_existing_path: Some(FileIdentity::from_metadata(metadata)),
+        }
+    }
+
+    fn existing_empty_directory(path: &Path, metadata: &Metadata) -> Self {
+        Self {
+            path: path.to_path_buf(),
+            kind: SyntheticMountTargetKind::EmptyDirectory,
+            pre_existing_path: Some(FileIdentity::from_metadata(metadata)),
+        }
+    }
+
+    pub(crate) fn preserves_pre_existing_path(&self) -> bool {
+        self.pre_existing_path.is_some()
+    }
+
+    pub(crate) fn path(&self) -> &Path {
+        &self.path
+    }
+
+    pub(crate) fn kind(&self) -> SyntheticMountTargetKind {
+        self.kind
+    }
+
+    pub(crate) fn should_remove_after_bwrap(&self, metadata: &Metadata) -> bool {
+        match self.kind {
+            SyntheticMountTargetKind::EmptyFile => {
+                if !metadata.file_type().is_file() || metadata.len() != 0 {
+                    return false;
+                }
+            }
+            SyntheticMountTargetKind::EmptyDirectory => {
+                if !metadata.file_type().is_dir() {
+                    return false;
+                }
+            }
+        }
+
+        match self.pre_existing_path {
+            Some(pre_existing_path) => pre_existing_path != FileIdentity::from_metadata(metadata),
+            None => true,
+        }
+    }
 }
 
 /// Wrap a command with bubblewrap so the filesystem is read-only by default,
@@ -126,6 +244,8 @@ pub(crate) fn create_bwrap_command_args(
             Ok(BwrapArgs {
                 args: command,
                 preserved_files: Vec::new(),
+                synthetic_mount_targets: Vec::new(),
+                protected_create_targets: Vec::new(),
             })
         } else {
             Ok(create_bwrap_flags_full_filesystem(command, options))
@@ -165,6 +285,8 @@ fn create_bwrap_flags_full_filesystem(command: Vec<String>, options: BwrapOption
     BwrapArgs {
         args,
         preserved_files: Vec::new(),
+        synthetic_mount_targets: Vec::new(),
+        protected_create_targets: Vec::new(),
     }
 }
 
@@ -179,6 +301,8 @@ fn create_bwrap_flags(
     let BwrapArgs {
         args: filesystem_args,
         preserved_files,
+        synthetic_mount_targets,
+        protected_create_targets,
     } = create_filesystem_args(
         file_system_sandbox_policy,
         sandbox_policy_cwd,
@@ -216,6 +340,8 @@ fn create_bwrap_flags(
     Ok(BwrapArgs {
         args,
         preserved_files,
+        synthetic_mount_targets,
+        protected_create_targets,
     })
 }
 
@@ -275,7 +401,7 @@ fn create_filesystem_args(
     unreadable_roots.sort();
     unreadable_roots.dedup();
 
-    let mut args = if file_system_sandbox_policy.has_full_disk_read_access() {
+    let args = if file_system_sandbox_policy.has_full_disk_read_access() {
         // Read-only root, then mount a minimal device tree.
         // In bubblewrap (`bubblewrap.c`, `SETUP_MOUNT_DEV`), `--dev /dev`
         // creates the standard minimal nodes: null, zero, full, random,
@@ -348,7 +474,12 @@ fn create_filesystem_args(
 
         args
     };
-    let mut preserved_files = Vec::new();
+    let mut bwrap_args = BwrapArgs {
+        args,
+        preserved_files: Vec::new(),
+        synthetic_mount_targets: Vec::new(),
+        protected_create_targets: Vec::new(),
+    };
     let mut allowed_write_paths = Vec::with_capacity(writable_roots.len());
     for writable_root in &writable_roots {
         let root = writable_root.root.as_path();
@@ -379,12 +510,7 @@ fn create_filesystem_args(
     unreadable_ancestors_of_writable_roots.sort_by_key(|path| path_depth(path));
 
     for unreadable_root in &unreadable_ancestors_of_writable_roots {
-        append_unreadable_root_args(
-            &mut args,
-            &mut preserved_files,
-            unreadable_root,
-            &allowed_write_paths,
-        )?;
+        append_unreadable_root_args(&mut bwrap_args, unreadable_root, &allowed_write_paths)?;
     }
 
     for writable_root in &sorted_writable_roots {
@@ -398,13 +524,13 @@ fn create_filesystem_args(
             .filter(|unreadable_root| root.starts_with(unreadable_root))
             .max_by_key(|unreadable_root| path_depth(unreadable_root))
         {
-            append_mount_target_parent_dir_args(&mut args, root, masking_root);
+            append_mount_target_parent_dir_args(&mut bwrap_args.args, root, masking_root);
         }
 
         let mount_root = symlink_target.as_deref().unwrap_or(root);
-        args.push("--bind".to_string());
-        args.push(path_to_string(mount_root));
-        args.push(path_to_string(mount_root));
+        bwrap_args.args.push("--bind".to_string());
+        bwrap_args.args.push(path_to_string(mount_root));
+        bwrap_args.args.push(path_to_string(mount_root));
 
         let mut read_only_subpaths: Vec<PathBuf> = writable_root
             .read_only_subpaths
@@ -415,9 +541,16 @@ fn create_filesystem_args(
         if let Some(target) = &symlink_target {
             read_only_subpaths = remap_paths_for_symlink_target(read_only_subpaths, root, target);
         }
+        append_protected_create_targets_for_writable_root(
+            &mut bwrap_args,
+            writable_root,
+            root,
+            symlink_target.as_deref(),
+            &read_only_subpaths,
+        );
         read_only_subpaths.sort_by_key(|path| path_depth(path));
         for subpath in read_only_subpaths {
-            append_read_only_subpath_args(&mut args, &subpath, &allowed_write_paths)?;
+            append_read_only_subpath_args(&mut bwrap_args, &subpath, &allowed_write_paths)?;
         }
         let mut nested_unreadable_roots: Vec<PathBuf> = unreadable_roots
             .iter()
@@ -430,12 +563,7 @@ fn create_filesystem_args(
         }
         nested_unreadable_roots.sort_by_key(|path| path_depth(path));
         for unreadable_root in nested_unreadable_roots {
-            append_unreadable_root_args(
-                &mut args,
-                &mut preserved_files,
-                &unreadable_root,
-                &allowed_write_paths,
-            )?;
+            append_unreadable_root_args(&mut bwrap_args, &unreadable_root, &allowed_write_paths)?;
         }
     }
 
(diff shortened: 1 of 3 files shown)


evawong-oai:
@codex review

evawong-oai:
Reviewer evidence update.

I extended the same preserved path rule to missing `.agents`. The existing `.agents` directory was already protected when present. The missing case now follows `.git` and `.codex` for the active workspace root, while explicit user write rules still win.

I also reran linked worktree validation on a Linux devbox with system `bwrap` installed at `/usr/bin/bwrap`, version `0.9.0`. The `.git` worktree pointer case passed. Normal writes inside the worktree succeeded. Overwriting `.git`, replacing `.git`, and writing through the resolved gitdir were blocked. The pointer file stayed unchanged and no `.git` directory was created.

I fixed the devbox validation gap where blocked missing `.codex` and `.agents` attempts left zero byte host files. The Linux bwrap path now tracks synthetic missing mount targets, runs bwrap in a child only when cleanup is needed, removes only zero byte synthetic targets after preflight and the real run, then preserves the bwrap exit status.

The rerun passed on the Linux devbox with system `bwrap` at `/usr/bin/bwrap`, version `0.9.0`, against commit `87f8ccb2bc22abb0e2c403da30866ae014941ae7`. Normal checkout and linked worktree cases passed for `.codex`, `.git`, and `.agents`. Missing `.codex` and `.agents` writes were blocked and left no host side file or directory immediately after the sandbox command returned. Normal writes still succeeded, the linked `.git` pointer stayed unchanged, and resolved gitdir writes stayed blocked. The validation transcript did not contain the previous proc mount setup error.

I also locally passed sandboxing crate tests, protocol crate tests, format check, and whitespace diff check.

The branch remains one commit, now `87f8ccb2bc22abb0e2c403da30866ae014941ae7`.

codex-rs/protocol/src/permissions.rs
```diff
@@ -1316,7 +1319,7 @@ fn default_read_only_subpaths_for_writable_root(
     // writable root itself.
     let top_level_git_is_file = top_level_git.as_path().is_file();
     let top_level_git_is_dir = top_level_git.as_path().is_dir();
-    if top_level_git_is_dir || top_level_git_is_file {
+    if top_level_git_is_dir || top_level_git_is_file || protect_missing_preserved_paths {
```
viyatb-oai: **[P1] Missing `.git` mask breaks parent repo discovery**

When the cwd is a subdirectory of a Git repo, this policy now adds a missing `<cwd>/.git` read-only subpath. On Linux, the bwrap layer materializes missing read-only subpaths as an empty file, so Git stops at that synthetic `<cwd>/.git` and fails instead of walking up to the parent repo.

Minimal repro for the behavior Git sees:

```sh
mkdir -p /tmp/codex-parent/sub
git -C /tmp/codex-parent init
git -C /tmp/codex-parent/sub status --short   # succeeds; discovers parent repo
cp /dev/null /tmp/codex-parent/sub/.git
git -C /tmp/codex-parent/sub status --short   # fatal: invalid gitfile format: .../sub/.git
```

That means running Codex from a repo subdirectory can break common Git operations after the new missing `.git` mask is applied. I think `.git` needs a different missing-path strategy than `.codex`/`.agents`, or the default preserved path rule should avoid materializing a missing `.git` under subdirectory workspace roots.

codex-rs/protocol/src/protocol.rs
```diff
@@ -1310,7 +1310,7 @@ fn default_read_only_subpaths_for_writable_root(
     // writable root itself.
     let top_level_git_is_file = top_level_git.as_path().is_file();
     let top_level_git_is_dir = top_level_git.as_path().is_dir();
-    if top_level_git_is_dir || top_level_git_is_file {
+    if top_level_git_is_dir || top_level_git_is_file || protect_missing_preserved_paths {
```
viyatb-oai: **[P2] Centralize preserved path policy to avoid drift**

This PR now keeps the same preserved-path policy in at least three places: this helper, the parallel `default_read_only_subpaths_for_writable_root` in `permissions.rs`, and the Linux-only `has_preserved_path_name` check in `bwrap.rs`. Because the security property depends on `.git`, `.agents`, `.codex`, and gitdir pointer handling staying in sync, can we centralize this into one shared protocol helper/constant plus a shared basename predicate for the Linux transient-file case? Otherwise the next preserved-path change can easily update one path but miss another.

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -787,6 +859,8 @@ fn append_mount_target_parent_dir_args(args: &mut Vec<String>, mount_target: &Pa
 
 fn append_read_only_subpath_args(
     args: &mut Vec<String>,
+    preserved_files: &mut Vec<File>,
+    synthetic_mount_targets: &mut Vec<SyntheticMountTarget>,
```
viyatb-oai: **[P3] Consider a small bwrap mount-arg context if this grows**

`append_read_only_subpath_args` and `append_unreadable_root_args` now both thread `args`, `preserved_files`, and `synthetic_mount_targets` together. If we keep adding missing-path cleanup behavior, a tiny local builder/context that owns those three mutable fields would make the mutation boundary clearer and reduce parameter sprawl. Not a blocker, but this is right at the point where the helper signatures are starting to carry implementation state.

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -562,6 +564,78 @@ fn resolve_true_command() -> String {
...
 
+fn run_or_exec_bwrap(bwrap_args: crate::bwrap::BwrapArgs) -> ! {
+    if bwrap_args.synthetic_mount_targets.is_empty() {
+        exec_bwrap(bwrap_args.args, bwrap_args.preserved_files);
+    }
+    run_bwrap_in_child_with_synthetic_mount_cleanup(bwrap_args);
+}
+
+fn run_bwrap_in_child_with_synthetic_mount_cleanup(bwrap_args: crate::bwrap::BwrapArgs) -> ! {
+    let synthetic_mount_targets = bwrap_args.synthetic_mount_targets.clone();
```
viyatb-oai: **[P3] Avoid cloning mount targets before fork**

Both this path and the proc preflight path clone `bwrap_args.synthetic_mount_targets` only so the parent can clean up after the child execs bwrap. We can avoid that allocation by destructuring `BwrapArgs` before `fork()`: keep `synthetic_mount_targets` in the parent scope, and move only `args`/`preserved_files` into `exec_bwrap` in the child. Tiny, but this path runs on every command that has synthetic mount targets.

evawong-oai:
@codex review

codex-rs/linux-sandbox/src/bwrap.rs
evawong-oai: Thanks, I am leaving this out of the BUGB fix. I added it to the MITM task list as follow up work and will handle the mount argument context cleanup in a separate PR.

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -562,6 +564,78 @@ fn resolve_true_command() -> String {
...
+        let err = std::io::Error::last_os_error();
+        panic!("failed to fork for bubblewrap: {err}");
+    }
+
+    if pid == 0 {
+        exec_bwrap(bwrap_args.args, bwrap_args.preserved_files);
+    }
+
+    let mut status: libc::c_int = 0;
+    let wait_res = unsafe { libc::waitpid(pid, &mut status as *mut libc::c_int, 0) };
```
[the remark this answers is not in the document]
evawong-oai: Mapped to commit `a1e35f301`. The forked bwrap parent now forwards termination signals while waiting, retries interrupted waits, and gives the child a parent death signal before exec.

codex-rs/protocol/src/permissions.rs
evawong-oai: Mapped to commit `a1e35f301`. The missing child `.git` mask is skipped only when a valid ancestor Git marker exists, so subdirectory workspaces keep parent repo discovery.

codex-rs/protocol/src/protocol.rs
evawong-oai: Mapped to commit `a1e35f301`. Preserved path names and the basename predicate now live in protocol and are shared by policy generation and Linux bwrap transient path handling.

codex-rs/linux-sandbox/src/linux_run_main.rs
evawong-oai: Mapped to commit `a1e35f301`. The bwrap args are destructured before fork, so the parent keeps synthetic mount targets and the child moves only exec args and preserved files.

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -562,6 +571,131 @@ fn resolve_true_command() -> String {
...
+        }
+    }
+}
+
+extern "C" fn forward_signal_to_bwrap_child(signal: libc::c_int) {
+    let pid = BWRAP_CHILD_PID.load(Ordering::SeqCst);
+    if pid > 0 {
+        unsafe {
+            libc::kill(pid, signal);
+        }
```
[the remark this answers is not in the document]
evawong-oai: Mapped to commit `a1e35f301`. The parent clears the forwarded child pid after wait returns and before synthetic target cleanup and status propagation.

codex-rs/protocol/src/permissions.rs
```diff
@@ -1387,8 +1410,17 @@ fn has_explicit_resolved_path_entry(
     entries.iter().any(|entry| &entry.path == path)
 }
 
+fn has_ancestor_git_metadata(writable_root: &AbsolutePathBuf) -> bool {
+    writable_root.as_path().parent().is_some_and(|parent| {
+        parent.ancestors().any(|ancestor| {
+            std::fs::symlink_metadata(ancestor.join(PRESERVED_GIT_PATH_NAME)).is_ok()
```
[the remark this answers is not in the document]
evawong-oai: Mapped to commit `a1e35f301`. Ancestor `.git` now suppresses the missing child `.git` mask only when it is a directory or a valid gitdir pointer. A stray file no longer disables the protection.

codex-rs/linux-sandbox/src/bwrap.rs
evawong-oai: Addressed in commit `36d53e598`. I rebased onto current `origin/main` and kept the cleanup in this PR. The bwrap mount helpers now mutate one `BwrapArgs` context instead of threading `args`, `preserved_files`, and `synthetic_mount_targets` through each helper.

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -562,6 +581,297 @@ fn resolve_true_command() -> String {
...
+                    target.target.path().display()
+                ),
+            }
+        }
+
+        for target in targets.iter().rev() {
+            if synthetic_mount_marker_dir_has_active_process(&target.marker_dir) {
+                continue;
+            }
+            remove_synthetic_mount_target(&target.target);
```
viyatb-oai, line 933: [P1] Synthetic preserved-path placeholders leak after Codex exits

The cleanup here does not reliably remove the host files that bwrap materializes for missing preserved paths during a real Codex workflow. I rebuilt `~/code/codex/codex-rs/target/debug/codex` from the PR branch on a linux machine (`68debc47917bf87cd405ea979c3d6bc48ea0a451`, `codex` timestamp `2026-04-22 01:52:36 +0000`) and ran it from a fresh non-git `~/y/codex-preserved-path-repro` with `--sandbox workspace-write --skip-git-repo-check -c approval_policy="on-request"`. The model first ran `git init` and `mkdir .codex`, then built a Go JSONL viewer. After `codex exec` exited, the host directory still contained `-r--r--r-- 0 .git`, `.codex`, and `.agents`; `git init` failed with `fatal: invalid gitfile format: /home/dev-user/y/codex-preserved-path-repro/.git`, and `mkdir .codex` failed with `File exists`. This leaves new projects permanently poisoned after Codex exits, so the synthetic mount cleanup needs to remove these placeholders before the PR lands.

viyatb-oai: Root cause in PR head:

First sandbox starts with missing .git/.codex/.agents, so it registers them as SyntheticMountTarget::missing(...).
While that sandbox is still alive, a second sandbox starts and sees the host-side zero-byte placeholder.
append_read_only_subpath_args() treats that as an “existing empty file” and calls append_existing_empty_file_bind_data_args(...).
That records the current inode as pre_existing_file.
First sandbox exits, sees another active marker, and correctly skips cleanup.
Second sandbox exits last, but should_remove_after_bwrap() returns false because the file matches its recorded “pre-existing” inode.
Marker dir is removed, placeholder remains forever.
So the broken edge is this pairing:

bwrap.rs:861-868: transient empty preserved file is registered as existing/pre-existing.
bwrap.rs:158-166: existing/pre-existing same inode means “do not remove.”
linux_run_main.rs:713-717: earlier missing-owner skips cleanup while another marker is active.
I confirmed with a minimal devbox repro using the built PR binary, no apply_patch involved:

```
~/y/codex-cleanup-concurrency-1776823706

Two overlapping codex sandbox linux --full-auto -- sh -c sleep ... invocations left:

-r--r--r-- 0 .git
-r--r--r-- 0 .codex
-r--r--r-- 0 .agents
```
this is triggered because multiple shell/fs-helper sandboxes overlap.

The fix should make “empty preserved file with active synthetic marker” join synthetic ownership, not classify as a real pre-existing file. Conservatively: only preserve an empty .git/.codex/.agents when there is no active marker/owned identity for that path. Do not blindly delete every empty preserved file, because a user could have had a real zero-byte file there before Codex started.

evawong-oai: Addressed in commit `13dff2c56`.

The cleanup now treats an empty preserved path with an active synthetic owner as synthetic ownership, so the last exiting sandbox removes the placeholder instead of preserving it as a real prior file.

I also kept the conservative guard you called out: real prior empty preserved files are still preserved when there is no active synthetic owner.

Coverage added:

1. `cleanup_synthetic_mount_targets_removes_transient_file_after_concurrent_owner_exits`
2. `cleanup_synthetic_mount_targets_preserves_real_pre_existing_empty_file`

CI is green on the rewritten two commit stack.

codex-rs/protocol/src/permissions.rs
```diff
@@ -1387,8 +1394,39 @@ fn has_explicit_resolved_path_entry(
...
 
+pub fn gitdir_pointer_for_child_git(path: &Path) -> Option<Vec<u8>> {
+    if path.file_name() != Some(OsStr::new(PRESERVED_GIT_PATH_NAME)) {
+        return None;
+    }
+    let parent = path.parent()?;
+    let gitdir = parent
+        .ancestors()
+        .find_map(|ancestor| git_metadata_dir(&ancestor.join(PRESERVED_GIT_PATH_NAME)))?;
+    Some(format!("gitdir: {}\n", gitdir.as_path().display()).into_bytes())
```
viyatb-oai: [P1] Keep child `.git` absent when an ancestor repo exists

This still breaks the repo-subdirectory workflow even though the invalid-gitfile crash is fixed. `gitdir_pointer_for_child_git()` synthesizes `<cwd>/.git` when `cwd` is under an existing repo, so Git stops there and treats `<cwd>` as a new worktree root instead of continuing to the parent. I re-smoked PR head `16eeec07` on Linux with the rebuilt PR binary: host `git -C repo/sub rev-parse --show-toplevel` returns `.../repo`, but sandboxed `git rev-parse --show-toplevel` from `repo/sub` returns `.../repo/sub`, and `git status --short` reports the synthetic `.agents`/`.codex` placeholders as untracked in that fake root. Suggested fix: do not reserve/synthesize a missing child `.git` when `writable_root` already has ancestor Git metadata; leave `<writable_root>/.git` absent so Git can keep walking upward. Only reserve missing `.git` when there is no ancestor repo, or when `.git` already exists at that root and must be protected.

evawong-oai: Addressed in commit `06ad68940d`. The default preserved path policy now leaves a missing child `.git` absent when the writable root is inside a valid ancestor repo, so Git can keep walking to the parent repo. Existing `.git` at the writable root is still protected, and roots without valid ancestor Git metadata still reserve missing `.git`. Validation passed: full codex protocol tests, Linux sandbox crate test target, and cargo fmt check.

viyatb-oai (approved):

codex-rs/cli/src/debug_sandbox.rs
```diff
@@ -275,6 +285,126 @@ async fn run_command_under_sandbox(
     handle_exit_status(status);
 }
 
+fn preserved_path_write_forbidden_reason(
```
bolinfest: Something feels off: this the logic that powers `codex sandbox`. This file should be small, as it should be calling into existing library code. This seems off.

codex-rs/cli/src/debug_sandbox.rs
```diff
@@ -275,6 +285,126 @@ async fn run_command_under_sandbox(
...
+    let program = command.first().map(|program| {
+        Path::new(program)
+            .file_name()
+            .and_then(|name| name.to_str())
+            .unwrap_or(program)
+    })?;
+
+    match program {
+        "git" => git_init_preserved_path_write(command, cwd, file_system_sandbox_policy),
+        "touch" | "mkdir" | "rm" | "rmdir" | "ln" | "mv" | "cp" | "install" => command
```
bolinfest: Why are we special-casing things?

evawong-oai: This is for resolving parent/child `.git` folders security concern.

codex-rs/cli/src/debug_sandbox.rs
evawong-oai: Agreed. I moved this out of the CLI sandbox entry point in the latest commit.

The CLI sandbox path now calls the shared preserved path write guard from the shell command crate, and the agent shell path calls the same helper. The debug sandbox file is back to being an entry point instead of owning the policy logic.

codex-rs/cli/src/debug_sandbox.rs
evawong-oai: I have since then moved this list into a policy configuration rather than having policy defined in code for separation of concerns.

evawong-oai:
Split this into a manual stack so each review surface is smaller.

1. #19846 policy primitive
2. #19847 macOS Seatbelt enforcement
3. #19848 shell preflight
4. #19849 runtime permission propagation
5. #18446 Linux bubblewrap enforcement

PR 18446 is now the Linux child at the top of the stack. The final tree is unchanged from the rebased full branch, but the review surfaces are separated by policy, macOS, preflight, runtime propagation, and Linux.

evawong-oai:
Follow up: the old auto merge setting carried over when I changed this PR base, so GitHub marked this old PR merged into the temporary parent branch.

I restored the parent branch to its scoped runtime permission commit and opened #19852 as the current Linux child for review. Use #19852 for the active top of the stack.

PR #18449, opened 2026-04-18
[codex] Describe uninstalled cross-repo plugin reads

xli-oai:
## Summary
- Populate `PluginDetail.description` in core for uninstalled cross-repo plugins when detailed fields are unavailable until install.
- Include the source Git URL plus optional path/ref/sha details in that fallback description.
- Keep `details_unavailable_reason` as the structured signal while app-server forwards the description normally.
- Add plugin-read coverage proving the response does not clone the remote source just to show the message.

## Why
Uninstalled cross-repo plugins intentionally return sparse detail data so listing/reading does not clone the plugin source. Without a description, Desktop and TUI detail pages look like an ordinary empty plugin. This gives users a concrete explanation and source pointer while keeping the existing structured reason available for callers.

## Validation
- `just fmt`
- `cargo test -p codex-core read_plugin_for_config_uninstalled_git_source_requires_install_without_cloning`
- `cargo test -p codex-app-server plugin_read --test all`
- `just fix -p codex-core`

Note: `cargo test -p codex-app-server` was also attempted before the latest refactor and failed broadly in unrelated v2 thread/realtime/review/skills suites; the new plugin-read test passed in that run as well.

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -239,7 +239,6 @@ use codex_core::find_thread_path_by_id_str;
 use codex_core::path_utils;
 use codex_core::plugins::MarketplaceAddError;
 use codex_core::plugins::OPENAI_CURATED_MARKETPLACE_NAME;
-use codex_core::plugins::PluginDetailsUnavailableReason as CorePluginDetailsUnavailableReason;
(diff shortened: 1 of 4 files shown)


xl-openai (approved):

xli-oai:
@codex review

PR #18452, opened 2026-04-18
feat: Support remote plugin list/read.

xl-openai:
Add a temporary internal remote_plugin feature flag that merges remote marketplaces into plugin/list and routes plugin/read through the remote APIs when needed, while keeping pure local marketplaces working as before.


codex-rs/app-server/src/codex_message_processor/plugins.rs
```diff
@@ -0,0 +1,640 @@
+use super::*;
+
+impl CodexMessageProcessor {
```
xl-openai, line 4: Try move plugin API out from codex_message_processor.rs as the file is too large now.

codex-rs/core-plugins/src/remote_legacy.rs
```diff
@@ -0,0 +1,313 @@
+use crate::remote::RemotePluginServiceConfig;
```
xl-openai, line 1: Git doesn’t detect it as a rename, but remote_legacy.rs is the old remote.rs moved over intact. It remains wired to the legacy plugin API and is currently only used for the initial sync path. The new implementation now lives in remote.rs.

codex-rs/app-server/src/codex_message_processor/plugins.rs
```diff
@@ -0,0 +1,641 @@
...
+                            },
+                        )
+                        .await;
+                    return;
+                }
+                let auth = self.auth_manager.auth().await;
+                let remote_plugin_service_config = RemotePluginServiceConfig {
+                    chatgpt_base_url: config.chatgpt_base_url.clone(),
+                };
+                let remote_plugin_id = format!("{plugin_name}@{remote_marketplace_name}");
```
viyatb-oai, line 300: [P2] `plugin_name` is request-controlled here, so concatenating it into `"{plugin_name}@{remote_marketplace_name}"` and then later interpolating that value into `"{base_url}/ps/plugins/{plugin_id}"` creates a path/query injection primitive against the authenticated ChatGPT plugin API. A crafted plugin name containing `/`, `?`, or `#` can rewrite the request target before `fetch_remote_plugin_detail()` does the GET. We should either carry the raw `plugin_name`/marketplace separately and build the URL with path-segment escaping (for example via `Url::path_segments_mut()`), or percent-encode `plugin_id` before appending it to the path.

xl-openai: Will add a safeguard but app-server is not a public facing service the caller should just be codex app/TUI.

xli-oai (approved):

xl-openai:
@codex review

codex-rs/core-plugins/src/remote.rs
```diff
@@ -1,317 +1,632 @@
...
-        .map_err(|source| RemotePluginFetchError::Request {
-            url: url.clone(),
-            source,
-        })?;
-    let status = response.status();
-    let body = response.text().await.unwrap_or_default();
-    if !status.is_success() {
-        return Err(RemotePluginFetchError::UnexpectedStatus { url, status, body });
+    let installed_plugin = fetch_installed_plugins_for_scope(config, auth, scope)
+        .await?
```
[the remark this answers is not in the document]
xl-openai: That's OK.

codex-rs/core-plugins/src/remote.rs
```diff
@@ -1,317 +1,632 @@
...
+) -> Result<Vec<RemoteMarketplace>, RemotePluginCatalogError> {
+    let auth = ensure_chatgpt_auth(auth)?;
+    let mut directory_by_scope =
+        BTreeMap::<RemotePluginScope, BTreeMap<String, RemotePluginDirectoryItem>>::new();
+    let mut installed_by_scope =
+        BTreeMap::<RemotePluginScope, BTreeMap<String, RemotePluginInstalledItem>>::new();
+
+    for scope in RemotePluginScope::all() {
+        let directory_plugins = fetch_directory_plugins_for_scope(config, auth, scope).await?;
+        let installed_plugins = fetch_installed_plugins_for_scope(config, auth, scope).await?;
```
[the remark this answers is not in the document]
xl-openai: That's OK.

codex-rs/core/src/plugins/manager_tests.rs
```diff
@@ -1819,9 +1816,7 @@ plugins = true
             name: "openai-curated".to_string(),
             path: AbsolutePathBuf::try_from(curated_root.join(".agents/plugins/marketplace.json"))
                 .unwrap(),
-            interface: Some(MarketplaceInterface {
-                display_name: Some(OPENAI_CURATED_MARKETPLACE_DISPLAY_NAME.to_string()),
-            }),
```
sayan-oai, line 1824: does this mean we no longer have the formatted `OpenAI Curated` in product surfaces and they see `openai-curated` instead?

xl-openai: I don’t understand the original purpose of this code — our marketplace already defines its own marketplace display name.

Also, in the Codex App we actually override the value to “Built by OpenAI,” but the lookup uses the marketplace name rather than the display name.

sayan-oai (commented):
@xl-openai sorry to dump, but codex findings below.

1 and 2 can be deferred if we are not doing TUI changes in this PR like you said.
3 seems small, worth checking.
4 seems like it's worth checking, but if we are maintaining all the remote plugins we can probably be more lax.
5 is a nit.

Findings

1. Remote plugin details are implemented in app-server, but the TUI never calls them.
In plugins.rs (line 1098), the TUI decides details are viewable only when marketplace.path.is_some(). Remote marketplaces intentionally have path: None, so plugins.rs (line 1134) builds no Enter action.

2. Installed remote plugins get a toggle that only writes local config, so it will not actually change remote state.

3. Remote pagination can loop forever.
remote.rs (line 519) and remote.rs (line 538) keep requesting pages until next_page_token is None, with no max page count and no repeated-token detection.

4. Remote default prompts bypass the documented caps.
The protocol says plugin default prompts are capped at 3 entries and 128 chars in v2.rs (line 3881). Local manifests enforce this in manifest.rs (line 244). Remote mapping just copies the prompt into a one-item vec in remote.rs (line 437).

5.: The app-server docs are stale for this API change.
README.md (line 189) still describes only local plugin list/read. It does not mention remote marketplaces, remoteMarketplaceName, or nullable marketplacePath / skill path. Example: a client following the README would still assume plugin/read requires marketplacePath.

codex-rs/app-server/src/codex_message_processor/plugins.rs
```diff
@@ -0,0 +1,671 @@
...
+        let roots = cwds.unwrap_or_default();
+        plugins_manager.maybe_start_non_curated_plugin_cache_refresh(&roots);
+
+        let config = match self.load_latest_config(/*fallback_cwd*/ None).await {
+            Ok(config) => config,
+            Err(err) => {
+                self.outgoing.send_error(request_id, err).await;
+                return;
+            }
+        };
```
sayan-oai: saw that later in this function we check if `Feature::Plugins` is enabled; im surprised we dont already know that by that point. why not just swap the config load and cache refresh and early return if the feature is disabled?

sayan-oai (approved):

PR #18499, opened 2026-04-18
Fix plugin cache panic when cwd is unavailable

etraut-openai:
## Summary

Fixes #16637. (I hit this bug after 11h of work on a long-running task.)

Plugin cache initialization could panic when an already-absolute cache path was normalized through `AbsolutePathBuf::from_absolute_path`, because that path still consulted `current_dir()`.

This changes absolute-path normalization so already-absolute paths do not depend on cwd, and makes plugin cache root construction available as a fallible path through `PluginStore::try_new()`. Plugin cache subpaths now use `AbsolutePathBuf::join()` instead of re-absolutizing derived absolute paths.

### DIFF
--- a/codex-rs/core-plugins/src/loader.rs
+++ b/codex-rs/core-plugins/src/loader.rs
@@ -129,7 +129,7 @@ pub fn refresh_curated_plugin_cache(
     plugin_version: &str,
     configured_curated_plugin_ids: &[PluginId],
 ) -> Result<bool, String> {
-    let store = PluginStore::new(codex_home.to_path_buf());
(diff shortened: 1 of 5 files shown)


xl-openai (approved):
Stamp to unblock, but I wonder if the root cause (current_dir() being unavailable) affects more than just plugins. Are we okay with that state?

etraut-openai:
Yes, unavailable cwd can surface elsewhere, but this PR covers the accidental dependency on cwd for already-absolute paths. I think we should be resilient to a panic in this case. The plugin code was a victim of this bug in `absolutize()`. The fix in that function is the most important part of this PR.

I also added additional hardening and error reporting specifically for the plugin code. That's less important, but it's useful for diagnosing any issue related to invalid plugin cache paths.

PR #18502, opened 2026-04-18
Support multiple cwd filters for thread list

acrognale-oai:
## Summary

- Teach app-server `thread/list` to accept either a single `cwd` or an array of cwd filters, returning threads whose recorded session cwd matches any requested path
- Add `useStateDbOnly` as an explicit opt-in fast path for callers that want to answer `thread/list` from SQLite without scanning JSONL rollout files
- Preserve backwards compatibility: by default, `thread/list` still scans JSONL rollouts and repairs SQLite state
- Wire the new cwd array and SQLite-only options through app-server, local/remote thread-store, rollout listing, generated TypeScript/schema fixtures, proto output, and docs

## Test Plan

- `cargo test -p codex-rollout`
- `cargo test -p codex-thread-store`
- `cargo test -p codex-app-server thread_list`
- `just fmt`
- `just fix -p codex-app-server-protocol -p codex-rollout -p codex-thread-store -p codex-app-server`


codex-rs/app-server-protocol/schema/json/ClientRequest.json
jif-oai, line 3065: the null seems to be actually lost. This is a breaking change

jif-oai (approved):
lgtm

PR #18504, opened 2026-04-18
Rebrand approvals reviewer config to auto-review

won-openai:
### Why

Auto-review is the user-facing name for the approvals reviewer, but the config/API value still exposed the old `guardian_subagent` name. That made new configs and generated schemas point users at Guardian terminology even though the intended product surface is Auto-review.

This PR updates the external `approvals_reviewer` value while preserving compatibility for existing configs and clients.

### What changed

- Makes `auto_review` the canonical serialized value for `approvals_reviewer`.
- Keeps `guardian_subagent` accepted as a legacy alias.
- Keeps `user` accepted and serialized as `user`.
- Updates generated config and app-server schemas so `approvals_reviewer` includes:
  - `user`
  - `auto_review`
  - `guardian_subagent`
- Updates app-server README docs for the reviewer value.
- Updates analytics and config requirements tests for the canonical auto_review value.

### Compatibility

Existing configs and API payloads using:

```toml
approvals_reviewer = "guardian_subagent"
```

continue to load and map to the Auto-review reviewer behavior. 

New serialization emits: 
```toml
approvals_reviewer = "auto_review" 
```

This PR intentionally does not rename the [features].guardian_approval key or broad internal Guardian symbols. Those are split out for a follow-up PR to keep this migration small and avoid touching large TUI/internal surfaces.

**Verification**
cargo test -p codex-protocol approvals_reviewer_serializes_auto_review_and_accepts_legacy_guardian_subagent
cargo test -p codex-app-server-protocol approvals_reviewer_serializes_auto_review_and_accepts_legacy_guardian_subagent


won-openai:
@codex review

won-openai:
@codex review

won-openai:
@codex review

won-openai:
@codex review

won-openai:
@codex review

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -303,19 +308,54 @@ impl From<CoreAskForApproval> for AskForApproval {
...
+    #[serde(rename = "auto_review", alias = "guardian_subagent")]
     GuardianSubagent,
 }
 
+impl JsonSchema for ApprovalsReviewer {
+    fn schema_name() -> String {
+        "ApprovalsReviewer".to_string()
+    }
+
+    fn json_schema(_generator: &mut SchemaGenerator) -> Schema {
```
won-openai, line 334: This is manual because the normal schema generator does not know about legacy aliases. The code can read both `auto_review` and old `guardian_subagent`, but without this manual schema the generated schema would only list `auto_review`. That could make old configs fail schema checks even though the app can still read them.

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -303,19 +308,54 @@ impl From<CoreAskForApproval> for AskForApproval {
...
+        string_enum_schema_with_description(
+            &["user", "auto_review", "guardian_subagent"],
+            "Configures who approval requests are routed to for review. Examples include sandbox escapes, blocked network access, MCP approval prompts, and ARC escalations. Defaults to `user`. `auto_review` uses a carefully prompted subagent to gather relevant context and apply a risk-based decision framework before approving or denying the request. The legacy value `guardian_subagent` is accepted for compatibility.",
+        )
+    }
+}
+
```
won-openai, line 341: This helper just writes out the schema by hand. It says: this value is a string, these are the allowed string values, and this is the description. We need that because the normal generated schema cannot include the old `guardian_subagent` value for us.

codex-rs/protocol/src/config_types.rs
```diff
@@ -69,22 +75,54 @@ pub enum SandboxMode {
...
+/// ARC escalations. Defaults to `user`. `auto_review` uses a carefully
 /// prompted subagent to gather relevant context and apply a risk-based
 /// decision framework before approving or denying the request.
 pub enum ApprovalsReviewer {
     #[default]
+    #[serde(rename = "user")]
     User,
+    #[serde(rename = "auto_review", alias = "guardian_subagent")]
+    #[strum(serialize = "auto_review")]
     GuardianSubagent,
```
dylan-hurd-oai, line 92: why not change the enum value to `AutoReview`

dylan-hurd-oai (commented):
1 comment about the enum value name itself, but otherwise lgtm

codex-rs/protocol/src/config_types.rs
won-openai: having this on the follow up PR!

dylan-hurd-oai (approved):

Issue #14081, opened 2026-03-09
Queue /fast while running


dat-lequoc:
### What variant of Codex are you using?

CLI

### What feature would you like to see?

<img width="593" height="182" alt="Image" src="https://github.com/user-attachments/assets/7b0d8a85-7a10-487a-8aa1-8d5212675330" />

/fast can be turned ON / OFF for next requests directly without the need to stop.

etraut-openai:
Thanks for the feature request. This will be included in the next release.

PR #18542, opened 2026-04-19
Queue slash and shell prompts in the TUI

etraut-openai:
## Why

Users have asked to queue follow-up slash commands while a task is running, including in #14081, #14588, #14286, and #13779. The previous TUI behavior validated slash commands immediately, so commands that are only meaningful once the current turn is idle could not be queued consistently.

The queue should preserve what the user typed and defer command parsing until the item is actually dispatched. This also gives `/fast`, `/review ...`, `/rename ...`, `/model`, `/permissions`, and similar slash workflows the same FIFO behavior as plain queued prompts.

## What Changed

- Added a queued-input action enum so queued items can be dispatched as plain prompts, slash commands, or user shell commands.
- Changed `Tab` queueing to accept slash-led prompts without validating them up front, then parse and dispatch them when dequeued.
- Added `!` shell-command queueing for `Tab` while a task is running, while preserving existing `Enter` behavior for immediate shell execution.
- Moved queued slash dispatch through shared slash-command parsing so inline commands, unavailable commands, unknown commands, and local config commands report at dequeue time.
- Continued queue draining after local-only actions and after slash menu cancellation or selection when no task is running.
- Preserved slash-popup completion behavior so `/mo<Tab>` completes to `/model ` instead of queueing the prefix.
- Updated pending-input preview snapshots to show queued follow-up inputs.

## Verification

I did a bunch of manual validation (and found and fixed a few bugs along the way).


fcoury-oai (approved):
Also tested and worked perfectly.

Approved.

PR #18573, opened 2026-04-19
feat(tui): show context used in plan implementation prompt

fcoury-oai:
# Summary

When a user finishes planning, the TUI asks whether to implement in the current conversation or start fresh with the approved plan. The clear-context choice is easier to evaluate when the prompt shows how much context has already been used, because the user can see when carrying the full prior conversation is likely to be less useful than preserving only the plan.    

image
    
This PR adds that context signal directly to the clear-context option while keeping the copy compact enough for the Plan-mode selection popup.

# What Changed

- Compute an optional context-usage label when opening the plan implementation prompt.
- Show the label only on `Yes, clear context and implement`, where it informs the cleanup decision.
- Prefer a percentage-used label when context-window information is available, with a compact token-used fallback when only token totals are known.
- Preserve the original option description when usage is unknown or effectively zero.
- Add rustdoc comments around the prompt-copy boundary so future changes keep the context label formatting and selection rendering responsibilities clear.

# Testing

- `cargo test -p codex-tui plan_implementation`

# Notes

The footer continues to show context remaining as ambient status. The implementation prompt intentionally shows context used because the user is choosing whether to clean up the current thread before implementation.

### DIFF
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -2480,15 +2480,39 @@ impl ChatWidget {
 
     fn open_plan_implementation_prompt(&mut self) {
         let default_mask = collaboration_modes::default_mode_mask(self.model_catalog.as_ref());
+        let context_usage_label = self.plan_implementation_context_usage_label();
 
         self.bottom_pane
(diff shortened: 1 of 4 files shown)


etraut-openai (approved):
Looks good.

Issue #5576, opened 2025-10-23
Output width remains truncated after resizing Codex CLI window back to full size


BobbyWang0120:
### What version of Codex is running?

0.47.0

### What subscription do you have?

Pro plan

### Which model were you using?

gpt-5-high

### What platform is your computer?

macOS

### What issue are you seeing?

When the Codex CLI is producing output and I resize the terminal window to a smaller (half) width, the new output lines wrap to fit the reduced width. However, when I resize the window back to the original/full width while Codex is still outputting, the subsequent output remains constrained to the smaller width. It does not expand to fill the full width of the terminal, so the text continues to wrap at the narrower width even though the window has been restored.

### What steps can reproduce the bug?

1. Launch the Codex CLI and start a task that prints multiple lines of output continuously.
2. While the CLI is outputting content, resize the terminal or window to approximately half of its original width.
3. Notice that the new output lines adjust to the smaller width as expected.
4. Without stopping the task, resize the terminal back to its original or a wider width.
5. Observe that the subsequent output still uses the narrower width, even though the window has been resized back to full size.

### What is the expected behavior?

After resizing the window back to its original or larger width, the Codex CLI should adjust the output to the new width. New lines of output should wrap according to the current window size, so they use the full available width rather than remaining restricted to the narrower width used during the temporary resize.

etraut-openai:
Thanks for posting. I think this may be a duplicate of #5259. If so, let's close this one. And feel free to upvote the existing issue.

PaulRBerg:
Still happening in Codex v0.75.0 in Warp terminal v0.2025.12.17.17.17.stable_02

**Initial Width of Terminal**

<img width="200" src="https://github.com/user-attachments/assets/b7de1bce-d533-4229-bca9-b7822ccb6634" />

**Expanded Width of Terminal**

Codex text has not resized according to the terminal

<img width="250" src="https://github.com/user-attachments/assets/8b8df71f-feee-4ba4-b396-ae2ee555d96d" />

mert-kurttutan:
Still happening in 0.77. I think this should be prioritized. It is very annoying when giving codex a chance to try out. Not a good a first impression

felixmin:
+1, this is extremely annoying and regularly makes me write "repeat your response"

PaulRBerg:
Still happening in v0.92

vprasanth:
This is pretty big issue for a tool built to run in the terminal

pthrasher:
[...]

Issue #5538, opened 2025-10-22
Input message disappears while Codex CLI is responding


BobbyWang0120:
### What version of Codex is running?

0.47.0

### What subscription do you have?

Pro plan

### Which model were you using?

gpt-5-high

### What platform is your computer?

macOS

### What issue are you seeing?

When using the Codex CLI, parts of my input message sometimes disappear while the model is generating its response. Lines or segments vanish randomly during processing. In some cases, parts of Codex's output also disappear unexpectedly while it continues to print output. This seems to occur about 20-30% of the time, and I haven't found reliable steps to reproduce it.

### What steps can reproduce the bug?

When using the Codex CLI, after I input a message, parts of my input sometimes disappear while Codex is outputting its response. Lines vanish one by one or segments of the input vanish randomly. Additionally, parts of the output generated by Codex may disappear unexpectedly while the rest of the output is still printing. This occurs randomly about 20-30% of the time and I haven't been able to find consistent steps to reproduce it.

### What is the expected behavior?

The CLI should keep my input and Codex's output visible while it is processing and generating responses. Neither the input message nor any part of the generated output should disappear; they should persist unchanged until the operation is complete.

tshmit:
I also have found parts of both input and output messages disappearing, seemingly at random. This issue has persisted though all recent versions up to and including the current version (0.47.0). I find no discernible pattern but it is common and the inability to reliably refer back in the conversation significantly impairs usability/utility.
(I do run multiple CLI sessions concurrently but find the bug appears even when interacting only with a single session)

ts-klassen:
I see output of AI  disappearing too and I think it's related.

#6599

jeck00119:
It wasn’t fixed, I still have this problem. My version is 0.61.0

distbit0:
same issue

gerrywastaken:
Examples. I have no idea why this isn't more widely reported as it makes Codex CLI unusable and makes the model output seem incoherent due to whole lines vanishing mid paragraph.

<img width="1916" height="989" alt="Image" src="https://github.com/user-attachments/assets/88034b26-8334-45a4-a1e8-8d1a62545b90" />

Jayashanker-Padishala:
I can reproduce this deterministically on my setup (not flaky).

  Environment:
  - OS: Microsoft Windows NT 10.0.26220.0 x64
  - Codex CLI: 0.98.0
  - Terminal: Windows Terminal + PowerShell

  Repro steps:
[...]

PR #18575, opened 2026-04-19
fix(tui): reflow scrollback on terminal resize

fcoury-oai:
Fixes multiple scrollback and terminal resize issues: #5538, #5576, #8352, #12223, #16165, and #15380.

## Why

Codex writes finalized transcript output into terminal scrollback after wrapping it for the current viewport width. A later terminal resize could leave that scrollback shaped for the old width, so wider windows kept narrow output and narrower windows could show stale wrapping artifacts until enough new output replaced the visible area.

This is also the foundation PR for responsive markdown tables. Table rendering needs finalized transcript content to be width-sensitive after insertion, not only while content is first streaming. Markdown table rendering itself stays in #18576.

## Stack

- PR1: resize backlog reflow and interrupt cleanup
- #18576: markdown table support

## What Changed

- Rebuild source-backed transcript history when the terminal width changes. `terminal_resize_reflow` is introduced through the experimental feature system, but is enabled by default for this rollout so we can validate behavior across real terminals.
- Preserve assistant and plan stream source so finalized streaming output can participate in resize reflow after consolidation.
- Debounce resize work, but force a final source-backed reflow when a resize happened during active or unconsolidated streaming output.
- Clear stale pending history lines on resize so old-width wrapped output is not emitted just before rebuilt scrollback.
- Bound replay work with `[tui.terminal_resize_reflow].max_rows`: omitted uses terminal-specific defaults, `0` keeps all rendered rows, and a positive value sets an explicit cap. The cap applies both while initially replaying a resumed transcript into scrollback and when rebuilding scrollback after terminal resize.
- Consolidate interrupted assistant streams before cleanup, then clear pending stream output and active-tail state consistently.
- Move resize reflow and thread event buffering helpers out of `app.rs` into dedicated TUI modules.
- Add focused coverage for resize reflow, feature-gated behavior, streaming source preservation, interrupted output cleanup, unicode-neutral text, terminal-specific row caps, and composer/layout stability.

## Runtime Bounds

Resize reflow keeps only the most recent rendered rows when a row cap is active. The default is `auto`, which maps to the detected terminal's default scrollback size where Codex can identify it: VS Code `1000`, Windows Terminal `9001`, WezTerm `3500`, and Alacritty `10000`. Terminals without a dedicated mapping use the conservative fallback of `1000` rows. Users can override this with `[tui.terminal_resize_reflow] max_rows = N`, or set `max_rows = 0` to disable row limiting.

## Validation

- `just fmt`
- `git diff --check`
- `cargo test --manifest-path codex-rs/Cargo.toml -p codex-tui reflow`
- `cargo test --manifest-path codex-rs/Cargo.toml -p codex-tui transcript_reflow`
- `just fix -p codex-tui`
- PR CI in progress on the squashed branch


codex-rs/features/src/lib.rs
```diff
@@ -657,6 +659,12 @@ pub const FEATURES: &[FeatureSpec] = &[
         stage: Stage::UnderDevelopment,
         default_enabled: false,
     },
+    FeatureSpec {
+        id: Feature::TerminalResizeReflow,
+        key: "terminal_resize_reflow",
+        stage: Stage::Stable,
```
etraut-openai: Should we make this experimental to begin with and then enable it by default once we get some feedback from users? Or are you confident enough to turn it on immediately for everyone?

codex-rs/tui/src/app/resize_reflow.rs
```diff
@@ -0,0 +1,242 @@
+use std::time::Instant;
```
etraut-openai, line 19: I've been making it a habit to add header comments to new modules so it's easier for humans (and agents) to quickly understand the intended scope and use of the module.

etraut-openai (approved):
This is a big change with relatively high regression risk, but I think it's worth it for the long-term benefits it gives us.

I recommend making this an experimental feature initially so we can get feedback from users across many platforms and terminals.

I built the code and did moderate manual testing across various scenarios, but only one platform and terminal. There was a noticeable streaming lag for very long threads upon resize, but that's to be expected, I guess.

codex-rs/tui/src/app/resize_reflow.rs
fcoury-oai: I just did a proper documentation pass.

codex-rs/features/src/lib.rs
fcoury-oai: Done. I changed `terminal_resize_reflow` back to an experimental opt-in feature (`Stage::Experimental`, `default_enabled: false`) and updated the PR body to describe the rollout that way.

codex-rs/config/src/types.rs
```diff
@@ -532,6 +532,20 @@ pub struct ModelAvailabilityNuxConfig {
     pub shown_count: HashMap<String, u32>,
 }
 
+pub const DEFAULT_TERMINAL_RESIZE_REFLOW_FALLBACK_MAX_ROWS: usize = 1_000;
```
etraut-openai, line 536: Nit: a comment would be good here to describe this constant.

codex-rs/config/src/types.rs
```diff
@@ -532,6 +532,20 @@ pub struct ModelAvailabilityNuxConfig {
...
+
+/// Experimental terminal resize-reflow tuning knobs.
```
etraut-openai: Is the intent to keep these knobs around long term? If so, probably remove the "experimental" part here.

codex-rs/tui/src/app.rs
```diff
@@ -481,12 +483,38 @@ fn list_skills_response_to_core(response: SkillsListResponse) -> ListSkillsRespo
     }
 }
 
+fn trailing_run_start<T: 'static>(transcript_cells: &[Arc<dyn HistoryCell>]) -> usize {
```
etraut-openai: Could we move this out of app.rs, perhaps into resize_reflow.rs? I recently cleaned up app.rs, and I'd like to keep it clean.

codex-rs/tui/tests/suite/resize_reflow.rs
```diff
@@ -0,0 +1,613 @@
+use std::path::Path;
```
etraut-openai, line 1: Nit: The name of this file seems odd. A "smoke test" is usually one that's conducted right before a build is released to make sure it minimally runs. This looks more like a standard test module, so maybe rename to "resize_reflow_test.rs"?

codex-rs/core/src/config/mod.rs
```diff
@@ -643,6 +646,22 @@ impl Default for MultiAgentV2Config {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum TerminalResizeReflowMaxRows {
```
etraut-openai, line 657: Do you plan to add more knobs under the "tui.terminal_resize_reflow" table? It seems odd to have just one knob under this table. If it's for future expansion, that's fine. If you don't have any plans to add more in the future, then let's collapse it to a single key under the "tui" table.

etraut-openai (approved):
This looks really good! I left a few comments for you to consider.

It looks like the optimization you apply for resizing (where you walk backward through the transcript to find the history cells that are needed to fill the configured line count) is not used for replaying the history when resuming a thread. I wonder if we should use that same optimization in that case? This isn't something that needs to be part of this PR — just something to consider for the future. 

I manually tested a bunch of resize scenarios using Terminal.app, iTerm, Kitty, the VS Code integrated terminal, and the Codex app integrated terminal. I tried a bunch of overlays (menus, status box, etc.), and I didn't find any issues. I didn't test on platforms other than macOS. I also didn't test the config mechanisms.

codex-rs/config/src/types.rs
fcoury-oai: Done. I added a doc comment explaining that this is the fallback row cap when Codex cannot identify a terminal-specific scrollback size.

Replied by codex.

codex-rs/config/src/types.rs
fcoury-oai: Done. I removed the “Experimental” wording from the config docs/schema since the tuning knob is intended to remain as config surface even while the feature itself is still experimental.

Replied by codex.

codex-rs/tui/src/app.rs
fcoury-oai: Done. I moved `trailing_run_start` into `app/resize_reflow.rs` so `app.rs` stays focused on top-level orchestration.

Replied by codex.

codex-rs/tui/tests/suite/resize_reflow.rs
fcoury-oai: Done. I renamed the module to `resize_reflow.rs`; that keeps it aligned with the surrounding `tests/suite` module naming while avoiding the “smoke test” label.

Replied by codex.

codex-rs/core/src/config/mod.rs
fcoury-oai: Done. I collapsed this to a single `[tui]` key: `terminal_resize_reflow_max_rows`. That keeps the current behavior but avoids introducing a one-field table.

Replied by codex.

PR #18576, opened 2026-04-19
feat(tui): render markdown tables responsively

fcoury-oai:
## Summary

Add responsive markdown pipe-table rendering for the TUI, including markdown-fence unwrapping for agent output and terminal-width-aware boxed table layout.

Streaming agent output now holds mutable table tails until finalization, so column widths can settle without committing malformed partial rows to scrollback. The live-tail path is gated behind `stream_table_live_tail_reflow` for speculative uncommitted table preview behavior.

## Validation

- `just fmt`
- `just write-config-schema`
- `cargo test -p codex-tui`
- `just fix -p codex-tui`


vincent-globalorange:
Thanks for pushing #18575/#18576. We reached the “try `glow` for tables” stage over here, so these fixes will surely be appreciated.

Generated by GPT-5.5. Approved by user.

PR #18580, opened 2026-04-19
Avoid redundant memory enable notice

### LINKED ISSUE #18554: /experimental shows redundant "Memories will be enabled in the next session" warning when Memories is already enabled

`codex-cli 0.122.0-alpha.10`

`ChatGPT Pro`

`gpt-5.4 xhigh`

`Darwin 25.2.0 arm64 arm`

`Ghostty 1.3.1`

When I open `/experimental`, the `Memories` checkbox is already checked. If I press Enter immediately without changing anything, Codex adds the warning `Memories will be enabled in the next session.`

That warning looks redundant in this case. Memories were already enabled for the current session, and I did not change the checkbox state.

1. Start a Codex CLI session where Memories is already enabled.
2. Run `/experimental`.
3. Confirm that `[x] Memories` is already checked.
4. Press Enter without toggling any options.
5. Observe the warning `Memories will be enabled in the next session.`

No warning should appear on a no-op save.

The warning should only appear if Memories actually transitions from disabled to enabled and a new session is required for that change to take effect.

### Additional information

This looks like the `/experimental` save path may be submitting the full feature snapshot, and the notice may be keyed off `Feature::MemoryTool = true` instead of an actual disabled-to-enabled transition.

The code path that seems most relevant is:

- `codex-rs/tui/src/bottom_pane/experimental_features_view.rs`
- `codex-rs/tui/src/app.rs`

etraut-openai:
## Summary

Fixes #18554.

The `/experimental` menu can submit the full experimental feature state even when the user presses Enter without toggling anything. Previously, Codex showed `Memories will be enabled in the next session.` whenever the submitted updates included `Feature::MemoryTool = true`, so sessions where Memories were already enabled could show a redundant warning on a no-op save.

This change records whether `Feature::MemoryTool` was enabled before applying feature updates and only emits the next-session notice when Memories actually transitions from disabled to enabled.


fcoury-oai (approved):
Replicated and confirmed the fix.

PR #18581, opened 2026-04-19
[1/4] Add executor HTTP request protocol

aibrahim-oai:
### Why
Remote streamable HTTP MCP needs a transport-shaped executor primitive before the MCP client can move network I/O to the executor. This layer keeps the executor unaware of MCP and gives later PRs an ordered streaming surface for response bodies.

### What
- Add typed `http/request` and `http/request/bodyDelta` protocol payloads.
- Add executor client helpers for buffered and streamed HTTP responses.
- Route body-delta notifications to request-scoped streams with sequence validation and cleanup when a stream finishes or is dropped.
- Document the new protocol constants, transport structs, public client methods, body-stream lifecycle, and request-scoped routing helpers.
- Add in-memory JSON-RPC client coverage for streamed HTTP response-body notifications, with comments spelling out what the test proves and each setup/exercise/assert phase.

### Stack
1. #18581 protocol
2. #18582 runner
3. #18583 RMCP client
4. #18584 manager wiring and local/remote coverage

### Verification
- `just fmt`
- `cargo check -p codex-exec-server -p codex-rmcp-client --tests`
- `cargo check -p codex-core --test all` compile-only
- `git diff --check`
- Online full CI is running from the `full-ci` branch, including the remote Rust test job.

Co-authored-by: Codex <noreply@openai.com>

PR #18582, opened 2026-04-19
[2/4] Implement executor HTTP request runner

aibrahim-oai:
### Why
Remote streamable HTTP MCP needs the executor to perform ordinary HTTP requests on the executor side. This keeps network placement aligned with `experimental_environment = "remote"` without adding MCP-specific executor APIs.

### What
- Add an executor-side `http/request` runner backed by `reqwest`.
- Validate request method and URL scheme, preserving the transport boundary at plain HTTP.
- Return buffered responses for ordinary calls and emit ordered `http/request/bodyDelta` notifications for streaming responses.
- Register the request handler in the exec-server router.
- Document the runner entrypoint, conversion helpers, body-stream bridge, notification sender, timeout behavior, and new integration-test helpers.
- Add exec-server integration tests with the existing websocket harness and a local TCP HTTP peer for buffered and streamed responses, with comments spelling out what each test proves and its setup/exercise/assert phases.

### Stack
1. #18581 protocol
2. #18582 runner
3. #18583 RMCP client
4. #18584 manager wiring and local/remote coverage

### Verification
- `just fmt`
- `cargo check -p codex-exec-server -p codex-rmcp-client --tests`
- `cargo check -p codex-core --test all` compile-only
- `git diff --check`
- Online full CI is running from the `full-ci` branch, including the remote Rust test job.

Co-authored-by: Codex <noreply@openai.com>


aibrahim-oai:
This change is part of the following stack:

- #18581
    - #18582 ◀
        - #18583
            - #18584

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

codex-rs/exec-server/src/server/handler.rs
```diff
@@ -147,6 +178,61 @@ impl ExecServerHandler {
         session.process().terminate(params).await
     }
 
+    pub(crate) async fn http_request(
+        &self,
+        params: HttpRequestParams,
+    ) -> Result<HttpRequestResponse, JSONRPCErrorError> {
+        self.require_initialized_for("http")?;
+        let request_id = if params.stream_response {
```
starr-openai: lets just make request_id required always?

codex-rs/exec-server/src/server/handler.rs
```diff
@@ -147,6 +178,61 @@ impl ExecServerHandler {
...
+            None
+        };
+        if let Some(request_id) = request_id.as_deref() {
+            self.reserve_http_body_stream(request_id).await?;
+        }
+        match run_http_request(params).await {
+            Ok((response, pending_stream)) => {
+                if let Some(pending_stream) = pending_stream {
+                    let mut http_body_streams = self.http_body_streams.lock().await;
+                    http_body_streams.pending.push_back(pending_stream);
```
starr-openai: can we just drive the body stream with a tokio task here? why do we enqueue the stream and drive them all separately?

codex-rs/exec-server/src/http_request.rs
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    let response = request
+        .send()
+        .await
+        .map_err(|err| internal_error(format!("http/request failed: {err}")))?;
+    let status = response.status().as_u16();
+    let headers = response_headers(response.headers());
+
+    if let Some(request_id) = request_id {
```
starr-openai: same as above - I think making request_id required simplifies the code a bit and is reasonable to always have

codex-rs/exec-server/src/server/handler.rs
starr-openai: if it's to ensure we clean-up successfully, maybe use a JoinSet + CancellationToken

codex-rs/exec-server/src/server/handler.rs
```diff
@@ -40,6 +50,7 @@ pub(crate) struct ExecServerHandler {
     session_registry: Arc<SessionRegistry>,
     notifications: RpcNotificationSender,
     session: StdMutex<Option<SessionHandle>>,
+    body_streams: Mutex<HashMap<String, Option<JoinHandle<()>>>>,
```
starr-openai: from quick read of the rust docs, if we can use a `tokio_util::task::TaskTracker` and don't need the key here, that might generalize so we can just use this for any background coros.

so e.g. `background_tasks: tokio_util::task::TaskTracker<...>`

then you can add the background body tailing tasks there and don't need the explicit cleanup in the request handler.

if it's not easy to adapt/use the current PR is fine tho

starr-openai (approved):

PR #18591, opened 2026-04-19
Surface parent thread status in side conversations

etraut-openai:
## Summary

Side conversations can hide important state changes from the parent conversation while the user is focused on the side thread. In particular, the parent may finish, fail, need user input, or require an approval while the side conversation remains visible. Users need a lightweight signal for those states, but parent approval overlays should not interrupt the side conversation itself.

This change adds parent-conversation status to the side conversation context label and defers parent interactive overlays while side mode is active. When the user exits side mode, pending parent approvals and input requests are restored in the main thread. The pending approval footer avoids duplicating the same parent approval status, and replayed notice cells are filtered when restoring a pending interactive request so tips or warnings do not crowd out the approval prompt.

The change is contained to the TUI side-conversation and thread replay paths.

Example 1: Approval pending
Screenshot 2026-04-19 at 12 56 07 PM

Example 2: Turn complete
Screenshot 2026-04-19 at 12 56 27 PM

### DIFF
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -176,6 +176,7 @@ mod app_server_adapter;
 pub(crate) mod app_server_requests;
 mod loaded_threads;
 mod pending_interactive_replay;
+mod replay_filter;
 mod side;
 
 use self::agent_navigation::AgentNavigationDirection;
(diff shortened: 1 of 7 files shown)


fcoury-oai (approved):
Tested and works as expected. Code looks good.

Approved 👍

PR #18597, opened 2026-04-19
Update realtime handoff transcript handling

guinness-oai:
## Summary

This PR aims to improve integration between the realtime model and the codex agent by sharing more context with each other. In particular, we now share full realtime conversation transcript deltas in addition to the delegation message. 

realtime_conversation.rs now turns a handoff into:
```
<realtime_delegation>
  <input>...</input>
  <transcript_delta>...</transcript_delta>
</realtime_delegation>
```

## Implementation notes

The transcript is accumulated in the realtime websocket layer as parsed realtime events arrive. When a background-agent handoff is requested, the current transcript snapshot is copied onto the handoff event and then serialized by `realtime_conversation.rs` into the hidden realtime delegation envelope that Codex receives as user-turn context.

For Realtime V2, the session now explicitly enables input audio transcription, and the parser handles the relevant input/output transcript completion events so the snapshot includes both user speech and realtime model responses. The delegation `<input>` remains the actual handoff request, while `<transcript_delta>` carries the surrounding conversation history for context.

Reviewers should note that the transcript payload is intended for Codex context sharing, not UI rendering. The realtime delegation envelope should stay hidden from the user-facing transcript surface, while still being included in the background-agent turn so Codex can answer with the same conversational context the realtime model had.

### DIFF
--- a/codex-rs/app-server/tests/suite/v2/realtime_conversation.rs
+++ b/codex-rs/app-server/tests/suite/v2/realtime_conversation.rs
@@ -687,7 +687,6 @@ async fn realtime_conversation_streams_v2_notifications() -> Result<()> {
     assert_eq!(
         handoff_item_added.item["active_transcript"],
         json!([
-            {"role": "assistant", "text": "hi"},
             {"role": "user", "text": "delegate now"},
             {"role": "assistant", "text": "working on it"}
         ])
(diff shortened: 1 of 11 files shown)


guinness-oai:
@codex

aibrahim-oai (approved):

Issue #18505, opened 2026-04-18
tui onboarding: warn when trusting git subdirectories


joycebeatriz:
## Summary

When a user runs Codex from a subdirectory within a Git repository and chooses to trust the directory, it's unclear that the trust decision applies to the repository root, not just the current directory.

## Proposal

Add a yellow warning message in the TUI onboarding screen that clarifies when the user is in a Git subdirectory, showing which directory will actually be trusted.

## Implementation

The fix has been implemented and tested at: https://github.com/joycebeatriz/codex/pull/1

- Added conditional warning message when `cwd != trust_target`
- Updated snapshot tests
- Included Contributor Covenant v3.0 adoption

itsmejay80:
Valid enhancement the codebase already has a `// TODO: add git warning.` comment at exactly the right spot (`onboarding_screen.rs:142`), and `TrustDirectoryWidget` already carries both `cwd` and `trust_target` fields, so the scaffolding is all there.

## Proposed Implementation

**1. `codex-rs/tui/src/onboarding/trust_directory.rs` — add warning block in `render_ref`**

After line 51 (the \"You are in ...\" line), insert a conditional yellow warning when `cwd != trust_target`:

```rust
// after the "You are in..." line
column.push("");

if self.cwd != self.trust_target {
    column.push(
        Paragraph::new(format!(
            "⚠  You are in a Git subdirectory. Trusting will apply to the repository root: {}",
            self.trust_target.display()
        ))
        .yellow()
        .wrap(Wrap { trim: true })
        .inset(Insets::tlbr(0, 2, 0, 0)),
    );
    column.push("");
}
```

**2. `codex-rs/tui/src/onboarding/onboarding_screen.rs:142` remove the TODO**

The TODO comment can be dropped once the widget handles the warning itself (it already receives `trust_target`).

**3. Add snapshot test**

```rust
#[test]
fn renders_snapshot_for_git_subdirectory() {
    let codex_home = TempDir::new().expect("temp home");
    let widget = TrustDirectoryWidget {
        codex_home: codex_home.path().to_path_buf(),
        cwd: PathBuf::from("/workspace/project/src"),        // subdirectory
        trust_target: PathBuf::from("/workspace/project"),   // git root
        show_windows_create_sandbox_hint: false,
        should_quit: false,
        selection: None,
        highlighted: TrustDirectorySelection::Trust,
        error: None,
    };

    let mut terminal =
[...]

PR #18602, opened 2026-04-20
Warn when trusting Git subdirectories

etraut-openai:
Addresses #18505

## Summary

When Codex is launched from a subdirectory of a Git repository, the onboarding trust prompt says it is trusting the current directory even though the persisted trust target is the repository root. That can make the scope of the trust decision unclear.

This updates the TUI trust prompt to show a yellow note only when the current directory differs from the resolved trust target, explaining that trust applies to the repository root and displaying that root. It also removes the stale onboarding TODO now that the warning is implemented.


fcoury-oai:
I found a potential issue. Even if I am not on a .git subdirectory (in this case I am in `packages/app`) it still says I am in a Git subdirectory. Since this happens for any subdirectory, maybe we need to change the message?

image

etraut-openai:
[codex] Addressed in the latest push: the note now says "You’re in a subdirectory of a Git project" instead of "Git subdirectory", so it applies to regular nested directories like `packages/app` without implying the directory itself contains `.git`.

fcoury-oai (approved):
Looks good 👍

PR #18610, opened 2026-04-20
Add verbose diagnostics for /mcp

### LINKED ISSUE #18539: /mcp output regressed starting in codex-cli 0.116.0: missing Status/Resources/Resource templates
## Summary
Starting in `codex-cli 0.116.0`, the `/mcp` slash command output regressed compared with `0.115.0`.

The MCP servers themselves can still be enabled/working, but `/mcp` no longer shows key diagnostic fields, which makes working MCP setups look broken.

## Environment
- OS: Windows
- Last known good: `0.115.0`
- First bad: `0.116.0`
- Also reproducible in `0.121.0`
- Same `~/.codex/config.toml`
- Same MCP server setup
- Reproduced in terminal, not only VS Code

## Expected behavior
`/mcp` should keep the diagnostic output format from `0.115.0`, including:
- `Status: enabled`
- `Resources`
- `Resource templates`

## Actual behavior starting in 0.116.0
`/mcp` switches to a reduced/different output format:
- `Status: enabled` disappears
- `Resources` disappears
- `Resource templates` disappears
- tool naming changes in some cases
- output can make working MCP servers look broken

### 0.115.0
`/mcp` shows full diagnostic output like:

```text
• context7
  • Status: enabled
  • Auth: Unsupported
  • Command: npx -y @upstash/context7-mcp@latest
  • Tools: query_docs, resolve_library_id
  • Resources: (none)
  • Resource templates: (none)
```

### 0.116.0+
`/mcp` no longer shows those fields and instead shows a reduced output format.

At the same time, MCP backends may still be functioning, so this appears to be a `/mcp` rendering/output regression rather than necessarily an MCP backend failure.

## Impact
- users think MCP stopped working after upgrade
- `/mcp` becomes much less useful for diagnostics
- enabled state is no longer visually confirmable from the slash command

## Request
Please restore the previous `/mcp` diagnostic output, or add a verbose mode that includes:
- status
- resources
- resource templates
- exact MCP state

etraut-openai:
Fixes #18539.

## Summary
The recent `/mcp` performance work kept the default command fast by avoiding resource and resource-template inventory probes, but it also removed useful diagnostics for users trying to confirm MCP server state.

This keeps bare `/mcp` on the fast tools/auth path and adds `/mcp verbose` for the slower diagnostic view. Verbose mode requests full MCP server status from the app-server and restores status, resources, and resource templates in the TUI output.

## Testing
In addition to running automation, I manually tested the feature to confirm that it works.

### DIFF
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -2054,14 +2054,18 @@ impl App {
     /// tradeoff because the effect is limited to stale inventory output in history,
     /// while request-token invalidation would add cross-cutting async state for a
     /// low-severity path.
-    fn fetch_mcp_inventory(&mut self, app_server: &AppServerSession) {
+    fn fetch_mcp_inventory(
+        &mut self,
+        app_server: &AppServerSession,
+        detail: McpServerStatusDetail,
+    ) {
         let request_handle = app_server.request_handle();
         let app_event_tx = self.app_event_tx.clone();
         tokio::spawn(async move {
-            let result = fetch_all_mcp_server_statuses(request_handle)
+            let result = fetch_all_mcp_server_statuses(request_handle, detail)
                 .await
                 .map_err(|err| err.to_string());
-            app_event_tx.send(AppEvent::McpInventoryLoaded { result });
+            app_event_tx.send(AppEvent::McpInventoryLoaded { result, detail });
         });
     }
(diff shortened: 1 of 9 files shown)


fcoury-oai (approved):
Works and code looks good. Approved!

PR #18626, opened 2026-04-20
Respect explicit untrusted project config

### LINKED ISSUE #18475: -c config override for projects.<path>.trust_level is not ephemeral and can   persist the opposite value to ~/.codex/config.toml

codex-cli 0.121.0

N/A

_No response_

Darwin 24.6.0 arm64 arm

Apple Terminal 2.14 (455.1)

Passing `-c` to override project trust at startup mutates `~/.codex/config.toml`, even though `--help` describes `-c` as a runtime override of values loaded from that file. In this case, codex even persisted the opposite value from the cli argument.

```
$ cat ~/.codex/config.toml 
# ...

[tui]
show_tooltips = false
theme = "ansi"

[features]
unified_exec = true
```

```
$ codex -c "projects={\"$PWD\"={trust_level=\"untrusted\"}}"
```

codex opens, no prompt is shown.

```
$ cat ~/.codex/config.toml
# ...

[tui]
show_tooltips = false
theme = "ansi"

[features]
unified_exec = true

[projects."/Users/a/b/c"]
trust_level = "trusted"
```

`-c` should not add anything to the config.toml (?) and certainly not the opposite of what it received via cli args.

  - `-c `should not persist changes into `~/.codex/config.toml`.
  - If the override is accepted, the effective value for that session should be `untrusted`, not `trusted`.

### Additional information

_No response_

etraut-openai:
## Why

Fixes #18475. A `-c` override such as `projects.<cwd>.trust_level = "untrusted"` is meant to be a runtime config override, but app-server thread startup treated any non-trusted project as eligible for automatic trust persistence when a permissive sandbox/cwd was requested. That meant an explicit `untrusted` session override could still cause `config.toml` to be updated with `trusted`.

## What changed

The app-server auto-trust path now runs only when the active project trust level is unknown. Explicit `trusted` and explicit `untrusted` values are both respected, regardless of whether they came from persisted config or session flags.

A focused `thread/start` test now covers the explicit `untrusted` case with a permissive sandbox request.

## Verification

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -2597,7 +2597,7 @@ impl CodexMessageProcessor {
             requested_permissions_trust_project(&typesafe_overrides, config.cwd.as_path());
 
         if requested_cwd.is_some()
-            && !config.active_project.is_trusted()
+            && config.active_project.trust_level.is_none()
             && (requested_permissions_trust_project
                 || matches!(
                     config.permissions.sandbox_policy.get(),
--- a/codex-rs/app-server/tests/suite/v2/thread_start.rs
+++ b/codex-rs/app-server/tests/suite/v2/thread_start.rs
@@ -803,6 +803,44 @@ async fn thread_start_with_read_only_sandbox_does_not_persist_project_trust() ->
 }
 
 #[tokio::test]
+async fn thread_start_preserves_untrusted_project_trust() -> Result<()> {
+    let server = create_mock_responses_server_repeating_assistant("Done").await;
+
+    let codex_home = TempDir::new()?;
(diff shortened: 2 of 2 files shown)


bolinfest (approved):
Is there an easy way to cover this with a test so we don't regress it?

etraut-openai:
[codex] Added a focused thread_start_preserves_untrusted_project_trust test covering explicit untrusted project trust with a WorkspaceWrite thread start. cargo test -p codex-app-server passed locally.

vieira:
@etraut-openai @bolinfest Is there anything I (OP of #18475) can do to help move this forward?

etraut-openai:
@vieira, thanks for the ping. I've been really busy the last few days, and this dropped off the radar. Working on it now.

Issue #18113, opened 2026-04-16
Codex CLI always uses sandbox


bogosmer:
### What version of Codex CLI is running?

0.121

### What subscription do you have?

Plus

### Which model were you using?

gpt-5.4

### What platform is your computer?

macOS Darwin 25.3.0 arm64 arm

### What terminal emulator and version are you using (if applicable)?

Terminal.app

### What issue are you seeing?

`codex exec` ignores explicit sandbox override flags.

I run:
env -i HOME="$HOME" USER="$USER" SHELL=/bin/zsh PATH="/opt/homebrew/bin:/usr/bin:/bin" \
  /opt/homebrew/bin/codex -p unsandboxed --sandbox danger-full-access -m gpt-5.4 exec "echo ok"

The command returns `ok`, but session logs show the resolved sandbox policy is still read-only.

### What steps can reproduce the bug?

1. Run the command above in a normal terminal.
2. Inspect the session log generated for that run.
3. In my case:
   ~/.codex/sessions/2026/04/16/rollout-2026-04-16T11-01-34-019d9586-36be-7d22-90cd-94cb0289e8c3.jsonl
4. `turn_context` shows:
   `"sandbox_policy":{"type":"read-only"}`

### What is the expected behavior?

With `-p unsandboxed --sandbox danger-full-access`, resolved sandbox policy should be danger-full-access (or at minimum not read-only).

### Additional information

This blocks workflows that require git writes/xcodebuild/network during automated loops, because the process behaves as sandboxed even when launched with unsandboxed/danger-full-access settings.

Project is marked as trusted and I have this in config.toml for unsandboxed profile: ```
[profiles.unsandboxed]
sandbox_mode = "danger-full-access"
approval_policy = "never"
```

bogosmer:
> Potential duplicates detected. Please review them and close your issue if it is a duplicate.
> 
> * [codex exec with danger-full-access still hits EPERM/loopback restrictions while interactive Codex in same shell succeeds #15696](https://github.com/openai/codex/issues/15696)
> 
> _Powered by [Codex Action](https://github.com/openai/codex-action)_

Not a duplicate although the underlying causes might be the same.

etraut-openai:
Thanks for the bug report. This will be addressed in the next release.

viyatb-oai:
fixed in `0.123.0`

PR #18630, opened 2026-04-20
Fix exec inheritance of root shared flags

etraut-openai:
Addresses #18113

Problem: Shared flags provided before the exec subcommand were parsed by the root CLI but not inherited by the exec CLI, so exec sessions could run with stale or default sandbox and model configuration.

Solution: Move shared TUI and exec flags into a common option block and merge root selections into exec before dispatch, while preserving exec's global subcommand flag behavior.

### DIFF
--- a/codex-rs/cli/src/main.rs
+++ b/codex-rs/cli/src/main.rs
@@ -717,6 +717,9 @@ async fn cli_main(arg0_paths: Arg0DispatchPaths) -> anyhow::Result<()> {
                 root_remote_auth_token_env.as_deref(),
                 "exec",
             )?;
+            exec_cli
+                .shared
+                .inherit_exec_root_options(&interactive.shared);
             prepend_config_flags(
                 &mut exec_cli.config_overrides,
                 root_config_overrides.clone(),
@@ -1242,6 +1245,7 @@ async fn run_debug_prompt_input_command(
     interactive: TuiCli,
     arg0_paths: Arg0DispatchPaths,
 ) -> anyhow::Result<()> {
+    let shared = interactive.shared.into_inner();
     let mut cli_kv_overrides = root_config_overrides
         .parse_overrides()
         .map_err(anyhow::Error::msg)?;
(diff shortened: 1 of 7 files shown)


codex-rs/cli/src/main.rs
fcoury-oai, line 736: Codex identified that we could also use interactive.shared options for this review branch. I think it might be a good idea, but not a blocker for this PR. Maybe we could have a follow up one for `codex review`?

codex-rs/utils/cli/src/shared_options.rs
```diff
@@ -0,0 +1,174 @@
...
+    #[arg(long = "sandbox", short = 's')]
+    pub sandbox_mode: Option<SandboxModeCliArg>,
+
+    /// Convenience alias for low-friction sandboxed automatic execution.
+    #[arg(long = "full-auto", default_value_t = false)]
+    pub full_auto: bool,
+
+    /// Skip all confirmation prompts and execute commands without sandboxing.
+    /// EXTREMELY DANGEROUS. Intended solely for running in environments that are externally sandboxed.
+    #[arg(
```
fcoury-oai, line 47: Another note from a Codex finding: 

After this change, the TUI no longer rejects conflicting permission flags. For example, this now parses:

```
codex --dangerously-bypass-approvals-and-sandbox --ask-for-approval on-request
```

Those two flags ask for opposite behaviors: one says “never ask and bypass the sandbox,” while the other says “ask before running commands.” The TUI then silently chooses the dangerous bypass behavior and ignores the approval setting.

The suggested fix is to make the TUI reject that conflicting flag combination again.

fcoury-oai (commented):
The first issue about replicating the shared options for Review is not a blocked and can be either addressed as a follow-up PR or in this one if you prefer.

I think we should fix the other issue of conflicting permissions.

There are also some merge conflicts now on this PR. Once you take another pass and decide how to address those issues, I can do a final pass.

codex-rs/utils/cli/src/shared_options.rs
etraut-openai: Fixed in the latest push by adding a TUI-specific shared-options wrapper that restores the conflict with `approval_policy`, plus regression coverage in `dangerous_bypass_conflicts_with_approval_policy`.

codex-rs/cli/src/main.rs
etraut-openai: Agreed this is good follow-up scope. I left `codex review` out of this PR to keep the fix focused on exec root flag inheritance.

fcoury-oai (approved):
Only did a sanity check locally with `exec` and the good looks good. Approved.

PR #18631, opened 2026-04-20
Remove simple TUI legacy_core reexports

etraut-openai:
## Problem
The TUI still imported path utilities and config-loader symbols through app-server-client's legacy_core facade even though those APIs already exist in utility/config crates. This is part of our ongoing effort to whittle away at these old dependencies.

## Solution
Rewire imports to avoid the TUI directly importing from the core crate and instead import from common lower-level crates. This PR doesn't include any functional changes; it's just a simple rewiring.


fcoury-oai (approved):
Looks good! 👍

PR #18632, opened 2026-04-20
Use app server metadata for fork parent titles

etraut-openai:
## Problem
The TUI resolved fork parent titles from local CODEX_HOME metadata, which could show missing or stale titles when app-server metadata is authoritative.

This is a lingering bug left over from the migration of the TUI to the app-server interface. I found it when I asked Codex to review all places where the TUI code was still directly accessing the local CODEX_HOME.

## Solution
Route fork parent title metadata through the app-server session state and render only that supplied title, with focused snapshot coverage for stale local metadata.

## Testing
I manually tested by renaming a thread then forking it and confirming that the "forked from" message indicated the parent thread's name.


fcoury-oai (approved):
Tested fork with a Codex App-generated thread and it worked flawlessly.

Code looks good! 👍

etraut-openai:
[codex] Addressed in b27f89daa0: handle_thread_session_quiet now calls on_session_configured_with_display_and_fork_parent_title and passes through the fork parent title.

PR #18633, opened 2026-04-20
Use app server thread names in TUI picker

etraut-openai:
## Problem

The TUI resume/fork picker was backfilling thread names from local rollout indexes. This was left over from before the TUI was moved to the app server. It should be using app-server APIs because the TUI might be connected to a remote connection.

This bug wasn't (yet) reported by a user. I found it by asking Codex to review places in the TUI code where it was still directly accessing the CODEX_HOME directory rather than going through app-server APIs.

## Solution

The resume picker and session lookups should use app-server thread APIs only. Remove legacy rollout name/list backfills, and avoid local name reads in fork history.

## Testing

I manually tested `codex resume` and `codex resume --all` to look for functional or performance regressions in the resume picker.

### DIFF
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -4597,13 +4597,7 @@ impl App {
                 tui.frame_requester().schedule_frame();
             }
             AppEvent::ResumeSessionByIdOrName(id_or_name) => {
-                match crate::lookup_session_target_with_app_server(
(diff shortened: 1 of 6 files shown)


fcoury-oai (approved):
Also tested and made sure the threads and thread names remained the same after this change, specially the ones created by the Codex App.

PR #18657, opened 2026-04-20
fix: auth.json leak in tests

jif-oai:
Before this some tests were leaking an auth.json file into `codex-rs/core`. This just fixes it


jif-oai:
@codex review

adrian-openai (approved):
Thanks for catching this!

PR #18659, opened 2026-04-20
chore: drop review prompt from TUI UX

jif-oai:
Due to the app-server rebase of the TUI, the review prompt was leaked into the transcript on the TUI
This is not a security issue but it was bad UX. This PR fixes this


jif-oai:
@codex review

fcoury-oai (approved):
I was able to replicate the issue and the fix. Code looks good!

PR #18661, opened 2026-04-20
feat: log client use min log level

jif-oai:
In the log client, use the log level filter as a minimum severity instead of exact match


jif-oai:
@codex review

PR #18662, opened 2026-04-20
feat: add metric to track the number of turns with memory usage

jif-oai:
Add a metric `codex.turn.memory` to know if a turn used memories or not. This is not part of the other turn metrics as a label to limit cardinality


jif-oai:
@codex review

PR #18713, opened 2026-04-20
protocol: preserve glob scan depth in permission profiles

bolinfest:
## Why

#18274 made `PermissionProfile` the canonical file-system permissions shape, but the round-trip from `FileSystemSandboxPolicy` to `PermissionProfile` still dropped one piece of policy metadata: `glob_scan_max_depth`.

That field is security-relevant for deny-read globs such as `**/*.env`. On Linux, bubblewrap sandbox construction uses it to bound unreadable glob expansion. If a profile copied from active runtime permissions loses this value and is submitted back as an override, the resulting `FileSystemSandboxPolicy` can behave differently even though the visible permission entries look equivalent.

## What changed

- Add `glob_scan_max_depth` to protocol `FileSystemPermissions` and preserve it when converting to/from `FileSystemSandboxPolicy`.
- Keep legacy `read`/`write` JSON for simple path-only permissions, but force canonical JSON when glob scan depth is present so the metadata is not silently dropped.
- Carry `globScanMaxDepth` through app-server `AdditionalFileSystemPermissions`, generated JSON/TypeScript schemas, and app-server/TUI conversion call sites.
- Preserve the metadata through sandboxing permission normalization, merging, and intersection.
- Carry the merged scan depth into the effective `FileSystemSandboxPolicy` used for command execution, so bounded deny-read globs reach Linux bubblewrap materialization.

## Verification

- `cargo test -p codex-sandboxing glob_scan -- --nocapture`
- `cargo test -p codex-sandboxing policy_transforms -- --nocapture`
- `just fix -p codex-sandboxing`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/18713).
* #18288
* #18287
* #18286
* #18285
* #18284
* #18283
* #18282
* #18281
* #18280
* #18279
* #18278
* #18277
* #18276
* #18275
* __->__ #18713

### DIFF
--- a/codex-rs/sandboxing/src/policy_transforms.rs
+++ b/codex-rs/sandboxing/src/policy_transforms.rs
@@ -13,6 +13,7 @@ use codex_protocol::protocol::SandboxPolicy;
 use codex_utils_absolute_path::AbsolutePathBuf;
 use codex_utils_absolute_path::canonicalize_preserving_symlinks;
 use std::collections::HashSet;
+use std::num::NonZeroUsize;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct EffectiveSandboxPermissions {
@@ -45,6 +46,7 @@ pub fn normalize_additional_permissions(
     let file_system = match additional_permissions.file_system {
         Some(file_system) => {
             let mut entries = Vec::with_capacity(file_system.entries.len());
+            let glob_scan_max_depth = file_system.glob_scan_max_depth;
             for entry in file_system.entries {
                 if matches!(&entry.path, FileSystemPath::GlobPattern { .. })
                     && entry.access != FileSystemAccessMode::None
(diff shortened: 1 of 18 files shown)


codex-rs/sandboxing/src/policy_transforms.rs
```diff
@@ -114,6 +118,9 @@ pub fn merge_permission_profiles(
             let file_system = match (base.file_system.as_ref(), permissions.file_system.as_ref()) {
                 (Some(base), Some(permissions)) => Some(FileSystemPermissions {
                     entries: merge_permission_entries(&base.entries, &permissions.entries),
+                    glob_scan_max_depth: base
+                        .glob_scan_max_depth
+                        .max(permissions.glob_scan_max_depth),
```
viyatb-oai: [P1] Carry glob scan depth into the effective sandbox policy

This keeps `glob_scan_max_depth` on the `PermissionProfile`, but the command/apply_patch path later materializes that profile with `effective_file_system_sandbox_policy`, and `merge_file_system_policy_with_additional_permissions` only appends entries to the cloned runtime policy. If the base policy has no depth and a reviewer grants root write with a bounded deny glob, Linux bwrap still sees the unbounded base depth when it expands the deny glob. The depth needs to be carried into the `FileSystemSandboxPolicy` at materialization time too, otherwise the approved cap does not actually reach execution.

viyatb-oai (approved):

PR #18717, opened 2026-04-20
[codex] Add tool interceptor support

cwolff-oai:
## Summary
- add local tool-interceptor server config for Codex exec
- route matching tool calls through the interceptor before normal execution
- support realistic red-teaming runs where connector/local tool calls can be replaced locally

## Validation
- built and used interceptor-enabled Codex binary for the scenario-pack red-team runs
- verified Gmail and Google Drive connector calls were intercepted rather than calling real connectors


cwolff-oai:
Closing this in favor of a Responses proxy approach that does not require changes in the codex-rs repo.

PR #18727, opened 2026-04-20
PR8: Revert agent identity fixture fix

adrian-openai:
Reverts #18697. This is PR8 in the reverse-order agent identity revert stack and should be merged first, before PR7. The original squash merge was tree-empty on main, so this PR intentionally carries an empty revert marker commit.


adrian-openai:

PR #18728, opened 2026-04-20
PR7: Revert background auth additional callsites

adrian-openai:
Reverts #18260. This is PR7 in the reverse-order agent identity revert stack and should be merged after PR8, before PR6.


adrian-openai:

PR #18729, opened 2026-04-20
PR6: Revert background agent task auth

adrian-openai:
Reverts #18094. This is PR6 in the reverse-order agent identity revert stack and should be merged after PR7, before PR5.


adrian-openai:

PR #18730, opened 2026-04-20
PR5: Revert downstream AgentAssertion auth

adrian-openai:
Reverts #17980. This is PR5 in the reverse-order agent identity revert stack and should be merged after PR6, before PR4.


adrian-openai:

PR #18731, opened 2026-04-20
PR4: Revert persisted agent task state

adrian-openai:
Reverts #17978. This is PR4 in the reverse-order agent identity revert stack and should be merged after PR5, before PR3.


adrian-openai:

PR #18732, opened 2026-04-20
PR3: Revert agent task registration

adrian-openai:
Reverts #17387. This is PR3 in the reverse-order agent identity revert stack and should be merged after PR4, before PR2.


adrian-openai:

PR #18733, opened 2026-04-20
PR2: Revert agent identity registration

adrian-openai:
Reverts #17386. This is PR2 in the reverse-order agent identity revert stack and should be merged after PR3, before PR1.


adrian-openai:

PR #18734, opened 2026-04-20
PR1: Revert use_agent_identity flag

adrian-openai:
Reverts #17385. This is PR1, the final PR in the reverse-order agent identity revert stack, and should be merged after PR2.


adrian-openai:

PR #18739, opened 2026-04-20
fix(tui): keep /copy aligned with rollback

### LINKED ISSUE #18718: /copy always copies most recent message, even after rewinding

codex-cli 0.121.0

Pro

gpt-5.4

Darwin 25.2.0 arm64 arm

ghostty

Using /copy always copies the *latest sent* message in a conversation, even when the Codex CLI has been rewound to an earlier state by using Esc and selecting an earlier message. This also means it's not possible to /copy an earlier message without re-generating it, since rewinding now requires sending a new message before I can copy from that branch again.

(cc @etraut-openai who was active on the original feature request)

- Send message "foo"
- Send message "bar"
- Rewind to "foo" by double-tapping Esc
- Send /copy
- Clipboard says "bar"

_No response_

### Additional information

_No response_

fcoury-oai:
## Why

Fixes #18718.

After rewinding a thread, `/copy` could still copy the latest assistant response from before the rewind. The transcript cells were rolled back, but the copy source was a single `last_agent_markdown` cache that was not synchronized with backtracking, so the visible conversation and copied content could diverge.

## What changed

`ChatWidget` now keeps a bounded copy history for the most recent 32 assistant responses, keyed by the visible user-turn count. When local rollback trims transcript cells, the copy cache is trimmed to the same surviving user-turn count so `/copy` uses the latest visible assistant response.

If the user rewinds past the retained copy window, `/copy` now reports:

```text
Cannot copy that response after rewinding. Only the most recent 32 responses are available to /copy.
```

The change also adds coverage for copying the latest surviving response after rollback and for the over-limit rewind message.

## Verification

- Manually resumed a synthetic 35-turn session, rewound within the retained window, and verified `/copy` copied the surviving response.
- Manually rewound past the retained window and verified `/copy` showed the 32-response limit message.
- `cargo test -p codex-tui slash_copy`
- `just fix -p codex-tui`
- `cargo insta pending-snapshots`

Note: `cargo test -p codex-tui` currently fails on unrelated model catalog and snapshot drift around the default model changing to `gpt-5.4`; the focused `/copy` tests pass after fixing the new test setup.


etraut-openai (approved):
Code looks good. I verified the bug before the change and the correct behavior after the change.

PR #18744, opened 2026-04-20
feat: add a built-in Amazon Bedrock model provider

celia-oai:
## Why

Codex needs a first-class `amazon-bedrock` model provider so users can select Bedrock without copying a full provider definition into `config.toml`. The provider has Codex-owned defaults for the pieces that should stay consistent across users: the display `name`, Bedrock `base_url`, and `wire_api`.

At the same time, users still need a way to choose the AWS credential profile used by their local environment. This change makes `amazon-bedrock` a partially modifiable built-in provider: code owns the provider identity and endpoint defaults, while user config can set `model_providers.amazon-bedrock.aws.profile`.

For example:

```toml
model_provider = "amazon-bedrock"

[model_providers.amazon-bedrock.aws]
profile = "codex-bedrock"
```

## What Changed

- Added `amazon-bedrock` to the built-in model provider map with:
  - `name = "Amazon Bedrock"`
  - `base_url = "https://bedrock-mantle.us-east-1.api.aws/v1"`
  - `wire_api = "responses"`
- Added AWS provider auth config with a profile-only shape: `model_providers.<id>.aws.profile`.
- Kept AWS auth config restricted to `amazon-bedrock`; custom providers that set `aws` are rejected.
- Allowed `model_providers.amazon-bedrock` through reserved-provider validation so it can act as a partial override.
- During config loading, only `aws.profile` is copied from the user-provided `amazon-bedrock` entry onto the built-in provider. Other Bedrock provider fields remain hard-coded by the built-in definition.
- Updated the generated config schema for the new provider AWS profile config.


celia-oai:
@codex review

codex-rs/model-provider-info/src/lib.rs
```diff
@@ -322,13 +395,15 @@ pub fn built_in_model_providers(
...
     let openai_provider = P::create_openai_provider(openai_base_url);
+    let amazon_bedrock_provider = P::create_amazon_bedrock_provider(/*aws*/ None);
 
     // We do not want to be in the business of adjucating which third-party
     // providers are bundled with Codex CLI, so we only include the OpenAI and
     // open source ("oss") providers by default. Users are encouraged to add to
     // `model_providers` in config.toml to add their own providers.
     [
         (OPENAI_PROVIDER_ID, openai_provider),
+        (AMAZON_BEDROCK_PROVIDER_ID, amazon_bedrock_provider),
```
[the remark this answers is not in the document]
celia-oai: this will be a follow-up

codex-rs/config/src/config_toml.rs
```diff
@@ -780,7 +782,10 @@ pub fn validate_reserved_model_provider_ids(
 ) -> Result<(), String> {
     let mut conflicts = model_providers
         .keys()
-        .filter(|key| RESERVED_MODEL_PROVIDER_IDS.contains(&key.as_str()))
+        .filter(|key| {
+            key.as_str() != AMAZON_BEDROCK_PROVIDER_ID
```
pakrym-oai, line 786: does overriding actually work?

do we want to support full override or just partial?

codex-rs/core/src/config/mod.rs
```diff
@@ -1807,7 +1808,16 @@ impl Config {
         let mut model_providers = built_in_model_providers(openai_base_url);
         // Merge user-defined providers into the built-in list.
         for (key, provider) in cfg.model_providers.into_iter() {
-            model_providers.entry(key).or_insert(provider);
+            if key == AMAZON_BEDROCK_PROVIDER_ID {
```
pakrym-oai: is there a better place for this logic?

codex-rs/core/src/config/config_tests.rs
```diff
@@ -369,6 +370,120 @@ command = "print-token"
...
 }
 
+#[test]
+fn rejects_provider_aws_for_custom_provider() {
+    let err = toml::from_str::<ConfigToml>(
+        r#"
+[model_providers.bedrock]
+name = "Amazon Bedrock"
+
+[model_providers.bedrock.aws]
```
pakrym-oai: tbh this feels a bit strange to nest aws under bedrock but not blocking

pakrym-oai (approved):

codex-rs/config/src/config_toml.rs
celia-oai: just partial. I can add stricter checks here to make it less confusing

codex-rs/core/src/config/mod.rs
celia-oai: added a helper function for this, might be slightly better?

PR #18747, opened 2026-04-20
[codex-analytics] add tool review event schema

rhan-oai:
## Why

We want to emit terminal review analytics for tool-related approval flows, but the event contract needs to exist before the reducer can publish anything.

This PR is the schema-only slice for the Codex review event family.

## What changed

- add the `ReviewEvent` analytics envelope in `codex-rs/analytics/src/events.rs`
- define the review subject kind, reviewer, trigger, terminal status, and post-review resolution enums
- define the review event payload with thread, turn, item, lineage, tool, and timing fields that the emitter stack will populate

## Verification

- stacked verification in dependent PRs: `cargo test -p codex-analytics analytics_client_tests --manifest-path codex-rs/Cargo.toml`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/18747).
* #18748
* #21434
* __->__ #18747
* #17090
* #17089
* #20514

### DIFF
--- a/codex-rs/analytics/src/events.rs
+++ b/codex-rs/analytics/src/events.rs
@@ -70,6 +70,8 @@ pub(crate) enum TrackEventRequest {
     CollabAgentToolCall(CodexCollabAgentToolCallEventRequest),
     WebSearch(CodexWebSearchEventRequest),
     ImageGeneration(CodexImageGenerationEventRequest),
+    #[allow(dead_code)]
+    ReviewEvent(CodexReviewEventRequest),
     PluginUsed(CodexPluginUsedEventRequest),
     PluginInstalled(CodexPluginEventRequest),
     PluginUninstalled(CodexPluginEventRequest),
@@ -442,7 +444,7 @@ pub(crate) struct CodexToolItemEventBase {
     pub(crate) item_id: String,
     pub(crate) app_server_client: CodexAppServerClientMetadata,
     pub(crate) runtime: CodexRuntimeMetadata,
-    pub(crate) thread_source: Option<&'static str>,
+    pub(crate) thread_source: Option<ThreadSource>,
     pub(crate) subagent_source: Option<String>,
     pub(crate) parent_thread_id: Option<String>,
     pub(crate) tool_name: String,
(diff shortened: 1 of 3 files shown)


codex-rs/analytics/src/events.rs
```diff
@@ -460,6 +462,74 @@ pub(crate) struct CodexToolItemEventBase {
...
+    McpToolCall,
+    Permissions,
+    NetworkAccess,
+}
+
+#[allow(dead_code)]
+#[derive(Clone, Copy, Debug, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum ToolReviewReviewer {
+    Guardian,
```
owenlin0, line 482: i was going to suggest `AutoReview` to make it inline with our user-facing naming, but i think the ship has already sailed for analytics and it's probably better to just be consistent with `Guardian` internally? thoughts?

codex-rs/analytics/src/events.rs
```diff
@@ -460,6 +462,74 @@ pub(crate) struct CodexToolItemEventBase {
...
 }
 
+#[allow(dead_code)]
+#[derive(Clone, Copy, Debug, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum ToolReviewToolKind {
+    CommandExecution,
+    FileChange,
+    McpToolCall,
+    Permissions,
```
owenlin0, line 474: hmm, why is `Permissions` a tool kind? it's not, is it?

codex-rs/analytics/src/events.rs
```diff
@@ -460,6 +462,74 @@ pub(crate) struct CodexToolItemEventBase {
...
+}
+
+#[allow(dead_code)]
+#[derive(Clone, Copy, Debug, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum ToolReviewTrigger {
+    Initial,
+    SandboxRetry,
+    NetworkRetry,
+    SubcommandExecve,
```
owenlin0: nit: i feel like `*Retry` sounds a bit confusing to me as a value for "review trigger". what do you think of

```
pub(crate) enum ToolReviewTrigger {
    Initial,
    SandboxDenial,
    NetworkPolicyDenial,
    ExecveIntercept,
}
```
?

to me this explains the reason behind the review a bit more clearly

codex-rs/analytics/src/events.rs
```diff
@@ -460,6 +462,74 @@ pub(crate) struct CodexToolItemEventBase {
...
+}
+
+#[allow(dead_code)]
+#[derive(Clone, Copy, Debug, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum ToolReviewStatus {
```
owenlin0: is it worth breaking out:
- whether the tool call was approved/denied/timed out
- what mutating actions taken afterwards
?

something like:
```
pub(crate) enum ToolReviewStatus {
    Approved,
    Denied,
    Aborted,
    TimedOut,
}

pub(crate) enum ToolReviewResolution {
    None,
    SessionApproval,
    ExecPolicyAmendment,
    NetworkPolicyAllow,
    NetworkPolicyDeny,
}
```

owenlin0: feels like it'd be cleaner to just be able to query what % of tool calls were approved or denied that way

codex-rs/analytics/src/events.rs
rhan-oai: considered this as well, but agree we should be consistent

owenlin0 (approved):

PR #18748, opened 2026-04-20
[codex-analytics] emit terminal review events

rhan-oai:
## Why

Review telemetry should describe reviews as first-class events, not only as counters denormalized onto terminal tool-item events. That lets us analyze guardian and user reviews consistently across command execution, file changes, permissions, and network access, while still preserving the terminal item summaries that existing tool analytics need.

To make those review events accurate, analytics also needs the observed completion time for each review and enough command metadata to distinguish `shell` from `unified_exec` reviews.

## What changed

- emit generic `codex_review_event` rows for completed user and guardian reviews, with review subjects, reviewer, trigger, terminal status, resolution, and observed duration
- reduce approval request / response / abort facts into review events for command execution, file change, and permissions flows
- keep denormalized review counts, final approval outcome, and permission-request flags on terminal tool-item events for item-associated reviews
- plumb review completion timing so user-review responses and aborts use app-server-observed completion times, while guardian analytics reuse the same terminal timestamps emitted on guardian assessment events
- carry command approval `source` through the protocol and app-server layers so review analytics can distinguish `shell` from `unified_exec`
- add analytics coverage for user-review emission, guardian-review emission, permission reviews that should not denormalize onto tool items, item-summary isolation across threads, and the serialized review-event shape

## Verification

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/18748).
* __->__ #18748
* #21434
* #18747
* #17090
* #17089
* #20514

### DIFF
--- a/codex-rs/analytics/src/reducer.rs
+++ b/codex-rs/analytics/src/reducer.rs
@@ -22,6 +22,8 @@ use crate::events::CodexMcpToolCallEventParams;
 use crate::events::CodexMcpToolCallEventRequest;
 use crate::events::CodexPluginEventRequest;
 use crate::events::CodexPluginUsedEventRequest;
+use crate::events::CodexReviewEventParams;
+use crate::events::CodexReviewEventRequest;
 use crate::events::CodexRuntimeMetadata;
 use crate::events::CodexToolItemEventBase;
 use crate::events::CodexTurnEventParams;
@@ -30,15 +32,20 @@ use crate::events::CodexTurnSteerEventParams;
 use crate::events::CodexTurnSteerEventRequest;
 use crate::events::CodexWebSearchEventParams;
 use crate::events::CodexWebSearchEventRequest;
+use crate::events::FinalApprovalOutcome;
 use crate::events::GuardianReviewEventParams;
 use crate::events::GuardianReviewEventPayload;
 use crate::events::GuardianReviewEventRequest;
+use crate::events::ReviewResolution;
+use crate::events::ReviewStatus;
+use crate::events::ReviewSubjectKind;
+use crate::events::ReviewTrigger;
+use crate::events::Reviewer;
 use crate::events::SkillInvocationEventParams;
 use crate::events::SkillInvocationEventRequest;
 use crate::events::ThreadInitializedEvent;
 use crate::events::ThreadInitializedEventParams;
 use crate::events::ToolItemFailureKind;
-use crate::events::ToolItemFinalApprovalOutcome;
 use crate::events::ToolItemTerminalStatus;
 use crate::events::TrackEventRequest;
 use crate::events::WebSearchActionKind;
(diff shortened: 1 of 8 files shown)


rhan-oai:
@codex

codex-rs/analytics/src/reducer.rs
```diff
@@ -320,13 +373,23 @@ impl AnalyticsReducer {
...
+                request,
+            } => {
+                self.ingest_server_request(connection_id, *request);
+            }
             AnalyticsFact::ServerResponse {
-                response: _response,
-                ..
-            } => {}
+                completed_at_ms,
+                response,
```
owenlin0, line 426: This path consumes `ServerResponse` facts that app-server records from the raw JSON-RPC result before the approval-specific handlers apply their effective fallback/validation. If the client returns a malformed command/file approval response, the handler later falls back to `Decline`, but analytics emits no review event because `response_from_result` never produced a typed response. For permissions, analytics can mark a raw non-empty grant as approved before the handler intersects/validates it and potentially grants nothing. Consider tracking the effective approval decision after the approval handler computes it, so analytics matches the behavior actually applied to the turn.

codex-rs/app-server-protocol/src/protocol/v2/item.rs
```diff
@@ -1270,6 +1274,11 @@ pub struct CommandExecutionRequestApprovalParams {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     #[ts(optional = nullable)]
     pub approval_id: Option<String>,
+    /// Originating command tool for this approval request.
+    ///
+    /// Uses `#[serde(default)]` for backwards compatibility with older senders.
+    #[serde(default = "default_guardian_command_source")]
+    pub source: GuardianCommandSource,
```
owenlin0: let's remove this, I don't want to add this implementation detail to the public app-server API.

can we populate codex_review_event.subject_name as just `command_execution` instead?

codex-rs/analytics/src/reducer.rs
owenlin0: ^ codex-generated - if it complicates the implementation too much I actually don't think it's worth doing, fwiw

owenlin0: feels like we should be tracking malformed client responses separately anyways

rhan-oai: fixed up permissions issue, agree on pushing off malformed response tracking

rhan-oai:
@codex

codex-rs/protocol/src/approvals.rs
```diff
@@ -231,6 +235,11 @@ pub struct ExecApprovalRequestEvent {
     pub turn_id: String,
     #[ts(type = "number")]
     pub started_at_ms: i64,
+    /// Originating command tool for this approval request.
+    ///
+    /// Uses `#[serde(default)]` for backwards compatibility with older senders.
+    #[serde(default = "default_guardian_command_source")]
+    pub source: GuardianCommandSource,
```
owenlin0: Can we remove this too? We also polluted the protocol a bit

codex-rs/core/src/guardian/review.rs
```diff
@@ -320,9 +322,10 @@ async fn run_guardian_review(
     ))
     .await;
 
-    let (assessment, count_denial_for_circuit_breaker) = match outcome {
+    let (assessment, count_denial_for_circuit_breaker, completed_at_ms) = match outcome {
         GuardianReviewOutcome::Completed(assessment) => {
             let approved = matches!(assessment.outcome, GuardianAssessmentOutcome::Allow);
+            let completed_at_ms = now_unix_timestamp_ms();
```
owenlin0: nit: we do this in each of the `match outcome` branch arms below. can we define it once right above the `match outcome`?

owenlin0 (approved):
just some small comments at this point, but preapproving

codex-rs/protocol/src/approvals.rs
rhan-oai: whoops yes my bad

PR #18757, opened 2026-04-20
fix: fully revert agent identity runtime wiring

efrazer-oai:
## Summary

This PR fully reverts the previously merged Agent Identity runtime integration from the old stack:
https://github.com/openai/codex/pull/17387/changes

It removes the Codex-side task lifecycle wiring, rollout/session persistence, feature flag plumbing, lazy `auth.json` mutation, background task auth paths, and request callsite changes introduced by that stack.

This leaves the repo in a clean pre-AgentIdentity integration state so the follow-up PRs can reintroduce the pieces in smaller reviewable layers.

## Stack

1. This PR: full revert
2. https://github.com/openai/codex/pull/18871: move Agent Identity business logic into a crate
3. https://github.com/openai/codex/pull/18785: add explicit AgentIdentity auth mode and startup task allocation
4. https://github.com/openai/codex/pull/18811: migrate auth callsites through AuthProvider

## Testing

Tests: targeted Rust checks, cargo-shear, Bazel lock check, and CI.

PR #18785, opened 2026-04-21
feat: add explicit AgentIdentity auth mode

efrazer-oai:
## Summary

This PR adds `CodexAuth::AgentIdentity` as an explicit auth mode.

An AgentIdentity auth record is a standalone `auth.json` mode. When `AuthManager::auth().await` loads that mode, it registers one process-scoped task and stores it in runtime-only state on the auth value. Header creation stays synchronous after that because the task is initialized before callers receive the auth object.

This PR also removes the old feature flag path. AgentIdentity is selected by explicit auth mode, not by a hidden flag or lazy mutation of ChatGPT auth records.

Reference old stack: https://github.com/openai/codex/pull/17387/changes

## Design Decisions

- AgentIdentity is a real auth enum variant because it can be the only credential in `auth.json`.
- The process task is ephemeral runtime state. It is not serialized and is not stored in rollout/session data.
- Account/user metadata needed by existing Codex backend checks lives on the AgentIdentity record for now.
- `is_chatgpt_auth()` remains token-specific.
- `uses_codex_backend()` is the broader predicate for ChatGPT-token auth and AgentIdentity auth.

## Stack

1. https://github.com/openai/codex/pull/18757: full revert
2. https://github.com/openai/codex/pull/18871: isolated Agent Identity crate
3. This PR: explicit AgentIdentity auth mode and startup task allocation
4. https://github.com/openai/codex/pull/18811: migrate Codex backend auth callsites through AuthProvider
5. https://github.com/openai/codex/pull/18904: accept AgentIdentity JWTs and load `CODEX_AGENT_IDENTITY`

## Testing

Tests: targeted Rust checks, cargo-shear, Bazel lock check, and CI.


efrazer-oai:
Codex review comments resolved.

efrazer-oai:
Codex r

> ### 💡 Codex Review
> https://github.com/openai/codex/blob/b23a44fd62024b18e3c2771e00222314bfbcb5b3/codex-rs/models-manager/src/manager.rs#L415
> 
> 
> **P1 Badge Treat AgentIdentity as ChatGPT auth for model refresh**
> `refresh_available_models` exits unless `auth_mode == Some(AuthMode::Chatgpt)`. After introducing explicit `AgentIdentity`, authenticated AgentIdentity users take this early-return path and skip remote model-catalog refresh (unless command auth is configured), leaving stale or missing model availability.
> 
> ℹ️ About Codex in GitHub

PR 3 is responsible for this fix as it does a broader migration of is_chatgpt_auth callsites

codex-rs/agent-identity/src/lib.rs
```diff
@@ -86,6 +107,41 @@ pub fn sign_task_registration_payload(
     Ok(BASE64_STANDARD.encode(signing_key.sign(payload.as_bytes()).to_bytes()))
 }
 
+pub fn register_agent_task_blocking(
```
pakrym-oai: should this be async/

efrazer-oai: Intentionally kept sync here as we need the task on startup and want to block on it, some of the alternatives are in PR description. But lmk if any of the other alternatives (e.g. making this async, propagating upwards into making AuthManager::new async) work better; i don't have a strong opinion and the choice was hard for me to reason about fully

codex-rs/core/src/client.rs
```diff
@@ -1727,7 +1727,9 @@ impl AuthRequestTelemetryContext {
         Self {
             auth_mode: auth_mode.map(|mode| match mode {
                 AuthMode::ApiKey => "ApiKey",
-                AuthMode::Chatgpt | AuthMode::ChatgptAuthTokens => "Chatgpt",
+                AuthMode::Chatgpt | AuthMode::ChatgptAuthTokens | AuthMode::AgentIdentity => {
```
pakrym-oai, line 1706: maybe worth mentioning agent identity

codex-rs/agent-identity/src/lib.rs
pakrym-oai: Can we resolve when .auth is called on AuthManager?

codex-rs/login/src/auth/manager.rs
```diff
@@ -1462,6 +1614,7 @@ impl AuthManager {
             config.cli_auth_credentials_store_mode(),
         );
         auth_manager.set_forced_chatgpt_workspace_id(config.forced_chatgpt_workspace_id());
+        auth_manager.set_chatgpt_backend_base_url(Some(config.chatgpt_base_url()));
```
pakrym-oai: Why don't we construct the manager with correct values and avoid RwLock?

codex-rs/login/src/auth/manager.rs
```diff
@@ -1424,6 +1528,54 @@ impl AuthManager {
...
+            .and_then(|guard| guard.clone())
+    }
+
+    /// Returns the default authorization header for ChatGPT backend requests.
+    pub async fn chatgpt_authorization_header(self: &Arc<Self>) -> Option<String> {
+        let auth = self.auth().await?;
+        self.chatgpt_authorization_header_for_auth(&auth).await
+    }
+
+    pub async fn chatgpt_authorization_header_for_auth(
```
pakrym-oai: do we still need this? I thought we are doing headers via AuthProvider ?

codex-rs/login/src/auth/manager.rs
```diff
@@ -1424,6 +1528,54 @@ impl AuthManager {
...
+    pub fn set_chatgpt_backend_base_url(&self, chatgpt_base_url: Option<String>) {
+        if let Ok(mut guard) = self.chatgpt_base_url.write()
+            && *guard != chatgpt_base_url
+        {
+            *guard = chatgpt_base_url;
+            self.auth_state_tx.send_replace(());
+        }
+    }
+
+    pub fn chatgpt_base_url(&self) -> Option<String> {
```
pakrym-oai: we shouldn't expose this publicly

codex-rs/login/src/auth/manager.rs
```diff
@@ -1424,6 +1528,54 @@ impl AuthManager {
             .and_then(|guard| guard.clone())
     }
 
+    /// Sets the ChatGPT backend URL override for future auth runtime initialization.
+    /// Passing `None` clears the override and returns future initialization to the
+    /// default backend URL.
+    pub fn set_chatgpt_backend_base_url(&self, chatgpt_base_url: Option<String>) {
```
pakrym-oai: should just be set in ctor

codex-rs/login/src/auth/manager.rs
```diff
@@ -1424,6 +1528,54 @@ impl AuthManager {
...
+
+    pub fn chatgpt_bearer_token_for_auth(auth: &CodexAuth) -> Option<String> {
+        auth.get_token().ok().filter(|token| !token.is_empty())
+    }
+
+    pub fn chatgpt_bearer_authorization_header_for_auth(auth: &CodexAuth) -> Option<String> {
+        Self::chatgpt_bearer_token_for_auth(auth).map(|token| format!("Bearer {token}"))
+    }
+
+    pub fn subscribe_auth_state(&self) -> watch::Receiver<()> {
```
pakrym-oai: Do we need this subscribe?

codex-rs/login/src/auth/manager.rs
```diff
@@ -73,7 +77,15 @@ struct ChatgptAuthState {
 
 impl PartialEq for CodexAuth {
     fn eq(&self, other: &Self) -> bool {
-        self.api_auth_mode() == other.api_auth_mode()
+        match (self, other) {
```
pakrym-oai: do we care about this implemenation?

efrazer-oai: No, fixed.

codex-rs/login/src/auth/manager.rs
efrazer-oai: No, was used for syncing chatgpt_url but we're passing in top level now, removed.

codex-rs/login/src/auth/manager.rs
efrazer-oai: Done

codex-rs/login/src/auth/manager.rs
efrazer-oai: Done

codex-rs/login/src/auth/manager.rs
efrazer-oai: Removed -> these were unused after PR after this, but no point having them here.

codex-rs/login/src/auth/manager.rs
efrazer-oai: Done

pakrym-oai (approved):

codex-rs/login/src/auth/manager.rs
```diff
@@ -1287,7 +1353,12 @@ impl AuthManager {
             tracing::error!("Failed to refresh token: {}", err);
             return Some(auth);
         }
-        self.auth_cached()
+        let auth = self.auth_cached()?;
+        if let Err(err) = auth.initialize_runtime(self.chatgpt_base_url.clone()).await {
```
adrian-openai, line 1357: AuthManager::auth() loads the cached AgentIdentity record and then returns None if initialize_runtime() fails. That makes valid pre-registered identity material indistinguishable from no auth whenever task registration has a transient network/TLS/proxy/backend failure or the task state is rejected. I wonder if we should keep the cached identity/account metadata visible and surface a distinct runtime-init/header error for request paths instead of flattening this into signed-out state.

codex-rs/agent-identity/src/lib.rs
```diff
@@ -86,6 +107,51 @@ pub fn sign_task_registration_payload(
...
+    chatgpt_base_url: &str,
+    key: AgentIdentityKey<'_>,
+) -> Result<String> {
+    let timestamp = Utc::now().to_rfc3339_opts(SecondsFormat::Secs, true);
+    let request = RegisterTaskRequest {
+        signature: sign_task_registration_payload(key, &timestamp)?,
+        timestamp,
+    };
+
+    let response = reqwest::Client::builder()
```
adrian-openai: By using this bare `reqwest::Client` with only a timeout, this startup-critical call misses the standard Codex client behavior used by ordinary backend traffic, including custom CA handling, default originator/user-agent headers, sandbox proxy policy, and future centralized transport behavior. I think we should use an injected/prepared client or move the HTTP call into codex-login so task registration uses the same transport path as other first-party backend calls.

adrian-openai (commented):
Had a few items to consider

codex-rs/agent-identity/src/lib.rs
efrazer-oai: Changing this!

PR #18786, opened 2026-04-21
Fallback display names for TUI skill mentions

canvrno-oai:
This updates TUI skill mentions to show a fallback label when a skill does not define a display name, so unnamed skills remain understandable in the picker without changing behavior for skills that already have one.

Screenshot 2026-04-20 at 6 25 15 PM


canvrno-oai:
@codex review

etraut-openai (approved):
Looks good. 

I built locally and tested with a custom plugin that contained a skill without a display name.

PR #18807, opened 2026-04-21
chore(app-server) linguist-generated

dylan-hurd-oai:
## Summary
Start marking app-server schema files as [linguist-generated](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github), so we can more easily parse reviews


sayan-oai (approved):
🙏

PR #18808, opened 2026-04-21
Add Codex Statsig integration

dgoya-oai:
## Summary

- Add a dedicated `codex-statsig` Rust crate around the Statsig Rust SDK.
- Map available Codex auth/session metadata into `StatsigUser`, including account/workspace IDs and `email_domain_type` instead of raw email.
- Configure Statsig event logging with ChatGPT auth headers and startup-resolved analytics opt-out behavior via `disable_network`.

## Test Plan

- `cargo test -p codex-statsig`
- `just bazel-lock-check`


codex-rs/statsig/src/lib.rs
```diff
@@ -0,0 +1,1020 @@
...
+        self.network.shutdown();
+        Ok(())
+    }
+
+    fn should_schedule_background_flush(&self) -> bool {
+        true
+    }
+}
+
+#[cfg(test)]
```
qw9-oai, line 450: i found that codex tends to put test into the same file as the actual lib. i had to explicitly tell it to split lib vs. test in my local codex config, but maybe it's time to put that into agent.md

dgoya-oai:

codex-rs/statsig/src/lib.rs
dgoya-oai: Split this out now: `src/lib.rs` keeps the implementation and declares `#[cfg(test)] mod tests;`, while the existing unit tests moved to `src/tests.rs` so they stay private-module tests without widening the public API.

PR #18817, opened 2026-04-21
[codex] Fix interrupt prompt history race

jgershen-oai:
## Summary

Fixes a race where a regular turn could emit `TurnStarted` before the initial user prompt was recorded. If the app interrupted in that window, core could persist `<turn_aborted>` without ever saving the prompt that started the turn.

This change queues accepted regular-turn startup input on the turn context, serializes startup transcript writes through a per-turn lock, and has the interrupt path drain that queue before writing the interrupted-turn marker.

## Root cause

The old regular-turn path recorded the initial user prompt inside `run_turn()`, after client-visible turn startup had already happened. The interrupt path could therefore run before prompt recording and append the abort marker to history first, or as the only item.

## What changed

- Add per-turn storage for startup transcript input and a per-turn serialization lock.
- Queue non-empty regular-turn input before spawning the regular task.
- Move startup prompt serialization into a shared drain helper that runs session-start hooks, user-prompt-submit hooks, prompt recording, and hook additional-context recording.
- Call the same drain helper from the interrupt path before appending `<turn_aborted>`.
- Add a deterministic regression test that asserts interrupted history contains the original user prompt before the abort marker.

## Validation

Passed:

- `just fmt`
- `just fix -p codex-core`
- `git diff --check`
- `cargo test -p codex-core abort_regular_task_records_prompt_before_interrupt_marker`
- `cargo test -p codex-core regular_turn_emits_turn_started_without_waiting_for_startup_prewarm`
- `cargo test -p codex-core interrupting_regular_turn_waiting_on_startup_prewarm_emits_turn_aborted`
- `cargo test -p codex-core blocked_user_prompt_submit_persists_additional_context_for_next_turn`
- `cargo test -p codex-core user_message_item_is_emitted`
- `cargo test -p codex-core abort_review_task_emits_exited_then_aborted_and_records_history`

Also attempted `cargo test -p codex-core`. The lib/unit portion passed (`1682 passed; 0 failed; 3 ignored`), but the integration `tests/all.rs` portion had local environment failures in approval/otel/user-shell env isolation tests. CI should provide the authoritative full-suite signal.

### DIFF
--- a/codex-rs/core/src/hook_runtime.rs
+++ b/codex-rs/core/src/hook_runtime.rs
@@ -273,8 +273,15 @@ pub(crate) async fn drain_turn_start_transcript_inputs(
     sess: &Arc<Session>,
     turn_context: &Arc<TurnContext>,
 ) -> bool {
-    let _guard = turn_context.transcript_serialization_lock.lock().await;
+    let Ok(_permit) = turn_context.transcript_serialization_lock.acquire().await else {
+        return false;
+    };
     if run_pending_session_start_hooks(sess, turn_context).await {
+        turn_context
+            .turn_start_transcript_inputs
+            .lock()
+            .await
+            .clear();
         return false;
     }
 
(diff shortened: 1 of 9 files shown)


jgershen-oai:
@codex review

PR #18820, opened 2026-04-21
Queue follow-up input during user shell commands

### LINKED ISSUE #17954: steer does not work with shell commands

v0.121.0-alpha.14 | v0.120.0

Pro

_No response_

_No response_

_No response_

Sending steed command while manually executed command is issued breaks codex

1. run any command like `!sleep 10`
2. send steer follow up like `hi`
3. wait for sleep to finish

After command is done, codex still does not continue and stays in `Working` counting to infinite.
Pressing ESC will lock up cli completely.

### Additional information

_No response_

etraut-openai:
Fixes #17954.

## Why
When a manual shell command like `!sleep 10` is running, submitting plain text such as `hi` currently sends that text as a steer for the active shell turn. User shell turns are not steerable like model turns, so the TUI can remain stuck in `Working` after the shell command finishes.

## What Changed
- Detect when the only active work is one or more `ExecCommandSource::UserShell` commands.
- Queue plain submitted input in that state so it drains after the shell command and shell turn complete.
- Preserve `!cmd` submissions during running work so explicit shell commands keep their existing behavior.
- Add regression coverage for the `!sleep 10` plus `hi` flow in `chatwidget::tests::exec_flow::user_message_during_user_shell_command_is_queued_not_steered`.

## Verification
- Manually confirmed hang before the fix and no hang after the fix

### DIFF
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -5411,6 +5411,12 @@ impl ChatWidget {
                         let should_submit_now =
                             self.is_session_configured() && !self.is_plan_streaming_in_tui();
                         if should_submit_now {
+                            if self.only_user_shell_commands_running()
+                                && !user_message.text.starts_with('!')
+                            {
+                                self.queue_user_message(user_message);
+                                return;
+                            }
                             // Submitted is emitted when user submits.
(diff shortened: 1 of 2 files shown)


fcoury-oai (approved):
Replicated the issue and confirmed it is fixed in this branch. 

Code looks good to me, approved!

PR #18848, opened 2026-04-21
feat: baseline lib

jif-oai:
This add with 2 entry point:
* `reset_git_repository` that takes a directory and set it as a new git root
* `diff_since_latest_init` this returns the diff for a given directory since the last `reset_git_repository`


jif-oai:
@codex review

PR #18867, opened 2026-04-21
sandboxing: materialize cwd-relative permission globs

bolinfest:
## Why

#18275 anchors session-scoped `:cwd` and `:project_roots` grants to the request cwd before recording them for reuse. Relative deny glob entries need the same treatment. Without anchoring, a stored session permission can keep a pattern such as `**/*.env` relative, then reinterpret that deny against a later turn cwd. That makes the persisted profile depend on the cwd at reuse time instead of the cwd that was reviewed and approved.

## What changed

`intersect_permission_profiles` now materializes retained `FileSystemPath::GlobPattern` entries against the request cwd, matching the existing materialization for cwd-sensitive special paths.

Materialized accepted grants are now deduplicated before deny retention runs. This keeps the sticky-grant preapproval shape stable when a repeated request is merged with the stored grant and both `:cwd = write` and the materialized absolute cwd write are present.

The preapproval check compares against the same materialized form, so a later request for the same cwd-relative deny glob still matches the stored anchored grant instead of re-prompting or rejecting.

Tests cover both the storage path and the preapproval path: a session-scoped `:cwd = write` grant with `**/*.env = none` is stored with both the cwd write and deny glob anchored to the original request cwd, cannot be reused from a later cwd, and remains preapproved when re-requested from the original cwd after merging with the stored grant.

## Verification

- `cargo test -p codex-sandboxing policy_transforms`
- `cargo test -p codex-core --lib relative_deny_glob_grants_remain_preapproved_after_materialization`
- `cargo clippy -p codex-sandboxing --tests -- -D clippy::redundant_clone`
- `cargo clippy -p codex-core --lib -- -D clippy::redundant_clone`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/18867).
* #18288
* #18287
* #18286
* #18285
* #18284
* #18283
* #18282
* #18281
* #18280
* #18279
* #18278
* #18277
* #18276
* __->__ #18867

### DIFF
--- a/codex-rs/core/src/tools/handlers/mod.rs
+++ b/codex-rs/core/src/tools/handlers/mod.rs
@@ -193,8 +193,7 @@ pub(super) async fn apply_granted_turn_permissions(
     );
     let permissions_preapproved = match (effective_permissions.as_ref(), granted_permissions) {
         (Some(effective_permissions), Some(granted_permissions)) => {
-            intersect_permission_profiles(effective_permissions.clone(), granted_permissions, cwd)
-                == *effective_permissions
+            permissions_are_preapproved(effective_permissions, granted_permissions, cwd)
         }
         _ => false,
     };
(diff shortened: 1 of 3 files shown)


evawong-oai (approved):

codex-rs/core/src/tools/handlers/mod.rs
```diff
@@ -213,17 +212,37 @@ pub(super) async fn apply_granted_turn_permissions(
...
+    granted_permissions: PermissionProfile,
+    cwd: &Path,
+) -> bool {
+    let materialized_effective_permissions = intersect_permission_profiles(
+        effective_permissions.clone(),
+        effective_permissions.clone(),
+        cwd,
+    );
+    intersect_permission_profiles(effective_permissions.clone(), granted_permissions, cwd)
+        == materialized_effective_permissions
```
viyatb-oai, line 226: This comparison still fails for the real sticky-grant shape this PR is trying to preserve. After a prior approval stores `:cwd` write plus `**/*.env` deny as `/repo` write plus `/repo/**/*.env`, a later command that repeats the original relative request gets merged with the stored grant before this helper runs. `intersect_permission_profiles(effective, effective, cwd)` then materializes both `:cwd` and `/repo` into duplicate `/repo` write entries, while `intersect_permission_profiles(effective, granted, cwd)` only has one. The equality check returns false, so an already-approved permission profile is no longer treated as preapproved and can be rejected when only request-permissions preapproval is enabled. De-duplicate after materializing entries, or compare normalized canonical profiles.

viyatb-oai:
```
let mut accepted_entries = Vec::new();
for entry in granted_file_system.entries.iter().filter(|entry| {
    granted_file_system_entry_within_request(
        &requested_file_system,
        &requested_policy,
        requested_read_deny_matcher.as_ref(),
        entry,
        cwd,
    )
}) {
    let entry = materialize_cwd_dependent_entry(entry, cwd);
    if !accepted_entries.contains(&entry) {
        accepted_entries.push(entry);
    }
}
let mut entries = accepted_entries.clone();
```

viyatb-oai (approved):
looks good - one small fix needed!

PR #18868, opened 2026-04-21
Add MITM hook config model

evawong-oai:
## Stack
1. This PR adds MITM hook config and model only.
2. Runtime follow up: #20659 wires hook enforcement into the proxy request path.
3. User facing config follow up: #18240 moves MITM policy into the PermissionProfile network tree.

## Why
1. Viyat asked for the original parent PR to be split so reviewers can inspect the policy model before request behavior changes.
2. This PR gives the proxy a typed MITM hook model, validation, matcher compilation, permissions TOML plumbing, schema support, and config tests.
3. This PR deliberately does not change CONNECT or MITM request handling.
4. Keeping runtime behavior out of this PR makes the review boundary simple: does the policy model parse, validate, compile, and lower correctly.

## Summary
1. Add the MITM hook config model and matcher compilation.
2. Validate hosts, methods, paths, query matchers, header matchers, secret sources, and reserved body matching.
3. Add wildcard matcher support for path, query value, and header value matching.
4. Add permissions TOML and schema support for flat runtime hook config.
5. Add config loader tests for MITM hook overlay behavior.

## Validation
1. Regenerated the config schema.
2. Ran the network proxy MITM hook unit tests.
3. Ran the core permission profile MITM hook parsing tests.
4. Ran the core config schema fixture test.
5. Ran the scoped Clippy fixer for the network proxy crate.
6. Ran the scoped Clippy fixer for the core crate.

## Notes
1. Runtime enforcement moved to #20659.
2. User facing PermissionProfile TOML shape remains in #18240.


codex-rs/network-proxy/src/mitm.rs
```diff
@@ -302,7 +327,43 @@ async fn mitm_blocking_response(
...
+
+    match policy
+        .app_state
+        .evaluate_mitm_hook_request(&policy.target_host, req)
+        .await?
+    {
+        HookEvaluation::Matched { actions } => {
+            return Ok(MitmPolicyDecision::Allow {
+                hook_actions: Some(actions),
+            });
```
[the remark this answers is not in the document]
evawong-oai: This is valid, but I am punting the enforcement flip to a separate before rollout PR because it changes user visible blocking semantics. The Notion rollout checklist still has block UX, sanitized request trace, production observability, feature gate, CA and client validation, and dogfood pending. Until those pieces are in place, this PR should only land runtime hook plumbing and should not start blocking newly matched writes.

I did clean up the full mode test wording so the branch no longer implies hooks should bypass the limited mode method clamp. The ordering change remains a blocker before any protected destination is dogfooded or gated on.

evawong-oai:
Codex review item about managed constraints for MITM hook fields: this is valid, and it is tracked in Linear as AGE 248.

Why it is not in this PR: PR 18868 is now only the typed MITM hook config model. Enforcing trusted constraints for `network.mitm` and `network.mitm_hooks` depends on the final protected destination policy shape and the PermissionProfile trusted layer rules. Putting that here would mix model parsing, precedence, and central policy enforcement in one review.

This remains blocked before dogfood or feature gate rollout. Trusted config must prevent user or project config from disabling or replacing managed MITM hooks before protected destinations are enabled.

The separate method policy ordering concern is tracked in Linear as AGE 247.

codex-rs/network-proxy/src/mitm.rs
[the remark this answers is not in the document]
evawong-oai: Resolving this outdated thread after the split. The punt is still intentional. PR 18868 is now the config model slice, and method policy ordering is tracked in Linear as AGE 247 before dogfood or feature gate rollout.

evawong-oai:
Thanks. This is intentional for this PR.

The MITM feature is still being built. PR 18868 only lands the typed config model. It does not turn on enforcement.

Enforcement will come in later PRs after the user flow, logging, feature gate, and protected destination policy are ready. That avoids surprising users while the rollout pieces are still being developed.

The follow up work is tracked in Linear:
1. AGE 247 for method policy ordering.
2. AGE 248 for managed config constraints.

Those tasks are required before dogfood or feature gate rollout.

codex-rs/network-proxy/src/mitm_hook.rs
```diff
@@ -0,0 +1,1047 @@
...
+            compile_path_matchers(&hook.matcher.path_prefixes).with_context(|| {
+                format!("invalid network.mitm_hooks[{hook_index}].match.path_prefixes")
+            })?;
+        if path_prefixes.is_empty() {
+            return Err(anyhow!(
+                "network.mitm_hooks[{hook_index}].match.path_prefixes must not be empty"
+            ));
+        }
+
+        if let Some(body) = hook.matcher.body.as_ref() {
```
winston-openai, line 203: we'll probably only want to do this on certain content-types (e.g. JSON, form data, graphql)

codex-rs/config/src/permissions_toml.rs
```diff
@@ -219,6 +263,12 @@ impl NetworkToml {
         if let Some(allow_local_binding) = self.allow_local_binding {
             config.network.allow_local_binding = allow_local_binding;
         }
+        if let Some(mitm) = self.mitm {
+            config.network.mitm = mitm;
+        }
+        if let Some(mitm_hooks) = self.mitm_hooks.as_ref() {
+            config.network.mitm_hooks = mitm_hooks.clone();
+        }
```
winston-openai: Is there any validation if these are overwritten?

codex-rs/network-proxy/src/config.rs
```diff
@@ -273,8 +278,8 @@ pub enum NetworkMode {
     /// blocked unless MITM is enabled so the proxy can enforce method policy on inner requests.
     /// SOCKS5 remains blocked in limited mode.
     Limited,
-    /// Full network access: all HTTP methods are allowed, and HTTPS CONNECTs are tunneled without
-    /// MITM interception.
+    /// Full network access: all HTTP methods are allowed. HTTPS CONNECTs are tunneled directly
+    /// unless MITM is needed for host-specific inner-request hooks.
```
winston-openai: I don’t think this is true. `http_connect_proxy` only enters `mitm_tunnel` when `mode == NetworkMode::Limited`. In full mode, CONNECT is tunneled directly even if `mitm = true` or `mitm_hooks` are configured. So hooks can’t currently make full mode CONNECT use MITM

codex-rs/network-proxy/src/mitm_hook.rs
evawong-oai: Agreed. Body matching is still reserved and not enabled in this PR.

When we add it, we should gate it by content type first. JSON, form data, and GraphQL are the sensible initial set. The current config model only reserves the field so we do not commit to broader body parsing semantics yet.

codex-rs/config/src/permissions_toml.rs
evawong-oai: Not yet. This PR keeps the normal config layering behavior, so later layers replace earlier values.

The managed constraint check is tracked as AGE 248. That follow up must make trusted config reject user or project config that disables or replaces managed MITM hooks before dogfood or feature gate rollout.

I am keeping that out of this PR because it belongs with the PermissionProfile trusted layer rules, not the config model slice.

codex-rs/network-proxy/src/config.rs
evawong-oai: Good catch. I updated the comment to match current behavior.

Full mode currently tunnels CONNECT directly. MITM hooks do not make full mode enter MITM today. Any future full mode hook enforcement should be a separate behavior change.

codex-rs/network-proxy/src/mitm_hook.rs
winston-openai: fwiw I'm not fully convinced we'll even need it here. Maybe we will, but I could also see a world where if something matches a url we forward it on to another proxy that does the deeper introspection/matching/replacing that's a bit more modular. We can punt on that decision though for now

winston-openai (approved):
LGTM, though you may still want to consider splitting this PR up into smaller, more reviewable chunks

winston-openai:
it may also be worth keeping the config shape internal rather than exposing it in the `PermissionProfile` given it's all getting redone in #18240

evawong-oai:
Done.

#18868 now keeps the config shape internal. The public `PermissionProfile` config is in #18240, where the new named hook and action shape lives.

pakrym-oai (approved):

PR #18876, opened 2026-04-21
[rollout_trace] Add rollout trace crate

cassirer-openai:
## Summary

Adds the standalone `codex-rollout-trace` crate, which defines the raw trace event format, replay/reduction model, writer, and reducer logic for reconstructing model-visible conversation/runtime state from recorded rollout data.

The crate-level design is documented in [`codex-rs/rollout-trace/README.md`](https://github.com/openai/codex/blob/codex/rollout-trace-crate/codex-rs/rollout-trace/README.md).

## Stack

This is PR 1/5 in the rollout trace stack.

- [#18876](https://github.com/openai/codex/pull/18876): Add rollout trace crate
- [#18877](https://github.com/openai/codex/pull/18877): Record core session rollout traces
- [#18878](https://github.com/openai/codex/pull/18878): Trace tool and code-mode boundaries
- [#18879](https://github.com/openai/codex/pull/18879): Trace sessions and multi-agent edges
- [#18880](https://github.com/openai/codex/pull/18880): Add debug trace reduction command

## Review Notes

This PR intentionally does not wire tracing into live Codex execution. It establishes the data model and reducer contract first, with crate-local tests covering conversation reconstruction, compaction boundaries, tool/session edges, and code-cell lifecycle reduction. Later PRs emit into this model.

The README is the best entry point for reviewing the intended trace format and reduction semantics before diving into the reducer modules.

PR #18877, opened 2026-04-21
[rollout_trace] Record core session rollout traces

cassirer-openai:
## Summary

Wires rollout trace recording into `codex-core` session and turn execution. This records the core model request/response, compaction, and session lifecycle boundaries needed for replay without yet tracing every nested runtime/tool boundary.

## Stack

This is PR 2/5 in the rollout trace stack.

- [#18876](https://github.com/openai/codex/pull/18876): Add rollout trace crate
- [#18877](https://github.com/openai/codex/pull/18877): Record core session rollout traces
- [#18878](https://github.com/openai/codex/pull/18878): Trace tool and code-mode boundaries
- [#18879](https://github.com/openai/codex/pull/18879): Trace sessions and multi-agent edges
- [#18880](https://github.com/openai/codex/pull/18880): Add debug trace reduction command

## Review Notes

This layer is the first live integration point. The important review question is whether trace recording is isolated from normal session behavior: trace failures should not become user-visible execution failures, and recording should preserve the existing turn/session lifecycle semantics.

The PR depends on the reducer/data model from the first stack entry and only introduces the core recorder surface that later PRs use for richer runtime and relationship events.


codex-rs/core/src/rollout_trace.rs
```diff
@@ -0,0 +1,250 @@
+//! Opt-in producer for the rollout trace bundle.
```
jif-oai: I think this should move in `codex_rollout_trace` and have it depends on `codex-protocol`

codex-rs/core/src/rollout_trace.rs
```diff
@@ -0,0 +1,250 @@
...
+        InferenceTraceContext::enabled(
+            Arc::clone(&self.writer),
+            thread_id,
+            codex_turn_id,
+            model,
+            provider_name,
+        )
+    }
+
+    /// Builds remote-compaction trace context for one checkpoint.
```
jif-oai: Remote only?

codex-rs/core/src/thread_manager.rs
```diff
@@ -957,6 +957,7 @@ impl ThreadManagerState {
             metrics_service_name,
             inherited_shell_snapshot,
             inherited_exec_policy,
+            inherited_rollout_trace: None,
```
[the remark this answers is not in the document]
cassirer-openai: [codex] This is handled in the later session/multi-agent layer of the stack (#18879), where `ThreadManagerState::inherited_rollout_trace_for_source` inherits the parent recorder for `SubAgent(ThreadSpawn)` sessions. PR #18877 only introduces the core recorder plumbing, so this thread is no longer relevant for this PR layer.

cassirer-openai:
[codex] The review note about emitting `ThreadEnded` / `RolloutEnded` is handled in the later session/multi-agent layer of the stack (#18879), which adds `record_thread_ended` on shutdown and restores the lifecycle tests. PR #18877 is now scoped to core recorder plumbing and leaves rollout status running at that layer intentionally.

codex-rs/core/src/session/session.rs
```diff
@@ -378,6 +379,40 @@ impl Session {
...
+            agent_role: session_configuration.session_source.get_agent_role(),
+            session_source: session_configuration.session_source.clone(),
+            cwd: session_configuration.cwd.to_path_buf(),
+            rollout_path: rollout_path.clone(),
+            model: session_configuration.collaboration_mode.model().to_string(),
+            provider_name: config.model_provider_id.clone(),
+            approval_policy: session_configuration.approval_policy.value().to_string(),
+            sandbox_policy: format!("{:?}", session_configuration.sandbox_policy.get()),
+        };
+        let rollout_trace = if let Some(rollout_trace) = inherited_rollout_trace {
```
jif-oai: We should find a way to avoid all those `if let Some()`. We will have some everywhere in the code. An idea would be to make functions such as `record_thread_started` depend on an Option. Either directly or through an inner

codex-rs/core/src/session/turn.rs
```diff
@@ -1905,15 +1905,27 @@ async fn try_run_sampling_request(
         auth_mode = sess.services.auth_manager.auth_mode(),
         features = sess.features.enabled_features(),
     );
+    let inference_trace = sess.services.rollout_trace.as_ref().map_or_else(
```
jif-oai: same comment

codex-rs/core/src/session/turn.rs
```diff
@@ -1905,15 +1905,27 @@ async fn try_run_sampling_request(
...
+        codex_rollout_trace::InferenceTraceContext::disabled,
+        |trace| {
+            trace.inference_trace_context(
```
jif-oai: To be idiomatic we should move all the clone in the helper and send reference (an `into` which is even better) here

codex-rs/core/src/rollout_trace.rs
cassirer-openai: Yeah remote only (at least for now) because that is the only one I cared about for the current use case. We can do followups to expand support for local compaction as well.

codex-rs/core/src/rollout_trace.rs
cassirer-openai: Done.

codex-rs/core/src/session/turn.rs
cassirer-openai: I've changed it so that we now always pass a `RolloutTraceRecorder` and made it possible to disable it.

codex-rs/core/src/session/session.rs
cassirer-openai: I've changed it so that we now always pass a RolloutTraceRecorder and made it possible to disable it.

codex-rs/core/src/session/turn.rs
cassirer-openai: Done.

codex-rs/core/src/session/session.rs
```diff
@@ -365,6 +366,38 @@ impl Session {
         let rollout_path = rollout_recorder
             .as_ref()
             .map(|rec| rec.rollout_path().to_path_buf());
+        let trace_agent_path = session_configuration
```
jif-oai, line 369: Ultra nit: add a small comment separator before this as it's a big init function

codex-rs/core/src/client.rs
```diff
@@ -469,10 +469,18 @@ impl ModelClient {
...
             self.state.conversation_id.to_string(),
         )));
-        client
-            .compact_input(&payload, extra_headers)
-            .await
-            .map_err(map_api_error)
+        let trace_attempt = compaction_trace.start_attempt(&payload);
+        match client.compact_input(&payload, extra_headers).await {
+            Ok(output) => {
+                trace_attempt.record_completed(&output);
```
jif-oai: Arguable, we could have:
```
let result = client.compact_input(&payload, extra_headers).await;
trace_attempt.record(&result);
result
```
And we move the match in `.record()` to avoid the pollution here

codex-rs/core/src/client.rs
```diff
@@ -1329,17 +1350,35 @@ impl ModelClientSession {
...
+                inference_trace.start_attempt()
+            };
+            inference_trace_attempt.record_started(&ws_request);
             let stream_result = self.websocket_session.connection.as_ref().ok_or_else(|| {
                 map_api_error(ApiError::Stream(
                     "websocket connection is unavailable".to_string(),
                 ))
             })?;
-            let stream_result = stream_result
+            let stream_result = match stream_result
```
jif-oai: This part is odd... we should have something like this instead:
```
let stream_result = stream_result
    .stream_request(ws_request, self.websocket_session.connection_reused())
    .await
    .map_err(|err| {
        let err = map_api_error(err);
        inference_trace_attempt.record_failed(&err);
        err
    })?;
```

codex-rs/rollout-trace/src/compaction.rs
```diff
@@ -118,6 +130,40 @@ impl CompactionTraceContext {
         attempt.record_started(request);
         attempt
     }
+
+    /// Records the point where compacted history becomes the live thread history.
+    ///
+    /// The checkpoint belongs to the same semantic compaction lifecycle as the
+    /// compact endpoint attempts, so the context reuses its stable compaction ID.
+    pub fn record_installed(&self, checkpoint: &CompactionCheckpointTracePayload<'_>) {
```
jif-oai, line 138: I'm not sure if this function shouldn't return a Result or an empty Option or so... Let's not make it for now but we might slightly iterate on this design

jif-oai (approved):
Ok after my comments

codex-rs/core/src/client.rs
cassirer-openai: Done.

codex-rs/core/src/session/session.rs
cassirer-openai: Done.

codex-rs/core/src/client.rs
cassirer-openai: Done.

codex-rs/rollout-trace/src/compaction.rs
cassirer-openai: Ok I'll leave it for now but we can def revisit this.

PR #18878, opened 2026-04-21
[rollout_trace] Trace tool and code-mode boundaries

cassirer-openai:
## Summary

Extends rollout tracing across tool dispatch and code-mode runtime boundaries. This records canonical tool-call lifecycle events and links code-mode execution/wait operations back to the model-visible calls that caused them.

## Stack

This is PR 3/5 in the rollout trace stack.

- [#18876](https://github.com/openai/codex/pull/18876): Add rollout trace crate
- [#18877](https://github.com/openai/codex/pull/18877): Record core session rollout traces
- [#18878](https://github.com/openai/codex/pull/18878): Trace tool and code-mode boundaries
- [#18879](https://github.com/openai/codex/pull/18879): Trace sessions and multi-agent edges
- [#18880](https://github.com/openai/codex/pull/18880): Add debug trace reduction command

## Review Notes

This PR is about attribution. Reviewers should focus on whether direct tool calls, code-mode-originated tool calls, waits, outputs, and cancellation boundaries are recorded with enough source information for deterministic reduction without coupling the reducer to live runtime internals.

The stack remains valid after this layer: tool and code-mode traces reduce through the existing crate model, while the broader session and multi-agent relationships are added in the next PR.


codex-rs/code-mode/src/runtime/mod.rs
```diff
@@ -41,6 +48,33 @@ pub struct WaitRequest {
 }
 
 #[derive(Debug, PartialEq)]
+pub enum WaitResponse {
```
jif-oai: I don't have enough context to review but it looks sus to add this in code-mode

codex-rs/core/src/tools/registry.rs
```diff
@@ -210,6 +229,60 @@ pub struct ToolRegistry {
...
 }
 
+/// No-op capable trace handle for one resolved tool dispatch.
+///
+/// The registry has several early-return paths after a handler is selected:
+/// pre-use hooks, handler execution, after-use hooks, and result status all
+/// affect the trace lifecycle. Keeping the trace eligibility and event writes
+/// behind this helper makes those paths say what happened instead of repeating
+/// the Direct/CodeMode/JsRepl/first-class-object policy at each branch.
+struct DispatchTrace {
```
jif-oai: this should move

codex-rs/core/src/tools/registry.rs
```diff
@@ -210,6 +229,60 @@ pub struct ToolRegistry {
...
+        ) && !handler.uses_first_class_trace_object(invocation);
+
+        let recorder = should_trace
+            .then(|| invocation.session.services.rollout_trace.clone())
+            .flatten();
+        Self { recorder }
+    }
+
+    fn record_started(&self, invocation: &ToolInvocation) {
+        if let Some(recorder) = &self.recorder {
```
jif-oai: this is a better pattern regarding the Option

codex-rs/core/src/rollout_trace.rs
```diff
@@ -70,6 +87,42 @@ pub(crate) struct CompactionCheckpointTracePayload<'a> {
...
+    tool_name: &'a str,
+    tool_namespace: Option<&'a str>,
+    payload: serde_json::Value,
+}
+
+/// Raw response payload for dispatch-level tool trace events.
+#[derive(Serialize)]
+#[serde(rename_all = "snake_case", tag = "type")]
+enum DispatchedToolTraceResponse<'a> {
+    DirectResponse {
```
jif-oai: this looks like a duplicate of what is defined above

codex-rs/core/src/rollout_trace.rs
```diff
@@ -136,6 +198,209 @@ impl RolloutTraceRecorder {
...
+            },
+        );
+    }
+
+    /// Emits the first response returned by the public code-mode `exec` tool.
+    ///
+    /// A yielded response returns control to the model while the cell keeps
+    /// running. A terminal response is followed by `CodeCellEnded` so the
+    /// reducer can distinguish "first model-visible output" from runtime end.
+    pub(crate) fn record_code_cell_initial_response(
```
jif-oai: Instead of re-sending thread_id and codex_turn_id, we could make record_code_cell_started return something and we carry this object

Something like:
```
let mut code_cell = tr.record_code_cell_started(...);
code_cell.record_initial_response(...);
code_cell.ended(...);
```

this would also make this code more organized (and I would still move this code in another crate)

jif-oai (commented):
Same comments as for the previous PR

codex-rs/code-mode/src/runtime/mod.rs
cassirer-openai: There is an explicit tool call `wait` where you wait on a code cell to complete/write outputs. It is important that we capture this one so that we can track which code cell the main loop is being blocked on and what it yielded back to the model.

codex-rs/core/src/tools/registry.rs
cassirer-openai: It becomes very awkward to move this into the tracing module since it would either introduce heavy dependencies on `ToolInvocation` and `AnyToolHandler` or force use to define awkward generic types inside of the module. All the heavy lifting is already handled by the RolloutTraceRecorder. This class only does the minimal wiring.

codex-rs/core/src/rollout_trace.rs
cassirer-openai: Good call.

codex-rs/core/src/rollout_trace.rs
cassirer-openai: Good call. This was too close to the raw event schema when it lived in `codex-core`, and it made the dispatch payload look like a duplicate of `ToolCallEnded`. I moved this into `codex-rollout-trace` so core now only passes a typed dispatch result into the trace crate. The remaining private serialized payload is the raw payload behind `result_payload`, while `ToolCallEnded` stays the lifecycle envelope.

codex-rs/core/src/tools/registry.rs
cassirer-openai: I'm going to move as much as I can.

codex-rs/code-mode/src/runtime/mod.rs
```diff
@@ -41,6 +48,33 @@ pub struct WaitRequest {
...
+    /// Non-yielding responses from this variant are terminal lifecycle points
+    /// for the matching code cell.
+    LiveCell(RuntimeResponse),
+    /// The requested cell was not live, so the response is only the result of
+    /// the `wait` tool call. It must not be treated as a code-cell lifecycle
+    /// event because there is no cell to complete.
+    MissingCell(RuntimeResponse),
+}
+
+impl WaitOutcome {
```
jif-oai: The fact that we need this is strange 

If we can't avoid it, let's at least have an `impl From<>`... (but ideally we could avoid it with the previous comment)

codex-rs/code-mode/src/service.rs
```diff
@@ -20,15 +20,27 @@ use crate::runtime::RuntimeCommand;
...
+use crate::runtime::WaitOutcome;
 use crate::runtime::WaitRequest;
 use crate::runtime::spawn_runtime;
 
+/// Nested tool request emitted by one code-mode cell.
+///
+/// Code mode owns the per-cell runtime id. Hosts should preserve it for
+/// provenance/debugging, but should still assign their own runtime tool call id
+/// if their tool-call graph requires globally unique ids.
+pub struct CodeModeToolInvocation {
```
jif-oai: Just to be sure, this does not exist anywhere else? This looks a bit like a sub-structure that should already be available

codex-rs/code-mode/src/service.rs
```diff
@@ -76,24 +88,55 @@ impl CodeModeService {
...
+            .clone()
+            .unwrap_or_else(|| self.allocate_cell_id());
+        let initial_yield_time_ms = request.yield_time_ms.unwrap_or(DEFAULT_EXEC_YIELD_TIME_MS);
+        let (event_rx, runtime_tx, runtime_terminate_handle, control_rx, response_tx, response_rx) = {
+            let mut sessions = self.inner.sessions.lock().await;
+            if sessions.contains_key(&cell_id) {
+                return Err(format!("exec cell {cell_id} already exists"));
+            }
+
+            let (event_tx, event_rx) = mpsc::unbounded_channel();
```
jif-oai: Create those before acquiring the lock please

codex-rs/core/src/tools/code_mode/execute_handler.rs
```diff
@@ -30,12 +31,23 @@ impl CodeModeExecuteHandler {
             .code_mode_service
             .stored_values()
             .await;
+        // Allocate before starting V8 so the trace can create the parent
+        // CodeCell before model-authored JavaScript issues nested tool calls.
+        let runtime_cell_id = exec.session.services.code_mode_service.allocate_cell_id();
+        let code_cell_trace = exec.session.services.rollout_trace.start_code_cell_trace(
```
jif-oai, line 37: kudo, this looks much better IMO

codex-rs/code-mode/src/service.rs
```diff
@@ -76,24 +88,55 @@ impl CodeModeService {
...
-            SessionHandle {
-                control_tx: control_tx.clone(),
-                runtime_tx: runtime_tx.clone(),
-            },
-        );
+    pub async fn execute(&self, request: ExecuteRequest) -> Result<RuntimeResponse, String> {
+        let cell_id = request
+            .cell_id
+            .clone()
+            .unwrap_or_else(|| self.allocate_cell_id());
```
jif-oai: I'm not sure I understand the workflow here. Why would it sometimes be attributed here: https://github.com/openai/codex/pull/18878/changes#diff-1c3341d6c9c447798b1942b49fd3679abecbf2d3d406e91593e3bd2567eda2eaR36 and sometimes not?! (and so arrives as null)

codex-rs/core/src/tools/code_mode/wait_handler.rs
```diff
@@ -72,13 +72,43 @@ impl ToolHandler for CodeModeWaitHandler {
...
+                .map_err(FunctionCallError::RespondToModel)
             }
             _ => Err(FunctionCallError::RespondToModel(format!(
                 "{WAIT_TOOL_NAME} expects JSON arguments"
             ))),
         }
     }
 }
+
+fn runtime_cell_id(response: &codex_code_mode::RuntimeResponse) -> &str {
```
jif-oai: Can we inline this?

codex-rs/core/src/tools/registry.rs
jif-oai: I mean this should move in a dedicated file. This has nothing to do with the registry

jif-oai: And ideally most of them become some `impl From<>...`

codex-rs/rollout-trace/src/recorder.rs
```diff
@@ -147,6 +159,80 @@ impl RolloutTraceRecorder {
         });
     }
 
+    /// Emits a turn-start lifecycle event.
+    ///
+    /// Most production turn lifecycle wiring lives outside this PR layer, but
+    /// trace-focused integration tests need a small explicit hook so reducer
+    /// inputs remain valid without exercising the full session loop.
+    #[doc(hidden)]
```
jif-oai: Why doc hidden?

codex-rs/rollout-trace/src/tool_dispatch.rs
```diff
@@ -0,0 +1,462 @@
...
+    match tool_namespace {
+        Some(namespace) => format!("{namespace}.{tool_name}"),
+        None => tool_name.to_string(),
+    }
+}
+
+impl ToolDispatchPayload {
+    fn log_payload(&self) -> String {
+        match self {
+            ToolDispatchPayload::Function { arguments } => arguments.clone(),
```
jif-oai: Side comments but all the code contains tons of clone. For this kind of features with high throughput, this can have an impact on latency

codex-rs/core/src/tools/router.rs
```diff
@@ -281,7 +281,7 @@ impl ToolRouter {
 
         let direct_js_repl_call = tool_name.namespace.is_none()
             && matches!(tool_name.name.as_str(), "js_repl" | "js_repl_reset");
-        if source == ToolCallSource::Direct
+        if matches!(source, ToolCallSource::Direct)
```
jif-oai: You should match by reference

codex-rs/core/src/tools/registry.rs
```diff
@@ -339,6 +457,8 @@ impl ToolRegistry {
             return Err(FunctionCallError::Fatal(message));
         }
 
+        let dispatch_trace = DispatchTrace::new(&invocation);
```
jif-oai: This starts tracing after unsupported-tool and incompatible-payload early returns, so those failed model tool calls never get ToolCallStarted/Ended

codex-rs/rollout-trace/src/tool_dispatch.rs
```diff
@@ -0,0 +1,462 @@
...
+
+fn dispatched_tool_kind(tool_name: &str, payload: &ToolDispatchPayload) -> ToolCallKind {
+    if let ToolDispatchPayload::Mcp { server, tool, .. } = payload {
+        return ToolCallKind::Mcp {
+            server: server.clone(),
+            tool: tool.clone(),
+        };
+    }
+
+    match tool_name {
```
jif-oai, line 274: send_input it's not there. I guess on purpose?

codex-rs/core/src/tools/code_mode/execute_handler.rs
```diff
@@ -45,6 +57,13 @@ impl CodeModeExecuteHandler {
             })
             .await
             .map_err(FunctionCallError::RespondToModel)?;
+        // The initial response is the model-visible custom-tool return.
+        // Yielded cells keep running, so terminal lifecycle is only emitted
+        // here when the first response also ended the runtime.
+        code_cell_trace.record_initial_response(&response);
```
jif-oai: This records the raw runtime response before handle_runtime_response clean image detail or truncates output, so the trace payload can differ from what the model actually saw.

codex-rs/rollout-trace/src/tool_dispatch.rs
cassirer-openai: Yeah it is coming in the next PR.

codex-rs/code-mode/src/runtime/mod.rs
cassirer-openai: I have clarified the need for this one in the docstring and changed it to the `iml From<>`

codex-rs/code-mode/src/service.rs
cassirer-openai: It didn't really exist in a way that we can import it but I've reorganized the code a little bit so we don't have to repeat it like this.

codex-rs/code-mode/src/service.rs
cassirer-openai: You are right this was stupid. It was backwards compatible (before the cell id was just allocated within the call but now we need to know the ID up front when tracing) in a silly way. I've changed it now so that it isn't optional anymore.

codex-rs/code-mode/src/service.rs
cassirer-openai: Done!

codex-rs/core/src/tools/code_mode/wait_handler.rs
cassirer-openai: Done.

codex-rs/core/src/tools/code_mode/execute_handler.rs
cassirer-openai: Very good catch. This behavior is actually what we want though but the comment is misleading so I'm updating it.

codex-rs/rollout-trace/src/recorder.rs
cassirer-openai: Codex slop.

codex-rs/core/src/tools/router.rs
cassirer-openai: Done.

codex-rs/core/src/tools/registry.rs
cassirer-openai: Very good catch! Fixed!

codex-rs/core/src/tools/registry.rs
cassirer-openai: This is addressed in spirit by moving the dispatch adapter out of `registry.rs`; the remaining conversions are now isolated in `tool_dispatch_trace.rs`. I don’t think `impl From` is clearly better for all of them: the main invocation conversion is intentionally optional because `JsRepl` should not create a dispatch trace, and the result conversion depends on both source and output formatting. `ToolPayload -> ToolDispatchPayload` is the one plausible `From` candidate, but I’d prefer not to add that unless we see it materially improves the clone/ownership cleanup.

codex-rs/rollout-trace/src/tool_dispatch.rs
cassirer-openai: Yeah you are right. I've taken a pass over the code and cleaned/avoided all clones I could and also changed the API/implementation slightly so that the construction of larger objects (mainly raw requests) never happen when the tracer is disabled.

codex-rs/core/src/tools/registry_tests.rs
```diff
@@ -49,3 +70,304 @@ fn handler_looks_up_namespaced_aliases_explicitly() {
             .is_some_and(|handler| Arc::ptr_eq(handler, &namespaced_handler))
     );
 }
+
+#[tokio::test]
+async fn dispatch_lifecycle_trace_records_direct_and_code_mode_requesters() -> anyhow::Result<()> {
```
jif-oai: I would move the tests as well

jif-oai (approved):

codex-rs/core/src/tools/registry_tests.rs
cassirer-openai: Done!

PR #18882, opened 2026-04-21
[codex] Route live thread writes through ThreadStore

wiltzius-openai:
Begin migrating the thread write codepaths to ThreadStore.

This starts using ThreadStore inside of core session code, not only in the app server code.

Rework the interfaces around thread recording/persistence. We're left with the following:

* `ThreadManager`: owns the process-level registry of loaded threads and handles cross-thread orchestration: start, resume, fork, lookup, remove, and route ops to running CodexThreads.
* `CodexThread`: represents one loaded/running thread from the outside. It is the handle app-server and callers use to submit ops, inspect session metadata, and shut the thread down.
* `LiveThread`: session-owned persistence lifecycle handle for one active thread. Core session code uses it to append rollout items, materialize lazy persistence, flush, shutdown, discard init-failed writers, and load that thread’s persisted history.
* `ThreadStore`: storage backend abstraction. It answers “how are threads persisted, read, listed, updated, archived?” Local and remote implementations live behind this trait.
* `LocalThreadStore`: local ThreadStore implementation. It owns the file/sqlite-specific details and keeps RolloutRecorder as a local implementation detail.

This is a few too many Thread abstractions for my liking, but they do all represent different concepts / needs / layers.

Migration note: in places where the core code explicitly requires a path, rather than a thread ID, throw an error if we're running with a remote store. 

Cover the new local live-writer lifecycle with focused tests and preserve app-server thread-start behavior, including ephemeral pathless sessions.


wiltzius-openai:
@codex review

codex-rs/thread-store/src/remote/mod.rs
```diff
@@ -52,24 +54,38 @@ impl ThreadStore for RemoteThreadStore {
...
     }
 
-    async fn resume_thread_recorder(
-        &self,
-        _params: ResumeThreadRecorderParams,
-    ) -> ThreadStoreResult<Box<dyn ThreadRecorder>> {
-        Err(not_implemented("resume_thread_recorder"))
+    async fn resume_thread(&self, _params: ResumeThreadParams) -> ThreadStoreResult<()> {
+        Err(not_implemented("resume_thread"))
     }
```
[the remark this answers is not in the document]
wiltzius-openai: yeah the remote thread store in non-functional, not worrying too much about it yet

codex-rs/core/src/session/session.rs
```diff
@@ -282,32 +261,42 @@ impl Session {
...
+                                thread_id: conversation_id,
+                                forked_from_id,
+                                source: session_source,
+                                base_instructions: BaseInstructions {
+                                    text: session_configuration.base_instructions.clone(),
+                                },
+                                dynamic_tools: session_configuration.dynamic_tools.clone(),
+                                event_persistence_mode,
+                            })
+                            .await?;
```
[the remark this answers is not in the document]
wiltzius-openai: addressed

codex-rs/thread-store/src/local/mod.rs
```diff
@@ -62,29 +97,116 @@ impl ThreadStore for LocalThreadStore {
...
+    }
+
+    async fn shutdown_thread(&self, thread_id: ThreadId) -> ThreadStoreResult<()> {
+        let recorder = self
+            .live_recorders
+            .lock()
+            .await
+            .remove(&thread_id)
+            .ok_or(ThreadStoreError::ThreadNotFound { thread_id })?;
+        recorder.shutdown().await.map_err(thread_store_io_error)
```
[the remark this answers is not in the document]
wiltzius-openai: addressed

wiltzius-openai:
@codex review

wiltzius-openai:
@codex review

wiltzius-openai:
@codex review

wiltzius-openai:
@codex review

wiltzius-openai:
@codex review

codex-rs/tui/src/app/tests.rs
```diff
@@ -1537,6 +1537,7 @@ async fn update_memory_settings_updates_current_thread_memory_mode() -> Result<(
     let (mut app, _app_event_rx, _op_rx) = make_test_app_with_channels().await;
     let codex_home = tempdir()?;
     app.config.codex_home = codex_home.path().to_path_buf().abs();
+    app.config.sqlite_home = codex_home.path().to_path_buf();
```
pakrym-oai: ?

codex-rs/core/src/session/mod.rs
```diff
@@ -3112,11 +3116,20 @@ impl Session {
...
+        match self.try_current_rollout_path().await {
+            Ok(path) => path,
+            Err(err) => {
+                warn!("{err}");
+                None
+            }
+        }
+    }
+
+    pub(crate) async fn try_current_rollout_path(&self) -> anyhow::Result<Option<PathBuf>> {
```
pakrym-oai: current_rollout_path vs try_current_rollout_path seem duplicative

codex-rs/core/src/live_thread.rs
```diff
@@ -0,0 +1,176 @@
...
+use codex_thread_store::UpdateThreadMetadataParams;
+use tracing::warn;
+
+/// Session-owned handle for the active thread's persistence lifecycle.
+///
+/// `LiveThread` keeps lifecycle decisions in core while delegating storage details to
+/// [`ThreadStore`]. Local stores may use a rollout file internally and remote stores may use a
+/// service, but session code should only need this handle for the active thread.
+#[derive(Clone)]
+pub(crate) struct LiveThread {
```
pakrym-oai: nit:should this live next to ThreadStore?

codex-rs/core/src/thread_manager.rs
```diff
@@ -260,6 +272,7 @@ impl ThreadManager {
             restriction_product,
         ));
         let skills_watcher = build_skills_watcher(Arc::clone(&skills_manager));
+        let thread_store = configured_thread_store(config);
```
pakrym-oai: nit: should this be selected at even higher level?

codex-rs/tui/src/app/tests.rs
wiltzius-openai: this is a small test harness fix, previously this harness didn't set up the sqlite home at all which caused some failures in the new codepaths.

codex-rs/core/src/live_thread.rs
wiltzius-openai: do you mean in the ThreadStore crate, or its own crate as a "peer" to ThreadStore?

codex-rs/core/src/session/session.rs
```diff
@@ -325,37 +305,68 @@ impl Session {
...
             }
         }
         .instrument(info_span!(
-            "session_init.rollout",
-            otel.name = "session_init.rollout",
+            "session_init.thread_persistence",
+            otel.name = "session_init.thread_persistence",
+            session_init.ephemeral = config.ephemeral,
+        ));
+        let state_db_fut = async {
```
pakrym-oai, line 355: worth having on the abstraction? at least for now?

codex-rs/core/src/session/mod.rs
wiltzius-openai: agreed, will fix

codex-rs/core/src/session/session.rs
pakrym-oai: downcast doesn't make it much better

wiltzius-openai: It went in and out a couple times as I revised.... agreed the downcast isn't much better but I wanted to push the failure / issue as early in the code flow as possible. I.e. make sure its clear that accessing state_db_fut in the session.rs code is not something the code out here should be doing; otherwise it's easy to just keep accessing it...

codex-rs/core/src/live_thread.rs
wiltzius-openai: moving it into ThreadStore

codex-rs/tui/src/app/tests.rs
wiltzius-openai: was able to remove

codex-rs/core/src/session/session.rs
```diff
@@ -397,533 +408,557 @@ impl Session {
...
-                    otel.name = "session_init.network_proxy",
-                    session_init.managed_network_requirements_enabled =
-                        managed_network_requirements_enabled,
-                ))
-                .await?;
-                (Some(network_proxy), Some(session_network_proxy))
-            } else {
-                (None, None)
+            });
+            let live_thread_cell = OnceCell::new();
```
pakrym-oai: Why do we need the OnceCell here? Doesn't Option work?

pakrym-oai (approved):

codex-rs/core/src/session/session.rs
wiltzius-openai: don't think we do, changed

codex-rs/core/src/session/session.rs
wiltzius-openai: to make it more obvious, I removed the default case that was still instantiating the state db. If you're running a local thread store you get one, otherwise (ephemeral or remote thread store) it's just None / we'll burn down what fails

PR #18888, opened 2026-04-21
hooks: emit Bash PostToolUse when exec_command completes via write_stdin

### LINKED ISSUE #16246: Hooks: PostToolUse is missing for tools that complete via exec session / polling path
## Summary

Codex emits `PreToolUse` for shell tools started via `exec_command`, but in some cases it never emits the matching `PostToolUse`.

This appears to happen specifically when the command does not complete on the initial `exec_command` call and instead completes later through the long-running session / polling path.

## Expected behavior

Every tool call that emits `PreToolUse` should emit exactly one matching `PostToolUse` when the tool finishes, regardless of whether it:

- completes immediately on the initial tool call, or
- completes later through the session continuation / polling flow

## Actual behavior

`PostToolUse` is emitted for commands that complete immediately, but is missing for commands that:

1. start via `exec_command`
2. return a running session
3. later complete through the session / polling path

## Evidence

In one reproduced case, a single turn produced:

- `7` `PreToolUse`
- `2` `PostToolUse`

The transcript for that same turn showed that the missing tool calls did in fact complete successfully later, including exit codes and output, but no `PostToolUse` event was emitted for them.

The missing cases were long-running shell commands that first returned a session and later completed through polling. The two cases that did get `PostToolUse` were commands that completed directly on the initial `exec_command` path.

## Impact

Consumers that rely on hooks to model tool lifecycle get incomplete traces:

- tools appear to start but never finish
- duration and status are missing
- downstream systems cannot reliably pair pre/post tool events

## Root-cause hypothesis

The hook integration appears to emit `PostToolUse` only for the immediate completion path, but not for tool completion that happens later through the exec session continuation / polling path.

## Suggested fix

Emit `PostToolUse` when the tool actually completes, regardless of whether completion happens:

- during the initial `exec_command`, or
- during later session polling / continuation

The emitted event should use the original tool call ID so consumers can pair it with the earlier `PreToolUse`.

eternal-openai:
Fixes #16246.

## Why

`exec_command` already emits `PreToolUse`, but long-running unified exec commands that finish on a later `write_stdin` poll could miss the matching `PostToolUse`. That left the Bash hook lifecycle inconsistent, broke expectations around `tool_use_id` and `tool_input.command`, and meant `PostToolUse` block/replacement feedback could fail to replace the final session output before it reached model context.

This keeps the fix scoped to the `exec_command` / `write_stdin` lifecycle. Broader non-Bash hook expansion is still out of scope here and remains tracked separately in #16732.

## What changed

- Compute and store `PostToolUsePayload` while handlers still have access to their concrete output type, and carry `tool_use_id` through that payload.
- Preserve the original hook-facing `exec_command` string through unified exec state (`ExecCommandRequest`, `ProcessEntry`, `PreparedProcessHandles`, and `ExecCommandToolOutput`) via `hook_command`, and remove the now-unused `session_command` output metadata.
- Emit exactly one Bash `PostToolUse` for long-running `exec_command` sessions when a later `write_stdin` poll observes final completion, using the original `exec_command` call id and hook-facing command.
- Keep one-shot `exec_command` behavior aligned with the same payload construction, including interactive completions that return a final result directly.
- Apply `PostToolUse` block/replacement feedback before the final `write_stdin` completion output is sent back to the model.
- Keep `write_stdin` itself out of `PreToolUse` matching so it continues to act as transport/polling for the original Bash tool call.
- Restore plain matcher behavior for tool-name matchers such as `Bash` and `Edit|Write`, while still treating patterns with regex characters (for example `mcp__.*`) as regexes.
- Add unit coverage for unified exec payload construction and parallel session separation, plus a core integration regression that verifies a blocked `PostToolUse` replaces the final `write_stdin` output in model context.

## Testing

- `cargo test -p codex-hooks`
- `cargo test -p codex-core post_tool_use_payload`
- `cargo test -p codex-core post_tool_use_blocks_when_exec_session_completes_via_write_stdin`


codex-rs/hooks/src/events/common.rs
```diff
@@ -109,7 +109,7 @@ pub(crate) fn matcher_pattern_for_event(
 }
 
 pub(crate) fn validate_matcher_pattern(matcher: &str) -> Result<(), regex::Error> {
-    if is_match_all_matcher(matcher) {
+    if is_match_all_matcher(matcher) || is_exact_matcher(matcher) {
```
abhinav-oai, line 112: I think I'm suffering from deja vu I thought this exact matcher was already here 😂

eternal-openai: Yknow, same, I even had codex research if there was some code reuse issue here. But apparently no

abhinav-oai: nvm realized it was on my own feature branch

codex-rs/core/src/tools/handlers/unified_exec.rs
```diff
@@ -143,20 +143,22 @@ impl ToolHandler for UnifiedExecHandler {
...
         };
 
-        let args = parse_arguments::<ExecCommandArgs>(arguments).ok()?;
-        if args.tty {
-            return None;
-        }
-
-        let tool_response = result.post_tool_use_response(call_id, payload)?;
+        let command = result.hook_command.clone()?;
+        let tool_use_id = if result.event_call_id.is_empty() {
```
abhinav-oai, line 157: [question] when would this be empty?

and isn't it the same as `call_id`?

abhinav-oai (approved):
LGTM 🚢

codex-rs/core/src/tools/handlers/unified_exec.rs
eternal-openai: Yeah, good ask - I think you're right that this should never happen. It’s only there as a defensive fallback because ExecCommandToolOutput has one synthetic constructor path (intercept_apply_patch) that doesn’t participate in the normal unified-exec lifecycle and leaves event_call_id empty. That path also never produces a PostToolUse payload, but I'll leave it in just in case

sayan-oai (approved):

PR #18892, opened 2026-04-21
Add remote thread config loader protos

rasmusrygaard:
## Why

Thread-scoped config needs a stable boundary between the app/session owner and the config stack. Instead of having call sites manually copy thread config fields into individual overrides, this adds the proto and Rust plumbing needed for a `ThreadConfigLoader` implementation to return typed sources that can be translated into ordinary config layer entries.

Keeping the remote payload typed also makes precedence easier to reason about: session-owned thread config maps back to the existing session config source, while user-owned thread config is represented separately without introducing a new config-layer source until it has TOML-backed fields.

## What changed

- Added the `codex.thread_config.v1` protobuf service and generated Rust module for loading thread config sources.
- Added `RemoteThreadConfigLoader`, which calls the gRPC service, parses `SessionThreadConfig` / `UserThreadConfig`, and validates provider fields such as `wire_api`, auth timeout, and absolute auth cwd.
- Added proto generation tooling under `config/scripts/generate-proto.sh` and `config/examples/generate-proto.rs`.
- Added `ThreadConfigLoader::load_config_layers`, plus static/no-op loader helpers, so tests and callers can use the same typed loader interface while config-layer translation stays centralized.

## Verification

- `cargo test -p codex-config thread_config`


codex-rs/config/examples/generate-proto.rs
```diff
@@ -0,0 +1,19 @@
+use std::path::PathBuf;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
wiltzius-openai, line 3: would love a better place to put the proto generation than `examples` but I couldn't figure out a better way to do it with the repo build rules

wiltzius-openai (commented):
seems reasonable

I think there's no local implementation yet? I opted to try to keep the local implementation of the interface ahead of the remote implementation, mainly because as I plumbed in the local version I found various details of the interface needed to change. Not a requirement just a thought about ordering.

rasmusrygaard:
Good flag! Local implementation is a no-op because since we don't inject any new config. I have the wiring in a stacked branch on top of this, but it just injects the NoOp loader

pakrym-oai (approved):

PR #18893, opened 2026-04-21
codex: support hooks in config.toml and requirements.toml

eternal-openai:
## Summary

Support the existing hooks schema in inline TOML so hooks can be configured from both `config.toml` and enterprise-managed `requirements.toml` without requiring a separate `hooks.json` payload.

This gives enterprise admins a way to ship managed hook policy through the existing requirements channel while still leaving script delivery to MDM or other device-management tooling, and it keeps `hooks.json` working unchanged for existing users.

This also lays the groundwork for follow-on managed filtering work such as #15937, while continuing to respect project trust gating from #14718. It does **not** implement `allow_managed_hooks_only` itself.

NOTE: yes, it's a bit unfortunate that the toml isn't formatted as closely as normal to our default styling. This is because we're trying to stay compatible with the spec for plugins/hooks that we'll need to support & the main usecase here is embedding into requirements.toml

## What changed

- moved the shared hook serde model out of `codex-rs/hooks` into `codex-rs/config` so the same schema can power `hooks.json`, inline `config.toml` hooks, and managed `requirements.toml` hooks
- added `hooks` support to both `ConfigToml` and `ConfigRequirementsToml`, including requirements-side `managed_dir` / `windows_managed_dir`
- treated requirements-managed hooks as one constrained value via `Constrained`, so managed hook policy is merged atomically and cannot drift across requirement sources
- updated hook discovery to load requirements-managed hooks first, then per-layer `hooks.json`, then per-layer inline TOML hooks, with a warning when a single layer defines both representations
- threaded managed hook metadata through discovered handlers and exposed requirements hooks in app-server responses, generated schemas, and `/debug-config`
- added hook/config coverage in `codex-rs/config`, `codex-rs/hooks`, `codex-rs/core/src/config_loader/tests.rs`, and `codex-rs/core/tests/suite/hooks.rs`

## Testing

- `cargo test -p codex-hooks`
- `cargo test -p codex-app-server config_api`

## Documentation

Companion updates are needed in the developers website repo for:

- the hooks guide
- the config reference, sample, basic, and advanced pages
- the enterprise managed configuration guide


codex-rs/hooks/src/engine/discovery.rs
```diff
@@ -1,22 +1,37 @@
...
 
 pub(crate) struct DiscoveryResult {
     pub handlers: Vec<ConfiguredHandler>,
     pub warnings: Vec<String>,
 }
 
+#[derive(Clone, Copy)]
+struct HookHandlerSource<'a> {
+    path: &'a AbsolutePathBuf,
+    is_managed: bool,
```
abhinav-oai, line 31: do we use `is_managed` anywhere?

bolinfest (commented):
Posting the Codex review findings requested by mbolin.

codex-rs/hooks/src/engine/discovery.rs
```diff
@@ -29,152 +44,332 @@ pub(crate) fn discover_handlers(config_layer_stack: Option<&ConfigLayerStack>) -
...
+        managed_hooks_source_path(managed_hooks.get(), managed_hooks.source.as_ref(), warnings)
+    else {
+        return;
+    };
+    append_hook_events(
+        handlers,
+        warnings,
+        display_order,
+        HookHandlerSource {
+            path: &source_path,
```
bolinfest, line 125: [P1] Managed hooks resolve relative commands in the workspace

Requirements-managed hooks validate `managed_dir`, but command execution still sets the child cwd to the active workspace and passes the raw shell command through unchanged. A managed policy using a relative command such as `python3 pre.py` will run whatever `pre.py` exists in the workspace instead of the MDM-managed directory, so an untrusted project can hijack or bypass the managed hook. Reject relative commands for managed handlers or run managed hooks from `managed_dir`, and add a regression test with a relative managed command plus a conflicting workspace script.

Anchored here because `command_runner.rs` is not part of this PR diff; the execution issue is at `codex-rs/hooks/src/engine/command_runner.rs:33-35`.

codex-rs/hooks/src/engine/mod.rs
```diff
@@ -170,3 +170,334 @@ impl ClaudeHooksEngine {
...
+                ..ConfigRequirements::default()
+            },
+            ConfigRequirementsToml {
+                hooks: Some(managed_hooks),
+                ..ConfigRequirementsToml::default()
+            },
+        )
+        .expect("config layer stack");
+
+        let engine = ClaudeHooksEngine::new(
```
bolinfest: [P2] Requirements hooks can be silently disabled

Sessions only enable the hook engine from `config.features.enabled(Feature::CodexHooks)`. Since `CodexHooks` defaults off, a `requirements.toml` or cloud requirements payload containing hooks alone produces `managed_hooks` but then `ClaudeHooksEngine` returns before discovery. The managed policy silently becomes a no-op unless admins also know to pin the `codex_hooks` feature. Either imply enablement when managed hooks are configured or reject/report this configuration, and add a core integration test that loads requirements hooks through a real session.

This test constructs the engine with `enabled = true`, so it does not cover the product session path at `codex-rs/core/src/session/session.rs:643-646`.

codex-rs/config/src/hook_config.rs
```diff
@@ -0,0 +1,202 @@
...
+    pub hooks: Vec<HookHandlerConfig>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+#[serde(tag = "type")]
+pub enum HookHandlerConfig {
+    #[serde(rename = "command")]
+    Command {
+        command: String,
+        #[serde(default, rename = "timeout", alias = "timeoutSec")]
```
bolinfest: [P3] Schema exposes `timeout` instead of `timeoutSec`

The new config schema is generated from `serde(rename = "timeout")`, so schema consumers see `timeout` as the property even though the added TOML tests and app-server API use `timeoutSec` and deserialization accepts it as the existing hook shape. This will make valid configs look invalid in editor/schema tooling. Make `timeoutSec` the canonical serialized/schema name with `timeout` as an alias, or provide a custom schema that includes both.

codex-rs/hooks/src/engine/discovery.rs
eternal-openai: this will be used later when we need to implement allow_managed_hooks_only -- for now this is a placeholder

codex-rs/hooks/src/engine/mod.rs
eternal-openai: Should be okay:

 - we are making hooks default on with this release
 - requirements.toml can also enable hooks if needed

codex-rs/hooks/src/engine/discovery.rs
eternal-openai: Thanks! Okay so:

- it's expected that the hook runs in the cwd of wherever codex is, that part is normal
- it's true that the system resolves hooks from the project dir too

So the potential problem here is like this:

1. admin ships a managed hook
2. the managed hook command is relative
3. the workspace contains a shadow script with that name

Therefore my judgement here is -- valid, could use hardening, but probably not a P1. Appears that the spec "recommends" using absolute paths for admins, but does not enforce it. Agree that we can do better

codex-rs/config/src/hook_config.rs
eternal-openai: Yeah, `timeout` is the correct schema, though timeoutSec is currently a valid alternative. Probably the solution is that we should remove timeoutSec entirely and just stick to timeout

codex-rs/app-server/src/message_processor/tracing_tests.rs
```diff
@@ -291,6 +292,28 @@ fn build_test_processor(
...
 
+fn run_current_thread_test_with_stack<F>(name: &str, future: F) -> Result<()>
+where
+    F: Future<Output = Result<()>> + Send + 'static,
+{
+    const TEST_STACK_SIZE_BYTES: usize = 4 * 1024 * 1024;
+
+    let handle = std::thread::Builder::new()
+        .name(name.to_string())
+        .stack_size(TEST_STACK_SIZE_BYTES)
```
bolinfest, line 303: I would argue this is a bad code smell. Please read https://github.com/openai/codex/pull/13429.

codex-rs/app-server/src/message_processor/tracing_tests.rs
```diff
@@ -291,6 +292,28 @@ fn build_test_processor(
...
+            let runtime = tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()?;
+            runtime.block_on(Box::pin(future))
+        })?;
+
+    match handle.join() {
+        Ok(result) => result,
+        Err(_) => Err(anyhow::anyhow!("{name} thread panicked")),
+    }
```
bolinfest, line 314: `map_err()` is a bit more canonical and I would include `err` using `?`, if necessary

```suggestion
    handle.join().map_err(|err| Err(anyhow::anyhow!("{name} thread panicked: {err?}"))
```

codex-rs/config/src/config_requirements.rs
```diff
@@ -2216,6 +2266,124 @@ allowed_approvals_reviewers = ["user"]
         Ok(())
     }
 
+    #[test]
+    fn deserialize_managed_hooks_requirements() -> Result<()> {
+        let toml_str = r#"
+            managed_dir = "/enterprise/hooks"
```
bolinfest: FYI, I would left-align the contents of this and the other strings so it's easier to copy-paste into a real `.toml` file for testing.

codex-rs/config/src/hook_config.rs
```diff
@@ -0,0 +1,202 @@
...
+
+impl HookEventsToml {
+    pub fn is_empty(&self) -> bool {
+        self.pre_tool_use.is_empty()
+            && self.permission_request.is_empty()
+            && self.post_tool_use.is_empty()
+            && self.session_start.is_empty()
+            && self.user_prompt_submit.is_empty()
+            && self.stop.is_empty()
+    }
```
bolinfest, line 47: I think this is a slightly better way to do this:

```suggestion
    pub fn is_empty(&self) -> bool {
        let Self = {
            pre_tool_use,
            permission_request,
            post_tool_use,
            session_start,
            user_prompt_submit,
            stop,
        } = self;
        pre_tool_use.is_empty()
            && permission_request.is_empty()
            && post_tool_use.is_empty()
            && session_start.is_empty()
            && user_prompt_submit.is_empty()
            && stop.is_empty()
    }
```

The reason is that if another field is added to `HookEventsToml`, you want to be sure to update this function. When you write it this way, if you add a new field, this function will fail to compile because not all the fields will be listed in the destructuring. This ensures you handle the new field in this function.

codex-rs/config/src/hook_config.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    pub fn handler_count(&self) -> usize {
+        [
```
bolinfest, line 58: Using the same destructuring trick as above also ensures all fields are handled.

codex-rs/config/src/hook_config.rs
```diff
@@ -0,0 +1,202 @@
...
+            &self.stop,
+        ]
+        .into_iter()
+        .flatten()
+        .map(|group| group.hooks.len())
+        .sum()
+    }
+
+    pub fn into_matcher_groups(self) -> [(HookEventName, Vec<MatcherGroup>); 6] {
+        [
```
bolinfest, line 73: Why not return a `HashMap` that uses `HookEventName` as a key?

codex-rs/config/src/hook_config.rs
```diff
@@ -0,0 +1,202 @@
...
+pub struct ManagedHooksRequirementsToml {
+    pub managed_dir: Option<PathBuf>,
+    pub windows_managed_dir: Option<PathBuf>,
+    #[serde(flatten)]
+    pub hooks: HookEventsToml,
+}
+
+impl ManagedHooksRequirementsToml {
+    pub fn is_empty(&self) -> bool {
+        self.managed_dir.is_none() && self.windows_managed_dir.is_none() && self.hooks.is_empty()
```
bolinfest: destructure `self` here too

codex-rs/config/src/hook_config.rs
```diff
@@ -0,0 +1,202 @@
...
+                [[PreToolUse.hooks]]
+                type = "command"
+                command = "python3 /tmp/pre.py"
+                timeoutSec = 10
+                statusMessage = "checking"
+            "#,
+        )
+        .expect("hook events TOML should deserialize");
+
+        assert_eq!(parsed.pre_tool_use.len(), 1);
```
bolinfest: I would favor doing one `assert_eq!` on `parsed.pre_tool_use` (or on `parsed` itself).

codex-rs/config/src/hook_config.rs
eternal-openai: I think the idea is to preserve ordering

codex-rs/config/src/hook_config.rs
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        #[cfg(not(windows))]
+        {
+            self.managed_dir.as_deref()
+        }
+    }
+}
+
+#[cfg(test)]
```
bolinfest, line 146: Prefer putting these in `hooks_tests.rs` and then referencing it here. Look how we do it in `codex-core`.

codex-rs/core/tests/suite/hooks.rs
```diff
@@ -1726,6 +1793,164 @@ async fn pre_tool_use_blocks_shell_command_before_execution() -> Result<()> {
...
+        "shell command output should still reach the model",
+    );
+
+    let json_hook_inputs = read_pre_tool_use_hook_inputs(test.codex_home_path())?;
+    let toml_hook_inputs = read_hook_inputs_from_log(
+        test.codex_home_path()
+            .join("pre_tool_use_toml_hook_log.jsonl")
+            .as_path(),
+    )?;
+    assert_eq!(json_hook_inputs.len(), 1);
```
bolinfest: Again, if it's not too hard, `assert_eq!(vec![ ... ], toml_hook_inputs)` so you assert everything at once and pretty_assertions is helpful in a failure

codex-rs/hooks/src/engine/discovery.rs
bolinfest: Interesting...so you can't ship a hook like `python3 ./myfile.py` if you want to "include" `myfile.py` as part of the hook?

bolinfest: For in-repo hooks, that maybe seems reasonable, but not from a plugin?

codex-rs/hooks/src/engine/discovery.rs
```diff
@@ -29,152 +44,332 @@ pub(crate) fn discover_handlers(config_layer_stack: Option<&ConfigLayerStack>) -
...
+    }
+
+    AbsolutePathBuf::from_absolute_path(source_path)
+        .inspect_err(|err| {
+            warnings.push(format!(
+                "skipping managed hooks from {source}: could not normalize managed hook directory {}: {err}",
+                source_path.display()
+            ));
+        })
+        .ok()
```
bolinfest: If you want to elide `return`:

```suggestion
    let source = requirement_source
        .map(ToString::to_string)
        .unwrap_or_else(|| "managed requirements".to_string());
    let Some(source_path) = managed_hooks.managed_dir_for_current_platform() else {
        warnings.push(format!(
            "skipping managed hooks from {source}: no managed hook directory is configured for this platform"
        ));
        return None;
    };
    if !source_path.is_absolute() {
        warnings.push(format!(
            "skipping managed hooks from {source}: managed hook directory {} is not absolute",
            source_path.display()
        ));
        None
    } else if !source_path.exists() {
        warnings.push(format!(
            "skipping managed hooks from {source}: managed hook directory {} does not exist",
            source_path.display()
        ));
        None
    } else if !source_path.is_dir() {
        warnings.push(format!(
            "skipping managed hooks from {source}: managed hook directory {} is not a directory",
            source_path.display()
        ));
        None
    } else {

    AbsolutePathBuf::from_absolute_path(source_path)
        .inspect_err(|err| {
            warnings.push(format!(
                "skipping managed hooks from {source}: could not normalize managed hook directory {}: {err}",
                source_path.display()
            ));
        })
        .ok()
    }        
```

codex-rs/hooks/src/engine/discovery.rs
```diff
@@ -29,152 +44,332 @@ pub(crate) fn discover_handlers(config_layer_stack: Option<&ConfigLayerStack>) -
...
+fn load_hooks_json(
+    config_folder: Option<&Path>,
+    warnings: &mut Vec<String>,
+) -> Option<(AbsolutePathBuf, HookEventsToml)> {
+    let source_path = config_folder?.join("hooks.json");
+    if !source_path.as_path().is_file() {
+        return None;
+    }
+
+    let contents = match fs::read_to_string(source_path.as_path()) {
```
bolinfest, line 187: @starr-openai @pakrym does this have to go through the `Environment` file reader?

codex-rs/config/src/hook_config.rs
eternal-openai: fixed - removed timeoutSec from the config-side hook schema and normalized the config examples/tests on timeout

codex-rs/config/src/config_requirements.rs
eternal-openai: fixed

codex-rs/config/src/hook_config.rs
eternal-openai: fixed

codex-rs/config/src/hook_config.rs
eternal-openai: fixed

codex-rs/config/src/hook_config.rs
eternal-openai: fixed

codex-rs/config/src/hook_config.rs
eternal-openai: fixed

codex-rs/config/src/hook_config.rs
eternal-openai: fixed

codex-rs/core/tests/suite/hooks.rs
eternal-openai: fixed

codex-rs/hooks/src/engine/mod.rs
```diff
@@ -170,3 +170,334 @@ impl ClaudeHooksEngine {
         crate::events::stop::run(&self.handlers, &self.shell, request).await
     }
 }
+
+#[cfg(test)]
```
bolinfest, line 174: Let's put this in `mod_tests.rs` or something?

codex-rs/hooks/src/engine/mod.rs
```diff
@@ -170,3 +170,334 @@ impl ClaudeHooksEngine {
...
+    use super::ClaudeHooksEngine;
+    use super::CommandShell;
+    use crate::events::pre_tool_use::PreToolUseRequest;
+
+    fn cwd() -> AbsolutePathBuf {
+        AbsolutePathBuf::current_dir().expect("current dir")
+    }
+
+    fn managed_hooks_for_current_platform<P: Into<std::path::PathBuf>>(
+        managed_dir: P,
```
bolinfest: You can drop the `<P: Into<std::path::PathBuf>>` this way:

```suggestion
        managed_dir: impl AsRef<Path>,
```

codex-rs/hooks/src/engine/discovery.rs
eternal-openai: Generally, the spec provides magic variables to this command that allows you to reference stuff like the plugin root or the project root. We don't have this implemented currently, so in the docs I have been recommending people to do something like `/usr/bin/python3 \"$(git rev-parse --show-toplevel)/.codex/hooks/pre_tool_use_policy.py\"` to find the correct path for the project. I'd like to get these variables implemented. Managed hooks should be absolute paths generally as discussed

codex-rs/hooks/src/engine/mod.rs
eternal-openai: fixed

codex-rs/hooks/src/engine/mod.rs
eternal-openai: fixed

codex-rs/app-server/src/message_processor/tracing_tests.rs
eternal-openai: resolved by your windows patch, thanks!

codex-rs/app-server/src/message_processor/tracing_tests.rs
eternal-openai: resolved by your windows patch, thanks!

codex-rs/hooks/src/engine/discovery.rs
eternal-openai: fixed

bolinfest (approved):

PR #18897, opened 2026-04-21
Add sticky environment API and thread state

starr-openai:
## Summary
- add sticky environment selections to app-server v2 thread/start and turn/start request flow
- carry thread-level selections through core session/thread state
- add app-server coverage for sticky selections and turn overrides

## Stack
1. This PR: API and thread persistence
2. #18898: config.toml named environment loading
3. #18899: downstream tool/runtime consumers

## Validation
- Not run locally; split only.


codex-rs/core/src/session/mod.rs
```diff
@@ -401,6 +402,7 @@ pub(crate) struct CodexSpawnArgs {
     pub(crate) inherited_exec_policy: Option<Arc<ExecPolicyManager>>,
     pub(crate) user_shell_override: Option<shell::Shell>,
     pub(crate) parent_trace: Option<W3cTraceContext>,
+    pub(crate) environments: Option<Vec<TurnEnvironmentSelection>>,
```
pakrym-oai: should this be non optional?

codex-rs/core/src/session/agent_task_lifecycle.rs
```diff
@@ -23,7 +23,9 @@ impl Session {
         rollout_items: &[RolloutItem],
     ) -> Option<Option<SessionAgentTask>> {
         rollout_items.iter().rev().find_map(|item| match item {
-            RolloutItem::SessionState(update) => Some(update.agent_task.clone()),
+            RolloutItem::SessionState(update) if !update.preserve_agent_task => {
```
pakrym-oai: SessionState should be deleted at this point and we shouldn't use it.

starr-openai: ah good to know - because we prefer to keep everything per turn?

codex-rs/core/src/session/mod.rs
starr-openai: ^ We need to keep these optional until we come up with a different signal for "no env enabled", because existing callers aren't setting the environment list.

let me see if we can move this up closer to the app-server protocol handling though - we could probably coerce None => {"local"} and make this required

codex-rs/core/src/session/turn_context.rs
```diff
@@ -559,52 +565,71 @@ impl Session {
...
+
+    pub(super) fn validate_turn_environment_selections(
+        environment_manager: &EnvironmentManager,
+        environments: &[TurnEnvironmentSelection],
+    ) -> CodexResult<()> {
+        if environments.is_empty() {
+            return Ok(());
+        }
+
+        if environment_manager.default_environment().is_none() {
```
pakrym-oai: Is this correct? Why do we validate explicit list against env var?

pakrym-oai: default_environment().is_none doesn't mean never enable environments, it's a shortcut for `environments:[]` in my understanding.

codex-rs/core/src/session/turn_context.rs
```diff
@@ -559,52 +565,71 @@ impl Session {
...
-        for environment_selection in environment_selections {
+        environments: Vec<TurnEnvironmentSelection>,
+    ) -> CodexResult<Vec<TurnEnvironment>> {
+        Self::validate_turn_environment_selections(
+            &self.services.environment_manager,
+            &environments,
+        )?;
+
+        let mut turn_environments = Vec::with_capacity(environments.len());
+        for selected_environment in environments {
```
pakrym-oai, line 579: validate_turn_environment_selections also checks for unknown names.

codex-rs/core/src/thread_manager.rs
```diff
@@ -1000,6 +1037,20 @@ impl ThreadManagerState {
     }
 }
 
+fn default_thread_environments(
+    environment_manager: &EnvironmentManager,
+    cwd: &AbsolutePathBuf,
+) -> Vec<TurnEnvironmentSelection> {
+    if environment_manager.default_environment().is_none() {
```
pakrym-oai: why don't we read environment_manager.default_environment() for anything else?

codex-rs/core/src/session/turn_context.rs
```diff
@@ -475,26 +487,18 @@ impl Session {
         &self,
         sub_id: String,
         updates: SessionSettingsUpdate,
-        environment_selections: Option<Vec<TurnEnvironmentSelection>>,
+        environments: TurnEnvironmentOverride,
```
pakrym-oai: should this be on SessionSettingsUpdate?

codex-rs/core/src/agent/control.rs
```diff
@@ -52,6 +53,7 @@ pub(crate) enum SpawnAgentForkMode {
 pub(crate) struct SpawnAgentOptions {
     pub(crate) fork_parent_spawn_call_id: Option<String>,
     pub(crate) fork_mode: Option<SpawnAgentForkMode>,
+    pub(crate) environments: Option<Vec<TurnEnvironmentSelection>>,
```
pakrym-oai, line 56: nit: non optional? non blocking

codex-rs/core/src/session/review.rs
```diff
@@ -111,6 +111,7 @@ pub(super) async fn spawn_review_thread(
         reasoning_summary,
         session_source,
         environment: parent_turn_context.environment.clone(),
+        environment_selections: parent_turn_context.environment_selections.clone(),
```
pakrym-oai: nit: why are we storing both environment_selections and environments? should we store only materialized list?

codex-rs/core/src/session/session.rs
```diff
@@ -238,6 +246,10 @@ pub(crate) struct SessionSettingsUpdate {
     pub(crate) reasoning_summary: Option<ReasoningSummaryConfig>,
     pub(crate) service_tier: Option<Option<ServiceTier>>,
     pub(crate) final_output_json_schema: Option<Option<Value>>,
+    /// Turn-local environment override. This is consumed while building a turn
+    /// context and does not mutate the sticky thread environments stored on
+    /// `SessionConfiguration`.
+    pub(crate) environments: TurnEnvironmentOverride,
```
pakrym-oai: Should this be Option<> ? like other fields/

codex-rs/core/src/session/turn_context.rs
```diff
@@ -565,59 +588,46 @@ impl Session {
                 sub_id,
                 session_configuration,
                 updates.final_output_json_schema,
+                effective_environments,
```
pakrym-oai: doesn't feel like we should put effective_environments on turn considering we have the entire resolved list.

codex-rs/core/src/session/turn_context.rs
```diff
@@ -488,26 +518,15 @@ impl Session {
...
-                .await;
-                return Err(err);
-            }
-        };
-        let update_result = {
+        let update_result: CodexResult<_> = {
             let mut state = self.state.lock().await;
             match state.session_configuration.clone().apply(&updates) {
                 Ok(next) => {
+                    let effective_environments =
```
pakrym-oai: Not sure I understand what is happening here.

In theory state.session_configuration.clone().apply(&updates)  will apply updates to the config and return the new environment list as part of `next`. Why do we need to `resolve_ref` here ?

codex-rs/core/src/agent/control.rs
starr-openai: this one still need Option for the 'use defaults' mode, which a few other callsites still use (e.g. memories/phase2.rs)

codex-rs/core/src/session/session.rs
starr-openai: good call - swapped to Option, which means we can drop what was previously an enum class serving pretty much the same purpose

starr-openai: (codex used an enum class to be more explicit on the two modes, but Option<> is more consistent with how we're using env everywehre else)

codex-rs/core/src/session/turn_context.rs
starr-openai: ^ that's just the thread level settings - don't we still need to apply the turn level overrides?

pakrym-oai (approved):

PR #18900, opened 2026-04-21
Migrate fork and resume reads to thread store

wiltzius-openai:
- Route cold thread/resume and thread/fork source loading through ThreadStore reads instead of direct rollout path operations
- Keep lookups that explicitly specify a rollout-path using the local thread store methods but return an invalid-request error for remote ThreadStore configurations
- Add some additional unit tests for code path coverage

### DIFF
--- a/codex-rs/thread-store/src/local/read_thread.rs
+++ b/codex-rs/thread-store/src/local/read_thread.rs
@@ -29,6 +29,13 @@ pub(super) async fn read_thread(
     let thread_id = params.thread_id;
     if let Some(metadata) = read_sqlite_metadata(store, thread_id).await
         && (params.include_archived || metadata.archived_at.is_none())
+        && (!params.include_history
+            || sqlite_rollout_path_can_load_history_for_thread(
+                store,
+                &metadata.rollout_path,
+                thread_id,
+            )
+            .await)
     {
         let mut thread = stored_thread_from_sqlite_metadata(store, metadata).await;
         attach_history_if_requested(&mut thread, params.include_history).await?;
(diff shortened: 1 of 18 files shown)


wiltzius-openai:
@codex review

wiltzius-openai:
@codex review

wiltzius-openai:
@codex review

wiltzius-openai:
@codex review

wiltzius-openai:
@codex review

wiltzius-openai:
@codex review

wiltzius-openai:
@codex review

wiltzius-openai:
@codex review

wiltzius-openai:
@codex review

codex-rs/thread-store/src/local/read_thread.rs
```diff
@@ -46,6 +53,19 @@ pub(super) async fn read_thread(
...
 
+async fn sqlite_rollout_path_can_load_history_for_thread(
+    store: &LocalThreadStore,
+    path: &std::path::Path,
+    thread_id: codex_protocol::ThreadId,
+) -> bool {
+    if !tokio::fs::try_exists(path).await.unwrap_or(false) {
+        return false;
+    }
+    read_thread_from_rollout_path(store, path.to_path_buf())
```
pakrym-oai, line 67: surprised we need to double check ids. not sure how we ever get the missalignment

pakrym-oai (approved):

codex-rs/thread-store/src/local/read_thread.rs
wiltzius-openai: in practice I hope it doesn't happen, but added a comment clarifying the case: basically this is a guard against rollout files being moved around without sqlite being updated. 

I am going to need to do a pass at some point on reducing the number of reads in all of this code, because its currently pretty chatty and I think is going to be quite slow doing full-history reads against sediment. But for now, migration+safety.

PR #18902, opened 2026-04-21
[hooks] clean up SessionStart and input hook handling

eternal-openai:
## Why
Pavel called out two cleanup opportunities in `run_turn`:

- The turn already runs pending `SessionStart` hooks before fresh `UserPromptSubmit` handling. The second `run_pending_session_start_hooks` call inside the sampling loop was leftover orchestration that no longer did anything once the pre-loop call consumed the pending source.
- Fresh user input and pending input had grown separate but very similar hook paths, which made the method harder to reason about.

While consolidating those paths, this branch also preserves whether pending input originally came from a user submission or from an internal `ResponseInputItem`. That matters for the hooks contract: user-submitted prompts should always run `UserPromptSubmit` before processing, even if the text happens to look like internal hook/context XML.

## What Changed
- Removed the dead loop-time `run_pending_session_start_hooks` call from `codex-rs/core/src/session/turn.rs`.
- Kept the existing pre-loop ordering where pending `SessionStart` hooks run before the turn's fresh `UserPromptSubmit` handling.
- Routed fresh input and pending input through shared `inspect_input_item` / `record_input_item` helpers in `codex-rs/core/src/hook_runtime.rs`.
- Added `PendingInputItem` so queued UI-submitted prompts keep their user-input provenance, while internal pending items continue to flow as `ResponseInputItem`s.
- Updated task-finish pending-input handling in `codex-rs/core/src/tasks/mod.rs` to use the same shared hook plumbing.
- Added regression coverage for fresh and queued XML-shaped user prompts so they still hit `UserPromptSubmit` before reaching the model.

## Verification
- `cargo test -p codex-core --test all hook_prompt_xml_text_still_runs_user_prompt_submit -- --nocapture`
- `cargo test -p codex-core --test all hooks:: -- --nocapture`


codex-rs/core/src/hook_runtime.rs
```diff
@@ -50,6 +52,8 @@ pub(crate) enum PendingInputHookDisposition {
 
 pub(crate) enum PendingInputRecord {
     UserMessage {
+        // Queued user input still carries UI-only text spans. Preserve that
```
pakrym-oai, line 55: useful?

codex-rs/core/src/hook_runtime.rs
```diff
@@ -277,8 +281,31 @@ pub(crate) async fn run_user_prompt_submit_hooks(
 pub(crate) async fn inspect_pending_input(
     sess: &Arc<Session>,
     turn_context: &Arc<TurnContext>,
-    pending_input_item: ResponseInputItem,
+    pending_input_item: PendingInputItem,
 ) -> PendingInputHookDisposition {
+    let pending_input_item = match pending_input_item {
```
pakrym-oai, line 286: this method is heavily duplicated between PendingInputItem::UserInput and Some(TurnItem::UserMessage(user_message)) = parse_turn_item

can it be streamlined?

codex-rs/core/src/hook_runtime.rs
```diff
@@ -277,8 +281,31 @@ pub(crate) async fn run_user_prompt_submit_hooks(
...
+                    response_item,
+                    additional_contexts: user_prompt_submit_outcome.additional_contexts,
+                },
+            ));
+        }
+        PendingInputItem::ResponseInput(input_item) => input_item,
+    };
+
     let response_item = ResponseItem::from(pending_input_item);
     if let Some(TurnItem::UserMessage(user_message)) = parse_turn_item(&response_item) {
```
pakrym-oai, line 310: Do we ever get real user messages as ResponseItems?

codex-rs/core/tests/suite/pending_input.rs
```diff
@@ -487,6 +556,149 @@ async fn user_input_does_not_preempt_after_reasoning_item() {
...
+                final_output_json_schema: None,
+                responsesapi_client_metadata: None,
+            })
+            .await?;
+    }
+
+    sleep(Duration::from_millis(100)).await;
+    let _ = gate_completed_tx.send(());
+
+    let requests = tokio::time::timeout(Duration::from_secs(30), async {
```
pakrym-oai, line 651: there must be a more reliable pattern to do this

codex-rs/core/tests/suite/pending_input.rs
```diff
@@ -487,6 +556,149 @@ async fn user_input_does_not_preempt_after_reasoning_item() {
...
+                    text: text.to_string(),
+                    text_elements: Vec::new(),
+                }],
+                final_output_json_schema: None,
+                responsesapi_client_metadata: None,
+            })
+            .await?;
+    }
+
+    sleep(Duration::from_millis(100)).await;
```
pakrym-oai, line 648: racy

codex-rs/core/src/state/turn.rs
```diff
@@ -219,10 +226,15 @@ impl TurnState {
     }
 
     pub(crate) fn push_pending_input(&mut self, input: ResponseInputItem) {
+        self.pending_input
+            .push(PendingInputItem::ResponseInput(input));
+    }
+
+    pub(crate) fn push_pending_input_item(&mut self, input: PendingInputItem) {
```
pakrym-oai, line 233: Should more callers to moved to user this method? for example steer_input

pakrym-oai: Ideally we capture all UserInputs correctly and you don't need to own the ResponseItem fallback

codex-rs/core/src/state/turn.rs
```diff
@@ -241,6 +253,17 @@ impl TurnState {
         }
     }
 
+    #[cfg(test)]
+    pub(crate) fn take_pending_input(&mut self) -> Vec<ResponseInputItem> {
```
pakrym-oai, line 257: can we delete this one?

pakrym-oai (commented):
I think we can simplify more by using the new item type you've added

PR #18811, opened 2026-04-21
refactor: route Codex auth through AuthProvider

efrazer-oai:
## Summary

This PR moves Codex backend request authentication from direct bearer-token handling to `AuthProvider`.

The new `codex-auth-provider` crate defines the shared request-auth trait. `CodexAuth::provider()` returns a provider that can apply all headers needed for the selected auth mode.

This lets ChatGPT token auth and AgentIdentity auth share the same callsite path:
- ChatGPT token auth applies bearer auth plus account/FedRAMP headers where needed.
- AgentIdentity auth applies AgentAssertion plus account/FedRAMP headers where needed.

Reference old stack: https://github.com/openai/codex/pull/17387/changes

## Callsite Migration

| Area | Change |
| --- | --- |
| backend-client | accepts an `AuthProvider` instead of a raw token/header |
| chatgpt client/connectors | applies auth through `CodexAuth::provider()` |
| cloud tasks | keeps Codex-backend gating, applies auth through provider |
| cloud requirements | uses Codex-backend auth checks and provider headers |
| app-server remote control | applies provider headers for backend calls |
| MCP Apps/connectors | gates on `uses_codex_backend()` and keys caches from generic account getters |
| model refresh | treats AgentIdentity as Codex-backend auth |
| OpenAI file upload path | rejects non-Codex-backend auth before applying headers |
| core client setup | keeps model-provider auth flow and allows AgentIdentity through provider-backed OpenAI auth |

## Stack

1. https://github.com/openai/codex/pull/18757: full revert
2. https://github.com/openai/codex/pull/18871: isolated Agent Identity crate
3. https://github.com/openai/codex/pull/18785: explicit AgentIdentity auth mode and startup task allocation
4. This PR: migrate Codex backend auth callsites through AuthProvider
5. https://github.com/openai/codex/pull/18904: accept AgentIdentity JWTs and load `CODEX_AGENT_IDENTITY`

## Testing

Tests: targeted Rust checks, cargo-shear, Bazel lock check, and CI.

PR #18904, opened 2026-04-21
feat: load AgentIdentity from JWT login/env

efrazer-oai:
## Summary

This PR lets programmatic AgentIdentity users provide one token through either stdin login or environment auth.

`codex login --with-agent-identity` reads an Agent Identity JWT from stdin, validates that it has the required claims, and stores that token as the `agent_identity` value in `auth.json`. The file format is token-only; the decoded account and key fields are runtime state, not hand-authored auth.json fields.

The Agent Identity JWT claim shape and decoder live in `codex-agent-identity`; `codex-login` only owns env/storage precedence and conversion into `CodexAuth::AgentIdentity`.

When env auth is enabled, `CODEX_AGENT_IDENTITY` can provide the same JWT without writing auth state to disk. `CODEX_API_KEY` still wins if both env vars are set.

Reference old stack: https://github.com/openai/codex/pull/17387/changes
Reference JWT/env stack: https://github.com/openai/codex/pull/18176

## Stack

1. https://github.com/openai/codex/pull/18757: full revert
2. https://github.com/openai/codex/pull/18871: isolated Agent Identity crate
3. https://github.com/openai/codex/pull/18785: explicit AgentIdentity auth mode and startup task allocation
4. https://github.com/openai/codex/pull/18811: migrate Codex backend auth callsites through AuthProvider
5. This PR: accept AgentIdentity JWTs through login/env

## Testing

Tests: targeted login and Agent Identity crate tests, CLI checks, scoped formatter/linter cleanup, and CI.


efrazer-oai:
> ### 💡 Codex Review
> https://github.com/openai/codex/blob/ca0aeace6b619f819a8afdc0cc603296d8e7420e/codex-rs/login/src/auth/manager.rs#L1023-L1030
> 
> 
> **P1 Badge Infer AgentIdentity mode when agent_identity is present**
> The new JWT/object deserialization for `auth.json.agent_identity` is ineffective unless `auth_mode` is also set, because `resolved_mode()` ignores `agent_identity`. A JWT-only `auth.json` falls back to `Chatgpt` mode and later behaves as missing token auth, which contradicts the “paste one token into auth.json” flow introduced here.
> 
> ℹ️ About Codex in GitHub

Disagree here, we should absolutely set `auth_mode` if we're copy-pasting in `auth.json` as the structure is tagged. If we're doing env variable, we don't break.

codex-rs/login/src/auth/storage.rs
```diff
@@ -41,7 +42,11 @@ pub struct AuthDotJson {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub last_refresh: Option<DateTime<Utc>>,
 
-    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        deserialize_with = "deserialize_agent_identity"
```
pakrym-oai: why are we not storing in a consistent format?

codex-rs/login/src/auth/storage.rs
```diff
@@ -56,6 +61,58 @@ pub struct AgentIdentityAuthRecord {
...
 
+#[derive(Deserialize)]
+#[serde(untagged)]
+enum AgentIdentityAuthInput {
+    Record(AgentIdentityAuthRecord),
+    Jwt(String),
+}
+
+#[derive(Deserialize)]
+struct AgentIdentityJwtClaims {
```
pakrym-oai: can this go into agent identity crate?

pakrym-oai (changes requested):
Let's not support editing auth.json directly

codex-rs/cli/src/main.rs
```diff
@@ -347,6 +349,12 @@ struct LoginCommand {
     )]
     with_api_key: bool,
 
+    #[arg(
+        long = "with-agent-identity",
```
pakrym-oai, line 372: should this say experimental for now?

pakrym-oai (approved):

PR #18907, opened 2026-04-21
respect workspace option for disabling plugins

zamoshchin-openai:
Respects the workspace setting for plugins in Codex

Plugins menu disappears
Plugins do not load
Plugins do not load in composer

no plugins loaded
Screenshot 2026-04-23 at 3 20 45 PM

no plugins in menu
Screenshot 2026-04-23 at 3 20 35 PM


zamoshchin-openai:

shijie-oai (approved):

codex-rs/chatgpt/src/workspace_settings.rs
```diff
@@ -0,0 +1,54 @@
...
+        return Ok(true);
+    }
+
+    let Some(account_id) = token_data.account_id.as_deref().filter(|id| !id.is_empty()) else {
+        return Ok(true);
+    };
+
+    let settings: WorkspaceSettingsResponse = chatgpt_get_request_with_timeout(
+        config,
+        format!("/accounts/{account_id}/settings"),
```
bolinfest: Do we have to sanitize or sanity check `account_id` or does `token_data` take care of that for us?

bolinfest (approved):
seems reasonable? but @xli-oai should have final say

xli-oai:
> seems reasonable? but @xli-oai should have final say

Guess you mean @xl-openai? 
But this LGTM, and later maybe we can skip this BG if there is no configured plugins

xl-openai:
So add a remote API call for most plugin API calls? My major concern is performance. Should we cache the result?

zamoshchin-openai:
added a 24 hour workspace settings cache

codex-rs/chatgpt/src/workspace_settings.rs
```diff
@@ -0,0 +1,152 @@
...
+
+use anyhow::Context;
+use codex_core::config::Config;
+use codex_login::CodexAuth;
+use serde::Deserialize;
+
+use crate::chatgpt_client::chatgpt_get_request_with_timeout;
+
+const WORKSPACE_SETTINGS_TIMEOUT: Duration = Duration::from_secs(10);
+const WORKSPACE_SETTINGS_CACHE_TTL: Duration = Duration::from_secs(24 * 60 * 60);
```
xl-openai: I think 15 mins is good enough.

xl-openai (approved):

PR #18908, opened 2026-04-21
Add remote thread config endpoint

rasmusrygaard:
## Why

App-server needs a way to fetch thread-scoped config from the remote thread config service when the user config opts into that behavior. This mirrors the existing experimental remote thread store endpoint while keeping local/noop behavior as the default.

Startup paths also need to avoid silently dropping the remote config endpoint after the first config load. The stdio app-server path discovers the endpoint from the initial config and installs the real thread config loader for later config builds, while in-process clients used by TUI/exec now select the same remote loader directly from their provided config.

## What changed

- Added `experimental_thread_config_endpoint` to `ConfigToml`, `Config`, and `core/config.schema.json`.
- Added config parsing coverage for the new setting.
- Updated app-server startup to select `RemoteThreadConfigLoader` from the initially loaded config, falling back to `NoopThreadConfigLoader` when unset.
- Let `ConfigManager` replace its thread config loader after startup discovery so later config loads use the selected loader.
- Updated in-process app-server client startup to pass `RemoteThreadConfigLoader` when its config has `experimental_thread_config_endpoint` set.

## Verification

- Added `experimental_thread_config_endpoint_loads_from_config_toml`.
- Added `runtime_start_args_use_remote_thread_config_loader_when_configured`.
- Ran `cargo check -p codex-app-server --lib`.
- Ran `cargo test -p codex-app-server-client`.

### DIFF
--- a/codex-rs/app-server/src/config_manager.rs
+++ b/codex-rs/app-server/src/config_manager.rs
@@ -131,7 +131,7 @@ impl ConfigManager {
         }
     }
 
-    pub(crate) fn current_thread_config_loader(&self) -> Arc<dyn ThreadConfigLoader> {
+    fn current_thread_config_loader(&self) -> Arc<dyn ThreadConfigLoader> {
         self.thread_config_loader
             .read()
             .map(|guard| Arc::clone(&*guard))
--- a/codex-rs/app-server/src/config_manager.rs
+++ b/codex-rs/app-server/src/config_manager.rs
@@ -32,7 +32,7 @@ pub(crate) struct ConfigManager {
     loader_overrides: LoaderOverrides,
     cloud_requirements: Arc<RwLock<CloudRequirementsLoader>>,
     arg0_paths: Arg0DispatchPaths,
-    thread_config_loader: Arc<dyn ThreadConfigLoader>,
+    thread_config_loader: Arc<RwLock<Arc<dyn ThreadConfigLoader>>>,
     host_name: Option<String>,
 }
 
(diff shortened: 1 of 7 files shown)


codex-rs/app-server/src/config_manager.rs
```diff
@@ -32,7 +32,7 @@ pub(crate) struct ConfigManager {
     loader_overrides: LoaderOverrides,
     cloud_requirements: Arc<RwLock<CloudRequirementsLoader>>,
     arg0_paths: Arg0DispatchPaths,
-    thread_config_loader: Arc<dyn ThreadConfigLoader>,
+    thread_config_loader: Arc<RwLock<Arc<dyn ThreadConfigLoader>>>,
```
pakrym-oai, line 35: can we create loader early enough not to need hot-replace?

pakrym-oai (changes requested):

codex-rs/app-server/src/config_manager.rs
rasmusrygaard: The tricky part is that we depend on the `config` to resolve config settings, similar to how the cloud requirements depend on `chatgpt_base_url`. We can resolve the config values again here if we want to but that also seems a bit dirty

codex-rs/app-server/src/config_manager.rs
```diff
@@ -120,6 +120,24 @@ impl ConfigManager {
...
+        thread_config_loader: Arc<dyn ThreadConfigLoader>,
+    ) {
+        if let Ok(mut guard) = self.thread_config_loader.write() {
+            *guard = thread_config_loader;
+        } else {
+            warn!("failed to update thread config loader");
+        }
+    }
+
+    pub(crate) fn current_thread_config_loader(&self) -> Arc<dyn ThreadConfigLoader> {
```
pakrym-oai: nit: make not pub?

pakrym-oai (approved):

codex-rs/app-server/src/config_manager.rs
pakrym-oai: Yeah, circular dependencies here are brutal 😢

PR #18910, opened 2026-04-22
Publish standalone installer release archives

efrazer-oai:
## Summary

This PR adds release files that are made for the standalone installer.

This PR publishes those files during the Rust release:

- `codex-standalone-<platform>-<version>.tar.gz`
- `codex-installer_SHA256SUMS`

Each standalone archive has the layout the installer needs directly. It contains `codex`, `rg`, and the Windows helper binaries where needed.

## Stack

1. #18910, this PR: publish standalone installer archives and their checksum file.
2. #18901: make the Unix and Windows installers download those archives and verify them before install.

## Design decisions

- Checksum generation stays as its own release step. The archive-staging script's job is to produce standalone archives. The checksum step's job is to publish a manifest over the final set of release archives that actually landed in `dist/installer`.
- The installer consumes one checksum manifest for the release instead of per-archive embedded metadata. This keeps the installer flow simple: download `codex-installer_SHA256SUMS`, look up the requested archive, verify it, then extract it.
- This follows the checksum-manifest pattern we already use for `rusty_v8`: stage artifacts, write a SHA-256 manifest next to them, and have consumers verify downloads against that manifest.

## Tests

Tests: workflow formatter check, Python compile checks, standalone archive staging smoke tests covering linux-x64, linux-arm64, darwin-x64, darwin-arm64, win32-x64, and win32-arm64 via synthetic vendor inputs, plus archive content checks for Unix and Windows layouts.


efrazer-oai:
@codex review

.github/workflows/rust-release.yml
```diff
@@ -503,13 +503,41 @@ jobs:
             --release-version "$RELEASE_VERSION" \
             --package codex \
             --package codex-responses-api-proxy \
-            --package codex-sdk
+            --package codex-sdk \
+            --standalone-output-dir dist/installer
```
bolinfest: I don't really love that this is a "feature" of the npm stuff rather than its own step. Can we assemble the right args so that `stage_standalone_installer_archives.py` is called in its own step? Would it be materially less efficient for some reason?

scripts/stage_standalone_installer_archives.py
```diff
@@ -0,0 +1,128 @@
...
+import argparse
+import importlib.util
+import shutil
+import tarfile
+import tempfile
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+BUILD_SCRIPT = REPO_ROOT / "codex-cli" / "scripts" / "build_npm_package.py"
```
bolinfest: Again, I don't think this should be conflated with what we are doing in npm. Or if anything, this should be the source of truth and npm should depend on this?

efrazer-oai:
@bolinfest fixed as per your feedback!

PR #18917, opened 2026-04-22
[codex] Support remote plugin install writes

xli-oai:
## Summary
- Add a remote plugin install write call that POSTs the selected remote plugin to the ChatGPT cloud plugin API.
- Align remote install with the latest remote read contract: `pluginName` carries the backend remote plugin id directly, for example `plugins~Plugin_linear`, and install no longer synthesizes `<name>@<marketplace>` ids.
- Validate remote install ids with the same character rules as remote read, return the same install response shape as local installs, and include mocked app-server coverage for the write path.

## Validation
- `just fmt`
- `cargo test -p codex-app-server --test all plugin_install`
- `just fix -p codex-core-plugins`

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor/plugins.rs
+++ b/codex-rs/app-server/src/codex_message_processor/plugins.rs
@@ -1,4 +1,5 @@
 use super::*;
+use codex_app_server_protocol::PluginInstallPolicy;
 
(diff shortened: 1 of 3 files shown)


xl-openai (approved):

xli-oai:
@codex review

PR #18950, opened 2026-04-22
feat: let model providers own model discovery

celia-oai:
## Why

`codex-models-manager` had grown to own provider-specific concerns: constructing OpenAI-compatible `/models` requests, resolving provider auth, emitting request telemetry, and deciding how provider catalogs should be sourced. That made the manager harder to reuse for providers whose model catalog is not fetched from the OpenAI `/models` endpoint, such as Amazon Bedrock.

This change moves provider-specific model discovery behind provider-owned implementations, so the models manager can focus on refresh policy, cache behavior, picker ordering, and model metadata merging.

## What Changed

- Introduced a `ModelsManager` trait with separate `OpenAiModelsManager` and `StaticModelsManager` implementations.
- Added `ModelsEndpointClient` so OpenAI-compatible HTTP fetching lives outside `codex-models-manager`.
- Moved `/models` request construction, provider auth resolution, timeout handling, and request telemetry into `codex-model-provider` via `OpenAiModelsEndpoint`.
- Added provider-owned `models_manager(...)` construction so configured OpenAI-compatible providers use `OpenAiModelsManager`, while static/catalog-backed providers can return `StaticModelsManager`.
- Added an Amazon Bedrock static model catalog for the GPT OSS Bedrock model IDs.
- Updated core/session/thread manager code and tests to depend on `Arc<dyn ModelsManager>`.
- Moved offline model test helpers into `codex_models_manager::test_support`.
## Metadata References

The Bedrock catalog metadata is based on the official Amazon Bedrock OpenAI model documentation:

- [Amazon Bedrock OpenAI models](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-openai.html) lists the Bedrock model IDs, text input/output modalities, and `128,000` token context window for `gpt-oss-20b` and `gpt-oss-120b`.
- [Amazon Bedrock `gpt-oss-120b` model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-oss-120b.html) lists the `bedrock-runtime` model ID `openai.gpt-oss-120b-1:0`, the `bedrock-mantle` model ID `openai.gpt-oss-120b`, text-only modalities, and `128K` context window.
- [OpenAI `gpt-oss-120b` model docs](https://developers.openai.com/api/docs/models/gpt-oss-120b) document configurable reasoning effort with `low`, `medium`, and `high`, plus text input/output modality.

The display names, default reasoning effort, and priority ordering are Codex-local catalog choices.

## Test Plan
- Manually verified app-server model listing with an AWS profile:

```shell
CODEX_HOME="$(mktemp -d)" cargo run -p codex-app-server-test-client -- \
  --codex-bin ./target/debug/codex \
  -c 'model_provider="amazon-bedrock"' \
  -c 'model_providers.amazon-bedrock.aws.profile="codex-bedrock"' \
  -c 'model_providers.amazon-bedrock.aws.region="us-west-2"' \
  model-list
```

The response returned the Bedrock catalog with `openai.gpt-oss-120b-1:0` as the default model and `openai.gpt-oss-20b-1:0` as the second listed model, both text-only and supporting low/medium/high reasoning effort.

### DIFF
--- a/codex-rs/core/src/session/mod.rs
+++ b/codex-rs/core/src/session/mod.rs
@@ -66,8 +66,8 @@ use codex_mcp::McpConnectionManager;
 use codex_mcp::McpRuntimeEnvironment;
 use codex_mcp::ToolInfo;
 use codex_mcp::codex_apps_tools_cache_key;
-use codex_models_manager::manager::ModelsManager;
 use codex_models_manager::manager::RefreshStrategy;
+use codex_models_manager::manager::SharedModelsManager;
 use codex_network_proxy::NetworkProxy;
 use codex_network_proxy::NetworkProxyAuditMetadata;
 use codex_network_proxy::normalize_host;
@@ -389,7 +389,7 @@ pub struct CodexSpawnOk {
 pub(crate) struct CodexSpawnArgs {
     pub(crate) config: Config,
     pub(crate) auth_manager: Arc<AuthManager>,
-    pub(crate) models_manager: Arc<dyn ModelsManager>,
+    pub(crate) models_manager: SharedModelsManager,
     pub(crate) environment_manager: Arc<EnvironmentManager>,
     pub(crate) skills_manager: Arc<SkillsManager>,
     pub(crate) plugins_manager: Arc<PluginsManager>,
--- a/codex-rs/core/src/session/mod.rs
+++ b/codex-rs/core/src/session/mod.rs
@@ -66,8 +66,6 @@ use codex_mcp::McpConnectionManager;
 use codex_mcp::McpRuntimeEnvironment;
 use codex_mcp::ToolInfo;
 use codex_mcp::codex_apps_tools_cache_key;
-#[cfg(test)]
-use codex_models_manager::collaboration_mode_presets::CollaborationModesConfig;
 use codex_models_manager::manager::ModelsManager;
 use codex_models_manager::manager::RefreshStrategy;
 use codex_network_proxy::NetworkProxy;
@@ -391,7 +389,7 @@ pub struct CodexSpawnOk {
 pub(crate) struct CodexSpawnArgs {
     pub(crate) config: Config,
     pub(crate) auth_manager: Arc<AuthManager>,
-    pub(crate) models_manager: Arc<ModelsManager>,
+    pub(crate) models_manager: Arc<dyn ModelsManager>,
     pub(crate) environment_manager: Arc<EnvironmentManager>,
     pub(crate) skills_manager: Arc<SkillsManager>,
     pub(crate) plugins_manager: Arc<PluginsManager>,
--- a/codex-rs/core/src/session/session.rs
+++ b/codex-rs/core/src/session/session.rs
@@ -270,7 +270,7 @@ impl Session {
         mut session_configuration: SessionConfiguration,
         config: Arc<Config>,
         auth_manager: Arc<AuthManager>,
-        models_manager: Arc<ModelsManager>,
+        models_manager: Arc<dyn ModelsManager>,
         exec_policy: Arc<ExecPolicyManager>,
         tx_event: Sender<Event>,
         agent_status: watch::Sender<AgentStatus>,
--- a/codex-rs/core/src/session/session.rs
+++ b/codex-rs/core/src/session/session.rs
@@ -270,7 +270,7 @@ impl Session {
         mut session_configuration: SessionConfiguration,
         config: Arc<Config>,
         auth_manager: Arc<AuthManager>,
-        models_manager: Arc<dyn ModelsManager>,
+        models_manager: SharedModelsManager,
         exec_policy: Arc<ExecPolicyManager>,
         tx_event: Sender<Event>,
         agent_status: watch::Sender<AgentStatus>,
--- a/codex-rs/core/src/session/turn_context.rs
+++ b/codex-rs/core/src/session/turn_context.rs
@@ -117,7 +117,11 @@ impl TurnContext {
         self.features.apps_enabled_for_auth(uses_codex_backend)
     }
 
-    pub(crate) async fn with_model(&self, model: String, models_manager: &ModelsManager) -> Self {
+    pub(crate) async fn with_model(
+        &self,
+        model: String,
+        models_manager: &dyn ModelsManager,
+    ) -> Self {
         let mut config = (*self.config).clone();
         config.model = Some(model.clone());
         let model_info = models_manager
@@ -381,7 +385,7 @@ impl Session {
         main_execve_wrapper_exe: Option<&PathBuf>,
         per_turn_config: Config,
         model_info: ModelInfo,
-        models_manager: &ModelsManager,
+        models_manager: &dyn ModelsManager,
         network: Option<NetworkProxy>,
         environment: Option<Arc<Environment>>,
         environments: Vec<TurnEnvironment>,
@@ -648,7 +652,7 @@ impl Session {
             self.services.main_execve_wrapper_exe.as_ref(),
             per_turn_config,
             model_info,
-            &self.services.models_manager,
+            self.services.models_manager.as_ref(),
             self.services
                 .network_proxy
                 .as_ref()
--- a/codex-rs/core/src/session/turn_context.rs
+++ b/codex-rs/core/src/session/turn_context.rs
@@ -120,7 +120,7 @@ impl TurnContext {
     pub(crate) async fn with_model(
         &self,
         model: String,
-        models_manager: &dyn ModelsManager,
+        models_manager: &SharedModelsManager,
     ) -> Self {
         let mut config = (*self.config).clone();
         config.model = Some(model.clone());
@@ -385,7 +385,7 @@ impl Session {
         main_execve_wrapper_exe: Option<&PathBuf>,
         per_turn_config: Config,
         model_info: ModelInfo,
-        models_manager: &dyn ModelsManager,
+        models_manager: &SharedModelsManager,
         network: Option<NetworkProxy>,
         environment: Option<Arc<Environment>>,
         environments: Vec<TurnEnvironment>,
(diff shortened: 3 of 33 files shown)


celia-oai:
@codex review

celia-oai:
@codex review

codex-rs/models-manager/src/manager.rs
```diff
@@ -3,138 +3,42 @@ use crate::collaboration_mode_presets::CollaborationModesConfig;
...
 
-impl RequestTelemetry for ModelsRequestTelemetry {
-    fn on_request(
+/// Remote endpoint used by the OpenAI-compatible model manager.
+///
+/// Implementations own provider-specific auth and transport details. The model
+/// manager owns refresh policy, cache behavior, and catalog merging; it calls
+/// this endpoint only when it decides a remote refresh should happen.
+#[async_trait]
+pub trait ModelsEndpointClient: fmt::Debug + Send + Sync {
```
pakrym-oai, line 32: do we need a second layer of abstraction here? If the OAI MM is the only one that consumes it

codex-rs/models-manager/src/manager.rs
```diff
@@ -164,97 +68,127 @@ impl fmt::Display for RefreshStrategy {
...
-            auth_manager,
-            model_catalog,
-            collaboration_modes_config,
-            ModelProviderInfo::create_openai_provider(/*base_url*/ None),
-        )
-    }
+/// Static model manager backed by an authoritative in-process catalog.
+#[derive(Debug)]
+pub struct StaticModelsManager {
+    remote_models: RwLock<Vec<ModelInfo>>,
```
pakrym-oai: why does static model maanger need rwlock?

codex-rs/models-manager/src/manager.rs
```diff
@@ -164,97 +68,127 @@ impl fmt::Display for RefreshStrategy {
...
+    collaboration_modes_config: CollaborationModesConfig,
+    auth_manager: Option<Arc<AuthManager>>,
```
pakrym-oai, line 195: why does static model maanger need AuthManager?

codex-rs/models-manager/src/manager.rs
```diff
@@ -531,21 +356,91 @@ impl ModelsManager {
...
-            .provider
-            .auth_manager()
-            .and_then(|auth_manager| auth_manager.auth_mode());
-        let chatgpt_mode = matches!(auth_mode, Some(AuthMode::Chatgpt));
-        presets = ModelPreset::filter_by_auth(presets, chatgpt_mode);
+    fn list_collaboration_modes_for_config(
+        &self,
+        collaboration_modes_config: CollaborationModesConfig,
+    ) -> Vec<CollaborationModeMask> {
+        builtin_collaboration_mode_presets(collaboration_modes_config)
```
pakrym-oai: does every implementation just call this method/

pakrym-oai: if so why does it have to be on MM? it doesn't use self.

codex-rs/models-manager/src/manager.rs
```diff
@@ -164,97 +68,127 @@ impl fmt::Display for RefreshStrategy {
...
+    /// List collaboration mode presets.
+    ///
+    /// Returns a static set of presets seeded with the configured model.
+    fn list_collaboration_modes(&self) -> Vec<CollaborationModeMask>;
+
+    fn list_collaboration_modes_for_config(
+        &self,
+        collaboration_modes_config: CollaborationModesConfig,
+    ) -> Vec<CollaborationModeMask>;
+
```
pakrym-oai: bit strange, we take CollaborationModesConfig into ctor and also as a parameter to this method.

alsmost feels like list_collaboration_modes doesn't belong here at all.

codex-rs/models-manager/src/manager.rs
```diff
@@ -310,84 +233,16 @@ impl ModelsManager {
...
-            .map(|model| model.model.clone())
-            .unwrap_or_default()
+        default_model_from_available(build_available_models(self.auth_mode(), remote_models))
     }
 
-    // todo(aibrahim): look if we can tighten it to pub(crate)
-    /// Look up model metadata, applying remote overrides and config adjustments.
     #[instrument(level = "info", skip(self, config), fields(model = model))]
-    pub async fn get_model_info(&self, model: &str, config: &ModelsManagerConfig) -> ModelInfo {
+    async fn get_model_info(&self, model: &str, config: &ModelsManagerConfig) -> ModelInfo {
```
pakrym-oai: can you implement this method on the trait directly?

codex-rs/models-manager/src/manager.rs
```diff
@@ -531,21 +356,91 @@ impl ModelsManager {
...
+    async fn get_remote_models(&self) -> Vec<ModelInfo> {
+        self.remote_models.read().await.clone()
+    }
+
+    fn try_get_remote_models(&self) -> Result<Vec<ModelInfo>, TryLockError> {
+        Ok(self.remote_models.try_read()?.clone())
+    }
+
+    #[cfg(test)]
+    async fn manipulate_cache_for_test<F>(&self, f: F) -> std::io::Result<()>
```
pakrym-oai: we didn't need these for tests before. did we? we can keep tests at the same scope

codex-rs/core/src/session/mod.rs
```diff
@@ -383,7 +381,7 @@ pub struct CodexSpawnOk {
 pub(crate) struct CodexSpawnArgs {
     pub(crate) config: Config,
     pub(crate) auth_manager: Arc<AuthManager>,
-    pub(crate) models_manager: Arc<ModelsManager>,
+    pub(crate) models_manager: Arc<dyn ModelsManager>,
```
pakrym-oai: nit: either dyn ModelsManager or SharedModelsManager

codex-rs/core/src/session/session.rs
```diff
@@ -259,7 +259,7 @@ impl Session {
         mut session_configuration: SessionConfiguration,
         config: Arc<Config>,
         auth_manager: Arc<AuthManager>,
-        models_manager: Arc<ModelsManager>,
+        models_manager: Arc<dyn ModelsManager>,
```
pakrym-oai: ditto

codex-rs/core/src/session/turn_context.rs
```diff
@@ -110,7 +110,11 @@ impl TurnContext {
         self.features.apps_enabled_for_auth(is_chatgpt_auth)
     }
 
-    pub(crate) async fn with_model(&self, model: String, models_manager: &ModelsManager) -> Self {
+    pub(crate) async fn with_model(
+        &self,
+        model: String,
+        models_manager: &dyn ModelsManager,
```
pakrym-oai: ditto

codex-rs/models-manager/src/manager.rs
```diff
@@ -531,21 +356,91 @@ impl ModelsManager {
...
+    }
 
-        ModelPreset::mark_default_by_picker_visibility(&mut presets);
+    fn try_list_models(&self) -> Result<Vec<ModelPreset>, TryLockError> {
+        let remote_models = self.try_get_remote_models()?;
+        Ok(build_available_models(self.auth_mode(), remote_models))
+    }
 
-        presets
+    async fn get_default_model(
```
pakrym-oai: can this be implemented on the trait?

pakrym-oai (changes requested):
Looks good overall:
1. Can we move more implementations of methods onto the trait directly?
2. list_collaboration_modes doesn't seem to fit into MM trait?
3. Is there a chance we can reduce churn in manager_tests so it's easier to review?
4. Do we even need a separate remote loader trait?

codex-rs/models-manager/src/manager.rs
celia-oai: good call

codex-rs/models-manager/src/manager.rs
celia-oai: this is kept the same behavior from the original code- will fix to make it more straightforward

codex-rs/models-manager/src/manager.rs
celia-oai: it's for this preset filtering logic:
```
let chatgpt_mode = matches!(auth_mode, Some(AuthMode::Chatgpt));
    presets = ModelPreset::filter_by_auth(presets, chatgpt_mode);
```
not a huge fan of having to pass down auth mode from auth manager for this, but I prefer to preserve the current functionality. We can think of how to refactor this in a better way later. we can't cache this auth_mode either due to this codex comment: https://github.com/openai/codex/pull/18950#discussion_r3127734928.

codex-rs/models-manager/src/manager.rs
celia-oai: it's just symmetry with the oai model manager path but seems safe to remove

codex-rs/models-manager/src/manager.rs
celia-oai: removed, this is from legacy implementations

codex-rs/models-manager/src/manager.rs
celia-oai: good catch, removing.

codex-rs/models-manager/src/manager.rs
celia-oai: the reason this lives in model provider layer instead of model manager is because it contains auth/ telemetry and transport dependencies, so it's a much cleaner split than having everything pushed down to model manager. I tried to move it into model manager but then we have to reconstruct model provider auth layer like resolve_provider_auth, which gets messy quite quickly. so i strongly prefer this split.

re manager_test change is large: it's mostly remove the http mock layer now that we split out ModelsEndpointClient. hopefully this is a one-time pain and the test is more straightforwardly targeting testing model cache/refresh/catelog logic

pakrym-oai (approved):

PR #18953, opened 2026-04-22
use long-lived sessions for codex sandbox windows

iceweasel-oai:
`codex sandbox windows` previously did a one-shot spawn for all commands.
This change uses the `unified_exec` session to spawn long-lived processes instead, and implements a simple bridge to forward stdin to the spawned session and stdout/stderr from the spawned session back to the caller.

It also fixes a bug with the new shared spawn context code where the "no-network env" was being applied to both elevated and unelevated sandbox spawns. It should only be applied for the unelevated sandbox because the elevated one uses firewall rules instead of an env-based network suppression strategy.

### DIFF
--- a/codex-rs/cli/src/debug_sandbox.rs
+++ b/codex-rs/cli/src/debug_sandbox.rs
@@ -341,7 +341,9 @@ async fn run_command_under_windows_session(
 
     let session = std::sync::Arc::new(spawned.session);
     let tokio_runtime = tokio::runtime::Handle::current();
-    let output_drain_timeout = std::time::Duration::from_millis(500);
+    // Give large or slow tail output a better chance to finish draining
+    // without letting rare EOF issues hang the wrapper indefinitely.
+    let output_drain_timeout = std::time::Duration::from_secs(5);
     // A helper thread watches our stdin. When the input source closes it,
     // the thread tells the main async code so we can also close stdin for
     // the sandboxed child process.
--- a/codex-rs/cli/src/debug_sandbox.rs
+++ b/codex-rs/cli/src/debug_sandbox.rs
@@ -152,14 +152,7 @@ async fn run_command_under_sandbox(
     if let SandboxType::Windows = sandbox_type {
         #[cfg(target_os = "windows")]
         {
-            run_command_under_windows_session(
-                &config,
-                command,
-                cwd,
-                sandbox_policy_cwd,
-                env,
-            )
-            .await;
+            run_command_under_windows_session(&config, command, cwd, sandbox_policy_cwd, env).await;
         }
         #[cfg(not(target_os = "windows"))]
         {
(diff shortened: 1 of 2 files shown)


codex-rs/cli/src/debug_sandbox.rs
```diff
@@ -152,79 +152,14 @@ async fn run_command_under_sandbox(
...
-                        base_dir.as_path(),
-                        command_vec,
-                        &cwd_clone,
-                        env_map,
-                        /*timeout_ms*/ None,
-                        config.permissions.windows_sandbox_private_desktop,
-                    )
-                }
-            })
+            run_command_under_windows_session(
```
iceweasel-oai: this is just pulling the windows part into its own function because it's kinda big

codex-rs/cli/src/debug_sandbox.rs
```diff
@@ -347,6 +275,128 @@ async fn run_command_under_sandbox(
...
+            cwd.as_path(),
+            env,
+            None,
+            /*tty*/ false,
+            /*stdin_open*/ true,
+            config.permissions.windows_sandbox_private_desktop,
+        )
+        .await
+    } else {
+        spawn_windows_sandbox_session_legacy(
```
iceweasel-oai, line 319: mostly these `spawn_windows_sandbox_session_*` changes are the key part. They use the same backend as `unified_exec` will use instead of the previous "one and done" spawn approach

codex-rs/cli/src/debug_sandbox.rs
```diff
@@ -347,6 +275,128 @@ async fn run_command_under_sandbox(
...
+        res = tokio::signal::ctrl_c() => {
+            if let Ok(()) = res {
+                session.request_terminate();
+            }
+            exit_rx.await.unwrap_or(-1)
+        }
+    };
+
+    stdin_close_task.abort();
+    let _ = tokio::time::timeout(output_drain_timeout, async {
```
iceweasel-oai, line 394: this part is just waiting for stdout/stderr from the child process to drain after we receive its exit code

codex-rs/windows-sandbox-rs/src/spawn_prep.rs
```diff
@@ -132,6 +129,30 @@ pub(crate) fn prepare_legacy_spawn_context(
...
+        policy_json_or_preset,
+        codex_home,
+        cwd,
+        env_map,
+        command,
+        inherit_path,
+        add_git_safe_directory,
+    )?;
+    if should_apply_network_block(&common.policy) {
+        apply_no_network_to_env(env_map)?;
```
iceweasel-oai, line 151: apply the "no network" changes to the env here, only in the legacy function, not the elevated one

codex-rs/cli/src/debug_sandbox.rs
```diff
@@ -347,6 +282,128 @@ async fn run_command_under_sandbox(
...
+        let _ = stdout_forwarder_done_rx.await;
+        let _ = stderr_forwarder_done_rx.await;
+    })
+    .await;
```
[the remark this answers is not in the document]
iceweasel-oai: I may address this in a follow-up but it's not blocking for now.

aaronl-openai (approved):
thank you!

when I have my windows machine tomorrow I'll test this against my changes to make sure everything works together

PR #18956, opened 2026-04-22
[Codex] Register browser requirements feature keys

khoi-oai:
## Summary
- register `in_app_browser` and `browser_use` as stable feature keys
- allow requirements/MDM feature requirements to pin those desktop browser controls
- add coverage for browser requirements being accepted by config loading

## Testing
- `cargo fmt --all` (`just fmt` unavailable locally; rustfmt warned about nightly-only `imports_granularity` config)
- `cargo test -p codex-features`
- `cargo test -p codex-core browser_feature_requirements_are_valid`
- Tested manually by setting in `requirements.toml` and seeing after app restart state to reflect the setting was correct (at the time hiding the `Browser Use` setting when the enterprise setting was set to false

### DIFF
--- a/codex-rs/features/src/lib.rs
+++ b/codex-rs/features/src/lib.rs
@@ -156,6 +156,14 @@ pub enum Feature {
     ToolSuggest,
     /// Enable plugins.
     Plugins,
+    /// Allow the in-app browser pane in desktop apps.
+    ///
+    /// Requirements-only gate: this should be set from requirements, not user config.
+    InAppBrowser,
+    /// Allow Browser Use agent integration in desktop apps.
+    ///
+    /// Requirements-only gate: this should be set from requirements, not user config.
+    BrowserUse,
     /// Temporary internal-only flag for PS-backed remote plugin catalog development.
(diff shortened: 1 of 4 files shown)


codex-rs/features/src/lib.rs
```diff
@@ -156,6 +156,10 @@ pub enum Feature {
     ToolSuggest,
     /// Enable plugins.
     Plugins,
+    /// Allow the in-app browser pane in desktop apps.
```
xl-openai, line 159: Maybe clarify this could / should only be set in requirements but not config.

xl-openai (approved):

codex-rs/features/src/lib.rs
sayan-oai: +1, need to figure out something better for this in the future. feels weird having empty features here just to pass through the app server endpoint.

sayan-oai (approved):

bschoepke-openai (approved):

khoi-oai:

PR #19013, opened 2026-04-22
Allow plugin-declared first-party connectors

tinocaer:
## Why

Codex currently blocks every `connector_openai_*` id as a blanket policy. There are some 1p connectors that we do want to allow. This PR updates the connector filter policy to allow 1p connectors that are explicitely bundled in a Codex plugin. 

## What Changed

- Updated the shared connector disallow policy in `connectors/src/filter.rs` to accept `plugin_declared_connector_ids`.
- Added a shared id-level helper so non-`AppInfo` paths, especially Codex Apps MCP tools, use the same connector-id policy.
- Threaded plugin-declared connector ids through ChatGPT connector inventory, app-server app listing, TUI app loading, core accessible connector listing, and tool-suggest discoverability.
- Updated Codex Apps MCP tool filtering to use plugin provenance from `ToolPluginProvenance` instead of the duplicate connector-id filter in `utils/plugins`.
- Re-applied current plugin provenance when reading cached Codex Apps tools and accessible connector caches, so stale cached first-party tools are hidden after their declaring plugin is no longer active.
- Removed the now-duplicated connector-id policy from `utils/plugins/src/mcp_connector.rs`.

## Verification

- `cargo test -p codex-connectors filter`
- `cargo test -p codex-chatgpt connectors_for_plugin_apps`
- `cargo test -p codex-chatgpt connectors`
- `cargo test -p codex-core connectors::tests`
- `cargo test -p codex-mcp codex_apps_tools_cache`
- `cargo test -p codex-mcp tool_plugin_provenance`
- `cargo test -p codex-tui apps_popup_stays_loading_until_final_snapshot_updates`


tinocaer:

codex-rs/chatgpt/src/connectors.rs
```diff
@@ -150,23 +165,29 @@ pub fn connectors_for_plugin_apps(
...
+    .into_iter()
+    .filter(|connector| plugin_app_ids.contains(connector.id.as_str()))
+    .collect()
 }
 
 pub fn merge_connectors_with_accessible(
     connectors: Vec<AppInfo>,
     accessible_connectors: Vec<AppInfo>,
     all_connectors_loaded: bool,
+    plugin_declared_connector_ids: &HashSet<String>,
```
ccy-oai, line 190: looks like 2 layers of function call plumber. Should fetching plugin be a caller's concern?

codex-rs/tui/src/chatwidget.rs
```diff
@@ -7924,12 +7924,15 @@ impl ChatWidget {
...
+                let plugin_declared_connector_ids =
+                    connectors::plugin_declared_connector_ids_for_config(&config).await;
                 let all_connectors =
                     connectors::list_all_connectors_with_options(&config, force_refetch).await?;
                 let connectors = connectors::merge_connectors_with_accessible(
                     all_connectors,
                     accessible_connectors,
                     /*all_connectors_loaded*/ true,
+                    &plugin_declared_connector_ids,
                 );
```
ccy-oai, line 7936: IIUC
`all_connectors` -> Full listing
`accessible_connectors` -> Usable ones
`plugin_declared_connector_ids` -> Allow through disable list

... to create a usable but not visible connector?

codex-rs/connectors/src/filter.rs
```diff
@@ -56,13 +65,95 @@ fn is_first_party_chat_originator(originator_value: &str) -> bool {
...
         DISALLOWED_CONNECTOR_IDS
     };
 
+    if disallowed_connector_ids.contains(&connector_id) {
+        return false;
+    }
+
     !connector_id.starts_with(DISALLOWED_CONNECTOR_PREFIX)
-        && !disallowed_connector_ids.contains(&connector_id)
+        || plugin_declared_connector_ids.contains(connector_id)
```
ccy-oai, line 85: Since we know all the 1p connectors, possible to just do a disallow except these const list vs loading from config?

ccy-oai (approved):

codex-rs/connectors/src/filter.rs
tinocaer: If we just manually declare that certain 1p connector always pass, then it means they may always be loaded to the user, even if they shouldn't be.

The benefit of this approach is that the user must install or enable a plugin for the connector bundled with it to be accessible to the model. IMO this is more correct since if a user disabled a plugin, we wouldn't want the harness to continue showing the bundled connectors to the model.

ccy-oai: accessible list is checking config, possible to just gate there?

codex-rs/tui/src/chatwidget.rs
tinocaer: We don't want connectors like appgen to be generally visible on its own, since that would mean it could be discoverable in marketplaces like ChatGPT.

By default, Codex filters out all 1p connectors, but if the connector is bundled in a plugin, we obviously want the connector to be included.

The connector is still visible to the user, since they interface with it through a plugin.

codex-rs/chatgpt/src/connectors.rs
tinocaer: For most cases, it's simpler to resolve the pass in the connector ids instead of the config and resolving it internally

PR #19014, opened 2026-04-22
Add excludeTurns parameter to thread/resume and thread/fork

ddr-oai:
For callers who expect to be paginating the results for the UI, they can now call thread/resume or thread/fork with excludeturns:true so it will not fetch any pages of turns, and instead only set up the subscription.  That call can be immediately followed by pagination requests to thread/turns/list to fetch pages of turns according to the UI's current interactions.


ddr-oai:
@codex review

codex-rs/app-server/src/codex_message_processor/token_usage_replay.rs
```diff
@@ -60,15 +61,19 @@ pub(super) async fn send_thread_token_usage_update_to_connection(
 
 pub(super) async fn latest_token_usage_turn_id_for_thread_path(thread: &Thread) -> Option<String> {
     let rollout_path = thread.path.as_deref()?;
-    latest_token_usage_turn_id_from_rollout_path(rollout_path, thread).await
+    latest_token_usage_turn_id_from_rollout_path(
+        rollout_path,
+        (!thread.turns.is_empty()).then_some(thread.turns.as_slice()),
```
bolinfest: I would maybe keep it simple and just always pass a slice and concede it could be empty?

The current design really has three possible states:

* `None`
* `Some` with an empty slice
* `Some` with a non-empty slice

But I think the first two states are always treated the same, so why not collapse them? Given that `latest_token_usage_turn_id_from_rollout_items()` is `pub(super)`, you can't guarantee that `Some` with an empty slice is never passed in.

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3296,6 +3296,11 @@ pub struct ThreadResumeParams {
     pub developer_instructions: Option<String>,
     #[ts(optional = nullable)]
     pub personality: Option<Personality>,
+    /// When false, return only thread metadata and live-resume state without
```
bolinfest: Can we rename this so unset is the same as `false`? It's always a little weird when unset means `true`.

bolinfest: Maybe `exclude_turns` then or `exclude_turn_metadata`?

codex-rs/app-server/src/codex_message_processor/token_usage_replay.rs
```diff
@@ -113,7 +105,22 @@ fn latest_token_usage_turn_owner_from_rollout_items(
...
+        rebuilt_turns.as_slice()
+    };
+
+    if turns.iter().any(|turn| turn.id == owner.id) {
+        return Some(owner.id);
+    }
+    owner
+        .position
+        .and_then(|position| turns.get(position))
+        .map(|turn| turn.id.clone())
```
bolinfest: Prefer:

```suggestion
    if turns.iter().any(|turn| turn.id == owner.id) {
        Some(owner.id)
    } else {
        owner
            .position
            .and_then(|position| turns.get(position))
            .map(|turn| turn.id.clone())
    }
```

codex-rs/app-server/README.md
```diff
@@ -269,6 +271,12 @@ Example:
     "personality": "friendly"
 } }
 { "id": 11, "result": { "thread": { "id": "thr_123", … } } }
+
+{ "method": "thread/resume", "id": 12, "params": {
+    "threadId": "thr_123",
+    "includeTurns": false
```
bolinfest: Also has to be updated assume you rename.

bolinfest (approved):

PR #18919, opened 2026-04-22
[app-server] type client responses

rhan-oai:
## Why

`pr17088` adds a typed seam for server-originated request/response traffic. Client-originated successful responses still lost their protocol variant once app-server serialized them into bare JSON-RPC `result` payloads, which forced analytics to rely on handler-local response tracking when it needed typed lifecycle data.

This PR keeps successful client responses typed until the common outgoing-response seam, so app-server can record them centrally while analytics still decides which response kinds it cares about downstream.

## What changed

- Adds `ClientResponsePayload` in `codex-rs/app-server-protocol` so successful client responses can carry their typed variant until JSON-RPC serialization.
- Routes app-server success paths through `ClientResponsePayload`, including `send_response` and `send_result`, instead of passing arbitrary serializable values to the outgoing sender.
- Records typed successful client-response payloads centrally from `OutgoingMessageSender`, while keeping `InterruptConversation` JSON-RPC-only because it does not correspond to a `ClientResponse` variant.
- Generalizes initialized client-request tracking in `MessageProcessor`, then filters request and response ingestion in `AnalyticsEventsClient` so only the currently relevant turn/thread variants enter the analytics queue.
- Adds a `ResponsePayload` analytics fact that converts the preserved payload back into `ClientResponse` inside the reducer, plus tests covering payload conversion and ignored unrelated request/response variants.

## Verification

- `cargo check -p codex-app-server`
- `cargo test -p codex-analytics`
- `cargo test -p codex-app-server outgoing_message::tests --lib`
- `cargo test -p codex-app-server-protocol client_response_payload --lib`
- `cargo test -p codex-app-server-protocol interrupt_conversation_payload_stays_jsonrpc_only --lib`

PR #19035, opened 2026-04-22
type client responses

rhan-oai:


rhan-oai:
Closing because this change belongs on #18919.

PR #19036, opened 2026-04-22
Overlay state DB git metadata for filtered thread lists

joeytrasatti-openai:
## Summary
- Factor the state DB `ThreadMetadata` to rollout `ThreadItem` mapping into a shared helper used by both DB pages and filesystem overlays
- Generalize filtered filesystem list overlays to fill missing thread list metadata from the state-derived `ThreadItem`, while preserving filesystem `path` and `thread_id`
- Add coverage for the merge behavior so existing filesystem values are not overwritten and future `ThreadItem` fields require an explicit decision

## Testing
- `just fmt` from `codex-rs`
- `git diff --check -- codex-rs/rollout/src/recorder.rs codex-rs/rollout/src/recorder_tests.rs`
- Attempted `cargo test -p codex-rollout thread_item_metadata` from `codex-rs`; blocked in dependency fetch/setup after updating crates.io and git submodules `https://github.com/livekit/protocol` and `https://chromium.googlesource.com/libyuv/libyuv`, so the focused tests did not run

### DIFF
--- a/codex-rs/rollout/src/recorder.rs
+++ b/codex-rs/rollout/src/recorder.rs
@@ -528,22 +528,16 @@ impl RolloutRecorder {
                     )
                     .await;
                 }
-                return Ok(page_from_filesystem_scan(
-                    fs_page,
-                    sort_direction,
-                    page_size,
-                    sort_key,
-                ));
+                let page = page_from_filesystem_scan(fs_page, sort_direction, page_size, sort_key);
+                return Ok(
+                    fill_missing_git_info_from_state_db(state_db_ctx.as_deref(), page).await,
+                );
             }
             return Ok(db_page.into());
         }
         if listing_has_metadata_filters {
-            return Ok(page_from_filesystem_scan(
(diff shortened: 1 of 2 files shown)


acrognale-oai (approved):

codex-rs/rollout/src/recorder.rs
```diff
@@ -978,6 +978,90 @@ fn page_from_filesystem_scan(
...
+                continue;
+            }
+        };
+        fill_missing_thread_item_metadata(item, thread_item_from_state_metadata(metadata));
+    }
+
+    page
+}
+
+fn fill_missing_thread_item_metadata(item: &mut ThreadItem, state_item: ThreadItem) {
```
jif-oai, line 1009: this looks bad and probably hide a bad design

jif-oai (commented):
lgtm. The bad design must be fixed but can be in a follow-up

jif-oai (approved):

PR #19040, opened 2026-04-22
[codex] Add unsandboxed process exec API

euroelessar:
## Why

App-server clients sometimes need argv-based local process execution while sandbox policy is controlled outside Codex. Those environments can reject sandbox-disabling paths before a command ever starts, even when the caller intentionally wants unsandboxed execution.

This PR adds a distinct `process/*` API for that use case instead of extending `command/exec` with another sandbox-disabling shape. Keeping the new surface separate also makes the future removal of `command/exec` simpler: clients that need explicit process lifecycle control can move to the newer handle-based API without depending on `command/exec` business logic.

## What changed

- Added v2 process lifecycle methods: `process/spawn`, `process/writeStdin`, `process/resizePty`, and `process/kill`.
- Added process notifications: `process/outputDelta` for streamed stdout/stderr chunks and `process/exited` for final exit status and buffered output.
- Made `process/spawn` intentionally unsandboxed and omitted sandbox-selection fields such as `sandboxPolicy` and `permissionProfile`.
- Added client-supplied, connection-scoped `processHandle` values for follow-up control requests and notification routing.
- Supported cwd, environment overrides, PTY mode and size, stdin streaming, stdout/stderr streaming, per-stream output caps, and timeout controls.
- Killed active process sessions when the originating app-server connection closes.
- Wired the implementation through the modular `request_processors/` app-server layout, with process-handle request serialization for follow-up control calls.
- Updated generated JSON/TypeScript schema fixtures and documented the new API in `codex-rs/app-server/README.md`.
- Added v2 app-server integration coverage in `codex-rs/app-server/tests/suite/v2/process_exec.rs` for spawn acknowledgement before exit, buffered output caps, and process termination.

## Verification


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3031,6 +3031,81 @@ pub struct CommandExecParams {
     pub sandbox_policy: Option<SandboxPolicy>,
 }
 
+/// Run a standalone command (argv vector) without a sandbox or thread/turn.
+///
+/// The final `command/execUnsandboxed` response is deferred until the process
```
bolinfest: I want to call this something that is distinct from `command/exec` and does not overlap with our shell tool names. We have `fs/` as a namespace already, so we could do `fs/spawn`, but perhaps we should do something more like Node.js and support `process/` with:

```
process/spawn
process/write_stdin
process/kill
process/resize_pty
```

Thoughts?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3031,6 +3031,81 @@ pub struct CommandExecParams {
...
+    /// When omitted, the server default applies. Cannot be combined with
+    /// `disableOutputCap`.
+    #[ts(type = "number | null")]
+    #[ts(optional = nullable)]
+    pub output_bytes_cap: Option<usize>,
+    /// Disable stdout/stderr capture truncation for this request.
+    ///
+    /// Cannot be combined with `outputBytesCap`.
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub disable_output_cap: bool,
```
bolinfest: Should we support these or simplify?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3031,6 +3031,81 @@ pub struct CommandExecParams {
...
+    /// Cannot be combined with `timeoutMs`.
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub disable_timeout: bool,
+    /// Optional timeout in milliseconds.
+    ///
+    /// When omitted, the server default applies. Cannot be combined with
+    /// `disableTimeout`.
+    #[ts(type = "number | null")]
+    #[ts(optional = nullable)]
+    pub timeout_ms: Option<i64>,
```
bolinfest: Why do we need both?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3031,6 +3031,81 @@ pub struct CommandExecParams {
...
+    /// Optional working directory. Defaults to the server cwd.
+    #[ts(optional = nullable)]
+    pub cwd: Option<PathBuf>,
```
bolinfest: Can we make this required? And use `AbsolutePathBuf`?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3031,6 +3031,81 @@ pub struct CommandExecParams {
...
+    /// Command argv vector. Empty arrays are rejected.
+    pub command: Vec<String>,
+    /// Optional client-supplied, connection-scoped process id.
+    ///
+    /// Required for `tty`, `streamStdin`, `streamStdoutStderr`, and follow-up
+    /// `command/exec/write`, `command/exec/resize`, and
+    /// `command/exec/terminate` calls. When omitted, buffered execution gets an
+    /// internal id that is not exposed to the client.
+    #[ts(optional = nullable)]
+    pub process_id: Option<String>,
```
bolinfest: What do you think about `process_handle`? I find it confusing that this is not a `PID`.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -658,6 +659,29 @@ fn configured_thread_store(config: &Config) -> Arc<dyn ThreadStore> {
...
+    output_bytes_cap: Option<usize>,
+    disable_output_cap: bool,
+    disable_timeout: bool,
+    timeout_ms: Option<i64>,
+    cwd: Option<PathBuf>,
+    env_overrides: Option<HashMap<String, Option<String>>>,
+    size: Option<codex_app_server_protocol::CommandExecTerminalSize>,
+}
+
+enum CommandExecMode {
```
bolinfest: Can we do something distinct and simple? It will make it much easier to remove the old codepath later if we decide we no longer have to support it.

codex-rs/app-server-protocol/src/protocol/v2.rs
euroelessar: output cap is needed, e.g. some git results can be gigantic

codex-rs/app-server-protocol/src/protocol/v2.rs
owenlin0: yes, i like this. i refactored to this API

codex-rs/app-server-protocol/src/protocol/v2.rs
owenlin0: removed `disable_output_cap` and `disable_timeout` to simplify but let me know if you think we need these?

euroelessar: we can merge them into non-disable fields (as we don't need backward compatibility & can have nicer api), but we still want to be able to disable them (e.g. by passing `null` or making them disabled by default) for the terminal use case

codex-rs/app-server/src/codex_message_processor.rs
owenlin0: yes, as part of introducing the `process/*` APIs, it also has its own implementation

codex-rs/app-server-protocol/src/protocol/v2.rs
owenlin0: sg, i'll use explicit `null` to mean to mean no output cap / timeout

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3558,6 +3558,203 @@ pub enum CommandExecOutputStream {
     Stderr,
 }
 
+/// PTY size in character cells for `process/spawn` PTY sessions.
```
bolinfest, line 3561: Incidentally, we should probably start to break up this file...

If we need the content defined in this file, we could use `#[path = "process.rs"]` though ideally we would just use the more standard `mod process`?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3558,6 +3558,203 @@ pub enum CommandExecOutputStream {
...
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ProcessTerminalSize {
+    /// Terminal height in character cells.
+    pub rows: u16,
+    /// Terminal width in character cells.
+    pub cols: u16,
+}
+
+/// Spawn a standalone process (argv vector) without a Codex sandbox.
```
bolinfest: I think it merits being extra explicit here.

```suggestion
/// Spawn a standalone process (argv vector) without a Codex sandbox on the host where the app server is running.
```

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3558,6 +3558,203 @@ pub enum CommandExecOutputStream {
...
+/// `processHandle` has been registered. Process output and exit are reported via
+/// `process/outputDelta` and `process/exited` notifications.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ProcessSpawnParams {
+    /// Command argv vector. Empty arrays are rejected.
+    pub command: Vec<String>,
+    /// Client-supplied, connection-scoped process handle.
+    pub process_handle: String,
```
bolinfest, line 3588: What happens if we get multiple requests with the same value here?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3558,6 +3558,203 @@ pub enum CommandExecOutputStream {
...
+    pub size: Option<ProcessTerminalSize>,
+}
+
+/// Successful response for `process/spawn`.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ProcessSpawnResponse {
+    /// Client-supplied, connection-scoped process handle.
+    pub process_handle: String,
```
bolinfest: Should we be sending it back?

codex-rs/app-server/src/request_processors/process_exec_processor.rs
```diff
@@ -0,0 +1,742 @@
...
+            outgoing,
+            process_exec_manager: ProcessExecManager::default(),
+        }
+    }
+
+    pub(crate) async fn process_spawn(
+        &self,
+        request_id: ConnectionRequestId,
+        params: ProcessSpawnParams,
+    ) -> Result<Option<ClientResponsePayload>, JSONRPCErrorError> {
```
bolinfest: Why is this the return type if it seems it can only return `Ok(None)`?

codex-rs/app-server/src/request_processors/process_exec_processor.rs
```diff
@@ -0,0 +1,742 @@
...
+            return Err(invalid_request("command must not be empty"));
+        }
+        if process_handle.is_empty() {
+            return Err(invalid_request("processHandle must not be empty"));
+        }
+        if size.is_some() && !tty {
+            return Err(invalid_params("process/spawn size requires tty: true"));
+        }
+        let mut env = create_env(
+            &self.config.permissions.shell_environment_policy,
```
bolinfest: I don't think `shell_environment_policy` should apply, should it? That's arguably designed to govern shell tools only?

codex-rs/app-server/src/request_processors/process_exec_processor.rs
```diff
@@ -0,0 +1,742 @@
...
+            let mut sessions = self.sessions.lock().await;
+            if sessions.contains_key(&process_key) {
+                return Err(invalid_request(format!(
+                    "duplicate active process handle: {}",
+                    serde_json::to_string(&process_handle)
+                        .unwrap_or_else(|_| format!("{process_handle:?}")),
+                )));
+            }
+            sessions.insert(process_key.clone(), ProcessSession { control_tx });
+        }
```
bolinfest, line 290: Ah, so this is where you check if the handle is reused: great!

Though also, prefer `entry()` so that you only do the lookup once, especially since you're holding a lock here:

```suggestion
{
    let mut sessions = self.sessions.lock().await;

    match sessions.entry(process_key.clone()) {
        std::collections::hash_map::Entry::Occupied(_) => {
            return Err(invalid_request(format!(
                "duplicate active process handle: {}",
                serde_json::to_string(&process_handle)
                    .unwrap_or_else(|_| format!("{process_handle:?}")),
            )));
        }
        std::collections::hash_map::Entry::Vacant(entry) => {
            entry.insert(ProcessSession { control_tx });
        }
    }
}
```

bolinfest: I would also consider skipping the `serde_json::to_string`: the debug version of `process_handle` should be sufficient for the caller to debug their logic issue.

codex-rs/app-server/src/message_processor.rs
```diff
@@ -1236,6 +1244,26 @@ impl MessageProcessor {
                     .command_exec_terminate(request_id.clone(), params)
                     .await
             }
+            ClientRequest::ProcessSpawn { params, .. } => {
```
bolinfest: Should all of these be dispatched to their own `tokio::task` here so that the main message processor is freed up? Also, I thought @euroelessar added some queues to help with this sort of thing?

codex-rs/app-server/tests/suite/v2/mod.rs
```diff
@@ -35,6 +35,8 @@ mod plugin_list;
 mod plugin_read;
 mod plugin_share;
 mod plugin_uninstall;
+#[cfg(unix)]
```
bolinfest: Shouldn't we test Windows, as well?

codex-rs/app-server/src/request_processors/process_exec_processor.rs
bolinfest: Also, I think this is the only reference to `self.config`, so we can potentially remove that field? Or if we decide we want this, I would store `shell_environment_policy` as a field instead of `Arc<Config>`.

codex-rs/app-server/src/request_processors/process_exec_processor.rs
```diff
@@ -0,0 +1,742 @@
...
+        &self,
+        connection_id: ConnectionId,
+        process_handle: String,
+        control: ProcessControl,
+    ) -> Result<(), JSONRPCErrorError> {
+        let process_key = ConnectionProcessHandle {
+            connection_id,
+            process_handle,
+        };
+        let session = {
```
bolinfest: You can omit the braces here.

codex-rs/app-server/src/request_processors/process_exec_processor.rs
```diff
@@ -0,0 +1,742 @@
...
+                            ProcessControl::Resize { size } => {
+                                handle_process_resize(&session, size)
+                            }
+                            ProcessControl::Kill => {
+                                session.request_terminate();
+                                Ok(())
+                            }
+                        };
+                        if let Some(response_tx) = response_tx {
+                            let _ = response_tx.send(result);
```
bolinfest: should we log if `Err`?

codex-rs/app-server/src/request_processors/process_exec_processor.rs
```diff
@@ -0,0 +1,742 @@
...
+                if matches!(expiration_outcome, Some(ExecExpirationOutcome::TimedOut)) {
+                    break EXEC_TIMEOUT_EXIT_CODE;
+                } else {
+                    break exit.unwrap_or(-1);
+                }
+            }
+        }
+    };
+
+    let timeout_handle = tokio::spawn(async move {
```
bolinfest, line 560: Add a comment here?

codex-rs/app-server/src/request_processors/process_exec_processor.rs
```diff
@@ -0,0 +1,742 @@
...
+                stdout: stdout.text,
+                stdout_cap_reached: stdout.cap_reached,
+                stderr: stderr.text,
+                stderr_cap_reached: stderr.cap_reached,
+            }),
+        )
+        .await;
+}
+
+fn spawn_process_output(
```
bolinfest: `collect_spawn_process_output()`?

codex-rs/app-server/src/request_processors/process_exec_processor.rs
```diff
@@ -0,0 +1,742 @@
...
+    if !delta.is_empty() {
+        session
+            .writer_sender()
+            .send(delta)
+            .await
+            .map_err(|_| invalid_request("stdin is already closed"))?;
+    }
+    if close_stdin {
+        session.close_stdin();
+    }
```
bolinfest, line 673: Is it correct that `writer_sender().send(delta)` sends a request to write to stdin, but there is no "ack" (so we don't know if stdin has been written yet) and then `session.close_stdin()` is called synchronously such that stdin could be closed before the async stdin write request has been processed?

Should close_stdin also be a request that is queued in the same way to `ProcessHandle`?

codex-rs/app-server/src/request_processors/process_exec_processor.rs
```diff
@@ -0,0 +1,742 @@
...
+) -> Result<(), JSONRPCErrorError> {
+    session
+        .resize(size)
+        .map_err(|err| invalid_request(format!("failed to resize PTY: {err}")))
+}
+
+fn terminal_size_from_protocol(
+    size: ProcessTerminalSize,
+) -> Result<TerminalSize, JSONRPCErrorError> {
+    if size.rows == 0 || size.cols == 0 {
```
bolinfest, line 689: Consider `std::num::NonZero`?

codex-rs/app-server/src/message_processor.rs
euroelessar: all rpc handlers are dispatched nowadays into their own task (or queue, if there is a serialization key)

codex-rs/app-server/src/request_processors/process_exec_processor.rs
owenlin0: from codex:
> The current ordering is safe because `send(delta).await` enqueues the bytes before `close_stdin()` drops our sender, and the writer task drains queued bytes before its receiver observes EOF and closes stdin. We don’t currently ack OS-level write/flush completion, which matches the existing command/exec behavior.

added a comment here to explain this for future readers though

codex-rs/app-server/src/request_processors/process_exec_processor.rs
owenlin0: IMO let's keep as is - so that it's a simple `number` in the API schema but we validate on the server

codex-rs/app-server-protocol/src/protocol/v2.rs
owenlin0: agreed... let's punt to a separate PR that mechanically splits up `v2.rs`?

codex-rs/app-server/src/request_processors/process_exec_processor.rs
```diff
@@ -0,0 +1,708 @@
...
+        if command.is_empty() {
+            return Err(invalid_request("command must not be empty"));
+        }
+        if process_handle.is_empty() {
+            return Err(invalid_request("processHandle must not be empty"));
+        }
+        if size.is_some() && !tty {
+            return Err(invalid_params("process/spawn size requires tty: true"));
+        }
+        let mut env = std::env::vars().collect::<HashMap<_, _>>();
```
bolinfest, line 90: Does this build on `std::env`? The caller shouldn't have to know the `PATH` on the app server host, right?

bolinfest (approved):
nice job: ty!

codex-rs/app-server/src/request_processors/process_exec_processor.rs
owenlin0: yep it does!

PR #19016, opened 2026-04-22
exec-server: expose arg0 alias root to fs sandbox

bolinfest:
## Why

The post-merge `rust-ci-full` run for #18999 still failed the Ubuntu remote `suite::remote_env` sandboxed filesystem tests. That run checked out merge commit `ddde50c611e4800cb805f243ed3c50bbafe7d011`, so the arg0 guard lifetime fix was present.

The Docker-backed failure had two remaining pieces:

- The sandboxed filesystem helper needs to execute Codex through the `codex-linux-sandbox` arg0 alias path. The helper sandbox was only granting read access to the real Codex executable parent, so the alias parent also has to be visible inside the helper sandbox.
- The remote-env tests were building sandbox contexts with `FileSystemSandboxContext::new()`, which captures the local test runner cwd. In the Docker remote exec-server, that host checkout path does not exist, so spawning the filesystem helper failed with `No such file or directory` before the helper could process the request.

While validating the PR, the Windows Bazel unit-test shard also exposed a marginal 2s timing assertion in `output_and_exit_are_retained_after_notification_receiver_closes`. The test command intentionally waits before emitting output on Windows, so the old deadline could fail under CI load even when the behavior was correct.

## What Changed

- Track all helper runtime read roots instead of a single root.
- Add both the real Codex executable parent and the `codex-linux-sandbox` alias parent to sandbox readable roots.
- Avoid sending an unused local cwd in remote filesystem sandbox contexts when the permission profile has no cwd-dependent entries.
- Build the Docker remote-env test sandbox contexts with a cwd path that exists inside the container.
- Relax the exec-server handler test close deadline from 2s to 5s.
- Add unit coverage for the alias-parent root and remote sandbox cwd handling.

## Verification

- `cargo test -p codex-exec-server`
- `cargo test -p codex-exec-server output_and_exit_are_retained_after_notification_receiver_closes`
- `cargo test -p codex-core remote_test_env_sandboxed_read_allows_readable_root`
- `just fix -p codex-exec-server`
- `just fix -p codex-core`

PR #19046, opened 2026-04-22
exec-server: require explicit filesystem sandbox cwd

bolinfest:
## Why

This is a cleanup PR for the `PermissionProfile` migration stack. #19016 fixed remote exec-server sandbox contexts so Docker-backed filesystem requests use a request/container `cwd` instead of leaking the local test runner `cwd`. That exposed the broader API problem: `FileSystemSandboxContext::new(SandboxPolicy)` could still reconstruct filesystem permissions by reading the exec-server process cwd with `AbsolutePathBuf::current_dir()`.

That made `cwd`-dependent legacy entries, such as `:cwd`, `:project_roots`, and relative deny globs, depend on ambient process state instead of the request sandbox `cwd`. As later PRs make `PermissionProfile` the primary permissions abstraction, sandbox contexts should be explicit about whether they carry a request `cwd` or are profile-only. Removing the implicit constructor prevents new call sites from accidentally rebuilding permissions against the wrong `cwd`.

## What changed

- Removed `FileSystemSandboxContext::new(SandboxPolicy)`.
- Kept production callers on explicit constructors: `from_legacy_sandbox_policy(..., cwd)`, `from_permission_profile(...)`, and `from_permission_profile_with_cwd(...)`.
- Updated exec-server test helpers to construct `PermissionProfile` values directly instead of routing through legacy `SandboxPolicy` projections.
- Updated the environment regression test to use an explicit restricted profile with no synthetic `cwd`.

## Verification

- `just fix -p codex-exec-server`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/19046).
* #18288
* #18287
* #18286
* #18285
* #18284
* #18283
* #18282
* #18281
* #18280
* __->__ #19046

### DIFF
--- a/codex-rs/exec-server/src/environment.rs
+++ b/codex-rs/exec-server/src/environment.rs
@@ -443,8 +443,11 @@ mod tests {
             std::env::current_exe().expect("current exe").as_path(),
         )
         .expect("absolute current exe");
-        let sandbox = crate::FileSystemSandboxContext::new(
-            codex_protocol::protocol::SandboxPolicy::new_read_only_policy(),
+        let sandbox = crate::FileSystemSandboxContext::from_permission_profile(
+            codex_protocol::models::PermissionProfile::from_runtime_permissions(
+                &codex_protocol::permissions::FileSystemSandboxPolicy::restricted(Vec::new()),
+                codex_protocol::permissions::NetworkSandboxPolicy::Restricted,
+            ),
         );
 
         let err = environment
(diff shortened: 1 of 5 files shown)


starr-openai (approved):
cc @viyatb-oai since you were also working on this recently

viyatb-oai (approved):

PR #19047, opened 2026-04-22
feat: add run task identity primitives

adrian-openai:
## Stack

This is PR 1 of the simplified HAI single-run-task stack:

## Summary

The simplified POC shape is one backend-owned task per Agent Identity run. This PR makes the first layer match that final shape directly instead of introducing task targets, caller-owned external task refs, or intermediate wrappers that later PRs would need to undo.

What changed:

- keeps the `AgentAssertion` wire payload as `agent_runtime_id`, `task_id`, `timestamp`, and `signature`
- exposes `register_agent_task` as the single task-registration helper for both existing Agent Identity JWT auth and the ChatGPT-registration path added later in the stack
- makes task registration send only the signed registration timestamp; the backend owns the returned opaque task id
- removes the unused target/task-kind/external-task-ref surfaces from `codex-agent-identity`
- keeps Agent Identity JWT JWKS lookup separate from agent/task registration URL derivation
- updates Agent Identity JWT auth to register one run task during auth construction and share that task across cloned auth handles

This PR intentionally does not enable ChatGPT-derived Agent Identity. That opt-in and config gate are added in the next PR.

## Testing

- `just test -p codex-agent-identity`

### DIFF
--- a/codex-rs/agent-identity/src/lib.rs
+++ b/codex-rs/agent-identity/src/lib.rs
@@ -34,6 +34,7 @@ const AGENT_TASK_REGISTRATION_TIMEOUT: Duration = Duration::from_secs(30);
 const AGENT_IDENTITY_JWKS_TIMEOUT: Duration = Duration::from_secs(10);
 const AGENT_IDENTITY_JWT_AUDIENCE: &str = "codex-app-server";
 const AGENT_IDENTITY_JWT_ISSUER: &str = "https://chatgpt.com/codex-backend/agent-identity";
+const AGENT_REGISTRATION_TIMEOUT: Duration = Duration::from_secs(15);
 
 /// Stored key material for a registered agent identity.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
@@ -47,6 +48,133 @@ pub struct AgentIdentityKey<'a> {
 pub struct AgentTaskAuthorizationTarget<'a> {
     pub agent_runtime_id: &'a str,
     pub task_id: &'a str,
+}
+
+/// Runtime identity that owns one or more registered agent tasks.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct AgentRuntimeId(String);
+
+impl AgentRuntimeId {
+    pub fn new(agent_runtime_id: impl Into<String>) -> Self {
+        Self(agent_runtime_id.into())
+    }
+
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    pub fn into_string(self) -> String {
+        self.0
+    }
+}
+
+impl From<String> for AgentRuntimeId {
+    fn from(agent_runtime_id: String) -> Self {
+        Self::new(agent_runtime_id)
+    }
+}
+
+impl From<&str> for AgentRuntimeId {
+    fn from(agent_runtime_id: &str) -> Self {
+        Self::new(agent_runtime_id)
+    }
+}
+
+/// Task identifier granted to an agent runtime for a scoped objective.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct AgentTaskId(String);
+
+impl AgentTaskId {
+    pub fn new(task_id: impl Into<String>) -> Self {
+        Self(task_id.into())
+    }
+
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    pub fn into_string(self) -> String {
+        self.0
+    }
+}
+
+impl From<String> for AgentTaskId {
+    fn from(task_id: String) -> Self {
+        Self::new(task_id)
+    }
+}
+
+impl From<&str> for AgentTaskId {
+    fn from(task_id: &str) -> Self {
+        Self::new(task_id)
+    }
+}
+
+/// Caller-owned stable reference that HAI can resolve to an opaque task id.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct AgentTaskExternalRef(String);
+
+impl AgentTaskExternalRef {
+    pub fn new(external_ref: impl Into<String>) -> Self {
+        Self(external_ref.into())
+    }
+
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    pub fn into_string(self) -> String {
+        self.0
+    }
+}
+
+impl From<String> for AgentTaskExternalRef {
+    fn from(external_ref: String) -> Self {
+        Self::new(external_ref)
+    }
+}
+
+impl From<&str> for AgentTaskExternalRef {
+    fn from(external_ref: &str) -> Self {
+        Self::new(external_ref)
+    }
+}
+
+/// Purpose of a registered task binding.
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum AgentTaskKind {
+    Thread,
+    Background,
+}
+
+/// Registered task binding used to authorize work for an agent runtime.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct RegisteredAgentTask {
+    pub agent_runtime_id: AgentRuntimeId,
+    pub task_id: AgentTaskId,
+    pub kind: AgentTaskKind,
+}
+
+impl RegisteredAgentTask {
+    pub fn new(
+        agent_runtime_id: impl Into<AgentRuntimeId>,
+        task_id: impl Into<AgentTaskId>,
+        kind: AgentTaskKind,
+    ) -> Self {
+        Self {
+            agent_runtime_id: agent_runtime_id.into(),
+            task_id: task_id.into(),
+            kind,
+        }
+    }
+
+    pub fn authorization_target(&self) -> AgentTaskAuthorizationTarget<'_> {
+        AgentTaskAuthorizationTarget {
+            agent_runtime_id: self.agent_runtime_id.as_str(),
+            task_id: self.task_id.as_str(),
+        }
+    }
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
@@ -86,9 +214,11 @@ struct AgentAssertionEnvelope {
 }
 
 #[derive(Serialize)]
-struct RegisterTaskRequest {
+struct RegisterTaskRequest<'a> {
     timestamp: String,
     signature: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    external_task_ref: Option<&'a str>,
 }
 
 #[derive(Deserialize)]
@@ -101,6 +231,18 @@ struct RegisterTaskResponse {
     encrypted_task_id: Option<String>,
     #[serde(default, rename = "encryptedTaskId")]
     encrypted_task_id_camel: Option<String>,
+}
+
+#[derive(Debug, Serialize)]
+struct RegisterAgentRequest {
+    abom: AgentBillOfMaterials,
+    agent_public_key: String,
+    capabilities: Vec<String>,
+}
+
+#[derive(Debug, Deserialize)]
+struct RegisterAgentResponse {
+    agent_runtime_id: String,
 }
 
 pub fn authorization_header_for_agent_task(
@@ -123,6 +265,13 @@ pub fn authorization_header_for_agent_task(
     };
     let serialized_assertion = serialize_agent_assertion(&envelope)?;
     Ok(format!("AgentAssertion {serialized_assertion}"))
+}
+
+pub fn authorization_header_for_registered_task(
+    key: AgentIdentityKey<'_>,
+    task: &RegisteredAgentTask,
+) -> Result<String> {
+    authorization_header_for_agent_task(key, task.authorization_target())
 }
 
 pub async fn fetch_agent_identity_jwks(
@@ -198,10 +347,21 @@ pub async fn register_agent_task(
     chatgpt_base_url: &str,
     key: AgentIdentityKey<'_>,
 ) -> Result<String> {
+    register_agent_task_with_external_ref(client, chatgpt_base_url, key, /*external_ref*/ None)
+        .await
+}
+
+pub async fn register_agent_task_with_external_ref(
+    client: &reqwest::Client,
+    chatgpt_base_url: &str,
+    key: AgentIdentityKey<'_>,
+    external_ref: Option<&AgentTaskExternalRef>,
+) -> Result<String> {
     let timestamp = Utc::now().to_rfc3339_opts(SecondsFormat::Secs, true);
     let request = RegisterTaskRequest {
         signature: sign_task_registration_payload(key, &timestamp)?,
         timestamp,
+        external_task_ref: external_ref.map(AgentTaskExternalRef::as_str),
     };
     let url = agent_task_registration_url(chatgpt_base_url, key.agent_runtime_id);
 
@@ -229,6 +389,45 @@ pub async fn register_agent_task(
         .context("failed to decode agent task registration response")?;
 
     task_id_from_register_task_response(key, response)
+}
+
+pub async fn register_agent_identity(
+    client: &reqwest::Client,
+    chatgpt_base_url: &str,
+    access_token: &str,
+    account_id: &str,
+    is_fedramp_account: bool,
+    key_material: &GeneratedAgentKeyMaterial,
+    abom: AgentBillOfMaterials,
+) -> Result<AgentRuntimeId> {
+    let url = agent_registration_url(chatgpt_base_url);
+    let request = RegisterAgentRequest {
+        abom,
+        agent_public_key: key_material.public_key_ssh.clone(),
+        capabilities: Vec::new(),
+    };
+
+    let mut request_builder = client
+        .post(&url)
+        .bearer_auth(access_token)
+        .header("ChatGPT-Account-ID", account_id)
+        .json(&request)
+        .timeout(AGENT_REGISTRATION_TIMEOUT);
+    if is_fedramp_account {
+        request_builder = request_builder.header("X-OpenAI-Fedramp", "true");
+    }
+
+    let response = request_builder
+        .send()
+        .await
+        .with_context(|| format!("failed to send agent identity registration request to {url}"))?
+        .error_for_status()
+        .with_context(|| format!("agent identity registration failed for {url}"))?
+        .json::<RegisterAgentResponse>()
+        .await
+        .with_context(|| format!("failed to parse agent identity response from {url}"))?;
+
+    Ok(AgentRuntimeId::new(response.agent_runtime_id))
 }
 
 fn task_id_from_register_task_response(
@@ -306,11 +505,6 @@ pub fn agent_task_registration_url(chatgpt_base_url: &str, agent_runtime_id: &st
     format!("{trimmed}/v1/agent/{agent_runtime_id}/task/register")
 }
 
-pub fn agent_identity_biscuit_url(chatgpt_base_url: &str) -> String {
-    let trimmed = chatgpt_base_url.trim_end_matches('/');
-    format!("{trimmed}/authenticate_app_v2")
-}
-
 pub fn agent_identity_jwks_url(chatgpt_base_url: &str) -> String {
     let trimmed = chatgpt_base_url.trim_end_matches('/');
     if trimmed.contains("/backend-api") {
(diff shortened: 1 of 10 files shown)


nicksteele-oai (approved):
This looks fine to me

tigrantsat-openai (commented):
Let's ensure full set of features + launching internally before enabling into codex.

codex-rs/agent-identity/src/lib.rs
```diff
@@ -49,6 +50,133 @@ pub struct AgentTaskAuthorizationTarget<'a> {
...
+
+impl From<&str> for AgentTaskId {
+    fn from(task_id: &str) -> Self {
+        Self::new(task_id)
+    }
+}
+
+/// Caller-owned stable reference that HAI can resolve to an opaque task id.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct AgentTaskExternalRef(String);
```
adrian-openai: Do we still need an external ref if we are just generating a single task? If we do, what are we pinning here?

codex-rs/agent-identity/src/lib.rs
```diff
@@ -49,6 +50,133 @@ pub struct AgentTaskAuthorizationTarget<'a> {
...
+impl From<&str> for AgentTaskExternalRef {
+    fn from(external_ref: &str) -> Self {
+        Self::new(external_ref)
+    }
+}
+
+/// Purpose of a registered task binding.
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum AgentTaskKind {
```
adrian-openai: If we only have a single task at all, do we still need this?

codex-rs/agent-identity/src/lib.rs
```diff
@@ -197,11 +346,22 @@ pub async fn register_agent_task(
     client: &reqwest::Client,
     chatgpt_base_url: &str,
     key: AgentIdentityKey<'_>,
+) -> Result<String> {
+    register_agent_task_with_external_ref(client, chatgpt_base_url, key, /*external_ref*/ None)
```
adrian-openai: Also probably no longer needed, no?

codex-rs/agent-identity/src/lib.rs
```diff
@@ -231,6 +260,45 @@ pub async fn register_agent_task(
...
+        capabilities: Vec::new(),
+    };
+
+    let mut request_builder = client
+        .post(&url)
+        .bearer_auth(access_token)
+        .header("ChatGPT-Account-ID", account_id)
+        .json(&request)
+        .timeout(AGENT_REGISTRATION_TIMEOUT);
+    if is_fedramp_account {
```
adrian-openai, line 377: What is this?

codex-rs/agent-identity/src/lib.rs
```diff
@@ -296,39 +341,82 @@ pub fn curve25519_secret_key_from_private_key_pkcs8_base64(
...
 }
 
-pub fn agent_identity_biscuit_url(chatgpt_base_url: &str) -> String {
-    let trimmed = chatgpt_base_url.trim_end_matches('/');
-    format!("{trimmed}/authenticate_app_v2")
-}
-
-pub fn agent_identity_jwks_url(chatgpt_base_url: &str) -> String {
-    let trimmed = chatgpt_base_url.trim_end_matches('/');
+pub fn agent_identity_jwks_url(agent_identity_jwt_base_url: &str) -> String {
```
Zemnmez, line 480: i think in future this should be the standard issuer derivation (i.e. https://[agent_id].something.something/.well-known) so this doesn't have to be so specific

codex-rs/agent-identity/src/lib.rs
```diff
@@ -103,34 +103,40 @@ struct RegisterTaskResponse {
...
-
     let timestamp = Utc::now().to_rfc3339_opts(SecondsFormat::Secs, true);
     let envelope = AgentAssertionEnvelope {
-        agent_runtime_id: target.agent_runtime_id.to_string(),
-        task_id: target.task_id.to_string(),
+        agent_runtime_id: key.agent_runtime_id.to_string(),
+        task_id: task_id.to_string(),
         timestamp: timestamp.clone(),
-        signature: sign_agent_assertion_payload(key, target.task_id, &timestamp)?,
+        signature: sign_agent_assertion_payload(key, task_id, &timestamp)?,
```
Zemnmez, line 238: In future let's remove the signature collision via `JSON.stringify([ field_1, field_2, field_3 ])` or something to that effect.

codex-rs/agent-identity/src/lib.rs
```diff
@@ -296,39 +341,82 @@ pub fn curve25519_secret_key_from_private_key_pkcs8_base64(
...
-pub fn agent_task_registration_url(chatgpt_base_url: &str, agent_runtime_id: &str) -> String {
-    let trimmed = chatgpt_base_url.trim_end_matches('/');
-    format!("{trimmed}/v1/agent/{agent_runtime_id}/task/register")
+pub fn agent_task_registration_url(
+    agent_identity_authapi_base_url: &str,
+    agent_runtime_id: &str,
+) -> String {
+    agent_identity_authapi_url(
+        agent_identity_authapi_base_url,
+        &format!("/v1/agent/{agent_runtime_id}/task/register"),
```
Zemnmez, line 476: the agent_runtime_id needs to be escaped, clamped or have other URL parts stripped

codex-rs/agent-identity/src/lib.rs
```diff
@@ -296,39 +341,82 @@ pub fn curve25519_secret_key_from_private_key_pkcs8_base64(
...
-    OsRng
-        .try_fill_bytes(&mut request_id_bytes)
-        .context("failed to generate agent identity request id")?;
-    Ok(format!(
-        "codex-agent-identity-{}",
-        URL_SAFE_NO_PAD.encode(request_id_bytes)
-    ))
+fn agent_identity_authapi_url(agent_identity_authapi_base_url: &str, api_path: &str) -> String {
+    let base_url = normalize_agent_identity_authapi_base_url(agent_identity_authapi_base_url);
+    format!("{base_url}{api_path}")
```
Zemnmez, line 491: would be great to do this via a proper URL construction rather than a naiive string concatenation

codex-rs/agent-identity/src/lib.rs
```diff
@@ -296,39 +341,82 @@ pub fn curve25519_secret_key_from_private_key_pkcs8_base64(
...
+}
+
+pub fn agent_identity_authapi_base_url_from_chatgpt_base_url(chatgpt_base_url: &str) -> String {
+    let mut base_url = chatgpt_base_url.trim_end_matches('/').to_string();
+    for suffix in [
```
Zemnmez: this seems very brittle but is probably fine for v0

codex-rs/agent-identity/src/lib.rs
```diff
@@ -296,39 +341,82 @@ pub fn curve25519_secret_key_from_private_key_pkcs8_base64(
...
+        "/wham/remote/control/server/enroll",
+        "/wham/remote/control/server",
+    ] {
+        if let Some(stripped) = base_url.strip_suffix(suffix) {
+            base_url = stripped.to_string();
+            break;
+        }
+    }
+    if matches!(
```
Zemnmez: perhaps we can build all these mappings statically? so there's no ambiguity at runtime?

Zemnmez (approved):

codex-rs/agent-identity/src/lib.rs
```diff
@@ -296,39 +420,82 @@ pub fn curve25519_secret_key_from_private_key_pkcs8_base64(
...
+    }
+}
+
+fn normalize_agent_identity_authapi_base_url(base_url: &str) -> String {
+    let trimmed = base_url.trim_end_matches('/');
+    if matches!(
+        trimmed,
+        "https://auth.openai.com" | "https://auth.api.openai.org"
+    ) {
+        format!("{trimmed}/api/accounts")
```
celia-oai: hmm, I'm not a huge fan of hiding this meaningful url rewrite logic in a function called "normalize...".  Could we just pass in the correct base_url in every callsite?

codex-rs/agent-identity/src/lib.rs
adrian-openai: This is going to be swapped out for a separate endpoint in the near_ish future.

codex-rs/agent-identity/src/lib.rs
adrian-openai: This is going to be swapped out for a different kind of header construct in the near-ish future.

codex-rs/agent-identity/src/lib.rs
adrian-openai: Agreed! This is something that we're working on.

codex-rs/agent-identity/src/lib.rs
adrian-openai: Agreed! This is getting cleaned up in the future as we standardize the paths which this will go to (and move things to the token endpoint).

codex-rs/agent-identity/src/lib.rs
adrian-openai: Agreed!

codex-rs/agent-identity/src/lib.rs
adrian-openai: This is going to be part of the cleanup here as we land onto the longer term token endpoints.

codex-rs/agent-identity/src/lib.rs
celia-oai: this + the above `agent_identity_authapi_base_url_from_chatgpt_base_url` logic doesn't give me huge confidence that this is not error prone. I can easily see `chatgpt_base_url` passed in is typoed and the translated url is just wrong. is the intention to support arbitrary chatgpt_base_url? if not, could we have an enum for all possible urls, something like:
```
enum ChatGptEnvironment {
    Production,
    Staging,
}

impl ChatGptEnvironment {
    fn chatgpt_base_url(&self) -> &'static str {
        match self {
            Self::Production => "https://chatgpt.com/backend-api",
            Self::Staging => "https://chatgpt-staging.com/backend-api",
        }
    }

    fn agent_identity_authapi_base_url(&self) -> &'static str {
        match self {
            Self::Production => "https://auth.openai.com/api/accounts",
            Self::Staging => "https://auth.api.openai.org/api/accounts",
        }
    }
}
```

codex-rs/agent-identity/src/lib.rs
```diff
@@ -260,10 +377,17 @@ pub fn decrypt_task_id_response(
 }
 
 pub fn generate_agent_key_material() -> Result<GeneratedAgentKeyMaterial> {
```
celia-oai, line 422: i don't fully understand what this change is doing here, but i trust you with it. might be good to explain what this code does vs. before more in in-line comment

codex-rs/login/src/auth/agent_identity.rs
```diff
@@ -1,43 +1,61 @@
...
-use std::env;
+use tokio::sync::OnceCell;
 
 use crate::default_client::build_reqwest_client;
 
 use super::storage::AgentIdentityAuthRecord;
 
-const PROD_AGENT_IDENTITY_AUTHAPI_BASE_URL: &str = "https://auth.openai.com/api/accounts";
-const CODEX_AGENT_IDENTITY_AUTHAPI_BASE_URL_ENV_VAR: &str = "CODEX_AGENT_IDENTITY_AUTHAPI_BASE_URL";
+const DEFAULT_CHATGPT_BACKEND_BASE_URL: &str = "https://chatgpt.com/backend-api";
```
celia-oai: what's the context behind url changes

codex-rs/login/src/auth/agent_identity.rs
```diff
@@ -1,43 +1,61 @@
...
 
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct AgentIdentityAuth {
     record: AgentIdentityAuthRecord,
```
celia-oai: should we just wrap the struct with Arc to simplify the clone workflow ([example](https://github.com/openai/codex/blob/dev/adrian/codex/hai-task-primitives/codex-rs/login/src/auth/external_bearer.rs#L18-L20))? presumably if we do this it also simplifies unit tests around clone behavior

codex-rs/login/src/auth/agent_identity.rs
```diff
@@ -59,82 +77,148 @@ impl AgentIdentityAuth {
...
     }
 
-    impl Drop for EnvVarGuard {
-        fn drop(&mut self) {
-            unsafe {
-                match &self.original {
-                    Some(value) => env::set_var(self.key, value),
-                    None => env::remove_var(self.key),
+    #[tokio::test]
+    async fn failed_run_task_registration_can_retry() -> anyhow::Result<()> {
```
celia-oai: super nit: this test name make it sound like `ensure_run_task` auto-retries, but i think it is closer to something like `failed_run_task_registration can be retried on the next call"

codex-rs/login/src/auth/agent_identity.rs
```diff
@@ -1,43 +1,61 @@
...
-        let agent_identity_authapi_base_url = agent_identity_authapi_base_url();
-        let process_task_id = register_agent_task(
-            &build_reqwest_client(),
-            &agent_identity_authapi_base_url,
-            key(&record),
-        )
-        .await
-        .map_err(std::io::Error::other)?;
-        Ok(Self {
+    pub fn new(record: AgentIdentityAuthRecord) -> Self {
```
celia-oai: why is ensure_run_task not part of the initialization? feels a little awkward that `AgentIdentityAuth::new(record)` is invalid until we run `ensure_run_task`, like this:
```
        let auth = AgentIdentityAuth::new(record);
        auth.ensure_run_task(Some(base_url)).await?;
        Ok(Self::AgentIdentity(auth))
```

codex-rs/agent-identity/src/lib.rs
adrian-openai: Yeah, agreed. I think this ended up too clever. The AuthAPI URL is a separate endpoint from the ChatGPT backend URL, and we should not hide meaningful environment mapping inside a `normalize...` helper.

I am going to change this so the Agent Identity registration/task-registration code takes an explicit `authapi_base_url` instead of deriving it from `chatgpt_base_url`. Then this helper can either go away or be reduced to very boring trailing-slash normalization on an already-AuthAPI URL.

adrian-openai: Agreed. The intent is not to support arbitrary ChatGPT base URLs for this path. For this POC we only need prod and staging.

I am going to add an explicit environment shape for this, probably `ChatGptEnvironment::{Production, Staging}`, with the known ChatGPT and AuthAPI URLs defined together. Then the callsites can choose prod/staging explicitly and pass the AuthAPI base URL through, rather than trying to infer it from an arbitrary string.

I am intentionally not going to support local/custom endpoints in this environment enum for now; tests that need a mock AuthAPI server can exercise the lower-level registration functions by passing the mock AuthAPI base URL directly.

codex-rs/agent-identity/src/lib.rs
adrian-openai: Makes sense. This is doing a slightly non-obvious thing now: we sample 64 bytes of seed material from `OsRng`, domain-separate/hash it with SHA-512, and then use the first 32 bytes as the Ed25519 seed because `ed25519_dalek` / PKCS#8 stores Ed25519 private keys as a 32-byte seed.

I will expand the inline comment so that is explicit and so it is clear we are not accidentally discarding half the entropy without using it.

codex-rs/login/src/auth/agent_identity.rs
adrian-openai: Context is that both new operations here, agent registration and task registration, go to AuthAPI, while much of Codex currently has a `chatgpt_base_url` threaded through config/session code. I tried to reuse that existing plumbing, but I agree that deriving AuthAPI from ChatGPT URL makes the code harder to reason about.

I am going to change the shape so AuthAPI URL is explicit. The environment/config edge can decide prod vs staging, and the registration code will just receive `authapi_base_url`. That should make this less surprising and remove the need for most of the URL rewrite logic.

codex-rs/login/src/auth/agent_identity.rs
adrian-openai: Yeah, I like that. Right now `AgentIdentityAuth` clones the record but shares the `run_task_id` cell, which is technically fine because the record is immutable, but it is a weird split.

I will switch it to an `Arc`-backed inner struct, similar to the external bearer auth pattern, so clones share the whole auth state. That should make the clone behavior clearer and remove the manual clone implementation.

codex-rs/login/src/auth/agent_identity.rs
adrian-openai: Good nit. The behavior being tested is that a failed `OnceCell` initialization leaves the cell unset, so a later call can try again. It is not testing that `ensure_run_task` itself retries internally.

I will rename this to something like `failed_run_task_registration_can_be_retried_on_next_call`.

codex-rs/login/src/auth/agent_identity.rs
adrian-openai: I agree with the invariant you are pointing at. Historically `AgentIdentityAuth` already existed for the JWT Agent Identity path, and the old `AgentIdentityAuth::load(record)` registered a task before returning an auth object. So the auth object was only usable after task registration had succeeded.

With this stack, both paths should keep that same invariant:

- JWT path: JWT gives us the runtime identity/private key/account metadata, then we register the run task, then construct `AgentIdentityAuth`.
- ChatGPT path: ChatGPT token registers the runtime identity, then registers the run task, then persists/constructs `AgentIdentityAuth`.

I am going to make that clearer in the type/API. Instead of `AgentIdentityAuth::new(record)` accepting a maybe-incomplete record, I will make construction require a complete record with `task_id` present, e.g. `from_complete_record(record)` or equivalent. Then the request/auth-provider path does not need to repair or lazily initialize task state; it can assume Agent Identity auth is ready to sign AgentAssertions.

codex-rs/login/src/auth/manager.rs
```diff
@@ -275,37 +261,40 @@ impl CodexAuth {
...
         jwt: &str,
         chatgpt_base_url: Option<&str>,
+        agent_identity_authapi_base_url: Option<&str>,
     ) -> std::io::Result<Self> {
         let base_url = chatgpt_base_url
-            .unwrap_or(DEFAULT_CHATGPT_BACKEND_BASE_URL)
+            .unwrap_or(ChatGptEnvironment::Production.chatgpt_base_url())
             .trim_end_matches('/')
             .to_string();
+        let agent_identity_authapi_base_url = agent_identity_authapi_base_url_for_auth(
```
celia-oai: this new code is incredibly confusing- it takes in a agent_identity_authapi_base_url but also return a agent_identity_authapi_base_url?

codex-rs/login/src/auth/manager.rs
```diff
@@ -580,53 +564,51 @@ async fn verified_agent_identity_record(
...
+    chatgpt_base_url: Option<&str>,
+    agent_identity_authapi_base_url: Option<&str>,
+) -> std::io::Result<String> {
+    if let Some(base_url) = agent_identity_authapi_base_url {
+        return Ok(base_url.trim_end_matches('/').to_string());
+    }
+
+    let chatgpt_base_url =
+        chatgpt_base_url.unwrap_or(ChatGptEnvironment::Production.chatgpt_base_url());
+    let environment = ChatGptEnvironment::from_chatgpt_base_url(chatgpt_base_url)
```
celia-oai: if we are gonna make Production the default anyways we can do something like this:

pub enum ChatGptEnvironment {
    #[default]
    Production,
    Staging,
}

in `from_chatgpt_base_url`, return
```
pub fn from_chatgpt_base_url(chatgpt_base_url: &str) -> Self {
    let chatgpt_base_url = chatgpt_base_url.trim_end_matches('/');

    match chatgpt_base_url {
        "https://chatgpt.com"
        | "https://chatgpt.com/backend-api"
        | "https://chatgpt.com/codex"
        | "https://chatgpt.com/backend-api/codex"
        | "https://chat.openai.com"
        | "https://chat.openai.com/backend-api"
        | "https://chat.openai.com/codex"
        | "https://chat.openai.com/backend-api/codex" => Self::Production,

        "https://chatgpt-staging.com"
        | "https://chatgpt-staging.com/backend-api"
        | "https://chatgpt-staging.com/codex"
        | "https://chatgpt-staging.com/backend-api/codex" => Self::Staging,

        _ => {
            tracing::warn!(
                %chatgpt_base_url,
                "unrecognized ChatGPT base URL; defaulting to production"
            );
             Self::default()
        }
    }
}
```
then you can get the url like:
```
chatgpt_base_url
    .map(ChatGptEnvironment::from_chatgpt_base_url)
    .unwrap_or_default()
    .agent_identity_authapi_base_url()
    .to_string()
```

codex-rs/login/src/auth/manager.rs
adrian-openai: Agreed, that was confusing. The input here is an optional AuthAPI base URL override, while the value we pass into `AgentIdentityAuth::load` is the resolved AuthAPI base URL after applying the override-or-environment logic. I am renaming the parameter to `agent_identity_authapi_base_url_override`, the local to `resolved_authapi_base_url`, and the helper to `resolve_agent_identity_authapi_base_url` so the direction of data is reviewable at the call site.

codex-rs/login/src/auth/manager.rs
adrian-openai: I agree with adding a real default on the environment enum, and I am taking that part: `ChatGptEnvironment` will derive `Default` with `Production` as the default, and missing `chatgpt_base_url` will use that default. I am intentionally not defaulting an explicit unknown/custom ChatGPT URL to prod, though. For this POC we only support prod and staging, and silently registering an unknown ChatGPT base URL against prod AuthAPI seems riskier than a clear unsupported-url error.

celia-oai (approved):

PR #19048, opened 2026-04-22
feat: expose AWS account state from account/read

celia-oai:
## Why

AWS/Bedrock mode currently reports `account: null` with `requiresOpenaiAuth: false` from `account/read`. That suppresses the OpenAI-auth requirement, but it does not let app clients distinguish AWS auth from any other non-OpenAI custom provider. For the prototype AWS provider UX, clients need a simple provider-derived signal so they can suppress ChatGPT/API-key login and token-refresh paths without hardcoding Bedrock checks.

## What changed

- Adds an `aws` variant to the v2 `Account` protocol union.
- Adds `ProviderAccountKind` to `codex-model-provider` so the runtime provider owns the app-visible account classification.
- Makes Amazon Bedrock return `ProviderAccountKind::Aws` from the model-provider layer.
- Updates app-server `account/read` to map `ProviderAccountKind` to the existing `GetAccountResponse` wire shape.
- Preserves the existing `account: null, requiresOpenaiAuth: false` behavior for other non-OpenAI providers.
- Regenerates the app-server protocol schema fixtures.
- Adds coverage for provider account classification and for the Amazon Bedrock `account/read` response.

## Testing

- `cargo test -p codex-model-provider`
- `cargo test -p codex-app-server get_account_with_aws_provider`

## Notes

I attempted `just bazel-lock-update` and `just bazel-lock-check`, but both are blocked in my local environment because `bazel` is not installed.


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1929,6 +1889,10 @@ pub enum Account {
     #[serde(rename = "chatgpt", rename_all = "camelCase")]
     #[ts(rename = "chatgpt", rename_all = "camelCase")]
     Chatgpt { email: String, plan_type: PlanType },
+
+    #[serde(rename = "aws", rename_all = "camelCase")]
+    #[ts(rename = "aws", rename_all = "camelCase")]
+    Aws {},
```
pakrym-oai: align with provider name?

codex-rs/model-provider/src/provider.rs
```diff
@@ -99,14 +115,24 @@ impl ModelProvider for ConfiguredModelProvider {
             None => None,
         }
     }
+
+    fn account_kind(&self) -> ProviderAccountKind {
```
pakrym-oai: shouuld this be returning an account instance?

pakrym-oai (commented):
Provider should probably be returning an account instance and encapsulate the logic that is used to populate it.

pakrym-oai (approved):

PR #19049, opened 2026-04-22
feat: opt ChatGPT auth into agent identity

adrian-openai:
## Stack

This is PR 2 of the simplified HAI single-run-task stack:

## Summary

This PR adds the disabled-by-default path for normal ChatGPT-login Codex sessions to obtain Agent Identity runtime auth through the Codex backend. Existing Agent Identity JWT startup mode remains a separate path and does not require the feature flag.

What changed:

- adds the experimental `use_agent_identity` feature flag and config schema entry
- adds an explicit `AgentIdentityAuthPolicy` so call sites choose `JwtOnly` or `ChatGptAuth` instead of passing a bare boolean
- stores standalone Agent Identity JWT credentials separately from backend-registered Agent Identity records
- persists the registered Agent Identity record, private key, and single run task id in `auth.json` so process restarts reuse the same identity
- derives the agent/task registration base URL from ChatGPT/Codex auth config while keeping JWT JWKS lookup separate
- provisions and caches ChatGPT-derived Agent Identity runtime auth when `use_agent_identity` is enabled
- reuses the shared run-task registration helper from PR1 rather than adding a second task-registration path

This PR intentionally does not switch model inference over to `AgentAssertion` auth. The provider-auth integration lands in the next PR.

## Testing

- `just test -p codex-login`

### DIFF
--- a/codex-rs/login/src/auth/agent_identity.rs
+++ b/codex-rs/login/src/auth/agent_identity.rs
@@ -1,43 +1,62 @@
+use std::sync::Arc;
+
 use codex_agent_identity::AgentIdentityKey;
+use codex_agent_identity::normalize_chatgpt_base_url;
 use codex_agent_identity::register_agent_task;
 use codex_protocol::account::PlanType as AccountPlanType;
-use std::env;
+use tokio::sync::OnceCell;
 
 use crate::default_client::build_reqwest_client;
 
 use super::storage::AgentIdentityAuthRecord;
 
-const PROD_AGENT_IDENTITY_AUTHAPI_BASE_URL: &str = "https://auth.openai.com/api/accounts";
-const CODEX_AGENT_IDENTITY_AUTHAPI_BASE_URL_ENV_VAR: &str = "CODEX_AGENT_IDENTITY_AUTHAPI_BASE_URL";
+const DEFAULT_CHATGPT_BACKEND_BASE_URL: &str = "https://chatgpt.com/backend-api";
 
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct AgentIdentityAuth {
     record: AgentIdentityAuthRecord,
-    process_task_id: String,
+    process_task_id: Arc<OnceCell<String>>,
+}
+
+impl Clone for AgentIdentityAuth {
+    fn clone(&self) -> Self {
+        Self {
+            record: self.record.clone(),
+            process_task_id: Arc::clone(&self.process_task_id),
+        }
+    }
 }
 
 impl AgentIdentityAuth {
-    pub async fn load(record: AgentIdentityAuthRecord) -> std::io::Result<Self> {
-        let agent_identity_authapi_base_url = agent_identity_authapi_base_url();
-        let process_task_id = register_agent_task(
-            &build_reqwest_client(),
-            &agent_identity_authapi_base_url,
-            key(&record),
-        )
-        .await
-        .map_err(std::io::Error::other)?;
-        Ok(Self {
+    pub fn new(record: AgentIdentityAuthRecord) -> Self {
+        Self {
             record,
-            process_task_id,
-        })
+            process_task_id: Arc::new(OnceCell::new()),
+        }
     }
 
     pub fn record(&self) -> &AgentIdentityAuthRecord {
         &self.record
     }
 
-    pub fn process_task_id(&self) -> &str {
-        &self.process_task_id
+    pub fn process_task_id(&self) -> Option<&str> {
+        self.process_task_id.get().map(String::as_str)
+    }
+
+    pub async fn ensure_runtime(&self, chatgpt_base_url: Option<String>) -> std::io::Result<()> {
+        self.process_task_id
+            .get_or_try_init(|| async {
+                let base_url = normalize_chatgpt_base_url(
+                    chatgpt_base_url
+                        .as_deref()
+                        .unwrap_or(DEFAULT_CHATGPT_BACKEND_BASE_URL),
+                );
+                register_agent_task(&build_reqwest_client(), &base_url, self.key())
+                    .await
+                    .map_err(std::io::Error::other)
+            })
+            .await
+            .map(|_| ())
     }
 
     pub fn account_id(&self) -> &str {
@@ -59,82 +78,10 @@ impl AgentIdentityAuth {
     pub fn is_fedramp_account(&self) -> bool {
         self.record.chatgpt_account_is_fedramp
     }
-}
-
-fn agent_identity_authapi_base_url() -> String {
-    env::var(CODEX_AGENT_IDENTITY_AUTHAPI_BASE_URL_ENV_VAR)
-        .ok()
-        .map(|base_url| base_url.trim().trim_end_matches('/').to_string())
-        .filter(|base_url| !base_url.is_empty())
-        .unwrap_or_else(|| PROD_AGENT_IDENTITY_AUTHAPI_BASE_URL.to_string())
-}
-
-fn key(record: &AgentIdentityAuthRecord) -> AgentIdentityKey<'_> {
-    AgentIdentityKey {
-        agent_runtime_id: &record.agent_runtime_id,
-        private_key_pkcs8_base64: &record.agent_private_key,
-    }
-}
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-    use serial_test::serial;
-
-    #[test]
-    #[serial(codex_auth_env)]
-    fn agent_identity_authapi_base_url_prefers_env_value() {
-        let _guard = EnvVarGuard::set(
-            CODEX_AGENT_IDENTITY_AUTHAPI_BASE_URL_ENV_VAR,
-            "https://authapi.example.test/api/accounts/",
-        );
-        assert_eq!(
-            agent_identity_authapi_base_url(),
-            "https://authapi.example.test/api/accounts"
-        );
-    }
-
-    #[test]
-    #[serial(codex_auth_env)]
-    fn agent_identity_authapi_base_url_uses_prod_authapi_by_default() {
-        let _guard = EnvVarGuard::remove(CODEX_AGENT_IDENTITY_AUTHAPI_BASE_URL_ENV_VAR);
-        assert_eq!(
-            agent_identity_authapi_base_url(),
-            PROD_AGENT_IDENTITY_AUTHAPI_BASE_URL
-        );
-    }
-
-    struct EnvVarGuard {
-        key: &'static str,
-        original: Option<std::ffi::OsString>,
-    }
-
-    impl EnvVarGuard {
-        fn set(key: &'static str, value: &str) -> Self {
-            let original = env::var_os(key);
-            unsafe {
-                env::set_var(key, value);
-            }
-            Self { key, original }
-        }
-
-        fn remove(key: &'static str) -> Self {
-            let original = env::var_os(key);
-            unsafe {
-                env::remove_var(key);
-            }
-            Self { key, original }
-        }
-    }
-
-    impl Drop for EnvVarGuard {
-        fn drop(&mut self) {
-            unsafe {
-                match &self.original {
-                    Some(value) => env::set_var(self.key, value),
-                    None => env::remove_var(self.key),
-                }
-            }
+    fn key(&self) -> AgentIdentityKey<'_> {
+        AgentIdentityKey {
+            agent_runtime_id: &self.record.agent_runtime_id,
+            private_key_pkcs8_base64: &self.record.agent_private_key,
         }
     }
 }
--- a/codex-rs/login/src/auth/agent_identity.rs
+++ b/codex-rs/login/src/auth/agent_identity.rs
@@ -1,81 +1,316 @@
+use std::future::Future;
 use std::sync::Arc;
 
 use codex_agent_identity::AgentIdentityKey;
+use codex_agent_identity::ChatGptEnvironment;
+use codex_agent_identity::build_abom;
+use codex_agent_identity::decode_agent_identity_jwt;
+use codex_agent_identity::fetch_agent_identity_jwks;
+use codex_agent_identity::generate_agent_key_material;
+use codex_agent_identity::is_retryable_registration_error;
+use codex_agent_identity::public_key_ssh_from_private_key_pkcs8_base64;
+use codex_agent_identity::register_agent_identity;
 use codex_agent_identity::register_agent_task;
 use codex_protocol::account::PlanType as AccountPlanType;
+use codex_protocol::protocol::SessionSource;
+use thiserror::Error;
 
 use crate::default_client::build_reqwest_client;
 
 use super::storage::AgentIdentityAuthRecord;
 
-#[derive(Clone, Debug)]
-pub struct AgentIdentityAuth {
-    inner: Arc<AgentIdentityAuthInner>,
+pub(super) const MAX_AGENT_IDENTITY_BOOTSTRAP_ATTEMPTS: usize = 3;
+
+pub(super) fn agent_identity_authapi_base_url(
+    chatgpt_base_url: Option<&str>,
+) -> std::io::Result<String> {
+    let environment = match chatgpt_base_url {
+        Some(chatgpt_base_url) => ChatGptEnvironment::from_chatgpt_base_url(chatgpt_base_url)
+            .map_err(std::io::Error::other)?,
+        None => ChatGptEnvironment::default(),
+    };
+    Ok(environment.agent_identity_authapi_base_url().to_string())
 }
 
-#[derive(Debug)]
-struct AgentIdentityAuthInner {
-    record: AgentIdentityAuthRecord,
-    run_task_id: String,
+pub(super) fn require_agent_identity_authapi_base_url(
+    agent_identity_authapi_base_url: Option<&str>,
+) -> std::io::Result<&str> {
+    agent_identity_authapi_base_url.ok_or_else(|| {
+        std::io::Error::other(
+            "Agent Identity only supports production and staging ChatGPT environments",
+        )
+    })
+}
+
+#[derive(Debug, Error)]
+pub enum AgentIdentityAuthError {
+    #[error(
+        "agent identity bootstrap unavailable after {attempts} attempts during {operation}: {message}"
+    )]
+    BootstrapUnavailable {
+        operation: &'static str,
+        attempts: usize,
+        message: String,
+    },
+}
+
+impl AgentIdentityAuthError {
+    pub fn is_bootstrap_unavailable(error: &std::io::Error) -> bool {
+        matches!(
+            error
+                .get_ref()
+                .and_then(|source| source.downcast_ref::<Self>()),
+            Some(Self::BootstrapUnavailable { .. })
+        )
+    }
+}
+
+#[derive(Debug, Error)]
+#[error("retryable agent identity registration failure: {message}")]
+pub(super) struct RetryableAgentIdentityRegistrationError {
+    message: String,
+}
+
+impl RetryableAgentIdentityRegistrationError {
+    pub(super) fn new(message: String) -> Self {
+        Self { message }
+    }
+}
+
+#[derive(Clone, Debug)]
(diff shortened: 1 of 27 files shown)


efrazer-oai:
High level not sure we want to persist the task id in rollout state. Ik in initial reviews of the old stack i flagged this as problematic due to the deeper problem (having different tasks for same session id), but idt this is the solution we'd want long term. 

### Reasoning
It seems orthogonal/like a (potentially unstable) implementation detail that our deeper auth logic shouldn't care about, and we can't gate rollout item changes on feature flag. Rollout data really matters & we'd want to keep very good hygiene.

If we can, let's default to making task id ephemeral and shield it as *much as possible* from core session code.


### Solution Musing

As a related note, it'd be *so* awesome if we could actually choose the task id name or establish some link between a session id and a task id. That way we get consistency backend-side rather than having to concern a lot of our app-server code with this stuff.


### Suggestion

So in the AuthEnum stack, we have pretty much everything related to auth abstracted into the 'Auth' provider. Almost every call basically only has to say 'hey give me an auth provider and have that inject headers'. 

So I think what we can do when enabling agent-identity feature is adding a *different* auth provider. What'd be amazing as an abstraction is if we could store some of this ephemeral stuff in-memory on the auth object (e.g. hashmap of session ids to task ids) and basically have the provider take our 'auth' and potentially a session id hint.

nicksteele-oai (approved):
Looks good on my end, might double check with codex

codex-rs/login/src/auth/manager.rs
```diff
@@ -59,6 +67,15 @@ pub enum CodexAuth {
     PersonalAccessToken(PersonalAccessTokenAuth),
 }
 
+/// Policy for resolving Agent Identity auth from a broader Codex auth snapshot.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AgentIdentityAuthPolicy {
+    /// Only use an existing Agent Identity JWT/runtime auth.
+    JwtOnly,
+    /// Use an Agent Identity JWT/runtime auth, or register one from managed ChatGPT auth.
```
adrian-openai: I think this has not been updated? Why would we accept both?

codex-rs/login/src/auth/manager.rs
```diff
@@ -87,19 +104,40 @@ pub struct ChatgptAuthTokens {
...
+    chatgpt_user_id: String,
+    email: String,
+    plan_type: AccountPlanType,
+    chatgpt_account_is_fedramp: bool,
+    access_token: String,
+}
+
 const TOKEN_REFRESH_INTERVAL: i64 = 8;
 const CHATGPT_ACCESS_TOKEN_REFRESH_WINDOW_MINUTES: i64 = 5;
+const DEFAULT_CHATGPT_BACKEND_BASE_URL: &str = "https://chatgpt.com/backend-api";
```
adrian-openai: I don't think that we need to move this from where it was before. Just leave it where it was on line 141.

Zemnmez (approved):

codex-rs/login/src/auth/agent_identity.rs
```diff
@@ -1,83 +1,170 @@
...
+    record: &AgentIdentityAuthRecord,
+    agent_identity_authapi_base_url: &str,
+) -> std::io::Result<String> {
+    retry_registration(|| async {
+        register_task_for_record(record, agent_identity_authapi_base_url).await
+    })
+    .await
+}
+
+async fn register_task_for_record(
```
celia-oai, line 295: this should be part of the initialization of AgentIdentityAuth? again, like the previous PR, we run into issue of invalid AgentIdentityAuth when initialized.

codex-rs/login/src/auth/manager.rs
```diff
@@ -93,9 +137,30 @@ pub struct ChatgptAuthTokens {
 #[derive(Debug, Clone)]
 struct ChatgptAuthState {
     auth_dot_json: Arc<Mutex<Option<AuthDotJson>>>,
+    agent_identity_auth: Arc<Mutex<Option<AgentIdentityAuth>>>,
```
celia-oai: we already have `AgentIdentityStorage` in auth_dot_json. doesn't this create two source of truth for identity auth?

codex-rs/login/src/auth/agent_identity.rs
adrian-openai: Good point. The current constructor validates that `task_id` exists, so we are not silently constructing an invalid `AgentIdentityAuth`, but the task-registration step is still split across a helper and the constructor. I agree that is too easy to misuse. I’m going to move the “ensure/register run task” step into an `AgentIdentityAuth` initialization path so callers receive only a complete usable auth object.

codex-rs/login/src/auth/manager.rs
adrian-openai: Yes, I think this is a fair point. This field was intended as an in-memory cache derived from `auth_dot_json.agent_identity`, not a second durable source of truth, but after the recent simplification it is no longer worth the extra state. I’m going to remove the cache and make `auth_dot_json.agent_identity` the single ChatGPT-managed source of truth. JWT storage will continue to stay as `AgentIdentityStorage::Jwt`, with runtime `AgentIdentityAuth` derived from it when loaded.

codex-rs/login/src/auth/manager.rs
```diff
@@ -96,6 +139,25 @@ struct ChatgptAuthState {
     client: CodexHttpClient,
 }
 
+impl ChatgptAuthState {
+    fn new(auth_dot_json: AuthDotJson) -> Self {
```
celia-oai: don't think this helper is super helpful. let's just change back to the old code

codex-rs/login/src/auth/manager.rs
```diff
@@ -221,25 +283,58 @@ impl CodexAuth {
...
                 return Err(std::io::Error::other(
-                    "agent identity auth is missing an agent identity token.",
+                    "agent identity auth is missing agent identity auth material.",
                 ));
             };
-            return Self::from_agent_identity_jwt(
-                &agent_identity,
-                chatgpt_base_url,
+            let base_url = chatgpt_base_url
+                .unwrap_or(ChatGptEnvironment::Production.chatgpt_base_url())
```
celia-oai: nit: let's do `ChatGptEnvironment::default`

codex-rs/login/src/auth/agent_identity.rs
```diff
@@ -1,83 +1,171 @@
...
+            .task_id
+            .as_deref()
+            .is_none_or(|task_id| task_id.trim().is_empty())
+        {
+            record.task_id = Some(
+                register_task_for_record_with_retries(&record, agent_identity_authapi_base_url)
+                    .await?,
+            );
+        }
+        Self::from_record(record)
```
celia-oai: don't think this is needed, we can return this directly:
```
Ok(Self {
            record: Arc::new(record),
        })
```

codex-rs/login/src/auth/agent_identity.rs
```diff
@@ -1,83 +1,171 @@
...
+    }
+
+    pub fn from_record(record: AgentIdentityAuthRecord) -> std::io::Result<Self> {
```
celia-oai: thoughts on just removing this and renaming the above one to `from_record`? don't think this one is needed. we shouldn't need to hit runtime error and we should encourage people to use the function above instead of giving them choices.

codex-rs/login/src/auth/manager.rs
```diff
@@ -221,25 +283,58 @@ impl CodexAuth {
...
+                .to_string();
+            let agent_identity_authapi_base_url = resolve_agent_identity_authapi_base_url(
+                Some(&base_url),
                 agent_identity_authapi_base_url,
-            )
-            .await;
+            )?;
+            match agent_identity {
+                AgentIdentityStorage::Jwt(jwt) => {
+                    let record = verified_agent_identity_record(&jwt, &base_url).await?;
```
celia-oai: this workflow can be part of AgentIdentityAuth? like
`AgentIdentityAuth::from_jwt(jwt, chatgpt_jwt_base_url, agent_identity_authapi_base_url)`

codex-rs/login/src/auth/manager.rs
```diff
@@ -221,25 +283,58 @@ impl CodexAuth {
...
+                    let auth = AgentIdentityAuth::from_record_with_registered_task(
+                        record,
+                        &agent_identity_authapi_base_url,
+                    )
+                    .await?;
+                    let storage = AgentIdentityStorage::Record(auth.record().clone());
+                    if auth_dot_json.agent_identity.as_ref() != Some(&storage) {
+                        let mut auth_dot_json = auth_dot_json;
+                        auth_dot_json.agent_identity = Some(storage);
+                        save_auth(
```
celia-oai: why are we saving auth in this code path? this workflow shouldn't be persisting auth.

codex-rs/login/src/auth/manager.rs
adrian-openai: Agreed. This helper was carrying more weight when `ChatgptAuthState` also owned the agent identity cache.

codex-rs/login/src/auth/manager.rs
adrian-openai: Nice catch.

codex-rs/login/src/auth/agent_identity.rs
adrian-openai: Agreed, and I think this folds into the next comment too. Once the production constructor owns task registration, the extra `Self::from_record(record)` validation hop is unnecessary. I’ll collapse this so the constructor registers/ensures the task and then directly returns `Ok(Self { record: Arc::new(record) })`.

codex-rs/login/src/auth/agent_identity.rs
adrian-openai: I agree. Having both constructors gives callers a choice we do not really want them to have. I’ll make the task-registering constructor the production `from_record(...)` path and remove the public sync constructor. Tests can keep a test-only initialized-record helper where they need to avoid network setup.

codex-rs/login/src/auth/manager.rs
adrian-openai: Yes, agreed. The manager should not need to know the details of “verify JWT, turn claims into a record, then ensure task registration.” I’ll move that workflow behind an `AgentIdentityAuth::from_jwt(...)` constructor so JWT and record initialization both live next to the invariant that `AgentIdentityAuth` is always task-ready.

codex-rs/login/src/auth/manager.rs
adrian-openai: Good catch. This save is leftover from trying to repair incomplete records while loading auth, but from_auth_dot_json should not mutate persistence. I’ll remove the save from this load path.

codex-rs/login/src/auth/agent_identity.rs
```diff
@@ -1,83 +1,186 @@
...
         Ok(Self {
-            inner: Arc::new(AgentIdentityAuthInner {
-                record,
-                run_task_id,
-            }),
+            record: Arc::new(record),
         })
     }
 
+    pub async fn from_jwt(
```
celia-oai, line 113: let's add unit test for this workflow too if it's easy?

codex-rs/login/src/auth/manager.rs
```diff
@@ -1548,6 +1822,11 @@ pub trait AuthManagerConfig {
 
     /// Returns the ChatGPT backend base URL used for first-party backend authorization.
     fn chatgpt_base_url(&self) -> String;
+
+    /// Returns an explicit AuthAPI base URL for Agent Identity registration, when configured.
+    fn agent_identity_authapi_base_url(&self) -> Option<String> {
+        None
```
celia-oai: this shouldn't return a default here. it should be managed by `codex_core::config::Config`. see docstring above

codex-rs/login/src/auth/manager.rs
```diff
@@ -1876,10 +2183,44 @@ impl AuthManager {
     }
 
     pub fn set_forced_chatgpt_workspace_id(&self, workspace_id: Option<Vec<String>>) {
```
celia-oai, line 2115: are changes in this function from unclean merge? doesn't seem to have to do with this PR?

codex-rs/login/src/auth/manager.rs
```diff
@@ -1734,6 +2007,40 @@ impl AuthManager {
...
+    pub async fn agent_identity_auth(
+        &self,
+        policy: AgentIdentityAuthPolicy,
+        session_source: SessionSource,
+    ) -> std::io::Result<Option<AgentIdentityAuth>> {
+        let Some(auth) = self.auth().await else {
+            return Ok(None);
+        };
+        if policy == AgentIdentityAuthPolicy::ChatGptAuth && matches!(auth, CodexAuth::Chatgpt(_)) {
+            let _permit = self
```
celia-oai: nit: no action item, but just in future would prefer we don't introduce dead variables like this that will be used by following PRs, and just add them in the PR that this gets used. this makes it a bit easier to review

codex-rs/login/src/auth/manager.rs
```diff
@@ -498,6 +568,126 @@ impl CodexAuth {
...
+            Self::AgentIdentity(auth) => Ok(Some(auth.clone())),
+            Self::ApiKey(_)
+            | Self::ChatgptAuthTokens(_)
+            | Self::PersonalAccessToken(_)
+            | Self::BedrockApiKey(_) => Ok(None),
+            Self::Chatgpt(_) => {
+                if policy == AgentIdentityAuthPolicy::JwtOnly {
+                    return Ok(None);
+                }
+                self.ensure_chatgpt_agent_identity(
```
celia-oai: could we move most of the agent identity specific logic to `agent_identity.rs` so we don't blow up the size of `manager.rs`? imo only the logic that has manager.rs dependencies (e.g. CodexAuth) should live here.

codex-rs/login/src/auth/manager.rs
```diff
@@ -498,6 +568,126 @@ impl CodexAuth {
...
+    ) -> std::io::Result<AgentIdentityAuth> {
+        let binding = ChatgptAgentIdentityBinding::from_auth(self, forced_chatgpt_workspace_id)
+            .ok_or_else(|| std::io::Error::other("ChatGPT auth is unavailable"))?;
+
+        let agent_identity_authapi_base_url = resolve_agent_identity_authapi_base_url(
+            chatgpt_base_url.as_deref(),
+            agent_identity_authapi_base_url.as_deref(),
+        )?;
+
+        if let Some(record) = self.stored_chatgpt_agent_identity_record(&binding.account_id)
```
celia-oai: this part seems to only work for record? should it also work for jwt?

codex-rs/login/src/auth/manager.rs
```diff
@@ -498,6 +568,126 @@ impl CodexAuth {
...
+            && agent_identity_record_matches_binding(&record, &binding)
+            && public_key_ssh_from_private_key_pkcs8_base64(&record.agent_private_key).is_ok()
+        {
+            let auth = map_retry_exhaustion_to_bootstrap_unavailable(
```
celia-oai: this wrapper function makes the code hard to read. could we do it more like this?
```
let auth = AgentIdentityAuth::from_record(record, &agent_identity_authapi_base_url)
    .await
    .map_err(|err| classify_bootstrap_error("agent task registration", err))?;
```

celia-oai: same with all other callsites of `map_retry_exhaustion_to_bootstrap_unavailable`. we should get rid of this

codex-rs/login/src/auth/manager.rs
```diff
@@ -498,6 +568,126 @@ impl CodexAuth {
...
+                "agent task registration",
+                AgentIdentityAuth::from_record(record, &agent_identity_authapi_base_url).await,
+            )?;
+            return self.persist_chatgpt_agent_identity_auth(auth);
```
celia-oai: why do we have to persist again if it already exist in storage?

codex-rs/login/src/auth/manager.rs
adrian-openai: nice catch

codex-rs/login/src/auth/manager.rs
adrian-openai: I agree in principle for sure! In this case the variable is intentionally holding the semaphore permit until bootstrap finishes, so it is doing work even though it is not referenced again. I'll rename it to `_bootstrap_permit` to make that clearer.

codex-rs/login/src/auth/manager.rs
adrian-openai: Agreed. I'll move the agent-identity-specific bootstrap helpers into `agent_identity.rs` and keep `manager.rs` focused on the pieces that need `CodexAuth`/storage access.

codex-rs/login/src/auth/manager.rs
adrian-openai: Agreed. I will remove `map_retry_exhaustion_to_bootstrap_unavailable` entirely and switch the callsites to direct `map_err(|err| classify_bootstrap_error("...", err))?` mapping.

codex-rs/login/src/auth/manager.rs
adrian-openai: Good catch. If the stored record is already complete, we should just return the auth. We only need to persist when `from_record(...)` fills in a missing `task_id`, so I will make that explicit.

codex-rs/login/src/auth/manager.rs
adrian-openai: Good catch. This is unrelated workspace-auth cache behavior and does not belong in this PR. I’ll revert this function back to the existing shape.

codex-rs/login/src/auth/manager.rs
adrian-openai: This branch is intentionally only for managed ChatGPT records. JWT auth is loaded earlier as `CodexAuth::AgentIdentity` and returns through the `Self::AgentIdentity(auth)` arm, so it does not reach this path. I’ll rename these helpers around “managed ChatGPT” and add a short comment so that boundary is explicit.

codex-rs/core/src/config/mod.rs
```diff
@@ -1146,6 +1146,29 @@ impl AuthManagerConfig for Config {
     fn chatgpt_base_url(&self) -> String {
         self.chatgpt_base_url.clone()
     }
+
+    fn agent_identity_authapi_base_url(&self) -> Option<String> {
```
celia-oai: previously i thought this is a managed config, so i suggested put it here, but since it's derived entirely from chatgpt_base_url, the better place for this logic to live in is:
1. `agent_identity_authapi_base_url` resolution will be a helper function that lives in `agent_identity.rs`;
2. `AuthManager::new` only needs to take in `chatgpt_base_url` and resolves `agent_identity_authapi_base_url` given the `chatgpt_base_url`. 

this way the boundry is clear: config only takes in managed configs, and `agent_identity_authapi_base_url` is derived in runtime based on managed configs.

adrian-openai: Agreed. Since this is derived from `chatgpt_base_url`, it should not live on the config boundary. I’ll remove `agent_identity_authapi_base_url` from `AuthManagerConfig`/constructor plumbing and derive it in the auth/agent-identity path from the ChatGPT base URL.

codex-rs/login/src/auth/manager.rs
```diff
@@ -1628,7 +1793,9 @@ impl AuthManager {
             keyring_backend_kind,
             forced_chatgpt_workspace_id: RwLock::new(forced_chatgpt_workspace_id),
             chatgpt_base_url,
+            agent_identity_authapi_base_url_override: None,
```
celia-oai: instead of passing agent_identity_authapi_base_url_override around, let's resolve agent_identity_authapi_base_url in this initialization and only pass resolved agent_identity_authapi_base_url in auth manager

adrian-openai: Yeah, that makes sense, sorry I misunderstood this from your previous comment

celia-oai (approved):

PR #19051, opened 2026-04-22
feat: use run agent task auth for inference

adrian-openai:
## Stack

This is PR 3 of the simplified HAI single-run-task stack:

## Summary

This PR moves Agent Identity usage into provider auth resolution. That keeps `AgentAssertion` auth tied to first-party OpenAI provider requests instead of applying a late session-wide override that could affect local, custom, Bedrock, API-key, or external-bearer providers.

What changed:

- adds a small `ProviderAuthScope` struct carrying the run auth policy and session source needed by provider-scoped auth resolution
- lets `Session` opt the existing `ModelClient` into `ChatGptAuth` policy when `use_agent_identity` is enabled, without adding a second model-client constructor
- resolves Agent Identity only for first-party OpenAI provider auth paths
- uses the persisted run task id from the `AgentIdentityAuth` record to build `AgentAssertion` auth for Responses requests
- routes shared request setup through scoped provider auth so unary compact requests use the same run-task assertion path as inference turns
- keeps local/custom/Bedrock/env-key/external-bearer provider auth unchanged
- lets missing run-task state surface through the existing model-request error path instead of silently falling back to bearer auth

This PR intentionally does not create thread-scoped, target-scoped, or background-scoped task identities. The run task is the only task Codex registers in this POC shape.

## Testing

- `just test -p codex-model-provider`
- `just test -p codex-core client::tests::provider_auth_scope_uses`
- `just test -p codex-core remote_compact_uses_agent_identity_assertion`


codex-rs/core/src/session/session.rs
```diff
@@ -1028,14 +1028,19 @@ impl Session {
                 live_thread: live_thread_init.as_ref().cloned(),
                 thread_store: Arc::clone(&thread_store),
                 attestation_provider: attestation_provider.clone(),
-                model_client: ModelClient::new(
+                model_client: ModelClient::new_with_agent_identity_policy(
```
adrian-openai: Why did we need to change the name of this?

codex-rs/core/src/session/session.rs
```diff
@@ -1234,6 +1237,8 @@ impl Session {
 
             // record_initial_history can emit events. We record only after the SessionConfiguredEvent is emitted.
             Box::pin(sess.record_initial_history(initial_history)).await;
+            sess.schedule_startup_prewarm(session_configuration.base_instructions.clone())
```
adrian-openai: Why did this get moved down?

codex-rs/model-provider/src/amazon_bedrock/mod.rs
```diff
@@ -105,6 +106,10 @@ impl ModelProvider for AmazonBedrockModelProvider {
         resolve_provider_auth(&self.aws).await
     }
 
+    async fn api_auth_for_scope(&self, _scope: ProviderAuthScope) -> Result<SharedAuthProvider> {
```
adrian-openai: Why was this changed?

codex-rs/model-provider/src/models_endpoint.rs
```diff
@@ -87,7 +88,13 @@ impl ModelsEndpointClient for OpenAiModelsEndpoint {
         let auth = self.auth().await;
         let auth_mode = auth.as_ref().map(CodexAuth::auth_mode);
         let api_provider = self.provider_info.to_api_provider(auth_mode)?;
-        let api_auth = resolve_provider_auth(auth.as_ref(), &self.provider_info)?;
+        let api_auth = resolve_provider_auth(
```
adrian-openai: Why do we need to make this change? JWT didn't need to make this change? Why would we make it??

Zemnmez (approved):

nicksteele-oai (approved):

codex-rs/core/src/client.rs
```diff
@@ -898,14 +927,26 @@ impl ModelClient {
     async fn current_client_setup(&self) -> Result<CurrentClientSetup> {
         let auth = self.state.provider.auth().await;
         let api_provider = self.state.provider.api_provider().await?;
-        let api_auth = self.state.provider.api_auth().await?;
+        let api_auth = self
+            .state
+            .provider
+            .api_auth_for_scope(ProviderAuthScope {
```
celia-oai, line 927: I did a quick search of the codebase and `provider.api_auth()` is still called in several places after `api_auth_for_scope` is introduced. is this split gonna cause issues for agent identity? or does only this call here need agent identity scope?

codex-rs/codex-api/src/auth.rs
```diff
@@ -59,6 +59,12 @@ pub trait AuthProvider: Send + Sync {
             Ok(request)
         })
     }
+
+    /// Returns safe Agent Identity metadata for request telemetry, when this
```
celia-oai: this shouldn't be added to the `pub trait AuthProvider` if it only exists for agent identity.

codex-rs/core/src/session/session.rs
```diff
@@ -1074,6 +1074,11 @@ impl Session {
                     /*item_ids_enabled*/ config.features.enabled(Feature::ItemIds),
                     attestation_provider,
                 )
+                .with_agent_identity_policy(if config.features.enabled(Feature::UseAgentIdentity) {
```
celia-oai: Can we just add agent_identity_policy as a field on SessionServices instead of adding a helper function?

codex-rs/core/src/session_startup_prewarm.rs
```diff
@@ -183,6 +183,12 @@ impl SessionStartupPrewarmHandle {
 impl Session {
     pub(crate) async fn schedule_startup_prewarm(self: &Arc<Self>, base_instructions: String) {
         if !self.services.model_client.responses_websocket_enabled() {
+            let model_client = self.services.model_client.clone();
+            tokio::spawn(async move {
+                if let Err(err) = model_client.prewarm_auth().await {
```
celia-oai, line 190: why are we adding this behavior? doesn't seem related to agent identity auth

codex-rs/core/src/client.rs
```diff
@@ -605,6 +622,7 @@ impl ModelClient {
         session_config: ApiRealtimeSessionConfig,
         mut extra_headers: ApiHeaderMap,
         api_provider_override: Option<ApiProvider>,
+        session_telemetry: &SessionTelemetry,
```
celia-oai: why are we threading the telemetry through realtime call?

codex-rs/model-provider/src/models_endpoint.rs
```diff
@@ -164,6 +168,8 @@ impl RequestTelemetry for ModelsRequestTelemetry {
             auth.error = response_debug.auth_error.as_deref(),
             auth.error_code = response_debug.auth_error_code.as_deref(),
             auth.mode = self.auth_mode.as_deref(),
+            auth.agent_id = self.agent_identity_telemetry.as_ref().map(|metadata| metadata.agent_id.as_str()),
```
celia-oai, line 176: could we add these telemetry through e.g. auth_env like the rest of the telemetry collected?

codex-rs/otel/src/events/session_telemetry.rs
```diff
@@ -578,6 +581,8 @@ impl SessionTelemetry {
                 auth.cf_ray = cf_ray,
                 auth.error = auth_error,
                 auth.error_code = auth_error_code,
+                auth.agent_id = agent_identity_telemetry.map(|metadata| metadata.agent_id.as_str()),
```
celia-oai, line 584: sam here

codex-rs/model-provider/src/auth.rs
```diff
@@ -95,6 +131,70 @@ pub(crate) fn resolve_provider_auth(
...
+    } = scope;
+    if let Some(CodexAuth::AgentIdentity(agent_identity_auth)) = auth {
+        return Ok(Arc::new(AgentIdentityAuthProvider {
+            auth: agent_identity_auth.clone(),
+        }));
+    }
+
+    let managed_chatgpt_auth = agent_identity_policy == AgentIdentityAuthPolicy::ChatGptAuth
+        && matches!(auth, Some(CodexAuth::Chatgpt(_)));
+    if !managed_chatgpt_auth {
```
celia-oai: nit: if we could do it in a natural and readable way let's combine a few of these `if` cases that all fall back to resolve_provider_auth to make it slightly more readable

codex-rs/model-provider/src/auth.rs
```diff
@@ -95,6 +131,70 @@ pub(crate) fn resolve_provider_auth(
...
+        session_source,
+        agent_identity_session_fallback,
+    } = scope;
+    if let Some(CodexAuth::AgentIdentity(agent_identity_auth)) = auth {
+        return Ok(Arc::new(AgentIdentityAuthProvider {
+            auth: agent_identity_auth.clone(),
+        }));
+    }
+
+    let managed_chatgpt_auth = agent_identity_policy == AgentIdentityAuthPolicy::ChatGptAuth
```
celia-oai: nit: let's create a helper function for this. I always have to think for a second what this means

codex-rs/core/src/client.rs
adrian-openai: This split is intentional. `api_auth_for_scope` is the session/run-scoped Codex request path, where managed ChatGPT auth may bootstrap Agent Identity and where the session fallback marker matters. The remaining plain `api_auth()`/`resolve_provider_auth()` callers are unscoped or provider-owned paths; explicit JWT Agent Identity still works there because the auth snapshot is already `CodexAuth::AgentIdentity`.

codex-rs/codex-api/src/auth.rs
adrian-openai: Re-reading this, I see what you mean: this should **not** be added to the public `AuthProvider` trait if it only exists for Agent Identity. Good point. I’ll keep the trait focused on applying auth and instead return the safe agent/task metadata alongside the resolved provider auth from `model-provider`.

codex-rs/core/src/client.rs
adrian-openai: Good catch. This was broader than needed: `/realtime/calls` did not have request telemetry before this PR, so threading `SessionTelemetry` here adds a new OTEL event rather than just populating Agent Identity fields on an existing one. I’ll remove this and keep the Agent Identity OTEL fields on the existing request/websocket telemetry paths.

codex-rs/core/src/session/session.rs
adrian-openai: Do you mean making this an explicit `ModelClient` construction input? I don’t think `SessionServices` should own a separate copy, since `ModelClient` is the thing that resolves request auth and is cloned/used independently of `SessionServices`. I’m happy to make the `ModelClient` ownership more explicit if that’s the concern, but I don’t think a second `SessionServices` field would simplify it.

codex-rs/core/src/session_startup_prewarm.rs
adrian-openai: This is the non-WebSocket equivalent of the existing startup prewarm. Agent Identity bootstrap/fallback happens inside `current_client_setup()`, so when websockets are disabled we otherwise would not attempt it until the first user request. I’ll add a short comment to make that relationship explicit.

codex-rs/otel/src/events/session_telemetry.rs
adrian-openai: Same answer here: these are request-auth fields, not environment fields. In the core request path they come from the resolved auth provider via AuthRequestTelemetryContext, then SessionTelemetry just emits the safe values into the existing OTEL slots.

codex-rs/model-provider/src/models_endpoint.rs
adrian-openai: I don’t think auth_env is the right bucket for these. auth_env is static environment/config presence, while agent_id / task_id come from the resolved request auth provider and can be absent after session fallback. I’d keep these request-auth scoped rather than storing them in env telemetry.

codex-rs/core/src/session/session.rs
celia-oai: oh yes, i misread the lines haha. please make it an explicit decalred field for model client yes

adrian-openai: Got it. I’ll make `agent_identity_policy` a required `ModelClient::new` argument and remove `with_agent_identity_policy`, so every construction site selects the policy explicitly.

codex-rs/login/src/auth/agent_identity.rs
```diff
@@ -60,12 +60,17 @@ pub enum AgentIdentityAuthError {
 
 impl AgentIdentityAuthError {
     pub fn is_bootstrap_unavailable(error: &std::io::Error) -> bool {
```
celia-oai: nit: we should just need to keep one of this or bootstrap_unavailable

codex-rs/login/src/auth/manager.rs
```diff
@@ -2006,15 +2024,64 @@ impl AuthManager {
                 .acquire()
                 .await
                 .map_err(std::io::Error::other)?;
-            return auth
+            let forced_chatgpt_workspace_id = self.forced_chatgpt_workspace_id();
```
celia-oai, line 2074: a cleaner/more readable way for this is to make cool down operations part of `impl AgentIdentityBootstrapCooldown`, like this:
```
#[derive(Debug)]
struct CachedAgentIdentityBootstrapFailure {
    account_id: String,
    authapi_base_url: String,
    retry_at: Instant,
    error: AgentIdentityAuthError,
}

#[derive(Debug, Default)]
struct AgentIdentityBootstrapCooldown {
    failure: Option<CachedAgentIdentityBootstrapFailure>,
}

impl AgentIdentityBootstrapCooldown {
    fn error_for(
        &mut self,
        account_id: &str,
        authapi_base_url: &str,
        now: Instant,
    ) -> Option<AgentIdentityAuthError> {}

    fn record_failure(
        &mut self,
        account_id: String,
        authapi_base_url: String,
        error: AgentIdentityAuthError,
        now: Instant,
    ) {}

      fn clear(&mut self) {}
```

codex-rs/login/src/auth/agent_identity.rs
adrian-openai: Agreed. `bootstrap_unavailable` is the one the cooldown path needs because it returns the typed error, so I’ll remove the boolean wrapper and update the tests to use `.is_some()` or `.is_none()`.

codex-rs/login/src/auth/manager.rs
adrian-openai: Agreed. I’ll split the cached failure from the cooldown wrapper and move lookup/expiry, recording, and clearing into `AgentIdentityBootstrapCooldown`, leaving `AuthManager::agent_identity_auth` focused on orchestrating bootstrap.

celia-oai (approved):

PR #19053, opened 2026-04-22
Default Fast service tier for eligible ChatGPT plans

shijie-oai:
## Why

Enterprise and business-like ChatGPT plans should get Codex's Fast service tier by default when the user or caller has not made an explicit service-tier choice. At the same time, callers need a durable way to choose standard routing without adding a new persisted `standard` service tier value. This keeps existing config compatibility while letting core own the managed default policy.

## What changed

- Resolve the effective service tier in core at session creation: explicit `fast` or `flex` wins, explicit null/clear or `[notice].fast_default_opt_out = true` resolves to standard routing, and otherwise eligible ChatGPT plans resolve to Fast when FastMode is enabled.
- Add `[notice].fast_default_opt_out` as the persisted opt-out marker for managed Fast defaults.
- Treat app-server/TUI `service_tier: null` as an explicit standard/clear choice by preserving that intent through config loading.
- Update TUI rendering to use core's effective service tier for startup and status surfaces while still keeping `config.service_tier` as the explicit configured choice.
- Update `/fast off` to clear `service_tier`, persist the opt-out marker, and send explicit standard for subsequent turns.

## Verification

- Added unit coverage for config override/notice handling, service-tier resolution, runtime null clearing, and `/fast off` turn propagation.
- `cargo build -p codex-cli`

Full test suite was not run locally per author request.

### DIFF
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -2031,14 +2031,24 @@ impl Config {
         let forced_login_method = cfg.forced_login_method;
 
         let model = model.or(config_profile.model).or(cfg.model);
-        let service_tier = service_tier_override
-            .unwrap_or_else(|| config_profile.service_tier.or(cfg.service_tier));
+        let mut notices = cfg.notice.unwrap_or_default();
+        let service_tier = match service_tier_override {
+            Some(Some(service_tier)) => Some(service_tier),
+            Some(None) => {
+                // Preserve explicit standard/clear intent after the nested override
+                // collapses into `Config.service_tier = None`.
+                notices.fast_default_opt_out = Some(true);
+                None
+            }
+            None => config_profile.service_tier.or(cfg.service_tier),
+        };
         let service_tier = match service_tier {
             Some(ServiceTier::Fast) if features.enabled(Feature::FastMode) => {
                 Some(ServiceTier::Fast)
             }
+            Some(ServiceTier::Fast) => None,
             Some(ServiceTier::Flex) => Some(ServiceTier::Flex),
-            _ => None,
+            None => None,
         };
 
         let compact_prompt = compact_prompt.or(cfg.compact_prompt).and_then(|value| {
@@ -2414,7 +2424,7 @@ impl Config {
             active_profile: active_profile_name,
             active_project,
             windows_wsl_setup_acknowledged: cfg.windows_wsl_setup_acknowledged.unwrap_or(false),
-            notices: cfg.notice.unwrap_or_default(),
+            notices,
             check_for_update_on_startup,
             disable_paste_burst: cfg.disable_paste_burst.unwrap_or(false),
             analytics_enabled: config_profile
--- a/codex-rs/core/src/session/mod.rs
+++ b/codex-rs/core/src/session/mod.rs
@@ -77,6 +77,7 @@ use codex_otel::current_span_w3c_trace_context;
 use codex_otel::set_parent_from_w3c_trace_context;
 use codex_protocol::ThreadId;
 use codex_protocol::ToolName;
+use codex_protocol::account::PlanType as AccountPlanType;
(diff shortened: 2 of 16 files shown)


codex-rs/tui/src/chatwidget/slash_dispatch.rs
```diff
@@ -172,7 +172,7 @@ impl ChatWidget {
                 self.open_model_popup();
             }
             SlashCommand::Fast => {
-                let next_tier = if matches!(self.config.service_tier, Some(ServiceTier::Fast)) {
+                let next_tier = if matches!(self.current_service_tier(), Some(ServiceTier::Fast)) {
```
shijie-oai, line 175: We cannot toggle based on the config value because enterprise users can be on fast but not have the service tier set because it is carried in memory.

codex-rs/tui/src/app/event_dispatch.rs
```diff
@@ -1029,14 +1029,24 @@ impl App {
...
                     .with_profile(profile)
-                    .set_service_tier(service_tier)
-                    .apply()
-                    .await
-                {
+                    .set_service_tier(service_tier);
+                if service_tier.is_none() {
+                    self.config.notices.fast_default_opt_out = Some(true);
+                    edits = edits.set_fast_default_opt_out(/*opted_out*/ true);
+                }
```
shijie-oai, line 1039: When serice_tier went from true to false, we would record the event which the users has explicitly set the service tier and enterprise default would not longer be applied.

codex-rs/core/src/session/mod.rs
```diff
@@ -784,6 +794,25 @@ impl Codex {
...
+    fast_mode_enabled: bool,
+) -> Option<ServiceTier> {
+    if configured_service_tier.is_some() || fast_default_opt_out || !fast_mode_enabled {
+        return configured_service_tier;
+    }
+
+    account_plan_type
+        .is_some_and(is_enterprise_default_service_tier_plan)
+        .then_some(ServiceTier::Fast)
+}
```
shijie-oai, line 810: This is the key of the logic, if the user does not have the config set and has not explicitly opt out and fast mode is enabled, we would then go ahead and default to fast mode.

codex-rs/core/src/config/mod.rs
```diff
@@ -2030,14 +2030,24 @@ impl Config {
...
+        let mut notices = cfg.notice.unwrap_or_default();
+        let service_tier = match service_tier_override {
+            Some(Some(service_tier)) => Some(service_tier),
+            Some(None) => {
+                // Preserve explicit standard/clear intent after the nested override
+                // collapses into `Config.service_tier = None`.
+                notices.fast_default_opt_out = Some(true);
+                None
+            }
+            None => config_profile.service_tier.or(cfg.service_tier),
```
shijie-oai, line 2043: Some(None) -> explicitly passing in null from client, we would like to treat that as the highest priority and passing `fast_default_opt_out` as true to override enterprise default behavior.
None -> omitted, we will have core takeover the behavior which is the calculated default value.

codex-rs/tui/src/chatwidget/status_surfaces.rs
```diff
@@ -482,7 +482,7 @@ impl ChatWidget {
             )),
             StatusLineItem::SessionId => self.thread_id.map(|id| id.to_string()),
             StatusLineItem::FastMode => Some(
-                if matches!(self.config.service_tier, Some(ServiceTier::Fast)) {
+                if matches!(self.current_service_tier(), Some(ServiceTier::Fast)) {
```
shijie-oai, line 485: No longer matching the exact config value but the value in memory.

shijie-oai: so that we show the correct status in TUI

dylan-hurd-oai (approved):

sayan-oai (approved):

PR #19054, opened 2026-04-22
obsolete: hidden background agent auth provider

adrian-openai:
> [!NOTE]
> This PR is obsolete for the current simplified single-run-task Agent Identity stack. It has already been merged, has no current review surface, and should not be treated as part of the active PR sequence.

The active stack is:

- [#19047](https://github.com/openai/codex/pull/19047) - run task identity primitives
- [#19049](https://github.com/openai/codex/pull/19049) - ChatGPT auth can register/reuse a persisted Agent Identity record
- [#19051](https://github.com/openai/codex/pull/19051) - first-party inference and compact requests use the run task `AgentAssertion` path

The text that used to be here described the earlier external-task-ref/background-helper design. That shape has been collapsed out of the stack, so reviewers should ignore this PR when reviewing the current design.

### DIFF
--- a/codex-rs/core/src/client.rs
+++ b/codex-rs/core/src/client.rs
@@ -116,8 +116,10 @@ use crate::util::emit_feedback_auth_recovery_tags;
 use codex_api::map_api_error;
 use codex_feedback::FeedbackRequestTags;
 use codex_feedback::emit_feedback_request_tags_with_auth_env;
+use codex_login::auth::AgentIdentityAuthPolicy;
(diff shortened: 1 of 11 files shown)


efrazer-oai (commented):
Any reason to add 'Background' when we already have 'Process'? They seem to do ~the same thing, might be able to cut out a few additions.

adrian-openai:
> Any reason to add 'Background' when we already have 'Process'? They seem to do ~the same thing, might be able to cut out a few additions.

Good catch @efrazer-oai - let me update this!

PR #19058, opened 2026-04-23
Add /auto-review-denials retry approval flow

won-openai:
## Why

Auto-review can deny an action that the user later decides they want to retry. Today there is no TUI surface for selecting a recent denial and sending explicit approval context back into the session, so users have to restate intent manually and the retry can be reviewed without the original denied action context.

This adds a narrow TUI-driven path for approving a recent denied action while still keeping the retry inside the normal auto-review flow.

## What Changed

- Added `/auto-review-denials` to open a picker of recent denied auto-review actions.
- Added a small in-memory TUI store for the 10 most recent denied auto-review events.
- Selecting a denial sends the structured denied event back through the existing core/app-server op path.
- Core now injects a developer message containing the approved action JSON rather than the full assessment event.
- Auto-review transcript collection now preserves this specific approval developer message so follow-up review sessions can see the user approval context.
- Added TUI snapshot/unit coverage for the picker and approval dispatch path.
- Added core coverage for retaining the approval developer message in the auto-review transcript.

## Verification

- `cargo test -p codex-core collect_guardian_transcript_entries_keeps_manual_approval_developer_message`
- `cargo test -p codex-tui auto_review_denials`
- `cargo test -p codex-tui approving_recent_denial_emits_structured_core_op_once`

## Notes

This intentionally keeps retries going through auto-review. The approval signal is context for the exact previously denied action, not a blanket bypass for similar future actions.

### DIFF
--- a/codex-rs/core/src/guardian/prompt.rs
+++ b/codex-rs/core/src/guardian/prompt.rs
@@ -14,6 +14,7 @@ use codex_utils_output_truncation::approx_bytes_for_tokens;
 use codex_utils_output_truncation::approx_token_count;
 use codex_utils_output_truncation::approx_tokens_from_byte_count;
 
+use super::AUTO_REVIEW_DENIED_ACTION_APPROVAL_DEVELOPER_PREFIX;
 use super::GUARDIAN_MAX_MESSAGE_ENTRY_TOKENS;
 use super::GUARDIAN_MAX_MESSAGE_TRANSCRIPT_TOKENS;
 use super::GUARDIAN_MAX_TOOL_ENTRY_TOKENS;
@@ -33,6 +34,7 @@ pub(crate) struct GuardianTranscriptEntry {
 
 #[derive(Debug, PartialEq, Eq)]
 pub(crate) enum GuardianTranscriptEntryKind {
+    Developer,
     User,
     Assistant,
     Tool(String),
@@ -41,6 +43,7 @@ pub(crate) enum GuardianTranscriptEntryKind {
 impl GuardianTranscriptEntryKind {
     fn role(&self) -> &str {
         match self {
+            Self::Developer => "developer",
             Self::User => "user",
             Self::Assistant => "assistant",
             Self::Tool(role) => role.as_str(),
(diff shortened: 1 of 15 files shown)


won-openai:
@codex review

etraut-openai:
`/auto-review-denials` is a pretty long slash command name — much longer than any others we currently have. I also don't think we should have built-in slash commands with hyphens in the name. Maybe shorten to just `/autoreview`?

won-openai:
> `/auto-review-denials` is a pretty long slash command name — much longer than any others we currently have. I also don't think we should have built-in slash commands with hyphens in the name. Maybe shorten to just `/autoreview`?

or we can do /denials, i think it's more intuitive! thoughts? @etraut-openai @dylan-hurd-oai

etraut-openai:
I guess I'd lean toward `/autoreview` over `/denials`. I suspect we may want to provide additional controls over auto-review parameters in the future, and this would allow us to expand the UI without adding another slash command.

codex-rs/core/src/guardian/prompt.rs
```diff
@@ -361,6 +364,15 @@ pub(crate) fn collect_guardian_transcript_entries(
                     content_entry(GuardianTranscriptEntryKind::User, content)
                 }
             }
+            ResponseItem::Message { role, content, .. } if role == "developer" => {
+                content_items_to_text(content).and_then(|text| {
+                    text.starts_with(AUTO_REVIEW_DENIED_ACTION_APPROVAL_DEVELOPER_PREFIX)
```
dylan-hurd-oai, line 372: this feels like a footgun. can we add a comment here outlining the intent?

won-openai:
@codex review

codex-rs/core/src/guardian/prompt.rs
```diff
@@ -361,6 +364,15 @@ pub(crate) fn collect_guardian_transcript_entries(
                     content_entry(GuardianTranscriptEntryKind::User, content)
                 }
             }
+            ResponseItem::Message { role, content, .. } if role == "developer" => {
```
won-openai, line 367: so that auto-review is able to send the override message

codex-rs/core/src/session/handlers.rs
```diff
@@ -1247,20 +1247,26 @@ async fn approve_guardian_denied_action(sess: &Arc<Session>, event: GuardianAsse
         return;
```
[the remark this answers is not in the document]
won-openai: ..?

codex-rs/tui/src/chatwidget.rs
```diff
@@ -9679,6 +9688,80 @@ impl ChatWidget {
...
+                    });
+                })],
+                dismiss_on_select: true,
+                ..Default::default()
+            }
+        }));
+
+        self.bottom_pane.show_selection_view(SelectionViewParams {
+            title: Some("Auto-review Denials".to_string()),
+            subtitle: Some("Select one denied action to approve for one retry.".to_string()),
```
dylan-hurd-oai: `Select a denied action to approve.` - let's keep this simple for now

codex-rs/tui/src/slash_command.rs
```diff
@@ -21,6 +21,8 @@ pub enum SlashCommand {
     #[strum(serialize = "sandbox-add-read-dir")]
     SandboxReadRoot,
     Experimental,
+    #[strum(to_string = "autoreview")]
+    AutoReviewDenials,
```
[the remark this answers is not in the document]
dylan-hurd-oai: @won-openai which one are we using here? Let's keep this simple

dylan-hurd-oai (approved):
Approved with a few small things to address

PR #19068, opened 2026-04-23
Unified mentions in TUI

canvrno-oai:
This PR replaces the TUI’s file-only `@mention` popup with a unified mentions experience. Typing `@...` now searches across filesystem matches, installed plugins, and skills in one popup, with result types clearly labeled and selectable from the same flow.

- Adds a unified `@mentions` popup that returns:
  - plugins
  - skills
  - files
  - directories

- Adds search modes so users can narrow the popup without changing their query:
  - All Results _(default/same as Codex App)_
  - Filesystem Only
  - Plugins _(...and skills)_

- Preserves existing insertion behavior:
  - selected file paths are inserted into the prompt
  - paths with spaces are quoted
  - image file selections still attach as images when possible
  - selecting a plugin or skill inserts the corresponding `$name`
  - the composer records the canonical mention binding, such as `plugin://...` or the skill path

- Expanded `@mentions` rendering:
  - type tags for Plugin, Skill, File, and Dir
  - distinct plugin/filesystem colors
  - stable fixed-height layout (8 rows)
  - truncation behavior for narrow terminals

Note:
- The unified mentions popup does not display app connectors under `@mention` results for Codex App parity. Connector mentions remain available through the existing `$mention` path.

https://github.com/user-attachments/assets/f93781ed-57d3-4cb5-9972-675bc5f3ef3f


etraut-openai:
I haven't yet reviewed the code, but I was playing with the new feature.
* I really like the unification behavior, the color coding, and the ability to filter based on type!
* One visual issue: the horizontal offset of the second column changes based on the max width of the first column, which means that it jumps around as I scroll. This looks unpolished. We've received bug reports for this in other parts of the TUI, and we've fixed those bugs.
* The ordering of recommendations seems quit suboptimal and not what I'd expect. For example, when I type a string that exactly matches a substring of a file name, I'd expect that file to be ranked above a skill whose name is a fuzzy or partial match for that string. Also, when I type an exact match for the first part of a filename, I'd expect that file to be above a different file where the string appears in the middle of the file name (e.g. if I type "foo", then "foobar.txt" should be ranked above "abcdefoo.txt".

fcoury-oai:
I also played with this feature a bit and I also really like the visuals. 

Adding to Eric's notes, one thing I think could be beneficial would be to keep the current type selection (All/Filesystem/Plugins) persisted throughout the current session. Not a blocker in any way, just a suggestion.

canvrno-oai:
@codex  review

fcoury-oai (approved):

etraut-openai (approved):

PR #19085, opened 2026-04-23
Persist target default reasoning on model upgrade

shijie-oai:
## Why

When the TUI upgrade flow moves a user to a newer model, the accepted migration should also persist the target model's default reasoning effort. That keeps the upgraded model and reasoning setting aligned instead of carrying forward a stale previously saved effort from the old model.

## What changed

- The accepted model migration path now updates in-memory config, TUI state, and persisted model selection with the target preset's `default_reasoning_effort`.
- The upgrade destructuring keeps `reasoning_effort_mapping` explicitly unused because mappings are no longer consulted on accepted migrations.
- Added a catalog test that starts with a pre-existing saved reasoning effort and verifies the accepted upgrade overwrites it with the target model default and emits the expected persistence events.
- Rebasing onto current `main` also updates a TUI thread-session test helper for the latest `permission_profile` field and `ApprovalsReviewer::AutoReview` rename so CI compiles on the new base.

## Verification

- `cargo test -p codex-tui model_catalog`
- `cargo test -p codex-tui permission_settings_sync_updates_active_snapshot_without_rewriting_side_thread`

### DIFF
--- a/codex-rs/tui/src/app/startup_prompts.rs
+++ b/codex-rs/tui/src/app/startup_prompts.rs
@@ -139,6 +139,28 @@ pub(super) fn target_preset_for_upgrade<'a>(
         .find(|preset| preset.model == target_model && preset.show_in_picker)
 }
 
+pub(super) fn apply_accepted_model_migration(
+    config: &mut Config,
+    app_event_tx: &AppEventSender,
+    from_model: String,
+    target_model: String,
+    target_default_effort: ReasoningEffortConfig,
+) {
+    app_event_tx.send(AppEvent::PersistModelMigrationPromptAcknowledged {
+        from_model,
+        to_model: target_model.clone(),
+    });
+
+    config.model = Some(target_model.clone());
+    config.model_reasoning_effort = Some(target_default_effort);
+    app_event_tx.send(AppEvent::UpdateModel(target_model.clone()));
+    app_event_tx.send(AppEvent::UpdateReasoningEffort(Some(target_default_effort)));
+    app_event_tx.send(AppEvent::PersistModelSelection {
+        model: target_model,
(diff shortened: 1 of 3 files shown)


codex-rs/tui/src/app/startup_prompts.rs
```diff
@@ -139,6 +139,28 @@ pub(super) fn target_preset_for_upgrade<'a>(
...
+
+    config.model = Some(target_model.clone());
+    config.model_reasoning_effort = Some(target_default_effort);
+    app_event_tx.send(AppEvent::UpdateModel(target_model.clone()));
+    app_event_tx.send(AppEvent::UpdateReasoningEffort(Some(target_default_effort)));
+    app_event_tx.send(AppEvent::PersistModelSelection {
+        model: target_model,
+        effort: Some(target_default_effort),
+    });
+}
```
shijie-oai, line 162: extract the logic out for testing.

codex-rs/tui/src/app/startup_prompts.rs
```diff
@@ -218,11 +240,11 @@ pub(super) async fn handle_model_migration_prompt_if_needed(
 
     if let Some(ModelUpgrade {
         id: target_model,
-        reasoning_effort_mapping,
         migration_config_key,
         model_link,
         upgrade_copy,
         migration_markdown,
+        ..
```
dylan-hurd-oai: nit: let's leave this as explicitly unused

dylan-hurd-oai (approved):

codex-rs/tui/src/app/startup_prompts.rs
shijie-oai: - [x] Address

PR #19098, opened 2026-04-23
feat: Compress skill paths with root aliases

xl-openai:
Add skill root tracking so model-visible skill lists can use short path aliases when absolute paths would exceed the metadata budget.


codex-rs/core-skills/src/render.rs
```diff
@@ -510,6 +608,262 @@ fn render_lines_with_description_budget(
...
+        SKILLS_HOW_TO_USE_WITH_ABSOLUTE_PATHS.to_string(),
+    ];
+    let aliased_lines = vec![
+        SKILLS_INTRO_WITH_ALIASES.to_string(),
+        "### Skill roots".to_string(),
+        SKILLS_HOW_TO_USE_WITH_ALIASES.to_string(),
+    ];
+
+    lines_cost(budget, &aliased_lines).saturating_sub(lines_cost(budget, &absolute_lines))
+}
```
sayan-oai: can we easily reuse the actual model-visible text constructed in AvailableSkillsInstructions::body() rather than constructing roughly the same text here?

codex-rs/core/src/session/mod.rs
```diff
@@ -2627,12 +2627,8 @@ impl Session {
             }
         }
         if turn_context.config.include_skill_instructions {
-            let implicit_skills = turn_context
-                .turn_skills
-                .outcome
-                .allowed_skills_for_implicit_invocation();
             let available_skills = build_available_skills(
-                &implicit_skills,
+                &turn_context.turn_skills.outcome,
```
sayan-oai, line 2631: we're changing model-visible context, can we have an integration test that asserts it takes the expected shape?

sayan-oai (approved):

sayan-oai (approved):
tests needed and one other comment but lgtm

PR #19160, opened 2026-04-23
Make apply_patch streaming parser stateful

akshaynathan:
Adds an incremental "streaming" parser for apply_patch. Instead of reparsing the entire patch every time, the incremental parser holds the current state and only parses the delta (line by line). Local n=1 benchmark shows this is 10-15x faster on reasonably sized patches.


pakrym-oai:
I think the streaming parser also requires a trailing newline which the old parser didn't

codex-rs/apply-patch/src/parser.rs
```diff
@@ -997,112 +780,3 @@ fn test_parse_patch_lenient() {
         ))
     );
 }
-
-#[test]
-fn test_parse_one_hunk() {
-    assert_eq!(
```
pakrym-oai, line 1003: should we keep this one?

pakrym-oai: and the one below?

pakrym-oai (approved):

PR #19170, opened 2026-04-23
Skip disabled rows in selection menu numbering and default focus

canvrno-oai:
Selection menus in the TUI currently let disabled rows interfere with numbering and default focus. This makes mixed menus harder to read and can land selection on rows that are not actionable. This change updates the shared selection-menu behavior in list_selection_view so disabled rows are not selected when these views open, and prevents them from being numbered like selectable rows.

- Disabled rows no longer receive numeric labels
- Digit shortcuts map to enabled rows only
- Default selection moves to the first enabled row in mixed menus
- Updated affected snapshot
- Added snapshot coverage for a plugin detail error popup
- Added a focused unit test for shared selection-view behavior


canvrno-oai:
@codex review

codex-rs/tui/src/bottom_pane/list_selection_view.rs
```diff
@@ -426,6 +435,7 @@ impl ListSelectionView {
                         .position(|idx| *idx == actual_idx)
                 })
             })
+            .or_else(|| self.first_enabled_visible_idx())
```
fcoury-oai, line 445: With Codex's help I found a potential issue:

If the user types to search an item, the code currently tries to keep the selection position. If the position now points to an item that is disabled, it doesn't move to the next enabled item. If the user presses Enter on that disabled item, nothing happens because the highlighted item can't be chosen.

In other words, filtering can leave the UI highlighting an item that cannot be selected, even though there are selectable items available.

The expected behavior is that whenever the visible list changes, the selected row should still be valid. If the preserved row is disabled, selection should move to the first enabled visible row, or otherwise use the same fallback logic that initial construction uses.

fcoury-oai (commented):
Codex found a small pending issue. Let me know how you want to proceed and I'll be glad to re-review whenever you're ready.

canvrno-oai:
@fcoury-oai I've addressed the issue that was identified by Codex.

fcoury-oai (approved):

PR #19176, opened 2026-04-23
Add conditional network proxy prompt guidance

viyatb-oai:
## Why
When a network proxy is active, the model needs enough context to keep proxy routing intact, distinguish explicit policy denials from ordinary network failures, and know when to request more network permission instead of trying to work around the sandbox. Without that guidance, it can misread proxy-mediated failures or override the environment that is enforcing policy.

## What changed
- Add a dedicated `# Network Proxy` section to the permissions instructions bundle when the current turn actually has an active proxy.
- Teach the model to honor `<network>` allow/deny entries, preserve proxy-related environment variables, and request more network permission for hosts outside policy.
- Spell out the useful denial signals (`blocked-by-allowlist`, `blocked-by-denylist`, and local/private target blocks) while clarifying that proxy-mediated clients can also time out or hang, so generic failures alone are not proof of a proxy denial.
- Keep the prompt plumbing explicit by grouping the feature-dependent render flags in `PermissionsInstructionOptions`.

## Testing
- `cargo test -p codex-core includes_network_proxy_guidance_when_proxy_is_active`
- `cargo test -p codex-core permissions_message_includes_writable_roots`


codex-rs/core/src/context/permissions_instructions.rs
```diff
@@ -105,6 +107,7 @@ impl PermissionsInstructions {
     ) -> Self {
         let mut text = String::new();
         append_section(&mut text, &sandbox_text(sandbox_mode, network_access));
+        append_section(&mut text, NETWORK_PROXY.trim_end());
```
dylan-hurd-oai: naively, this feels like a reasonable spot for this

PR #19180, opened 2026-04-23
use a version-specific suffix for command runner binary in .sandbox-bin

iceweasel-oai:
we copy `codex-command-runner.exe` into `CODEX_HOME/.sandbox-bin/` so that it can be executed by the sandbox user.
We also detect if that version is stale and copy a new one in if so.

This can fail when you are running multiple versions of the app - the file in `.sandbox-bin` can look stale because it comes from another app build.

This change allows us to have multiple versions in there for different CLI versions, and it fallsback to a `size+mtime` hash in the filename for dev builds that don't report a real CLI version.


iceweasel-oai:
@codex review

bschoepke-openai (approved):
ty!

PR #19184, opened 2026-04-23
fix: handle deferred network proxy denials

viyatb-oai:
## Why

This bug is exposed by Guardian/auto-review approvals. With the managed network proxy enabled, a blocked network request can be reported back through the network approval service as an approval denial after the command has already started. Before this change, the shell and unified exec runtimes registered those network approval calls, but did not have a way to observe an async proxy denial as a cancellation/failure signal for the running process.

The result was confusing: Guardian/auto-review could correctly deny network access, but the command path could keep running or unregister the approval without surfacing the denial as the command failure.

## What Changed

- `NetworkApprovalService` now attaches a cancellation token to active and deferred network approvals.
- Proxy-denial outcomes are recorded only for active registrations, cancel the owning token, and are consumed when the approval is finalized.
- The shell runtime combines the normal command timeout with the network-denial cancellation token.
- Unified exec stores the deferred network approval object, terminates tracked processes when the proxy denial arrives, and returns the denial as a process failure while polling or completing the process.
- Tool orchestration passes the active network approval cancellation token into the sandbox attempt and preserves deferred approval errors instead of silently unregistering them.
- App-server `command/exec` now handles the combined timeout-or-cancellation expiration variant used by the runtime.

## Verification

- `cargo test -p codex-core network_approval --lib`
- `cargo clippy -p codex-app-server --all-targets -- -D warnings`


dylan-hurd-oai (commented):
generally looks good, but i'd like @jif-oai to look at unified exec. would be great to have slightly more detail in the denial messages

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -69,6 +69,7 @@ const UNIFIED_EXEC_ENV: [(&str, &str); 10] = [
     ("GH_PAGER", "cat"),
     ("CODEX_CI", "1"),
 ];
+const NETWORK_ACCESS_DENIED_MESSAGE: &str = "network access was denied";
```
dylan-hurd-oai: can we make this piece clearer?

dylan-hurd-oai (approved):
I think we're okay to proceed for now, but would like the denial messages to be a bit clearer about where the network access denial is coming from. I think lack of specificity will lead to model confusion here

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -217,17 +221,102 @@ impl UnifiedExecProcessManager {
...
             session
                 .services
                 .network_approval
-                .unregister_call(network_approval_id)
+                .unregister_call(network_approval.registration_id())
                 .await;
         }
     }
 
+    async fn finish_network_approval_for_entry(
```
jif-oai: Does all those functions really need to be part of the process manager impl? 
None of those depend on self...

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -352,6 +459,11 @@ impl UnifiedExecProcessManager {
                     ..
                 } => (Some(process_id), exit_code),
                 ProcessStatus::Exited { exit_code, .. } => {
+                    Self::finish_deferred_network_approval_for_session(
```
jif-oai: There is a race where a deferred network denial can be consumed by `finish_deferred_network_approval_for_session` after the earlier `is_cancelled` check, but before `failure_message` is set... In that path `exec_command` returns `ProcessFailed` without calling `fail_and_terminate`

jif-oai (commented):
Unified exec still builds ExecOptions { expiration: ExecExpiration::DefaultTimeout, ... } without wiring attempt.network_denial_cancellation_token into the spawned request. The manager-side watcher covers the normal process after spawn, but anything inside the spawn/ preparation path, especially zsh-fork escalation setup, is not cancelled through the same expiration path as shell commands. That makes unified exec less robust than the shell path added in this PR.

codex-rs/core/src/unified_exec/process_manager.rs
viyatb-oai: good catch, fixing!

codex-rs/core/src/unified_exec/process_manager.rs
viyatb-oai: moving these out

viyatb-oai:
@codex review

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -178,6 +182,88 @@ fn exec_server_process_id(process_id: i32) -> String {
...
+) {
+    let network_cancelled = deferred.cancellation_token();
+    let process_exited = process.cancellation_token();
+    tokio::spawn(async move {
+        tokio::select! {
+            _ = network_cancelled.cancelled() => {}
+            _ = process_exited.cancelled() => {
+                if !network_cancelled.is_cancelled() {
+                    return;
+                }
```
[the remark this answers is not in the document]
jif-oai: That sounds quite right to me

codex-rs/core/src/tools/network_approval.rs
```diff
@@ -215,29 +243,30 @@ impl NetworkApprovalService {
...
-        self.call_outcomes.lock().await.remove(registration_id);
+        self.remove_call(registration_id).await;
     }
 
     async fn resolve_single_active_call(&self) -> Option<Arc<ActiveNetworkApprovalCall>> {
-        let active_calls = self.active_calls.lock().await;
-        if active_calls.len() == 1 {
-            return active_calls.values().next().cloned();
+        let calls = self.calls.lock().await;
+        if calls.active_calls.len() == 1 {
```
jif-oai, line 295: This makes denial cancellation work only when exactly one network-backed command is active. Can't we have to live execs? (or more ofc)

codex-rs/core/src/exec.rs
```diff
@@ -184,10 +197,52 @@ impl ExecExpiration {
             ExecExpiration::Timeout(duration) => Some(duration.as_millis() as u64),
             ExecExpiration::DefaultTimeout => Some(DEFAULT_EXEC_COMMAND_TIMEOUT_MS),
             ExecExpiration::Cancellation(_) => None,
+            ExecExpiration::TimeoutOrCancellation { timeout, .. } => {
+                Some(timeout.as_millis() as u64)
+            }
```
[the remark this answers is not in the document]
jif-oai: This sounds correct. If you close the comment please explain why

codex-rs/core/src/unified_exec/process_manager.rs
```diff
@@ -320,27 +428,50 @@ impl UnifiedExecProcessManager {
...
-                    message.clone(),
-                    wall_time,
-                )
-                .await;
-            }
+        if deferred_network_approval
+            .as_ref()
+            .is_some_and(DeferredNetworkApproval::is_cancelled)
+        {
+            let message = network_denial_message_for_session(
```
jif-oai, line 469: This can race with the background denial task, because both consume the stored denial message... I'm not exactly sure what would be the impact (and codex does not know) but please confirm this is not problematic

codex-rs/core/src/unified_exec/process_manager.rs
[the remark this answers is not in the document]
viyatb-oai: Fixed in 4a5751acbe. Unified exec now keeps a bounded 100 ms post-exit watch for a late network-denial cancellation before finalizing the deferred registration, and the process-denial watcher uses the same grace path. Added `late_network_denial_grace_observes_cancellation_after_exit`; the short-lived/background denial regressions still pass as well.

codex-rs/core/src/exec.rs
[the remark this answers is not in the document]
viyatb-oai: fixed in https://github.com/openai/codex/pull/19880, which handles the case for windows. i split the pr

codex-rs/core/src/tools/network_approval.rs
viyatb-oai: Yes. Multiple live network-backed commands are possible. right now the proxy observer gives us an unattributed `BlockedRequest`, so this path only cancels when exactly one active call exists; with multiple active calls we cannot safely know which registration to fail, and intentionally avoid canceling the wrong one.

I've tried to fix this before but it ends up being a messy fix in the proxy code with registration/call ids in the headers and its brittle (an id via proxy credentials in the proxy URL).
I'll call this out explicitly in the code.

PR #19193, opened 2026-04-23
Support Codex Apps auth elicitations

mzeng-openai:
## Summary

- request URL-mode MCP elicitations when Codex Apps tool calls fail with connector auth metadata
- route Codex Apps auth URL elicitations into the TUI app-link flow

## Test plan

- `just fmt`
- `cargo test -p codex-core mcp_tool_call::tests`
- `cargo test -p codex-mcp`
- `cargo test -p codex-tui bottom_pane::app_link_view::tests`
- `just fix -p codex-core`
- `just fix -p codex-mcp`
- `just fix -p codex-tui`

Also attempted broader local runs:

- `cargo test -p codex-core` fails in unrelated config/request-permission/proxy-sensitive tests under the current Codex Desktop environment.
- `cargo test -p codex-tui` fails in unrelated status snapshots/trust-default tests because the ambient environment renders workspace-write/network permission defaults.


codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -472,28 +473,294 @@ fn record_server_fields(span: &Span, url: Option<&str>) {
...
+            key.to_string(),
+            serde_json::Value::String(value.to_string()),
+        );
+    }
+}
+
+fn codex_apps_auth_elicitation_message(auth_failure: &CodexAppsConnectorAuthFailure) -> String {
+    match auth_failure.auth_reason.as_deref() {
+        Some("oauth_upgrade_required") => format!(
+            "Reconnect {} on ChatGPT to grant the permissions needed for this request.",
```
zamoshchin-openai: nit: but on ChatGPT we are using this message, do you want to match it? 

Image




i.e. "Codex needs additional access to {}"

zamoshchin-openai (approved):

viyatb-oai:
This auth-URL elicitation path looks like a trust-boundary expansion: any MCP server can now supply a URL that the TUI later passes straight to `webbrowser::open(...)`. A malicious or compromised MCP server could present this as a normal sign-in prompt while driving the user to an attacker-controlled HTTPS page, or to a custom protocol URL if the platform/browser handler accepts it. The existing app-link flow was effectively pinned to first-party ChatGPT app URLs, so generic MCP URL elicitations should not inherit that same trusted browser-open path.

Suggested fix: validate before opening, ideally at the browser-open sink rather than relying on each caller to pass safe strings. Parse with `url::Url`, reject non-absolute URLs, reject schemes other than `https`, reject embedded credentials, and for `codex_apps` require an allowlisted ChatGPT host. For generic MCP auth URLs, keep the exact host visible in the UI and route them through an explicitly external/untrusted sign-in flow instead of the first-party app-link flow.

mzeng-openai:
> This auth-URL elicitation path looks like a trust-boundary expansion: any MCP server can now supply a URL that the TUI later passes straight to `webbrowser::open(...)`. A malicious or compromised MCP server could present this as a normal sign-in prompt while driving the user to an attacker-controlled HTTPS page, or to a custom protocol URL if the platform/browser handler accepts it. The existing app-link flow was effectively pinned to first-party ChatGPT app URLs, so generic MCP URL elicitations should not inherit that same trusted browser-open path.
> 
> Suggested fix: validate before opening, ideally at the browser-open sink rather than relying on each caller to pass safe strings. Parse with `url::Url`, reject non-absolute URLs, reject schemes other than `https`, reject embedded credentials, and for `codex_apps` require an allowlisted ChatGPT host. For generic MCP auth URLs, keep the exact host visible in the UI and route them through an explicitly external/untrusted sign-in flow instead of the first-party app-link flow.

Added https check and trusted domains.

codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -476,28 +477,298 @@ fn record_server_fields(span: &Span, url: Option<&str>) {
...
+
+    let Some(auth_failure) = codex_apps_connector_auth_failure(&result, metadata) else {
+        return result;
+    };
+
+    let request_id = rmcp::model::RequestId::String(codex_apps_auth_elicitation_id(call_id).into());
+    let params =
+        build_codex_apps_auth_elicitation_request(sess, turn_context, call_id, &auth_failure);
+    let response = sess
+        .request_mcp_server_elicitation(turn_context, request_id, params)
```
viyatb-oai, line 652: This auth prompt path bypasses the MCP elicitation policy. Normal MCP elicitations go through `ElicitationRequestManager`, which declines when `approval_policy = never` or granular `mcp_elicitations = false`; this new path only checks `Feature::AuthElicitation` and then directly emits an `ElicitationRequest` via `request_mcp_server_elicitation`.

Suggested fix: gate this before building/sending the URL elicitation, using the same policy semantics as normal MCP elicitations, and add regression tests for `AskForApproval::Never` plus granular `mcp_elicitations: false`.

```rust
fn auth_elicitation_is_rejected_by_policy(approval_policy: codex_protocol::protocol::AskForApproval) -> bool {
    match approval_policy {
        codex_protocol::protocol::AskForApproval::Never => true,
        codex_protocol::protocol::AskForApproval::Granular(config) => {
            !config.allows_mcp_elicitations()
        }
        codex_protocol::protocol::AskForApproval::OnFailure
        | codex_protocol::protocol::AskForApproval::OnRequest
        | codex_protocol::protocol::AskForApproval::UnlessTrusted => false,
    }
}

if auth_elicitation_is_rejected_by_policy(turn_context.approval_policy.value()) {
    return result;
}
```

Without this, an admin/user setting intended to suppress MCP elicitations can still get a Codex Apps sign-in prompt.

codex-rs/tui/src/bottom_pane/app_link_view.rs
```diff
@@ -66,6 +90,189 @@ pub(crate) struct AppLinkViewParams {
...
+            thread_id,
+            server_name,
+            request_id,
+            AuthUrlParts {
+                meta: meta.as_ref(),
+                message,
+                url,
+                elicitation_id,
+            },
+            AuthUrlRouting::WithGenericFallback,
```
viyatb-oai: This app-server TUI path re-enables generic URL auth elicitations for non-`codex_apps` servers. The active-thread path uses `CodexAppsOnly`, and `codex-mcp` now declines custom-server URL elicitations, but inactive app-server requests deliberately use `WithGenericFallback`, so an inactive custom MCP URL request can still render as a trusted sign-in link.

Suggested fix: keep the app-server path Codex-Apps-only too, then delete/update the generic fallback tests and snapshot (`non_codex_apps_app_server_auth_url_elicitation_uses_generic_app_link`, `inactive_thread_url_elicitation_routes_to_app_link`, and the generic auth snapshot). If custom MCP URL auth is needed later, it should land behind a separate feature/spec gate.

```suggestion
            AuthUrlRouting::CodexAppsOnly,
```

mzeng-openai:
@viyatb addressed both. Auth elicitations now respect approval-policy blockers (never and granular MCP elicitations disabled skip the flow), and app-server URL auth is Codex Apps-only with the generic app-link fallback removed.

codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -476,28 +478,309 @@ fn record_server_fields(span: &Span, url: Option<&str>) {
...
+        .get(CONNECTOR_AUTH_FAILURE_IS_AUTH_FAILURE_KEY)
+        .and_then(serde_json::Value::as_bool)
+        != Some(true)
+    {
+        return None;
+    }
+
+    let connector_id =
+        string_auth_failure_field(auth_failure, CONNECTOR_AUTH_FAILURE_CONNECTOR_ID_KEY)
+            .or_else(|| metadata.and_then(|metadata| metadata.connector_id.clone()))?;
```
viyatb-oai: Non-blocking security note: this prefers the connector id reported by the tool-result metadata over the connector id from the invoked tool metadata. If a buggy or malicious Codex Apps connector returns an auth-failure block for a different connector, Codex can show an auth prompt labeled with the invoked connector while opening a ChatGPT auth URL for the metadata-specified connector.

Suggested fix: bind the auth elicitation to the invoked tool identity. Require `metadata.connector_id` to be present, reject the auth prompt if the result metadata includes a different `connector_id`, and then build the install URL from the trusted metadata connector id/name rather than letting result metadata choose the target.

viyatb-oai (commented):
Non-blocking security review note.

viyatb-oai (approved):

codex-rs/core/src/mcp_tool_call.rs
mzeng-openai: Addressed in b5029ed50c: auth elicitations now require the invoked tool metadata connector_id, reject a mismatched result connector_id, and build the install URL from the invoked connector metadata.

codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -476,28 +478,319 @@ fn record_server_fields(span: &Span, url: Option<&str>) {
...
+    sess: &Session,
+    turn_context: &TurnContext,
+    call_id: &str,
+    server: &str,
+    metadata: Option<&McpToolApprovalMetadata>,
+    result: CallToolResult,
+) -> CallToolResult {
+    if server != CODEX_APPS_MCP_SERVER_NAME {
+        return result;
+    }
```
bolinfest: I have commented on this in the past and I want us to stop doing this. We should be using capabilities, not MCP server names because MCP server names can be spoofed, which makes this code misleading.

codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -535,28 +540,133 @@ fn truncate_str_to_char_boundary(value: &str, max_chars: usize) -> &str {
...
+    match manager.hard_refresh_codex_apps_tools_cache().await {
+        Ok(mcp_tools) => connectors::refresh_accessible_connectors_cache_from_mcp_tools(
+            &turn_context.config,
+            auth.as_ref(),
+            &mcp_tools,
+        ),
+        Err(err) => {
+            tracing::warn!("failed to refresh Codex Apps tools after connector auth: {err:#}");
+        }
+    }
```
bolinfest, line 687: Can this be done instead? Does the lock on `auth` need to be held during sess.services.mcp_connection_manager.read()?

```suggestion
async fn refresh_codex_apps_after_connector_auth(sess: &Session, turn_context: &TurnContext) {
    let mcp_tools_result = {
        let manager = sess.services.mcp_connection_manager.read().await;
        manager.hard_refresh_codex_apps_tools_cache().await
    };

    match mcp_tools_result {
        Ok(mcp_tools) => {
            let auth = sess.services.auth_manager.auth().await;
            connectors::refresh_accessible_connectors_cache_from_mcp_tools(
                &turn_context.config,
                auth.as_ref(),
                &mcp_tools,
            );
        }
        Err(err) => {
            tracing::warn!("failed to refresh Codex Apps tools after connector auth: {err:#}");
        }
    }
```

bolinfest (approved):

codex-rs/core/src/mcp_tool_call.rs
mzeng-openai: Switched this path to gate on `McpConnectionManager::is_host_owned_codex_apps_server`, so the auth flow no longer trusts the server name alone.

PR #19206, opened 2026-04-23
app-server: persist device key bindings in sqlite

euroelessar:
## Why

Device-key providers should only own platform key material. The account/client binding used to authorize a signing payload is app-server state, and keeping that state in provider-specific metadata makes the same check harder to audit and harder to share across platform implementations.

Persisting the binding in the shared state database gives the device-key crate a platform-neutral source of truth before it asks a provider to sign. It also lets app-server move potentially blocking key operations off the main message processor path, which matters once providers may wait for OS authentication prompts.

## What changed

- Add a `device_key_bindings` state migration plus `StateRuntime` helpers keyed by `key_id`.
- Add an async `DeviceKeyBindingStore` abstraction to `codex-device-key` and use it from `DeviceKeyStore::create` and `DeviceKeyStore::sign`.
- Keep provider calls behind async store methods and run the synchronous provider work through `spawn_blocking`.
- Wire app-server device-key RPC handling to the SQLite-backed binding store and spawn response/error delivery tasks for device-key requests.
- Run the turn-start tracing test on the existing larger current-thread test harness after the larger async surface made the default test stack too small locally.

## Validation

- `cargo test -p codex-state device_key`
- `cargo test -p codex-state`
- `cargo test -p codex-app-server device_key`
- `cargo test -p codex-app-server message_processor::tracing_tests::turn_start_jsonrpc_span_parents_core_turn_spans`
- `just fix -p codex-device-key`
- `just fix -p codex-state`
- `just bazel-lock-update`
- `just bazel-lock-check`
- `git diff --check`

### DIFF
--- a/codex-rs/device-key/src/lib.rs
+++ b/codex-rs/device-key/src/lib.rs
@@ -1,3 +1,4 @@
+use async_trait::async_trait;
 use base64::Engine;
 use base64::engine::general_purpose::URL_SAFE_NO_PAD;
 use p256::pkcs8::EncodePublicKey;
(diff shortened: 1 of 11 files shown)


codex-rs/device-key/src/lib.rs
```diff
@@ -211,70 +212,154 @@ pub enum DeviceKeyError {
...
         request: DeviceKeySignRequest,
     ) -> Result<DeviceKeySignature, DeviceKeyError> {
         let protection_class = validate_key_id(&request.key_id)?;
         validate_payload(&request.payload)?;
-        let binding = self.provider.binding(&request.key_id, protection_class)?;
+        let binding = self
+            .bindings
+            .get_binding(&request.key_id)
+            .await?
+            .ok_or(DeviceKeyError::KeyNotFound)?;
```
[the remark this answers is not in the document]
euroelessar: not applicable yet as no users yet

codex-rs/device-key/src/lib.rs
```diff
@@ -211,47 +212,66 @@ pub enum DeviceKeyError {
...
+        &self,
+        request: DeviceKeyCreateRequest,
+    ) -> Result<DeviceKeyInfo, DeviceKeyError> {
         let key_id_random = random_key_id_random();
         validate_binding(&request.binding.account_user_id, &request.binding.client_id)?;
-        self.provider.create(ProviderCreateRequest {
-            key_id_random: &key_id_random,
-            protection_policy: request.protection_policy,
-            binding: &request.binding,
+        let provider = Arc::clone(&self.provider);
```
etraut-openai, line 232: `create` now creates the provider key first and writes the SQLite binding afterward. If `put_binding` fails after the provider succeeds, the API returns an error but leaves private key material behind with no binding record so future signing cannot reconcile or clean it up. This is an unlikely failure, so not sure how important it is. You could add a delete on failure.

etraut-openai (approved):
Looks mostly good to me. I found one small hole that you might want to fix — up to you.

codex-rs/device-key/src/lib.rs
```diff
@@ -211,47 +212,66 @@ pub enum DeviceKeyError {
...
+        let info = spawn_provider_call(move || {
+            provider.create(ProviderCreateRequest {
+                key_id_random,
+                protection_policy: request.protection_policy,
+            })
         })
+        .await?;
+        self.bindings
```
viyatb-oai: Non-blocking: `create()` now creates provider key material before writing the SQLite binding. If the binding write fails after provider creation succeeds, this returns an error but leaves behind a stranded key with no binding row, so future signing cannot reconcile or use it. I do not think this is a security blocker, but it would be cleaner to add cleanup/delete-on-failure or otherwise make this failure mode recoverable.

viyatb-oai (approved):

PR #19207, opened 2026-04-23
[codex] Forward Codex Apps tool call IDs to backend metadata

rreichel3-oai:
## Summary
- include the outer tool `call_id` in Codex Apps MCP request metadata under `_meta._codex_apps.call_id`
- preserve existing Codex Apps metadata like `resource_uri` and `contains_mcp_source`
- add request metadata coverage for both the existing-metadata and no-existing-metadata cases

## Why
The paired backend change in [openai/openai#850796](https://github.com/openai/openai/pull/850796) updates MCP compliance logging to prefer `_meta._codex_apps.call_id` instead of the JSON-RPC request id. This client change sends that outer tool call id so the backend can record the model/tool call identifier when it is available.

This is wire-compatible with older backends because `_meta._codex_apps` is already reserved backend-only metadata. Backends that do not read `call_id` will ignore the extra field.

## Testing
- `cargo test -p codex-core request_meta`
- `just fmt`
- `just fix -p codex-core`

### DIFF
--- a/codex-rs/core/src/mcp_tool_call_tests.rs
+++ b/codex-rs/core/src/mcp_tool_call_tests.rs
@@ -719,13 +719,13 @@ async fn codex_apps_tool_call_request_meta_includes_turn_metadata_and_codex_apps
         build_mcp_tool_call_request_meta(
             &turn_context,
             CODEX_APPS_MCP_SERVER_NAME,
-            "call-calendar",
+            "call_abc123xyz789",
             Some(&metadata),
         ),
         Some(serde_json::json!({
             crate::X_CODEX_TURN_METADATA_HEADER: expected_turn_metadata,
             MCP_TOOL_CODEX_APPS_META_KEY: {
-                "call_id": "call-calendar",
+                "call_id": "call_abc123xyz789",
                 "resource_uri": "connector://calendar/tools/calendar_create_event",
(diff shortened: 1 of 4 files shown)


mzeng-openai (approved):

codex-rs/core/src/mcp_tool_call_tests.rs
```diff
@@ -668,9 +668,13 @@ async fn mcp_tool_call_request_meta_includes_turn_metadata_for_custom_server() {
     )
     .expect("turn metadata json");
 
-    let meta =
-        build_mcp_tool_call_request_meta(&turn_context, "custom_server", /*metadata*/ None)
-            .expect("custom servers should receive turn metadata");
+    let meta = build_mcp_tool_call_request_meta(
+        &turn_context,
+        "custom_server",
+        "call-custom",
```
bolinfest, line 674: I know this is a test, but isn't a more representative `call_id` something like `"call_abc123xyz789"`?

bolinfest (approved):

codex-rs/core/src/mcp_tool_call_tests.rs
rreichel3-oai: Fixed, thanks for the callout

PR #19209, opened 2026-04-23
[codex] Emit usage-limit prompt analytics from TUI

jchu-oai:
## Summary

- emits a `shown` product analytics event when the member-facing credits-depleted or usage-limit-reached terminal prompt opens
- emits a `cta_clicked` product analytics event when the member selects `Yes`
- preserves the existing owner-nudge email flow; selecting `No` only dismisses the prompt

## Notes

This is stacked on top of the generic app-server/protocol/analytics plumbing PR. It is intentionally limited to the TUI producer layer: credits depleted asks whether to notify the workspace owner, and usage limit reached asks whether to request an increase from the owner.


jchu-oai:
@codex review

PR #15578, opened 2026-03-23
Add Windows sandbox unified exec runtime support

iceweasel-oai:
## Summary

This is the runtime/foundation half of the Windows sandbox unified-exec work.

- add Windows sandbox `unified_exec` session support in `windows-sandbox-rs` for both:
  - the legacy restricted-token backend
  - the elevated runner backend
- extend the PTY/process runtime so driver-backed sessions can support:
  - stdin streaming
  - stdout/stderr separation
  - exit propagation
  - PTY resize hooks
- add Windows sandbox runtime coverage in `codex-windows-sandbox` / `codex-utils-pty`

This PR does **not** enable Windows sandbox `UnifiedExec` for product callers yet because hooking this up to app-server comes in the next PR.

Windows sandbox advertising is intentionally kept aligned with `main`, so sandboxed Windows callers still fall back to `ShellCommand`.

This PR isolates the runtime/session layer so it can be reviewed independently from product-surface enablement.

PR #19211, opened 2026-04-23
Improve Windows process management edge cases

iceweasel-oai:
## Summary

Some improvements to Windows process-management issues from https://github.com/openai/codex/pull/15578

- bound the elevated runner pipe-connect handshake instead of waiting forever on blocking pipe connects
- terminate the spawned runner if that handshake fails, so timeout/error paths do not leave a stray `codex-command-runner.exe`
- loop on partial `WriteFile` results when forwarding stdin in the elevated runner, so input is not silently truncated
- fix the concrete HANDLE/SID cleanup paths in the runner setup code
- keep draining driver-backed stdout/stderr after exit until the backend closes, instead of dropping the tail after a fixed 200ms grace period
- reuse `LocalSid` for SID ownership and add more explanatory comments around the ownership/concurrency-sensitive code paths

## Why

The original PR fixed a lot of Windows session plumbing, but there were still a few sharp process-lifecycle edges:

- some elevated runner handshakes could block forever
- the new timeout path could still orphan the spawned runner process
- stdin forwarding still assumed a single `WriteFile` consumed the whole buffer
- a few raw HANDLE/SID error paths still leaked
- driver-backed output could still lose the last chunk of stdout/stderr on slower backends

## Validation

- `cargo fmt -p codex-windows-sandbox -p codex-utils-pty`
- `cargo test -p codex-utils-pty`
- `cargo test -p codex-windows-sandbox finish_driver_spawn`
- `cargo test -p codex-windows-sandbox runner_`

Ran a local test matrix of unified-exec and shell_tool tests, all passing

### DIFF
--- a/codex-rs/windows-sandbox-rs/src/elevated/runner_client.rs
+++ b/codex-rs/windows-sandbox-rs/src/elevated/runner_client.rs
@@ -12,6 +12,7 @@ use crate::runner_pipe::find_runner_exe;
 use crate::runner_pipe::pipe_pair;
 use crate::winutil::quote_windows_arg;
 use crate::winutil::to_wide;
+use anyhow::Context;
 use anyhow::Result;
 use std::ffi::c_void;
 use std::fs::File;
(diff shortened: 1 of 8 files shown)


iceweasel-oai:
@codex review

codex-rs/windows-sandbox-rs/src/elevated/runner_client.rs
```diff
@@ -136,28 +278,40 @@ pub(crate) fn spawn_runner_transport(
...
             CloseHandle(h_pipe_out);
         }
         return Err(err);
     }
 
+    unsafe {
+        if pi.hProcess != 0 {
+            // After the handshake succeeds we no longer need the runner process HANDLE. The
+            // pipes are now the lifetime anchor for the transport.
+            CloseHandle(pi.hProcess);
```
jif-oai, line 343: We still close the runner process handle before `send_spawn_request` / `read_spawn_ready`; if that later spawn_ready wait times out, the caller can no longer terminate the runner, so this can still leave a stray `codex-command-runner.exe`

jif-oai (approved):
The startup ownership boundary still feels split: `spawn_runner_transport` owns the runner process only through pipe connect, but `send_spawn_request` / `read_spawn_ready` are still part of startup and can fail or time out after the process handle is closed

codex-rs/windows-sandbox-rs/src/elevated/runner_client.rs
iceweasel-oai: addressed in https://github.com/openai/codex/pull/19211/changes/094c3069fbc592dc162f78844cf76d39d4c6b65f

PR #19216, opened 2026-04-23
Allow any granular approval requirement

abhinav-oai:
## Why
Customer needs managed requirements to allow granular approval policies without enumerating every combination of granular flags. Before this change, `allowed_approval_policies` compared concrete `AskForApproval` objects exactly, so allowing any granular approval-policy shape meant listing every permutation.

This adds a scoped wildcard for the granular policy family only. It does not add a global `any` approval-policy escape hatch.

## What Changed
- `requirements.toml` can now include `{ granular = "any" }` in `allowed_approval_policies`.
- Config parsing stores approval allow-list entries as `ApprovalPolicyConstraint`, where concrete policies still exact-match and `{ granular = "any" }` matches any `AskForApproval::Granular(_)`.
- `configRequirements/read` now exposes the same constraint shape as `ApprovalPolicyConstraint`, preserving `{ "granular": "any" }` for clients instead of collapsing it into one concrete granular policy.
- App-server schema and generated TypeScript fixtures were regenerated for the new constraint type.

## Validation
- `cargo test -p codex-config allowed_approval_policies`
- `cargo test -p codex-app-server map_requirements_toml_to_api_preserves_any_granular_approval_constraint`


codex-rs/config/src/config_requirements.rs
```diff
@@ -642,6 +643,118 @@ pub struct ConfigRequirementsToml {
...
+pub enum ApprovalPolicyConstraint {
+    #[serde(rename = "untrusted")]
+    UnlessTrusted,
+    OnFailure,
+    OnRequest,
+    Granular(GranularApprovalConstraint),
+    Never,
+}
+
+impl ApprovalPolicyConstraint {
```
abhinav-oai, line 657: this file has the core changes

alexsong-oai (approved):

codex-rs/config/src/config_requirements.rs
alexsong-oai: LGTM

PR #19229, opened 2026-04-23
Add agent graph store interface

rasmusrygaard:
## Summary

Persisted subagent parent/child topology currently leaks through `StateRuntime`'s SQLite-specific thread-spawn helpers. This PR introduces a narrow `AgentGraphStore` boundary so follow-up work can route graph operations through a local or remote store without coupling orchestration code directly to the state DB graph API.

## Changes

- Adds the new `codex-agent-graph-store` crate.
- Defines a flat `AgentGraphStore` trait for the v1 graph surface: upsert edge, set edge status, list direct children, and list descendants.
- Adds public graph types for `ThreadSpawnEdgeStatus`, `AgentGraphStoreError`, and `AgentGraphStoreResult`.
- Implements `LocalAgentGraphStore` on top of an existing `codex_state::StateRuntime`, preserving today's SQLite-backed `thread_spawn_edges` behavior.
- Registers the crate in Cargo/Bazel metadata.

This PR only adds the local contract and implementation; call-site migration and the remote gRPC store are left to the follow-up PRs in the stack.

## Testing

- `cargo test -p codex-agent-graph-store`

The new unit tests cover local parity with the existing `StateRuntime` graph methods, `Open`/`Closed` filtering, status updates, and stable breadth-first descendant ordering.

### DIFF
--- a/codex-rs/agent-graph-store/src/local.rs
+++ b/codex-rs/agent-graph-store/src/local.rs
@@ -155,9 +155,9 @@ mod tests {
         let fixture = state_runtime().await;
         let state_db = fixture.state_db;
         let store = LocalAgentGraphStore::new(state_db.clone());
-        let parent_thread_id = thread_id(1);
-        let first_child_thread_id = thread_id(2);
-        let second_child_thread_id = thread_id(3);
+        let parent_thread_id = thread_id(/*suffix*/ 1);
+        let first_child_thread_id = thread_id(/*suffix*/ 2);
+        let second_child_thread_id = thread_id(/*suffix*/ 3);
 
         store
             .upsert_thread_spawn_edge(
@@ -177,7 +177,7 @@ mod tests {
             .expect("open child edge should insert");
 
         let all_children = store
-            .list_thread_spawn_children(parent_thread_id, None)
+            .list_thread_spawn_children(parent_thread_id, /*status_filter*/ None)
             .await
             .expect("all children should load");
         assert_eq!(
@@ -211,8 +211,8 @@ mod tests {
         let fixture = state_runtime().await;
         let state_db = fixture.state_db;
         let store = LocalAgentGraphStore::new(state_db);
-        let parent_thread_id = thread_id(10);
-        let child_thread_id = thread_id(11);
+        let parent_thread_id = thread_id(/*suffix*/ 10);
+        let child_thread_id = thread_id(/*suffix*/ 11);
 
         store
             .upsert_thread_spawn_edge(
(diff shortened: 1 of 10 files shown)


codex-rs/agent-graph-store/src/local.rs
```diff
@@ -0,0 +1,340 @@
...
+        &self,
+        parent_thread_id: ThreadId,
+        child_thread_id: ThreadId,
+        status: ThreadSpawnEdgeStatus,
+    ) -> AgentGraphStoreResult<()> {
+        self.state_db
+            .upsert_thread_spawn_edge(parent_thread_id, child_thread_id, to_state_status(status))
+            .await
+            .map_err(internal_error)
+    }
```
rasmusrygaard, line 44: Implementation should match what's already in the repo. Will point those callers at this code eventually

codex-rs/agent-graph-store/src/local.rs
```diff
@@ -0,0 +1,340 @@
...
+    ) -> AgentGraphStoreResult<Vec<ThreadId>> {
+        if let Some(status) = status_filter {
+            return self
+                .state_db
+                .list_thread_spawn_children_with_status(parent_thread_id, to_state_status(status))
+                .await
+                .map_err(internal_error);
+        }
+
+        let mut open_children = self
```
jif-oai: This "all children" branch does two separate reads. If a child flips status between them it can disappear or show up twice here, and any future/unknown status gets silently dropped instead of being returned as "all"

codex-rs/agent-graph-store/src/store.rs
```diff
@@ -0,0 +1,38 @@
+use async_trait::async_trait;
+use codex_protocol::ThreadId;
+
+use crate::AgentGraphStoreResult;
+use crate::ThreadSpawnEdgeStatus;
+
+/// Storage-neutral boundary for persisted thread-spawn parent/child topology.
+#[async_trait]
+pub trait AgentGraphStore: Send + Sync {
```
jif-oai, line 12: Could we doc a bit more the trait? For example the impact of `status_filter` on `list_thread_spawn_descendants` is not trivial to me

codex-rs/agent-graph-store/src/types.rs
```diff
@@ -0,0 +1,11 @@
+use serde::Deserialize;
+use serde::Serialize;
+
+/// Lifecycle status attached to a directional thread-spawn edge.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
```
jif-oai, line 5: Nit, the existing `DirectionalThreadSpawnEdgeStatus` is using `snake_case` for serialization. We should unify otherwise it's an open door to later bugs

jif-oai (approved):
Ok after my comments

codex-rs/agent-graph-store/src/local.rs
rasmusrygaard: Good catch! Seems like a behavior we should fix in list_agent_subtree_thread_ids and upload_feedback where we call the logical method here. Will fix in the interface for now

PR #19161, opened 2026-04-23
ci: derive cache-stable Windows Bazel PATH

bolinfest:
## Why

The BuildBuddy runs for PR #19086 and the later `main` build had the same source tree, but their Windows Bazel action and test cache keys did not line up. Comparing the downloaded execution logs showed the full GitHub-hosted Windows runner `PATH` had changed from `apache-maven-3.9.14` to `apache-maven-3.9.15`.

This repo is not using Maven; the Maven entry was just ambient hosted-runner state. The problem was that Windows Bazel CI was still forwarding the whole runner `PATH` into Bazel via `--action_env=PATH`, `--host_action_env=PATH`, and `--test_env=PATH`, which made otherwise reusable cache entries sensitive to unrelated runner image churn.

After discussion with the Bazel and BuildBuddy folks, the better shape for this change was to stop asking Bazel to inherit the ambient Windows `PATH` and instead compute one explicit cache-stable `PATH` in the Windows setup action that already prepares the CI toolchain environment.

## What

- remove Windows `PATH` passthrough from `.bazelrc`
- export `CODEX_BAZEL_WINDOWS_PATH` from `.github/actions/setup-bazel-ci/action.yml`
- move the PATH derivation logic into `.github/scripts/compute-bazel-windows-path.ps1` so the allow-list is easier to review and document
- keep only the Windows tool locations these Bazel jobs actually need: MSVC and SDK paths, Git, PowerShell, Node, DotSlash, and the standard Windows system directories
- update `.github/scripts/run-bazel-ci.sh` to require that explicit value and forward it to Bazel action, host action, and test environments
- log the derived `CODEX_BAZEL_WINDOWS_PATH` in the setup step to simplify cache-key debugging

## Windows `%PATH%`

On the Windows CI jobs for this PR, note the `%CODEX_BAZEL_WINDOWS_PATH%` we derive is quite long, _and_ many of the path entries have version numbers, so upgrading any of them would result in a full rebuild because the change to `%PATH%` would invalidate the existing cache entries:

```
Derived CODEX_BAZEL_WINDOWS_PATH entries:
  C:\Program Files\PowerShell\7
  C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.44.35207\bin\HostX64\x64
  C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\IDE\VC\VCPackages
[...]

PR #19232, opened 2026-04-23
ci: reuse Bazel CI startup for target-discovery queries

bolinfest:
## Why

A rerun of the Windows Bazel clippy job after [#19161](https://github.com/openai/codex/pull/19161) had exactly the cache behavior we wanted in BuildBuddy: zero action-cache misses. Even so, the GitHub job still took a little over five minutes.

The problem was that the job was paying for two separate Bazel startup paths:

1. a `bazel query` to discover extra lint targets
2. the real `bazel build --config=clippy ...` invocation

On Windows, that query was bypassing the CI Bazel wrapper, so it did not reuse the same `--output_user_root`, CI config, or remote-cache setup as the real build. In practice that meant the rerun could still cold-start a separate Bazel server before the actual clippy build even began.

## What

- add `.github/scripts/run-bazel-query-ci.sh` to run CI-side Bazel queries with the same startup and cache-related flags as the main Bazel command
- switch `scripts/list-bazel-clippy-targets.sh` to use that helper for manual `rust_test` target discovery
- switch `tools/argument-comment-lint/list-bazel-targets.sh` to use the same helper
- simplify `.github/scripts/run-argument-comment-lint-bazel.sh` so its Windows-only query path also goes through the shared helper

This keeps the target-discovery queries aligned with the later build/test invocation instead of treating them as a separate cold Bazel session.

## Verification

- `bash -n .github/scripts/run-bazel-query-ci.sh`
- `bash -n scripts/list-bazel-clippy-targets.sh`
- `bash -n tools/argument-comment-lint/list-bazel-targets.sh`
- `bash -n .github/scripts/run-argument-comment-lint-bazel.sh`
- mocked a Windows invocation of `run-bazel-query-ci.sh` and verified it forwards `--output_user_root`, `--config=ci-windows`, the BuildBuddy auth header, and the repository cache flags

## Docs

No documentation updates are needed.

### DIFF
--- a/.github/scripts/run-argument-comment-lint-bazel.sh
+++ b/.github/scripts/run-argument-comment-lint-bazel.sh
@@ -2,16 +2,6 @@
 
 set -euo pipefail
 
-ci_config=ci-linux
-case "${RUNNER_OS:-}" in
-  macOS)
-    ci_config=ci-macos
-    ;;
-  Windows)
-    ci_config=ci-windows
-    ;;
-esac
-
 bazel_lint_args=("$@")
 if [[ "${RUNNER_OS:-}" == "Windows" ]]; then
   has_host_platform_override=0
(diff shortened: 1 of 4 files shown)


etraut-openai (approved):
Nice find! I don't have experience with bazel, but this looks good to me.

PR #19247, opened 2026-04-24
chore: apply truncation policy to unified_exec

sayan-oai:
we were not respecting turn's `truncation_policy` to clamp output tokens for `unified_exec` and `write_stdin`.

this meant truncation was only being applied by `ContextManager` before the output was stored in-memory (so it _was_ being truncated from model-visible context), but the full output was persisted to rollout on disk.

now we respect that `truncation_policy` and `ContextManager`-level truncation remains a backup.

### Tests
added tests, tested locally.

### DIFF
--- a/codex-rs/core/src/tools/handlers/unified_exec_tests.rs
+++ b/codex-rs/core/src/tools/handlers/unified_exec_tests.rs
@@ -2,6 +2,7 @@ use super::*;
 use crate::shell::default_user_shell;
 use crate::tools::handlers::parse_arguments_with_base_path;
 use crate::tools::handlers::resolve_workdir_base_path;
+use crate::unified_exec::DEFAULT_MAX_OUTPUT_TOKENS;
 use codex_protocol::models::FileSystemPermissions;
 use codex_protocol::models::PermissionProfile;
 use codex_tools::UnifiedExecShellMode;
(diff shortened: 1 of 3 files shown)


codex-rs/core/tests/suite/unified_exec.rs
```diff
@@ -1230,6 +1230,196 @@ async fn exec_command_reports_chunk_and_exit_metadata() -> Result<()> {
...
+    let outputs = collect_tool_outputs(&bodies)?;
+    let start_output = outputs
+        .get(start_call_id)
+        .expect("missing start exec_command output");
+    assert!(
+        start_output.process_id.is_some(),
+        "start command should leave a running process for write_stdin"
+    );
+
+    let stdin_output = outputs
```
pakrym-oai: let's do a stricter regex match, hardcode exact token numbers too

pakrym-oai: search for assert_regex_match

codex-rs/core/tests/suite/unified_exec.rs
```diff
@@ -1230,6 +1230,196 @@ async fn exec_command_reports_chunk_and_exit_metadata() -> Result<()> {
...
+        "expected write_stdin output to be truncated by policy despite large model request: {:?}",
+        stdin_output.output
+    );
+    assert!(
+        stdin_output.output.len() < 1_000,
+        "expected policy-sized write_stdin output, got {} bytes",
+        stdin_output.output.len()
+    );
+    assert!(
+        stdin_output.original_token_count.unwrap_or_default() > 50,
```
pakrym-oai: How do we know it's not global truncation?

codex-rs/core/tests/suite/unified_exec.rs
pakrym-oai: This is so the test is more resistant to accidental changes in logic.

codex-rs/core/src/tools/handlers/unified_exec_tests.rs
```diff
@@ -23,6 +24,34 @@ use crate::tools::registry::ToolHandler;
 use crate::turn_diff_tracker::TurnDiffTracker;
 use tokio::sync::Mutex;
 
+#[test]
+fn max_output_tokens_resolves_default_and_clamps_to_policy() {
```
pakrym-oai: I don't know how useful this test is considering the integration one.

codex-rs/core/tests/suite/unified_exec.rs
sayan-oai: changed to look at `RawResponseItem`s, since those arent truncated by `ContextManager`. if its truncated by then, it must have happened at the `unified_exec` level, as intended.

pakrym-oai (approved):

PR #19280, opened 2026-04-24
[codex] Migrate thread turns list to thread store

wiltzius-openai:
- migrate `thread/turns/list` to ThreadStore. Uses ThreadStore for most data now but merges in the in-memory state from thread manager
- keep v2 `thread/list` pathless-store friendly by converting `StoredThread` directly to API `Thread`
- add regression coverage for pathless store history/listing

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -261,7 +261,6 @@ use codex_core::exec::ExecParams;
 use codex_core::exec_env::create_env;
 use codex_core::find_archived_thread_path_by_id_str;
 use codex_core::find_thread_name_by_id;
-use codex_core::find_thread_names_by_ids;
 use codex_core::find_thread_path_by_id_str;
 use codex_core::path_utils;
 use codex_core::plugins::PluginInstallError as CorePluginInstallError;
@@ -371,7 +370,6 @@ use codex_state::ThreadMetadata;
 use codex_state::ThreadMetadataBuilder;
 use codex_state::log_db::LogDbLayer;
 use codex_thread_store::ArchiveThreadParams as StoreArchiveThreadParams;
-#[cfg(debug_assertions)]
 use codex_thread_store::InMemoryThreadStore;
 use codex_thread_store::ListThreadsParams as StoreListThreadsParams;
 use codex_thread_store::LocalThreadStore;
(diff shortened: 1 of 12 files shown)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -4271,95 +4260,38 @@ impl CodexMessageProcessor {
...
                 // the server has to rebuild the full turn list until turn metadata is indexed
                 // separately.
-                let has_live_in_progress_turn =
-                    match self.thread_manager.get_thread(thread_uuid).await {
-                        Ok(thread) => matches!(thread.agent_status().await, AgentStatus::Running),
-                        Err(_) => false,
-                    };
-                let turns = reconstruct_thread_turns_from_rollout_items(
+                let loaded_thread = self.thread_manager.get_thread(thread_uuid).await.ok();
+                let has_live_in_progress_turn = match loaded_thread.as_ref() {
```
starr-openai: from codex:
```
thread/turns/list only merges the in-memory active turn when agent_status == Running. The existing resume path also checks
     active_turn_snapshot().status == InProgress because turn events can arrive before runtime status catches up. In that window this API can omit the current turn from paginated results. Reuse the resume-path predicate and merge the
     snapshot when present.
```

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -4271,95 +4260,38 @@ impl CodexMessageProcessor {
...
+                    // Persisted history may not yet include the currently running turn. The
+                    // app-server listener has already projected live turn events into ThreadState,
+                    // so merge that in-memory snapshot before paginating.
+                    let thread_state = self.thread_state_manager.thread_state(thread_uuid).await;
+                    let active_turn = {
+                        let state = thread_state.lock().await;
+                        state.active_turn_snapshot()
+                    };
+                    if let Some(active_turn) = active_turn {
+                        merge_turn_history_with_active_turn(&mut turns, active_turn);
```
starr-openai: ^ is this logic that should be consolidated into ThreadStore directly?

wiltzius-openai: I don’t think this should move into ThreadStore directly. ThreadStore is interface for persisted history/metadata state, while this merge is app-server specific: it combines persisted history with the in-memory "live" turn state from thread manager.

There's def an argument to be made that we should have an interface to do that merging, although I've added a lot of interface already so don't know if I'm going too overboard.

codex-rs/app-server/src/codex_message_processor.rs
wiltzius-openai: Added a regression test (that now passes) for this case!

starr-openai (approved):

PR #19283, opened 2026-04-24
check PID of named pipe consumer

iceweasel-oai:
## Why
The elevated Windows command runner currently trusts the first process that connects to its parent-created named pipes. Tightening the pipe ACL already narrows who can reach that boundary, but verifying the connected client PID gives the parent one more fail-closed check: it only accepts the exact runner process it just spawned.

## What changed
- validate `GetNamedPipeClientProcessId` after `ConnectNamedPipe` and reject clients whose PID does not match the spawned runner
- also did some code de-duplication to route the one-shot elevated capture flow in `windows-sandbox-rs/src/elevated_impl.rs` through `spawn_runner_transport()` so both elevated codepaths use the same pipe bootstrap and PID validation

Using the transport unification here also reduces duplication in the elevated Windows IPC bootstrap, so future hardening to the runner handshake only needs to land in one place.

## Validation
- manual testing: one-shot elevated path via `target/debug/codex.exe exec` running a randomized shell command and confirming captured output
- manual testing: elevated session path via `target/debug/codex.exe -c 'windows.sandbox="elevated"' sandbox windows -- python -u -c ...` with stdin/stdout round-trips (`READY`, then `GOT:...` for two input lines)

### DIFF
--- a/codex-rs/windows-sandbox-rs/src/elevated_impl.rs
+++ b/codex-rs/windows-sandbox-rs/src/elevated_impl.rs
@@ -180,7 +180,7 @@ mod windows_impl {
             allow_null_device(psid_to_use);
         }
 
-        let result = (|| -> Result<CaptureResult> {
+        (|| -> Result<CaptureResult> {
             let spawn_request = SpawnRequest {
                 command: command.clone(),
                 cwd: cwd.to_path_buf(),
@@ -240,9 +240,7 @@ mod windows_impl {
                 stderr,
                 timed_out,
             })
-        })();
-
-        result
+        })()
     }
 
     #[cfg(test)]
--- a/codex-rs/windows-sandbox-rs/src/elevated_impl.rs
+++ b/codex-rs/windows-sandbox-rs/src/elevated_impl.rs
@@ -24,57 +24,23 @@ mod windows_impl {
     use crate::env::ensure_non_interactive_pager;
     use crate::env::inherit_path_env;
     use crate::env::normalize_null_device_env;
-    use crate::helper_materialization::HelperExecutable;
-    use crate::helper_materialization::resolve_helper_for_launch;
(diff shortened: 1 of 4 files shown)


codex-rs/windows-sandbox-rs/src/elevated_impl.rs
```diff
@@ -302,133 +180,26 @@ mod windows_impl {
...
+                tty: false,
+                stdin_open: false,
+                use_private_desktop,
             };
-            write_frame(&mut pipe_write, &spawn_request)?;
-            read_spawn_ready(&mut pipe_read)?;
+            let mut transport =
+                spawn_runner_transport(codex_home, cwd, &sandbox_creds, logs_base_dir)?;
+            transport.send_spawn_request(spawn_request)?;
+            transport.read_spawn_ready()?;
```
viyatb-oai, line 201: [P2] Capture path can abandon a live runner during spawn

The one-shot capture path now calls RunnerTransport::read_spawn_ready, which has a hard-coded 15s pre-spawn timeout. A red-team path is to make runner startup or child creation exceed 15s: the parent returns an error and drops the pipe handles while the runner may already have spawned the restricted child and is about to send SpawnReady. Cleanup then depends on the runner reaching its best-effort job assignment/error path, and valid slow Windows launches can fail outside the requested command timeout. The capture path should preserve the old blocking read semantics, derive this from the command timeout, or explicitly terminate the runner on handshake abandonment.

viyatb-oai (approved):

codex-rs/windows-sandbox-rs/src/elevated_impl.rs
iceweasel-oai: fixed in https://github.com/openai/codex/pull/19283/changes/20949c08d8fcd8237e944a92bf671b2712723dfa

PR #19292, opened 2026-04-24
Reject unsupported js_repl image MIME types

etraut-openai:
## Summary

`codex.emitImage` accepted arbitrary image MIME types for byte payloads and data URLs. That allowed a value like `image/rgba` to be wrapped as an `input_image`, even though it is not a supported encoded image format, so the invalid image could reach the model-input path and trigger output sanitization.

This results in a panic in debug builds because the output sanitization is meant as a final safety net, not a primary means of rejecting invalid image types. I've hit this case multiple times when executing certain long-running tasks.

This PR rejects unsupported image MIME types before they are emitted from `js_repl`.

## Changes

- Validate `codex.emitImage({ bytes, mimeType })` in the JS kernel so only encoded PNG, JPEG, WebP, or GIF payloads are accepted.
- Apply the same MIME allowlist to direct image data URLs, including the Rust host-side validation path.
- Clarify the JS REPL instructions so agents know byte payloads must already be encoded as PNG/JPEG/WebP/GIF.


codex-rs/core/src/tools/js_repl/kernel.js
```diff
@@ -1240,9 +1241,37 @@ function normalizeEmitImageUrl(value) {
...
+    throw new Error("codex.emitImage expected a valid image data URL");
+  }
+  const mediaType = dataUrl.slice(5, commaIndex).split(";")[0];
+  if (!mediaType) {
+    throw new Error("codex.emitImage expected image data URL to include a MIME type");
+  }
+  return mediaType;
+}
+
+function validateEmitImageMimeType(mimeType) {
```
bolinfest: I would consider naming `assertEmitImageMimeType()` to underscore the fact that this will throw.

I'm so used to Rust: should we return something `Result`-like instead of throwing? I'm not familiar with how we handle error-handling in this code.

codex-rs/core/src/tools/js_repl/kernel.js
```diff
@@ -1240,9 +1241,37 @@ function normalizeEmitImageUrl(value) {
...
+  validateEmitImageMimeType(mimeType);
   return value;
 }
 
+function parseDataUrlMimeType(dataUrl) {
+  const commaIndex = dataUrl.indexOf(",");
+  if (commaIndex < 0) {
+    throw new Error("codex.emitImage expected a valid image data URL");
+  }
+  const mediaType = dataUrl.slice(5, commaIndex).split(";")[0];
```
bolinfest: Is this where the `5` comes from?

```suggestion
  const mediaType = dataUrl.slice('data:'.length, commaIndex).split(";")[0];
```

codex-rs/core/src/tools/js_repl/kernel.js
```diff
@@ -1240,9 +1241,37 @@ function normalizeEmitImageUrl(value) {
...
+  if (
+    normalized !== "image/png" &&
+    normalized !== "image/jpeg" &&
+    normalized !== "image/webp" &&
+    normalized !== "image/gif"
+  ) {
+    throw new Error(
+      "codex.emitImage only supports image/png, image/jpeg, image/webp, or image/gif",
+    );
+  }
```
bolinfest: Alternative to the repeated `!==`?

```suggestion
const SUPPORTED_IMAGE_TYPES = [
  "image/png",
  "image/jpeg",
  "image/webp",
  "image/gif",
];

if (!SUPPORTED_IMAGE_TYPES.includes(normalized)) {
  throw new Error(
    `codex.emitImage only supports ${SUPPORTED_IMAGE_TYPES.join(", ")}`,
  );
}
```

codex-rs/core/src/tools/js_repl/mod.rs
```diff
@@ -1798,13 +1798,30 @@ fn emitted_image_content_item(
 }
 
 fn validate_emitted_image_url(image_url: &str) -> Result<(), String> {
-    if image_url
+    if !image_url
         .get(..5)
         .is_some_and(|scheme| scheme.eq_ignore_ascii_case("data:"))
     {
+        return Err("codex.emitImage only accepts data URLs".to_string());
+    }
```
bolinfest, line 1806: `strip_prefix()` to the rescue!

```suggestion
    let data_payload = image_url.strip_prefix("data:").ok_or_else(|| anyhow!("expected data: URI but got `{image_url}`"))?;
```

Oh, but you need to support `eq_ignore_ascii_case()`? Is that a thing for URIs?

bolinfest (approved):
I don't feel strongly about any of these comments, FYI!

PR #19351, opened 2026-04-24
Add agents.interrupt_message for interruption markers

jif-oai:
## Why

Agent interruptions currently always persist a model-visible interrupted-turn marker before emitting `TurnAborted`. That marker is useful by default because it gives the next model turn context about a deliberately interrupted task, but some deployments need to suppress that history injection entirely while still keeping the client-visible interruption event.

## What changed

- Add `[agents] interrupt_message = false` to disable the model-visible interrupted-turn marker.
- Resolve the setting into `Config::agent_interrupt_message_enabled`, defaulting to `true` so existing behavior is unchanged.
- Apply the setting to both live interrupted turns and interrupted fork snapshots.
- Keep emitting `TurnAborted` even when the history marker is disabled.
- Regenerate `core/config.schema.json` for the new `agents.interrupt_message` field.

## Testing

- `cargo test -p codex-core load_config_resolves_agent_interrupt_message -- --nocapture`
- `cargo test -p codex-core disabled_interrupted_fork_snapshot_appends_only_interrupt_event -- --nocapture`
- `cargo test -p codex-core multi_agent_v2_interrupted_marker_uses_developer_input_message -- --nocapture`
- `cargo test -p codex-core multi_agent_v2_followup_task_can_disable_interrupted_marker -- --nocapture`
- `cargo test -p codex-core multi_agent_v2_followup_task_interrupts_busy_child_without_losing_message -- --nocapture`
- `cargo check -p codex-core`

### DIFF
--- a/codex-rs/config/src/config_toml.rs
+++ b/codex-rs/config/src/config_toml.rs
@@ -566,6 +566,9 @@ pub struct AgentsToml {
     /// Default maximum runtime in seconds for agent job workers.
     #[schemars(range(min = 1))]
     pub job_max_runtime_seconds: Option<u64>,
+    /// Whether to record a model-visible message when an agent turn is interrupted.
+    /// Defaults to true.
+    pub interrupt_message: Option<bool>,
 
     /// User-defined role declarations keyed by role name.
     ///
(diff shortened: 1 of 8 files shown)


jif-oai:
@codex review

jif-oai:
@codex review

PR #19389, opened 2026-04-24
Guard npm update readiness

shijie-oai:
## Why
For npm/Bun-managed installs, the update prompt was treating the latest GitHub release as ready to install. During the `0.124.0` release, GitHub and npm visibility were not atomic: the root npm wrapper could become visible before the npm registry marked that version as the package `latest`. That left a window where users could be prompted to upgrade before npm was ready for the release.

## What changed
- Keep GitHub Releases as the candidate latest-version source for npm/Bun installs, but only write the existing `version.json` cache after npm registry metadata proves that same root version is ready.
- Add `codex-rs/tui/src/npm_registry.rs` to validate npm readiness by checking `dist-tags.latest` and root package `dist` metadata for the GitHub candidate version.
- Move version parsing helpers into `codex-rs/tui/src/update_versions.rs` so that logic can be tested without compiling the release-only `updates.rs` module under tests.
- Update `.github/workflows/rust-release.yml` so the six known platform tarballs publish before the root `@openai/codex` wrapper. Other npm tarballs publish before the root wrapper, and the SDK publishes after the root package it depends on.

### DIFF
--- a/.github/workflows/rust-release.yml
+++ b/.github/workflows/rust-release.yml
@@ -651,11 +651,59 @@ jobs:
             prefix="${NPM_TAG}-"
           fi
 
+          root_tarball="dist/npm/codex-npm-${VERSION}.tgz"
+          sdk_tarball="dist/npm/codex-sdk-npm-${VERSION}.tgz"
+          # Keep this list in sync with CODEX_PLATFORM_PACKAGES in
+          # codex-cli/scripts/build_npm_package.py. The root wrapper advances
+          # @openai/codex@latest as soon as it publishes, so every platform
+          # package it aliases must already exist in the registry first.
+          platform_tarballs=(
+            "dist/npm/codex-npm-linux-x64-${VERSION}.tgz"
+            "dist/npm/codex-npm-linux-arm64-${VERSION}.tgz"
+            "dist/npm/codex-npm-darwin-x64-${VERSION}.tgz"
+            "dist/npm/codex-npm-darwin-arm64-${VERSION}.tgz"
+            "dist/npm/codex-npm-win32-x64-${VERSION}.tgz"
+            "dist/npm/codex-npm-win32-arm64-${VERSION}.tgz"
+          )
+
+          for required_tarball in "${platform_tarballs[@]}" "${root_tarball}"; do
+            if [[ ! -f "${required_tarball}" ]]; then
+              echo "Missing npm tarball: ${required_tarball}"
+              exit 1
+            fi
+          done
+
           shopt -s nullglob
-          tarballs=(dist/npm/*-"${VERSION}".tgz)
-          if [[ ${#tarballs[@]} -eq 0 ]]; then
-            echo "No npm tarballs found in dist/npm for version ${VERSION}"
-            exit 1
+          other_tarballs=()
+          for tarball in dist/npm/*-"${VERSION}".tgz; do
+            if [[ "${tarball}" == "${root_tarball}" || "${tarball}" == "${sdk_tarball}" ]]; then
+              continue
+            fi
+
+            is_platform_tarball=false
+            for platform_tarball in "${platform_tarballs[@]}"; do
+              if [[ "${tarball}" == "${platform_tarball}" ]]; then
+                is_platform_tarball=true
+                break
+              fi
+            done
+            if [[ "${is_platform_tarball}" == true ]]; then
+              continue
+            fi
+
+            other_tarballs+=("${tarball}")
+          done
+
+          # Publish the platform packages before the root CLI wrapper. The root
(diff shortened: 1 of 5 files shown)


codex-rs/tui/src/updates.rs
```diff
@@ -90,17 +96,20 @@ async fn check_for_update(version_file: &Path) -> anyhow::Result<()> {
...
                 .send()
                 .await?
                 .error_for_status()?
-                .json::<ReleaseInfo>()
+                .json::<NpmPackageInfo>()
                 .await?;
-            extract_version_from_latest_tag(&latest_tag_name)?
+            npm_registry::ensure_version_ready(&package_info, &latest_version)?;
+            latest_version
+        }
```
shijie-oai, line 110: Previouly we only rely on github latest tag but given recent issue there is a clear lag between github package readiness and npm status. Here we are adding a cross validation to make sure both are ready.

.github/workflows/rust-release.yml
```diff
@@ -651,11 +651,59 @@ jobs:
...
+            other_tarballs+=("${tarball}")
+          done
+
+          # Publish the platform packages before the root CLI wrapper. The root
+          # wrapper advances @openai/codex@latest, so it should only publish
+          # after the optional dependency versions it references exist.
+          tarballs=(
+            "${platform_tarballs[@]}"
+            "${other_tarballs[@]}"
+            "${root_tarball}"
```
shijie-oai, line 703: Root tarball is what determines when npm says a version is ready. So even with the cross validation logic below against github latest, we can still run into user wants to upgrade yet it is actually no 100% ready.

codex-rs/tui/src/updates.rs
```diff
@@ -120,18 +129,17 @@ async fn check_for_update(version_file: &Path) -> anyhow::Result<()> {
...
 }
 
-fn is_newer(latest: &str, current: &str) -> Option<bool> {
-    match (parse_version(latest), parse_version(current)) {
-        (Some(l), Some(c)) => Some(l > c),
-        _ => None,
-    }
-}
-
-fn extract_version_from_latest_tag(latest_tag_name: &str) -> anyhow::Result<String> {
```
shijie-oai, line 130: refactored into its own file.

codex-rs/tui/src/updates.rs
```diff
@@ -168,68 +176,3 @@ pub async fn dismiss_version(config: &Config, version: &str) -> anyhow::Result<(
     tokio::fs::write(version_file, json_line).await?;
     Ok(())
 }
-
-fn parse_version(v: &str) -> Option<(u64, u64, u64)> {
```
shijie-oai, line 172: Refactored out into update_version

.github/workflows/rust-release.yml
```diff
@@ -651,11 +651,59 @@ jobs:
...
+            "dist/npm/codex-npm-darwin-arm64-${VERSION}.tgz"
+            "dist/npm/codex-npm-win32-x64-${VERSION}.tgz"
+            "dist/npm/codex-npm-win32-arm64-${VERSION}.tgz"
+          )
+
+          for required_tarball in "${platform_tarballs[@]}" "${root_tarball}"; do
+            if [[ ! -f "${required_tarball}" ]]; then
+              echo "Missing npm tarball: ${required_tarball}"
+              exit 1
+            fi
```
shijie-oai, line 673: So hardcoding the platform is a bit ugly but we can make sure that all tarballs are covered. Alternatively we can use a sort given that we didn't really care about existence prviously.

etraut-openai (approved):
Thanks for doing this!

This PR includes two pieces: a build script change and a client-side change. Of the two, I think the build script change is much more important. 

I have a small concern about the client change. It hardcodes the address of the public npm registry. Most of our users will go through the public registry, but some enterprises may use internal npm registries for supply chain controls. I guess that these enterprises will typically disable our auto-update check using `check_for_update_on_startup = false`.

shijie-oai:
> Thanks for doing this!
> 
> This PR includes two pieces: a build script change and a client-side change. Of the two, I think the build script change is much more important.
> 
> I have a small concern about the client change. It hardcodes the address of the public npm registry. Most of our users will go through the public registry, but some enterprises may use internal npm registries for supply chain controls. I guess that these enterprises will typically disable our auto-update check using `check_for_update_on_startup = false`.

Hm that is a good point with the internal npm registries but there should not be any behavioral difference. Really the only case is check_for_update_on_startup == true (or unset) and distributed through internal registry. With previous approach of only checking against github latest release we would still show upgrade disregarding internal registry status. With the adding cross valiation against public registry that behavior does not change - not ideal but no regression.

PR #19391, opened 2026-04-24
permissions: make runtime config profile-backed

bolinfest:
## Why

`PermissionProfile` is now the canonical permissions shape after #19231 because it can distinguish `Managed`, `Disabled`, and `External` enforcement while also carrying filesystem rules that legacy `SandboxPolicy` cannot represent cleanly. Core config and session state still needed to accept profile-backed permissions without forcing every profile through the strict legacy bridge, which rejected valid runtime profiles such as direct write roots.

## What Changed

- Adds `Permissions.permission_profile` and `SessionConfiguration.permission_profile` as constrained runtime state, while keeping `sandbox_policy` as a legacy compatibility projection.
- Introduces profile setters that keep `PermissionProfile`, split filesystem/network policies, and legacy `SandboxPolicy` projections synchronized.
- Uses a compatibility projection for requirement checks and legacy consumers instead of rejecting profiles that cannot round-trip through `SandboxPolicy` exactly.
- Updates config loading, config overrides, session updates, turn context plumbing, prompt permission text, sandbox tags, and exec request construction to carry profile-backed runtime permissions.
- Preserves configured deny-read entries and `glob_scan_max_depth` when command/session profiles are narrowed.
- Adds `PermissionProfile::read_only()` and `PermissionProfile::workspace_write()` presets that match legacy defaults.

## Verification

- `cargo test -p codex-core direct_write_roots`
- `cargo test -p codex-core runtime_roots_to_legacy_projection`
- `cargo test -p codex-app-server requested_permissions_trust_project_uses_permission_profile_intent`

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/19391).
* #19395
* #19394
* #19393
* #19392
* __->__ #19391
* #19604

PR #19392, opened 2026-04-24
permissions: derive compatibility policies from profiles

bolinfest:
## Why

After #19391, `PermissionProfile` and the split filesystem/network policies could still be stored in parallel. That creates drift risk: a profile can preserve deny globs, external enforcement, or split filesystem entries while a cached projection silently loses those details. This PR makes the profile the runtime source and derives compatibility views from it.

## What Changed

- Removes stored filesystem/network sandbox projections from `Permissions` and `SessionConfiguration`; their accessors now derive from the canonical `PermissionProfile`.
- Derives legacy `SandboxPolicy` snapshots from profiles only where an older API still needs that field.
- Updates MCP connection and elicitation state to track `PermissionProfile` instead of `SandboxPolicy` for auto-approval decisions.
- Adds semantic filesystem-policy comparison so cwd changes can preserve richer profiles while still recognizing equivalent legacy projections independent of entry ordering.
- Updates config/session tests to assert profile-derived projections instead of parallel stored fields.

## Verification

- `cargo test -p codex-core direct_write_roots`
- `cargo test -p codex-core runtime_roots_to_legacy_projection`
- `cargo test -p codex-app-server requested_permissions_trust_project_uses_permission_profile_intent`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/19392).
* #19395
* #19394
* #19393
* __->__ #19392


codex-rs/core/src/session/session.rs
```diff
@@ -127,11 +134,17 @@ impl SessionConfiguration {
...
+        let current_file_system_sandbox_policy = self.file_system_sandbox_policy();
+        let current_network_sandbox_policy = self.network_sandbox_policy();
+        let legacy_file_system_projection =
+            FileSystemSandboxPolicy::from_legacy_sandbox_policy_preserving_deny_entries(
+                &current_sandbox_policy,
                 &self.cwd,
+                &current_file_system_sandbox_policy,
             );
+        let file_system_policy_matches_legacy = current_file_system_sandbox_policy
+            .is_semantically_equivalent_to(&legacy_file_system_projection, &self.cwd);
```
viyatb-oai, line 151: [P1] Preserve absolute cwd-equal write roots across cwd updates

This semantic comparison makes a direct absolute write grant that happens to equal the current cwd indistinguishable from a legacy `:cwd` / `WorkspaceWrite` grant. If the active profile is root read plus `/repo/a` write while cwd is `/repo/a`, this returns true against the legacy projection. A later cwd-only update to `/repo/b` then enters the rederive branch below and rebuilds the profile from `WorkspaceWrite` for the new cwd, silently moving write permission from `/repo/a` to `/repo/b`.

Suggested fix: keep the order-insensitive comparison, but only allow the cwd-rederive path for profiles that are structurally cwd-bound, e.g. require an explicit `FileSystemPath::Special(CurrentWorkingDirectory)` or `ProjectRoots { subpath: None }` write entry before treating the profile as a legacy projection that may be rebound. That preserves the entry-order fix without folding an absolute `Path(old_cwd)` grant into `:cwd`. Please also add a regression test where a profile grants `Path(/repo/a)` write at cwd `/repo/a`, then a cwd-only update to `/repo/b` keeps `/repo/a` writable and does not grant `/repo/b`.

viyatb-oai (approved):

PR #19435, opened 2026-04-24
stop blocking unified_exec on Windows

iceweasel-oai:
## Summary
- remove the Windows-specific unified-exec environment block from tool selection
- keep `unified_exec` default-off on Windows unless the feature is explicitly enabled
- normalize model-provided `shell_type = unified_exec` to `shell_command` when the feature is disabled
- drop obsolete tests tied to the removed environment gate and keep the feature-flag regression coverage

## Why
Now that the session/long-lived process backend is implemented for the Windows sandbox, we don't need to hard disable it anymore. We will be rolling out slowly using a feature gate.

## Impact
This allows manual Windows opt-in in CLI and app-backed flows while preserving the existing default-off behavior for Windows users.


iceweasel-oai:
@codex review

codex-rs/tools/src/tool_config.rs
```diff
@@ -168,26 +166,25 @@ impl ToolsConfig {
...
+                ConfigShellToolType::ShellCommand
+            }
+            other => other,
+        };
         let shell_type = if !features.enabled(Feature::ShellTool) {
             ConfigShellToolType::Disabled
         } else if features.enabled(Feature::ShellZshFork) {
             ConfigShellToolType::ShellCommand
-        } else if features.enabled(Feature::UnifiedExec) && unified_exec_allowed {
+        } else if unified_exec_enabled {
```
jif-oai, line 178: Please add a RestrictedToken/workspace-write regression test here

codex-rs/core/src/tools/spec_tests.rs
```diff
@@ -296,31 +295,6 @@ fn build_specs_with_unavailable_tools(
     )
 }
 
-#[tokio::test]
```
jif-oai, line 299: Should we have a model_provided_unified_exec_respects_platform_default_feature_flag ?

jif-oai (approved):
Ok after my commebnts

codex-rs/core/src/tools/spec_tests.rs
iceweasel-oai: I think that test is covered here
https://github.com/openai/codex/pull/19435/changes/BASE..f4976eeac19b4ae46978b0bb5588d4c02d2fc724#diff-61e43250a78341d7d74201ac34fe28160c73a45729db3e9c0e73e3ddb7ddf683R53

basically tests that even if the model says unified_exec, we don't choose it if the feature is disabled

PR #19442, opened 2026-04-24
feat: disable capabilities by model provider

celia-oai:
## Why

Unsupported features must fail closed and Codex must not expose OpenAI-hosted fallback paths when the active provider cannot support them. In practice, Bedrock should not surface app connectors, MCP servers, tool search/suggestions, image generation, web search, or JS REPL until those paths are explicitly supported for that provider.

This PR moves that decision into provider-owned capability metadata instead of scattering Bedrock-specific checks across callers.

## What changed

- Adds `ProviderCapabilities` to `codex-model-provider`, with default support for existing providers and a Bedrock override that disables unsupported launch surfaces.
- Adds `ToolCapabilityBounds` to `codex-tools` so provider capability limits can clamp otherwise-enabled tool config.
- Applies capability bounds when building session and review-thread tool config.
- Routes MCP/app connector configuration through `McpManager::mcp_config`, which filters configured MCP servers and app connectors based on the active provider.
- Updates app-server MCP list/read paths to use the filtered MCP config.
- Adds coverage for default provider capabilities, Bedrock disabled capabilities, and optional tool-surface clamping.

## Testing

built locally and verified that bedrock responses api now return without errors calling unsupported tools.

### DIFF
--- a/codex-rs/core/src/connectors.rs
+++ b/codex-rs/core/src/connectors.rs
@@ -17,7 +17,6 @@ use codex_connectors::DirectoryListResponse;
 use codex_exec_server::EnvironmentManager;
 use codex_exec_server::EnvironmentManagerArgs;
 use codex_exec_server::ExecServerRuntimePaths;
-use codex_model_provider::create_model_provider;
 use codex_protocol::models::PermissionProfile;
 use codex_tools::DiscoverableTool;
 use rmcp::model::ToolAnnotations;
@@ -232,15 +231,6 @@ pub async fn list_accessible_connectors_from_mcp_tools_with_environment_manager(
     let plugins_manager = Arc::new(PluginsManager::new(config.codex_home.to_path_buf()));
     let mcp_manager = McpManager::new(Arc::clone(&plugins_manager));
     let tool_plugin_provenance = mcp_manager.tool_plugin_provenance(config).await;
-    let capabilities =
-        create_model_provider(config.model_provider.clone(), /*auth_manager*/ None).capabilities();
-    if !capabilities.app_connectors {
-        return Ok(AccessibleConnectorsStatus {
-            connectors: Vec::new(),
-            codex_apps_ready: true,
-        });
-    }
-
     if !force_refetch && let Some(cached_connectors) = read_cached_accessible_connectors(&cache_key)
     {
         let cached_connectors = codex_connectors::filter::filter_disallowed_connectors(
--- a/codex-rs/core/src/connectors.rs
+++ b/codex-rs/core/src/connectors.rs
@@ -17,7 +17,9 @@ use codex_connectors::DirectoryListResponse;
 use codex_exec_server::EnvironmentManager;
 use codex_exec_server::EnvironmentManagerArgs;
 use codex_exec_server::ExecServerRuntimePaths;
+use codex_model_provider::create_model_provider;
 use codex_protocol::models::PermissionProfile;
+use codex_protocol::protocol::SandboxPolicy;
 use codex_tools::DiscoverableTool;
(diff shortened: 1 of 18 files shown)


celia-oai:
@codex review

celia-oai:
chatted with Pavel offline, this can't live on config level cuz feature flag will be gone. reworking it so core depends on model-provider for toggling features

codex-rs/core/src/session/mcp.rs
```diff
@@ -218,7 +218,17 @@ impl Session {
             .mcp_manager
             .tool_plugin_provenance(config.as_ref())
             .await;
-        let mcp_servers = with_codex_apps_mcp(mcp_servers, auth.as_ref(), &mcp_config);
+        let capabilities = turn_context.provider.capabilities();
+        let mcp_servers = if capabilities.mcp_servers {
```
pakrym-oai: Why are we disabling mcp servers? They are exposed like normal function tools to the model. What exactly isn't supported about MCPs?

codex-rs/core/src/session/review.rs
```diff
@@ -41,6 +41,9 @@ pub(super) async fn spawn_review_thread(
         permission_profile: &parent_turn_context.permission_profile,
         windows_sandbox_level: parent_turn_context.windows_sandbox_level,
     })
+    .with_capability_bounds(tool_capability_bounds(
```
pakrym-oai: can we add separate methods per tool/capability/

pakrym-oai: Is this to avoid a dependency?

codex-rs/tools/src/tool_config.rs
```diff
@@ -114,6 +114,25 @@ pub struct ToolsConfig {
...
+    pub tool_suggest: bool,
+    pub image_generation: bool,
+    pub web_search: bool,
+}
+
+impl Default for ToolCapabilityBounds {
+    fn default() -> Self {
+        Self {
+            tool_search: true,
+            tool_suggest: true,
```
pakrym-oai: why isn't tool_suggest supported? Isn't it a normal function tool?

codex-rs/core/src/connectors.rs
```diff
@@ -231,6 +232,15 @@ pub async fn list_accessible_connectors_from_mcp_tools_with_environment_manager(
     let plugins_manager = Arc::new(PluginsManager::new(config.codex_home.to_path_buf()));
     let mcp_manager = McpManager::new(Arc::clone(&plugins_manager));
     let tool_plugin_provenance = mcp_manager.tool_plugin_provenance(config).await;
+    let capabilities =
+        create_model_provider(config.model_provider.clone(), /*auth_manager*/ None).capabilities();
+    if !capabilities.app_connectors {
```
pakrym-oai: how do we even get here with CodexAuth::uses_codex_backend check above?

pakrym-oai (changes requested):
Are we trying to disable more than we need?

codex-rs/core/src/session/review.rs
celia-oai: yes it's to avoid a codex-provider dependency.

codex-rs/tools/src/tool_config.rs
celia-oai: yes you are right - adding it back

codex-rs/core/src/session/review.rs
celia-oai: actually this dependency alr exists. refactored

pakrym-oai (approved):

PR #19449, opened 2026-04-24
permissions: remove legacy read-only access modes

bolinfest:
## Why

`ReadOnlyAccess` was a transitional legacy shape on `SandboxPolicy`: `FullAccess` meant the historical read-only/workspace-write modes could read the full filesystem, while `Restricted` tried to carry partial readable roots. The partial-read model now belongs in `FileSystemSandboxPolicy` and `PermissionProfile`, so keeping it on `SandboxPolicy` makes every legacy projection reintroduce lossy read-root bookkeeping and creates unnecessary noise in the rest of the permissions migration.

This PR makes the legacy policy model narrower and explicit: `SandboxPolicy::ReadOnly` and `SandboxPolicy::WorkspaceWrite` represent the old full-read sandbox modes only. Split readable roots, deny-read globs, and platform-default/minimal read behavior stay in the runtime permissions model.

## What changed

- Removes `ReadOnlyAccess` from `codex_protocol::protocol::SandboxPolicy`, including the generated `access` and `readOnlyAccess` API fields.
- Updates legacy policy/profile conversions so restricted filesystem reads are represented only by `FileSystemSandboxPolicy` / `PermissionProfile` entries.
- Keeps app-server v2 compatible with legacy `fullAccess` read-access payloads by accepting and ignoring that no-op shape, while rejecting legacy `restricted` read-access payloads instead of silently widening them to full-read legacy policies.
- Carries Windows sandbox platform-default read behavior with an explicit override flag instead of depending on `ReadOnlyAccess::Restricted`.
- Refreshes generated app-server schema/types and updates tests/docs for the simplified legacy policy shape.

## Verification

- `cargo check -p codex-app-server-protocol --tests`
- `cargo check -p codex-windows-sandbox --tests`
- `cargo test -p codex-app-server-protocol sandbox_policy_`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/19449).
* #19395
* #19394
* #19393
* #19392
* #19391
* __->__ #19449


codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1802,19 +1777,84 @@ pub enum SandboxPolicy {
...
+            SandboxPolicyDeserialize::WorkspaceWrite {
+                writable_roots,
+                read_only_access,
+                network_access,
+                exclude_tmpdir_env_var,
+                exclude_slash_tmp,
+            } => {
+                if read_only_access.is_present() {
+                    return Err(serde::de::Error::custom(
+                        "workspaceWrite.readOnlyAccess is no longer supported; use permissionProfile for restricted reads",
```
viyatb-oai: [P1] Keep legacy sandbox fields compatible until app clients switch

The Codex app still sends the legacy read-access fields on `sandboxPolicy` today. This deserializer rejects any `readOnly.access` or `workspaceWrite.readOnlyAccess` field, including the old `fullAccess` no-op shape, so a current app talking to this server can fail `turn/start` during params decoding before the turn starts. Please keep a compatibility path here until the app has switched to `permissionProfile`: ignore the legacy `fullAccess` shape and preserve or translate restricted reads instead of rejecting the field outright.

viyatb-oai (approved):

PR #19468, opened 2026-04-25
Fix Bazel cargo_bin runfiles paths

fjord-oai:
## Summary

Fix a Bazel-only path resolution bug in `codex_utils_cargo_bin::cargo_bin`.

Under Bazel runfiles, `rlocation` can return a relative `bazel-out/...` path even though `cargo_bin()` documents that it returns an absolute path. That can break callers that store the returned binary path and later spawn it after changing cwd, because the relative path is resolved from the wrong directory.

This patch absolutizes the runfiles-resolved path before returning it.


bolinfest (approved):
Seems reasonable since CI looks green?

PR #19471, opened 2026-04-25
[codex] Use serial_test for Windows sandbox tests

dylan-hurd-oai:
## Why

Follow-up to #19453. The merged fix serialized the legacy Windows sandbox process tests with a crate-local mutex. The earlier version in dc8f32f used serial_test for this, which keeps the intent clearer and avoids bespoke synchronization code.

## What Changed

- Added serial_test as a dev-dependency for codex-windows-sandbox.
- Replaced the local Mutex guard with use serial_test::serial.
- Applied one named serial group to the same legacy cmd/PowerShell sandbox tests.

## Verification

- just fmt
- cargo test -p codex-windows-sandbox
- just bazel-lock-update
- just bazel-lock-check

### DIFF
--- a/codex-rs/Cargo.lock
+++ b/codex-rs/Cargo.lock
@@ -3771,6 +3771,7 @@ dependencies = [
  "rand 0.8.5",
  "serde",
  "serde_json",
+ "serial_test",
  "tempfile",
  "tokio",
  "windows 0.58.0",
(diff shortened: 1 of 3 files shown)


bolinfest (approved):
I was hoping that https://github.com/openai/codex/pull/18943 would fix this, but let's give this a shot?

PR #19481, opened 2026-04-25
Remove ghost snapshots 

pakrym-oai:
## Summary
- Remove `ghost_snapshot` / `GhostCommit` from the Responses API surface and generated SDK/schema artifacts.
- Keep legacy config loading compatible, but make undo a no-op that reports the feature is unavailable.
- Clean up core history, compaction, telemetry, rollout, and tests to stop carrying ghost snapshot items.

## Testing
- Unit tests passed for `codex-protocol`, `codex-core` targeted undo and compaction flows, `codex-rollout`, and `codex-app-server-protocol`.
- Regenerated config and app-server schemas plus Python SDK artifacts and verified they match the checked-in outputs.


jif-oai (approved):
This will break any session with ghost snapshot in it but we shouldn't have had any in a while so good for me

pakrym-oai:
>This will break any session with ghost snapshot in it but we shouldn't have had any in a while so good for me

We filter shapshots on load. see strip_legacy_ghost_snapshot_rollout_line

PR #19487, opened 2026-04-25
[codex] Move config loading into codex-config

pakrym-oai:
## Why

Config loading had become split across crates: `codex-config` owned the config types and merge logic, while `codex-core` still owned the loader that assembled the layer stack. This change consolidates that responsibility in `codex-config`, so the crate that defines config behavior also owns how configs are discovered and loaded.

To make that move possible without reintroducing the old dependency cycle, the shell-environment policy types and helpers that `codex-exec-server` needs now live in `codex-protocol` instead of flowing through `codex-config`.

This also makes the migrated loader tests more deterministic on machines that already have managed or system Codex config installed by letting tests override the system config and requirements paths instead of reading the host's `/etc/codex`.

## What Changed

- moved the config loader implementation from `codex-core` into `codex-config::loader` and deleted the old `core::config_loader` module instead of leaving a compatibility shim
- moved shell-environment policy types and helpers into `codex-protocol`, then updated `codex-exec-server` and other downstream crates to import them from their new home
- updated downstream callers to use loader/config APIs from `codex-config`
- added test-only loader overrides for system config and requirements paths so loader-focused tests do not depend on host-managed config state
- cleaned up now-unused dependency entries and platform-specific cfgs that were surfaced by post-push CI

## Testing

- `cargo test -p codex-core config_loader_tests::`
- `cargo test -p codex-protocol -p codex-exec-server -p codex-cloud-requirements -p codex-rmcp-client --lib`
- `cargo test --lib -p codex-app-server-client -p codex-exec`
- `cargo test --no-run --lib -p codex-app-server`
- `cargo test -p codex-linux-sandbox --lib`
- `cargo shear`
- `just bazel-lock-check`

## Notes

- I did not chase unrelated full-suite failures outside the migrated loader surface.
- `cargo test -p codex-core --lib` still hits unrelated proxy-sensitive failures on this machine, and Windows CI still shows unrelated long-running/timeouting test noise outside the loader migration itself.


codex-rs/config/src/loader/mod.rs
```diff
@@ -494,16 +450,36 @@ fn system_requirements_toml_file() -> io::Result<AbsolutePathBuf> {
...
+) -> io::Result<AbsolutePathBuf> {
+    match &overrides.system_requirements_path {
+        Some(path) => AbsolutePathBuf::from_absolute_path(path),
+        None => system_requirements_toml_file(),
+    }
+}
+
 #[cfg(unix)]
-fn system_config_toml_file() -> io::Result<AbsolutePathBuf> {
+#[doc(hidden)]
```
bolinfest: I noticed Codex likes to add these sometimes. We should add something to our `AGENTS.md` about this? Please don't add them: they're just distracting in our case and I worry others will copy the pattern.

bolinfest: Current status:

```
$ rg '#\[doc\(hidden\)\]' | wc -l
      12
```

codex-rs/protocol/src/shell_environment.rs
```diff
@@ -26,17 +26,8 @@ where
     let mut env_map = populate_env(vars, policy, thread_id);
 
     if cfg!(target_os = "windows") {
-        // This is a workaround to address the failures we are seeing in the
```
bolinfest, line 29: Is this correct to remove?

codex-rs/exec-server/src/local_process.rs
```diff
@@ -706,7 +706,7 @@ fn notification_sender(inner: &Inner) -> Option<RpcNotificationSender> {
 #[cfg(test)]
```
bolinfest, line 706: Drive-by comment: please adopt the `_tests.rs` pattern in the `exec-server` crate!

codex-rs/core/src/lib.rs
```diff
@@ -25,7 +25,8 @@ mod codex_delegate;
 mod command_canonicalization;
 mod commit_attribution;
 pub mod config;
-pub mod config_loader;
+#[cfg(test)]
```
bolinfest: I guess this is OK, but it's weird we only do this for this and `git_info_tests`. Normally we just include unit tests from the file with the business logic?

bolinfest (approved):
Thanks for the refactor!

PR #19510, opened 2026-04-25
Hide rewind preview when no user message exists

### LINKED ISSUE #19508: Rewind (double escape) should not be shown if there is no user's message.

v0.125.0

Free

gpt-5.4-mini

Darwin 25.4.0 arm64 arm

Ghostty

Tool UX Feedback: Rewind (double escape) should not be shown if there is no user's message.

Currently, when opening Codex CLI and not typing any message, double-pressing the Esc key triggers the Transcript Logs Viewer tool, but it includes no actual message, just a header info TUI box => so showing the transcript doesn't do anything. I think we should have a way to check and prevent the Transcript tool from showing up when there are no messages (both user messages and agent messages). And showing a TUI info message when this happens. Just a quality-of-life improvement suggestion.

Current (no message)
Image

Expect (as normal, has user's messages)

Image

Uploaded thread: 019dc328-31eb-7752-863a-fa87af7be199

Expect: (as normal, has user's messages) => Transcript log is showing only when there are User's message only.

Image

### Additional information

_No response_

etraut-openai:
## Why

Fixes #19508.

In a fresh TUI session, pressing `Esc` twice entered the rewind transcript overlay even though there was no user message to rewind to. That produced an empty header-only transcript view and exposed a rewind flow that could not select a valid target.

## What changed

The backtrack flow now checks whether a user-message rewind target exists before opening the transcript preview. If no target exists, Codex stays in the main TUI and shows `No previous message to edit.` instead of opening an empty overlay.

The same guard applies when starting rewind preview from the transcript overlay, and the first `Esc` no longer advertises the “edit previous message” hint when there is no previous message available.

Snapshot coverage was added for the unavailable rewind info message, along with a small target-detection test.


vinhnx:
Awesome, thanks for the quick fix @etraut-openai!

fcoury-oai (approved):
Tested it for both cases: no messages and with some messages. 

Code looks good as well. Approved.

PR #19513, opened 2026-04-25
Delay approval prompts while typing

etraut-openai:
## Why

Fixes #7744. Approval modals can currently appear while the user is typing ahead in the TUI composer, which lets plain letters like `y` or `a` get consumed as approval shortcuts instead of staying in the draft input.

## What changed

- Track recent composer typing activity in `bottom_pane/mod.rs`.
- Delay new approval overlays for 1 second while the composer is active, keeping delayed requests queued until the user is idle.
- Preserve the existing active-overlay behavior so approvals that arrive while an approval modal is already open are still queued into that overlay.
- Prune delayed approvals when app-server resolution says the request has already been handled.

## Verification

Added unit coverage for immediate approvals, delayed approvals, idle deadline reset, typed shortcut letters staying in the composer, shortcut handling after the delay, and resolved delayed-request pruning.

Focused `codex-tui` test groups pass locally. The full `cargo test -p codex-tui` run currently aborts in `app::tests::attach_live_thread_for_selection_rejects_unmaterialized_fallback_threads`; that same test also fails when run alone with the same stack overflow.

Manual reviewer check:

1. Start the TUI from the repo root:

   ```bash
   RUST_LOG=trace just codex \
     -c log_dir=<temp-log-dir> \
     --ask-for-approval untrusted \
     --sandbox workspace-write
   ```

2. Submit this prompt:

   ```text
   create a file text.txt on my desktop
   ```

3. While the agent is preparing the approval request, immediately type text such as `ya this should stay in the composer`.
4. Confirm the typed-ahead `y`/`a` remains in the composer instead of approving the request.
5. Stop typing for about 1 second; the approval modal should then appear.
6. Once the modal is visible, press `y` and confirm the approval shortcut works normally.


codex-rs/tui/src/bottom_pane/mod.rs
```diff
@@ -512,7 +568,23 @@ impl BottomPane {
                 self.request_redraw();
                 return InputResult::None;
             }
+            let records_composer_activity =
```
fcoury-oai, line 570: Codex found an edge case here where: on Windows, typing normal characters with AltGr is reported by the terminal as Ctrl+Alt.

This code treats Ctrl or Alt as "not typing", so AltGr characters like <sup>2</sup> do get inserted in the composer but they don't reset the "user is typing" timer. So the prompt can still appear automatically as the user is typing that class of characters.

The fix is to use the existing helper that knows AltGr is real text input:

```rust
  && !key_event
      .modifiers
      .intersects(KeyModifiers::CONTROL | KeyModifiers::ALT)
```

  to:

```rust
  && !key_hint::has_ctrl_or_alt(key_event.modifiers)
```

fcoury-oai (approved):
Just curious: were you able to replicate this problem by hand? I tried by couldn't.

Codex found an edge case that is very simple to fix. I am proactively approving it.

codex-rs/tui/src/bottom_pane/mod.rs
etraut-openai: [codex] Addressed in 8e567d28ee by switching this check to `key_hint::has_ctrl_or_alt(key_event.modifiers)`, so AltGr text input now resets the composer activity timer instead of being treated like a Ctrl/Alt shortcut.

PR #19537, opened 2026-04-25
[mcp] Fix plugin MCP approval policy.

mzeng-openai:
Plugin MCP servers are loaded from plugin manifests rather than top-level `[mcp_servers]`, so their tool approval preferences need to be stored and applied through the owning plugin config. Without this, choosing "Always allow" for a plugin MCP tool could write a preference that was not reliably used on later tool calls.

## Summary
- Add plugin-scoped MCP policy config under `plugins.<plugin>.mcp_servers`, including server enablement, tool allow/deny lists, server defaults, and per-tool approval modes.
- Overlay plugin MCP policy onto manifest-provided server configs when plugins are loaded.
- Route persistent "Always allow" writes for plugin MCP tools back to the owning `plugins.<plugin>.mcp_servers.<server>.tools.<tool>` config entry.
- Reload user config after persisting an approval and make the plugin load cache config-aware so stale plugin MCP policy is not reused after `config.toml` changes.
- Regenerate the config schema and add coverage for plugin MCP policy loading, approval lookup, persistence, and stale-cache prevention.

## Testing
- `cargo test -p codex-core --lib plugin_mcp`


codex-rs/core/src/plugins/manager.rs
```diff
@@ -401,7 +408,9 @@ impl PluginsManager {
             return PluginLoadOutcome::default();
         }
 
-        if !force_reload && let Some(outcome) = self.cached_enabled_outcome() {
+        let config_version = version_for_toml(&config.config_layer_stack.effective_config());
```
mzeng-openai, line 446: Invalidate plugin cache when the actual config changes.

codex-rs/config/src/types.rs
```diff
@@ -653,6 +653,50 @@ pub use crate::skills_config::SkillsConfig;
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub default_tools_approval_mode: Option<AppToolApproval>,
+
+    /// Explicit allow-list of tools exposed from this server.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub enabled_tools: Option<Vec<String>>,
+
+    /// Explicit deny-list of tools. These tools are removed after applying `enabled_tools`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub disabled_tools: Option<Vec<String>>,
```
shijie-oai, line 747: I am not 100% understanding why we need to separate list here.

codex-rs/config/src/types.rs
```diff
@@ -653,6 +653,50 @@ pub use crate::skills_config::SkillsConfig;
...
+
+    /// Per-tool approval settings keyed by tool name.
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub tools: HashMap<String, McpServerToolConfig>,
```
shijie-oai, line 751: Maybe better name for this? `ApprovalModelByTool`?

codex-rs/config/src/types.rs
mzeng-openai: this is following the existing definition in McpServerConfig

codex-rs/config/src/types.rs
mzeng-openai: I would keep it generic so that we can add more fields into later!

xl-openai (approved):

owenlin0 (approved):

shijie-oai (approved):

PR #19595, opened 2026-04-25
[codex] Bypass managed network for escalated exec

viyatb-oai:
## Why

`sandbox_permissions = "require_escalated"` is treated as an explicit request to approve the command and run it outside the filesystem/platform sandbox. Before this change, shell and unified exec still registered managed network approval context and could inject Codex-managed proxy state into the child process, which meant an approved escalated command could still hit a second network approval path.

This PR makes that escalation boundary consistent: once a command is explicitly approved to run outside the sandbox, Codex does not also route that process through the managed network proxy.

## Security impact

Command/filesystem sandbox approval now implies network approval for that command. If an untrusted command or script is allowed to run with `require_escalated`, its network calls are unsandboxed: Codex-managed network allowlists and denylists are not respected for that process, so the command can exfiltrate any data it can read.

## What changed

- Skip managed network approval specs for `SandboxPermissions::RequireEscalated`.
- Pass `network: None` into shell, zsh-fork shell, and unified exec sandbox preparation for explicitly escalated requests.
- Strip Codex-managed proxy environment variables when `CODEX_NETWORK_PROXY_ACTIVE` is present, while preserving user proxy env when the Codex marker is absent.
- Add regression coverage for the prepared exec request so the old behavior cannot silently reappear.

## Verification

- `cargo test -p codex-core explicit_escalation`

### DIFF
--- a/codex-rs/core/src/tools/runtimes/mod.rs
+++ b/codex-rs/core/src/tools/runtimes/mod.rs
@@ -6,6 +6,7 @@ small and focused and reuses the orchestrator for approvals + sandbox + retry.
 */
 use crate::exec_env::CODEX_THREAD_ID_ENV_VAR;
 use crate::path_utils;
+use crate::sandboxing::SandboxPermissions;
 use crate::shell::Shell;
 use crate::tools::sandboxing::ToolError;
 #[cfg(target_os = "macos")]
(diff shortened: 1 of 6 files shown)


dylan-hurd-oai:
@codex review

codex-rs/core/src/tools/runtimes/mod.rs
```diff
@@ -43,6 +44,28 @@ pub(crate) fn build_sandbox_command(
...
+        for key in PROXY_ENV_KEYS {
+            env.remove(*key);
+        }
+        #[cfg(target_os = "macos")]
+        if env
+            .get(PROXY_GIT_SSH_COMMAND_ENV_KEY)
+            .is_some_and(|command| command.starts_with(CODEX_PROXY_GIT_SSH_COMMAND_MARKER))
+        {
+            env.remove(PROXY_GIT_SSH_COMMAND_ENV_KEY);
+        }
```
bolinfest, line 65: Maybe add a comment here, particularly why it's only for Mac?

codex-rs/core/src/tools/runtimes/mod_tests.rs
```diff
@@ -16,6 +34,23 @@ use std::sync::Arc;
 use tempfile::tempdir;
 use tokio::sync::watch;
 
+struct StaticReloader;
+
+#[async_trait::async_trait]
```
bolinfest, line 39: I was going to ask if we can we avoid `#[async_trait::async_trait]` because I'm a bit wary of it since #16630, though I see this is an existing contract, so it's not in scope to change it in this PR.

bolinfest:
I think this is the right change, though I'm a little uncomfortable that these three files require similar changes:

```
codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
codex-rs/core/src/tools/runtimes/shell.rs
codex-rs/core/src/tools/runtimes/unified_exec.rs
```

I don't see a great way to consolidate them, but it admittedly takes effort to ensure we are applying this sort of logic consistently across our various shell tools.

bolinfest (approved):

PR #19597, opened 2026-04-25
Fix TUI attach fallback test stack overflow

### LINKED ISSUE #19596: Windows: codex-tui app test stack overflows on clean HEAD
## Summary

On Windows, `codex-tui` can abort with `STATUS_STACK_OVERFLOW` when running the `app::tests::attach_live_thread_for_selection_rejects_unmaterialized_fallback_threads` test. This reproduces from a clean detached worktree at upstream `HEAD`, so it is not caused by unrelated local TUI skill-popup changes.

## Environment

- OS: Windows
- Repo: `openai/codex`
- Clean HEAD tested: `6de6eaa0c` (`[4/4] Honor Streamable HTTP MCP placement (#18584)`)
- Test path used: short path under `C:\Users\stefa\codex-clean-head-proof-*` to avoid Windows path-length failures from deeper worktree paths

## Reproduction

From a clean detached worktree at `HEAD`:

```powershell
cd C:\Users\stefa\codex-clean-head-proof-8868\codex-rs
cargo test -p codex-tui app::tests::attach_live_thread_for_selection_rejects_unmaterialized_fallback_threads -- --test-threads=1
```

## Actual Result

```text
running 1 test
test app::tests::attach_live_thread_for_selection_rejects_unmaterialized_fallback_threads ...

thread 'app::tests::attach_live_thread_for_selection_rejects_unmaterialized_fallback_threads' has overflowed its stack
error: test failed, to rerun pass `-p codex-tui --lib`

Caused by:
  process didn't exit successfully: ... (exit code: 0xc00000fd, STATUS_STACK_OVERFLOW)
```

A full local run of `cargo test -p codex-tui` also aborts with the same class of stack overflow before the suite can complete.

## Expected Result

The test should pass or fail with a normal assertion. It should not abort the process with a stack overflow.

## Notes

This was separated from a focused TUI skill mention popup change because the failure reproduces on clean `HEAD`. Keeping this as its own issue should make the root cause easier to triage without mixing it into unrelated UI behavior changes.

etraut-openai:
## Why

Fixes #19596.

The `codex-tui` fallback tests for `attach_live_thread_for_selection` exercise the embedded app-server path, which can exceed the default test-thread stack in debug builds. When that happens, the process aborts with a stack overflow before the tests can report a normal assertion result. The issue was reported on Windows, and the same failure reproduced locally without increasing `RUST_MIN_STACK`.

## What changed

The two affected fallback tests now run through a small test helper that creates an explicit 4 MiB thread stack and drives the async test body with a current-thread Tokio runtime. The test behavior is unchanged: both cases still assert that empty or unmaterialized fallback threads are rejected without creating a local thread event channel.

## Verification

- `cargo test -p codex-tui attach_live_thread_for_selection_rejects_ -- --test-threads=1`
- `cargo test -p codex-tui` no longer aborts with the reported stack overflow; it progressed past these tests and later failed on unrelated sandbox-sensitive status snapshot/config expectations.

### DIFF
--- a/codex-rs/tui/src/app/tests.rs
+++ b/codex-rs/tui/src/app/tests.rs
@@ -1349,65 +1349,100 @@ async fn open_agent_picker_marks_loaded_threads_open() -> Result<()> {
     Ok(())
 }
 
-#[tokio::test]
-async fn attach_live_thread_for_selection_rejects_empty_non_ephemeral_fallback_threads()
--> Result<()> {
-    let mut app = make_test_app().await;
-    let mut app_server = crate::start_embedded_app_server_for_picker(app.chat_widget.config_ref())
-        .await
-        .expect("embedded app server");
-    let started = app_server
-        .start_thread(app.chat_widget.config_ref())
-        .await?;
-    let thread_id = started.session.thread_id;
-    app.agent_navigation.upsert(
-        thread_id,
-        Some("Scout".to_string()),
-        Some("worker".to_string()),
-        /*is_closed*/ false,
-    );
+fn run_stack_heavy_app_server_test<F, Fut>(test_name: &'static str, test_body: F) -> Result<()>
+where
+    F: FnOnce() -> Fut + Send + 'static,
+    Fut: std::future::Future<Output = Result<()>>,
+{
+    // The embedded app-server path can exceed the default test-thread stack in debug builds.
+    const TEST_STACK_SIZE_BYTES: usize = 4 * 1024 * 1024;
 
-    let err = app
-        .attach_live_thread_for_selection(&mut app_server, thread_id)
-        .await
-        .expect_err("empty fallback should not attach as a blank replay-only thread");
+    let handle = std::thread::Builder::new()
+        .name(test_name.to_string())
+        .stack_size(TEST_STACK_SIZE_BYTES)
+        .spawn(move || -> Result<()> {
+            let runtime = tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()?;
+            runtime.block_on(test_body())
+        })?;
 
-    assert_eq!(
-        err.to_string(),
-        format!("Agent thread {thread_id} is not yet available for replay or live attach.")
-    );
-    assert!(!app.thread_event_channels.contains_key(&thread_id));
-    Ok(())
+    match handle.join() {
+        Ok(result) => result,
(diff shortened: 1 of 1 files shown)


BaseInfinity:
Thanks for jumping on this! I tested the same class of fix locally and wanted to share one extra data point from Windows.

The per-test large-stack wrapper fixes the originally reported attach fallback tests, but on my machine `cargo test -p codex-tui` then continued far enough to hit the same `STATUS_STACK_OVERFLOW` class in other debug-heavy app tests, including:

- `app::tests::discard_side_thread_removes_agent_navigation_entry`
- `app::tests::interrupt_without_active_turn_is_treated_as_handled`
- `app::tests::open_agent_picker_marks_loaded_threads_open`

A broader approach that worked locally was setting `RUST_MIN_STACK=16777216` for Windows `codex-tui` tests from `tui/build.rs`:

```rust
fn main() {
    if std::env::var_os("CARGO_CFG_WINDOWS").is_some() {
        // Match the CLI runtime's worker stack so debug-heavy app tests do
        // not overflow Windows libtest threads before assertions run.
        println!("cargo:rustc-env=RUST_MIN_STACK=16777216");
    }
}
```

Local verification:

- RED on clean `HEAD`: `cargo test -p codex-tui app::tests::attach_live_thread_for_selection_rejects_unmaterialized_fallback_threads -- --test-threads=1` aborted with `STATUS_STACK_OVERFLOW`.
- With the `RUST_MIN_STACK` build-script env, the originally failing exact test passed.
- An unwrapped test that later overflowed in full-suite runs, `app::tests::open_agent_picker_marks_loaded_threads_open`, also passed with the build-script env.
- `cargo test -p codex-tui` no longer aborted with stack overflow locally; it progressed to normal assertion failures in unrelated `status::helpers::*compose_agents_summary*` path-format expectations on my Windows environment.

I don't want to derail this PR if the narrower wrapper is preferred, but if the intent is to unblock local Windows `codex-tui` runs generally, the crate-level test stack setting may avoid needing to wrap more app tests as they surface.

Thanks again for responding and getting back so quickly, I've been ramping up my codex cli usage on Windows and love to give back to such a great tool! Thanks for making it open source!

etraut-openai:
OK, then this problem is more involved. I'm going to abandon this approach. This will need to take a back seat since it's not affecting our CI or other users.

BaseInfinity:
Thanks for looking into this! I'll let you know from my end if my local fix was enough

PR #19610, opened 2026-04-25
Support end_turn in response.completed

andmis:
Some providers of Responses API forward a model-defined `end_turn` boolean indicating explicitly the model's indication of whether it would like to end the turn or to be inferenced again. In this PR, we update the sampling loop to use this field correctly if it's set. If the field is not set by the provider, we fall back to the existing sampling logic.

### DIFF
--- a/codex-rs/core/src/session/turn.rs
+++ b/codex-rs/core/src/session/turn.rs
@@ -2144,8 +2144,8 @@ async fn try_run_sampling_request(
                 sess.update_token_usage_info(&turn_context, token_usage.as_ref())
                     .await;
                 should_emit_turn_diff = true;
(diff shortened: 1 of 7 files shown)


codex-rs/core/src/session/turn.rs
```diff
@@ -2143,7 +2144,9 @@ async fn try_run_sampling_request(
                 sess.update_token_usage_info(&turn_context, token_usage.as_ref())
                     .await;
                 should_emit_turn_diff = true;
-
+                if let Some(end_turn) = end_turn {
```
tibo-openai: Suggest

if let Some(false) = end_turn {
    needs_follow_up = true;
}

tibo-openai (approved):

PR #19616, opened 2026-04-26
feat(tui): add codex logo to session header

fcoury-oai:
## Summary
- Render the braille Codex logo to the left of the startup session header on wide terminals.
- Keep the existing text-only header as the fallback for narrower terminal widths.
- Pick the logo gradient from the terminal background so it remains legible on light and dark themes.

image
image

## Tests
- Not run in this turn; PR was opened from the existing branch diff.


fcoury-oai:
Replaced by #19617.

PR #19618, opened 2026-04-26
Persist shell mode commands in prompt history

### LINKED ISSUE #19613: Bash mode history
### What variant of Codex are you using?

CLI

### What feature would you like to see?

Once in bash mode, I would like to be able to press up on by keyboard to get recent commands.

I would like the history of previous commands, to persist even after starting a new session `/new`

P.S. Maybe should should rename it `shell mode` for windows powershell users.

### Additional information

_No response_

etraut-openai:
## Why

`!` shell commands are currently surfaced as "Bash mode", which is misleading for users running shells such as PowerShell or zsh. Those commands also bypass the persistent prompt history path, so they cannot be recalled after starting a new session.

Fixes #19613.

## What changed

- Rename the TUI footer label and related test wording from "Bash mode" to "Shell mode".
- Persist accepted `!` shell commands to prompt history with the leading `!`, so recall restores the composer into shell mode across sessions.
- Add coverage for immediate and queued shell-command submissions emitting the prompt-history update.

## Verification

- `cargo test -p codex-tui bang_shell`
- `cargo test -p codex-tui shell_command_uses_shell_accent_style`
- `cargo test -p codex-tui footer_mode_snapshots`
- `cargo insta pending-snapshots --manifest-path tui/Cargo.toml`

Manually verified fix after confirming presence of bug prior to fix.

### DIFF
--- a/codex-rs/tui/src/bottom_pane/chat_composer.rs
+++ b/codex-rs/tui/src/bottom_pane/chat_composer.rs
@@ -1160,7 +1160,7 @@ impl ChatComposer {
 
     /// Convert canonical composer text into the textarea's internal representation.
     ///
-    /// Bash mode stores the leading `!` as prompt state instead of editable text,
+    /// Shell mode stores the leading `!` as prompt state instead of editable text,
     /// so full-buffer imports must absorb that prefix before rebuilding the textarea.
     fn imported_text_for_textarea(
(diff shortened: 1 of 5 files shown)


fcoury-oai (approved):
Tested before and after and codex confirmed no findings.

Issue #19581, opened 2026-04-25
WebSocket fails with non-ASCII workspace paths (UTF-8 in x-codex-turn-metadata header)


Kamicyus:
# Codex CLI UTF-8 Path Bug — Bug Report Draft

**Bug version**: codex-cli 0.125.0-alpha.3
**Date discovered**: 2026-04-25
**Reproducer**: Kamer (kamercanizvermez@gmail.com)

## Summary

Codex CLI fails to establish WebSocket connection when the workspace path contains non-ASCII Unicode characters (specifically tested with Turkish characters: `ı`, `ğ`, `ş`). The HTTP header `x-codex-turn-metadata` requires ASCII-only values per RFC 7230, but the workspace path is included unencoded in the `workspaces` field of the JSON metadata.

## Reproduction

```bash
# Setup: directory with Turkish characters
mkdir ~/Agentlarım && cd ~/Agentlarım
git init && git commit --allow-empty -m "test"

# Trigger
echo "hello" | codex exec -

# Expected: response from Codex
# Actual: 5x reconnect attempts, fallback succeeds but slow
```

## Error Output

```
ERROR codex_api::endpoint::responses_websocket: failed to connect to websocket:
UTF-8 encoding error: failed to convert header to a str for header name
'x-codex-turn-metadata' with value: "{\"session_id\":\"...\",\"thread_source\":\"user\",
\"turn_id\":\"...\",\"workspaces\":{\"/Users/kamer/Agentlar\xc4\xb1m\":{...}},
\"sandbox\":\"seatbelt\"}"
url: wss://chatgpt.com/backend-api/codex/responses
```

The `\xc4\xb1` is the UTF-8 byte sequence for `ı` (U+0131 LATIN SMALL LETTER DOTLESS I).

## Workaround

Run Codex outside the affected workspace:

```bash
cd /tmp && codex exec --skip-git-repo-check
```

This avoids triggering git workspace discovery which embeds the path in the WebSocket header.

## Suggested Fix

In the `x-codex-turn-metadata` header serialization code, percent-encode the workspace path values per RFC 3986, or use a separate header that supports UTF-8 (e.g., `x-codex-turn-metadata-b64` with base64-encoded JSON).

Alternative: use a body field or HTTP/2 trailer instead of a header for binary-safe metadata.

## Impact

- Affects all users with non-ASCII characters in workspace paths
- Common in non-English locales (Turkish, German, French, Japanese, Chinese, etc.)
[...]

PR #19620, opened 2026-04-26
Escape turn metadata headers as ASCII JSON

### LINKED ISSUE #17468: Official codex-cli 0.120.0 sends raw multibyte JSON in x-codex-turn-metadata, breaking HTTP interoperability

codex-cli 0.120.0

ChatGPT Plus

gpt-5.1-codex-mini

Microsoft Windows NT 10.0.26200.0 x64

PowerShell 7.6.0

  x-codex-turn-metadata is emitted by the official CLI as raw JSON rather than an ASCII-safe encoding. When that metadata
  contains non-ASCII characters, this breaks HTTP interoperability.

  I gathered three independent repros:

  1. Standard client stack repro:
     Python urllib.request fails before send with:
     UnicodeEncodeError: 'latin-1' codec can't encode characters...
     when the same logical turn metadata JSON is placed directly in x-codex-turn-metadata.
  2. Hosted API repro:
     Node fetch against Azure OpenAI fails before send for the raw header with:
     TypeError: Cannot convert argument to a ByteString because the character at index 31 has a value of 26481 which is
     greater than 255.
     The same logical metadata succeeds when Base64-encoded. Against:
     GET https://example-resource.openai.azure.com/openai/v1/models
     the Base64 version returned 200 OK.
  3. Official CLI repro:
     I ran the official npm-installed codex-cli 0.120.0 from a Git repo whose absolute path contains Japanese characters:
```text
C:\Users\<redacted>\AppData\Local\Temp\codex-turn-meta-repro-東京\repo
```
and pointed it at a local mock /v1/responses endpoint.

The mock server captured this request header:
```text
{
  "path": "/v1/responses",
  "turn_metadata": "{\"session_id\":\"019d...\",\"turn_id\":\"019d...\",\"workspaces\":{\"C:\\\\Users\\\\<redacted>\\\\AppData\
\\\Local\\\\Temp\\\\codex-turn-meta-repro-\u00e6\u009d\u00b1\u00e4\u00ba\u00ac\\\\repo\":{\"latest_git_commit_hash\":
\"ab1e986b4d8d86bdf5b69f6c614e8840789b6457\",\"has_changes\":false}},\"sandbox\":\"none\"}"
}
```

  That value is not Base64. It is raw JSON placed directly in the header, and the Japanese path appears as mojibake bytes
  in the captured header value instead of being transport-safe.

  I also compared the official CLI with a locally built binary from my fork of Codex:

  - repo: https://github.com/amabile4/codex
  - branch: azure/release-0.120.0

  That fork only changes the wire encoding of x-codex-turn-metadata from raw JSON to Base64. It does not change the
  logical metadata payload.

  Under the same machine / same repo / same prompt / same local mock server:

  - the official codex-cli 0.120.0 sends raw JSON in x-codex-turn-metadata
  - the fork build sends Base64 in x-codex-turn-metadata

  The mock server decodes the fork header back to the same logical JSON, including the original Japanese workspace path.
  So the fork is not changing the meaning of the metadata, only making the transport encoding safe.

  1. Confirm the official CLI version:
```powershell
codex --version
```
  2. Create a Git repo whose absolute path contains non-ASCII characters, for example Japanese:
```powershell
$root = Join-Path $env:TEMP 'codex-turn-meta-repro-東京'
$repo = Join-Path $root 'repo'
New-Item -ItemType Directory -Force -Path $repo | Out-Null
Set-Location $repo
git init
Set-Content README.md 'repro'
git add README.md
git -c user.name='repro' -c user.email='repro@example.com' commit -m 'init'
```
  3. Start a local mock HTTP server that logs request headers for /v1/responses.
  4. Run the official CLI against that local mock provider:
```powershell
codex exec --json `
  -C $repo `
  -c model_provider='mock' `
  -c 'model_providers.mock.name="Mock"' `
  -c 'model_providers.mock.base_url="http://127.0.0.1:8012/v1"' `
  -c 'model_providers.mock.wire_api="responses"' `
  -m 'gpt-5.1-codex-mini' `
  'Reply with exactly OK. Do not use any tools.'
```
  5. Inspect the mock server log. The official CLI sends raw JSON in x-codex-turn-metadata, including the non-ASCII
     workspace path.
  6. As an interoperability check, send the same logical metadata using a minimal repro client:
      - raw JSON in x-codex-turn-metadata
      - Base64-encoded JSON in x-codex-turn-metadata
  7. Observe:
      - raw fails in standard client stacks
      - Base64 succeeds
  8. Optional A/B confirmation:
     run a locally built binary from amabile4/codex branch azure/release-0.120.0, where the only relevant change is Base64
     encoding of x-codex-turn-metadata. Under the same repo and same mock server, the fork build sends Base64 while the
     official CLI sends raw JSON.

  x-codex-turn-metadata should be encoded in an ASCII-safe form before it is placed on the wire.

  A Base64-encoded JSON blob is sufficient. The official CLI should not emit raw multibyte JSON header values that can:

  - fail in standard HTTP clients before the request is sent
  - break reconnect or request handling on some platforms
  - fail against providers or gateways that enforce ASCII-safe header handling

### Additional information

  - Confirmed examples currently gathered:
      - Python urllib.request: raw fails, Base64 succeeds
      - Node fetch -> Azure OpenAI GET /openai/v1/models: raw fails before send, Base64 succeeds with 200 OK
      - Official codex-cli 0.120.0 -> local mock /v1/responses: raw JSON is actually emitted in x-codex-turn-metadata
      - Local fork build from amabile4/codex branch azure/release-0.120.0 -> local mock /v1/responses: Base64 is emitted
        instead and decodes cleanly back to the same JSON
  - This is not only an Azure-specific complaint. Azure is one impacted environment, but the underlying bug is that the
    header value is not transport-safe across multiple HTTP stacks.

etraut-openai:
## Why

`x-codex-turn-metadata` is sent as an HTTP/WebSocket header, but Codex was serializing the metadata JSON with raw UTF-8 string contents. When a workspace path contains non-ASCII characters, common HTTP stacks can reject or corrupt that header before the request reaches the provider.

Fixes #17468. Also addresses the duplicate WebSocket report in #19581.

## What changed

- Added `codex_utils_string::to_ascii_json_string`, a shared helper that serializes JSON normally while escaping non-ASCII string content as `\uXXXX`.
- Switched turn metadata header serialization, including merged Responses API client metadata, to use the ASCII-safe JSON helper.
- Added coverage for non-ASCII workspace paths and non-ASCII client metadata while preserving the same parsed JSON values.

## Verification

- `cargo test -p codex-utils-string`
- `cargo test -p codex-core turn_metadata`
- `just bazel-lock-check`

### DIFF
--- a/codex-rs/utils/string/src/json.rs
+++ b/codex-rs/utils/string/src/json.rs
@@ -21,7 +21,7 @@ impl serde_json::ser::Formatter for AsciiJsonFormatter {
             }
 
             if start < index {
-                writer.write_all(fragment[start..index].as_bytes())?;
+                writer.write_all(&fragment.as_bytes()[start..index])?;
             }
 
             let mut utf16 = [0; 2];
@@ -32,7 +32,7 @@ impl serde_json::ser::Formatter for AsciiJsonFormatter {
         }
 
         if start < fragment.len() {
-            writer.write_all(fragment[start..].as_bytes())?;
+            writer.write_all(&fragment.as_bytes()[start..])?;
         }
 
         Ok(())
--- a/codex-rs/utils/string/src/json.rs
+++ b/codex-rs/utils/string/src/json.rs
@@ -1,0 +1,85 @@
+//! JSON serialization helpers for output that must remain parseable as JSON
+//! while staying safe for ASCII-only transports.
+
+use std::io;
+
+use serde::Serialize;
+
+struct AsciiJsonFormatter;
+
+impl serde_json::ser::Formatter for AsciiJsonFormatter {
+    // serde_json has no ensure_ascii flag; this formatter keeps its serializer
+    // in charge and only escapes non-ASCII string fragments.
+    fn write_string_fragment<W>(&mut self, writer: &mut W, fragment: &str) -> io::Result<()>
+    where
+        W: ?Sized + io::Write,
+    {
+        let mut start = 0;
+        for (index, ch) in fragment.char_indices() {
+            if ch.is_ascii() {
+                continue;
+            }
+
+            if start < index {
+                writer.write_all(fragment[start..index].as_bytes())?;
+            }
+
+            let mut utf16 = [0; 2];
+            for code_unit in ch.encode_utf16(&mut utf16) {
+                write!(writer, "\\u{code_unit:04x}")?;
+            }
+            start = index + ch.len_utf8();
+        }
+
+        if start < fragment.len() {
+            writer.write_all(fragment[start..].as_bytes())?;
+        }
+
+        Ok(())
+    }
+}
+
+/// Serialize JSON while escaping non-ASCII string content as `\uXXXX`.
+///
+/// This is useful when JSON needs to remain parseable as JSON but must be
+/// carried through ASCII-safe transports such as HTTP headers.
+pub fn to_ascii_json_string<T>(value: &T) -> serde_json::Result<String>
+where
+    T: Serialize + ?Sized,
+{
+    let mut bytes = Vec::new();
+    let mut serializer = serde_json::Serializer::with_formatter(&mut bytes, AsciiJsonFormatter);
+    value.serialize(&mut serializer)?;
+    String::from_utf8(bytes)
+        .map_err(|err| serde_json::Error::io(io::Error::new(io::ErrorKind::InvalidData, err)))
+}
+
+#[cfg(test)]
+mod tests {
+    use pretty_assertions::assert_eq;
+    use serde_json::Value;
+    use serde_json::json;
+
+    use super::to_ascii_json_string;
+
+    #[test]
+    fn to_ascii_json_string_escapes_non_ascii_strings() {
+        let value = json!({
+            "workspaces": {
+                "/tmp/東京": {
+                    "label": "Agentlarım",
+                    "emoji": "🚀"
+                }
+            }
+        });
+
+        let serialized = to_ascii_json_string(&value).expect("serialize ascii json");
+
+        assert!(serialized.is_ascii());
+        assert!(!serialized.contains("東京"));
+        assert!(!serialized.contains("Agentlarım"));
+        assert!(!serialized.contains("🚀"));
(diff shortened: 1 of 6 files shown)


codex-rs/utils/string/src/json.rs
```diff
@@ -0,0 +1,85 @@
...
+                    "label": "Agentlarım",
+                    "emoji": "🚀"
+                }
+            }
+        });
+
+        let serialized = to_ascii_json_string(&value).expect("serialize ascii json");
+
+        assert!(serialized.is_ascii());
+        assert!(!serialized.contains("東京"));
```
pakrym-oai, line 116: can you assert the exact expected string?

codex-rs/utils/string/src/json.rs
```diff
@@ -0,0 +1,85 @@
...
+use std::io;
+
+use serde::Serialize;
+
+struct AsciiJsonFormatter;
+
+impl serde_json::ser::Formatter for AsciiJsonFormatter {
+    // serde_json has no ensure_ascii flag; this formatter keeps its serializer
+    // in charge and only escapes non-ASCII string fragments.
+    fn write_string_fragment<W>(&mut self, writer: &mut W, fragment: &str) -> io::Result<()>
```
pakrym-oai, line 13: Have you considered percent encoding?

pakrym-oai (approved):
Percent encoding might be a more HTTP-native solution for this.

codex-rs/utils/string/src/json.rs
etraut-openai: Here's Codex's justification for not using percent encoding. I think it's a reasonable point, so I'm going to keep the current implementation.

I’d push back gently on percent encoding. It’s HTTP-native for opaque header values, but this header’s payload is JSON. Encoding the whole JSON blob would require receiver-side percent-decoding before JSON parse; encoding only string contents would change the JSON values unless every consumer knew to decode them. JSON Unicode escapes are standard JSON, ASCII-safe, and preserve the existing wire contract for JSON parsers.

PR #19624, opened 2026-04-26
fix(tui): make skill context budget warning actionable

fcoury-oai:
## Summary

I was getting this message often but I had no clue on what to do other than asking Codex how to fix it. I believe other users will have the same annoyance. This PR tries to make the message actionable, so the underlying trigger can be fixed.

image

This PR updates the skills metadata budget warning so it explains what changed and what remains available. Description-only truncation now says all skills are still available and lists the largest truncated descriptions. Cases that omit skills now list the omitted skill names, capped with an `and N more` suffix.

## Why

The previous warning reported only aggregate truncation details, which made it hard for users to understand whether skills were unavailable or what they could inspect next.

## Validation

- Manual testing of the before and after as can be seen above.
- `just fmt`
- `cargo test -p codex-core-skills`
- `cargo test -p codex-core skill_metrics_records`
- `cargo test -p codex-core build_initial_context_emits_thread_start_skill_warning_on_repeated_builds`
- `cargo test -p codex-tui live_app_server_warning_notification_renders_message`
- `cargo test -p codex-app-server turn_start_emits_thread_scoped_warning_notification_for_trimmed_skills`
- `git diff --check`

### DIFF
--- a/codex-rs/core-skills/src/render.rs
+++ b/codex-rs/core-skills/src/render.rs
@@ -17,10 +17,8 @@ use codex_utils_output_truncation::approx_token_count;
 const DEFAULT_SKILL_METADATA_CHAR_BUDGET: usize = 8_000;
 const SKILL_METADATA_CONTEXT_WINDOW_PERCENT: usize = 2;
 const SKILL_DESCRIPTION_TRUNCATION_WARNING_THRESHOLD_CHARS: usize = 10;
+const MAX_WARNING_SKILL_NAMES: usize = 5;
 const APPROX_BYTES_PER_TOKEN: usize = 4;
-pub const SKILL_DESCRIPTION_TRUNCATED_WARNING_PREFIX: &str = "Warning: Exceeded skills context budget. Loaded skill descriptions were truncated by an average of";
-pub const SKILL_DESCRIPTIONS_REMOVED_WARNING_PREFIX: &str =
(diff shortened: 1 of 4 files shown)


etraut-openai (approved):
LGTM. Let's run it past @xl-openai to make sure he's good with the new wording.

codex-rs/core-skills/src/render.rs
```diff
@@ -1066,7 +1111,7 @@ mod tests {
         assert_eq!(
             rendered.warning_message,
             Some(
-                "Warning: Exceeded skills context budget. Loaded skill descriptions were truncated by an average of 14 characters per skill."
```
xl-openai, line 1069: I think the number of char being truncated is an important information.

codex-rs/core-skills/src/render.rs
```diff
@@ -1116,7 +1161,7 @@ mod tests {
         assert_eq!(
             rendered.warning_message,
             Some(
-                "Warning: Exceeded skills context budget. All skill descriptions were removed and 2 additional skills were not included in the model-visible skills list."
+                "Skills metadata exceeded the context budget. Some skills were omitted from the model-visible skills list. Omitted skills: repo-skill, user-skill."
```
xl-openai, line 1164: I have some concern for this. The goal is not to ask user to disable/fix those omitted skills, they are just unluckily being in the last part of the skill list.

PR #19625, opened 2026-04-26
Reset TUI keyboard reporting on exit

### LINKED ISSUE #19553: iTerm2: Ctrl+C exit can intermittently leave terminal stuck in CSI-u mode and print `9;5:3u`

`codex-cli 0.125.0`

_No response_

_No response_

macOS 26.4.1 / Darwin 25.4.0 arm64

iTerm2 `3.6.10`

Relevant environment:

```text
SHELL=/bin/zsh
TERM=xterm-256color
TERM_PROGRAM=iTerm.app
TERM_PROGRAM_VERSION=3.6.10
```

When running Codex CLI in iTerm2 on macOS, pressing `Ctrl+C` to exit Codex can intermittently leave the parent terminal input state corrupted. I have not found a stable reproduction pattern yet, and it does not happen on every `Ctrl+C` exit. However, the failure has occurred after exiting Codex with `Ctrl+C`.

After Codex exits, the shell can receive or display a raw CSI-u style keyboard sequence such as:

```text
9;5:3u
```

This looks like Codex enables enhanced keyboard reporting / CSI-u / modifyOtherKeys style input, but does not always fully restore the terminal state on the `Ctrl+C` exit path. After Codex exits, the parent shell should be back in normal text input mode instead of still emitting raw enhanced-keyboard fragments.

This appears related in shape to #16517, but the environment and trigger are different:

- terminal emulator: iTerm2, not Ghostty
- trigger observed so far: local `Ctrl+C` exit from Codex, not an abnormal SSH disconnect
- reproducibility: intermittent; I do not yet know the exact timing or state needed to trigger it
- observed raw fragment: `9;5:3u`

I do not currently have a fully deterministic reproduction sequence. The observed trigger path is:

1. Open iTerm2 on macOS.
2. Start `codex`.
3. While the Codex TUI is active, press `Ctrl+C` to exit.
4. Return to the same iTerm2 tab/window's shell.
5. Continue typing or pressing keys.

Observed intermittent result:

the shell can receive or display raw CSI-u style text such as:

```text
9;5:3u
```

Codex should restore the terminal keyboard/input state on `Ctrl+C` / SIGINT exit paths, including disabling or popping any enhanced keyboard / CSI-u / modifyOtherKeys modes it enabled.

After Codex exits, the parent shell should immediately behave normally, without requiring the user to run `reset`, `stty sane`, or manual terminal control sequences.

### Additional information

A possible manual recovery command is:

```bash
printf '\033[<u\033[>4;0m' && stty sane && reset
```

The fact that this kind of reset helps suggests the issue may be related to keyboard enhancement flags / CSI-u / modifyOtherKeys state not being reliably restored.

Suggested area to inspect: the Codex TUI cleanup path for SIGINT, early return, panic, task cancellation, and any other abnormal or semi-abnormal exits. In particular, please check whether keyboard enhancement flags are always restored symmetrically with the setup path.

etraut-openai:
## Why

Codex enables enhanced keyboard reporting while the TUI owns the terminal. In iTerm2, exiting the TUI with Ctrl+C can intermittently leave the parent shell receiving raw CSI-u / `modifyOtherKeys` fragments instead of normal key input.

Final terminal cleanup should put the parent shell back into normal keyboard reporting even if the terminal misses the usual stack pop.

Fixes #19553.

## What Changed

- Move TUI keyboard enhancement setup and detection into `tui/src/tui/keyboard_modes.rs`.
- Add an exit-only `restore_after_exit()` path that performs the normal keyboard enhancement pop plus unconditional keyboard enhancement and `modifyOtherKeys` resets.
- Keep temporary restore paths, such as external-editor handoff, using the balanced stack pop behavior.

## Confidence

Medium. This is a speculative fix: I was not able to reproduce the reported iTerm2 behavior manually, but the symptoms line up with terminal keyboard reporting state surviving Codex exit. The added reset sequences are scoped to final TUI shutdown and should be harmless when the terminal is already clean.

### DIFF
--- a/codex-rs/tui/src/lib.rs
+++ b/codex-rs/tui/src/lib.rs
@@ -1576,7 +1576,7 @@ pub(crate) async fn resolve_cwd_for_resume_or_fork(
     reason = "TUI should no longer be displayed, so we can write to stderr."
 )]
 fn restore() {
-    if let Err(err) = tui::restore() {
+    if let Err(err) = tui::restore_after_exit() {
         eprintln!(
             "failed to restore terminal. Run `reset` or restart your terminal to recover: {err}"
         );
@@ -1595,7 +1595,7 @@ impl TerminalRestoreGuard {
     #[cfg_attr(debug_assertions, allow(dead_code))]
     fn restore(&mut self) -> color_eyre::Result<()> {
         if self.active {
-            crate::tui::restore()?;
+            crate::tui::restore_after_exit()?;
             self.active = false;
         }
         Ok(())
--- a/codex-rs/tui/src/tui.rs
+++ b/codex-rs/tui/src/tui.rs
@@ -19,9 +19,6 @@ use crossterm::event::DisableFocusChange;
 use crossterm::event::EnableBracketedPaste;
 use crossterm::event::EnableFocusChange;
 use crossterm::event::KeyEvent;
-use crossterm::event::KeyboardEnhancementFlags;
-use crossterm::event::PopKeyboardEnhancementFlags;
-use crossterm::event::PushKeyboardEnhancementFlags;
 use crossterm::terminal::EnterAlternateScreen;
 use crossterm::terminal::LeaveAlternateScreen;
 use crossterm::terminal::supports_keyboard_enhancement;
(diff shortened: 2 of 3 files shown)


fcoury-oai (approved):
Ran some sanity checks in iTerm2 on macOS:

- Repeated `Ctrl+C` exits from the idle composer
- `Ctrl+C` exit while Codex was active
- External-program handoff and return
- `Ctrl+Z` / `fg` suspend-resume before exiting
- After each run, kept using the same shell session and verified normal typing, arrows, tab completion, and control keys still behaved normally

I am comfortable with approving this as a speculative fix.

PR #19631, opened 2026-04-26
Color TUI statusline from active theme

etraut-openai:
## Why

Users have shared that the TUI can feel too visually flat because themes mostly show up in code syntax highlighting. The configurable statusline is a natural place to make the active theme more visible, while still letting users keep the existing monotone statusline if they prefer it.

## What Changed

- Added a statusline styling helper that builds the rendered statusline from `(StatusLineItem, text)` segments, preserving item identity while keeping the plain text output unchanged.
- Derived foreground accent colors from the active syntax theme by looking up TextMate scopes through the existing syntax highlighter, with conservative ANSI fallbacks when a scope does not provide a foreground.
- Tuned theme-derived colors to keep the accents visible without making the statusline feel overly bright.
- Added `[tui].status_line_use_colors`, defaulting to `true`, plus a separated `/statusline` toggle so users can enable or disable theme-derived statusline colors from the setup UI.
- Updated the live statusline and `/statusline` preview to use the same styled builder, while keeping terminal-title preview text plain.
- Kept statusline separators and active-agent add-ons subdued while removing blanket dimming from the whole passive statusline.

## Verification

- `cargo test -p codex-tui status_line`
- `cargo test -p codex-tui theme_picker`
- `cargo test -p codex-tui foreground_style_for_scopes`
- `cargo test -p codex-tui`
- `cargo test -p codex-core status_line_use_colors`
- `cargo insta pending-snapshots --manifest-path tui/Cargo.toml`

## Visual

Screenshot 2026-04-30 at 6 16 08 PM

Screenshot 2026-04-30 at 6 16 02 PM


canvrno-oai (approved):
Tested and reviewed, this is a nice addition.

PR #19705, opened 2026-04-26
Discover hooks bundled with plugins

abhinav-oai:
## Why

Plugins can bundle lifecycle hooks, but Codex previously only discovered hooks from user, project, and managed config layers. This adds the plugin discovery and runtime plumbing needed for plugin-bundled hooks while keeping execution behind the `plugin_hooks` feature flag.

## What

- Discovers plugin hook sources from each plugin's default `hooks/hooks.json`.
- Supports `plugin.json` manifest `hooks` entries as either relative paths or inline hook objects.
- Plumbs discovered plugin hook sources through plugin loading into the hook runtime when `plugin_hooks` is enabled.
- Marks plugin-originated hook runs as `HookSource::Plugin`.
- Injects `PLUGIN_ROOT` and `CLAUDE_PLUGIN_ROOT` into plugin hook command environments.
- Updates generated schemas and hook source metadata for the plugin hook source.

## Stack

1. This PR - openai/codex#19705
2. openai/codex#19778
3. openai/codex#19840
4. openai/codex#19882

## Reviewer Notes

- Core logic is in `codex-rs/core-plugins/src/loader.rs` and `codex-rs/hooks/src/engine/discovery.rs`
- Moved existing / adding new tests to `codex-rs/core-plugins/src/loader_tests.rs` hence the large diff there
- Otherwise mostly plumbing and minor schema updates

### Core Changes

The `codex-rs/core` changes are limited to wiring plugin hook support into existing core flows:

- `core/src/session/session.rs` conditionally pulls effective plugin hook sources and plugin hook load warnings from `PluginsManager` when `plugin_hooks` is enabled, then passes them into `HooksConfig`.
- `core/src/hook_runtime.rs` adds the `plugin` metric tag for `HookSource::Plugin`.
- `core/config.schema.json` picks up the new `plugin_hooks` feature flag, and `core/src/plugins/manager_tests.rs` updates fixtures for the added plugin hook fields.


codex-rs/hooks/src/engine/discovery.rs
```diff
@@ -125,11 +147,45 @@ fn append_managed_requirement_handlers(
...
+    for source in plugin_hook_sources {
+        let PluginHookSource {
+            plugin_root,
+            source_path,
+            hooks,
+            ..
+        } = source;
+        let mut env = HashMap::new();
+        let plugin_root_value = plugin_root.display().to_string();
+        env.insert("AGENTS_PLUGIN_ROOT".to_string(), plugin_root_value.clone());
```
abhinav-oai: can we just make this one `PLUGIN_ROOT` actually? be sure to update all other references/usage

codex-rs/hooks/src/engine/discovery.rs
```diff
@@ -125,11 +146,46 @@ fn append_managed_requirement_handlers(
...
     );
 }
 
+fn append_plugin_hook_sources(
+    handlers: &mut Vec<ConfiguredHandler>,
+    warnings: &mut Vec<String>,
+    display_order: &mut i64,
+    plugin_hook_sources: Vec<PluginHookSource>,
+) {
+    // TODO(abhinav): check enabled/trusted state here before plugin hooks become runnable.
```
abhinav-oai, line 162: in future PRs will add
1. notion of a hook being trusted in order to run, similar to how a project is trusted
2. ability to toggle hooks, similar to skills

eternal-openai:
Generally looks very good to me, though: PLUGIN_ROOT calls for inline substitution in addition to being injected as a real env variable. The reason for this is that Powershell and co don't really do env this way, so it's better if the string is replaced inline. Also note PLUGIN_DATA which is a separate concept that we should eventually support

Also, we usually present a warning to the user if hooks are invalid, but I think here it silently fails. not very important, but worth noting

abhinav-oai:
Addressed in the latest push: I kept env injection and added inline substitution for `PLUGIN_ROOT` / `CLAUDE_PLUGIN_ROOT`, added `PLUGIN_DATA` / `CLAUDE_PLUGIN_DATA` backed by `~/.codex/plugins/data/<plugin>-<marketplace>/`, and restored startup warnings when plugin hook files fail to load or parse.

codex-rs/core/tests/suite/hooks.rs
```diff
@@ -732,7 +732,7 @@ fn request_message_input_texts(body: &[u8], role: &str) -> Vec<String> {
         .collect()
 }
 
-#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+#[tokio::test]
```
abhinav-oai, line 735: Switched the hooks integration suite back to plain `#[tokio::test]` after checking whether the two-worker runtime was buying us anything

- `cargo test -p codex-core suite::hooks::`
  - multi-thread: `27.84s`
  - single-thread: `6.69s`
- `cargo test -p codex-core suite::hooks:: -- --test-threads=1`
  - multi-thread: `26.56s`
  - single-thread: `11.43s`

eternal-openai (approved):

pakrym-oai (approved):

PR #19706, opened 2026-04-26
Preserve TUI markdown list spacing after code blocks

### LINKED ISSUE #19702: TUI markdown renderer collapses loose list spacing after code blocks

codex-cli 0.125.0

Pro

gpt-5.5

Darwin 24.6.0 arm64 arm

Ghostty

Markdown list items that contain block content, especially fenced code blocks, can render without a visual blank separator before the next list item.

The model output includes the blank line in the stored session JSONL, so this appears to be a TUI markdown rendering issue rather than the model omitting the newline.

Given source markdown like:

````markdown
1. First:

   ```rust
   fn foo() {}
   ```

2. Second:
````

Actual TUI rendering is visually closer to:

```text
1. First:

   fn foo() {}
2. Second:
```

Expected TUI rendering:

```text
1. First:

   fn foo() {}

2. Second:
```

Render a markdown response in the Codex TUI with an ordered list where one item contains a fenced code block and the next item follows after a blank line:

````markdown
1. First:

   ```rust
   fn foo() {}
   ```

2. Second:
````

The source markdown has a blank line between the first item content and the next marker, but the rendered transcript can omit the visual separator.

The TUI should preserve a readable blank separator before the next list marker when the previous list item contains block content such as a fenced code block.

Simple or tight lists should remain compact:

```markdown
1. First

2. Second
```

can still render as:

```text
1. First
2. Second
```

The goal is not to add extra vertical space to every list item, only to avoid visually attaching the next marker to block-heavy list content.

### Additional information

A local fix has been drafted on fork branch:

https://github.com/joshka/codex/tree/joshka/markdown-list-spacing

The draft change tracks when a list item contains block content and preserves a separator before the next marker only in that case. It includes regression coverage for both paths: code-block list items keep the separator, and simple list items stay compact.

etraut-openai:
## Why

Fixes #19702.

The TUI markdown renderer could visually attach the next list marker to a fenced code block inside the previous list item, even when the source markdown included a blank line before the next item. That made block-heavy loose lists harder to read, while the desired behavior is still to keep simple lists compact.

## What changed

- Track whether the current rendered list item contains a code block.
- Preserve one blank separator before the following list marker only when the previous item contained a code block.
- Add regression coverage for both paths: code-block list items keep the separator, and simple loose list items stay compact.

## Verification

- `cargo test -p codex-tui markdown_render`

I also manually verified that the bug exists before and is fixed after.

## Before
Screenshot 2026-04-26 at 1 19 01 PM

## After
Screenshot 2026-04-26 at 1 18 54 PM

### DIFF
--- a/codex-rs/tui/src/markdown_render.rs
+++ b/codex-rs/tui/src/markdown_render.rs
@@ -154,6 +154,8 @@ where
     inline_styles: Vec<Style>,
     indent_stack: Vec<IndentContext>,
     list_indices: Vec<Option<u64>>,
+    list_needs_blank_before_next_item: Vec<bool>,
+    list_item_contains_code_block: Vec<bool>,
     link: Option<LinkState>,
     needs_newline: bool,
     pending_marker_line: bool,
@@ -184,6 +186,8 @@ where
             inline_styles: Vec::new(),
             indent_stack: Vec::new(),
             list_indices: Vec::new(),
+            list_needs_blank_before_next_item: Vec::new(),
+            list_item_contains_code_block: Vec::new(),
             link: None,
             needs_newline: false,
             pending_marker_line: false,
@@ -292,6 +296,11 @@ where
             TagEnd::CodeBlock => self.end_codeblock(),
             TagEnd::List(_) => self.end_list(),
             TagEnd::Item => {
+                if self.list_item_contains_code_block.pop().unwrap_or(false)
+                    && let Some(needs_blank) = self.list_needs_blank_before_next_item.last_mut()
+                {
+                    *needs_blank = true;
+                }
                 self.indent_stack.pop();
                 self.pending_marker_line = false;
             }
@@ -476,15 +485,26 @@ where
             self.push_line(Line::default());
         }
         self.list_indices.push(index);
+        self.list_needs_blank_before_next_item.push(false);
     }
 
     fn end_list(&mut self) {
         self.list_indices.pop();
+        self.list_needs_blank_before_next_item.pop();
         self.needs_newline = true;
     }
 
     fn start_item(&mut self) {
+        if self
+            .list_needs_blank_before_next_item
+            .last_mut()
+            .map(std::mem::take)
+            .unwrap_or(false)
+        {
+            self.push_blank_line();
+        }
         self.pending_marker_line = true;
+        self.list_item_contains_code_block.push(false);
         let depth = self.list_indices.len();
         let is_ordered = self
             .list_indices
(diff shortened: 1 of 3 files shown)


codex-rs/tui/src/markdown_render.rs
```diff
@@ -524,6 +544,9 @@ where
     }
 
     fn start_codeblock(&mut self, lang: Option<String>, indent: Option<Span<'static>>) {
+        if let Some(item_contains_code_block) = self.list_item_contains_code_block.last_mut() {
```
fcoury-oai: Codex found another edge case that survives the fix:

```
Reply with exactly this markdown and nothing else:

1. First:
   - Nested:

     ```rust
     fn first() {}
     ```

2. Second:
```

This is the result in this branch:

<img width="487" height="396" alt="Image" src="https://github.com/user-attachments/assets/e2be32e1-f13b-4d0b-b650-c10f2caa506b" />

The solution: when a code block starts, mark every active enclosing list item as containing a code block, not just the innermost one. Then add a regression test where an outer list item contains a nested child with a fenced code block and is followed by another outer sibling.

The smallest fix is probably just this in `start_codeblock`:

```rust
fn start_codeblock(&mut self, lang: Option<String>, indent: Option<Span<'static>>) {
    for item_contains_code_block in &mut self.list_item_contains_code_block {
        *item_contains_code_block = true;
    }

    self.flush_current_line();
    // ...
}
```

Right now the code does this:

```rust
if let Some(item_contains_code_block) = self.list_item_contains_code_block.last_mut() {
    *item_contains_code_block = true;
}
```

That only marks the deepest list item. Iterating over the whole stack marks the nested child and every enclosing parent item, so when any of those items ends, the next sibling at that level
can keep the blank separator.

A matching regression test could look like this:

```rust
#[test]
fn outer_list_item_after_nested_code_block_keeps_blank_separator() {
    let md = "1. First:\n\
              \n\
                 - Nested:\n\
              \n\
                   ```rust\n\
                   fn first() {}\n\
                   ```\n\
              \n\
              2. Second:\n";

    let text = render_markdown_text(md);
    let lines = plain_lines(&text);

    assert_eq!(
        lines,
        vec![
            "1. First:",
            "    - Nested:",
            "",
            "      fn first() {}",
            "",
            "2. Second:",
        ]
    );
}
```

fcoury-oai (commented):
I think it's a good idea to use the presented solution to catch other edge cases.

codex-rs/tui/src/markdown_render.rs
etraut-openai: [codex] Thanks, agreed. I updated `start_codeblock` to mark every active enclosing list item and added `outer_list_item_after_nested_code_block_keeps_blank_separator` to cover the nested fenced-code case.

PR #19707, opened 2026-04-26
Fix remote compaction with deferred dynamic tools

### LINKED ISSUE #19486: [P0 / Blocker] Remote compact task fails 100% with "tools.defer_loading requires tools.tool_search" — GPT-5.5 unusable at context limit, no client-side workaround
## Severity: Blocker / P0

Codex becomes unusable the moment any non-trivial conversation reaches the auto-compact threshold. **100% reproducible.** **No viable client-side workaround** — rolling back the desktop app does not help because the failing call is constructed server-side.

This bug effectively makes the headline GPT-5.5 launch experience on Codex desktop dead on arrival for anyone running the default plugin set.

## Environment

- Codex desktop (macOS, Apple Silicon)
  - Reproduced on **26.422.30944** (current)
  - Reproduced on **26.422.21459** after manual rollback — identical failure
- Model: **GPT-5.5**, reasoning effort \`xhigh\`
- Plugins enabled (default-ish set): \`github\`, \`browser-use\`, \`documents\`, \`spreadsheets\`, \`presentations\`, \`computer-use\`

## Error

As soon as the conversation hits the compact threshold:

\`\`\`
Error running remote compact task: {
  "error": {
    "message": "Invalid Value: 'tools.defer_loading'. Deferred tools require tools.tool_search.",
    "type": "invalid_request_error",
    "param": "tools.defer_loading",
    "code": null
  }
}
\`\`\`

The session stalls. Every retry reproduces the same 400. The conversation cannot progress.

## Reproduction

1. Install Codex desktop 26.422.x on Apple Silicon.
2. Select GPT-5.5.
3. Keep default plugins enabled so tool descriptions cross the ~10% context threshold that auto-enables \`tool_search\` + \`defer_loading\`.
4. Work normally until the conversation approaches the compact threshold.
5. The remote compact endpoint returns the 400 above — every time, with no recovery path.

## Likely root cause

The server-side compact task (\`chatgpt.com/backend-api/codex/responses/compact\`) rebuilds the Responses API payload with \`defer_loading\` on individual tool entries but **omits the paired \`tool_search\` entry** at the top level of the \`tools\` array. The Responses API then rejects its own internally-constructed request. This regression coincides with \"tool discovery enabled by default\" in the recent changelog.

In other words: the server is failing its own schema validation against itself.

## Why the obvious workarounds do not work

- **Rolling back the desktop client (30944 → 21459):** identical failure. Compact is remote — client version is irrelevant.
- **Starting new threads:** resets context, but any substantive session hits the same wall again.
- **Switching to GPT-5.4:** \`defer_loading\` was introduced in 5.4, so it almost certainly exercises the same path.
- **Disabling plugins** until tool descriptions drop under the ~10% threshold is the only thing that *might* avoid it — i.e., asking paying users to disable advertised features to get the product to function.

## Impact

- GPT-5.5 + Codex desktop is **non-functional for any user with the default plugin set** as soon as a conversation fills.
- This is the flagship model from last week's launch.
- There is no preserved-functionality workaround.
- The fix does not require a client release — it is a pure server-side correction in the compact-task request builder.

## Requested fix

Either:
- (a) Include \`tool_search\` in the rebuilt compact request whenever any tool entry carries \`defer_loading\`, **or**
- (b) Strip \`defer_loading\` from tool entries in the rebuilt compact request when \`tool_search\` is intentionally omitted.

Please prioritize. This is a top-of-funnel regression on a paid flagship model.

## Related

- #19009 — Compact fails in CLI and VS Code Extension
- #19279 — Experimental feature check disabling app startup
- #19295 — White screen on v26.422.21637 (Pro)
- #19370 — GPT-5.5 not usable in Codex App for remote projects

pash-openai:
Remote compaction was rebuilding its compact prompt with the raw model-visible tool list, while normal model turns first remove any dynamic tool specs marked for deferred loading. After tool discovery started enabling deferred dynamic tools more often, compact requests could carry `defer_loading` entries into a compact path that may not preserve the paired top-level `tool_search` tool, producing the `tools.defer_loading requires tools.tool_search` failure at compaction time.

This moves the normal-turn prompt filtering into a shared helper and uses it when remote compaction builds its prompt. The compact request now sees the same model-visible tool set as an ordinary turn: `tool_search` remains available for discovery, and deferred dynamic tool definitions are kept out of the prompt until searched for.

I smoke-tested the API contract against the real compact endpoint using the exact Codex App dynamic tool introduced in openai/openai#849632: `codex_app::automation_update` with `deferLoading: true`, alongside the app's other dynamic tools. A compact request that included the deferred `automation_update` tool but no top-level `tool_search` failed with the reported 400: `Invalid Value: 'tools.defer_loading'. Deferred tools require tools.tool_search.` The same request shape after filtering out deferred dynamic tool definitions returned 200 with a compacted `message` and `compaction_summary`.

As a control, the live endpoint currently accepts the raw tool list when `tool_search` is present. The bug is the fragile coupling: if `defer_loading` survives into any compact request shape where `tool_search` is absent, compaction hard-fails before model work starts. Matching normal-turn filtering removes that failure mode entirely.

Fixes #19486.

### DIFF
--- a/codex-rs/core/src/compact_remote.rs
+++ b/codex-rs/core/src/compact_remote.rs
@@ -12,6 +12,7 @@ use crate::context_manager::estimate_response_item_model_visible_bytes;
 use crate::context_manager::is_codex_generated_item;
 use crate::session::session::Session;
 use crate::session::turn::built_tools;
+use crate::session::turn::model_visible_specs_for_prompt;
 use crate::session::turn_context::TurnContext;
 use codex_analytics::CompactionImplementation;
 use codex_analytics::CompactionPhase;
@@ -165,7 +166,7 @@ async fn run_remote_compact_task_inner_impl(
     .await?;
     let prompt = Prompt {
         input: prompt_input,
-        tools: tool_router.model_visible_specs(),
+        tools: model_visible_specs_for_prompt(tool_router.as_ref(), turn_context.as_ref()),
         parallel_tool_calls: turn_context.model_info.supports_parallel_tool_calls,
         base_instructions,
         personality: turn_context.personality,
--- a/codex-rs/core/src/session/turn.rs
+++ b/codex-rs/core/src/session/turn.rs
@@ -945,21 +945,7 @@ pub(crate) fn build_prompt(
     turn_context: &TurnContext,
     base_instructions: BaseInstructions,
 ) -> Prompt {
-    let deferred_dynamic_tools = turn_context
-        .dynamic_tools
-        .iter()
-        .filter(|tool| tool.defer_loading)
-        .map(|tool| ToolName::new(tool.namespace.clone(), tool.name.clone()))
-        .collect::<HashSet<_>>();
-    let tools = if deferred_dynamic_tools.is_empty() {
-        router.model_visible_specs()
-    } else {
-        router
-            .model_visible_specs()
-            .into_iter()
-            .filter_map(|spec| filter_deferred_dynamic_tool_spec(spec, &deferred_dynamic_tools))
-            .collect()
-    };
+    let tools = model_visible_specs_for_prompt(router, turn_context);
 
     Prompt {
         input,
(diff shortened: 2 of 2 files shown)


pash-openai:
@codex review

PR #19708, opened 2026-04-26
Load cloud requirements for agent identity

shijie-oai:
## Why

Agent Identity sessions can represent Business and Enterprise ChatGPT workspaces, but cloud requirements were skipped before fetch. That meant workspace-managed requirements were not loaded for Agent Identity even when the JWT carried the same account identity and plan information that normal ChatGPT token auth exposes.

This PR now sits on top of the Agent Identity stack through [#19764](https://github.com/openai/codex/pull/19764). Because [#19763](https://github.com/openai/codex/pull/19763) moved task registration into Agent Identity auth loading, cloud requirements no longer needs a separate runtime-initialization step before building the backend client.

## What changed

- Stop skipping `CodexAuth::AgentIdentity` in the cloud requirements loader.
- Share the cloud requirements eligibility check between startup load and background cache refresh.
- Rely on eagerly loaded Agent Identity auth so backend requests can attach task-scoped `AgentAssertion` headers.
- Decode Agent Identity JWT `plan_type` as the auth-layer plan type, then convert it through a shared `auth::PlanType` -> `account::PlanType` mapping.
- Add the missing serde alias for the `education` plan string and add coverage for raw Agent Identity plan aliases such as `hc` and `education`.

## Testing

- `cargo test -p codex-agent-identity -p codex-login -p codex-cloud-requirements -p codex-protocol`

### DIFF
--- a/codex-rs/cloud-requirements/src/lib.rs
+++ b/codex-rs/cloud-requirements/src/lib.rs
@@ -179,6 +179,14 @@ fn auth_identity(auth: &CodexAuth) -> (Option<String>, Option<String>) {
     (auth.get_chatgpt_user_id(), auth.get_account_id())
 }
 
+fn cloud_requirements_eligible_auth(auth: &CodexAuth) -> bool {
+    let Some(plan_type) = auth.account_plan_type() else {
+        return false;
+    };
+    auth.uses_codex_backend()
+        && (plan_type.is_business_like() || matches!(plan_type, PlanType::Enterprise))
+}
+
 fn cache_payload_bytes(payload: &CloudRequirementsCacheSignedPayload) -> Option<Vec<u8>> {
     serde_json::to_vec(&payload).ok()
 }
@@ -329,17 +337,7 @@ impl CloudRequirementsService {
         let Some(auth) = self.auth_manager.auth().await else {
             return Ok(None);
         };
-        if matches!(auth, CodexAuth::AgentIdentity(_)) {
-            // AgentIdentity does not carry a human bearer token, and identity-edge
-            // only allowlists task-scoped AgentAssertion calls for the Codex runtime.
-            return Ok(None);
-        }
-        let Some(plan_type) = auth.account_plan_type() else {
-            return Ok(None);
-        };
-        if !auth.uses_codex_backend()
-            || !(plan_type.is_business_like() || matches!(plan_type, PlanType::Enterprise))
-        {
+        if !cloud_requirements_eligible_auth(&auth) {
             return Ok(None);
         }
         let (chatgpt_user_id, account_id) = auth_identity(&auth);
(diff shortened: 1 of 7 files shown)


codex-rs/login/src/auth/manager.rs
```diff
@@ -321,10 +320,7 @@ impl CodexAuth {
         }
     }
 
-    pub async fn initialize_runtime(
-        &self,
-        _chatgpt_base_url: Option<String>,
```
shijie-oai: Not used.

codex-rs/cloud-requirements/src/lib.rs
```diff
@@ -326,20 +348,10 @@ impl CloudRequirementsService {
...
-        let Some(auth) = self.auth_manager.auth().await else {
-            return Ok(None);
-        };
-        if matches!(auth, CodexAuth::AgentIdentity(_)) {
-            // AgentIdentity does not carry a human bearer token, and identity-edge
-            // only allowlists task-scoped AgentAssertion calls for the Codex runtime.
-            return Ok(None);
-        }
-        let Some(plan_type) = auth.account_plan_type() else {
+        let Some(auth) = self.auth_manager.auth_snapshot().await else {
```
shijie-oai: The reason to `auth_snapshot` so that we are not getting a task identity because it is not necessary for us to use the exiting auth state to validate and apply the cloud requirements. Previously auth() failure would fail open even if the agent identity is tied to an enterprise or a business liked plan.

codex-rs/cloud-requirements/src/lib.rs
```diff
@@ -329,17 +337,7 @@ impl CloudRequirementsService {
         let Some(auth) = self.auth_manager.auth().await else {
             return Ok(None);
         };
-        if matches!(auth, CodexAuth::AgentIdentity(_)) {
```
shijie-oai, line 332: For agent identities - we would like to fetch cloud requirements as expected.

codex-rs/login/src/auth/auth_tests.rs
```diff
@@ -945,6 +952,13 @@ fn fake_agent_identity_jwt(record: &AgentIdentityAuthRecord) -> std::io::Result<
 
 fn signed_agent_identity_jwt(
     record: &AgentIdentityAuthRecord,
+) -> jsonwebtoken::errors::Result<String> {
+    signed_agent_identity_jwt_with_plan_type(record, json!(record.plan_type))
+}
+
+fn signed_agent_identity_jwt_with_plan_type(
```
efrazer-oai: Let's just make signed_agent_identity_jwt accept PlanType instead of helper below, otherwise lgtm

efrazer-oai (approved):

codex-rs/login/src/auth/auth_tests.rs
shijie-oai: Addressed

sayan-oai (approved):

PR #19709, opened 2026-04-26
Render delegated patch approval details

### LINKED ISSUE #19632: Subagent hit sandbox, TUI wouldn't show me the command

codex-cli 0.125.0

plus

5.3-codex

Microsoft Windows NT 10.0.26200.0 x64

Windows Terminal

Subagent hit sandbox, TUI wouldn't show me the command it wanted to run

Image

Uploaded thread: 019dc7d0-09ce-7131-a2d4-ac73733b7592

_No response_

### Additional information

_No response_

etraut-openai:
## Why

Fixes #19632.

When a delegated agent requests approval for an in-progress file change, the parent TUI handles that request from an inactive thread. The app server already sent the `FileChange` item with the proposed diff, but the inactive-thread approval path was not recovering and rendering it the same way as the active-thread path.

The result was an inconsistent approval prompt: main-thread edits show a normal patch preview history item before the approval modal, while delegated edits did not show that preview in the transcript flow.

## What Changed

- Recover buffered or historical `FileChange` item changes when building inactive-thread file-change approval requests.
- Reuse the app-server file-change conversion helper for both live transcript rendering and inactive-thread approvals.
- Render recovered delegated patches as a normal patch preview history cell before the approval modal.
- Keep apply-patch approval modals focused on the decision prompt and optional metadata; they do not render a synthetic command line or embed the diff body.

## Manual Repro And Verification

I manually reproduced the issue using a file under `~/Desktop` so the write would require approval.

Before the fix:

1. Ask the main thread: `Use apply_patch, not shell redirection or Python, to create ~/Desktop/bug1.txt with three short lines.`
2. Observe the expected TUI shape: the transcript shows a normal patch preview such as `• Added ~/Desktop/bug1.txt (+N -0)` above the approval modal, and the modal contains only the approval prompt/options without a synthetic command line.
3. Ask for the delegated path: `Spawn a worker. Have it use apply_patch, not shell redirection or Python, to create ~/Desktop/bug1.txt with four short lines.`
4. Observe the delegated approval is inconsistent: the parent view does not render the proposed patch as the normal transcript preview before the modal, so the diff context is missing from the stream or appears inside the modal instead of in the history flow.

After the fix:

1. Repeat the delegated worker prompt with `apply_patch`.
2. Confirm the parent view renders the same normal patch preview history cell (`• Added ~/Desktop/bug1.txt (+N -0)` plus the diff) immediately before the approval modal.
3. Confirm the approval modal remains focused on the decision prompt. For delegated approvals it may show the worker thread label, but it should not show a `$ apply_patch` command line or embed the diff body in the modal.

### DIFF
--- a/codex-rs/tui/src/app/tests.rs
+++ b/codex-rs/tui/src/app/tests.rs
@@ -30,6 +30,9 @@ use codex_app_server_protocol::AdditionalPermissionProfile;
 use codex_app_server_protocol::AgentMessageDeltaNotification;
 use codex_app_server_protocol::CommandExecutionRequestApprovalParams;
 use codex_app_server_protocol::ConfigWarningNotification;
+use codex_app_server_protocol::FileChangeRequestApprovalParams;
+use codex_app_server_protocol::FileUpdateChange;
+use codex_app_server_protocol::ItemStartedNotification;
 use codex_app_server_protocol::JSONRPCErrorError;
 use codex_app_server_protocol::McpServerStartupState;
 use codex_app_server_protocol::McpServerStatusUpdatedNotification;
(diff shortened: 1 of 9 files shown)


fcoury-oai (approved):
Thanks for the replication instructions, it made my life much easier. I indeed replicated the issue and verified it works on this branch.

Code looks good! Approved.

PR #19726, opened 2026-04-26
Fix codex-core config test type paths

pakrym-oai:
Summary:
- Update config tests to reference config requirement types from codex_config after the loader split.

Tests:
- just fmt
- cargo build -p codex-core --tests
- cargo clippy -p codex-core --tests -- -D warnings


aibrahim-oai (approved):

bolinfest (approved):
Thanks for the quick fix!

PR #19739, opened 2026-04-27
inline hostname resolution for remote sandbox config

abhinav-oai:
# Why

Requirements support host-specific `remote_sandbox_config.hostname_patterns`, but config loading previously resolved and passed the system hostname through every config-loading path even when no requirements layer used `remote_sandbox_config`. On machines where hostname lookup is slow, startup and app-server config reads paid for a feature that was not active.

We only need the hostname when a requirements layer actually declares `remote_sandbox_config`, so this moves hostname resolution to the single requirements merge point and keeps all other config callers unaware of hostname matching.

# What

- Removed the eager `host_name` plumbing from `load_config_layers_state`, `load_requirements_toml`, `ConfigBuilder`, app-server `ConfigManager`, network proxy loading, and related call sites.
- Resolve the hostname inside `merge_requirements_with_remote_sandbox_config` only when the incoming requirements contain `remote_sandbox_config`.


codex-rs/config/src/loader/mod.rs
```diff
@@ -602,9 +588,11 @@ pub(super) fn merge_requirements_with_remote_sandbox_config(
     target: &mut ConfigRequirementsWithSources,
     source: RequirementSource,
     mut requirements: ConfigRequirementsToml,
-    host_name: Option<&str>,
 ) {
-    requirements.apply_remote_sandbox_config(host_name);
+    if requirements.remote_sandbox_config.is_some() {
+        let host_name = crate::host_name();
```
abhinav-oai, line 593: basically just inlined it here

abhinav-oai:
@codex review

etraut-openai (approved):
Nice! This not only addresses the regression but also simplifies the code quite a bit.

PR #19751, opened 2026-04-27
Add opaque identity and make available to pass through to remote contracts

bryanashley:
## What

Add an opaque app-server identity key that can be provided at startup and forwarded to remote contract implementations.

This change:
- Adds an `IdentityKey` wrapper type in `codex-app-server`
- Adds optional `--identity-key` support to the standalone app-server binary and `codex app-server`
- Plumbs the optional key through app-server startup, in-process startup, and message processor args
- Adds RemoteThreadStore support for forwarding the identity key as binary gRPC metadata
- Re-exports the remote identity metadata header for remote contract implementations
- Adds focused coverage for opaque byte preservation, CLI parsing, and remote thread-store metadata forwarding

## Why

Some app-server deployments need to pass an opaque identity value through to remote contract implementations. The value must be preserved as bytes, including non-UTF-8 argv data on Unix, and should not be interpreted by the app-server itself.

## How

`IdentityKey` stores the key as raw bytes and converts from `OsString` at the CLI boundary. `RemoteThreadStore` now has an identity-key-aware constructor and attaches the key using a tonic interceptor with the `x-codex-app-server-identity-key-bin` metadata header.

Existing startup APIs remain compatible through `run_main_with_transport`; callers that need identity forwarding can use `run_main_with_transport_options`.

## Testing

- `just fmt`
- `cargo test -p codex-app-server identity_key`
- `cargo test -p codex-thread-store list_threads_forwards_identity_key_as_metadata`

### DIFF
--- a/codex-rs/app-server-client/src/lib.rs
+++ b/codex-rs/app-server-client/src/lib.rs
@@ -405,6 +405,7 @@ impl InProcessClientStartArgs {
             session_source: self.session_source,
             enable_codex_api_key_env: self.enable_codex_api_key_env,
             initialize,
+            identity_key: None,
             channel_capacity: self.channel_capacity,
         }
     }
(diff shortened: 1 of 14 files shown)


bryanashley:

PR #19761, opened 2026-04-27
Forward macOS DeviceCheck attestation from Codex app-server helper

jiamingz42:
> **DO NOT SUBMIT:** This prototype now uses the dedicated helper bundle ID `com.openai.codex.app-server`; keep the PR draft-only until that bundle ID is finalized for production use.

## Why
This adds the macOS plumbing needed for Codex to attach DeviceCheck evidence from a small signed app-server helper even when the runtime is outside the desktop app bundle. The intent is to make DeviceCheck available as an additional weak attestation / risk signal for ChatGPT-authenticated Codex traffic while keeping token generation isolated in a small signed app-server helper with its own bundle identity.

## Summary

image

This PR has three pieces:

1. **Add a macOS DeviceCheck app-server helper.**
   - Introduces `DeviceCheckProbe.app`, a small app-server helper that generates DeviceCheck tokens on macOS using bundle ID `com.openai.codex.app-server`.
   - `codex-rs` discovers the helper from the packaged runtime layout, invokes it when available, and falls back to a structured failure payload when token generation is unavailable instead of failing the request.

2. **Send DeviceCheck evidence on the Codex request paths that need it.**
   - ChatGPT-authenticated Codex requests now attach `x-oai-attestation` for the three provider endpoints that should carry the signal:
     - `/responses`
     - `/responses/compact`
     - `/realtime/calls`
   - The header is only added for the ChatGPT Codex provider base, not generic API-key traffic.

3. **Ship the signed helper in release artifacts.**
   - Extends the macOS signing action to handle app bundles.
   - Builds, signs, and packages `DeviceCheckProbe.app` in the Apple Silicon release flow.
   - Includes the helper in the Darwin arm64 npm package, with CI coverage for the expected staging layout.

## Validation
- `cargo test -p codex-core`
- `python3 codex-cli/scripts/test_build_npm_package.py`
- Built the helper with `./codex-rs/devicecheck-probe/build-probe.sh --target aarch64-apple-darwin --out /tmp/devicecheck-probe-release-test`
- Verified a real Codex request to `https://chatgpt.com/backend-api/codex/responses` carried `x-oai-attestation`, and the token passed live sa-server validation with `is_ok=true`.

### DIFF
--- a/.github/actions/macos-code-sign/action.yml
+++ b/.github/actions/macos-code-sign/action.yml
@@ -7,6 +7,10 @@ inputs:
   binaries:
     description: Space-delimited binary basenames to sign and notarize.
     default: "codex codex-responses-api-proxy"
+  app-bundles:
+    description: Space-delimited macOS app bundle basenames to sign and notarize.
+    required: false
+    default: ""
   sign-binaries:
     description: Whether to sign and notarize the macOS binaries.
     required: false
@@ -123,6 +127,7 @@ runs:
       env:
         TARGET: ${{ inputs.target }}
         BINARIES: ${{ inputs.binaries }}
+        APP_BUNDLES: ${{ inputs.app-bundles }}
       run: |
         set -euo pipefail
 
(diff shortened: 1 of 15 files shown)


jiamingz42:

PR #19769, opened 2026-04-27
Add proactive Codex usage prompt

jchu-oai:
## Summary
- Add a 90% proactive Codex usage prompt in the TUI using existing rate-limit snapshots
- Branch the CTA for Plus, Pro/ProLite, and explicit workspace-owner limit cases
- Open public ChatGPT upgrade or usage settings destinations with CLI UTM attribution
- Leave analytics and experiment plumbing out of this PR

## Notes
- Workspace members are intentionally excluded from direct purchase prompts
- UTM naming and product copy can be adjusted while this remains draft

### DIFF
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -201,6 +201,7 @@ use codex_protocol::protocol::Op;
 use codex_protocol::protocol::PatchApplyBeginEvent;
 use codex_protocol::protocol::RateLimitReachedType;
 use codex_protocol::protocol::RateLimitSnapshot;
+use codex_protocol::protocol::RateLimitWindow;
 use codex_protocol::protocol::ReviewRequest;
 use codex_protocol::protocol::ReviewTarget;
 use codex_protocol::protocol::SkillMetadata as ProtocolSkillMetadata;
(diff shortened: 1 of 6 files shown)


codex-rs/tui/src/chatwidget/tests/status_and_layout.rs
```diff
@@ -542,6 +541,138 @@ async fn rate_limit_switch_prompt_popup_snapshot() {
...
+    chat.maybe_show_pending_rate_limit_prompt();
+    chat.handle_key_event(KeyEvent::new(KeyCode::Char('y'), KeyModifiers::NONE));
+
+    let (url, message) = next_open_url_and_info_message_events(&mut rx);
+    assert_eq!(url, PROACTIVE_USAGE_PROMPT_PRO_URL);
+    assert_eq!(message, PROACTIVE_USAGE_PROMPT_BROWSER_MESSAGE);
+}
+
+#[tokio::test]
+async fn proactive_usage_prompt_no_suppresses_repeat_prompts() {
```
rahult-oai, line 619: I think `proactive_usage_prompt_no_suppresses_repeat_prompts` may not be proving what its name says. It only asserts that no URL/info event was sent, but a repeat prompt would also send no event unless the user selected “Yes”. Could we also assert that the prompt is not visible after the later 95% snapshot?

codex-rs/tui/src/chatwidget.rs
```diff
@@ -495,6 +480,11 @@ fn is_standard_tool_call(parsed_cmd: &[ParsedCommand]) -> bool {
 const RATE_LIMIT_WARNING_THRESHOLDS: [f64; 3] = [75.0, 90.0, 95.0];
 const NUDGE_MODEL_SLUG: &str = "gpt-5.4-mini";
 const RATE_LIMIT_SWITCH_PROMPT_THRESHOLD: f64 = 90.0;
+const PROACTIVE_USAGE_PROMPT_THRESHOLD: f64 = 90.0;
+const PROACTIVE_USAGE_PROMPT_USAGE_URL: &str = "https://chatgpt.com/codex/settings/usage?utm_source=codex_cli&utm_medium=cli&utm_campaign=usage_limit_prompt&utm_content=threshold_90";
+const PROACTIVE_USAGE_PROMPT_PRO_URL: &str = "https://chatgpt.com/explore/pro?utm_source=codex_cli&utm_medium=cli&utm_campaign=usage_limit_prompt&utm_content=threshold_90";
```
rahult-oai, line 485: Nit: Tracking params are copy pasted in 2 places, might we wanna have a helper 
`const USAGE_PROMPT_UTM: &str =
    "utm_source=codex_cli&utm_medium=cli&utm_campaign=usage_limit_prompt&utm_content=threshold_90";`

PR #19771, opened 2026-04-27
fix: filter dynamic deferred tools from model_visible_specs

### LINKED ISSUE #19486: [P0 / Blocker] Remote compact task fails 100% with "tools.defer_loading requires tools.tool_search" — GPT-5.5 unusable at context limit, no client-side workaround

sayan-oai:
fixes #19486

### Problem
Right now dynamic deferred tools are filtered at normal-turn prompt building time, rather than upstream while building the `ToolRouter` itself. This causes issues because dynamic deferred tools are then wrongly included in the router's `model_visible_specs`, which is what the compaction request-building flow relies on.

### Fix
Move the dynamic deferred tool filtering to `ToolRouter` creation time to solve this problem for every request that relies on `ToolRouter` for `model_visible_specs`, which solves the issue generically.

### Tests
Added unit + integration tests to ensure dynamic deferred tools are omitted from `model_visible_specs` and compaction request respectively.

Tested against live `/compact` endpoint; raw deferred dynamic tools without `tool_search` returned `400` (current bug), while the filtered payload (this fix) returns `200`.


codex-rs/core/src/tools/router.rs
```diff
@@ -71,23 +71,26 @@ impl ToolRouter {
...
-                    }
-                })
-                .collect()
-        } else {
-            specs
-                .iter()
-                .map(|configured_tool| configured_tool.spec.clone())
-                .collect()
-        };
+        let deferred_dynamic_tools = dynamic_tools
```
pakrym-oai, line 74: instead of post filtering can we not add deferred tools to specs?

sayan-oai: code mode builds its `ALL_TOOLS` list from `router.specs()`, and the specs are used to create code mode tool definitions (name, schema, description, etc.). so the tools have to be present in `specs()`, but not `model_visible_specs()`.

for MCPs, code mode registers all MCP tools as direct while building its nested router so that `specs()` is fully populated and can be used to build `ALL_TOOLS` as described.

pakrym-oai (approved):
--

PR #19778, opened 2026-04-27
Add hooks/list app-server RPC

abhinav-oai:
## Why

We need a way to list the available hooks to expose via the TUI and App so users can view and manage their hooks

## What

- Adds `hooks/list` for one or more `cwd` values that returns discovered hook metadata

## Stack

1. openai/codex#19705
2. This PR - openai/codex#19778
3. openai/codex#19840
4. openai/codex#19882

## Review Notes

The generated schema files account for most of the raw diff, these files have the core change:

- `hooks/src/engine/discovery.rs` builds the inventory entries during hook discovery while leaving runtime handlers focused on execution.
- `app-server/src/codex_message_processor.rs` wires `hooks/list` into the app-server flow for each requested `cwd`.
- `app-server-protocol/src/protocol/v2.rs` defines the new v2 request/response payloads exposed on the wire.

### Core Changes

`core/src/plugins/manager.rs` adds `plugins_for_layer_stack(...)` so `skills/list` and `hooks/list`can resolve plugin state for each requested `cwd`


codex-rs/core/src/plugins/manager.rs
```diff
@@ -332,7 +332,7 @@ pub struct PluginsManager {
     featured_plugin_ids_cache: RwLock<Option<CachedFeaturedPluginIds>>,
     configured_marketplace_upgrade_state: RwLock<ConfiguredMarketplaceUpgradeState>,
     non_curated_cache_refresh_state: RwLock<NonCuratedCacheRefreshState>,
-    cached_enabled_outcome: RwLock<Option<PluginLoadOutcome>>,
+    cached_enabled_outcome: RwLock<Option<(bool, PluginLoadOutcome)>>,
```
abhinav-oai: need this in case the flag changes, we can't return the same cached results

temporary while `plugin_hooks` is under development, will rip all this out once the safety + trust has landed

abhinav-oai:
@codex review

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -6193,6 +6199,77 @@ impl CodexMessageProcessor {
...
+            // gates are enabled for this workspace.
+            let (plugin_hook_sources, plugin_hook_load_warnings) = plugins_manager
+                .effective_plugin_hooks_for_layer_stack(
+                    &config_layer_stack,
+                    plugins_enabled,
+                    config.features.enabled(Feature::PluginHooks),
+                )
+                .await;
+            let hooks = codex_hooks::list_hooks(codex_hooks::HooksConfig {
+                feature_enabled: config.features.enabled(Feature::CodexHooks),
```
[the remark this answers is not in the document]
abhinav-oai: I think codex is just getting tripped up by the naming

1. we check the features from the actual `cwd` that codex is running from
2. the `cwd`s that get provided to `hooks/list` (also `skills/list`) is just a list of dirs to look for `hooks.json` files. we **don't** wanna respect any potential features in those dirs

codex-rs/core/src/plugins/manager.rs
xl-openai: Can you add a comment suggesting the meaning of this bool so people don't need to read the code?

eternal-openai:
PR reads well! One minor concern. Can you pull on this thread:

> hooks/list resolves a per-cwd ConfigLayerStack, but then uses a single process-global Config snapshot to decide plugins, plugin_hooks, and codex_hooks. If the app-server was started in repo A and the client asks for hooks in repo B, differing feature flags in B's .codex/config.toml are ignored: hooks can disappear when B enables them, or be exposed when B disables them. Because the RPC is explicitly multi-cwd, its enablement checks need to be derived from each requested cwd's effective config/layer stack.

So my understanding of this issue:

- Start app-server while sitting in repo A
- Repo A has .codex/config.toml with codex_hooks = false
- Repo B has .codex/config.toml with codex_hooks = true, plus hooks configured
- Call hooks/list for repo B

What happens now:

The RPC correctly loads repo B’s hook definitions via repo B’s ConfigLayerStack. Then it passes feature_enabled: config.features.enabled(Feature::CodexHooks) from the global config, which still reflects repo A

Result: hooks/list says “no hooks” for repo B

pakrym-oai (approved):

PR #19807, opened 2026-04-27
[codex] Promote Codex App on Windows in tooltips

vb-openai:
## Summary
- Mention macOS and Windows in the README Codex App promo.
- Update startup tooltip copy to promote the Codex App on macOS and Windows.
- Rotate Free/Go users on supported desktop platforms between the free-plan promo and the Codex App promo.
- Keep Codex App tips hidden on unsupported platforms and add focused tooltip tests.

## Validation
- `just fmt`
- `git diff --check`
- `cargo test -p codex-tui tooltips::tests`

Note: full `cargo test -p codex-tui` still aborts in an unrelated stack overflow test: `app::tests::attach_live_thread_for_selection_rejects_unmaterialized_fallback_threads`.


codex-rs/tui/src/tooltips.rs
```diff
@@ -110,6 +109,14 @@ fn pick_paid_tooltip<R: Rng + ?Sized>(
     }
 }
 
+fn pick_free_go_tooltip<R: Rng + ?Sized>(rng: &mut R, supports_desktop_app: bool) -> &'static str {
```
jif-oai, line 112: No we shouldn't re-implement one more randomization path here. This code is becoming extremely messy!
Please just use toolstips.txt and if you need something broader, then use the announcement thing

PR #19836, opened 2026-04-27
disallow fileparams metadata for custom mcps

colby-oai:
## Summary
Disallow fileParams metadata for custom MCPs 

Restricts Codex openai/fileParams handling to the first-party codex_apps MCP server. Custom MCP servers may still advertise the metadata, but Codex now ignores it for upload rewriting, preventing non-Apps tools from receiving signed OpenAI file refs for local paths. Added a regression test for the allowed and denied cases.

### DIFF
--- a/codex-rs/core/src/mcp_tool_call.rs
+++ b/codex-rs/core/src/mcp_tool_call.rs
@@ -1206,11 +1206,9 @@ fn openai_file_input_params_for_server(
     server: &str,
     meta: Option<&serde_json::Map<String, serde_json::Value>>,
 ) -> Option<Vec<String>> {
-    if server != CODEX_APPS_MCP_SERVER_NAME {
-        return None;
-    }
-
-    Some(declared_openai_file_input_param_names(meta)).filter(|params| !params.is_empty())
+    (server == CODEX_APPS_MCP_SERVER_NAME)
+        .then_some(declared_openai_file_input_param_names(meta))
+        .filter(|params| !params.is_empty())
 }
 
 fn get_mcp_app_resource_uri(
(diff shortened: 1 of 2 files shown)


codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -1194,13 +1194,24 @@ pub(crate) async fn lookup_mcp_tool_metadata(
             .and_then(|meta| meta.get(MCP_TOOL_CODEX_APPS_META_KEY))
             .and_then(serde_json::Value::as_object)
             .cloned(),
-        openai_file_input_params: Some(declared_openai_file_input_param_names(
+        openai_file_input_params: openai_file_input_params_for_server(
```
soheil-oai, line 1198: nit: can we add some comment notes on the reason behind this for future?

soheil-oai (approved):

codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -1194,13 +1194,24 @@ pub(crate) async fn lookup_mcp_tool_metadata(
...
+    server: &str,
+    meta: Option<&serde_json::Map<String, serde_json::Value>>,
+) -> Option<Vec<String>> {
+    if server != CODEX_APPS_MCP_SERVER_NAME {
+        return None;
+    }
+
+    Some(declared_openai_file_input_param_names(meta)).filter(|params| !params.is_empty())
+}
+
```
caseychow-oai, line 1213: this is not very idomatic rust imo, would suggest using `then_some` to help inline this: https://stackoverflow.com/a/62166509

mzeng-openai (approved):

owenlin0 (approved):

PR #19839, opened 2026-04-27
[codex] Trace cancelled inference streams

cassirer-openai:
Records cancelled inference streams when Codex stops consuming a provider response before `response.completed`, preserving complete output items observed before cancellation.

Also closes still-running inference calls when the owning turn ends, so reduced rollout traces do not leave stale `Running` inference nodes.

Covered by focused reducer coverage and a core stream-drop test for partial output preservation.

### DIFF
--- a/codex-rs/core/src/client.rs
+++ b/codex-rs/core/src/client.rs
@@ -1623,6 +1623,9 @@ fn parent_thread_id_header_value(session_source: &SessionSource) -> Option<Strin
     }
 }
 
+const RESPONSE_STREAM_CHANNEL_CAPACITY: usize = 1600;
+const STREAM_DROPPED_REASON: &str = "response stream dropped before provider terminal event";
+
 fn map_response_stream<S>(
     api_stream: S,
     session_telemetry: SessionTelemetry,
@@ -1634,7 +1637,8 @@ where
         + Send
         + 'static,
 {
-    let (tx_event, rx_event) = mpsc::channel::<Result<ResponseEvent>>(1600);
+    let (tx_event, rx_event) =
+        mpsc::channel::<Result<ResponseEvent>>(RESPONSE_STREAM_CHANNEL_CAPACITY);
     let (tx_last_response, rx_last_response) = oneshot::channel::<LastResponse>();
     let consumer_dropped = CancellationToken::new();
     let consumer_dropped_for_stream = consumer_dropped.clone();
@@ -1647,10 +1651,7 @@ where
         loop {
             let event = tokio::select! {
                 _ = consumer_dropped.cancelled() => {
-                    inference_trace_attempt.record_cancelled(
-                        "response stream dropped before provider terminal event",
-                        &items_added,
-                    );
+                    inference_trace_attempt.record_cancelled(STREAM_DROPPED_REASON, &items_added);
                     return;
                 }
                 event = api_stream.next() => event,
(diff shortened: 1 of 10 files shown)


jif-oai:
@codex review

codex-rs/core/src/client.rs
```diff
@@ -1634,13 +1636,28 @@ where
...
 
     tokio::spawn(async move {
         let mut logged_error = false;
         let mut tx_last_response = Some(tx_last_response);
         let mut items_added: Vec<ResponseItem> = Vec::new();
         let mut api_stream = api_stream;
-        while let Some(event) = api_stream.next().await {
+        loop {
+            let event = tokio::select! {
+                _ = consumer_dropped.cancelled() => {
```
jif-oai, line 1653: This only records cancellation while the mapper is waiting on api_stream.next(). If the consumer is dropped while we’re blocked in one of the `tx_event.send(...).await` paths, we loose it. So an interrupted stream can lose what we're trying to do here

jif-oai (commented):
Can we make the mapper own one “finish once” path for every non-completed stream exit? and keep the reducer fallback as a last-resort repair only?

cassirer-openai:
>Can we make the mapper own one “finish once” path for every non-completed stream exit? and keep the reducer fallback as a last-resort repair only?

Yes. I updated the stream mapper so current streams should now close themselves explicitly instead of depending on reducer repair.

`map_response_stream` now records a terminal trace event for each non-completed exit it can observe:

API/provider stream error records InferenceFailed, carrying any completed output items observed before the error.
Downstream consumer drop / closed response channel records `InferenceCancelled`, also carrying any completed output items observed before the drop.
Provider stream EOF before `response.completed` records `InferenceFailed`.
The attempt now has a terminal guard, so whichever path wins records exactly once. The reducer turn-end closure remains only as a last-resort repair for malformed/older traces where an inference is still Running when the owning turn ends.

codex-rs/core/src/client.rs
cassirer-openai: Yep, good catch. The token only covers the `api_stream.next()` wait. I added explicit cancellation recording on the `tx_event.send(...).await` error paths too, so if the receiver is dropped while a send is blocked we still close the inference as cancelled and keep the items observed so far.

codex-rs/rollout-trace/src/reducer/mod.rs
```diff
@@ -254,6 +254,20 @@ impl TraceReducer {
...
+            RawTraceEventPayload::InferenceCancelled {
+                inference_call_id,
+                partial_response_payload,
+                ..
+            } => {
+                self.complete_inference_call(
+                    event.seq,
+                    event.wall_time_unix_ms,
+                    inference_call_id,
+                    ExecutionStatus::Cancelled,
```
jif-oai, line 266: This can still race with turn-end cleanup: if the turn ends as failed/ aborted first, a late async InferenceCancelled can overwrite that status back to cancelled. Can we keep the existing terminal status here and only add the partial response?

jif-oai (approved):
lgtm after this

codex-rs/rollout-trace/src/reducer/mod.rs
cassirer-openai: Good catch! Fixed.

PR #19841, opened 2026-04-27
permissions: remove cwd special path

bolinfest:
## Why

The experimental `PermissionProfile` API had both `:cwd` and `:project_roots` special filesystem paths, which made the permission root ambiguous. This PR removes the unstable `current_working_directory` special path before the permissions API is stabilized, so callers use `:project_roots` for symbolic project-root access.

## What changed

- Removes `FileSystemSpecialPath::CurrentWorkingDirectory` from protocol and app-server protocol models, plus regenerated app-server JSON/TypeScript schemas.
- Replaces internal `:cwd` permission entries with `:project_roots` entries.
- Keeps the existing cwd-update behavior for legacy-shaped workspace-write profiles, while removing the deleted `CurrentWorkingDirectory` case from that compatibility path.
- Keeps `PermissionProfile::workspace_write()` as the reusable symbolic workspace-write helper, with docs noting that `:project_roots` entries resolve at enforcement time.
- Updates app-server docs/examples and approval UI labeling to stop advertising `:cwd` as a permission token.

## Compatibility

Persisted rollout items may contain the old `{"kind":"current_working_directory"}` tag from earlier experimental `permissionProfile` snapshots. This PR keeps that tag as a deserialize-only alias for `ProjectRoots { subpath: None }`, while continuing to serialize only the new `project_roots` tag.

## Follow-up

This PR intentionally does not introduce an explicit project-root set on `SessionConfiguration` or runtime sandbox resolution. Today, the resolver still uses the active cwd as the single implicit project root. A follow-up should model project roots separately from tool cwd so `:project_roots` entries can resolve against the configured project roots, and resolve to no entries when there are no project roots.

## Verification

- `cargo test -p codex-protocol permissions:: --lib`
- `cargo test -p codex-sandboxing -p codex-exec-server --lib`
- `cargo test -p codex-core session_configuration_apply_ --lib`
- `cargo test -p codex-app-server command_exec_permission_profile_project_roots_use_command_cwd --test all`
- `cargo test -p codex-tui thread_read_session_state_does_not_reuse_primary_permission_profile --lib`
- `cargo test -p codex-tui preset_matching_accepts_workspace_write_with_extra_roots --lib`
- `cargo test -p codex-config --lib`


codex-rs/protocol/src/permissions.rs
```diff
@@ -80,7 +80,6 @@ impl FileSystemAccessMode {
 pub enum FileSystemSpecialPath {
     Root,
     Minimal,
-    CurrentWorkingDirectory,
     ProjectRoots {
```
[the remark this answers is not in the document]
bolinfest: [codex] Fixed in d85ef7c by accepting the legacy `current_working_directory` tag as a serde alias for `project_roots` at both the core and app-server protocol layers, with regression tests for the persisted JSON shape.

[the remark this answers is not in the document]
bolinfest: [codex] Update: removed that compatibility alias again in 876bda9. This `permissionProfile` shape is still unstable API, so the old `current_working_directory` tag is intentionally dropped rather than preserved as backwards-compat surface.

viyatb-oai, line 84: This is fine as the new canonical shape, but removing the old variant outright also removes the only deserialization path for profiles already persisted with `{"kind":"current_working_directory"}`. Base `main` still emits that tag from `PermissionProfile::workspace_write()`, and every persisted `TurnContextItem` stores `permission_profile: Some(...)`; on this head, `RolloutRecorder::load_rollout_items()` can no longer deserialize those lines and drops them as parse errors during resume/fork, so existing threads can lose their saved turn-context baseline after upgrade.

A small compatibility-only fix is to keep accepting the old tag while only serializing the new one:

```rust
    #[serde(alias = "current_working_directory")]
    ProjectRoots {
        #[ts(optional)]
        subpath: Option<PathBuf>,
    },
```

Please also add a regression test that deserializes `{"kind":"current_working_directory"}` to `ProjectRoots { subpath: None }` and reserializes it as `project_roots`.

viyatb-oai (commented):
Confirmed this against persisted rollout behavior on the current head.

codex-rs/protocol/src/permissions.rs
bolinfest: [codex] Acknowledged, but we are intentionally not preserving the old `current_working_directory` tag in this PR. `permissionProfile` is still unstable surface area, and the goal here is to fully remove the ambiguous cwd special path rather than retain it as a deserialize-only alias. The PR body now calls out the remaining follow-up: model project roots separately from tool cwd so `:project_roots` no longer implicitly means the active cwd.

viyatb-oai (approved):

codex-rs/protocol/src/permissions.rs
bolinfest: [codex] Addressed in f7d5c241cd8c by adding a deserialize-only `#[serde(alias = "current_working_directory")]` on `ProjectRoots` in both the core protocol and v2 app-server protocol mirrors. I also added regression coverage that decodes the old tag to `ProjectRoots { subpath: None }` and reserializes it using the canonical `project_roots` tag.

PR #19846, opened 2026-04-27
[sandbox] Enforce protected workspace metadata paths

evawong-oai:
## Summary

Make FileSystemSandboxPolicy the semantic source of truth for project root metadata protection. Under writable roots, `.git`, `.codex`, and `.agents` stay protected unless user policy grants an explicit write rule for that metadata path.

## Scope

1. Add `protected_metadata_names` to `WritableRoot`.
2. Teach `FileSystemSandboxPolicy::can_write_path_with_cwd` to reject protected metadata writes under writable roots unless explicitly allowed.
3. Default workspace write profiles to protect `.git`, `.codex`, and `.agents`.
4. Add the Linux fallback setup needed before Linux enforcement lands later in the stack.

## Reviewer Focus

1. The policy decision belongs in FileSystemSandboxPolicy, not shell command parsing.
2. Legacy SandboxPolicy remains a compatibility projection, not the source of the new rule.
3. Explicit user write rules can still opt into these metadata paths.

## Stack

1. Policy primitive: this PR
2. macOS Seatbelt adapter: #19847
3. Shell preflight UX: #19848
4. Runtime profile propagation: #19849
5. Linux bubblewrap adapter: #19852

## Validation

1. codex protocol permissions tests
2. formatting for codex protocol and codex linux sandbox
3. diff whitespace check

### DIFF
--- a/codex-rs/linux-sandbox/src/bwrap.rs
+++ b/codex-rs/linux-sandbox/src/bwrap.rs
@@ -259,6 +259,7 @@ fn create_filesystem_args(
         writable_roots.push(WritableRoot {
             root: AbsolutePathBuf::from_absolute_path("/")?,
             read_only_subpaths: Vec::new(),
+            protected_metadata_names: Vec::new(),
         });
     }
     let missing_auto_metadata_read_only_project_root_subpaths: HashSet<PathBuf> =
--- a/codex-rs/protocol/src/permissions.rs
+++ b/codex-rs/protocol/src/permissions.rs
@@ -19,6 +19,62 @@ use crate::protocol::NetworkAccess;
 use crate::protocol::SandboxPolicy;
 use crate::protocol::WritableRoot;
 
+const PROTECTED_METADATA_GIT_PATH_NAME: &str = ".git";
+const PROTECTED_METADATA_AGENTS_PATH_NAME: &str = ".agents";
+const PROTECTED_METADATA_CODEX_PATH_NAME: &str = ".codex";
+
+/// Top-level workspace metadata paths that stay protected under writable roots.
+pub const PROTECTED_METADATA_PATH_NAMES: &[&str] = &[
+    PROTECTED_METADATA_GIT_PATH_NAME,
+    PROTECTED_METADATA_AGENTS_PATH_NAME,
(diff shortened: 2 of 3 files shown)


evawong-oai:
Fixed in 3ed97f5caa.

The active contract is now represented on `WritableRoot.protected_metadata_names`. `get_writable_roots_with_cwd()` fills that field for `.git`, `.agents`, and `.codex`, and explicit writable rules for those paths remove the matching name from the parent root.

`needs_direct_runtime_enforcement()` now checks that field before the legacy semantic comparison. If legacy `SandboxPolicy` cannot express the metadata protection contract, we keep the direct enforcement path. That removes the hidden dependency on `can_write_path_with_cwd()` for the bridge decision.

The stale inline nested repo concern is also covered by the current contract. This only protects the first component under each writable root, so normal nested work below a regular directory is not blocked by this field.

Validation passed locally, and the devbox 46 case suite passed with `just c sandbox linux` at `8cf759200dd2355df3a76fbfcb0dc9f8619f33f5`.

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -259,6 +259,7 @@ fn create_filesystem_args(
         writable_roots.push(WritableRoot {
             root: AbsolutePathBuf::from_absolute_path("/")?,
             read_only_subpaths: Vec::new(),
+            protected_metadata_names: Vec::new(),
```
viyatb-oai, line 262: nit: `protected_metadata_dirs`?

viyatb-oai (approved):
can you preserve comments from the old code?

codex-rs/linux-sandbox/src/bwrap.rs
evawong-oai: Keeping `protected_metadata_dirs` to match Michael

evawong-oai: Clarifying my note above: this keeps `protected_metadata_dirs` aligned with Michael terminology around metadata directories so the stack reads consistently with the related sandbox changes. It is still scoped to protected metadata, but this call site builds directory masks, so the name is easier to read here.

PR #19848, opened 2026-04-27
Add workspace metadata shell preflight

evawong-oai:
## Summary

Add an early user facing check that explains simple literal shell redirections into protected project root metadata. The check uses FileSystemSandboxPolicy for the protected metadata names and message.

## Scope

1. Add a shared shell command helper for literal protected metadata redirection targets.
2. Use that helper in debug sandbox, legacy shell, and UnifiedExec paths.
3. Use the effective FileSystemSandboxPolicy after additional permissions are merged.
4. Keep direct command creation attempts such as mkdir .codex enforced by FileSystemSandboxPolicy and the platform sandbox adapters instead of duplicating that policy in command preflight.

## Reviewer Focus

1. This PR improves the error message for simple parseable shell redirections such as printf data into .codex config.
2. It does not weaken .git, .agents, or .codex protections. Security remains in FileSystemSandboxPolicy and the platform sandbox adapters.
3. Command preflight is intentionally not a second security policy. Direct command attempts such as mkdir .codex should still be denied by sandbox enforcement rather than by a command name deny list.

## Stack

1. Policy primitive: #19846
2. macOS Seatbelt adapter: #19847
3. Shell preflight UX: this PR
4. Runtime profile propagation: #19849
5. Linux bubblewrap adapter: #19852

## Validation

1. targeted core protected metadata preflight tests
2. targeted shell command metadata detector tests
3. Rust format check
4. diff whitespace check

### DIFF
--- a/codex-rs/cli/src/debug_sandbox.rs
+++ b/codex-rs/cli/src/debug_sandbox.rs
@@ -35,6 +35,10 @@ use crate::exit_status::handle_exit_status;
 #[cfg(target_os = "macos")]
 use seatbelt::DenialLogger;
 
+#[cfg(target_os = "linux")]
+const LINUX_SANDBOX_FORWARDED_SIGNALS: &[libc::c_int] =
+    &[libc::SIGHUP, libc::SIGINT, libc::SIGQUIT, libc::SIGTERM];
+
 #[cfg(target_os = "macos")]
 pub async fn run_command_under_seatbelt(
     command: SeatbeltCommand,
@@ -142,6 +146,13 @@ async fn run_command_under_sandbox(
     // sandbox policy. In the future, we could add a CLI option to set them
     // separately.
     let sandbox_policy_cwd = cwd.clone();
+    if let Some(reason) = codex_shell_command::metadata_write_forbidden_reason(
+        &command,
+        cwd.as_path(),
+        &config.permissions.file_system_sandbox_policy(),
+    ) {
+        anyhow::bail!("{reason}");
+    }
 
     let env = create_env(
         &config.permissions.shell_environment_policy,
@@ -261,6 +272,9 @@ async fn run_command_under_sandbox(
         denial_logger.on_child_spawn(&child);
     }
 
+    #[cfg(target_os = "linux")]
+    let status = wait_for_debug_sandbox_child(&mut child).await?;
+    #[cfg(not(target_os = "linux"))]
     let status = child.wait().await?;
 
     #[cfg(target_os = "macos")]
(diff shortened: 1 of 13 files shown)


codex-rs/core/src/tools/handlers/shell.rs
```diff
@@ -531,6 +531,14 @@ impl ShellHandler {
                 prefix_rule,
             })
             .await;
+        let exec_approval_requirement = codex_shell_command::metadata_write_forbidden_reason(
+            &exec_params.command,
+            &exec_params.cwd,
+            &file_system_sandbox_policy,
+        )
+        .map_or(exec_approval_requirement, |reason| {
+            crate::tools::sandboxing::ExecApprovalRequirement::Forbidden { reason }
```
viyatb-oai: [P1] Preflight misses the default UnifiedExec path

This preflight only wraps the legacy ShellHandler path. UnifiedExec is stable/default on non-Windows, and UnifiedExecProcessManager creates its own ExecApprovalRequirement without calling `metadata_write_forbidden_reason`, so default `exec_command` can still run `printf pwned > .git` or write `.codex/config` without the intended preflight.

Suggested fix: share the metadata-write preflight with UnifiedExec, or call it from `UnifiedExecProcessManager::open_session_with_sandbox` before constructing `UnifiedExecToolRequest`, using the same effective policy and cwd inputs that the runtime will enforce.

codex-rs/core/src/tools/handlers/shell.rs
```diff
@@ -531,6 +531,14 @@ impl ShellHandler {
                 prefix_rule,
             })
             .await;
+        let exec_approval_requirement = codex_shell_command::metadata_write_forbidden_reason(
+            &exec_params.command,
+            &exec_params.cwd,
+            &file_system_sandbox_policy,
+        )
```
viyatb-oai: [P2] Preflight uses the wrong policy view for workdirs and grants

This passes `exec_params.cwd` and the base turn file-system policy into `forbidden_agent_metadata_write`. That helper uses the cwd both to resolve the redirection target and to resolve special policy roots, while the runtime resolves sandbox policy against `turn.cwd` and then merges additional permissions. With `workdir: subdir`, `printf x > ../.codex/config.toml` can miss the project-root metadata preflight, and an approved explicit `.codex` additional permission is still denied because it is absent from the base policy.

Suggested fix: split target resolution from policy-root resolution. Resolve the redirection target against the command cwd, evaluate writable-root metadata relative to the sandbox policy cwd (`turn.cwd`), and call the helper with the effective file-system policy after additional permissions are merged.

codex-rs/cli/src/debug_sandbox.rs
```diff
@@ -438,13 +452,96 @@ async fn spawn_debug_sandbox_child(
...
+    {
+        let parent_pid = unsafe { libc::getpid() };
+        // SAFETY: `pre_exec` runs in the child immediately before exec. The
+        // closure only adjusts the child signal mask, installs a parent-death
+        // signal, and checks the inherited parent pid to close the fork/exec
+        // race.
+        unsafe {
+            cmd.pre_exec(move || {
+                block_linux_sandbox_forwarded_signals()?;
+                if libc::prctl(libc::PR_SET_PDEATHSIG, libc::SIGTERM) == -1 {
```
viyatb-oai: [P2] Forwarded signals are blocked in the Linux child

This `pre_exec` path blocks SIGHUP/SIGINT/SIGQUIT/SIGTERM in the child and never unblocks them before `exec`. Signal masks survive exec, so the spawned `codex-linux-sandbox` and payload inherit those signals blocked; when the parent later forwards Ctrl-C/SIGTERM with `kill()`, the child keeps waiting instead of terminating, and the `PR_SET_PDEATHSIG` SIGTERM is also blocked.

Suggested fix: avoid changing the child signal mask here, or block signals only in the parent and explicitly restore/unblock the child mask before exec. If forwarding needs to cover descendants, prefer process-group signaling while keeping SIGTERM/SIGINT deliverable.

codex-rs/core/src/tools/sandboxing_tests.rs
```diff
@@ -138,3 +139,112 @@ fn guardian_bypasses_sandbox_for_explicit_escalation_on_first_attempt() {
...
+}
+
+#[test]
+fn metadata_write_preflight_forbids_only_sandboxed_skip() {
+    let repo = tempdir().expect("create tempdir");
+    std::fs::create_dir(repo.path().join(".git")).expect("create .git");
+    let file_system_sandbox_policy = FileSystemSandboxPolicy::from_legacy_sandbox_policy_for_cwd(
+        &SandboxPolicy::new_workspace_write_policy(),
+        repo.path(),
+    );
```
bolinfest: Please don't introduce new uses of `from_legacy_sandbox_policy_for_cwd`.

codex-rs/core/src/tools/sandboxing.rs
```diff
@@ -194,6 +195,50 @@ impl ExecApprovalRequirement {
     }
 }
 
+/// Applies metadata-write preflight only to commands that would otherwise run in
```
bolinfest: What does "metadata-write preflight" mean?

codex-rs/core/src/tools/sandboxing_tests.rs
evawong-oai: Updated in the latest push. The new tests now construct FileSystemSandboxPolicy directly with workspace_write, so this PR no longer adds any calls to from_legacy_sandbox_policy_for_cwd.

codex-rs/core/src/tools/sandboxing.rs
evawong-oai: Agreed, the old wording was too dense. I renamed the helper to apply_protected_metadata_write_preflight and updated the comment to say this is an early user facing check for protected metadata writes before sandboxed execution. The security boundary remains FileSystemSandboxPolicy and the platform sandbox adapters.

codex-rs/core/src/tools/handlers/shell.rs
evawong-oai: Addressed. The helper now separates command cwd from sandbox policy cwd, and the handlers pass the effective FileSystemSandboxPolicy after additional permissions are merged.

codex-rs/core/src/tools/handlers/shell.rs
evawong-oai: Addressed. UnifiedExec now calls the shared apply_protected_metadata_write_preflight path before it builds the ExecCommandRequest.

codex-rs/cli/src/debug_sandbox.rs
evawong-oai: Addressed. The debug sandbox child path no longer mutates the Unix signal mask before exec.

PR #19852, opened 2026-04-27
Enforce workspace metadata protections in Linux sandbox

evawong-oai:
## Summary

Enforce FileSystemSandboxPolicy protected metadata names in the Linux bubblewrap adapter so `.git`, `.agents`, and `.codex` remain read only inside writable workspace roots unless the policy grants an explicit write carveout.

## Scope

1. Translate protected metadata names from FileSystemSandboxPolicy into bubblewrap masks for existing metadata paths.
2. Represent missing protected metadata paths as guarded mount targets so agents cannot create `.git`, `.agents`, or `.codex` under writable roots.
3. Preserve normal git discovery for existing repos, worktrees, and parent repos.
4. Keep explicit user write grants working when policy allows a protected metadata path directly.

## Not in scope

1. No shell preflight UX.
2. No TUI runtime profile propagation.
3. No macOS Seatbelt changes in this PR.

## Reviewer focus

1. This should be reviewed as the Linux enforcement adapter for the policy primitive from PR 19846.
2. macOS enforcement already landed in PR 19847.
3. The important invariant is that `FileSystemSandboxPolicy` is the source of truth for `.git`, `.agents`, and `.codex`.

## Validation

1. `git diff` whitespace check passed.
2. `cargo fmt` check passed with the existing stable rustfmt warning about `imports_granularity`.
3. Full Linux sandbox Cargo test suite passed on the devbox.
4. Devbox forty six case suite passed at head `012accb703c13bd28df5b40079a9bf183036336a`.
5. Devbox summary: pass 46, fail 0.
6. The devbox suite was run through `just c sandbox linux`.
7. Focused repo test for Viyat parent repo case passed on the devbox.

### DIFF
--- a/codex-rs/linux-sandbox/src/bwrap.rs
+++ b/codex-rs/linux-sandbox/src/bwrap.rs
@@ -579,6 +579,7 @@ fn create_filesystem_args(
         append_metadata_path_masks_for_writable_root(
             &mut read_only_subpaths,
             root,
+            mount_root,
             &protected_metadata_names,
         );
         if let Some(target) = &symlink_target {
@@ -654,11 +655,12 @@ fn append_protected_create_targets_for_writable_root(
 fn append_metadata_path_masks_for_writable_root(
     read_only_subpaths: &mut Vec<PathBuf>,
     root: &Path,
+    mount_root: &Path,
     protected_metadata_names: &[String],
 ) {
     for name in protected_metadata_names {
         let path = root.join(name);
-        if should_leave_missing_git_for_parent_repo_discovery(root, name, &path) {
+        if should_leave_missing_git_for_parent_repo_discovery(mount_root, name) {
             continue;
         }
         if !read_only_subpaths.iter().any(|subpath| subpath == &path) {
@@ -667,17 +669,17 @@ fn append_metadata_path_masks_for_writable_root(
     }
 }
 
-fn should_leave_missing_git_for_parent_repo_discovery(
-    root: &Path,
-    name: &str,
-    path: &Path,
-) -> bool {
+fn should_leave_missing_git_for_parent_repo_discovery(mount_root: &Path, name: &str) -> bool {
+    let path = mount_root.join(name);
     name == ".git"
         && matches!(
             path.symlink_metadata(),
             Err(err) if err.kind() == io::ErrorKind::NotFound
         )
-        && root.ancestors().skip(1).any(ancestor_has_git_metadata)
+        && mount_root
+            .ancestors()
+            .skip(1)
+            .any(ancestor_has_git_metadata)
 }
 
 fn ancestor_has_git_metadata(ancestor: &Path) -> bool {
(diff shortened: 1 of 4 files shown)


codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -838,28 +1024,107 @@ fn append_read_only_subpath_args(
...
+    bwrap_args.args.push(path_to_string(path));
+}
+
+fn append_missing_read_only_subpath_args(bwrap_args: &mut BwrapArgs, path: &Path) -> Result<()> {
+    if path.file_name().is_some_and(is_protected_metadata_name) {
+        append_empty_directory_args(bwrap_args, path);
+        bwrap_args
+            .synthetic_mount_targets
+            .push(SyntheticMountTarget::missing_empty_directory(path));
+        return Ok(());
```
viyatb-oai, line 1100: [P1] Missing `.git` directory mask breaks parent repo discovery

This branch handles every missing protected metadata name, including `.git`, by calling `append_empty_directory_args`. When Codex runs from a subdirectory of a Git repo, that materializes `<cwd>/.git` in the sandbox; Git stops there instead of walking up to the parent `.git`, so normal commands like `git rev-parse --show-toplevel` and `git status` regress. The test below locks in that broken behavior. Keep missing `.agents` and `.codex` as empty directory masks, but do not materialize missing child `.git` when an ancestor repo exists; use the protected-create cleanup/preflight path for `.git` creation instead.

evawong-oai: Fixed. Missing child `.git` under an ancestor repo now stays out of the bubblewrap read only empty dir masks, so Git can keep walking to the parent repo. `.agents` and `.codex` still use the empty dir masks. The missing child `.git` path is registered as a protected create target, and Linux now watches the parent directory while the sandboxed command is running so `git init` fails during the command instead of only during cleanup.

Validation:
1. Devbox forty six case suite passed at head `74981ffa802e105a53f35ad11ecded077816ff9d`.
2. Targeted Linux unit test passed for `missing_child_git_under_parent_repo_uses_protected_create_target`.

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -481,12 +1319,12 @@ fn run_bwrap_in_child_capture_stderr(bwrap_args: crate::bwrap::BwrapArgs) -> Str
...
     }
 
-    let mut status: libc::c_int = 0;
-    let wait_res = unsafe { libc::waitpid(pid, &mut status as *mut libc::c_int, 0) };
-    if wait_res < 0 {
-        let err = std::io::Error::last_os_error();
-        panic!("waitpid failed for bubblewrap child: {err}");
-    }
+    wait_for_bwrap_child(pid);
+    let cleanup_signal_mask = ForwardedSignalMask::block();
```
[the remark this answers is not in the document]
evawong-oai: Fixed in `512f76c278`. The preflight path now keeps the waited status, restores the temporary signal forwarders before returning, and exits with the child signal status when the probe is terminated. That keeps cancel and timeout behavior from continuing into the real sandbox launch.

codex-rs/linux-sandbox/src/linux_run_main.rs
```diff
@@ -467,9 +1303,11 @@ fn run_bwrap_in_child_capture_stderr(bwrap_args: crate::bwrap::BwrapArgs) -> Str
             close_fd_or_panic(write_fd, "close write end in bubblewrap child");
         }
 
-        exec_bwrap(bwrap_args.args, bwrap_args.preserved_files);
+        exec_bwrap(args, preserved_files);
     }
 
+    install_bwrap_signal_forwarders(pid);
+    setup_signal_mask.restore();
```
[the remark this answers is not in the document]
evawong-oai: Fixed in `512f76c278`. The signal forwarder install now returns the prior handlers, and both the real launch path and the preflight path restore them after the child exits and before signals are unblocked. That removes the window where later signals could be handled by stale bwrap forwarding state.

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -484,18 +627,95 @@ fn create_filesystem_args(
...
+    cwd: &Path,
+) -> Vec<String> {
+    let mut names = writable_root.protected_metadata_names.clone();
+    for name in PROTECTED_METADATA_PATH_NAMES {
+        if names.iter().any(|existing| existing == name) {
+            continue;
+        }
+        let path = writable_root.root.join(*name);
+        if !file_system_sandbox_policy.can_write_path_with_cwd(path.as_path(), cwd) {
+            names.push((*name).to_string());
```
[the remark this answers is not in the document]
evawong-oai: Fixed in `512f76c278`. Linux bwrap now uses `WritableRoot.protected_metadata_names` directly from `FileSystemSandboxPolicy`; it no longer derives extra protected metadata names locally with `can_write_path_with_cwd`. That keeps symlink alias root decisions owned by the policy layer.

codex-rs/linux-sandbox/src/bwrap.rs
```diff
@@ -484,18 +622,77 @@ fn create_filesystem_args(
...
+}
+
+fn append_metadata_path_masks_for_writable_root(
+    read_only_subpaths: &mut Vec<PathBuf>,
+    root: &Path,
+    protected_metadata_names: &[String],
+) {
+    for name in protected_metadata_names {
+        let path = root.join(name);
+        if should_leave_missing_git_for_parent_repo_discovery(root, name, &path) {
```
viyatb-oai: [P2] Symlinked workspace still masks parent repo `.git`

`append_metadata_path_masks_for_writable_root` decides whether to skip the child `.git` mask from the logical `root`, but symlinked writable roots are rebound to their real target and the command cwd is canonicalized before exec. For a workspace like `/tmp/link/sub -> /repo/sub`, this check sees no ancestor repo under `/tmp/link` and adds the mask; the later remap turns it into `/repo/sub/.git`, so Git inside the sandbox still stops before reaching `/repo/.git`.

Suggested fix: evaluate the parent-repo exception against the effective mount root too, e.g. thread `symlink_target.as_deref().unwrap_or(root)` into the `.git` skip decision (or perform the skip after remapping), then add a regression test that combines the existing symlinked writable-root case with `sandbox_keeps_parent_repo_discovery_while_blocking_child_metadata`.

evawong-oai: Fixed in `4a98b3f914`.

The `.git` parent repo exception now evaluates against the effective bwrap mount root, `symlink_target.as_deref().unwrap_or(root)`, while the mask path still goes through the existing logical path remap. That means a symlinked writable root like `link_repo/sub` pointing at `repo/sub` sees the real parent repo before deciding whether to skip the missing child `.git` mask.

Added coverage:
1. `symlinked_missing_child_git_under_parent_repo_uses_effective_mount_root`
2. Devbox case 47, `Symlinked Parent Repo Subdir Keeps Git Root And Allows Normal Coding`

Validation:
1. Linux sandbox crate tests passed on evawong 1: 80 unit tests and 25 integration tests.
2. The forty seven case `just c` suite passed on evawong 1 at `4a98b3f91499b50c196ff02cf99e6461bb5bfd89`.

viyatb-oai (approved):

evawong-oai:
Thanks. I am keeping this PR scoped to the Linux sandbox protection fix. A follow up cleanup PR will modularize the bwrap code so this logic is easier to review and maintain.

PR #19859, opened 2026-04-27
Show plugin hooks in plugin details

abhinav-oai:
# Why

PR #19705 adds discovery for hooks bundled with plugins, but `/plugins` still only shows skills, apps, and MCP servers. This follow-up makes bundled hooks visible in the same plugin detail view so users can inspect the full plugin surface in one place.

We'll also need this `PluginHookSummary` to populate the Plugin Hooks in the App. We can't use `hooks/list` there  because we need to be able to show the hooks for disabled plugins as well.

# What

- extend `plugin/read` with `PluginHookSummary` entries containing each hook event name and handler count
- summarize bundled plugin hooks while loading plugin details
- render a `Hooks` row in the `/plugins` detail popup

CleanShot 2026-04-27 at 11 45 34@2x


codex-rs/core/src/plugins/manager.rs
```diff
@@ -1469,6 +1486,39 @@ impl PluginsManager {
     }
 }
 
+fn summarize_plugin_hooks(hook_sources: &[PluginHookSource]) -> Vec<PluginHookSummary> {
```
abhinav-oai: can we do this in a cleaner way that will automatically do all possible hooks instead of hardcoding?

codex-rs/core/src/plugins/manager.rs
```diff
@@ -172,11 +175,18 @@ pub struct PluginDetail {
...
     pub apps: Vec<AppConnectorId>,
     pub mcp_server_names: Vec<String>,
     pub details_unavailable_reason: Option<PluginDetailsUnavailableReason>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PluginHookSummary {
+    pub event_name: HookEventName,
+    pub handler_count: usize,
+}
```
abhinav-oai, line 196: will expand with more metadata in the future

abhinav-oai: I think the follow-up shape should be per-handler rather than per-event:

```rust
PluginHookSummary {
    key: String,
    event_name: HookEventName,
    matcher: Option<String>,
    enabled: bool,
    status_message: Option<String>,
    definition: serde_json::Value,
    display_order: i64,
}
```

That is the smallest shape the app detail page needs to render the hook rows directly: `event_name` for the title, `matcher` + optional `status_message` for the compact description, `enabled` for the toggle state, `definition` for the JSON details modal, `display_order` so clients preserve backend/authored order, and `key` as the opaque handle for toggling one specific handler.

abhinav-oai: then from this we can derive the event+count we need for the TUI `/plugins`

abhinav-oai:
@codex review

xl-openai (approved):

PR #19863, opened 2026-04-27
[codex-analytics] remove ga flag

rhan-oai:


owenlin0 (approved):
config-loading is tolerant if someone still has:
```
[features]
general_analytics = false/true
```

in their config.toml, right? I assume so but just wanted to double check

PR #19878, opened 2026-04-27
Ingest node_repl stderr telemetry spans

charley-openai:
## Why

So that we can debug latency in long node_repl spans with function specific breakdowns.

[Example trace](https://openai.datadoghq.com/apm/trace/0936629243861388d65ba1a16777753a?graphType=waterfall&shouldShowLegend=true&spanID=7820977976046608013&traceQuery=) 
Screenshot 2026-04-28 at 12 46 58 PM

Note: this change should go in alongside the [browser-use plugin change](https://github.com/openai/openai/pull/864367)

## Summary
- add node_repl/js MCP _meta trace context for stderr span telemetry
- parse @codex-telemetry stderr lines from local stdio MCP servers and route node_repl telemetry into OTel
- reconstruct sanitized Browser Use/node_repl spans with explicit trace/span/parent ids so emitted child spans nest under node_repl.js
- keep traceparent fallback for older records and add js_repl tool-call span coverage

## Tests
- cargo test -p codex-otel stderr_span_telemetry
- cargo test -p codex-rmcp-client stderr_telemetry_parser
- cargo test -p codex-core mcp_tool_call
- cargo test -p codex-core otel_stderr_spans
- cargo test -p codex-mcp
- just fix -p codex-otel
- just fix -p codex-rmcp-client -p codex-mcp -p codex-core
- just fmt
- git diff --check

### DIFF
--- a/codex-rs/codex-mcp/src/rmcp_client.rs
+++ b/codex-rs/codex-mcp/src/rmcp_client.rs
@@ -7,7 +7,9 @@
 //! [`crate::connection_manager`].
 
 use std::borrow::Cow;
+use std::collections::BTreeMap;
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::env;
 use std::ffi::OsString;
 use std::sync::Arc;
@@ -50,6 +52,8 @@ use codex_rmcp_client::ExecutorStdioServerLauncher;
 use codex_rmcp_client::LocalStdioServerLauncher;
 use codex_rmcp_client::RmcpClient;
 use codex_rmcp_client::StdioServerLauncher;
+use codex_rmcp_client::StdioServerTelemetry;
+use codex_rmcp_client::StdioServerTelemetrySink;
 use futures::future::BoxFuture;
 use futures::future::FutureExt;
 use futures::future::Shared;
@@ -57,6 +61,7 @@ use rmcp::model::ClientCapabilities;
 use rmcp::model::ElicitationCapability;
 use rmcp::model::Implementation;
 use rmcp::model::InitializeRequestParams;
+use rmcp::model::JsonObject;
 use rmcp::model::ProtocolVersion;
 use rmcp::model::Tool as RmcpTool;
 use tokio_util::sync::CancellationToken;
@@ -65,6 +70,12 @@ use tracing::warn;
 /// MCP server capability indicating that Codex should include [`SandboxState`]
 /// in tool-call request `_meta` under this key.
 pub const MCP_SANDBOX_STATE_META_CAPABILITY: &str = "codex/sandbox-state-meta";
+/// MCP server capability indicating that Codex should include W3C trace context
+/// in tool-call request `_meta` so the server can emit reconstructed child spans.
+/// See `codex-rs/docs/mcp_subspan_tracing.md` for the protocol contract.
+pub const MCP_SUBSPAN_TRACING_CAPABILITY: &str = "codex/subspan-tracing";
+pub const MCP_SUBSPAN_TRACING_TRANSPORT_STDERR_JSONL: &str = "stderr-jsonl";
+pub const MCP_SUBSPAN_TRACING_VERSION: u64 = 1;
 
 pub(crate) const MCP_TOOLS_LIST_DURATION_METRIC: &str = "codex.mcp.tools.list.duration_ms";
 pub(crate) const MCP_TOOLS_FETCH_UNCACHED_DURATION_METRIC: &str =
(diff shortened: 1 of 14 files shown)


codex-rs/codex-mcp/src/rmcp_client.rs
```diff
@@ -72,6 +74,7 @@ pub(crate) const MCP_TOOLS_FETCH_UNCACHED_DURATION_METRIC: &str =
     "codex.mcp.tools.fetch_uncached.duration_ms";
 pub(crate) const DEFAULT_STARTUP_TIMEOUT: Duration = Duration::from_secs(30);
 pub(crate) const DEFAULT_TOOL_TIMEOUT: Duration = Duration::from_secs(120);
+const NODE_REPL_MCP_SERVER_NAME: &str = "node_repl";
```
aaronl-openai: I don't think we can build this kind of knowledge of `node_repl` into codex core. I think there are a couple approaches we could take instead:
- in the host, we can read the `otel.*` config and pass it through via an env var. then we could have node_repl read that and directly emit the metrics
- we could loosen the restrictions that are making this be node_repl specific. a better way to model this might be through an experimental capability like sandbox. I also wonder if instead of using stderr to communicate span data out of band, we could return all of the spans in the _meta of the tool call response. that feels a bit cleaner: an experimental tracing capability that both provides and receives tracing data via _meta.

charley-openai: I agree, I switched the approach here to a "mcp server capability" contract instead.

codex-rs/otel/src/stderr_span_telemetry.rs
```diff
@@ -0,0 +1,756 @@
...
+    "browser_use.playwright.dom_snapshot",
+    "browser_use.tab.goto",
+    "browser_use.tab.click",
+    "browser_use.tab.type",
+    "browser_use.tab.screenshot",
+    "browser_use.cdp.execute",
+    "browser_use.tab.wait_for_load_state",
+];
+
+const ALLOWED_ATTRIBUTE_PREFIXES: &[&str] = &["browser_use.", "node_repl.", "js."];
```
jif-oai, line 35: This isn’t really the browser-use-v1 allowlist. The paired emitter sends keys like tool.name, browser.action, url.origin, tab.id, timeout_ms, and error.kind, which this side silently drops, while any future browser_use.* / js.* key gets through. Can we use the actual negotiated profile instead of a prefix approximation?

codex-rs/otel/src/stderr_span_telemetry.rs
```diff
@@ -0,0 +1,756 @@
...
+    name: String,
+    span_id: Option<SpanId>,
+    trace_id: Option<TraceId>,
+    parent_span_id: Option<SpanId>,
+    trace_flags: Option<TraceFlags>,
+    traceparent: Option<String>,
+    tracestate: Option<String>,
+    start_time: SystemTime,
+    end_time: SystemTime,
+    attributes: Vec<KeyValue>,
```
jif-oai, line 63: (found by Codex)
The paired emitter sends a top-level status, but we never parse or apply it here, so failed child spans lost their error status on ingest

codex-rs/codex-mcp/src/rmcp_client.rs
```diff
@@ -492,6 +570,13 @@ async fn start_server_task(
         .as_ref()
         .and_then(|exp| exp.get(MCP_SANDBOX_STATE_META_CAPABILITY))
         .is_some();
+    let subspan_tracing_capability = initialize_result
+        .capabilities
+        .experimental
+        .as_ref()
+        .and_then(|exp| exp.get(MCP_SUBSPAN_TRACING_CAPABILITY))
+        .and_then(McpSubspanTracingCapability::from_experimental_capability);
+    subspan_tracing_enabled.store(subspan_tracing_capability.is_some(), Ordering::Release);
```
jif-oai, line 579: Negotiation is tool-scoped, but ingestion becomes server-scoped here... that feels pretty odd

codex-rs/codex-mcp/src/rmcp_client.rs
```diff
@@ -637,11 +752,40 @@ async fn make_rmcp_client(
...
+    server_name: &str,
+    enabled: Arc<AtomicBool>,
+) -> StdioServerTelemetrySink {
+    let server_name = server_name.to_string();
+    Arc::new(move |telemetry: StdioServerTelemetry| {
+        if !enabled.load(Ordering::Acquire) {
+            return;
+        }
+
+        if let Err(error) = codex_otel::emit_mcp_subspan_telemetry(telemetry.payload) {
```
jif-oai, line 765: Can we bind these records back to the active call we advertised? Right now any opted-in server can print arbitrary trace/span ids on stderr and we’ll ingest them, so a buggy or custom or any change in server can attach spans to unrelated traces

codex-rs/rmcp-client/src/stdio_server_launcher.rs
```diff
@@ -578,6 +591,38 @@ impl ExecutorStdioServerLauncher {
...
+        return;
+    };
+
+    let Some(telemetry) = parse_stderr_telemetry_line(line) else {
+        info!("MCP server stderr ({program_name}): {line}");
+        return;
+    };
+
+    match telemetry {
+        Ok(telemetry) => sink(telemetry),
```
jif-oai, line 612: This sink runs inline... this is dangerous under high load

codex-rs/codex-mcp/src/rmcp_client.rs
```diff
@@ -88,9 +99,52 @@ pub(crate) struct ManagedClient {
...
+                })
+            });
+        if !supports_stderr_jsonl {
+            return None;
+        }
+
+        let tools = capability
+            .get("tools")
+            .and_then(serde_json::Value::as_object)
+            .map(|tools| tools.keys().cloned().collect());
```
jif-oai, line 136: We parse the per-tool capability object and then throw away everything except the tool names, including `attributeProfile`. So the API looks generic, but the host still hardcodes Browser Use / node_repl policy  elsewhere instead of actually negotiating the profile. This sounds strange and it seem to break the genericity that we're trying to have here

codex-rs/otel/src/stderr_span_telemetry.rs
```diff
@@ -0,0 +1,756 @@
...
+    "browser_use.tab.goto",
+    "browser_use.tab.click",
+    "browser_use.tab.type",
+    "browser_use.tab.screenshot",
+    "browser_use.cdp.execute",
+    "browser_use.tab.wait_for_load_state",
+];
+
+const ALLOWED_ATTRIBUTE_PREFIXES: &[&str] = &["browser_use.", "node_repl.", "js."];
+const ALLOWED_ATTRIBUTE_KEYS: &[&str] = &["error.type", "error.message"];
```
jif-oai, line 36: Also, why are we admitting arbitrary error.message here?

jif-oai (changes requested):
I feel this PR is trying to present itself as a generic MCP tracing extension while still being a bespoke Browser Use pipeline...
The paired emitter and host disagree on attrs/status, negotiated scope gets discarded on ingest, and the supposedly out-of-band path is now on transport-critical loops
Maybe I'm missing something here but I feel we should choose one direction and stick to it

charley-openai:
closing for now while i'm out on pat leave.

PR #19895, opened 2026-04-28
External agent session support

stefanstokic-oai:
## Summary

This extends external agent detection/import beyond config artifacts so Codex can detect recent sessions files from the external agent home and import them into Codex rollout history.

## What changed

- Added a focused `external_agent_sessions` module for:
  - session discovery
  - source-record parsing
  - rollout construction
  - import ledger tracking
- Wired session detection/import into the app-server external agent config API.
- Added compaction handling so large imported sessions can be resumed safely before the first follow-up turn.

## Testing

Added coverage for:
- recent-session detection
- custom-title handling
- recency filtering
- dedupe and re-detect-after-source-change behavior
- visible imported turn construction
- backward-compatible import payload deserialization
- end-to-end RPC import flow
- rejection of undetected session paths
- repeat-import behavior
- large-session compaction before first follow-up

Ran:
- `cargo test -p codex-app-server external_agent_config_import_ --test all`


stefanstokic-oai:

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -1123,6 +1149,39 @@ pub struct ExternalAgentConfigMigrationItem {
     pub details: Option<MigrationDetails>,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct ExternalAgentConfigImportMigrationItem {
```
alexsong-oai: Can just reuse ExternalAgentConfigMigrationItem?

codex-rs/core/src/codex_thread.rs
```diff
@@ -272,6 +272,14 @@ impl CodexThread {
         self.codex.session.token_usage_info().await
     }
 
+    pub async fn recompute_token_usage(&self) {
```
alexsong-oai: What does this for?

stefanstokic-oai: Seeds token usage for the imported historical rollout so auto-compaction can work on the first follow-up when the imported session is over the model’s context limit. Otherwise, the imported rollout has no prior model-reported usage yet and can look empty to the compaction check.

Alternatively, we can truncate imported session to fit within the context window.

codex-rs/app-server-protocol/src/protocol/v2.rs
stefanstokic-oai: addressed in https://github.com/openai/codex/pull/19895/changes/7986ca1209c65537d61cc17ca67ba9f6ab89213c

alexsong-oai (approved):

pakrym-oai (approved):

codex-rs/app-server/src/external_agent_sessions/import.rs
```diff
@@ -0,0 +1,224 @@
+use super::ConversationMessage;
```
pakrym-oai: please move all conversion logic to a separate crate

stefanstokic-oai: addressed in https://github.com/openai/codex/commit/a46f7e42a6dd8e4c65998632043f8e45ad61a59e

PR #19899, opened 2026-04-28
app-server-protocol: mark permission profiles experimental

bolinfest:
## Why

`PermissionProfile` is now the canonical internal permissions representation, but the app-server wire shape is still intentionally unstable while the migration continues. Stable app-server clients should not see or generate code for these fields until the wire format settles.

## What changed

- Marks every app-server v2 field that sends `PermissionProfile` as experimental, including `command/exec`, `thread/start`, `thread/resume`, `thread/fork`, and `turn/start` request/response payloads.
- Enables per-field experimental inspection for `command/exec`, so `permissionProfile` is gated without making the entire method experimental.
- Fixes the generated TypeScript schema filter to be comment-aware. The previous scanner treated apostrophes inside doc comments as string delimiters, so some experimental fields leaked into stable TypeScript even though stable JSON was filtered correctly.

## Verification

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/19899).
* #19900
* __->__ #19899


viyatb-oai:
@codex review

codex-rs/app-server-protocol/schema/typescript/v2/CommandExecutionRequestApprovalParams.ts
```diff
@@ -19,40 +16,25 @@ export type CommandExecutionRequestApprovalParams = { threadId: string, turnId:
...
-/**
+proposedExecpolicyAmendment?: ExecPolicyAmendment | null, /**
  * Optional proposed network policy amendments (allow/deny host) for future requests.
  */
-proposedNetworkPolicyAmendments?: Array<NetworkPolicyAmendment> | null,
-/**
- * Ordered list of decisions the client may present for this prompt.
- */
-availableDecisions?: Array<CommandExecutionApprovalDecision> | null, };
+proposedNetworkPolicyAmendments?: Array<NetworkPolicyAmendment> | null};
```
viyatb-oai, line 40: [P2] Strip all experimental approval fields for stable clients

This stable generated type now hides `availableDecisions`, but runtime compatibility stripping only clears `additional_permissions`. `bespoke_event_handling` still populates `available_decisions`, so non-experimental clients can receive a field that the stable schema/TS no longer exposes.

Suggested fix: keep the source runtime stripping in sync with this generated shape by also clearing `available_decisions`:

```rust
pub fn strip_experimental_fields(&mut self) {
    self.additional_permissions = None;
    self.available_decisions = None;
}
```

If `availableDecisions` is intended to stay visible to stable clients, then the alternative is to make that field non-experimental and keep it in the stable generated schemas.

viyatb-oai (approved):

PR #19900, opened 2026-04-28
permissions: add built-in default profiles

bolinfest:
## Why

The migration away from `SandboxPolicy` needs new configs to start from permissions profiles instead of deriving profiles from legacy sandbox modes. Existing users can have empty `config.toml` files, and we should not rewrite user-owned config files that may live in shared repositories.

This PR introduces built-in profile names so an empty config can resolve to a canonical `PermissionProfile`, while explicit named `[permissions]` profiles still behave predictably.

## What changed

- Adds built-in `default_permissions` profile names:
  - `:read-only` maps to `PermissionProfile::read_only()`.
  - `:workspace` maps to the workspace-write profile, including project-root metadata carveouts.
  - `:danger-no-sandbox` maps to `PermissionProfile::Disabled`, preserving the distinction between no sandbox and a broad managed sandbox.
- Reserves the `:` prefix for built-in profiles so user-defined `[permissions]` profiles cannot collide with future built-ins.
- Allows `default_permissions` to reference a built-in profile without requiring a `[permissions]` table.
- Makes an otherwise empty config choose a built-in profile by trust/platform context: trusted or untrusted project roots use `:workspace` when the platform supports that sandbox, while roots without a trust decision use `:read-only`.
- Keeps legacy `sandbox_mode` configs on the legacy path, and still rejects user-defined `[permissions]` profiles that omit `default_permissions` so we do not silently guess among custom profiles.
- Preserves compatibility behavior for implicit defaults: bare `network.enabled = true` allows runtime network without starting the managed proxy, explicit profile proxy policy still starts the proxy, and implicit workspace/add-dir roots keep legacy metadata carveouts.

## Verification

- `cargo test -p codex-core builtin --lib`
- `cargo test -p codex-core profile_network_proxy_config`
- `cargo test -p codex-core implicit_builtin_workspace_profile_preserves_add_dir_metadata_carveouts`
- `cargo test -p codex-core permissions_profiles_network_enabled_allows_runtime_network_without_proxy`
- `cargo test -p codex-core permissions_profiles_proxy_policy_starts_managed_network_proxy`

## Documentation

Public Codex config docs should mention these built-in names when the `[permissions]` config format is ready to document as stable.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/19900).
* #20041
* #20040
* #20037
* #20035
* #20034
* #20033
* #20032
* #20030
* #20028
* #20027
* #20026
* #20024
* #20021
* #20018
* #20016
* #20015
* #20013
* #20011
* #20010
* #20008
* __->__ #19900

### DIFF
--- a/codex-rs/core/src/config/mod.rs
+++ b/codex-rs/core/src/config/mod.rs
@@ -99,9 +99,12 @@ use std::path::Path;
 use std::path::PathBuf;
 use std::sync::Arc;
 
-use crate::config::permissions::compile_permission_profile;
+use crate::config::permissions::builtin_permission_profile;
+use crate::config::permissions::compile_permission_profile_selection;
+use crate::config::permissions::default_builtin_permission_profile_name;
 use crate::config::permissions::get_readable_roots_required_for_codex_runtime;
-use crate::config::permissions::network_proxy_config_from_profile_network;
+use crate::config::permissions::network_proxy_config_for_profile_selection;
+use crate::config::permissions::validate_user_permission_profile_names;
 use codex_network_proxy::NetworkProxyConfig;
 use toml::Value as TomlValue;
 use toml_edit::DocumentMut;
@@ -315,6 +318,18 @@ impl Permissions {
 
         self.permission_profile.set(permission_profile)?;
         Ok(())
+    }
+}
+
+// A profile override only inherits the selected profile's proxy/allowlist config
+// when Codex is still responsible for the network policy. `Disabled` means no
+// outer sandbox, so starting the managed proxy would narrow the override.
+fn profile_allows_configured_network_proxy(permission_profile: &PermissionProfile) -> bool {
+    match permission_profile {
+        PermissionProfile::Managed { network, .. } | PermissionProfile::External { network } => {
+            network.is_enabled()
+        }
+        PermissionProfile::Disabled => false,
     }
 }
 
@@ -1861,6 +1876,7 @@ impl Config {
             .permissions
             .as_ref()
             .is_some_and(|profiles| !profiles.is_empty());
+        validate_user_permission_profile_names(cfg.permissions.as_ref())?;
         if has_permission_profiles
             && !matches!(
                 permission_config_syntax,
@@ -1891,8 +1907,8 @@ impl Config {
         let profiles_are_active = matches!(
             permission_config_syntax,
             Some(PermissionConfigSyntax::Profiles)
-        ) || (permission_config_syntax.is_none()
-            && has_permission_profiles);
+        ) || permission_config_syntax.is_none();
+        let using_implicit_builtin_profile = permission_config_syntax.is_none();
         let (
             configured_network_proxy_config,
             permission_profile,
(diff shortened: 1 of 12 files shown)


codex-rs/core/src/config/mod.rs
```diff
@@ -1947,32 +1945,27 @@ impl Config {
...
+                    cfg.permissions.as_ref(),
                     default_permissions,
                     resolved_cwd.as_path(),
                     &mut startup_warnings,
                 )?;
-            let mut permission_profile = PermissionProfile::from_runtime_permissions(
-                &file_system_sandbox_policy,
-                network_sandbox_policy,
-            );
+            let mut permission_profile = builtin_permission_profile(default_permissions)
```
viyatb-oai: [P1] Preserve legacy workspace-write settings for implicit trusted defaults

When no `sandbox_mode` or `default_permissions` is configured, trusted projects now enter the profile path and select the built-in `:workspace` profile. That bypasses `ConfigToml::derive_permission_profile()`, so legacy `[sandbox_workspace_write]` settings are no longer applied for the default trusted-project workspace-write case. Existing configs that relied on the trust default plus `sandbox_workspace_write.network_access = true`, `writable_roots`, or tmp exclusions now silently get the plain built-in workspace profile instead.

I verified this with a temporary regression test in a detached `/tmp/codex-pr19900-review` worktree: trusted project, no explicit `sandbox_mode`, `[sandbox_workspace_write] network_access = true`, and an extra writable root. It failed because the loaded config had restricted network before it even reached the extra-root assertions.

Suggested fix: make the built-in workspace profile construction go through the legacy workspace-write settings when those settings are present, rather than always using `PermissionProfile::workspace_write()`. For example, thread `cfg.sandbox_workspace_write.as_ref()` into the built-in helper and use the existing conversion shape:

```rust
fn builtin_permission_profile(
    profile_name: &str,
    workspace_write: Option<&SandboxWorkspaceWrite>,
) -> Option<PermissionProfile> {
    match profile_name {
        BUILT_IN_WORKSPACE_PROFILE => Some(match workspace_write {
            Some(SandboxWorkspaceWrite {
                writable_roots,
                network_access,
                exclude_tmpdir_env_var,
                exclude_slash_tmp,
            }) => PermissionProfile::workspace_write_with(
                writable_roots,
                if *network_access {
                    NetworkSandboxPolicy::Enabled
                } else {
                    NetworkSandboxPolicy::Restricted
                },
                *exclude_tmpdir_env_var,
                *exclude_slash_tmp,
            ),
            None => PermissionProfile::workspace_write(),
        }),
        // existing arms...
    }
}
```

Then use that same helper in both `compile_permission_profile_selection(...)` and the later `permission_profile` reconstruction so runtime policy, stored profile, and compatibility projection stay aligned.

codex-rs/core/src/config/mod.rs
```diff
@@ -1947,32 +1945,27 @@ impl Config {
...
-            let profile = resolve_permission_profile(permissions, default_permissions)?;
-            let configured_network_proxy_config =
-                network_proxy_config_from_profile_network(profile.network.as_ref());
+            let default_permissions = cfg.default_permissions.as_deref().unwrap_or_else(|| {
+                default_builtin_permission_profile_name(&active_project, windows_sandbox_level)
+            });
+            let configured_network_proxy_config = network_proxy_config_for_profile_selection(
+                cfg.permissions.as_ref(),
+                default_permissions,
+            )?;
```
viyatb-oai, line 1969: [P2] Keep profile network enablement separate from proxy startup

One network-side detail to keep separate from the legacy sandbox mapping: `[permissions.<profile>.network].enabled = true` should map to the sandbox network bit (`NetworkSandboxPolicy::Enabled`, and legacy `network_access = true` when projected), but it should not by itself start the managed network proxy. Users coming from legacy `network_access = true` expect either direct/all network allowed or no network, not an implicit proxy path.

Right now this selection helper builds `configured_network_proxy_config` from the selected profile's `NetworkToml`, and `NetworkToml::apply_to_network_proxy_config()` copies `network.enabled` into `NetworkProxyConfig.network.enabled`. Later config loading keeps `Config.permissions.network = Some(...)` whenever that proxy config is enabled, so a profile that only wants sandbox network enabled can also activate the proxy.

Suggested fix: treat `NetworkToml.enabled` as input to `compile_network_sandbox_policy()` only, and do not let it flip `NetworkProxyConfig.network.enabled` in `network_proxy_config_for_profile_selection()` / `network_proxy_config_from_profile_network()`. Keep proxy startup behind a separate explicit knob such as `enable_network_proxy` (or managed network requirements), while still allowing the other profile network fields (`domains`, sockets, proxy URL, local binding) to overlay proxy policy when that separate proxy path is enabled.

codex-rs/core/src/config/mod.rs
bolinfest: [codex] Addressed in 346a2a10dbb4. The implicit `:workspace` path now threads `sandbox_workspace_write` through built-in profile construction, so existing `writable_roots`, `network_access`, and tmp exclusion settings are preserved. Added regression coverage in `implicit_builtin_workspace_profile_preserves_sandbox_workspace_write_settings`.

codex-rs/core/src/config/mod.rs
bolinfest: [codex] Addressed in 346a2a10dbb4. Profile `network.enabled` now maps to the runtime network sandbox bit without starting the managed proxy; managed network requirements remain the separate proxy-start path and the affected integration tests now inject those requirements during config construction. Added config/helper regression coverage for this.

viyatb-oai (approved):

PR #19762, opened 2026-04-27
refactor: make auth loading async

efrazer-oai:
## Summary

Auth loading used to expose synchronous construction helpers in several places even though some auth sources now need async work. This PR makes the auth-loading surface async and updates the callers to await it.

This is intentionally only plumbing. It does not change how AgentIdentity tokens are decoded, how task runtime ids are allocated, or how JWT signatures are verified.

## Stack

1. **This PR:** [refactor: make auth loading async](https://github.com/openai/codex/pull/19762)
2. [refactor: load AgentIdentity runtime eagerly](https://github.com/openai/codex/pull/19763)
3. [feat: verify AgentIdentity JWTs with JWKS](https://github.com/openai/codex/pull/19764)

## Important call sites

| Area | Change |
| --- | --- |
| `codex-login` auth loading | `CodexAuth` and `AuthManager` construction paths now await auth loading. |
| app-server startup | Auth manager construction is awaited during initialization. |
| CLI/TUI/exec/MCP/chatgpt callers | Existing auth-loading calls now await the same behavior. |
| cloud requirements storage loader | The loader becomes async so it can share the same auth construction path. |
| auth tests | Tests that load auth now run in async contexts. |

## Testing

Tests: targeted Rust auth test compilation, formatter, scoped Clippy fix, and Bazel lock check.

PR #19904, opened 2026-04-28
fix: configure AgentIdentity AuthAPI base URL

efrazer-oai:
## Summary

AgentIdentity runtime loading currently registers tasks against a single hardcoded AuthAPI base URL. That works for production, but local and staging validation may need registration to target a different authapi-login-provider without baking internal staging service URLs into the OSS binary.

This PR adds a small config surface for `agent_identity_authapi_base_url` and threads it through the existing auth-loading path as a direct argument. Explicit config wins. Without config, task registration keeps using the production AuthAPI URL, matching the current default behavior.

## Stack

1. openai/codex#19762 - `refactor: make auth loading async` (merged)
2. openai/codex#19763 - `refactor: load agent identity runtime eagerly`
3. This PR - `fix: configure AgentIdentity AuthAPI base URL`
4. openai/codex#19764 - `feat: verify agent identity JWTs with JWKS`

## Design decisions

- Keep the existing auth-loading shape and pass the new value as an argument. This avoids another wrapper loader and keeps the call path readable.
- Add config instead of embedding internal staging URLs. Environments that need a non-production AuthAPI can configure it explicitly.
- Keep the default AuthAPI registration URL as production. `chatgpt_base_url` remains separate and is used by the follow-up JWKS verification PR for fetching public keys from the ChatGPT backend route.
- Resolve the AuthAPI base URL inside AgentIdentity loading, because task registration is the only consumer of this value.

## Testing

Tests: targeted Rust checks, AgentIdentity auth tests, config schema regeneration, formatter/fix pass, and whitespace diff check.


shijie-oai (approved):
this is purely for testing against staging etc right? Just want to confirm because previously I did make it to be able to accept env variable but took it out cause did not really see the point of it but now I think we shuold. I am a bit hesitating with adding it to config.toml though but can go either way.

sayan-oai (commented):
agree with shijie's comment; if this is just for testing purposes it shouldnt be in config.toml. it should be an env var that can be set as needed

codex-rs/cloud-requirements/src/lib.rs
```diff
@@ -738,6 +738,7 @@ pub async fn cloud_requirements_loader_for_storage(
         enable_codex_api_key_env,
         credentials_store_mode,
         Some(chatgpt_base_url.clone()),
+        /*agent_identity_authapi_base_url*/ None,
```
sayan-oai: question-- is there ever a legit case someone would want to override the agent_identity_authapi_base_url via cloud_requirements? doesnt seem like it but wanted to double check

efrazer-oai: ack, set it to an env variable as discussed offline!

codex-rs/login/src/auth/agent_identity.rs
```diff
@@ -58,9 +61,77 @@ impl AgentIdentityAuth {
...
+        );
+        assert_eq!(
+            agent_identity_authapi_base_url(),
+            "https://authapi.example.test/api/accounts"
+        );
+    }
+
+    #[test]
+    fn agent_identity_authapi_base_url_uses_prod_authapi_by_default() {
+        let _guard = EnvVarGuard::remove(CODEX_AGENT_IDENTITY_AUTHAPI_BASE_URL_ENV_VAR);
```
sayan-oai, line 100: might need to serialize these two given they mutate env

sayan-oai (approved):

codex-rs/login/src/auth/agent_identity.rs
efrazer-oai: done!

PR #19905, opened 2026-04-28
Add compact lifecycle hooks (started by vincentkoc - external contrib)

eternal-openai:
Based on work from Vincent K - https://github.com/openai/codex/pull/19060

CleanShot 2026-04-29 at 20 47 40@2x

## Why

Compaction rewrites the conversation context that future model turns receive, but hooks currently have no deterministic lifecycle point around that rewrite. This adds compact lifecycle hooks so users can audit manual and automatic compaction, surface hook messages in the UI, and run post-compaction follow-up without overloading tool or prompt hooks.

## What Changed

- Added `PreCompact` and `PostCompact` hook events across hook config, discovery, dispatch, generated schemas, app-server notifications, analytics, and TUI hook rendering.
- Added trigger matching for compact hooks with the documented `manual` and `auto` matcher values.
- Wired `PreCompact` before both local and remote compaction, and `PostCompact` after successful local or remote compaction.
- Kept compact hook command input to lifecycle metadata: session id, Codex turn id, transcript path, cwd, hook event name, model, and trigger.
- Made compact stdout handling consistent with other hooks: plain stdout is ignored as debug output, while malformed JSON-looking stdout is reported as failed hook output.
- Added integration coverage for compact hook dispatch, trigger matching, post-compact execution, and the audited behavior that `decision:"block"` does not block compaction.

## Out of Scope

- Hook-specific compaction blocking is not implemented; `decision:"block"` and exit-code-2 blocking semantics are intentionally unsupported for `PreCompact`.
- Custom compaction instructions are not exposed to compact hooks in this PR.
- Compact summaries, summary character counts, and summary previews are not exposed to compact hooks in this PR.

## Verification

- `cargo test -p codex-hooks`
- `cargo test -p codex-core manual_pre_compact_block_decision_does_not_block_compaction`
- `cargo test -p codex-app-server hooks_list`
- `cargo test -p codex-core config_schema_matches_fixture`
- `cargo test -p codex-tui hooks_browser`

## Docs

The developer documentation for Codex hooks should be updated alongside this feature to document `PreCompact` and `PostCompact`, the `manual`/`auto` matcher values, and the compact hook payload fields.


codex-rs/core/src/hook_runtime.rs
```diff
@@ -250,6 +251,75 @@ pub(crate) async fn run_post_tool_use_hooks(
...
+    Stopped { reason: Option<String> },
+    Blocked { reason: String },
+}
+
+pub(crate) enum PostCompactHookOutcome {
+    Continue,
+    Stopped,
+}
+
+pub(crate) async fn run_post_compact_hooks(
```
pakrym-oai, line 296: very non blocking nit: these methods look very boilerplate (most of request parameters are repeated, event emmiting, etc.)

pakrym-oai (approved):

PR #19907, opened 2026-04-28
Clarify network approval auto-review prompts

maja-openai:
## Why

Network access approval prompts were showing the generic retry reason, which made auto-review focus on the blocked connection instead of the command that caused it. This makes network approvals easier to assess by telling the reviewer to evaluate whether the triggering command was authorised by the user and within policy, and to treat the network call as acceptable when it is a reasonable consequence of that command.

## What changed

- Split guardian approval request prompt rendering so `NetworkAccess` has a dedicated branch.
- For network requests, show `Network approval context` and `Network access JSON` instead of `Retry reason` / `Planned action JSON`.
- Added regression coverage for the network approval prompt wording and for omitting retry reason in this case.

## Verification

- `cargo test -p codex-core guardian::tests::build_guardian_prompt_items_explains_network_access_review_scope`

### DIFF
--- a/codex-rs/core/src/guardian/prompt.rs
+++ b/codex-rs/core/src/guardian/prompt.rs
@@ -172,17 +172,41 @@ pub(crate) async fn build_guardian_prompt_items(
     if let Some(note) = omission_note {
         push_text(format!("\n{note}\n"));
     }
-    push_text(headings.action_intro.to_string());
-    push_text(">>> APPROVAL REQUEST START\n".to_string());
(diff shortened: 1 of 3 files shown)


dylan-hurd-oai (approved):
LGTM! could we add a snapshot test for this as well?

maja-openai:
[codex] Addressed in the latest branch updates: this PR now includes an insta snapshot test for the network access guardian prompt layout (codex_core__guardian__tests__network_access_guardian_prompt_layout).

PR #19917, opened 2026-04-28
Allow /statusline and /title slash commands during active turns

canvrno-oai:
- Marks `/title` and `/statusline` as available during active tasks.
- Extends the existing slash-command availability test coverage to include these commands alongside `/goal`.

### DIFF
--- a/codex-rs/tui/src/slash_command.rs
+++ b/codex-rs/tui/src/slash_command.rs
@@ -196,6 +196,8 @@ impl SlashCommand {
             | SlashCommand::Mcp
             | SlashCommand::Apps
             | SlashCommand::Plugins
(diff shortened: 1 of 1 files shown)


canvrno-oai:
@codex review

etraut-openai:
I like this change in general. I tested it manually, and it seems to work fine.

One scenario that I tested was an approval prompt that appears while the `/title` or `/statusline` menu is displayed. The approval prompt temporarily replaces the menu, which makes sense. Once the approval prompt is dismissed, the menu reappears. It was a little jarring to have the approval prompt suddenly appear, but I think this situation will be sufficiently rare that it's not a problem.

etraut-openai (approved):

PR #19919, opened 2026-04-28
app-server: notify clients of remote-control status changes

euroelessar:
## Why

Remote-control app-server enrollments have both an internal server id and the environment id exposed to remote-control clients. App-server clients need one current status snapshot that says whether remote control is usable and which environment id, if any, is exposed.

A temporary websocket disconnect is not itself an identity change. Account changes, stale enrollment invalidation, successful re-enrollment, and missing ChatGPT auth are meaningful status changes. Disabled remote control remains `disabled` regardless of auth or SQLite state. SQLite startup failure disablement and enrollment persistence failures are handled in #20068; this PR reports the resulting effective status to clients.

## What changed

- Adds v2 `remoteControl/status/changed` carrying `state` and `environmentId`.
- Adds `RemoteControlConnectionState` values: `disabled`, `connecting`, `connected`, and `errored`.
- Exposes remote-control status updates through `RemoteControlHandle` using a Tokio watch channel.
- Always sends the current remote-control status snapshot to newly initialized app-server clients.
- Broadcasts status changes to initialized app-server clients when state or environment id changes.
- Treats missing ChatGPT auth as an `errored` status while leaving it retryable because auth can change at runtime.
- Clears `environmentId` when enrollment is cleared for account changes, auth loss, stale backend invalidation, or disabled remote control.
- Updates app-server protocol schema fixtures, generated TypeScript, app-server README, remote-control tests, and TUI exhaustive notification matches.

## Stack

- Builds on #20068.

## Verification

- `cargo test -p codex-app-server transport::remote_control --lib`
- `cargo check -p codex-tui`
- `just fix -p codex-tui`


SagarJadhav30 (commented):
best intreseting

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2877,6 +2877,25 @@ pub struct DeviceKeyPublicResponse {
     pub protection_class: DeviceKeyProtectionClass,
 }
 
+/// Current remote-control connection state and environment id exposed to clients.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct RemoteControlStatusChangedNotification {
+    pub state: RemoteControlConnectionState,
```
owenlin0: nit: `status`? `state` to me feels like it should be a struct whereas `status` sounds more like an enum

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -2877,6 +2877,25 @@ pub struct DeviceKeyPublicResponse {
...
+    pub environment_id: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(rename_all = "camelCase", export_to = "v2/")]
+pub enum RemoteControlConnectionState {
```
owenlin0: `RemoteControlConnectionStatus`?

owenlin0 (approved):
focused mostly on the API and skimmed the impl given that it is isolated to remote_control. codex flagged some comments though, worth addressing?

PR #19929, opened 2026-04-28
TUI: use cumulative turn duration for worked-for separator

### LINKED ISSUE #19814: Codex no longer shows reliable full prompt elapsed time

codex-cli 0.125.0

API key

gpt-5.5

Darwin 25.4.0 arm64 arm

iTerm2

Codex used to make it easy to see how long a full prompt/turn took, from submission until the agent stopped. Recently, the displayed Worked for ... duration is missing or no longer reliable. For long-running prompts, I need the total wall-clock time for the whole turn, not a reset/truncated status duration.

This seems related to existing reports about Worked for ... resetting or not being cumulative.  

1. Start Codex CLI/TUI.
2. Submit a prompt that takes several minutes or longer.
3. Wait until the turn fully completes.
4. Check the final displayed duration.

Observed: the UI does not reliably show the full time from prompt submission to completion.

Codex should show a stable cumulative elapsed time for each turn: from when I submit the prompt until Codex fully stops.

### Additional information

_No response_

etraut-openai:
## Why

Fixes #19814.

The TUI's current `Worked for ...` timing behavior is a leftover from #9599. At that point, models could emit multiple assistant messages in one turn for preambles/commentary, but the TUI did not yet have a reliable signal that an assistant message was the final answer when it started streaming. To avoid showing an ever-growing elapsed time on each preamble separator, #9599 made the separator timer incremental by tracking elapsed time since the previous separator.

That workaround is no longer the right model for the final completed-turn display. Since then, #16638 added protocol-native turn timing, including `duration_ms` on turn completion. With that cumulative duration available at the point where the TUI renders the completed-turn separator, the UI can show the actual turn duration directly instead of carrying per-separator timing state.

## What Changed

- Thread `duration_ms` into `ChatWidget::on_task_complete` from both legacy `TurnCompleteEvent` handling and app-server `TurnCompleted` notifications.
- Use `duration_ms` for the final `Worked for ...` separator, falling back to the status indicator timer only when the protocol duration is unavailable.
- Keep mid-turn separators before later assistant text as plain visual dividers instead of clocked `Worked for ...` separators.
- Remove the old incremental separator timer state and helper (`last_separator_elapsed_secs` / `worked_elapsed_from`).
- Add a snapshot regression test for a turn that runs a command and then completes with a final answer, verifying the final separator uses the cumulative turn duration.

## Verification

- `cargo test -p codex-tui final_worked_for_uses_cumulative_turn_duration_snapshot`
- `just fix -p codex-tui`

Manual repro prompt:

```text
Manual timing repro. First send a short preamble/commentary sentence before using tools. Then run exactly this shell command: sleep 75; echo MANUAL_TIMING_DONE. After the command finishes, give a final answer that says "done". Do not skip the preamble.
```

After this change, the mid-turn break before the final answer should be a plain divider, and the final completed-turn separator should show `Worked for ...` using the cumulative turn duration.

Before:
Screenshot 2026-04-27 at 10 09 01 PM

After:
Screenshot 2026-04-27 at 10 09 07 PM

### DIFF
--- a/codex-rs/tui/src/chatwidget.rs
+++ b/codex-rs/tui/src/chatwidget.rs
@@ -991,13 +991,12 @@ pub(crate) struct ChatWidget {
     // Whether the next streamed assistant content should be preceded by a final message separator.
     //
     // This is set whenever we insert a visible history cell that conceptually belongs to a turn.
-    // The separator itself is only rendered if the turn recorded "work" activity (see
-    // `had_work_activity`).
+    // The separator itself is only rendered if the turn recorded "work" activity.
     needs_final_message_separator: bool,
     // Whether the current turn performed "work" (exec commands, MCP tool calls, patch applications).
     //
     // This gates rendering of the "Worked for …" separator so purely conversational turns don't
-    // show an empty divider. It is reset when the separator is emitted.
+    // show an empty divider.
     had_work_activity: bool,
     // Whether the current turn emitted a plan update.
     saw_plan_update_this_turn: bool,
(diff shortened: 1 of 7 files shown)


fcoury-oai (approved):
I was able to replicate the issue on alpha codex and confirmed the fix works.

PR #19931, opened 2026-04-28
Move local /resume cwd filtering into thread/list

canvrno-oai:
Move local resume and fork cwd filtering to `thread/list` instead of filtering in the TUI. This makes the `/resume` menu feel slightly faster to load when working in repos with many historical threads, and centralizes the cwd filtering in app-server.

**Affected:**
- /resume from inside the TUI.
- codex resume with no session ID and without --last
- codex resume --all
- codex fork with no session ID and without --last
- codex fork --all

**Not affected:**
- codex resume <id>
- codex fork <id>
- codex resume --last
- codex fork --last

Steps to test performance improvement in a real Codex environment:
- Launch `codex resume` using compiled binary in a directory that has seen many threads.
- Launch `codex resume` using release binary in same directory.
- Observe difference in time-to-full-page as threads load.

### DIFF
--- a/codex-rs/tui/src/resume_picker.rs
+++ b/codex-rs/tui/src/resume_picker.rs
@@ -143,8 +143,8 @@ struct PickerPage {
 /// sessions appear during pagination.
 ///
 /// Filtering happens in two layers:
-/// 1. Provider and source filtering at the backend.
-/// 2. Working-directory filtering at the picker (unless `--all` is passed).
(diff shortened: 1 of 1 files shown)


etraut-openai (approved):
A nice code cleanup! It also appears to improve performance of the resume picker in the tui. 

I built locally and ran it through some manual tests to look for regressions, and I didn't find any problems.

Issue #19925, opened 2026-04-28
Missing message/progression bar


Whyaremax:
### What version of Codex CLI is running?

codex-cli 0.125.0

### What subscription do you have?

Pro

### Which model were you using?

gpt-5.4

### What platform is your computer?

Linux 6.19.8 x86_64 unknown (Debian 13, no idea why its "unknown")

### What terminal emulator and version are you using (if applicable)?

GNOME Terminal

### What issue are you seeing?

When model are summarying up, or reporting to user, example: `I did... I changed...`, it would cause `working` message be disappeared

### What steps can reproduce the bug?

(Thread ID are not available due to sensitive info)
Let agent do a thing
When summarying up, 
queue next message
Suppose, you should be seeing no working message initally
But Agent will still do its job, and the working message gonna came back after first tool call

### What is the expected behavior?

Immediately able to see `working`

etraut-openai:
I'm not understanding your bug report. Could you provide more specific repro steps? Or a screen shot?

canvrno-oai:
@etraut-openai I was able to duplicate this:
- Create a new thread and task the agent with doing something that will result in a long output.
- While output is streaming, queue a steer.
- When the output from the first message ends, the steer _**is**_ sent, but there is no "Working..." text or visual indication that the steer was sent.

PR #19939, opened 2026-04-28
Restore TUI working status after steer message is set

canvrno-oai:
Fix for #19925

Restore the `Working` indicator after a streamed final answer finishes when a user steer message is sent.
Add regression coverage for long output plus a mid-stream steer:
  `cargo test -p codex-tui final_answer_completion_restores_status_indicator_for_pending_steer`

Duplication/testing steps:
1. Start a new thread and ask for a long response.
2. While the response is streaming, submit a steer message.
3. When the first response finishes, observe whether `Working...` is shown while waiting for the steer message response.


etraut-openai (approved):
Looks good!

PR #19965, opened 2026-04-28
[codex] Sync remote plugin config toggles

xli-oai:
## Summary
- Translate remote plugin `plugins.<remotePluginId>.enabled` config writes into remote install/uninstall calls instead of persisting backend IDs into `config.toml`.
- Preserve normal local plugin enable/disable behavior through config writes.
- Add app-server coverage for remote enable/install/cache materialization and remote disable/uninstall/cache removal, plus README notes for the compatibility path.

## Test plan
- `just fmt`


xl-openai:
Let's bring it offline:
1. Do we need a different API for enable/disable. Adding it into batchWrite is very strange.
2. Do we really need to support disable a remote plugin?

PR #19966, opened 2026-04-28
Require remote plugin detail before uninstall

xli-oai:
## Summary
- Fetch remote plugin detail before sending the uninstall request.
- Use the detail response to derive the marketplace namespace and plugin name for cache cleanup.
- Stop the uninstall before the backend POST if detail lookup fails, so backend state and local cache state do not diverge.

## Testing
- `just fmt`
- `cargo test -p codex-app-server plugin_uninstall`
- `git diff --check`


xli-oai:
@codex review

xl-openai (approved):

PR #19999, opened 2026-04-28
fix(network-proxy): recheck network proxy connect targets

viyatb-oai:
## Why
The proxy checks the requested host before opening the upstream connection, but DNS can resolve an allowed hostname to a loopback, private, or other non-public address after that first decision. Without a final check on the actual socket target, a request that looks acceptable at the hostname layer can still connect to a local service once resolution completes.

## What changed
- add a shared TCP connector check for direct proxy egress
- use that path for HTTP, `CONNECT`, SOCKS5, and MITM upstream connections
- keep configured upstream proxy hops on the existing proxy path
- add direct-connector coverage for allowed and rejected local targets

## Security impact
Direct proxy egress now rechecks the resolved socket address before connecting, closing the gap between hostname policy evaluation and the final network target.

## Verification

### DIFF
--- a/codex-rs/network-proxy/src/connect_policy.rs
+++ b/codex-rs/network-proxy/src/connect_policy.rs
@@ -1,0 +1,192 @@
+use crate::policy::is_non_public_ip;
+use crate::state::NetworkProxyState;
+use rama_core::Service;
+use rama_core::error::BoxError;
+use rama_core::error::ErrorExt as _;
+use rama_core::error::OpaqueError;
+use rama_core::extensions::ExtensionsMut;
+use rama_net::address::ProxyAddress;
+use rama_net::client::EstablishedClientConnection;
+use rama_net::transport::TryRefIntoTransportContext;
+use rama_tcp::TcpStream;
+use rama_tcp::client::TcpStreamConnector;
+use rama_tcp::client::service::TcpConnector;
+use std::io;
+use std::net::SocketAddr;
+use std::sync::Arc;
+
+#[derive(Clone)]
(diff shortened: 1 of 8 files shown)


evawong-oai (approved):
LGTM

evawong-oai:

1. Confirmed this maps to the DNS rebinding item in the readiness tracker.
2. Confirmed the issue is the DNS rebinding TOCTOU where policy checks an allowed host before connect, then the final TCP connect can resolve to loopback or private IP.
3. Reviewed the new TargetCheckedTcpConnector and confirmed it rejects the final resolved SocketAddr when allow_local_binding is false and the target IP is non public.
4. Confirmed HTTP CONNECT uses the target checked connector.
5. Confirmed plain HTTP direct egress uses the target checked connector.
6. Confirmed SOCKS5 TCP uses the target checked connector.
7. Confirmed MITM upstream config carries allow_local_binding so MITM egress also uses the same final target policy.
8. Verified the PR head I reviewed was e0d08ebf9216b37da0a33c858763d7d88e2048b8.
9. Verified required GitHub checks were green, including Ubuntu build, clippy, and release checks.
10. Ran cargo test for codex network proxy locally and got 135 passed and 0 failed.
11. Ran git diff check locally and it passed.
12. Approved and squash merged. Merge commit is e1ba87ccb28e59139f451cc038550473b34dd57b.

Remaining readiness gate: verify the target event build includes the merge commit and rerun the original rbndr.us PoC or an equivalent real Codex managed proxy regression showing the old local target read path fails.

PR #20001, opened 2026-04-28
fix(network-proxy): harden linux proxy bridge helpers

viyatb-oai:
## Why
The Linux managed-proxy bridge helpers are long-lived child processes in the sandbox networking path. Before this change they stayed dumpable and the network seccomp profile did not block cross-process memory syscalls, so another same-user process could potentially inspect or modify bridge memory instead of interacting only through the intended proxy interface.

## What changed
- reuse the shared `codex-process-hardening` helper to mark bridge helper children non-dumpable before they begin serving
- deny `process_vm_readv` and `process_vm_writev` in the existing network seccomp filter

## Security impact
Bridge helpers are less exposed to same-user cross-process inspection or memory writes, which reduces the chance that sandboxed code can interfere with proxy support processes outside the intended IPC path.

## Verification
- `cargo test -p codex-process-hardening`
- attempted `cargo check -p codex-linux-sandbox --target x86_64-unknown-linux-gnu`; blocked on missing `x86_64-linux-gnu-gcc` on this macOS host

### DIFF
--- a/codex-rs/Cargo.lock
+++ b/codex-rs/Cargo.lock
@@ -2801,6 +2801,7 @@ dependencies = [
  "cc",
  "clap",
  "codex-core",
+ "codex-process-hardening",
  "codex-protocol",
  "codex-sandboxing",
  "codex-utils-absolute-path",
--- a/codex-rs/linux-sandbox/Cargo.toml
+++ b/codex-rs/linux-sandbox/Cargo.toml
@@ -17,6 +17,7 @@ workspace = true
 
 [target.'cfg(target_os = "linux")'.dependencies]
 clap = { workspace = true, features = ["derive"] }
+codex-process-hardening = { workspace = true }
 codex-protocol = { workspace = true }
 codex-sandboxing = { workspace = true }
 codex-utils-absolute-path = { workspace = true }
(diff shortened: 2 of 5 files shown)


evawong-oai (approved):
LGTM

evawong-oai:

1. Confirmed this maps to BUGB 16151 and BUGB 16152 in the Pwn2Own readiness tracker.
2. Reviewed the Linux bridge change and confirmed both host and local bridge helpers call harden_bridge_process before serving.
3. Confirmed harden_bridge_process preserves the parent death signal and calls codex_process_hardening::disable_process_dumping, which sets PR_SET_DUMPABLE to 0.
4. Confirmed network seccomp now denies ptrace, process_vm_readv, and process_vm_writev.
5. Verified required GitHub checks were green, including Ubuntu build, clippy, and release checks.
6. Ran local validation: cargo test for process hardening passed, cargo test for the Linux sandbox crate completed on macOS but had no Linux tests to run, and diff check passed.
7. Noted remaining readiness gate: verify the target event build includes commit 3377afd84a420193839a06e237118e7a267e9f94 and rerun a Linux PoC or regression showing the old bridge memory manipulation path fails.

PR #20002, opened 2026-04-28
fix(network-proxy): tighten network proxy bypass defaults

viyatb-oai:
## Why
Managed sessions use `NO_PROXY` to keep a small set of destinations on the direct path by default. The old default also bypassed all IPv4 link-local addresses in `169.254.0.0/16`, which includes metadata endpoints such as `169.254.169.254`. Because `NO_PROXY` is evaluated by the client before the request reaches the managed proxy, requests to that range could skip proxy-side allowlist and local-binding checks entirely. On hosts where a link-local metadata service is reachable, that creates a path to sensitive environment metadata or credentials outside the intended enforcement point.

## What changed
- remove the default IPv4 link-local `169.254.0.0/16` bypass from the managed proxy environment
- keep the existing loopback and private-network defaults unchanged
- update the regression assertion to lock in the narrower default

## Security impact
Link-local requests now stay on the managed-proxy path by default, so the proxy can apply configured policy before they reach metadata-style endpoints or other link-local services.

## Verification

### DIFF
--- a/codex-rs/network-proxy/src/proxy.rs
+++ b/codex-rs/network-proxy/src/proxy.rs
@@ -422,7 +422,6 @@ pub const NO_PROXY_ENV_KEYS: &[&str] = &[
 
 pub const DEFAULT_NO_PROXY_VALUE: &str = concat!(
     "localhost,127.0.0.1,::1,",
-    "169.254.0.0/16,",
     "10.0.0.0/8,",
     "172.16.0.0/12,",
     "192.168.0.0/16"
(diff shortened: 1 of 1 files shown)


evawong-oai (approved):
LGTM

evawong-oai:

1. Confirmed this maps to the link local proxy bypass item in the readiness tracker.
2. Reviewed the proxy default bypass list and confirmed 169.254.0.0/16 was removed from DEFAULT_NO_PROXY_VALUE.
3. Confirmed the change means link local IMDS requests are no longer silently excluded from proxy policy by the default no proxy list.
4. Verified required GitHub checks were green before merge.
5. Ran the targeted network proxy validation locally and confirmed the relevant tests passed.
6. Recorded the merged commit 3afb185a4f02dab00927ad597996f3e5528cea45 in the readiness tracker.

Remaining readiness gate: verify the target event build includes this commit and validate link local IMDS requests route through proxy policy with allow_local_binding false.

PR #20008, opened 2026-04-28
tui: use permission profiles for sandbox state

bolinfest:
## Summary
- Move TUI permission state from legacy `SandboxPolicy` values to canonical `PermissionProfile` values across presets, app events, chat widget state, app commands, thread routing, and cached thread session state.
- Keep app-server compatibility boundaries explicit: embedded sessions send `permissionProfile`, while remote sessions send only a legacy `sandbox` projection and fall back to read-only when a custom profile cannot be projected.
- Update status/add-dir UI summaries and snapshots to render the active permission profile, including workspace profiles selected by the new built-in defaults.

## Verification
- `rg '\bSandboxPolicy\b' codex-rs/tui -n` returns no matches.
- `cargo test -p codex-tui`
- `cargo test -p codex-tui additional_dirs`
- `just fmt`
- `just fix -p codex-tui`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/20008).
* #20041
* #20040
* #20037
* #20035
* #20034
* #20033
* #20032
* #20030
* #20028
* #20027
* #20026
* #20024
* #20021
* #20018
* #20016
* #20015
* #20013
* #20011
* #20010
* __->__ #20008

### DIFF
--- a/codex-rs/tui/src/additional_dirs.rs
+++ b/codex-rs/tui/src/additional_dirs.rs
@@ -1,23 +1,35 @@
-use codex_protocol::protocol::SandboxPolicy;
+use codex_protocol::models::PermissionProfile;
 use std::path::PathBuf;
 
 /// Returns a warning describing why `--add-dir` entries will be ignored for the
-/// resolved sandbox policy. The caller is responsible for presenting the
+/// resolved permission profile. The caller is responsible for presenting the
 /// warning to the user (for example, printing to stderr).
 pub fn add_dir_warning_message(
     additional_dirs: &[PathBuf],
-    sandbox_policy: &SandboxPolicy,
+    permission_profile: &PermissionProfile,
+    cwd: &std::path::Path,
 ) -> Option<String> {
     if additional_dirs.is_empty() {
         return None;
     }
 
-    match sandbox_policy {
-        SandboxPolicy::WorkspaceWrite { .. }
-        | SandboxPolicy::DangerFullAccess
-        | SandboxPolicy::ExternalSandbox { .. } => None,
-        SandboxPolicy::ReadOnly { .. } => Some(format_warning(additional_dirs)),
+    if matches!(
+        permission_profile,
+        PermissionProfile::Disabled | PermissionProfile::External { .. }
+    ) {
+        return None;
     }
+
+    let file_system_policy = permission_profile.file_system_sandbox_policy();
+    if file_system_policy.has_full_disk_write_access() {
+        return None;
+    }
+
+    if file_system_policy.can_write_path_with_cwd(cwd, cwd) {
+        return None;
+    }
+
+    Some(format_warning(additional_dirs))
 }
 
 fn format_warning(additional_dirs: &[PathBuf]) -> String {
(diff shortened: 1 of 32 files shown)


codex-rs/tui/src/status/card.rs
```diff
@@ -553,6 +536,51 @@ impl StatusHistoryCell {
...
+            let network_policy = permission_profile.network_sandbox_policy();
+            if permission_profile.to_legacy_sandbox_policy(cwd).is_err() {
+                let mut summary = "custom permissions".to_string();
+                append_network_suffix(&mut summary, network_policy);
+                return summary;
+            }
+
+            if file_system_policy.has_full_disk_write_access() {
+                let mut summary = "workspace-write".to_string();
+                append_network_suffix(&mut summary, network_policy);
```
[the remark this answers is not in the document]
bolinfest: [codex] Addressed in 804f1f934949. The status permission label now derives from the existing legacy summary path, so full-disk managed profiles report as danger-full-access or external-sandbox instead of being forced to workspace-write; added coverage for both network cases.

codex-rs/tui/src/app/platform_actions.rs
```diff
@@ -18,11 +18,21 @@ impl App {
...
+            let Ok(sandbox_policy) = permission_profile.to_legacy_sandbox_policy(cwd.as_path())
+            else {
+                tx.send(AppEvent::OpenWorldWritableWarningConfirmation {
+                    preset: None,
+                    sample_paths: Vec::new(),
+                    extra_count: 0usize,
+                    failed_scan: true,
+                });
+                return;
+            };
```
bolinfest: Hmm, I need to look into this...

codex-rs/tui/src/additional_dirs.rs
```diff
@@ -1,23 +1,38 @@
...
+
+    let file_system_policy = permission_profile.file_system_sandbox_policy();
+    if file_system_policy.has_full_disk_write_access() {
+        return None;
     }
+
+    if !file_system_policy
+        .get_writable_roots_with_cwd(cwd)
+        .is_empty()
+    {
```
fcoury-oai: Should we use `file_system_policy.can_write_path_with_cwd(cwd, cwd)` here instead like we do in `mod.rs`?

codex-rs/tui/src/chatwidget.rs
```diff
@@ -9799,36 +9796,37 @@ impl ChatWidget {
...
-                },
-                SandboxPolicy::WorkspaceWrite {
-                    network_access: preset_network_access,
-                    ..
-                },
-            ) => current_network_access == preset_network_access,
-            _ => false,
+        match preset.id {
+            "full-access" => matches!(current_permission_profile, PermissionProfile::Disabled),
+            "read-only" => {
```
fcoury-oai, line 9894: This one was identified by Codex but here's my understanding of it:

It seems like this classifies any managed profile that can't write `cwd` as Read Only. Isn't it broader than the actual read-only access? 

A custom profile that can write `/tmp/foo` while not writing the repo still satisfies this check, so `/permissions` would show Read Only even though the session is actually _writable_ elsewhere?

Could this required a real `read-only` policy, rather than only `!can_write_path_with_cwd(cwd, cwd)`?

fcoury-oai (approved):
I reviewed the code and performed some sanity checks on the areas touched by this PR:

- `/status` reports Default for the normal workspace-write preset

- Switching `/permissions` from Default to Full Access updates the history line and `/status` to `Permissions: Full Access`.

- Starting with `--sandbox read-only --add-dir /tmp/codex-pr20008-extra` rejects the extra writable root.

- Starting with `--sandbox workspace-write --add-dir ...` and `--sandbox danger-full-access --add-dir ...` starts successfully without the add-dir error.

- After switching to Full Access, the next submitted turn is actually sent with `permission_profile: disabled` and `sandbox-policy: danger-full-access`, so the runtime permission change is taking effect. Verified it directly on the rollout file.

One unrelated behavior I noticed while checking this: on `main` as well as this branch, starting with a fresh `CODEX_HOME` and running the first `/status` shows:

```text
Permissions:          Custom (workspace-write, on-request)
```

If I switch to another permission level and then back to Default, /status then reports Default as expected. I am not familiar enough with the current implementation to tell whether that is intentional.

I am approving it but I also left two review comments on potential issues I noticed in the diff. I will leave it to you to decide whether either is worth addressing here.

codex-rs/tui/src/app/platform_actions.rs
bolinfest: Updating so that we can avoid the spawn_blocking() altogether in the event of an early return.

codex-rs/tui/src/additional_dirs.rs
bolinfest: Addressed in d6fe246e8766. `--add-dir` warnings now check whether the active filesystem policy can write the current workspace (`can_write_path_with_cwd(cwd, cwd)`) instead of treating any writable root as sufficient, and I added a regression test for a profile that can write elsewhere but not `cwd`.

codex-rs/tui/src/chatwidget.rs
bolinfest: Addressed in d6fe246e8766. The Read Only preset now requires a managed profile with no full-disk write access and no writable roots, rather than only checking that `cwd` is not writable. I also added coverage for a managed profile that can write `/tmp/writable` but not the workspace so it is no longer classified as Read Only.

PR #20010, opened 2026-04-28
core tests: submit turns with permission profiles

bolinfest:
## Summary

- Add `PermissionProfile`-based turn submission helpers to `core_test_support`, while keeping the legacy `SandboxPolicy` helper for tests that intentionally exercise legacy fallback behavior.
- Switch the default `TestCodex::submit_turn()` path to send a real `PermissionProfile` plus the required legacy compatibility projection in `Op::UserTurn`.
- Migrate straightforward app/search/shell/truncation tests from `SandboxPolicy::{DangerFullAccess, ReadOnly}` to `PermissionProfile::{Disabled, read_only}`.
- Add a TUI compatibility projection helper for legacy app-server fields so non-legacy writable roots are preserved instead of being downgraded to read-only.
- Fix remote start/resume/fork sandbox-mode projection to classify any managed profile with writable roots as workspace-write, not only profiles that can write `cwd`.
- Reduce `SandboxPolicy` references in `codex-rs/core/tests` from 47 files to 41 files without changing production behavior.

## Testing

- `cargo test -p codex-tui compatibility_profile_preserves_unbridgeable_write_roots`
- `cargo test -p codex-tui sandbox_mode_preserves_non_cwd_write_roots_for_remote_sessions`
- `just fmt`
- `just fix -p core_test_support`
- `just fix -p codex-core`

### DIFF
--- a/codex-rs/tui/src/app_server_session.rs
+++ b/codex-rs/tui/src/app_server_session.rs
@@ -3,6 +3,7 @@ use crate::bottom_pane::FeedbackAudience;
 use crate::legacy_core::append_message_history_entry;
 use crate::legacy_core::config::Config;
 use crate::legacy_core::message_history_metadata;
+use crate::permission_compat::legacy_compatible_permission_profile;
 use crate::status::StatusAccountDisplay;
 use crate::status::plan_type_display_name;
 use codex_app_server_client::AppServerClient;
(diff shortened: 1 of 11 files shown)


bolinfest:
[codex] Addressed in 0cf46843ab65. Remote turn/start now uses a legacy-compatible workspace-write projection for profiles that cannot be represented directly, preserving non-legacy writable roots instead of falling back to read-only; added focused coverage for the compatibility projection.

codex-rs/tui/src/app_server_session.rs
```diff
@@ -1124,7 +1118,10 @@ fn sandbox_mode_from_permission_profile(
                     .network_sandbox_policy()
                     .is_enabled()
                     .then_some(codex_app_server_protocol::SandboxMode::DangerFullAccess)
-            } else if file_system_policy.can_write_path_with_cwd(cwd, cwd) {
+            } else if !file_system_policy
```
etraut-openai: This doesn't seem to be a precise translation. If the profile has any writable roots at all, we're translating this to "WorkspaceWrite", which has historically meant "all roots are writable". Is this intended?

codex-rs/tui/src/permission_compat.rs
```diff
@@ -0,0 +1,102 @@
+use codex_protocol::models::PermissionProfile;
```
etraut-openai, line 4: Nit: Add header comment indicating the purpose of this module?

etraut-openai (approved):
Overall, looks good. One potential hole in the translation logic and one nit for your consideration.

codex-rs/tui/src/permission_compat.rs
bolinfest: Addressed in 631451a504ba by adding a module-level comment explaining that `permission_compat` contains compatibility projections from canonical permission profiles into legacy shapes still required by older or remote app-server APIs.

codex-rs/tui/src/app_server_session.rs
bolinfest: Addressed in 631451a504ba. The remote legacy projection now only emits `workspace-write` when the active profile can write the configured cwd/project root. Profiles with only non-cwd writable roots fall back to `read-only` because the legacy remote field cannot represent arbitrary roots safely; added coverage for both cases.

viyatb-oai (approved):
+1 to @etraut-openai's comment but otherwise lgtm

PR #20013, opened 2026-04-28
core tests: migrate more turns to permission profiles

bolinfest:
## Summary
- Migrate another batch of direct `Op::UserTurn` test construction from legacy `SandboxPolicy` values to `PermissionProfile` inputs via `turn_permission_fields()`.
- Replace a one-off read-only `SandboxPolicy` bridge in the macOS exec test with `PermissionProfile::read_only()`.
- Reduce `SandboxPolicy` references in `codex-rs/core/tests` from 32 files at the start of the cleanup stack to 27 files.

## Testing
- `just fmt`
- `just fix -p codex-core`


bolinfest:
[codex] Addressed in 5f618ba516fc. Remote start/resume/fork sandbox-mode mapping now treats any managed profile with writable roots as workspace-write instead of only checking whether cwd itself is writable; added coverage for a non-cwd write root.

PR #20015, opened 2026-04-28
core tests: configure profiles directly

bolinfest:
## Summary
- Replace legacy sandbox config setup in delegate and telemetry tests with direct `PermissionProfile` configuration.
- Move no-sandbox and read-only test turns in `tools.rs`, `code_mode.rs`, `user_shell_cmd.rs`, and `model_visible_layout.rs` from legacy `SandboxPolicy` values to `PermissionProfile` helpers, while leaving the deny-glob read-only compatibility case for a later targeted cleanup.
- Use `PermissionProfile::read_only()` where tests need managed read-only behavior and `PermissionProfile::Disabled` where they intentionally need no sandbox.
- Reduce `SandboxPolicy` references in `codex-rs/core/tests` from 27 files after #20013 to 22 files.

## Testing
- `just fmt`

### DIFF
--- a/codex-rs/core/tests/suite/code_mode.rs
+++ b/codex-rs/core/tests/suite/code_mode.rs
@@ -11,10 +11,10 @@ use codex_models_manager::bundled_models_response;
 use codex_protocol::dynamic_tools::DynamicToolCallOutputContentItem;
 use codex_protocol::dynamic_tools::DynamicToolResponse;
 use codex_protocol::dynamic_tools::DynamicToolSpec;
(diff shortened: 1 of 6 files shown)


bolinfest:
[codex] Addressed in 5f618ba516fc, which is in this stack below #20015. Remote start/resume/fork sandbox-mode mapping now treats any managed profile with writable roots as workspace-write instead of only checking whether cwd itself is writable; added coverage for a non-cwd write root.

PR #20023, opened 2026-04-28
Respect filesystem read policy for fileParams uploads

colby-oai:
Adds a filesystem read-policy check before Codex uploads local files for `openai/fileParams` MCP tool arguments. If a resolved file path is not readable under the current turn’s sandbox policy, the tool call now fails before opening or uploading the file. Includes regression coverage for deny-read behavior.

### DIFF
--- a/codex-rs/core/src/mcp_openai_file.rs
+++ b/codex-rs/core/src/mcp_openai_file.rs
@@ -112,19 +112,6 @@ async fn build_uploaded_local_argument_value(
     file_path: &str,
 ) -> Result<JsonValue, String> {
     let resolved_path = turn_context.resolve_path(Some(file_path.to_string()));
-    if !turn_context
-        .file_system_sandbox_policy()
-        .can_read_path_with_cwd(resolved_path.as_path(), turn_context.cwd.as_path())
-    {
-        return Err(match index {
-            Some(index) => format!(
-                "reading `{file_path}` for `{field_name}[{index}]` is blocked by the filesystem sandbox policy"
-            ),
-            None => format!(
-                "reading `{file_path}` for `{field_name}` is blocked by the filesystem sandbox policy"
-            ),
-        });
-    }
     let Some(auth) = auth else {
         return Err(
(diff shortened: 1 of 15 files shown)


codex-rs/core/src/mcp_openai_file.rs
```diff
@@ -103,6 +103,19 @@ async fn build_uploaded_local_argument_value(
     file_path: &str,
 ) -> Result<JsonValue, String> {
     let resolved_path = turn_context.resolve_path(Some(file_path.to_string()));
+    if !turn_context
+        .file_system_sandbox_policy()
+        .can_read_path_with_cwd(resolved_path.as_path(), turn_context.cwd.as_path())
```
viyatb-oai: This still performs a policy pre-check in core and then `upload_local_file()` reopens the path with ordinary local filesystem APIs (`tokio::fs::metadata` / `File::open`). That is the unsafe part: the actual dereference/read is not happening through the sandboxed exec-server filesystem path, so a symlink target or swapped path can differ from what the pre-check approved.

Fix: make the file read go through the exec-server filesystem abstraction with the turn sandbox context, then upload the bytes that came back from that operation. Concretely:

- build the sandbox context from the turn, e.g. `let sandbox = turn_context.file_system_sandbox_context(/*additional_permissions*/ None);`;
- get the active environment filesystem and call its sandboxed read, e.g. `environment.get_filesystem().read_file(&resolved_path, Some(&sandbox)).await`;
- feed those bytes, size, and file name into a new `codex_api` upload helper that uploads an already-open/read byte stream instead of reopening the path;
- remove the separate `can_read_path_with_cwd()` guard and the later unsandboxed `tokio::fs::metadata` / `File::open` path.

This preserves symlink behavior safely: if the symlink target is readable under the sandbox, the exec-server fs read returns the bytes; if it points at denied content, that read fails. Please include a regression test for a readable-looking symlink that targets a denied file.

starr-openai (approved):

PR #20049, opened 2026-04-28
feat: expose provider capability bounds to app server clients

celia-oai:
follow up of #19442. The app server now exposes provider-derived bounds through a new v2 `modelProvider/read` method. The response reports the configured provider map key as `modelProvider` and returns the effective capability booleans so clients can align their UI with the same provider-owned limits used by core.


codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -496,6 +496,10 @@ client_request_definitions! {
         params: v2::ModelListParams,
         response: v2::ModelListResponse,
     },
+    ModelProviderRead => "modelProvider/read" {
```
shijie-oai: Should we do something like `activeModelProvider/read`? `modelProvider/read` sounds like we can pass in a model provider name and get the object back. Also wondering if we want to scope it down more (i.e. `modelProvider/capability/read`) or keep it high level for future extendability.

celia-oai: good call, will change it to `currentModelProvider/read`

codex-rs/app-server-protocol/src/protocol/common.rs
```diff
@@ -496,6 +496,10 @@ client_request_definitions! {
         params: v2::ModelListParams,
         response: v2::ModelListResponse,
     },
+    CurrentModelProviderRead => "currentModelProvider/read" {
```
pakrym-oai: modelProvider/read?

celia-oai: i changed it from that because of Shijie's comment above haha: https://github.com/openai/codex/pull/20049#discussion_r3156546428

owenlin0: I'm not a fan of `current*` because you can apply that reasoning to things like `currentAccount/read`, `currentConfig/read`, etc. and it just feels boilerplatey

owenlin0: `modelProviderCapabilities/read` ?

celia-oai: let's do `modelProvider/capabilities/read`

owenlin0: agreed

pakrym-oai (approved):

PR #20050, opened 2026-04-28
[app-server] type client response payloads

rhan-oai:
## Why

`pr17088` adds typed server-originated request/response plumbing, but successful client responses are still erased into bare JSON-RPC `result` values before app-server can make any typed decision about them.

This precursor PR keeps successful client responses typed until the outgoing response seam. It is intentionally limited to protocol/app-server plumbing so the analytics behavior change can review separately on top.

## What changed

- Add `ClientResponsePayload` as the pre-serialization client response body type.
- Route app-server successful response paths through the typed payload seam while preserving existing handler-local analytics behavior.
- Keep `InterruptConversation` JSON-RPC-only because it has no `ClientResponse` variant.
- Move the new payload conversion tests into a dedicated protocol test module.

## Verification

- `cargo check -p codex-app-server`

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -43,6 +43,7 @@ use codex_app_server_protocol::CancelLoginAccountResponse;
 use codex_app_server_protocol::CancelLoginAccountStatus;
 use codex_app_server_protocol::ClientRequest;
 use codex_app_server_protocol::ClientResponse;
+use codex_app_server_protocol::ClientResponsePayload;
 use codex_app_server_protocol::CodexErrorInfo;
 use codex_app_server_protocol::CollaborationModeListParams;
 use codex_app_server_protocol::CollaborationModeListResponse;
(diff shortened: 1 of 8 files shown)


codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -1395,7 +1376,9 @@ impl CodexMessageProcessor {
             .await
             .map(|()| LoginAccountResponse::ApiKey {});
         let logged_in = result.is_ok();
-        self.outgoing.send_result(request_id, result).await;
+        self.outgoing
+            .send_result(request_id, result, ClientResponsePayload::LoginAccount)
+            .await;
```
owenlin0: Is there a way to avoid updating every callsite that sends an RPC response with the payload type? This seems boilerplatey, especially since the `result` is typed already.

Codex suggests:
> Yes. The cleanest reduction is to make the protocol macro generate From<ResponseType> for ClientResponsePayload for response types that are unambiguous, then change send_result/send_response to accept Into<ClientResponsePayload>.

owenlin0: There's one annoying gotcha, which is that we have at least one reused response type:
```
ConfigValueWrite => response: v2::ConfigWriteResponse
ConfigBatchWrite => response: v2::ConfigWriteResponse
```

but maybe there's a fix/workaround for this one?

owenlin0 (approved):

PR #20053, opened 2026-04-28
Add generic plugin backend bridge

alandelong-oai:
## Why

Some trusted local plugins need to call authenticated backend routes while keeping the user's ChatGPT credentials inside the Codex host. Without a host-owned handoff, each plugin would either need to grow its own auth flow or receive credentials directly, which is a worse boundary for local plugin processes.

## What changed

- add config-driven named loopback bridges under `plugin_backend_bridges`
- write a random-token state file for each configured bridge so a local plugin can discover its listener without receiving user credentials
- forward only requests under the configured local/backend path prefixes, using the signed-in Codex account for backend auth
- keep the feature opt-in: no bridge starts unless it is explicitly configured
- document the config shape and regenerate the config schema

## Verification

- `cargo test -p codex-app-server plugin_backend_bridge`
- `cargo test -p codex-core config_schema`

### DIFF
--- a/codex-rs/Cargo.lock
+++ b/codex-rs/Cargo.lock
@@ -1892,6 +1892,7 @@ dependencies = [
  "opentelemetry_sdk",
  "owo-colors",
  "pretty_assertions",
+ "rand 0.9.3",
  "reqwest",
  "rmcp",
  "serde",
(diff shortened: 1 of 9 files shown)


alandelong-oai:

alandelong-oai:
Closing this for now. We are pivoting the prototype to reuse existing Codex auth state directly while we align with the Codex team on the longer-term plugin design path.

PR #20058, opened 2026-04-28
Add environment provider snapshot

starr-openai:
## Summary
- Change `EnvironmentProvider` to return concrete `Environment` instances instead of `EnvironmentConfigurations`.
- Make `DefaultEnvironmentProvider` provide the provider-visible `local` environment plus optional `remote` environment from `CODEX_EXEC_SERVER_URL`.
- Keep `EnvironmentManager` as the concrete cache while exposing its own explicit local environment for `local_environment()` fallback paths.

## Validation
- `just fmt`
- `git diff --check`


codex-rs/exec-server/src/environment_provider.rs
```diff
@@ -0,0 +1,264 @@
...
+    pub fn new(exec_server_url: Option<String>) -> Self {
+        Self { exec_server_url }
+    }
+
+    /// Builds a provider by reading `CODEX_EXEC_SERVER_URL`.
+    pub fn from_env() -> Self {
+        Self::new(std::env::var(CODEX_EXEC_SERVER_URL_ENV_VAR).ok())
+    }
+
+    pub(crate) fn environment_configurations(&self) -> EnvironmentConfigurations {
```
pakrym-oai: sholdn't this method be async?

codex-rs/exec-server/src/environment.rs
```diff
@@ -83,27 +85,56 @@ impl EnvironmentManager {
             local_runtime_paths,
```
pakrym-oai, line 91: should we drop exec_server_url from EnvironmentManagerArgs now that provider owns that?

codex-rs/exec-server/src/environment_provider.rs
starr-openai: good catch

codex-rs/exec-server/src/environment.rs
```diff
@@ -75,35 +68,93 @@ impl EnvironmentManager {
...
+        .to_string();
+        let mut environments = HashMap::new();
+        for (id, configuration) in environment_configurations {
+            let environment = Environment::remote_inner(
+                configuration.exec_server_url,
+                Some(local_runtime_paths.clone()),
+            );
+            environments.insert(id, Arc::new(environment));
+        }
+        environments.insert(
```
pakrym-oai: interesting that this is not returned by the provider

pakrym-oai (approved):
DefaultProvider should probably also return the local one.
Can we return the full `Environment`?

codex-rs/exec-server/src/environment.rs
starr-openai: we have some places that depend on local env being there via `local_environment`, so I put this in EnvironmentManager itself to guarantee its set and configured properly.  Then the env provider just provides external ones.

this is why:
> DefaultProvider should probably also return the local one.
doesn't.

PR #19432, opened 2026-04-24
[codex] Add token usage to turn tracing spans

charley-openai:
## Why

Slow Codex turns are easier to debug when token usage is visible in the trace itself, without joining against separate analytics. This adds token usage to existing turn-handling spans for regular user turns only.

[Example turn](https://openai.datadoghq.com/apm/trace/9d353efa2cb5de1f4c5b93dc33c3df04?colorBy=service&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=3555541504891512675&spanViewType=metadata&traceQuery=)
Screenshot 2026-04-24 at 3 03 07 PM

## What Changed

Added response-level token fields on completed handle_responses spans:

gen_ai.usage.input_tokens
gen_ai.usage.cache_read.input_tokens
gen_ai.usage.output_tokens
codex.usage.reasoning_output_tokens
codex.usage.total_tokens
Added aggregate token fields on regular turn spans:

codex.turn.token_usage.*
Added an explicit regular-turn opt-in via SessionTask::records_turn_token_usage_on_span() so this is not coupled to span-name strings.

## Testing

- `cargo test -p codex-otel`
- `cargo test -p codex-core turn_and_completed_response_spans_record_token_usage`
- `just fmt`
- `just fix -p codex-core`
- `just fix -p codex-otel`
- Manual local Electron/app-server smoke test: regular user turn emits the new span fields

Known status: `cargo test -p codex-core` was attempted and failed in unrelated existing areas: config approvals, request-permissions, git-info ordering, and subagent metadata persistence.

PR #20060, opened 2026-04-28
Add reasoning effort to turn tracing spans

charley-openai:
Why
#19432 added token usage to the turn and response spans. This follow-up adds the configured reasoning effort so performance traces can be filtered by model effort.

[example trace](https://openai.datadoghq.com/apm/trace/1ff708a87159ff4898bdc8bd6091ec18?graphType=waterfall&shouldShowLegend=true&spanID=6596351544047485652&traceQuery=)
Screenshot 2026-04-28 at 3 52 12 PM

What Changed
- Adds `codex.turn.reasoning_effort` to the turn span.
- Adds `codex.request.reasoning_effort` to `handle_responses`.
- Extends the span test to cover explicit `high` effort with token usage.

Testing
- `cargo test -p codex-core turn_and_completed_response_spans_record_token_usage`
- `cargo test -p codex-otel`
- `just fmt`
- `just fix -p codex-core`
- `just fix -p codex-otel`

### DIFF
--- a/codex-rs/core/src/tasks/mod.rs
+++ b/codex-rs/core/src/tasks/mod.rs
@@ -368,12 +368,14 @@ impl Session {
         let task_cancellation_token = cancellation_token.child_token();
         // Task-owned turn spans keep a core-owned span open for the
         // full task lifecycle after the submission dispatch span ends.
+        let reasoning_effort = turn_context.effective_reasoning_effort_for_tracing();
         let task_span = info_span!(
             "turn",
             otel.name = span_name,
             thread.id = %self.conversation_id,
             turn.id = %turn_context.sub_id,
             model = %turn_context.model_info.slug,
+            codex.turn.reasoning_effort = %reasoning_effort,
             codex.turn.token_usage.input_tokens = field::Empty,
             codex.turn.token_usage.cached_input_tokens = field::Empty,
             codex.turn.token_usage.non_cached_input_tokens = field::Empty,
--- a/codex-rs/core/tests/suite/otel.rs
+++ b/codex-rs/core/tests/suite/otel.rs
@@ -1,6 +1,7 @@
 use codex_core::config::Constrained;
 use codex_features::Feature;
 use codex_protocol::models::PermissionProfile;
+use codex_protocol::openai_models::ReasoningEffort;
 use codex_protocol::protocol::AskForApproval;
 use codex_protocol::protocol::EventMsg;
 use codex_protocol::protocol::Op;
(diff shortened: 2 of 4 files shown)


codex-rs/core/src/tasks/mod.rs
```diff
@@ -661,6 +662,11 @@ impl Session {
             };
             if records_turn_token_usage_on_span {
```
jif-oai, line 664: Can we record codex.turn.reasoning_effort when the span is created instead?

codex-rs/core/tests/suite/otel.rs
```diff
@@ -806,22 +831,24 @@ async fn record_responses_sets_span_fields_for_response_events() {
...
+        let from_field = from.map(|from| format!("from=\"{from}\""));
+        let tool_name_field = tool_name.map(|tool_name| format!("tool_name=\"{tool_name}\""));
+
         assert!(
-            logs.contains(&format!("handle_responses{{otel.name=\"{name}\"")),
-            "missing otel.name={name}\nlogs:\n{logs}"
+            logs.lines().any(|line| {
+                line.contains("handle_responses{")
+                    && line.contains(&otel_name)
+                    && line.contains("codex.request.reasoning_effort=xhigh")
```
jif-oai: This test is now coupled to gpt-5.5's current default effort. A catalog-only metadata change will fail it without any tracing regression, so I'd rather set the effort explicitly than assert the model default

codex-rs/core/src/tasks/mod.rs
charley-openai: yes!

codex-rs/core/tests/suite/otel.rs
charley-openai: good point, changed.

codex-rs/core/src/tasks/mod.rs
```diff
@@ -366,21 +366,34 @@ impl Session {
...
-            turn.id = %turn_context.sub_id,
-            model = %turn_context.model_info.slug,
-            codex.turn.token_usage.input_tokens = field::Empty,
-            codex.turn.token_usage.cached_input_tokens = field::Empty,
-            codex.turn.token_usage.non_cached_input_tokens = field::Empty,
-            codex.turn.token_usage.output_tokens = field::Empty,
-            codex.turn.token_usage.reasoning_output_tokens = field::Empty,
-            codex.turn.token_usage.total_tokens = field::Empty,
-        );
+        let task_span = if records_turn_token_usage_on_span {
```
jif-oai: Why this `if`?

jif-oai: This does not seem related to this PR

jif-oai (approved):
Ok after my comment

codex-rs/core/src/tasks/mod.rs
charley-openai: good call, i'll remove it.

PR #20064, opened 2026-04-28
Include auto-review rollout in feedback uploads

won-openai:
## Summary

- include the live auto-review trunk rollout when `/feedback` uploads logs
- upload that attachment as `auto-review-rollout-<parent-thread-id>.jsonl` so it is distinguishable from the parent rollout
- show the same auto-review attachment name in the TUI consent popup

## Scope

- this only covers the live cached auto-review trunk for the current parent thread
- it does not add durable historical parent->auto-review lookup
- it does not add persisted rollout support for ephemeral parallel review forks

## UI 

Screenshot 2026-04-28 at 1 17 18 PM

## Validation

- `cargo test -p codex-core cached_guardian_subagent_exposes_its_rollout_path`
- `cargo test -p codex-feedback`
- `cargo test -p codex-tui feedback_upload_consent_popup_snapshot`
- `cargo test -p codex-tui feedback_good_result_consent_popup_includes_connectivity_diagnostics_filename`

## Known unrelated local failures

- `cargo test -p codex-core` currently fails in the pre-existing proxy env snapshot test `tools::runtimes::tests::maybe_wrap_shell_lc_with_snapshot_keeps_user_proxy_env_when_proxy_inactive`
- `cargo test -p codex-tui` currently hits pre-existing `status::*` snapshot drift unrelated to this change

## Follow-Up 
- persist parallel auto-review fork sessions so /feedback can include their rollout history too
- attach each persisted fork as its own clearly named file, for example auto-review-rollout-<parent-thread-id>-fork <n>.jsonl, instead of merging multiple Guardian sessions into one attachment
- keep the same live-session-only scope initially; durable historical parent -> auto-review lookup can remain a    separate decision if we later need feedback from resumed sessions

### DIFF
--- a/codex-rs/feedback/src/lib.rs
+++ b/codex-rs/feedback/src/lib.rs
@@ -338,12 +338,17 @@ pub struct FeedbackSnapshot {
     pub thread_id: String,
 }
 
+pub struct FeedbackAttachmentPath {
+    pub path: PathBuf,
+    pub filename: Option<String>,
+}
+
 pub struct FeedbackUploadOptions<'a> {
     pub classification: &'a str,
     pub reason: Option<&'a str>,
     pub tags: Option<&'a BTreeMap<String, String>>,
     pub include_logs: bool,
-    pub extra_attachment_paths: &'a [PathBuf],
+    pub extra_attachment_paths: &'a [FeedbackAttachmentPath],
     pub session_source: Option<SessionSource>,
     pub logs_override: Option<Vec<u8>>,
 }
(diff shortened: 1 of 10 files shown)


codex-rs/feedback/src/lib.rs
```diff
@@ -338,12 +338,17 @@ pub struct FeedbackSnapshot {
     pub thread_id: String,
 }
 
+pub struct FeedbackAttachmentPath {
+    pub path: PathBuf,
+    pub filename: Option<String>,
```
dylan-hurd-oai: It seems like this is an optional overwrite for the filename on `path`? Can we make this clearer, e.g. at least documenting on the struct / renaming the var?

dylan-hurd-oai (commented):
quick clarification about FeedbackAttachmentPath, but otherwise looks reasonable!

codex-rs/feedback/src/lib.rs
won-openai: yes. sorry for the lack of clarification!

dylan-hurd-oai (approved):

PR #20069, opened 2026-04-28
Bypass review for always-allow MCP tools in auto-review

maja-openai:
## Why

When an MCP or app tool is configured with approval mode `approve` (always allow), users expect that decision to be authoritative. In guardian auto-review mode, ARC could still return `ask-user`, which then routed the approval question into guardian with the ARC reason as context. That meant a tool explicitly configured as always allowed still went through both safety monitors before running.

This change keeps the existing ARC behavior for non-auto-review sessions, but avoids the ARC-to-guardian sequence when `approvals_reviewer = auto_review` and the tool approval mode is `approve`.

## What changed

- Short-circuit MCP tool approval handling when `approval_mode == approve` and `approvals_reviewer == auto_review`.
- Updated the MCP approval regression test so the auto-review case asserts neither ARC nor guardian is called.
- Preserved existing tests that verify ARC can still block always-allow MCP tools outside guardian auto-review mode.

## Verification

- `cargo test -p codex-core --lib mcp_tool_call`


codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -888,6 +889,12 @@ async fn maybe_request_mcp_tool_approval(
         return None;
     }
 
+    if approval_mode == AppToolApproval::Approve
```
mzeng-openai: nit: Can we move this into mcp_permission_prompt_is_auto_approved?

maja-openai: Done, but that does require me tho thread in more context. PTAL

mzeng-openai (approved):

codex-rs/codex-mcp/src/mcp/mod.rs
```diff
@@ -67,7 +69,17 @@ pub fn qualified_mcp_tool_name_prefix(server_name: &str) -> String {
 pub fn mcp_permission_prompt_is_auto_approved(
     approval_policy: AskForApproval,
     permission_profile: &PermissionProfile,
+    context: McpPermissionPromptAutoApproveContext,
 ) -> bool {
+    if matches!(
+        approval_policy,
+        AskForApproval::OnRequest | AskForApproval::Granular(_)
+    ) && context.approvals_reviewer == Some(ApprovalsReviewer::AutoReview)
```
owenlin0, line 77: hmm do we need this line? `&& context.approvals_reviewer == Some(ApprovalsReviewer::AutoReview)`

should `approve` in the tool context just mean approve whether it's a human or auto-reviewer? cc @mzeng-openai

owenlin0 (approved):

codex-rs/codex-mcp/src/mcp/mod.rs
owenlin0: oh, is it due to this behavior? 

> Preserved existing tests that verify ARC can still block always-allow MCP tools outside guardian auto-review mode.

mzeng-openai: That's my read as well. I'd get feedback from @fouad-openai and see if we can get rid of this as well as a follow up, but not blocking this change.

maja-openai: [codex] Yes, that guard is intentional for this PR scope. The requested behavior is to bypass ARC only for always-allow tools when approvals are routed to guardian auto-review; non-auto-review approval modes should preserve the existing ARC path so ARC can still block. I agree the broader policy could be revisited as a follow-up with @fouad-openai.

PR #20072, opened 2026-04-28
Support disabling tool suggest for specific tools.

mzeng-openai:
## Summary
- Add `disable_tool_suggest` to app and plugin config, schema, and TypeScript output
- Exclude disabled connectors and plugins from tool suggestion discovery
- Persist "never show again" tool-suggestion choices back into `config.toml`
- Update config docs and add coverage for connector and plugin suppression

## Testing
- Added and updated unit tests for config persistence and tool-suggest filtering
- Not run (not requested)

### DIFF
--- a/codex-rs/core/src/tools/handlers/tool_suggest.rs
+++ b/codex-rs/core/src/tools/handlers/tool_suggest.rs
@@ -1,6 +1,7 @@
 use std::collections::HashSet;
 
 use codex_app_server_protocol::AppInfo;
+use codex_config::types::ToolSuggestDisabledTool;
 use codex_mcp::CODEX_APPS_MCP_SERVER_NAME;
 use codex_rmcp_client::ElicitationAction;
 use codex_rmcp_client::ElicitationResponse;
@@ -18,7 +19,6 @@ use codex_tools::filter_tool_suggest_discoverable_tools_for_client;
 use codex_tools::verified_connector_suggestion_completed;
 use rmcp::model::RequestId;
 use serde_json::Value;
-use toml_edit::value;
 use tracing::warn;
 
 use crate::config::edit::ConfigEdit;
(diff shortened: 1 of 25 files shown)


xli-oai (approved):

codex-rs/core/src/tools/handlers/tool_suggest.rs
```diff
@@ -123,6 +130,9 @@ impl ToolHandler for ToolSuggestHandler {
         let response = session
             .request_mcp_server_elicitation(turn.as_ref(), request_id, params)
             .await;
+        if let Some(response) = response.as_ref() {
+            maybe_persist_tool_suggest_disable(&session, &turn, &tool, response).await;
+        }
```
sayan-oai, line 135: what happens if the elicitation action is `Accept` and `persist = "always"`? should we gate on action `Decline`?

mzeng-openai: good catch, we don't need to persist accept.
I've also refactored the config shape since codex found out that creating ghost entries for apps and plugins can cause more side effects than I thought. Right now consolidating the disabled tool list under [tool_suggest]

codex-rs/core/src/tools/handlers/tool_suggest.rs
```diff
@@ -158,6 +168,64 @@ impl ToolHandler for ToolSuggestHandler {
...
+        DiscoverableTool::Connector(connector) => ("apps", connector.id.as_str()),
+        DiscoverableTool::Plugin(plugin) => ("plugins", plugin.id.as_str()),
+    };
+
+    ConfigEdit::SetPath {
+        segments: vec![
+            root.to_string(),
+            tool_id.to_string(),
+            "disable_tool_suggest".to_string(),
+        ],
```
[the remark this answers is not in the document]
mzeng-openai: refactored the config so that disabled tools are listed under [tool_suggest]

codex-rs/core/src/tools/handlers/tool_suggest.rs
sayan-oai: gotcha. is the list only created in user-level config, or does it map back to the config layer the tool comes from?

codex says: if the list always writes to user config, a list defined in a higher-precedence layer (like project-level) would clobber the user config list because they don't merge.

mzeng-openai: That makes sense! I think it's more of a "how to merge" issue rather than config location thing, added logic to support merging the disabled tool list from higher precedence.

sayan-oai (approved):

mzeng-openai:
Made some small edits so that we can hot-reload any disabled_tools changes after the tool suggest elicitation.

xl-openai (approved):

PR #20080, opened 2026-04-28
[codex-analytics] prevent stale guardian events from satisfying reused reviews

rhan-oai:
## Why

Reused Guardian review trunks can still have older child-turn events queued when a later review starts. The review waiter currently accepts the first terminal event it sees from the shared child session, so a stale `TurnComplete` can be attributed to the new review. That produces impossible analytics combinations such as non-null TTFT with sub-10 ms completion latency and zero token deltas on `trunk_reused` reviews.

## What changed

- Preserve the child turn id returned by the Guardian review `Op::UserTurn` submission.
- Restrict Guardian review waiting to events correlated with that submitted child turn.
- Restrict timeout/abort draining to terminal events for the same child turn.
- Add regression coverage for stale prior-turn completions, stale prior-turn errors, and interrupt draining in `codex-rs/core/src/guardian/review_session.rs`.

## Verification

- `cargo test -p codex-core guardian::review_session::tests::`
- `cargo clippy -p codex-core --tests -- -D warnings`

### DIFF
--- a/codex-rs/core/src/guardian/review_session.rs
+++ b/codex-rs/core/src/guardian/review_session.rs
@@ -13,6 +13,7 @@ use codex_protocol::models::PermissionProfile;
 use codex_protocol::models::ResponseItem;
 use codex_protocol::openai_models::ReasoningEffort as ReasoningEffortConfig;
 use codex_protocol::protocol::AskForApproval;
+use codex_protocol::protocol::Event;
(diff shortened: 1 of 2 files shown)


dylan-hurd-oai (commented):
Can we add an integration test for this?

dylan-hurd-oai (approved):

PR #20081, opened 2026-04-28
[codex] Expose used plugin IDs in turn metadata

pab-oai:
## Why
Pulse targeting needs to be able to make request-time decisions based on which plugins were used for a Codex turn. Today that information is not present in `x-codex-turn-metadata`, so Pulse cannot inspect it from the live request context.

## What changed
- Add `plugin_ids_used` to serialized turn metadata.
- Populate it with the canonical plugin IDs used for the turn.
- Keep the emitted IDs deduplicated so request-time consumers receive a stable field.

## Validation
- `cargo fmt --check`
- `cargo test -p codex-core turn_metadata_state_ --lib`


pab-oai:

PR #20085, opened 2026-04-28
fix: don't auto approve git -C ...

owenlin0:
It's safer to make sure these commands go through approval flows.

### DIFF
--- a/codex-rs/shell-command/src/command_safety/is_dangerous_command.rs
+++ b/codex-rs/shell-command/src/command_safety/is_dangerous_command.rs
@@ -58,7 +58,8 @@ fn is_git_global_option_with_inline_value(arg: &str) -> bool {
 pub(crate) fn git_global_option_requires_prompt(arg: &str) -> bool {
     matches!(
         arg,
-        "-c" | "--config-env"
+        "-C" | "-c"
+            | "--config-env"
             | "--exec-path"
             | "--git-dir"
             | "--namespace"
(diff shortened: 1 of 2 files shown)


codex-rs/shell-command/src/command_safety/is_safe_command.rs
```diff
@@ -335,7 +335,7 @@ mod tests {
 
         assert_eq!(
             is_known_safe_command(&vec_str(&["git", "-C", ".", "branch", "--show-current"])),
-            true
+            false
```
bolinfest: I know you have to update this to make the test pass, but I don't think this illustrated what this test was trying to prove anymore. I think the idea is that `git branch --show-current` should be safe/read-only whereas `git branch -d feature` is not.

codex-rs/shell-command/src/command_safety/is_dangerous_command.rs
```diff
@@ -58,15 +58,16 @@ fn is_git_global_option_with_inline_value(arg: &str) -> bool {
 pub(crate) fn git_global_option_requires_prompt(arg: &str) -> bool {
     matches!(
         arg,
-        "-c" | "--config-env"
+        "-C" | "-c"
```
bolinfest, line 63: Somewhere in here we should comment on why we reject when we see `-C`.

codex-rs/shell-command/src/command_safety/is_safe_command.rs
owenlin0: yeah good call, let me rewrite this to preserve that intention

bolinfest: Incidentally, Rust prefers using `assert!(thing)` to `assert_eq!(thing, true)` and using `assert!(!thing)` to `assert_eq!(!thing, false)` and `pretty_assertions::assert_eq` isn't doing much for us here AFAICT so it's weird that it's imported just for this function.

bolinfest (approved):

PR #20086, opened 2026-04-28
Fix plugin list workspace settings test isolation

canvrno-oai:
Fixes test that often fails locally when running `cargo test`
- Add an app-server test helper that combines managed-config isolation with custom env overrides.
- Isolate `HOME` / `USERPROFILE` in plugin-list workspace settings tests so host home marketplaces do not affect results.


canvrno-oai:
@codex review

abhinav-oai (approved):

Issue #18248, opened 2026-04-17
Windows app and CLI child processes start with missing core env vars, breaking dotnet/NuGet and git networking


albertc71:
## Summary
On Windows x64, both Codex App and Codex CLI appear to launch child processes with an incomplete/broken environment block.

This breaks common tools in ways that do not reproduce in a normal Windows terminal on the same machine.

The clearest symptoms I reproduced were:
- `dotnet build` / restore failing inside NuGet with `Value cannot be null. (Parameter 'path1')`
- `git fetch` failing with `getaddrinfo() thread failed to start`
- `dotnet --info` / `dotnet nuget locals all -l` behaving incorrectly until core Windows env vars are manually restored

I can reproduce the same issue in both the Codex App and Codex CLI.

See the follow-up comment(s) on this issue for sanitized failing and succeeding command logs that may be useful for debugging.

## Versions / Environment
- Codex App: `26.409.61251`
- Codex CLI: `0.121.0` (`codex -V` -> `codex-cli 0.121.0`)
- Subscription: `Pro x5`
- Platform: `Windows x64`
- CLI config file: `%USERPROFILE%\.codex\config.toml`

## Relevant Codex config
This reproduces even with a non-restrictive config.

Sanitized relevant settings from `%USERPROFILE%\.codex\config.toml`:

```toml
model = "gpt-5.4"
model_reasoning_effort = "xhigh"
sandbox_mode = "danger-full-access"
approval_policy = "never"

[shell_environment_policy]
inherit = "core"

[sandbox_workspace_write]
network_access = true

[windows]
sandbox = "unelevated"
```

This is useful context because the failure still occurs with:
- `sandbox_mode = "danger-full-access"`
- `approval_policy = "never"`
- `network_access = true`

So the issue does not appear to require a restrictive filesystem sandbox to reproduce.

I also explicitly tried using:

```toml
[windows]
sandbox = "elevated"
```

Per the official Windows docs, `elevated` is the preferred native Windows sandbox and `unelevated` is the fallback. In my environment, `elevated` does not work correctly either, and Codex appears to fall back / fail during sandbox setup rather than resolving the underlying issue.

## What appears to be wrong
Inside the Codex-launched process environment, standard Windows/user env vars are empty by default.

Observed empty vars from the Codex-run shell:
- `APPDATA=`
- `LOCALAPPDATA=`
- `ProgramFiles=`
- `ProgramFiles(x86)=`
- `SystemRoot=`
- `ComSpec=`
[...]

PR #20089, opened 2026-04-28
expand the set of core shell env vars for Windows.

iceweasel-oai:
https://github.com/openai/codex/issues/13917 and https://github.com/openai/codex/issues/18248 correctly identify that 

```
[shell_environment_policy]
inherit = "core"
```
is not functional on Windows because it carries an insufficient set of env vars. 
This PR expands that to match the more functional set from the MCP client

### DIFF
--- a/codex-rs/protocol/src/shell_environment.rs
+++ b/codex-rs/protocol/src/shell_environment.rs
@@ -58,27 +58,17 @@ where
         ShellEnvironmentPolicyInherit::None => HashMap::new(),
         ShellEnvironmentPolicyInherit::Core => {
             #[cfg(not(target_os = "windows"))]
-            {
-                let core_vars: std::collections::HashSet<&str> = COMMON_CORE_VARS
-                    .iter()
-                    .copied()
-                    .chain(PLATFORM_CORE_VARS.iter().copied())
-                    .collect();
-                vars.into_iter()
-                    .filter(|(k, _)| core_vars.contains(k.as_str()))
-                    .collect()
-            }
-
+            let core_env_vars = UNIX_CORE_VARS;
             #[cfg(target_os = "windows")]
-            {
-                vars.into_iter()
-                    .filter(|(k, _)| {
-                        WINDOWS_CORE_ENV_VARS
-                            .iter()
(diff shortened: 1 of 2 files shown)


codex-rs/protocol/src/shell_environment.rs
```diff
@@ -113,10 +119,104 @@ where
...
+    // Temp locations
+    "TEMP",
+    "TMP",
+    "TMPDIR",
+    // Common shells/pwsh hints
+    "POWERSHELL",
+    "PWSH",
+];
+
+#[cfg(not(target_os = "windows"))]
```
bolinfest: This should be listed first.

codex-rs/protocol/src/shell_environment.rs
```diff
@@ -113,10 +119,104 @@ where
...
+const COMMON_CORE_VARS: &[&str] = &["PATH", "SHELL", "TMPDIR", "TEMP", "TMP"];
 
-#[cfg(unix)]
+#[cfg(not(target_os = "windows"))]
```
bolinfest: It's a bit confusing that `PLATFORM_CORE_VARS` is defined for only one platform now?

codex-rs/protocol/src/shell_environment.rs
```diff
@@ -58,21 +57,28 @@ where
...
+                    .collect()
+            }
+
+            #[cfg(target_os = "windows")]
+            {
+                vars.into_iter()
+                    .filter(|(k, _)| {
+                        WINDOWS_CORE_ENV_VARS
+                            .iter()
+                            .any(|allowed| allowed.eq_ignore_ascii_case(k))
```
bolinfest: The `eq_ignore_ascii()` thing was originally written as Windows-only, though some utilities like `curl` recognizes both lowercase and uppercase for certain variables on UNIX:

```
HTTP_PROXY
http_proxy
HTTPS_PROXY
https_proxy
NO_PROXY
no_proxy
```

so maybe we should simplify things and do the `eq_ignore_ascii_case()` check on both platforms?

iceweasel-oai: done

bolinfest (approved):
Nice: ty!

PR #20091, opened 2026-04-28
[tool_suggest] Improve tool_suggest triggering conditions.

mzeng-openai:
## Summary
- Tighten `tool_suggest` guidance so it prefers explicit plugin install requests, while still allowing a connector install when the relevant plugin is already installed and a needed connector from that plugin is missing.
- Tell the model not to call `tool_suggest` in parallel with other tools.

## Testing
- `cargo test -p codex-tools tool_suggest`
- `cargo test -p codex-core tool_suggest`


codex-rs/tools/src/tool_discovery.rs
```diff
@@ -293,9 +292,10 @@ pub fn create_tool_suggest_tool(discoverable_tools: &[ToolSuggestEntry]) -> Tool
         ),
         (
             "tool_id".to_string(),
-            JsonSchema::string(Some(format!(
-                "Connector or plugin id to suggest. Must be one of: {discoverable_tool_ids}."
-            ))),
+            JsonSchema::string_enum(
```
mzeng-openai: Hard constrain the valid tool ids.

codex-rs/tools/src/tool_registry_plan.rs
```diff
@@ -309,7 +309,7 @@ pub fn build_tool_registry_plan(
     {
         plan.push_spec(
             create_tool_suggest_tool(&collect_tool_suggest_entries(discoverable_tools)),
-            /*supports_parallel_tool_calls*/ true,
+            /*supports_parallel_tool_calls*/ false,
```
mzeng-openai: Disable parallel tool call

codex-rs/tools/src/tool_discovery.rs
pakrym-oai: does this value dynamically change during the thread timeline?

codex-rs/tools/src/tool_registry_plan.rs
pakrym-oai: We shouldn't do this. This controls how harness executes the tool, not how the model samples it.

codex-rs/tools/src/tool_discovery.rs
mzeng-openai: If the installable/installed set of plugins change, then yes. Would you suggest we change to to unconstrained?

sayan-oai: maybe revert to soft constraint as per discussion in slack?

mzeng-openai: reverted

codex-rs/tools/src/tool_discovery.rs
```diff
@@ -287,15 +282,11 @@ pub fn create_tool_suggest_tool(discoverable_tools: &[ToolSuggestEntry]) -> Tool
         ),
         (
             "action_type".to_string(),
-            JsonSchema::string(Some(
-                "Suggested action for the tool. Use \"install\" or \"enable\".".to_string(),
-            )),
+            JsonSchema::string(Some("Suggested action for the tool. Use \"install\".".to_string())),
```
sayan-oai, line 285: we no longer support `enable`? is there an associated logic change required, or did `enable` map to `install` logic anyways?

codex-rs/tools/src/tool_discovery.rs
```diff
@@ -287,15 +282,11 @@ pub fn create_tool_suggest_tool(discoverable_tools: &[ToolSuggestEntry]) -> Tool
...
         ),
         (
             "tool_id".to_string(),
-            JsonSchema::string(Some(format!(
-                "Connector or plugin id to suggest. Must be one of: {discoverable_tool_ids}."
-            ))),
+            JsonSchema::string(Some("Connector or plugin id to suggest.".to_string())),
```
sayan-oai, line 289: why are we omitting the discoverable_tool_ids here? IIUC we are telling the model the current list per-turn anyways via `{{discoverable_tools}}` in `codex-rs/core/templates/search_tool/tool_suggest_description.md`.

sayan-oai (approved):

codex-rs/tools/src/tool_discovery.rs
mzeng-openai: Was going to update the PR description, I'm leaning towards simplify the tool behavior here since "enable" is rarely useful here but adds lots of boilerplate and implementation complexity.

sayan-oai: ok, didnt see the logic being removed so was confused. can review again after that change is in

codex-rs/tools/src/tool_discovery.rs
mzeng-openai: Yeah the main intention is to remove boilerplate and reduce model confusion, we have code to validate the tool ids before triggering the elicitation so if model misbehaves we can catch it.

sayan-oai (approved):

PR #20092, opened 2026-04-28
Return None when auth refresh fails

gpeal:
Right now, if Codex winds up in a state with auth but it can't refresh the token, the user is left with an unhelpful message that says to log out and log back in again.

Ultimately, we should prevent that from happening but if it does, returning None will allow the caller to redirect the user back to the login page


celia-oai (approved):
lgtm. will investigate how we got into this inconsistent state in the first place

PR #20098, opened 2026-04-28
fix: ignore dangerous project-level config keys

owenlin0:
## Description
Ignore these top-level config keys when loading project-scoped config.toml files:
```
    "openai_base_url",
    "chatgpt_base_url",
    "model_provider",
    "model_providers",
    "notify",
    "profile",
    "profiles",
    "experimental_realtime_ws_base_url",
```

## What changed

- Add a project-local config denylist for credential-routing fields such as `openai_base_url`, `chatgpt_base_url`, `model_provider`, `model_providers`, `profile`, `profiles`, and `experimental_realtime_ws_base_url`.
- Strip those fields from project config layers before they participate in effective config merging, while leaving safe project-local settings intact.
- Track ignored project-local keys on config layers and surface a startup warning telling users to move those settings to user-level `config.toml` if they intentionally need them.
- Update profile behavior coverage so project-local `profile` / `profiles` entries are ignored instead of overriding user-level profile selection.

## Verification

- `cargo test -p codex-core project_layer_ignores_unsupported_config_keys`
- `cargo test -p codex-core project_profiles_are_ignored`
- `cargo test -p codex-core config::config_loader_tests`

### DIFF
--- a/codex-rs/config/src/loader/mod.rs
+++ b/codex-rs/config/src/loader/mod.rs
@@ -47,6 +47,21 @@ const SYSTEM_CONFIG_TOML_FILE_UNIX: &str = "/etc/codex/config.toml";
 #[cfg(windows)]
 const DEFAULT_PROGRAM_DATA_DIR_WINDOWS: &str = r"C:\ProgramData";
 
+// Project-local config comes from repository contents, so it should not get to
+// choose where a user's credentials are sent or which local commands are run.
+// These settings are still supported from user, system, managed, and runtime
+// config layers.
+const PROJECT_LOCAL_CONFIG_DENYLIST: &[&str] = &[
+    "openai_base_url",
+    "chatgpt_base_url",
+    "model_provider",
+    "model_providers",
+    "notify",
+    "profile",
+    "profiles",
+    "experimental_realtime_ws_base_url",
+];
+
 async fn first_layer_config_error_from_entries(layers: &[ConfigLayerEntry]) -> Option<ConfigError> {
     typed_first_layer_config_error_from_entries::<ConfigToml>(layers, CONFIG_TOML_FILE).await
 }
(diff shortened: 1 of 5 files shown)


codex-rs/config/src/loader/mod.rs
```diff
@@ -691,16 +704,35 @@ fn project_layer_entry(
...
         dot_codex_folder: dot_codex_folder.clone(),
     };
 
-    if let Some(reason) = disabled_reason {
+    let mut entry = if let Some(reason) = disabled_reason {
         ConfigLayerEntry::new_disabled(source, config, reason)
     } else {
         ConfigLayerEntry::new(source, config)
+    };
+    entry.ignored_project_config_keys = ignored_project_config_keys;
```
bolinfest: I don't love using a type's `new()` method to create something as `mut` and then immediately setting a field because it feels like something that should have been settable in the constructor.

I see that `ConfigLayerEntry::new()` has a lot of call sites, which makes it a pain to change the signature, but also: is it really the case that this is the only place where `ignored_project_config_keys` should be set explicitly for `ConfigLayerEntry`?

Should we be returning a small struct instead like:

```rust
struct SanitizedConfigLayerEntry {
    entry: ConfigLayerEntry,
    ignored_project_config_keys: Vec<String>,
}
```

codex-rs/core/src/config/mod.rs
```diff
@@ -1788,6 +1819,7 @@ impl Config {
         let user_instructions = AgentsMdManager::load_global_instructions(Some(&codex_home))
             .map(|loaded| loaded.contents);
         let mut startup_warnings = Vec::new();
+        push_project_config_ignored_key_warnings(&config_layer_stack, &mut startup_warnings);
```
bolinfest: In practice, this is the only place we do anything with the warnings, right? I think it might be cleaner if the warnings lived on `ConfigLayerStack`, but I'm not sure how hard it is to do.

codex-rs/config/src/loader/mod.rs
```diff
@@ -47,6 +47,19 @@ const SYSTEM_CONFIG_TOML_FILE_UNIX: &str = "/etc/codex/config.toml";
...
 const DEFAULT_PROGRAM_DATA_DIR_WINDOWS: &str = r"C:\ProgramData";
 
+// Project-local config comes from repository contents, so it should not get to
+// choose where a user's OpenAI or ChatGPT credentials are sent. These settings
+// are still supported from user, system, managed, and runtime config layers.
+const PROJECT_LOCAL_CONFIG_DENYLIST: &[&str] = &[
+    "openai_base_url",
+    "chatgpt_base_url",
+    "model_provider",
+    "model_providers",
```
bolinfest, line 58: I think `notify` should be on this list.

codex-rs/config/src/state.rs
```diff
@@ -66,6 +66,7 @@ pub struct ConfigLayerEntry {
     pub raw_toml: Option<String>,
     pub version: String,
     pub disabled_reason: Option<String>,
+    pub warnings: Vec<String>,
```
bolinfest: Let's change this to `Option<Vec<String>>` so:

- `None` means "not checked for warnings" (default)
- `Some(vec![])` means "checked and there were no warnings"
- otherwise: checked and here are the warnings

Incidentally, when the type is `Option<Vec<T>>`, `Some(vec![])` and `vec![]` would both take up three words because the first word is a pointer (which must be non-zero), so I believe the `Option<Vec<String>>` shouldn't take up more space than `Vec<String>` in this case.

codex-rs/config/src/loader/mod.rs
```diff
@@ -696,16 +711,51 @@ fn project_layer_entry(
...
         dot_codex_folder: dot_codex_folder.clone(),
     };
 
-    if let Some(reason) = disabled_reason {
+    let entry = if let Some(reason) = disabled_reason {
         ConfigLayerEntry::new_disabled(source, config, reason)
     } else {
         ConfigLayerEntry::new(source, config)
+    };
+    entry.with_warnings(warnings)
```
bolinfest: thanks, this feels better

codex-rs/config/src/loader/mod.rs
```diff
@@ -696,16 +711,51 @@ fn project_layer_entry(
...
+
+fn project_ignored_config_keys_warning(
+    dot_codex_folder: &AbsolutePathBuf,
+    ignored_keys: &[String],
+) -> String {
+    let config_path = dot_codex_folder.join(CONFIG_TOML_FILE);
+    let ignored_keys = ignored_keys.join(", ");
+    format!(
+        concat!(
+            "Ignored unsupported project-local config keys in {}: {}. ",
```
bolinfest: consider

```suggestion
            "Ignored unsupported project-local config keys in {config_path}: {ignored_keys}. ",

...

     config_path = config_path.display(),
```

bolinfest (approved):

PR #20100, opened 2026-04-28
Increase plugin hook env test timeout

abhinav-oai:
# Why

`plugin_hook_sources_run_with_plugin_env_and_plugin_source` can still fail on Windows after the earlier file-based assertion cleanup because the hook process itself occasionally exceeds the old 5s timeout under CI load. When that happens, the hook run ends as `Failed` before the test can inspect its structured output.

The Windows Bazel failure showed the hook run itself failing after nearly 8 seconds:

```text
---- engine::tests::plugin_hook_sources_run_with_plugin_env_and_plugin_source stdout ----
thread 'engine::tests::plugin_hook_sources_run_with_plugin_env_and_plugin_source' panicked at hooks/src\engine\mod_tests.rs:428:5:
assertion failed: `(left == right)`
Diff < left / right > :
<Failed
>Completed
...
test result: FAILED. 78 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 7.96s
```

# What

- raise the flaky plugin hook env test timeout from 5s to 10s so it matches the other executed hook tests in this module

# Validation

- `cargo test -p codex-hooks`

### DIFF
--- a/codex-rs/hooks/src/engine/mod_tests.rs
+++ b/codex-rs/hooks/src/engine/mod_tests.rs
@@ -373,7 +373,7 @@ print(json.dumps({
                 matcher: Some("Bash".to_string()),
                 hooks: vec![HookHandlerConfig::Command {
                     command: format!("python3 {}", script_path.display()),
-                    timeout_sec: Some(15),
+                    timeout_sec: Some(10),
                     r#async: false,
                     status_message: None,
                 }],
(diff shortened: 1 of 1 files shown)


eternal-openai (approved):
approved though I'm concerned that if this took 8s it could still in theory fail at 10s sometimes

PR #20106, opened 2026-04-29
linux-sandbox: switch helper plumbing to PermissionProfile

bolinfest:
## Why

`PermissionProfile` is the canonical runtime permission model in the Rust workspace, but the Linux sandbox helper still accepted a legacy `SandboxPolicy` plus separate filesystem and network policy flags. That translation layer made the helper interface harder to reason about and left `linux-sandbox`-specific callers and tests coupled to the legacy policy representation.

This change moves the helper onto `PermissionProfile` directly so the Linux sandbox plumbing matches the rest of the permission stack.

## What changed

- changed `codex-linux-sandbox` to accept `--permission-profile` and derive the runtime filesystem and network policies internally
- updated the in-process seccomp and legacy Landlock path in `codex-rs/linux-sandbox` to operate on `PermissionProfile`
- updated Linux sandbox argv construction in `codex-rs/sandboxing`, `codex-rs/core`, and the CLI debug sandbox path to pass the canonical profile instead of serializing compatibility policy projections
- simplified the Linux sandbox tests to build the exact permission profile under test, including the managed-proxy path and direct-runtime-enforcement carveout coverage
- removed helper-local `SandboxPolicy` usage from `bwrap` tests where `FileSystemSandboxPolicy` is already the value being exercised

## Testing

- `cargo test -p codex-linux-sandbox` (on this macOS host, the crate compiled cleanly and its Linux-only tests were cfg-gated)
- `cargo test -p codex-core --no-run`
- `cargo test -p codex-cli --no-run`

### DIFF
--- a/codex-rs/cli/src/debug_sandbox.rs
+++ b/codex-rs/cli/src/debug_sandbox.rs
@@ -16,7 +16,8 @@ use codex_core::spawn::CODEX_SANDBOX_ENV_VAR;
 use codex_core::spawn::CODEX_SANDBOX_NETWORK_DISABLED_ENV_VAR;
 use codex_protocol::config_types::SandboxMode;
 use codex_protocol::permissions::NetworkSandboxPolicy;
-use codex_sandboxing::landlock::create_linux_sandbox_command_args_for_policies;
+use codex_sandboxing::landlock::allow_network_for_proxy;
+use codex_sandboxing::landlock::create_linux_sandbox_command_args_for_permission_profile;
 #[cfg(target_os = "macos")]
 use codex_sandboxing::seatbelt::CreateSeatbeltCommandArgsParams;
 #[cfg(target_os = "macos")]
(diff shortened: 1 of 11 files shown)


viyatb-oai (approved):

evawong-oai (approved):
Non blocking note since this looks pre existing, not introduced by this refactor.

This callsite still passes allow_network_for_proxy as false even when managed_network_requirements_enabled is true. That means codex sandbox linux can start the managed proxy but invoke the Linux helper without proxy only network enforcement, so an enabled network profile can still get full direct network access.

Normal SandboxManager Linux execution threads enforce_managed_network through allow_network_for_proxy(enforce_managed_network). If debug sandbox is supposed to mirror that path, this probably wants the same boolean. Otherwise I would split this into a follow up since it is outside the refactor intent.

bolinfest:
[codex] Addressed in 2bc25571ee1e. The Linux debug-sandbox path now passes allow_network_for_proxy(managed_network_requirements_enabled), matching the regular SandboxManager managed-network behavior.

evawong-oai (approved):
Fresh approval on current head.

Issue #20017, opened 2026-04-28
Codex CLI shell commands hang indefinitely during bwrap sandbox initialization on autofs-backed mounts


mahendra202602:
### What version of Codex CLI is running?

v0.125.0

### What subscription do you have?

Corporate

### Which model were you using?

gpt-5.5, gpt-5.4

### What platform is your computer?

Oracle Linux 8

### What issue are you seeing?

Codex CLI shell commands are hanging indefinitely. Even simple commands like echo test, pwd, date, and ls never return. Instead, they remain as background terminal requests and keep piling up.

This does not appear to be caused by the model loop or by the shell command itself. Investigation suggests Codex is blocking during Bubblewrap sandbox initialization. The sandbox probe appears to run a command similar to:

/usr/bin/bwrap --unshare-user --unshare-net --ro-bind / / /bin/true

That bwrap process then blocks while setting up mounts. The stuck child process was observed in autofs_wait, which suggests it is waiting on an autofs-backed mountpoint encountered while binding the host filesystem.

As a result, Codex becomes unusable because no shell commands complete.

### What steps can reproduce the bug?

Run Codex CLI in a Linux environment that has autofs-backed or lazy-mounted workspace  paths.
Ask Codex to run a simple shell command, for example:
echo test
date
pwd
ls
Observe that the command does not return.
Run /ps in Codex.
Observe that multiple background terminal requests are still pending and continue to accumulate.
Trace the Codex process or its child processes.
Observe that Codex is waiting on a bwrap sandbox probe, and the bwrap child is blocked during mount setup, with process state showing autofs_wait.

### What is the expected behavior?

Codex should complete simple shell commands normally.

If the sandbox cannot be initialized, Codex should fail fast with a clear error message instead of leaving commands running indefinitely in the background.

Ideally, Codex should also avoid blocking forever when bwrap --ro-bind / / encounters an autofs-backed mountpoint.

hipotures:
I found the local root cause and a workaround.

  The problem was a stale/broken NFS systemd automount at `/arch03`. I had an `/etc/fstab` entry like:

  192.168.1.102:/arch03 /arch03 ...

  After reboot I had not mounted it manually, but it still existed as a systemd autofs trigger:

  /arch03 systemd-1 autofs

  The NFS server at `192.168.1.102` was powered off / unavailable at the time, so the automount target
  could not respond. This meant that any process touching `/arch03` could block while the system tried to
  resolve/mount the unavailable NFS export.

[...]

PR #20111, opened 2026-04-29
fix(sandboxing): Bound advisory system bwrap startup probe

### LINKED ISSUE #19828: codex cli startup hangs 20-30s on Linux systems where bwrap bind-mounting / is slow (NFS, autofs)

0.119.0 and later (bug introduced in commit 806e5f7c6 / PR #15893, Apr 6 2026)

Enterprise

gpt-5.4

Linux 5.14.21-150400.24.184-default x86_64 x86_64

Mate Terminal

  Since commit 806e5f7c6 / PR #16215 , codex exec takes ~30 seconds to start on Linux systems with NFS-mounted or autofs-managed filesystems. No error messages are printed — the process simply hangs silently before printing the session header.

```zsh
  $ time codex exec 'hi'
  # ... 30 second silence ...
  OpenAI Codex v0.1.15 (research preview)
  --------
  workdir: /home/user/project
  model: ...
  --------
  user
  hi
  codex
  Hey! How can I help you today?

  real    0m31s
  user    0m27s
  sys     0m2s

```
The root cause is system_bwrap_has_user_namespace_access() in codex-rs/sandboxing/src/bwrap.rs, added in PR #15893. It probes system bwrap on every startup by running:

```zsh
bwrap --unshare-user --unshare-net --ro-bind / / /bin/true
```

via Command::output() with no timeout. On systems with NFS or autofs mounts, binding / causes bwrap to traverse thousands of automount points, taking 20-30 seconds — even though it ultimately succeeds and the sandbox works fine. This blocks the entire startup path synchronously.

  The hang can be confirmed in isolation:

```zsh
  time bwrap --unshare-user --unshare-net --ro-bind / / /bin/true
  # real  0m26s

```

1. Use a Linux system where NFS or autofs mounts are present under / (common in enterprise/HPC environments with network home directories or project mounts)
2. Confirm the hang is present in isolation: time bwrap --unshare-user --unshare-net --ro-bind / / /bin/true — if this takes >1s, you will hit the bug
3. Run time codex exec 'hi'
4. Observe ~30s wall-clock delay before the session header prints

  The delay scales with how long the bwrap command takes on your system. On a standard desktop Linux with no NFS mounts, bwrap completes in <50ms and the bug is not visible.

`codex exec 'hi'` should complete in ~3 seconds as it did prior to commit 806e5f7c6 / PR #16215. The bwrap probe's purpose is only to detect broken user-namespace configurations (e.g. No permissions to create a new namespace). If bwrap is present and ultimately succeeds, the probe should not add perceptible latency to startup.

A 500ms timeout on the probe is sufficient: on a healthy system it completes in <50ms; on a slow-bwrap system it times out, kills the child, and conservatively returns true (assume bwrap works) — which is the correct behavior since bwrap does work, it's just slow.

### Additional information

  Suggested fix in codex-rs/sandboxing/src/bwrap.rs — replace the blocking Command::output() call with a spawn + poll loop that kills the child and returns true after 500ms:

```rust
  use std::time::Duration;

  const BWRAP_CHECK_TIMEOUT: Duration = Duration::from_millis(500);

  fn system_bwrap_has_user_namespace_access(system_bwrap_path: &Path) -> bool {
      let mut child = match Command::new(system_bwrap_path)
          .args(["--unshare-user", "--unshare-net", "--ro-bind", "/", "/", "/bin/true"])
          .stdout(std::process::Stdio::null())
          .stderr(std::process::Stdio::piped())
          .spawn()
      {
          Ok(child) => child,
          Err(_) => return true,
      };

      let deadline = std::time::Instant::now() + BWRAP_CHECK_TIMEOUT;
      loop {
          match child.try_wait() {
              Ok(Some(status)) => {
                  let stderr = child.stderr.take().map_or_else(Vec::new, |mut r| {
                      use std::io::Read;
                      let mut buf = Vec::new();
                      let _ = r.read_to_end(&mut buf);
                      buf
                  });
                  let output = Output { status, stdout: Vec::new(), stderr };
                  return output.status.success() || !is_user_namespace_failure(&output);
              }
              Ok(None) => {
                  if std::time::Instant::now() >= deadline {
                      let _ = child.kill();
                      return true;
                  }
                  std::thread::sleep(Duration::from_millis(50));
              }
              Err(_) => return true,
          }
      }
  }

```
With this fix applied locally, startup on affected systems drops from ~30s to ~3s.

viyatb-oai:
## Why

Linux startup runs an advisory system `bwrap` warning probe on each launch. On hosts with NFS or autofs mounts, its `--ro-bind / /` probe can take tens of seconds before Codex prints anything, matching #19828. Because this probe only decides whether to surface a warning, it should not be allowed to stall startup.

Relevant pre-change path: [`codex-rs/sandboxing/src/bwrap.rs`](https://github.com/openai/codex/blob/de2ccf94735a3d8a2a7077e6a5292026413867cf/codex-rs/sandboxing/src/bwrap.rs#L64-L80)

## What changed

- Bound the advisory system `bwrap` probe to 500 ms.
- Preserve the existing warning behavior when `bwrap` promptly reports a known user-namespace failure.
- Kill and reap the probe child on timeout, then suppress the advisory warning instead of blocking startup.
- Read probe stderr with a bounded nonblocking drain so descendants that inherit the pipe cannot extend startup after the probe child exits.
- Add regression coverage for both a deliberately slow fake `bwrap` process and a fake probe whose descendant keeps stderr open.

## Security

This only bounds the advisory startup probe. It does not change the command execution path or add a fail-open sandbox fallback. The related command-side hang in #20017 remains separate from this PR.

## Verification

- Added `system_bwrap_probe_times_out_without_reporting_a_warning`.
- Added `system_bwrap_probe_does_not_wait_for_descendants_holding_stderr_open`.
- `cargo clippy -p codex-sandboxing --all-targets -- -D warnings`

Fixes #19828
Related: #20017

### DIFF
--- a/codex-rs/sandboxing/src/bwrap.rs
+++ b/codex-rs/sandboxing/src/bwrap.rs
@@ -1,9 +1,14 @@
 use crate::policy_transforms::should_require_platform_sandbox;
 use codex_protocol::models::PermissionProfile;
+use std::io::Read;
 use std::path::Path;
 use std::path::PathBuf;
 use std::process::Command;
 use std::process::Output;
+use std::process::Stdio;
+use std::thread;
+use std::time::Duration;
+use std::time::Instant;
 
 const SYSTEM_BWRAP_PROGRAM: &str = "bwrap";
 const MISSING_BWRAP_WARNING: &str = concat!(
@@ -26,6 +31,8 @@ const USER_NAMESPACE_FAILURES: [&str; 4] = [
     "setting up uid map: Permission denied",
     "No permissions to create a new namespace",
 ];
+const SYSTEM_BWRAP_PROBE_TIMEOUT: Duration = Duration::from_millis(500);
+const SYSTEM_BWRAP_PROBE_POLL_INTERVAL: Duration = Duration::from_millis(50);
 
 pub fn system_bwrap_warning(permission_profile: &PermissionProfile) -> Option<String> {
     if !should_warn_about_system_bwrap(permission_profile) {
@@ -54,15 +61,15 @@ fn system_bwrap_warning_for_path(system_bwrap_path: Option<&Path>) -> Option<Str
         return Some(MISSING_BWRAP_WARNING.to_string());
     };
 
-    if !system_bwrap_has_user_namespace_access(system_bwrap_path) {
+    if !system_bwrap_has_user_namespace_access(system_bwrap_path, SYSTEM_BWRAP_PROBE_TIMEOUT) {
         return Some(USER_NAMESPACE_WARNING.to_string());
     }
 
     None
 }
 
-fn system_bwrap_has_user_namespace_access(system_bwrap_path: &Path) -> bool {
-    let output = match Command::new(system_bwrap_path)
+fn system_bwrap_has_user_namespace_access(system_bwrap_path: &Path, timeout: Duration) -> bool {
+    let mut child = match Command::new(system_bwrap_path)
         .args([
             "--unshare-user",
             "--unshare-net",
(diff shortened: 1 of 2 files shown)


rreichel3-oai (approved):
Looks good, discussed offline to ensure that this wouldn't cause timeouts for /dev and /proc and other special FS mount points.

PR #20117, opened 2026-04-29
feat(cli): add explicit sandbox permission profiles

viyatb-oai:
## Why

`codex sandbox` is useful for exercising sandbox behavior directly, but before this stack the CLI
only picked up permission profiles indirectly from the active config. The existing debug-sandbox path
already compiled `[permissions]` profiles through normal config loading, as covered by the existing
profile tests in [`debug_sandbox.rs`](https://github.com/openai/codex/blob/de2ccf94735a3d8a2a7077e6a5292026413867cf/codex-rs/cli/src/debug_sandbox.rs#L715-L760).

This adds the smallest stable entry point first: an explicit profile selector that reuses the same
config machinery as normal Codex config, so standalone testing becomes possible without changing
current no-selector behavior.

## What changed

- Add additive `--permissions-profile NAME` support to `codex sandbox macos|linux|windows`.
- Resolve built-in and user-defined profile names by feeding `default_permissions` through the
  existing config compilation path instead of inventing a sandbox-only parser.
- Make an explicit selector win over an ambient active profile's legacy `sandbox_mode`.
- Keep the existing no-selector behavior unchanged.

## Stack

1. #20117 `sandbox-ui-profile` --> this PR
2. #20118 `sandbox-ui-config`

Both PRs are additive. Replay JSON is intentionally deferred to a follow-up design pass.

## Tests ran

- `cargo test -p codex-cli sandbox_macos_parses_permissions_profile`
- `cargo test -p codex-core cli_override_takes_precedence_over_profile_sandbox_mode`
- macOS branch-binary smoke on the rebased top of stack: built-in `:workspace` and user-defined
  profiles both executed successfully through `--permissions-profile`.
- Linux devbox branch-binary smoke on the rebased top of stack: built-in `:workspace` and
  user-defined profiles both executed successfully through `--permissions-profile`.


codex-rs/cli/src/debug_sandbox.rs
```diff
@@ -189,7 +196,9 @@ async fn run_command_under_sandbox(
 
     let mut child = match sandbox_type {
         #[cfg(target_os = "macos")]
-        SandboxType::Seatbelt => {
+        SandboxType::Seatbelt {
+            allow_unix_sockets, ..
```
bolinfest: Slight preference for this so that if we add another field to the `Seatbelt` variant, the compiler will force us to deal with it here.

```suggestion
            allow_unix_sockets, log_denials: _,
```

bolinfest (approved):

PR #20118, opened 2026-04-29
feat(cli): add sandbox profile config controls

viyatb-oai:
## Why

The explicit profile path from #20117 is meant for standalone testing, but it still inherited the
shell cwd and all managed requirements implicitly. The pre-existing launcher path even called out
that it did not support a separate cwd yet in
[`debug_sandbox.rs`](https://github.com/openai/codex/blob/509453f688a30929432be866402d1ea46aa12169/codex-rs/cli/src/debug_sandbox.rs#L174-L179).

For a standalone command, the useful default is to let the caller choose the project directory being
tested and to avoid administrator-provided constraints unless the caller explicitly wants to test
those too.

## What changed

- Add explicit-profile-only `-C/--cd DIR`, and use that cwd for both profile resolution and command
  execution.
- Add explicit-profile-only `--include-managed-config`.
- Make explicit profile mode skip managed requirement sources by default, including cloud
  requirements, MDM requirements, `/etc/codex/requirements.toml`, and the legacy managed-config
  requirements projection.
- Preserve all existing invocations outside the explicit-profile path.

## Stack

1. #20117 `sandbox-ui-profile`
2. #20118 `sandbox-ui-config` --> this PR

Both PRs are additive. Replay JSON is intentionally deferred to a follow-up design pass.

## Tests ran

- `cargo test -p codex-cli sandbox_macos_`
- `cargo test -p codex-core load_config_layers_can_ignore_managed_requirements`
- `cargo test -p codex-core load_config_layers_includes_cloud_requirements`
- macOS branch-binary smoke on the rebased top of stack: `-C` changed execution cwd, explicit
  profile mode omitted managed proxy env under `env -i`, and `--include-managed-config` restored it.
- Linux devbox branch-binary smoke on the rebased top of stack: `-C` changed execution cwd for
  built-in and user-defined explicit profiles.

### DIFF
--- a/codex-rs/cli/src/debug_sandbox.rs
+++ b/codex-rs/cli/src/debug_sandbox.rs
@@ -43,7 +43,6 @@ pub async fn run_command_under_seatbelt(
     codex_linux_sandbox_exe: Option<PathBuf>,
 ) -> anyhow::Result<()> {
     let SeatbeltCommand {
-        full_auto,
         permissions_profile,
         cwd,
         include_managed_config,
@@ -52,20 +51,22 @@ pub async fn run_command_under_seatbelt(
         config_overrides,
         command,
     } = command;
+    let managed_requirements_mode = ManagedRequirementsMode::for_profile_invocation(
+        &permissions_profile,
+        include_managed_config,
+    );
     run_command_under_sandbox(
         DebugSandboxConfigOptions {
-            full_auto,
             permissions_profile,
             cwd,
-            include_managed_config,
+            managed_requirements_mode,
         },
         command,
         config_overrides,
         codex_linux_sandbox_exe,
-        SandboxType::Seatbelt {
-            allow_unix_sockets,
-            log_denials,
-        },
+        SandboxType::Seatbelt,
+        log_denials,
+        &allow_unix_sockets,
     )
     .await
 }
@@ -83,24 +84,28 @@ pub async fn run_command_under_landlock(
     codex_linux_sandbox_exe: Option<PathBuf>,
 ) -> anyhow::Result<()> {
     let LandlockCommand {
-        full_auto,
         permissions_profile,
         cwd,
         include_managed_config,
         config_overrides,
         command,
     } = command;
+    let managed_requirements_mode = ManagedRequirementsMode::for_profile_invocation(
+        &permissions_profile,
+        include_managed_config,
+    );
     run_command_under_sandbox(
         DebugSandboxConfigOptions {
-            full_auto,
             permissions_profile,
             cwd,
-            include_managed_config,
+            managed_requirements_mode,
         },
         command,
         config_overrides,
         codex_linux_sandbox_exe,
         SandboxType::Landlock,
+        /*log_denials*/ false,
+        &[],
     )
     .await
 }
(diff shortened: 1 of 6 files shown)


codex-rs/core/src/config/config_loader_tests.rs
```diff
@@ -1084,6 +1084,52 @@ async fn load_config_layers_includes_cloud_requirements() -> anyhow::Result<()>
...
+        overrides,
+        cloud_requirements,
+        &codex_config::NoopThreadConfigLoader,
+    )
+    .await?;
+
+    assert_eq!(
+        layers.requirements_toml(),
+        &ConfigRequirementsToml::default()
+    );
```
bolinfest, line 1129: I think a more compelling assertion would be getting the `Config` and then trying to set a value on a `Constrained<T>` that would normally be disallowed by the managed config, but is allowed in this case because of `overrides.ignore_managed_requirements = true`.

I would also add assertion messages to `assert!` for additional emphasis on the invariant we are trying to maintain.

codex-rs/cli/src/main.rs
```diff
@@ -1950,6 +1950,41 @@ mod tests {
         assert_eq!(command.command, vec!["echo"]);
     }
 
+    #[test]
```
bolinfest: I don't know these sorts of tests give me a ton of additional confidence: I feel like the integration tests that invoke the command are most important.

codex-rs/cli/src/lib.rs
```diff
@@ -60,6 +78,23 @@ pub struct LandlockCommand {
     #[arg(long = "permissions-profile", value_name = "NAME")]
     pub permissions_profile: Option<String>,
 
+    /// Working directory used for profile resolution and command execution.
```
bolinfest, line 75: Is there a world in which we would want separate arguments for the `cwd` used to load the config and the `cwd` used to spawn the process in the sandbox?

codex-rs/cli/src/lib.rs
```diff
@@ -29,6 +30,23 @@ pub struct SeatbeltCommand {
...
```
bolinfest, line 31: Somewhere in the near future, I think we should drop the platform arg after `codex sandbox macos|linux|windows` and then we won't have to repeat these 3x?

codex-rs/cli/src/debug_sandbox.rs
```diff
@@ -124,9 +143,22 @@ enum SandboxType {
     Windows,
 }
 
-async fn run_command_under_sandbox(
+#[derive(Debug)]
+struct DebugSandboxConfigOptions {
     full_auto: bool,
     permissions_profile: Option<String>,
+    cwd: Option<PathBuf>,
+    include_managed_config: bool,
```
bolinfest: Should `ManagedRequirementsMode` be the type of this field?

bolinfest (approved):

codex-rs/cli/src/lib.rs
viyatb-oai: I do not think so for this standalone path. `-C` is meant to behave like "act as if I am in this project": it should drive both profile resolution and the child process cwd. Splitting those would make it easier to accidentally test one project profile while executing somewhere else, and I do not see a concrete use case here that justifies the extra surface.

PR #20119, opened 2026-04-29
feat(cli): add sandbox replay json mode

viyatb-oai:
## Why

After #20117 and #20118, callers can derive a sandbox from config, but there is still no way to
faithfully replay an already-resolved sandbox state without reading ambient config again. The current
launcher builds effective runtime state from loaded config before starting the network proxy and
platform sandbox in
[`debug_sandbox.rs`](https://github.com/openai/codex/blob/731ec516921229984d6aff3ca422a430d60c1d4d/codex-rs/cli/src/debug_sandbox.rs#L205-L290).

This is the first point in the stack where a replay payload is actually needed, so this PR keeps that
payload local to the sandbox CLI instead of promoting a shared MCP-facing contract before there is a
second consumer.

## What changed

- Add mutually exclusive `--permissions-json JSON` and `--permissions-json-file FILE`.
- Introduce a sandbox-CLI-local replay payload containing the effective permissions, managed-network
  behavior, cwd, environment, and platform launcher knobs needed for standalone execution.
- Execute replay mode directly from that payload without reading ambient config, and reject
  config-derived flags that would make replay ambiguous.
- Keep child stdio, exit-code propagation, OS subcommands, and macOS `--log-denials` behavior on the
  existing launcher path.
- Document explicit-profile mode and replay mode in `codex-rs/README.md`.

## Example

```shell
codex sandbox macos --permissions-json-file replay.json -- echo hello
```

## Stack

1. #20117 `sandbox-ui-profile`
2. #20118 `sandbox-ui-config`
3. #20119 `sandbox-ui-json` --> this PR
4. #20120 `sandbox-ui-cleanup`

PRs 1-3 are additive. #20120 is the only intentional behavior break in the stack.

## Tests ran

- `cargo test -p codex-cli sandbox_macos_`
- `cargo test -p codex-core network_proxy_spec`
- macOS branch-binary smoke on the rebased top of stack: file-backed replay JSON honored replay
  cwd, conflicting config-derived flags were rejected, and child exit status `7` propagated.
- Linux devbox branch-binary smoke on the rebased top of stack: file-backed replay JSON honored
  replay cwd, conflicting config-derived flags were rejected, and child exit status `7` propagated.
- macOS branch-binary smoke on this PR head: replay mode still worked and legacy `--full-auto`
  still executed before the cleanup PR removes it.


codex-rs/cli/src/debug_sandbox.rs
```diff
@@ -151,46 +176,185 @@ struct DebugSandboxConfigOptions {
...
+        };
+
+        if let Some(replay) = replay {
+            if !overrides.raw_overrides.is_empty() {
+                anyhow::bail!("sandbox replay JSON cannot be combined with -c/--config overrides");
+            }
+            return Ok(Self::Replay(Box::new(replay)));
+        }
+
+        Ok(Self::Config { options, overrides })
```
bolinfest, line 218:
```suggestion
        if let Some(replay) = replay {
            if !overrides.raw_overrides.is_empty() {
                anyhow::bail!("sandbox replay JSON cannot be combined with -c/--config overrides");
            }
            Ok(Self::Replay(Box::new(replay)))
        } else {
            Ok(Self::Config { options, overrides })
        }
```

codex-rs/cli/src/debug_sandbox/replay.rs
```diff
@@ -0,0 +1,37 @@
...
+use codex_network_proxy::NetworkProxyConfig;
+use codex_protocol::config_types::WindowsSandboxLevel;
+use codex_protocol::models::PermissionProfile;
+use codex_utils_absolute_path::AbsolutePathBuf;
+use serde::Deserialize;
+use serde::Serialize;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub(super) struct SandboxReplayPayload {
```
bolinfest, line 15: Docstring to explain what we mean by "replay."

bolinfest, line 15: Should more of these fields implement `#[default]`?

codex-rs/cli/src/debug_sandbox/replay.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub(super) permission_profile: PermissionProfile,
+    pub(super) network_proxy: Option<SandboxReplayNetworkProxy>,
+    pub(super) managed_network_requirements_enabled: bool,
+    pub(super) sandbox_cwd: AbsolutePathBuf,
+    pub(super) codex_home: AbsolutePathBuf,
+    pub(super) env: HashMap<String, String>,
+    pub(super) codex_linux_sandbox_exe: Option<PathBuf>,
+    pub(super) use_legacy_landlock: bool,
+    pub(super) windows_sandbox_level: WindowsSandboxLevel,
```
bolinfest, line 24: `cfg(windows)` for these?

codex-rs/cli/src/debug_sandbox/replay.rs
```diff
@@ -0,0 +1,37 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub(super) struct SandboxReplayPayload {
+    pub(super) permission_profile: PermissionProfile,
+    pub(super) network_proxy: Option<SandboxReplayNetworkProxy>,
+    pub(super) managed_network_requirements_enabled: bool,
+    pub(super) sandbox_cwd: AbsolutePathBuf,
+    pub(super) codex_home: AbsolutePathBuf,
+    pub(super) env: HashMap<String, String>,
+    pub(super) codex_linux_sandbox_exe: Option<PathBuf>,
```
bolinfest, line 22: `cfg(linux)` for these?

bolinfest (commented):
I'm not sure we should do this until we have a way for the user to reasonably construct this argument.

Perhaps the JSON version of `PermissionProfile` is all we should support? I don't feel like we should be using all of these fields...(for example, why would we need `codex_home` in this case?)

I suppose if we do support taking `PermissionProfile` as JSON, we would have to disallow the `extends` option unless it points to a built-in profile?

PR #20123, opened 2026-04-29
[rollout-tracer] Match analysis messages on encrypted id.

cassirer-openai:
In some setups the summary or raw content can be dropped between requests. This triggers a check in the reducer which expects that the messages should remain identical between requests.

This PR relaxes the checks to only focus on the encrypted ID instead. It also changes the reducer to keep the most rich version of the message observed during the rollout (this ensures that we don't accidentally lose the CoT nor summary when available).

### DIFF
--- a/codex-rs/rollout-trace/src/reducer/conversation.rs
+++ b/codex-rs/rollout-trace/src/reducer/conversation.rs
@@ -6,6 +6,7 @@
 //! later model-facing payload carries their content.
 
 use anyhow::Context;
+use anyhow::Ok;
 use anyhow::Result;
 use anyhow::bail;
 use serde_json::Value;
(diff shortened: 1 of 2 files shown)


codex-rs/rollout-trace/src/reducer/conversation.rs
```diff
@@ -695,6 +661,22 @@ fn readable_reasoning_parts(body: &ConversationBody) -> Vec<&ConversationPart> {
...
+    existing: &ConversationBody,
+    incoming: &ConversationBody,
+) -> bool {
+    let existing_parts = readable_reasoning_parts(existing);
+    let incoming_parts = readable_reasoning_parts(incoming);
+    // Keep the richest non-conflicting readable observation. Treat "richer" as
+    // an ordered superset so a later sighting can add summary/text while a
+    // conflicting sighting cannot overwrite the first readable body.
+    let mut incoming_iter = incoming_parts.iter();
+    existing_parts.len() < incoming_parts.len()
```
jif-oai: This still drops complementary readable reasoning. If one sighting has only raw text and a later sighting has only summary for the same encrypted_content, neither body is a superset, so we never keep both. Can we merge non-conflicting text/summary observations instead of requiring the incoming body to contain all existing readable parts?

jif-oai (approved):

codex-rs/rollout-trace/src/reducer/conversation.rs
cassirer-openai: Yeah I don't think that will ever happen in practice but seems reasonable so updated the code.

PR #20133, opened 2026-04-29
chore(cli) deprecate --full-auto

dylan-hurd-oai:
## Summary
Starts the process of getting rid of `--full-auto`, with some concessions:
1. Fully removes the command from the tui, since it just resolves to the default permissions there, and encourages users to use the one-time trust flow if they're not in a trusted repo.
2. Marks the command as deprecated in `codex exec`, in case users are actively relying on this. We'll remove in an upcoming n+X release.
3. Cleans up some of the `codex sandbox` cli logic, to keep supporting legacy sandbox policies for now.

This isn't the cleanest setup, but I think it is worthwhile to warn users for one release before hard-removing it.

## Testing 
- [x] Updated unit tests


codex-rs/cli/src/debug_sandbox.rs
```diff
@@ -593,9 +579,9 @@ async fn load_debug_sandbox_config(
 async fn load_debug_sandbox_config_with_codex_home(
     cli_overrides: Vec<(String, TomlValue)>,
     codex_linux_sandbox_exe: Option<PathBuf>,
-    full_auto: bool,
     codex_home: Option<PathBuf>,
 ) -> anyhow::Result<Config> {
+    let uses_legacy_sandbox_mode_override = cli_overrides_use_legacy_sandbox_mode(&cli_overrides);
```
etraut-openai, line 589: It took me a while to understand why this new override check was necessary. I think I get it, but this could benefit from a comment.

etraut-openai (approved):
Looks good! There was one part that I needed to stare at for a while to convince myself that it was correct. Could maybe use a comment — up to you.

codex-rs/cli/src/debug_sandbox.rs
dylan-hurd-oai: Very fair! Added

PR #20141, opened 2026-04-29
Add ThreadManager sample crate

pakrym-oai:
Summary:
- Add codex-thread-manager-sample, a one-shot binary that starts a ThreadManager thread, submits a prompt, and prints the final assistant output.
- Pass ThreadStore into ThreadManager::new and expose thread_store_from_config for existing callsites.
- Build the sample Config directly with only --model and prompt inputs.

Verification:
- just fmt
- cargo check -p codex-thread-manager-sample -p codex-app-server -p codex-mcp-server
- git diff --check

Tests: Not run per request.


codex-rs/core/src/thread_manager.rs
```diff
@@ -349,6 +353,14 @@ impl ThreadManager {
             Ok(codex_home) => codex_home,
             Err(err) => panic!("test codex_home should be absolute: {err}"),
         };
+        let cwd = std::env::current_dir().unwrap_or_else(|_| codex_home.clone());
```
bolinfest: Why is process `cwd` preferred to `codex_home`?

codex-rs/core/src/thread_manager.rs
```diff
@@ -272,6 +274,7 @@ impl ThreadManager {
         session_source: SessionSource,
         collaboration_modes_config: CollaborationModesConfig,
         environment_manager: Arc<EnvironmentManager>,
+        thread_store: Arc<dyn ThreadStore>,
```
bolinfest: So we want to support cases where this is created some way other than `thread_store_from_config()`?

Since `new()` already takes `config`, should this be `Option<Arc<dyn ThreadStore>>` and if `None, it calls `thread_store_from_config()`?

bolinfest (approved):

codex-rs/core/src/thread_manager.rs
pakrym-oai: >So we want to support cases where this is created some way other than thread_store_from_config()?

Yes, some consumers will directly create and provider an implementation of ThreadStore. 

I'd like to make it responsibility of the caller to provide a thread store and remove thread_store_from_config eventually so would rather not keep the fallback.

PR #20172, opened 2026-04-29
TUI: Remove core protocol dependency [1/7]

etraut-openai:
## Why

This is part 1 of a 7-PR stack to remove direct `codex_protocol::protocol` usage from `codex-tui` while keeping each layer reviewable and shippable.

This first layer reduces the size of the later `chatwidget` diff by mechanically moving MCP startup bookkeeping out of the central widget file without changing the event shapes or behavior.

## What changed

- Extracted MCP startup status handling into `tui/src/chatwidget/mcp_startup.rs`.
- Kept the existing core event types in place for this purely mechanical move.
- Updated the MCP startup tests to import the moved test-only event types directly.

## Verification

- `cargo test -p codex-tui chatwidget::tests::mcp_startup`


fcoury-oai (approved):
Did a smoke test with one MCP server that is always broken both on the current stable version and in this branch. Also compared the output of `/mcp` after boot and they are identical.

Code looks good, approved.

PR #20173, opened 2026-04-29
TUI: Remove core protocol dependency [2/7]

etraut-openai:
## Why

This is part 2 of a 7-PR stack to remove direct `codex_protocol::protocol` usage from `codex-tui` while keeping each layer reviewable and shippable.

Before the TUI event bus can stop carrying core `Op` values, `AppCommand` needs to be an owned TUI command shape rather than a thin wrapper around `Op`.

## What changed

- Replaced the opaque `AppCommand(Op)` wrapper with explicit owned variants for the commands the TUI submits.
- Preserved `into_core()` so this layer does not yet change the app/thread submission boundary.
- Kept existing core leaf types for now so this remains a mechanical command-shape refactor.

## Verification

- `cargo check -p codex-tui`

### DIFF
--- a/codex-rs/tui/src/app_command.rs
+++ b/codex-rs/tui/src/app_command.rs
@@ -25,8 +25,87 @@ use serde_json::Value;
 
 use crate::permission_compat::legacy_compatible_permission_profile;
 
+#[allow(clippy::large_enum_variant)]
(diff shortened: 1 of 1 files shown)


fcoury-oai (approved):
Code looks good.

I ran smoke tests for submitting a message, changing permissions, and interrupting an active conversation.

PR #20174, opened 2026-04-29
TUI: Remove core protocol dependency [3/7]

etraut-openai:
## Why

This is part 3 of a 7-PR stack to remove direct `codex_protocol::protocol` usage from `codex-tui` while keeping each layer reviewable and shippable.

With `AppCommand` now explicit, the internal app event bus can carry TUI commands directly instead of bouncing through core `Op` values.

## What changed

- Changed `AppEvent::CodexOp` and `AppEvent::SubmitThreadOp` to carry `AppCommand`.
- Updated app-event senders and direct emitters to submit `AppCommand` values.
- Adjusted tests to match `AppCommand` or convert back through `into_core()` where they intentionally assert legacy payload equality.

## Verification

- `cargo test -p codex-tui --no-run`

### DIFF
--- a/codex-rs/tui/src/app/config_persistence.rs
+++ b/codex-rs/tui/src/app/config_persistence.rs
@@ -351,8 +351,8 @@ impl App {
             #[cfg(target_os = "windows")]
             {
                 let windows_sandbox_level = WindowsSandboxLevel::from_config(&self.config);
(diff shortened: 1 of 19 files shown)


fcoury-oai (approved):
Ran the following manual testing: interrupting a running conversion, renaming a session and using the new name to restart it and /compact. They all worked as expected.

Code looks good. Approved.

PR #20239, opened 2026-04-29
[codex-analytics] replay protocol-native item timing

rhan-oai:
## Why

Once protocol-native lifecycle timing is available on the public app-server surface, replay consumers need to preserve it instead of rebuilding partial legacy-only events. Keeping that adapter layer separate makes it clear which changes are about consumption rather than production or wire shape.

## What changed

- Carries native timing through the exec JSONL bridge.
- Replays timestamped app-server items into the TUI legacy event path for command execution, collaboration, web search, image generation, and patch flows.
- Updates TUI and exec tests so replayed items retain the new lifecycle fields.

## Verification

- `cargo test -p codex-exec event_processor_with_json_output`
- `cargo test -p codex-tui --no-run`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/20239).
* #18748
* #18747
* #17090
* #17089
* __->__ #20239
* #20515
* #20514
* #20300


rhan-oai:
Superseded by the collapsed two-PR timing split; the remaining compile-required changes now live in #20515.

PR #20257, opened 2026-04-29
Add thread metadata

evan-oai:
Adds an optional thread-scoped `execution_environment` value to V2 `thread/start`, `thread/resume`, and `thread/fork` requests. Clients that know where a thread is running can classify it as `local`, `ssh`, or `remote_control`; omitted stays unset for older or unspecified callers.

The app-server analytics reducer stores that value with thread metadata and carries it into `codex_thread_initialized`, `codex_turn_event`, and `codex_turn_steer_event`, so downstream analytics can group usage by execution environment.

The new request fields use the existing app-server `experimentalApi` gate, matching other newer V2 request fields.

Validation:
- `cargo test -p codex-analytics track_request_only_enqueues_analytics_relevant_requests`
- `cargo test -p codex-analytics thread_execution_environment_flows_to_thread_turn_and_steer_events`


alexsong-oai (approved):

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -8427,6 +8431,11 @@ async fn handle_pending_thread_resume_request(
         active_permission_profile,
         reasoning_effort,
     };
+    pending.analytics_events_client.track_response(
```
rhan-oai, line 8434: [blocking] centralized response tracking already exists in outgoing_message.rs (in fn send_response_as), this duplicates the send

rhan-oai:
if possible, i would prefer for execution_environment be represented as thread state in codex_app_server_protocol::Thread. this way the protocol maps cleanly to analytics, and we don't have to store request state in the reducer for a field that isn't inherently request-native

rhan-oai (changes requested):

PR #20260, opened 2026-04-29
fix(core): truncate large mcp tool outputs in rollouts

owenlin0:
## Why
Large MCP tool call outputs can make rollout JSONL files enormous. In the session that motivated this change, the biggest JSONL records were:
- `event_msg/mcp_tool_call_end`
- `response_item/function_call_output`

both containing the same unbounded MCP payloads - just 3 MCP tool calls that each were multi-hundred MBs 😱 

This PR truncates both of those JSONL records.

#### For `response_item/function_call_output`
Unified exec already bounds tool output before it is injected into model-facing history, which also keeps the corresponding rollout `response_item/function_call_output` records small. 

MCP should follow the same pattern: truncate the model-facing tool output at the tool-output boundary, while leaving code-mode/raw hook consumers alone.

#### For `event_msg/mcp_tool_call_end`
`McpToolCallEnd` also needs its own bounded event copy because it is the app-server/replay/UI event shape that backs `ThreadItem::McpToolCall`. Unfortunately this is _not_ downstream of the `ToolOutput` trait.

## Model behavior 
Model behavior is actually unchanged as a result of this PR. 

Before this PR, MCP output was:
1. Converted to `FunctionCallOutput`.
2. Recorded into in-memory history.
3. Truncated by `ContextManager::record_items()` before later model turns saw it.

After this branch, MCP output is truncated earlier, in `McpToolOutput::response_payload()`, using the same helper. Then `ContextManager::record_items()` sees an already-truncated output and effectively has little/no additional work to do.

So the model should still see the same kind of truncated function-call output. The practical difference is where truncation happens: earlier, before rollout persistence/app-server emission can see the giant payload.

## Verification

- `cargo test -p codex-core mcp_tool_output`
- `cargo test -p codex-core mcp_tool_call::tests::truncate_mcp_tool_result_for_event`
- `cargo test -p codex-core mcp_post_tool_use_payload_uses_model_tool_name_args_and_result`
- `just fmt`
- `just fix -p codex-core`
- `git diff --check`


codex-rs/core/src/mcp_tool_call.rs
```diff
@@ -647,6 +651,50 @@ fn sanitize_mcp_tool_result_for_model(
...
+            //
+            // This budget applies to the preview text, not the final event JSON.
+            // The preview is itself serialized into a JSON string, so quotes and
+            // backslashes can be escaped again and the stored event may end up
+            // somewhat larger than this byte budget.
+            let truncated = truncate_text(
+                &serialized,
+                TruncationPolicy::Bytes(MCP_TOOL_CALL_EVENT_RESULT_MAX_BYTES),
+            );
+            Ok(CallToolResult {
```
pakrym-oai, line 681: feels a bit strange to build a fake result rather then truncate where we consume the result or store a raw string.

pakrym-oai (approved):
Can we add an integration test that asserts that truncated result arrives in the app-server event.

PR #20267, opened 2026-04-29
Emit analytics for remote plugin installs

xli-oai:
## Summary

- emit `codex_plugin_installed` after a remote plugin install succeeds
- keep local installs unchanged, but let remote installs override the analytics `plugin_id` with the backend remote plugin id (`plugins~Plugin_...`)
- preserve the local/display identity in `plugin_name` and `marketplace_name`, plus capability metadata from the installed bundle
- add regression coverage for local install analytics, remote install analytics, and analytics id override serialization

## Testing

- `just fmt`

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor/plugins.rs
+++ b/codex-rs/app-server/src/codex_message_processor/plugins.rs
@@ -581,7 +581,7 @@ impl CodexMessageProcessor {
 
         let mut plugin_metadata =
             plugin_telemetry_metadata_from_root(&result.plugin_id, &result.installed_path).await;
-        plugin_metadata.analytics_plugin_id = Some(remote_plugin_id);
+        plugin_metadata.remote_plugin_id = Some(remote_plugin_id);
         self.analytics_events_client
             .track_plugin_installed(plugin_metadata);
 
--- a/codex-rs/app-server/src/codex_message_processor/plugins.rs
+++ b/codex-rs/app-server/src/codex_message_processor/plugins.rs
@@ -505,7 +505,7 @@ impl CodexMessageProcessor {
     async fn remote_plugin_install_response(
         &self,
         remote_marketplace_name: String,
-        plugin_name: String,
+        remote_plugin_id: String,
     ) -> Result<PluginInstallResponse, JSONRPCErrorError> {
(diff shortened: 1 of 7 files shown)


codex-rs/app-server/src/codex_message_processor/plugins.rs
```diff
@@ -470,6 +470,10 @@ impl CodexMessageProcessor {
                 Some(self.effective_plugins_changed_callback(config.clone())),
             );
 
+        self.analytics_events_client.track_plugin_installed(
+            plugin_telemetry_metadata_from_root(&result.plugin_id, &result.installed_path).await,
```
xl-openai: What's the format of this plugin_id?

xli-oai: `result.plugin_id` is the local Codex `PluginId`, formatted `<plugin name>@<remote marketplace name>` after the bundle detail has been validated. For this path that means something like `linear@chatgpt-global`, derived from `remote_detail.summary.name` plus the remote marketplace name.

The backend Plugin Service id is still the request/detail id like `plugin_linear`; that is what we send to `/ps/plugins/{plugin_id}/install`. The analytics event intentionally uses the local cache/config id so remote installs report the same `plugin_id`, `plugin_name`, and `marketplace_name` shape as locally materialized plugins.

codex-rs/plugin/src/lib.rs
```diff
@@ -42,13 +42,17 @@ pub struct PluginHookSource {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct PluginTelemetryMetadata {
     pub plugin_id: PluginId,
+    /// Optional event-facing id when the analytics source has a more specific
+    /// upstream identifier than the local plugin cache id.
+    pub analytics_plugin_id: Option<String>,
```
xl-openai: I would call it remote_plugin_id.

xl-openai (approved):

codex-rs/plugin/src/lib.rs
xli-oai: yup 👍 , i will start a renaming across all weird namings thx

xli-oai:
@codex review

PR #20268, opened 2026-04-29
Sync remote installed plugin bundles

xli-oai:
## Summary
- Download missing remote installed plugin bundles during app-server startup and plugin/list refresh.
- Upgrade cached remote installed bundles when the backend installed version changes.
- Remove stale remote installed bundle caches without writing remote plugin state into config.toml.

## Review note
This is a clean PR branch cut from the current diff on top of latest `origin/main`. The diff intentionally has no `codex-rs/core/**` files, so CODEOWNERS should not request the core-directory owner review from stale PR history.

## Validation
Already run on the source branch before creating this clean PR:
- `just fmt`
- `cargo test -p codex-app-server --test all app_server_startup_sync_downloads_remote_installed_plugin_bundles -- --nocapture`
- `cargo test -p codex-app-server --test all plugin_list_sync_upgrades_and_removes_remote_installed_plugin_bundles -- --nocapture`
- `cargo test -p codex-app-server --test all app_server_startup_remote_plugin_sync_runs_once -- --nocapture`
- `just fix -p codex-core-plugins`
- `git diff --check`

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor/plugins.rs
+++ b/codex-rs/app-server/src/codex_message_processor/plugins.rs
@@ -607,6 +607,12 @@ impl CodexMessageProcessor {
             )));
         }
         let actual_remote_marketplace_name = remote_detail.marketplace_name.clone();
+        let _remote_plugin_cache_mutation =
+            codex_core_plugins::remote::mark_remote_plugin_cache_mutation_in_flight(
+                config.codex_home.as_path(),
+                &actual_remote_marketplace_name,
+                &remote_detail.summary.name,
+            );
         let validated_bundle = codex_core_plugins::remote_bundle::validate_remote_plugin_bundle(
             &plugin_name,
             &actual_remote_marketplace_name,
--- a/codex-rs/app-server/src/codex_message_processor/plugins.rs
+++ b/codex-rs/app-server/src/codex_message_processor/plugins.rs
@@ -3,7 +3,65 @@ use crate::error_code::internal_error;
 use crate::error_code::invalid_request;
 use codex_app_server_protocol::PluginInstallPolicy;
 
+fn maybe_start_remote_installed_plugin_bundle_sync_for_config(
+    thread_manager: Arc<ThreadManager>,
+    config: Config,
+    auth: Option<CodexAuth>,
+    on_effective_plugins_changed: Option<Arc<dyn Fn() + Send + Sync + 'static>>,
+) {
+    if !config.features.enabled(Feature::Plugins) || !config.features.enabled(Feature::RemotePlugin)
+    {
+        return;
+    }
+
+    let remote_plugin_service_config = RemotePluginServiceConfig {
+        chatgpt_base_url: config.chatgpt_base_url.clone(),
+    };
+    let plugins_manager = thread_manager.plugins_manager();
+    let config_for_refresh = config.clone();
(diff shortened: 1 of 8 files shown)


codex-rs/core-plugins/src/remote.rs
```diff
@@ -651,6 +869,96 @@ fn remove_remote_plugin_cache(
...
+                )
+            })?;
+            if installed_plugin_names.contains(&plugin_name) {
+                continue;
+            }
+
+            let cache_path = entry.path();
+            if cache_path.is_dir() {
+                fs::remove_dir_all(&cache_path).map_err(|err| {
+                    format!(
```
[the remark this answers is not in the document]
xli-oai: [codex] Fixed in 31459836a6. Direct remote install now holds a ref-counted cache mutation guard from bundle download through backend install, and stale cleanup skips guarded plugin cache entries instead of deleting solely from one installed-plugins snapshot. Added a focused core-plugins regression test for this path.

codex-rs/core-plugins/src/remote.rs
```diff
@@ -408,6 +455,177 @@ pub async fn fetch_remote_installed_plugins(
...
+    let Some(auth) = auth else {
+        return;
+    };
+    let key = RemoteInstalledPluginBundleSyncKey {
+        codex_home: codex_home.clone(),
+        chatgpt_base_url: config.chatgpt_base_url.clone(),
+        account_id: auth.get_account_id(),
+        chatgpt_user_id: auth.get_chatgpt_user_id(),
+    };
+    if !mark_remote_installed_plugin_bundle_sync_in_flight(key.clone()) {
```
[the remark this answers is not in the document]
xli-oai: [codex] Fixed in 31459836a6. The in-flight sync key is now the local plugin cache root, not auth/account identity, so two auth contexts cannot run concurrent installed-bundle syncs against the same `plugins/cache` tree. Added a focused regression test for cache-root deduping.

codex-rs/app-server/src/codex_message_processor/plugins.rs
```diff
@@ -434,6 +500,12 @@ impl CodexMessageProcessor {
                 "remote plugin {plugin_name} is not available for install"
             )));
         }
+        let _remote_plugin_cache_mutation =
```
xl-openai, line 546: Why do you need to call this outside maybe_start_remote_installed_plugin_bundle_sync?

xli-oai: [codex] Answered in 4868ed79f1 with a call-site comment. `maybe_start_remote_installed_plugin_bundle_sync` only deduplicates/protects the background installed-bundle sync. Direct remote install is a separate writer to the same `plugins/cache` tree: it materializes the bundle before the backend installed-plugin snapshot can include it, so a concurrent sync could otherwise see that cache entry as stale and remove it. The guard has to start before the direct local bundle write and stay alive through the backend install call.

riskpumpfun-debug:
Who is still holding and haven’t received there supplies?

codex-rs/app-server/src/codex_message_processor/plugins.rs
xl-openai: I see this is for direct install.

xl-openai (approved):

xli-oai:
@codex review

PR #20270, opened 2026-04-29
[codex] Fix elevated Windows sandbox named-pipe access

iceweasel-oai:
## Summary
- add elevated-only token constructors that include the current token user SID in the restricted SID list
- switch the elevated Windows command runner to use those constructors
- leave the unelevated restricted-token path unchanged

## Why
Windows named pipes created by tools like Ninja use the platform's default named-pipe ACL when no explicit security descriptor is provided. In the elevated sandbox, the pipe owner has access, but the write-restricted token can still fail its restricted-SID access check because the sandbox user SID was not in the restricting SID set. That causes child processes to exit successfully while Ninja never receives the expected pipe completion/close behavior and hangs.

Including the elevated sandbox user's SID in the restricting SID list lets the restricted check succeed for these owner-scoped pipe objects without broadening the unelevated sandbox to the real signed-in user.

## Impact
- fixes the minimal Ninja hang repro in the elevated Windows sandbox
- preserves the existing unelevated sandbox behavior and write protections
- keeps the change scoped to the elevated runner rather than changing shared token semantics
- this does not affect file-writes for the sandbox because the sandbox users themselves do not receive any additional permissions over what the capability SIDs already have. In fact we don't even explicitly grant the sandbox user ACLs anywhere.

## Validation
- `cargo build -p codex-windows-sandbox --quiet`
- verified the stock `ninja.exe` minimal repro exits normally on host and in the elevated sandbox
- verified the same repro still hangs in the unelevated sandbox, which is the intended scope of this change

### DIFF
--- a/codex-rs/windows-sandbox-rs/src/elevated/command_runner_win.rs
+++ b/codex-rs/windows-sandbox-rs/src/elevated/command_runner_win.rs
@@ -28,8 +28,8 @@ use codex_windows_sandbox::SpawnRequest;
 use codex_windows_sandbox::StderrMode;
 use codex_windows_sandbox::StdinMode;
 use codex_windows_sandbox::allow_null_device;
-use codex_windows_sandbox::create_readonly_token_with_caps_from;
-use codex_windows_sandbox::create_workspace_write_token_with_caps_from;
+use codex_windows_sandbox::create_readonly_token_with_caps_and_user_from;
+use codex_windows_sandbox::create_workspace_write_token_with_caps_and_user_from;
 use codex_windows_sandbox::decode_bytes;
 use codex_windows_sandbox::encode_bytes;
 use codex_windows_sandbox::get_current_token_for_restriction;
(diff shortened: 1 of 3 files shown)


iceweasel-oai:
@codex review

codex-rs/windows-sandbox-rs/src/elevated/command_runner_win.rs
```diff
@@ -242,10 +242,10 @@ fn spawn_ipc_process(req: &SpawnRequest) -> Result<IpcSpawnedProcess> {
     let h_token = OwnedWinHandle::new(unsafe {
         match &policy {
             SandboxPolicy::ReadOnly { .. } => {
-                create_readonly_token_with_caps_from(base.raw(), &cap_psid_ptrs)
+                create_readonly_token_with_caps_and_user_from(base.raw(), &cap_psid_ptrs)
```
viyatb-oai, line 245: Non-blocking, and I do not think this has security impact, but one small side effect worth calling out: this is a little broader than the named-pipe case. For example, an object that already grants `CodexSandboxOffline` access, such as a file owned by that sandbox account under its own profile, will now also satisfy the restricted-SID pass. If we ever wanted to keep `read-only` narrower, the follow-up would be to leave `ReadOnly` on the old helper and only opt it into this path once we have a concrete `read-only` repro that needs the broader behavior too. I think the current tradeoff is still reasonable because this is the dedicated sandbox account, not the real signed-in user.

viyatb-oai (approved):
lgtm with a non blocking comment

codex-rs/windows-sandbox-rs/src/elevated/command_runner_win.rs
iceweasel-oai: @viyatb-oai yep, it is broader for sure. I addressed this in the PR with

```this does not affect file-writes for the sandbox because the sandbox users themselves do not receive any additional permissions over what the capability SIDs already have. In fact we don't even explicitly grant the sandbox user ACLs anywhere.```

PR #20275, opened 2026-04-29
fix: show correct Bedrock runtime endpoint in /status

celia-oai:
## Why

`/status` was showing the configured `ModelProviderInfo.base_url` for Amazon Bedrock, which can be stale or misleading because the actual Bedrock Mantle endpoint is derived at runtime from the resolved AWS region. This made sessions report the wrong provider endpoint even though requests used the correct runtime URL.

## What changed

- Added `ModelProvider::runtime_base_url()` so provider implementations can expose the request-time base URL through the shared runtime provider abstraction.
- Moved Bedrock region-to-Mantle URL resolution into `amazon_bedrock::mantle::runtime_base_url()`, keeping region resolution private to the Mantle module.
- Overrode `runtime_base_url()` for Amazon Bedrock so it returns the resolved Mantle endpoint instead of the configured default.
- Resolved and cached the runtime provider base URL during TUI startup, then used that cached value when rendering `/status`.
- Added status coverage that verifies Bedrock displays the runtime URL and ignores the configured Bedrock `base_url` when they differ.

## Verification
model provider is resolved correctly in local build:
Screenshot 2026-04-29 at 5 01 36 PM

### DIFF
--- a/codex-rs/tui/src/app.rs
+++ b/codex-rs/tui/src/app.rs
@@ -125,6 +125,8 @@ use codex_config::types::ApprovalsReviewer;
 use codex_config::types::ModelAvailabilityNuxConfig;
 use codex_exec_server::EnvironmentManager;
 use codex_features::Feature;
+use codex_model_provider::create_model_provider;
+use codex_model_provider_info::ModelProviderInfo;
 use codex_models_manager::model_presets::HIDE_GPT_5_1_CODEX_MAX_MIGRATION_PROMPT_CONFIG;
 use codex_models_manager::model_presets::HIDE_GPT5_1_MIGRATION_PROMPT_CONFIG;
 use codex_otel::SessionTelemetry;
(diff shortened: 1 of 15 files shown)


codex-rs/tui/src/app.rs
```diff
@@ -598,6 +600,17 @@ fn active_turn_not_steerable_turn_error(error: &TypedRequestError) -> Option<App
     .then_some(turn_error)
 }
 
+async fn resolve_runtime_model_provider_base_url(provider: &ModelProviderInfo) -> Option<String> {
+    let provider = create_model_provider(provider.clone(), /*auth_manager*/ None);
```
pakrym-oai, line 516: why don't we do this in format_model_provider directly?

celia-oai: because this runtime url resolution needs to make aws sdk call, so it's pretty heavy. Ideally we keep create_model_provider lightweight, and only make sdk call when needed. it also forces `create_model_provider` to be async, which feels like a bit of an overkill

pakrym-oai (approved):

PR #20277, opened 2026-04-29
[codex] tighten exec safety parsing and sandbox path review

olliem-oai:
## Summary

- Decode backslash escapes in unquoted Bash/Zsh words before command-safety matching, so shell-escaped tokens such as `-e\xec` are evaluated as the argv Bash will actually execute.
- Require approval when a command requests a sandbox override and the executable is an explicit path inside a sandbox-writable location, preventing basename-safe paths like `/tmp/cat` from bypassing review.
- Suppress execpolicy amendment suggestions for those sandbox-writable executable paths.

## Root Cause

Two approval bypasses came from comparing the wrong executable surface:

- Bash parser safety checks used raw tree-sitter word text, while Bash removes unquoted backslash escapes before execution.
- Safe-command matching trusted executable basenames even when the executable path itself was controlled by the sandboxed workspace/tmp area and the command requested escalated execution.

## Validation

- `cargo test -p codex-shell-command`
- `cargo test -p codex-core exec_policy::tests::sandbox_writable_safe_command_path_requires_approval_for_escalation`
- `cargo test -p codex-core exec_policy::tests::`

### DIFF
--- a/codex-rs/core/src/exec_policy.rs
+++ b/codex-rs/core/src/exec_policy.rs
@@ -250,7 +250,16 @@ impl ExecPolicyManager {
         // Keep heredoc prefix parsing for rule evaluation so existing
         // allow/prompt/forbidden rules still apply, but avoid auto-derived
         // amendments when only the heredoc fallback parser matched.
-        let auto_amendment_allowed = !used_complex_parsing;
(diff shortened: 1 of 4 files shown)


olliem-oai:

PR #20281, opened 2026-04-29
Use selected turn environments for runtime context

starr-openai:
## Summary
- make selected turn environments the source of truth for session runtime cwd and MCP runtime environment selection
- keep local/no-selection fallback behavior intact
- add coverage for duplicate selected environments, cwd resolution, and MCP runtime environment selection

## Validation
- git diff --check
- rustfmt was run on touched Rust files during the implementation workflow

CI should provide the full Bazel/test signal.


codex-rs/core/src/session/turn_context.rs
```diff
@@ -106,6 +105,10 @@ impl TurnContext {
         self.permission_profile.network_sandbox_policy()
     }
 
+    pub(crate) fn primary_environment(&self) -> Option<&TurnEnvironment> {
+        self.environments.first()
```
pakrym-oai, line 109: does any of the callers care about the environment id?

pakrym-oai: we can reduce boilerplate by returning Environment directly

codex-rs/core/src/session/mod.rs
```diff
@@ -471,12 +472,14 @@ impl Codex {
         validate_environment_selections(environment_manager.as_ref(), &environments)?;
         let environment =
             selected_primary_environment(environment_manager.as_ref(), &environments)?;
+        let mut load_config = config.clone();
+        load_config.cwd = primary_selected_cwd_or_fallback(&environments, &config.cwd);
```
pakrym-oai: instead of adding another method can we return a tuple from selected_primary_environment?

codex-rs/core/src/environment_selection.rs
```diff
@@ -25,7 +26,14 @@ pub(crate) fn validate_environment_selections(
     environment_manager: &EnvironmentManager,
     environments: &[TurnEnvironmentSelection],
 ) -> CodexResult<()> {
+    let mut seen_environment_ids = HashSet::with_capacity(environments.len());
     for selected_environment in environments {
+        if !seen_environment_ids.insert(selected_environment.environment_id.as_str()) {
```
pakrym-oai, line 62: should we just dedup?

codex-rs/core/src/thread_manager.rs
```diff
@@ -1086,13 +1086,20 @@ impl ThreadManagerState {
         user_shell_override: Option<crate::shell::Shell>,
     ) -> CodexResult<NewThread> {
         let is_resumed_thread = matches!(&initial_history, InitialHistory::Resumed(_));
+        validate_environment_selections(self.environment_manager.as_ref(), &environments)?;
         let environment =
             selected_primary_environment(self.environment_manager.as_ref(), &environments)?;
+        let effective_cwd = environments
```
pakrym-oai: isn't this missaligned with primary?

codex-rs/core/src/thread_manager.rs
```diff
@@ -1086,13 +1086,20 @@ impl ThreadManagerState {
         user_shell_override: Option<crate::shell::Shell>,
     ) -> CodexResult<NewThread> {
         let is_resumed_thread = matches!(&initial_history, InitialHistory::Resumed(_));
+        validate_environment_selections(self.environment_manager.as_ref(), &environments)?;
         let environment =
```
pakrym-oai: Feels like we are resolving environments in too many places.

We resolve primary here, then throw it away, then re-resolve in spawn_internal.

codex-rs/core/src/session/turn_context.rs
```diff
@@ -641,6 +647,10 @@ impl Session {
         &self,
         environments: &[TurnEnvironmentSelection],
     ) -> CodexResult<Vec<TurnEnvironment>> {
+        crate::environment_selection::validate_environment_selections(
```
pakrym-oai: do we need validate_environment_selections? 

we fail with "unknown turn environment" below.

codex-rs/core/src/session/tests.rs
```diff
@@ -4397,32 +4495,206 @@ async fn empty_turn_environments_clear_primary_environment() {
...
+    assert!(err.to_string().contains("duplicate"));
+    assert_eq!(current_configuration.cwd, original_configuration.cwd);
+    assert_eq!(
+        current_configuration.environments,
+        original_configuration.environments
+    );
+}
+
+#[tokio::test]
+async fn mcp_runtime_environment_uses_primary_selected_environment() {
```
pakrym-oai: is this correct? I thought we have an environment specified on mcp servers.

pakrym-oai: I'm not sure these tests add a ton of value tbh considering integration tests

codex-rs/core/src/session/session.rs
```diff
@@ -183,26 +183,29 @@ impl SessionConfiguration {
             next_configuration.windows_sandbox_level = windows_sandbox_level;
         }
 
+        let effective_environments = updates
+            .environments
+            .as_deref()
+            .unwrap_or(&self.environments);
         let absolute_cwd = updates
```
pakrym-oai, line 195: we should consider rejecting updates/thread starts that have both environments and cwd.

codex-rs/core/src/session/mcp.rs
```diff
@@ -234,13 +278,7 @@ impl Session {
...
             self.get_tx_event(),
             turn_context.permission_profile(),
-            McpRuntimeEnvironment::new(
-                turn_context
-                    .environment
-                    .clone()
-                    .unwrap_or_else(|| self.services.environment_manager.local_environment()),
-                turn_context.cwd.to_path_buf(),
-            ),
+            self.mcp_runtime_environment_for_turn_context(turn_context),
```
pakrym-oai: don't we have access to configuration here?

pakrym-oai: to use mcp_runtime_environment_for_configuration

codex-rs/core/src/session/mcp.rs
```diff
@@ -1,6 +1,50 @@
...
+            self.services.environment_manager.local_environment(),
+            turn_context.cwd.to_path_buf(),
+        )
+    }
+
+    pub(crate) fn mcp_runtime_environment_for_configuration(
+        &self,
+        session_configuration: &SessionConfiguration,
+    ) -> CodexResult<McpRuntimeEnvironment> {
+        if let Some(turn_environment) = session_configuration.environments.first() {
```
pakrym-oai: it's a bit confusing to do the same thing in differnt ways in so many places.

codex-rs/core/src/session/mcp.rs
```diff
@@ -1,6 +1,50 @@
 use super::*;
 
 impl Session {
+    pub(crate) fn mcp_runtime_environment_for_turn_context(
+        &self,
+        turn_context: &TurnContext,
+    ) -> McpRuntimeEnvironment {
+        if let Some(turn_environment) = turn_context.primary_environment() {
```
pakrym-oai: A little surprised we have logic like this, expected each mcp server to define an env.

pakrym-oai: but this is outside of scope of the pr.

codex-rs/codex-mcp/src/runtime.rs
```diff
@@ -50,11 +50,11 @@ impl McpRuntimeEnvironment {
         }
     }
 
-    pub(crate) fn environment(&self) -> Arc<Environment> {
+    pub fn environment(&self) -> Arc<Environment> {
         Arc::clone(&self.environment)
     }
 
-    pub(crate) fn fallback_cwd(&self) -> PathBuf {
+    pub fn fallback_cwd(&self) -> PathBuf {
```
pakrym-oai: did we start using these publicly?

codex-rs/core/src/session/turn_context.rs
starr-openai: ^ some callers needed cwd from this

codex-rs/core/src/session/mod.rs
starr-openai: hmm good idea. i did this helper mostly since there was a lot of copy/paste handling for "use turn cwd else fallback"

codex-rs/core/src/environment_selection.rs
starr-openai: we're relying on ordering for the moment to choose primary, so we cant dedupe with like a HashSet on turncontext itself yet. but yeah by this point we should have already validated out dupes... let me try to consolidate the validation/resolve logic again

codex-rs/core/src/session/turn_context.rs
starr-openai: let me consolidate validation then we can skip that here

starr-openai:
Follow-up rationale for the latest cwd/environment cleanup:

- `SessionConfiguration.cwd` and `SessionConfiguration.environments[*].cwd` now remain separate pieces of state.
- A turn/session-level `cwd` update no longer rewrites the stored sticky environment cwd.
- Environment cwd changes should come from explicit environment selections (`TurnEnvironmentSelection.cwd` / app-server `TurnEnvironmentParams.cwd`), while `config.cwd` remains the fallback when no primary environment is selected.
- MCP startup/refresh now keeps the prior default-environment-or-local fallback inline, using the primary turn/stored environment only when one is explicitly selected.

This keeps the PR pointed in the direction of relying on explicit env cwd rather than deriving or mutating env cwd from unrelated turn-level cwd updates.

codex-rs/core/src/session/session.rs
```diff
@@ -207,11 +208,6 @@ impl SessionConfiguration {
 
         let cwd_changed = absolute_cwd.as_path() != self.cwd.as_path();
         next_configuration.cwd = absolute_cwd.clone();
-        if cwd_changed
-            && let Some(primary_environment) = next_configuration.environments.first_mut()
-        {
-            primary_environment.cwd = absolute_cwd;
-        }
```
starr-openai, line 214: re previous comments about cwd handling:

im removing places where we took turn level cwd and overrode primary env cwd.  instead, we'll rely on the explicit TurnEnvironmentParams.cwd for this, and only use turn level cwd as a fallback if none was provided.

codex-rs/app-server/src/codex_message_processor.rs
```diff
@@ -2965,6 +2951,27 @@ impl CodexMessageProcessor {
...
+                .into_iter()
+                .map(|environment| TurnEnvironmentSelection {
+                    environment_id: environment.environment_id,
+                    cwd: environment.cwd,
+                })
+                .collect::<Vec<_>>()
+        });
+        if let Some(environment_selections) = environment_selections.as_ref() {
+            self.thread_manager
+                .validate_environment_selections(environment_selections)
```
pakrym-oai, line 3017: nit, move this method from thread_manager into message processor?

codex-rs/core/src/thread_manager.rs
```diff
@@ -419,7 +418,8 @@ impl ThreadManager {
         &self,
         environments: &[TurnEnvironmentSelection],
     ) -> CodexResult<()> {
-        validate_environment_selections(self.state.environment_manager.as_ref(), environments)
+        resolve_environment_selections(self.state.environment_manager.as_ref(), environments)
+            .map(|_| ())
```
pakrym-oai, line 436: nit: what does this map do?

codex-rs/core/src/session/turn_context.rs
```diff
@@ -672,8 +656,6 @@ impl Session {
         turn_environments: Vec<TurnEnvironment>,
     ) -> Arc<TurnContext> {
         let primary_turn_environment = turn_environments.first();
```
pakrym-oai, line 664: nit: not for this round, but we are still having multiple places deciding what primary is

codex-rs/core/src/session/mcp.rs
```diff
@@ -221,6 +221,19 @@ impl Session {
         let mcp_servers = with_codex_apps_mcp(mcp_servers, auth.as_ref(), &mcp_config);
         let auth_statuses =
             compute_auth_statuses(mcp_servers.iter(), store_mode, auth.as_ref()).await;
+        let mcp_runtime_environment = match turn_context.primary_environment() {
```
pakrym-oai, line 224: is this a copy of logic in session.rs?

pakrym-oai (approved):

codex-rs/core/src/thread_manager.rs
starr-openai: we're just calling resolve_environment_selections for validation here but dropping the result

PR #20282, opened 2026-04-29
tui: return from side chat on Ctrl-D

### LINKED ISSUE #20264: ctrl+d in /side should exit the side chat, not exit codex
### What variant of Codex are you using?

cli

### What feature would you like to see?

I was in a side chat and hit ctrl+d expecting it to exit out of the layer into the root layer of chat, but codex exited. Side chats are similar to a subshell in a shell, so ctrl+d should be an unwind / exit layer rather than a quit command here (or at least prompt). The Impact of not doing this is loss of CoT on the main chat because it's not reloaded when running resume.

### Additional information

_No response_

etraut-openai:
## Why

Fixes #20264.

Side conversations are an ephemeral layer on top of the main chat. Pressing `Ctrl+D` from an empty side-chat composer should unwind back to the parent thread, matching the existing side-return behavior, instead of falling through to the global quit shortcut and exiting Codex.

## What changed

The side-return shortcut matcher now treats `Ctrl+D` the same way it already treats `Esc` and `Ctrl+C`. Because app-level side-return handling runs before the chat widget's global quit handling, this returns from `/side` while preserving normal `Ctrl+D` quit behavior outside side conversations.

The existing shortcut coverage was updated to include lowercase and uppercase `Ctrl+D` key events.

## Verification

- `cargo test -p codex-tui side_return_shortcuts_match_esc_ctrl_c_and_ctrl_d`
- `cargo test -p codex-tui` starts successfully and the new shortcut test passes, but the broader suite later aborts in the unrelated existing test `app::tests::attach_live_thread_for_selection_rejects_unmaterialized_fallback_threads` with a stack overflow.

### DIFF
--- a/codex-rs/tui/src/app/platform_actions.rs
+++ b/codex-rs/tui/src/app/platform_actions.rs
@@ -65,7 +65,11 @@ pub(super) fn side_return_shortcut_matches(key_event: KeyEvent) -> bool {
             modifiers,
             kind: KeyEventKind::Press,
             ..
-        } if modifiers.contains(KeyModifiers::CONTROL) && c.eq_ignore_ascii_case(&'c') => true,
+        } if modifiers.contains(KeyModifiers::CONTROL)
+            && (c.eq_ignore_ascii_case(&'c') || c.eq_ignore_ascii_case(&'d')) =>
+        {
+            true
+        }
         _ => false,
     }
 }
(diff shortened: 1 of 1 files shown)


canvrno-oai (approved):

fcoury-oai (approved):
Code looks good, ran a smoke test: started a main thread, ran `/side` and used Ctrl+D.

With latest alpha codex I got a complete exit and with this code I successfully exited back to the main thread.

Approved!

PR #20291, opened 2026-04-30
app-server: remove dead api version handling from bespoke events

pakrym-oai:
Remove ApiVersion::V1


etraut-openai (approved):
Codex reordered event handlers in bespoke_event_handling, so I wasn't able to verify with certainty that the new code is equivalent, but I spot checked it, and it looks good.

PR #20300, opened 2026-04-30
[codex-analytics] centralize thread analytics state

rhan-oai:
## Why

Several analytics event families need the same per-thread attribution state: the app-server client/runtime associated with a thread and, for lifecycle-oriented events, the thread metadata captured during initialization. Keeping connection ids and lifecycle metadata in separate maps made each consumer rebuild the same thread context and made subagent attribution harder to resolve consistently.

## What changed

- Replaces the separate thread connection and metadata maps with one reducer-owned `threads` map.
- Routes guardian, compaction, turn-steer, and turn analytics through shared thread-state lookups while preserving turn-origin attribution for turn events and request-origin attribution for steer events.
- Lets newly observed spawned subagent threads inherit their parent thread connection so later thread-scoped analytics can resolve through the same state model.
- Adds regression coverage for standalone `SubAgentThreadStarted` publication plus the `SubAgentSource::ThreadSpawn` parent fallback through a thread-scoped consumer that depends on inherited connection state.

## Verification

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/20300).
* #18748
* #18747
* #17090
* #17089
* #20515
* #20514
* __->__ #20300

### DIFF
--- a/codex-rs/analytics/src/analytics_client_tests.rs
+++ b/codex-rs/analytics/src/analytics_client_tests.rs
@@ -303,10 +303,10 @@ fn sample_turn_completed_notification(
     })
 }
 
-fn sample_turn_resolved_config(turn_id: &str) -> TurnResolvedConfigFact {
+fn sample_turn_resolved_config(thread_id: &str, turn_id: &str) -> TurnResolvedConfigFact {
     TurnResolvedConfigFact {
         turn_id: turn_id.to_string(),
-        thread_id: "thread-2".to_string(),
+        thread_id: thread_id.to_string(),
         num_input_images: 1,
         submission_type: None,
         ephemeral: false,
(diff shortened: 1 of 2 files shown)


rhan-oai:
@codex

owenlin0 (commented):
i find that there are a lot of concepts to understand, can we simplify?

seems like the branch added a fairly abstract mini-framework:
- AnalyticsDropSite: where the event was dropped
- MissingAnalyticsContext: why it was dropped
- AnalyticsConnectionSource: how to find the connection
- ThreadMetadataRequirement: whether metadata is needed
- ResolvedAnalyticsContext: the resolved output

are all those necessary?

wonder if we can have a `resolve_analytics_context` that looks something like:
```
let Some((connection_state, thread_metadata)) =
    self.thread_context_or_warn("compaction", &input.thread_id, Some(&input.turn_id))
else {
    return;
};
```

or something like that

owenlin0 (approved):

PR #20305, opened 2026-04-30
fix(exec-policy) use is_known_safe_command less

dylan-hurd-oai:
## Summary
Restricts behavior of `is_known_safe_command` only to modes where it is explicitly part of the documented behavior:
- when `environment_lacks_sandbox_protections`
- in `AskForApproval::UnlessTrusted`

Notably, as a result of this, escalations for commands that pass `is_known_safe_commands` are no longer auto-approved in AskForApproval::OnRequest or AskForApproval::Granular.

## Testing
- [x] Updated unit tests
- [x] Updated approvals scenario tests.

### DIFF
--- a/codex-rs/core/src/exec_policy.rs
+++ b/codex-rs/core/src/exec_policy.rs
@@ -649,9 +649,6 @@ pub(crate) fn render_decision_for_unmatched_command(
             codex_shell_command::is_safe_command::is_safe_powershell_words(command)
         }
     };
-    if is_known_safe && !used_complex_parsing {
-        return Decision::Allow;
-    }
 
     // On Windows, ReadOnly sandbox is not a real sandbox, so special-case it
     // here.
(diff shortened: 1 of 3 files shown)


evawong-oai (approved):
Security review done. I approve this PR.

Non blocking: while reviewing this file I noticed a pre existing issue in exec policy loading. A malformed rules file currently falls back to an empty policy, which can drop explicit rules for the session. This does not look introduced by this PR, and this PR narrows the approval bypass surface, so I would not block this change on it. It should be tracked as a separate follow up.

Research done:

1. Reviewed PR 20305 from base f3e0a64efc0adba396672cf1167f04da6127a56d to head 116859a1fc395fe826405a4d2463692200a74ef7.
2. Triage covered the three changed files. The runtime review focused on the exec policy implementation. The test files were used as behavior evidence.
3. Confirmed the PR makes known safe command auto allow narrower. Under OnRequest, known safe sandbox escalation now prompts. Under Granular with sandbox approval disabled, it is forbidden.
4. Found one Medium reviewed file issue. It appears pre existing and not introduced by this PR.
5. Dedupe was skipped because there was one finding. Rerank completed. High severity corruption and RCE validation was skipped because there were no high severity corruption or RCE findings.
6. GitHub checks are passing on the current head.

owenlin0 (approved):

PR #20319, opened 2026-04-30
Add allow_managed_hooks_only hook requirement

eternal-openai:
## Why

Enterprise-managed hook policy needs a narrow way to require Codex to ignore user-controlled lifecycle hooks without adopting the broader trust-precedence model from earlier hook work. This keeps the policy anchored in `requirements.toml`, so admins can opt into managed hooks only while normal `config.toml` files cannot enable the restriction themselves.

## What changed

- Added `allow_managed_hooks_only` to the requirements data flow and preserved explicit `false` values.
- Also adds it to /debug-config
- Marked MDM, system, and legacy managed config layers as managed for hook discovery.
- Updated hook discovery so `allow_managed_hooks_only = true`:
  - keeps managed requirements hooks and managed config-layer hooks,
  - skips user/project/session `hooks.json` and `[hooks]` entries with concise startup warnings,
  - skips current unmanaged plugin hooks,
  - ignores any `allow_managed_hooks_only` key placed in ordinary `config.toml` layers.


codex-rs/hooks/src/engine/discovery.rs
```diff
@@ -65,13 +85,27 @@ pub(crate) fn discover_handlers(
...
+                path: &policy_path,
+                key_source: policy_path.display().to_string(),
+                source: hook_source,
+                is_managed,
+                hook_states: &hook_states,
+                env: HashMap::new(),
+                plugin_id: None,
+            };
+            if !policy.allows(&policy_source) {
+                continue;
```
sayan-oai, line 107: no user facing warning?

codex-rs/hooks/src/engine/discovery.rs
```diff
@@ -350,7 +376,12 @@ fn append_hook_events(
     display_order: &mut i64,
     source: HookHandlerSource<'_>,
     hook_events: HookEventsToml,
+    policy: HookDiscoveryPolicy,
 ) {
+    if !policy.allows(&source) {
+        return;
```
sayan-oai, line 382: no user facing warning?

codex-rs/hooks/src/engine/discovery.rs
```diff
@@ -56,6 +67,15 @@ pub(crate) fn discover_handlers(
     let mut warnings = plugin_hook_load_warnings;
     let mut display_order = 0_i64;
     let hook_states = hook_states_from_stack(config_layer_stack);
+    let policy = HookDiscoveryPolicy {
```
sayan-oai, line 70: codex claims this doesnt cover the legacy `notify` hook path. codex regurgitation:

`A user can still set notify = ["sh", "-c", "..."] in their config and have Codex spawn it after turns even when allow_managed_hooks_only = true`

sayan-oai (approved):

codex-rs/hooks/src/engine/discovery.rs
eternal-openai: Yep, that's a good point - the plan is to deprecate / remove this mechanism, so I think I'll leave that as is

PR #20321, opened 2026-04-30
hook trust metadata and enforcement

abhinav-oai:
# Why

We want shared hook trust that both the app and the TUI can build on, but the metadata is only useful if runtime behavior agrees with it. This PR adds a single backend trust model for hooks so unmanaged hooks cannot run until the current definition has been reviewed, while managed hooks remain runnable and non-configurable.

# What

- persist `trusted_hash` alongside hook state in `config.toml`
- expose `currentHash` and derived `trustStatus` through `hooks/list`
- derive trust from normalized hook definitions so equivalent hooks from `config.toml` and `hooks.json` share the same trust identity
- gate unmanaged hooks on trust before they enter the runnable handler set

# Reviewer Notes

- key file to review is `codex-rs/hooks/src/engine/discovery.rs`
- the only **core** change is schema related


abhinav-oai:
@codex review

codex-rs/hooks/src/engine/discovery.rs
```diff
@@ -396,10 +402,17 @@ fn append_matcher_groups(
                         ));
                         continue;
                     }
+                    let timeout_sec = timeout_sec.unwrap_or(600).max(1);
```
abhinav-oai: ?

codex-rs/hooks/src/engine/discovery.rs
```diff
@@ -36,7 +39,8 @@ struct HookHandlerSource<'a> {
     path: &'a AbsolutePathBuf,
     key_source: String,
     source: HookSource,
-    disabled_hook_keys: &'a HashSet<String>,
+    is_managed: bool,
```
abhinav-oai, line 43: adding explicit `is_managed` instead of inferring from the `HookSource`

we could have `RequirementSource::Unknown` from the managed lane so it gets a bit muddy

codex-rs/hooks/src/engine/discovery.rs
```diff
@@ -478,6 +571,16 @@ fn hook_source_for_config_layer_source(source: &ConfigLayerSource) -> HookSource
     }
 }
 
+fn is_managed_config_layer_source(source: &ConfigLayerSource) -> bool {
```
abhinav-oai: since we have an explicit `is_managed` piped through, do we still need to compute this?

abhinav-oai:
@codex review

codex-rs/hooks/src/engine/discovery.rs
```diff
@@ -396,10 +402,17 @@ fn append_matcher_groups(
...
+                    let timeout_sec = timeout_sec.unwrap_or(600).max(1);
+                    let current_hash = command_hook_hash(
+                        event_name,
+                        matcher,
+                        &command,
+                        timeout_sec,
+                        status_message.as_deref(),
+                    );
                     let command = source.env.iter().fold(command, |command, (key, value)| {
                         command.replace(&format!("${{{key}}}"), value)
```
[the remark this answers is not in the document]
abhinav-oai: I think this is fine since we're doing the substitution and we're trying to protect against changes to the actual command not just path drift

eternal-openai (commented):
https://github.com/openai/codex/pull/20319 is another PR that adds managed flagging, but I think I should probably rebase it on top of yours

please make sure that the TUI change related to this lands with good timing

looking at the hooks spec, I think we should avoid a hand-built trust hash field list. The spec has more security-relevant fields coming (if, shell, asyncRewake, HTTP/MCP/prompt/agent hook fields, etc.), and it’d be easy to add one to the config struct later without remembering to update command_hook_hash. So my rec is to use something derived from the config types (including event_name & matcher so you cant just move a hook to a different event type). Then I think it's possible to save some code by doing something like:

```rust
let value = TomlValue::try_from(normalized_hook_identity)?;
let current_hash = version_for_toml(&value);
```

later we should think about if there's any gains that can be made by hashing entire plugins -- or hooks scripts referenced by the config, but I recognize that as a swamp that doesn't entirely work (maybe with the llm-based scanning idea). but in this, an obvious problem is that replacing the actual script being executed doesn't reset trust

abhinav-oai:
Addressed in ffcc9cc118bec5ec32adc7cb77f8f0e11b37cc12: the trust hash now serializes a typed `NormalizedHookIdentity` built from the config structs (including event name + effective matcher) instead of maintaining a hand-built field list. I kept the existing normalization step before hashing so runtime-equivalent hooks still converge.

pakrym-oai (approved):

PR #20324, opened 2026-04-30
Remove core protocol dependency [1/2]

etraut-openai:
## Why

This stack moves `codex-tui` away from the core protocol event surface and toward app-server API shapes plus TUI-owned local models. This first PR sets up the lower-risk foundation: it introduces the local model surface and extracts app-server event routing into focused TUI modules while preserving the existing behavior for the larger migration in PR2.

This PR is part 1 of a 2-PR stack:

1. Add TUI-owned replacement models and extract app-server event routing.
2. Move the active TUI flow to app-server notifications and delete obsolete adapter code.

## What changed

- Added TUI-owned approval, diff, session state, session resume, token usage, and user-message models.
- Added `app/app_server_event_targets.rs` and `app/app_server_events.rs` to hold app-server event targeting and dispatch logic outside `app.rs`.
- Updated app/status tests to use the local model layer and added focused routing coverage.
- Boxed a few large async TUI test futures so this base layer remains checkable without overflowing the default test stack.

## Verification


fcoury-oai (approved):
Ran a focused local smoke test for the TUI refactor path on this PR: started the PR branch’s TUI in a disposable workspace, exercised a normal read-only turn, an approval-backed file edit through the patch/diff UI, the `/status` surface, and a resume flow. 

Everything worked as expected. Code looks good, approved.

PR #20334, opened 2026-04-30
Make missing config clears no-ops

### LINKED ISSUE #20145: app-server config/value/write returns configPathNotFound when clearing an unset nested config value

`config/value/write` appears to treat JSON `null` as a clear/unset operation, but clearing a missing nested config value currently fails when an intermediate parent table is absent.

For example, clearing `features.personality` from an empty `config.toml` returns `configPathNotFound` / `Path not found` because the `[features]` table does not exist.

This makes reset/clear flows brittle for app-server clients, since a client has to distinguish between “value exists” and “value is already absent” before sending an otherwise idempotent clear.

Start with an empty user `config.toml`.

Call `config/value/write` with:

```json
{
  "keyPath": "features.personality",
  "value": null,
  "mergeStrategy": "replace"
}
```

Actual result:

```text
configPathNotFound / Path not found
```

I reproduced this locally with a focused regression test against `ConfigManager::write_value`: clearing `features.personality` from an empty config fails before the fix because the missing `features` parent table is treated as a path-not-found error.

Clearing an already-unset nested config value should succeed as an idempotent no-op.

The response should be successful, the config file should remain unchanged, and `config/value/write` should not return `configPathNotFound` just because the parent table is absent.

### Additional information

I searched for existing issues/PRs using terms such as `config/value/write`, `configPathNotFound`, `ConfigPathNotFound`, `features.personality`, `null`, and `clear missing nested config path`, and I did not find an exact duplicate.

Related but different issues I found include #10499 and #11728, but those appear to cover different config-write failure modes.

I have a small regression test and narrow fix direction prepared if maintainers agree this should be treated as a no-op.

etraut-openai:
## Why

Fixes #20145.

`config/value/write` treats a JSON `null` value as a request to clear the config key. Clearing a key that is already absent should be idempotent, but clearing a nested key such as `features.personality` from an empty `config.toml` returned `configPathNotFound` because `clear_path` treated the missing `features` parent table as an error.

That makes app-server reset flows brittle because clients have to read first and avoid sending a clear request unless the parent path already exists.

## What Changed

- Updated app-server config clearing so missing intermediate tables, or non-table parents, are treated as an unchanged no-op.
- Removed the now-unreachable `MergeError::PathNotFound` path from config write merging.
- Added a regression test covering `features.personality = null` against an empty user config.

## Verification

- `cargo test -p codex-app-server clear_missing_nested_config_is_noop`
- `cargo test -p codex-app-server` was run; the config manager unit suite passed, but one unrelated integration test failed because `turn_start_emits_thread_scoped_warning_notification_for_trimmed_skills` expected `7` trimmed skills and observed `8`.

### DIFF
--- a/codex-rs/app-server/src/config_manager_service.rs
+++ b/codex-rs/app-server/src/config_manager_service.rs
@@ -244,10 +244,6 @@ impl ConfigManager {
 
             apply_merge(&mut user_config, &segments, parsed_value.as_ref(), strategy).map_err(
                 |err| match err {
-                    MergeError::PathNotFound => ConfigManagerError::write(
-                        ConfigWriteErrorCode::ConfigPathNotFound,
-                        "Path not found",
-                    ),
                     MergeError::Validation(message) => ConfigManagerError::write(
                         ConfigWriteErrorCode::ConfigValidationError,
                         message,
@@ -413,7 +409,6 @@ fn parse_key_path(path: &str) -> Result<Vec<String>, String> {
 
 #[derive(Debug)]
 enum MergeError {
-    PathNotFound,
     Validation(String),
 }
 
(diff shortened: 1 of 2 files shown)


jif-oai (approved):
make sense

PR #20336, opened 2026-04-30
execpolicy: unwrap PowerShell -Command wrappers on Windows

iceweasel-oai:
## Why
On Windows, Codex runs shell commands through a top-level `powershell.exe -NoProfile -Command ...` wrapper. `execpolicy` was matching that wrapper instead of the inner command, so prefix rules like `["git", "push"]` did not fire for PowerShell-wrapped commands even though the same normalization already happens for `bash -lc` on Unix.

This change makes the Windows shell wrapper transparent to rule matching while preserving the existing Windows unmatched-command safelist and dangerous-command heuristics.

## What changed
- add `parse_powershell_command_plain_commands()` in `shell-command/src/powershell.rs` to unwrap the top-level PowerShell `-Command` body with `extract_powershell_command()` and parse it with the existing PowerShell AST parser
- update `core/src/exec_policy.rs` so `commands_for_exec_policy()` treats top-level PowerShell wrappers like `bash -lc` and evaluates rules against the parsed inner commands
- carry a small `ExecPolicyCommandOrigin` through unmatched-command evaluation and expose `is_safe_powershell_words()` / `is_dangerous_powershell_words()` so Windows safelist and dangerous-command checks still work after unwrap
- add Windows-focused tests for wrapped PowerShell prompt/allow matches, wrapper parsing, and unmatched safe/dangerous inner commands, and re-enable the end-to-end `execpolicy_blocks_shell_invocation` test on Windows

## Testing
- `cargo test -p codex-shell-command`

### DIFF
--- a/codex-rs/core/src/exec_policy.rs
+++ b/codex-rs/core/src/exec_policy.rs
@@ -25,8 +25,10 @@ use codex_protocol::permissions::FileSystemSandboxKind;
 use codex_protocol::permissions::FileSystemSandboxPolicy;
 use codex_protocol::protocol::AskForApproval;
 use codex_shell_command::is_dangerous_command::command_might_be_dangerous;
+#[cfg(windows)]
 use codex_shell_command::is_dangerous_command::is_dangerous_powershell_words;
 use codex_shell_command::is_safe_command::is_known_safe_command;
+#[cfg(windows)]
 use codex_shell_command::is_safe_command::is_safe_powershell_words;
 use thiserror::Error;
 use tokio::fs;
@@ -105,6 +107,7 @@ static BANNED_PREFIX_SUGGESTIONS: &[&[&str]] = &[
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub(crate) enum ExecPolicyCommandOrigin {
     Direct,
+    #[cfg(windows)]
     PowerShell,
 }
 
@@ -604,6 +607,7 @@ pub(crate) fn render_decision_for_unmatched_command(
 ) -> Decision {
     let is_known_safe = match command_origin {
         ExecPolicyCommandOrigin::Direct => is_known_safe_command(command),
+        #[cfg(windows)]
         ExecPolicyCommandOrigin::PowerShell => is_safe_powershell_words(command),
     };
     if is_known_safe && !used_complex_parsing {
@@ -627,6 +631,7 @@ pub(crate) fn render_decision_for_unmatched_command(
     // forbid the command.
     let command_is_dangerous = match command_origin {
         ExecPolicyCommandOrigin::Direct => command_might_be_dangerous(command),
+        #[cfg(windows)]
         ExecPolicyCommandOrigin::PowerShell => is_dangerous_powershell_words(command),
     };
     if command_is_dangerous || environment_lacks_sandbox_protections {
--- a/codex-rs/core/src/exec_policy.rs
+++ b/codex-rs/core/src/exec_policy.rs
@@ -25,7 +25,9 @@ use codex_protocol::permissions::FileSystemSandboxKind;
 use codex_protocol::permissions::FileSystemSandboxPolicy;
 use codex_protocol::protocol::AskForApproval;
 use codex_shell_command::is_dangerous_command::command_might_be_dangerous;
+use codex_shell_command::is_dangerous_command::is_dangerous_powershell_words;
 use codex_shell_command::is_safe_command::is_known_safe_command;
+use codex_shell_command::is_safe_command::is_safe_powershell_words;
 use thiserror::Error;
 use tokio::fs;
 use tokio::sync::Semaphore;
@@ -99,6 +101,12 @@ static BANNED_PREFIX_SUGGESTIONS: &[&[&str]] = &[
     &["lua", "-e"],
     &["osascript"],
 ];
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum ExecPolicyCommandOrigin {
+    Direct,
+    PowerShell,
+}
 
 pub(crate) fn child_uses_parent_exec_policy(parent_config: &Config, child_config: &Config) -> bool {
     fn exec_policy_config_folders(config: &Config) -> Vec<AbsolutePathBuf> {
@@ -248,7 +256,7 @@ impl ExecPolicyManager {
             prefix_rule,
         } = req;
         let exec_policy = self.current();
-        let (commands, used_complex_parsing) = commands_for_exec_policy(command);
+        let (commands, used_complex_parsing, command_origin) = commands_for_exec_policy(command);
         // Keep heredoc prefix parsing for rule evaluation so existing
         // allow/prompt/forbidden rules still apply, but avoid auto-derived
         // amendments when only the heredoc fallback parser matched.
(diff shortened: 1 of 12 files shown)


iceweasel-oai:
@codex review

codex-rs/core/src/exec_policy.rs
```diff
@@ -705,6 +707,13 @@ fn commands_for_exec_policy(command: &[String]) -> (Vec<Vec<String>>, bool) {
         return (commands, false);
     }
 
+    #[cfg(windows)]
+    if let Some(commands) = parse_powershell_command_plain_commands(command)
+        && !commands.is_empty()
+    {
+        return (commands, false);
```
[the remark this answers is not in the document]
iceweasel-oai: fixed in https://github.com/openai/codex/pull/20336/changes/25baa6db8b7b04772e79584647204ed9fed58eef

iceweasel-oai:
@codex review

codex-rs/core/src/tools/runtimes/shell/unix_escalation.rs
```diff
@@ -674,13 +674,16 @@ fn evaluate_intercepted_exec_policy(
...
     let fallback = |cmd: &[String]| {
         crate::exec_policy::render_decision_for_unmatched_command(
-            approval_policy,
-            &permission_profile,
-            file_system_sandbox_policy,
-            sandbox_cwd,
             cmd,
-            sandbox_permissions,
-            used_complex_parsing,
+            crate::exec_policy::UnmatchedCommandContext {
```
iceweasel-oai, line 678: this struct was introduced to fix a clippy warning about too many parameters due to me adding one new one (`command_origin`)

codex-rs/core/src/exec_policy.rs
bolinfest, line 1038: Can you create a separate file `exec_policy_windows_tests.rs` and include it here in the same way, but with `#[cfg(all(windows, test))]` so you don't have to `#[cfg(windows)]` in `exec_policy_windows_tests` itself?

codex-rs/core/src/exec_policy.rs
```diff
@@ -698,18 +739,31 @@ fn default_policy_path(codex_home: &Path) -> PathBuf {
     codex_home.join(RULES_DIR_NAME).join(DEFAULT_POLICY_FILE)
 }
 
-fn commands_for_exec_policy(command: &[String]) -> (Vec<Vec<String>>, bool) {
+fn commands_for_exec_policy(
+    command: &[String],
+) -> (Vec<Vec<String>>, bool, ExecPolicyCommandOrigin) {
```
bolinfest: I take responsibility for the original tuple return value being bad, so since you're touching this code, can you declare a small struct before this function to use as a return value and return that?

codex-rs/core/src/exec_policy.rs
```diff
@@ -25,7 +25,11 @@ use codex_protocol::permissions::FileSystemSandboxKind;
 use codex_protocol::permissions::FileSystemSandboxPolicy;
 use codex_protocol::protocol::AskForApproval;
 use codex_shell_command::is_dangerous_command::command_might_be_dangerous;
+#[cfg(windows)]
```
bolinfest: If you only use the function in one block in this file and that block already has `#[cfg(windows)]`, I would favor doing the import in that block so the top-level imports are less noisy.

We should add that to `AGENTS.md`...

codex-rs/core/src/exec_policy.rs
```diff
@@ -98,6 +104,24 @@ static BANNED_PREFIX_SUGGESTIONS: &[&[&str]] = &[
     &["osascript"],
 ];
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum ExecPolicyCommandOrigin {
```
bolinfest, line 110: This needs a docstring because I'm a little fuzzy on what this means.

`ExecPolicyCommandOrigin::PowerShell` to me implies the command was targeted to be run under PowerShell, but looking at the code, it feels like it actually means "the type of shell parser that was used to parse the command." It's possible the user is on Windows but could be using a shell that isn't PowerShell, isn't it? Or no?

To that end, I'm not clear why the other variant is named `Direct`.

codex-rs/core/src/exec_policy_tests.rs
```diff
@@ -656,11 +656,65 @@ async fn evaluates_bash_lc_inner_commands() {
...
+                "-Command".to_string(),
+                "echo blocked".to_string(),
+            ],
+            approval_policy: AskForApproval::Never,
+            sandbox_policy: SandboxPolicy::DangerFullAccess,
+            file_system_sandbox_policy: unrestricted_file_system_sandbox_policy(),
+            sandbox_permissions: SandboxPermissions::UseDefault,
+            prefix_rule: None,
+        },
+        ExecApprovalRequirement::Forbidden {
```
bolinfest: I'm confused, why isn't this `ExecApprovalRequirement::NeedsApproval`?

codex-rs/core/tests/suite/exec_policy.rs
```diff
@@ -81,11 +81,6 @@ fn assert_no_matched_rules_invariant(output_item: &Value) {
 
 #[tokio::test]
 async fn execpolicy_blocks_shell_invocation() -> Result<()> {
-    // TODO execpolicy doesn't parse powershell commands yet
```
bolinfest, line 84: Well, that explains that...

codex-rs/shell-command/src/powershell.rs
```diff
@@ -68,6 +69,16 @@ pub fn extract_powershell_command(command: &[String]) -> Option<(&str, &str)> {
     None
 }
 
+/// Parse the script body from a top-level PowerShell wrapper into argv-like commands.
+///
+/// This is intentionally narrower than the Windows safe-command parser: it only unwraps the
+/// `-Command`/`-c` body from a PowerShell invocation we already recognize, then delegates the
+/// script itself to the PowerShell AST parser.
+pub fn parse_powershell_command_plain_commands(command: &[String]) -> Option<Vec<Vec<String>>> {
```
bolinfest: Maybe `parse_powershell_command_into_plain_commands()`?

codex-rs/shell-command/src/powershell.rs
```diff
@@ -68,6 +69,16 @@ pub fn extract_powershell_command(command: &[String]) -> Option<(&str, &str)> {
...
+    let (shell, script) = extract_powershell_command(command)?;
+    try_parse_powershell_ast_commands(shell, script)
```
bolinfest: I think `executable` is the more appropriate name here.

```suggestion
    let (executable, script) = extract_powershell_command(command)?;
    try_parse_powershell_ast_commands(executable, script)
```

codex-rs/shell-command/src/powershell.rs
```diff
@@ -186,4 +199,18 @@ mod tests {
         let (_shell, script) = extract_powershell_command(&cmd).expect("extract");
         assert_eq!(script, "Get-ChildItem | Select-String foo");
     }
+
+    #[cfg(windows)]
+    #[test]
+    fn parses_plain_powershell_commands() {
+        let commands = parse_powershell_command_plain_commands(&[
```
bolinfest: Also add a test for a case where `commands` has more than one entry?

bolinfest (approved):

PR #20348, opened 2026-04-30
Move plugin out of core.

xl-openai:


xl-openai:
@codex review this PR with a focus on migration feature parity. Please verify that the behavior after the move is exactly the same as before, and flag any regressions, missing paths, or subtle API/behavior differences introduced by the migration.

xl-openai:
Re `app-server-client::legacy_core::plugins`: this removal is intentional; the legacy shim should not continue exposing plugin manager APIs after the migration, and callers should depend on `codex-core-plugins` directly.

codex-rs/core/src/lib.rs
```diff
@@ -67,7 +67,7 @@ pub use message_history::history_metadata as message_history_metadata;
 pub use message_history::lookup as lookup_message_history_entry;
 pub use utils::path_utils;
 pub mod personality_migration;
-pub mod plugins;
+pub(crate) mod plugins;
```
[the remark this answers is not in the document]
xl-openai: This is intentional for this PR: `codex_core::plugins` should stop being a public plugin API surface, and callers should import from `codex-core-plugins` directly.

codex-rs/core-api/src/lib.rs
```diff
@@ -35,7 +35,6 @@ pub use codex_core::config::Permissions;
 pub use codex_core::config::TerminalResizeReflowConfig;
 pub use codex_core::config::ThreadStoreConfig;
 pub use codex_core::config::find_codex_home;
-pub use codex_core::plugins::PluginsManager;
 pub use codex_core::skills::SkillsManager;
```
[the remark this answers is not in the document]
xl-openai: This removal is intentional: `core-api` should not continue re-exporting `PluginsManager` now that the plugin manager lives in `codex-core-plugins`.

xl-openai:
@codex review this PR for behavioral parity only. Treat the removed `codex_core::plugins`, `core-api`, and `app-server-client::legacy_core::plugins` re-exports as intentional and out of scope. Do not flag issues already present on `origin/main`; only flag behavior changed by this PR.

codex-rs/core-plugins/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 pub mod installed_marketplaces;
 pub mod loader;
+pub mod manager;
```
sayan-oai: does this need to be pub?

codex-rs/core/src/plugins/mod.rs
```diff
@@ -1,45 +1,31 @@
...
 
+pub(crate) fn plugins_config_input_from_config(config: &Config) -> PluginsConfigInput {
+    PluginsConfigInput::new(
+        config.config_layer_stack.clone(),
+        config.features.enabled(Feature::Plugins),
+        config.features.enabled(Feature::RemotePlugin),
+        config.features.enabled(Feature::PluginHooks),
+        config.chatgpt_base_url.clone(),
+    )
+}
```
sayan-oai: this logic is defined 5 times across the code. maybe make it a public helper on Config, or put it in some shared crate, but we shouldn't duplicate it so much

codex-rs/core-plugins/src/manager.rs
```diff
@@ -72,14 +72,40 @@ use std::sync::atomic::AtomicBool;
...
 use tracing::info;
 use tracing::warn;
 
 static CURATED_REPO_SYNC_STARTED: AtomicBool = AtomicBool::new(false);
 const FEATURED_PLUGIN_IDS_CACHE_TTL: std::time::Duration =
     std::time::Duration::from_secs(60 * 60 * 3);
 
+#[derive(Debug, Clone)]
+pub struct PluginsConfigInput {
+    pub config_layer_stack: ConfigLayerStack,
```
sayan-oai, line 84: codex:

ConfigLayerStack is owned by PluginsConfigInput in this setup, and ConfigLayerStack is a big handle. 

make this a borrowed view, e.g. PluginsConfigInput<'a> { config_layer_stack: &'a ConfigLayerStack, chatgpt_base_url: &'a str, ... }

sayan-oai (approved):
thanks for cleaning this up 🙏

please test very thoroughly!

codex-rs/core-plugins/src/manager.rs
xl-openai: Agreed this would be cleaner, but `PluginsConfigInput` is currently moved into startup/auto-upgrade/background tasks, so making it borrowed requires a larger owned-snapshot split; I’m leaving that for a follow-up to keep this migration scoped.

codex-rs/core-plugins/src/lib.rs
xl-openai: Done — changed `manager` to a private module and kept the intended public surface through the existing top-level `pub use manager::...` exports.

codex-rs/core/src/plugins/mod.rs
xl-openai: Done — added `Config::plugins_config_input()` and updated the core/app-server/chatgpt/cli/tui callsites to use it instead of duplicating the same construction.

PR #20437, opened 2026-04-30
feat: add `session_id`

jif-oai:
## Summary

Related to https://openai.slack.com/archives/C095U48JNL9/p1777537279707449
TLDR:
We update the meaning of session ids and thread ids:
* thread_id stays as now
* session_id become a shared id between every thread under a /root thread (i.e. every sub-agent share the same session id)

This PR introduces an explicit `SessionId` and threads it through the protocol/client boundary so `session_id` and `thread_id` can diverge when they need to, while preserving compatibility for older serialized `session_configured` events.


jif-oai:
@codex review

codex-rs/tui/src/app_server_session.rs
```diff
@@ -1791,6 +1791,7 @@ mod tests {
         let forked_from_id = ThreadId::new();
         let read_only_profile = PermissionProfile::read_only();
         let response = ThreadResumeResponse {
+            session_id: ThreadId::new().to_string(),
```
pakrym-oai, line 1816: do we need to expose session id via app-server?

pakrym-oai (approved):

jif-oai:
Flake come from `threads_started_via_app_server_are_visible_through_typed_requests` which is unrelated. Will fix in a follow-up

PR #20447, opened 2026-04-30
Fix MCP status/list lifecycle leak

jif-oai:
## Summary

Follow-up to #19753 after investigating reports of stale MCP server processes and app-server logs like `Failed to terminate MCP process group`.

`mcpServerStatus/list` is declared as a globally serialized RPC, but the handler was spawning the actual MCP inventory work into a detached task and returning immediately. That let multiple status-list requests overlap outside the serialization queue and connection RPC gate, with each request able to create a temporary `McpConnectionManager` and start MCP clients. Similar detached-task behavior existed for app-server MCP resource reads and tool calls, which could keep thread/MCP state alive after the original RPC handler was considered complete.

This change keeps the MCP work inside the request future so the existing serialization and connection lifecycle tracking actually cover it.

## Changes

- Await `mcpServerStatus/list`, `mcpServer/resource/read`, and `mcpServer/tool/call` work directly in the app-server RPC handlers instead of detaching it with `tokio::spawn`.
- Explicitly shut down temporary MCP managers used by threadless resource reads and status snapshots after their work completes.
- Harden stdio MCP process termination so local termination failures are surfaced, keep the process handle retryable, and report failed `taskkill` status on Windows.
- Add a regression test proving overlapping `mcpServerStatus/list` requests do not run concurrent inventory work.

## Verification

- `cargo test -p codex-app-server mcp_server_status_list_serializes_inventory_work -- --nocapture`
- `cargo test -p codex-mcp`
- `cargo test -p codex-rmcp-client shutdown_kills_initialized_stdio_server_with_in_flight_operation -- --nocapture`

### DIFF
--- a/codex-rs/app-server/src/codex_message_processor.rs
+++ b/codex-rs/app-server/src/codex_message_processor.rs
@@ -3604,7 +3604,7 @@ impl CodexMessageProcessor {
 
         let (thread_id, thread) = self.load_thread(&thread_id).await?;
 
-        let request = request_id.clone();
+        let request = request_id;
 
         let rollback_already_in_progress = {
             let thread_state = self.thread_state_manager.thread_state(thread_id).await;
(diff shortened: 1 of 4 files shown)


jif-oai:
@codex review

PR #20463, opened 2026-04-30
feat(rollouts): store EventMsg::ApplyPatchEnd in limited history mode

owenlin0:
The Codex App treats apply patch tool calls quite load-bearing in the UI (always shown on a completed turn), so we'd like to persist `EventMsg::ApplyPatchEnd` to guarantee that when a client reconnects to app-server mid-turn, we always have the full diff to display at the end of that turn.


codex-rs/rollout/src/policy.rs
```diff
@@ -223,4 +227,29 @@ mod tests {
             EventPersistenceMode::Limited
         ));
     }
+
+    #[test]
+    fn persists_patch_apply_end_events_in_limited_mode() {
```
jif-oai: tbh I don't think this test brings anything but load on the CI pipeline

jif-oai (approved):

codex-rs/rollout/src/policy.rs
owenlin0: fair, down to remove

owenlin0: deleted the entire suite since they all did similar things

PR #20464, opened 2026-04-30
feat(rollouts): store EventMsg::McpToolCallEnd in limited history mode

owenlin0:
Similar to https://github.com/openai/codex/pull/20463, we want to persist MCP tool calls even in limited history mode because we want a client to be able to reconnect to a running app-server that has called, let's say, a slack MCP and the slack MCP responded with an elicitation, we want to be able to show that elicitation when a client reconnects.


owenlin0:
putting this to draft since i am not sure we want this, actually

PR #20502, opened 2026-04-30
fix(tui): set persist_extended_history: false

owenlin0:
Large rollouts are no good. This updates the TUI to behave the same as the Codex App, which is also turning it off.


etraut-openai (approved):
lg

PR #20514, opened 2026-04-30
[codex-analytics] add item lifecycle timing

rhan-oai:
## Why

Tool families already disagree on what their existing `duration` fields mean, so lifecycle latency should live on the shared item envelope instead of being inferred from per-tool execution fields. Carrying that envelope through app-server notifications gives downstream consumers one reusable timing signal without pretending every tool has the same execution semantics.

## What changed

- Adds `started_at_ms` to core `ItemStartedEvent` values and `completed_at_ms` to core `ItemCompletedEvent` values.
- Populates those timestamps in the shared session lifecycle emitters, so protocol-native items get timing without each producer tracking its own clock state.
- Exposes `startedAtMs` on app-server `item/started` notifications and `completedAtMs` on `item/completed` notifications.
- Maps the lifecycle timestamps through the app-server boundary while leaving legacy-converted notifications nullable when no lifecycle timestamp exists.
- Regenerates the app-server JSON schema and TypeScript fixtures for the notification-envelope change and updates downstream fixtures that construct those notifications directly.
- Extends the existing web-search and image-generation integration flows to assert the new lifecycle timestamps on the native item events.

## Verification

- `cargo check -p codex-protocol -p codex-core -p codex-app-server-protocol -p codex-app-server -p codex-tui -p codex-exec -p codex-app-server-client`
- `cargo test -p codex-core --test all web_search_item_is_emitted`
- `cargo test -p codex-core --test all image_generation_call_event_is_emitted`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/20514).
* #18748
* #18747
* #17090
* #17089
* __->__ #20514


codex-rs/protocol/src/protocol.rs
```diff
@@ -1828,6 +1828,9 @@ pub struct ItemStartedEvent {
     pub thread_id: ThreadId,
     pub turn_id: String,
     pub item: TurnItem,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub started_at_ms: Option<i64>,
```
owenlin0: let's make this non-optional. it's safe because we never persisted `ItemStartedEvent` in rollouts :)

codex-rs/protocol/src/protocol.rs
```diff
@@ -1854,6 +1857,9 @@ pub struct ItemCompletedEvent {
     pub thread_id: ThreadId,
     pub turn_id: String,
     pub item: TurnItem,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[ts(optional)]
+    pub completed_at_ms: Option<i64>,
```
owenlin0: Let's make this non-optional too. Can we do something like this?

```
fn unknown_completed_at_ms() -> i64 {
    0
}

pub struct ItemCompletedEvent {
    // other fields
    // ...
    #[serde(default = "unknown_completed_at_ms")]
    pub completed_at_ms: i64,
}
```

I think we need this since we read `EventMsg::ItemCompleted(TurnItem::Plan(_))` from rollout files, so we want to keep that deserializing properly with a `serde(default)`.

owenlin0: and i think it's totally fine to return 0 since app-server only exposes `item/completed` notification while streaming a thread live. APIs that return thread history doesn't use this.

owenlin0 (commented):
let's see if we can make these timestamps non-optional. first at the core layer, then at app-server (if possible). want to give it a shot? this would be much nicer instead of keeping around optional forever 🙏

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -6928,6 +6925,8 @@ pub struct ItemStartedNotification {
     pub item: ThreadItem,
     pub thread_id: String,
     pub turn_id: String,
+    /// Unix timestamp (in milliseconds) when this item lifecycle started.
+    pub started_at_ms: i64,
```
owenlin0, line 6933: let's make sure to add the `#[ts(type = "number")]` annotation for this field

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -6990,6 +6989,8 @@ pub struct ItemCompletedNotification {
     pub item: ThreadItem,
     pub thread_id: String,
     pub turn_id: String,
+    /// Unix timestamp (in milliseconds) when this item lifecycle completed.
+    pub completed_at_ms: i64,
```
owenlin0, line 6998: here too

codex-rs/protocol/src/protocol.rs
```diff
@@ -1854,6 +1855,14 @@ pub struct ItemCompletedEvent {
     pub thread_id: ThreadId,
     pub turn_id: String,
     pub item: TurnItem,
+    // Old rollout files may not have this field. Default to 0 so they still
+    // deserialize after tightening the core event contract.
```
owenlin0: nit: can we call out `ItemCompleted for PlanItem` in particular? that's the only one that ever got persisted

owenlin0 (approved):
small comments but overall lgtm!

PR #20515, opened 2026-04-30
[codex-analytics] expose item timing through app server

rhan-oai:
## Why

Once the core lifecycle events carry timestamps, app-server clients need the same envelope fields on the public notifications. Keeping the timing on `item/started` and `item/completed` avoids expanding every `ThreadItem` variant with lifecycle fields, and leaves any existing execution-duration fields free to keep their narrower semantics.

## What changed

- Exposes `startedAtMs` on `item/started` and `completedAtMs` on `item/completed` notifications.
- Maps protocol-native lifecycle timestamps through the app-server boundary while leaving legacy-converted notifications nullable when no lifecycle timestamp exists.
- Regenerates the app-server JSON schema and TypeScript fixtures for the notification-envelope changes only.
- Updates downstream fixtures and consumers that construct the public notification structs directly.

## Verification

- `cargo test -p codex-core -p codex-app-server-protocol -p codex-app-server -p codex-tui -p codex-exec -p codex-app-server-client --no-run`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/openai/codex/pull/20515).
* #18748
* #18747
* #17090
* #17089
* #20514
* __->__ #20515
* #20300


rhan-oai:
Collapsed into #20514 so the lifecycle timing work is reviewed as one PR.

PR #20524, opened 2026-04-30
deprecate legacy notify

abhinav-oai:
# Why

`notify` is the remaining compatibility surface from the legacy hook implementation. The newer lifecycle hook engine now owns the active hook system, so we should start steering users away from adding new `notify` configs before removing the old path entirely. This also adds a lightweight watchpoint for the deprecation so we can see how much legacy usage remains before the clean drop.

# What

- emit a startup deprecation notice when a non-empty `notify` command is configured
- emit `codex.notify.configured` when a session starts with legacy `notify` configured
- emit `codex.notify.run` when the legacy notify path fires after a completed turn
- mark `notify` as deprecated in the config schema and repo docs
- remove the orphaned `codex-rs/hooks/src/user_notification.rs` file that is no longer compiled
- add regression coverage for the new deprecation notice

# Next steps

A follow-up PR can remove the legacy notify path entirely once we are ready for the clean drop. Before then, we can watch `codex.notify.configured` and `codex.notify.run` to understand the deprecation impact and remaining active usage. The cleanup PR should then delete the `notify` config field, the `legacy_notify` implementation, the old compatibility dispatch types and callsites that only exist for the legacy path, and the remaining compatibility docs/tests.

# Testing

- `cargo test -p codex-hooks`
- `cargo test -p codex-core emits_deprecation_notice_for_notify`


abhinav-oai:
@codex review

eternal-openai (approved):

codex-rs/core/src/session/session.rs
```diff
@@ -558,6 +559,24 @@ impl Session {
...
+                .is_some_and(|argv| !argv.is_empty() && !argv[0].is_empty());
+            if legacy_notify_configured {
+                post_session_configured_events.push(Event {
+                    id: INITIAL_SUBMIT_ID.to_owned(),
+                    msg: EventMsg::DeprecationNotice(DeprecationNoticeEvent {
+                        summary:
+                            "`notify` is deprecated and will be removed in a future release."
+                                .to_string(),
+                        details: Some(
+                            "Use lifecycle hooks for new automation instead of adding new `notify` configurations."
```
etraut-openai: I think this needs to be clearer. Ideally, the user can figure out how to address the problem based entirely on this advice. Could we say something like "Switch to `stop` hook". Maybe even offer a link to the documentation page?

docs/config.md
```diff
@@ -60,7 +60,9 @@ disabled_tools = [
 
 ## Notify
 
-Codex can run a notification hook when the agent finishes a turn. See the configuration reference for the latest notification settings:
```
etraut-openai, line 63: We shouldn't be adding documentation to the OSS repo. The docs live in a different (external) repo, and we don't want to try to keep two copies of the docs in sync. 

No need to fix in this PR. It looks like others have been adding docs here too. I'll do a separate clean-up pass after this PR lands. I'll also update the AGENTS.md guidance because that's probably causing codex to add docs here.

etraut-openai (approved):
Overall, looks good. I think the deprecation warning text could be improved.

PR #20527, opened 2026-04-30
Support PreToolUse updatedInput rewrites

abhinav-oai:
## Why

`PreToolUse` already exposes `updatedInput` in its hook output schema, but Codex currently rejects it instead of applying the rewrite. That leaves hook authors unable to make the documented pre-execution adjustment to a tool call before it runs.

## What

- Accept `updatedInput` from `PreToolUse` hooks when paired with `permissionDecision: "allow"`.
- Apply the rewritten input before dispatch so the tool executes the updated payload, not the original one.
- Preserve the stable hook-facing compatibility shapes that participating tool handlers expose:
  - Bash-like tools (`shell`, `container.exec`, `local_shell`, `shell_command`, `exec_command`) use `{ "command": ... }`.
  - `apply_patch` exposes its patch body through the same command-shaped hook contract.
  - MCP tools expose their JSON argument object directly.
- Keep each participating tool handler responsible for translating hook-facing `updatedInput` back into its concrete invocation shape.

## Verification

Direct Bash-like rewrite coverage:

- `pre_tool_use_rewrites_shell_before_execution`
- `pre_tool_use_rewrites_container_exec_before_execution`
- `pre_tool_use_rewrites_local_shell_before_execution`
- `pre_tool_use_rewrites_shell_command_before_execution`
- `pre_tool_use_rewrites_exec_command_before_execution`

These cases assert that each supported Bash-like surface runs only the rewritten command while the hook still observes the original `{ "command": ... }` input.

`pre_tool_use_rewrites_apply_patch_before_execution`

- Model emits one patch.
- Hook swaps in a different patch.
- Asserts only the rewritten file is created, and the hook saw the original patch.

`pre_tool_use_rewrites_code_mode_nested_exec_command_before_execution`

- Model runs one nested shell command from code mode.
- Hook rewrites it.
- Asserts only the rewritten command runs, and the hook saw the original nested input.

`pre_tool_use_rewrites_mcp_tool_before_execution`

- Model calls the RMCP echo tool.
- Hook rewrites the MCP arguments.
- Asserts the MCP server receives and returns the rewritten message, not the original one.


abhinav-oai:
@codex review

abhinav-oai:
@codex review

abhinav-oai:
> https://github.com/openai/codex/blob/04ae865e76e43e38bd529702ffcf5c9e448ee1f0/codex-rs/hooks/src/events/pre_tool_use.rs#L103-L106
**<sub><sub>P1 Badge</sub></sub>  Re-run policy hooks against rewritten input**

I think this is working as intended under the trust model for hooks
- `updatedInput` is only produced by hooks the user has explicitly trusted and opted into
- so `permissionDecision: "allow"` with an `updatedInput` is the hook author saying "allow this call, but execute the rewritten payload instead."

Re-running the same trusted `PreToolUse` hook set against that rewritten payload would add another pass, but it would not strengthen the boundary we are relying on here as the rewrite itself came from trusted hook code.

codex-rs/core/src/tools/registry.rs
```diff
@@ -260,14 +288,12 @@ impl ToolRegistry {
...
         &self,
-        invocation: ToolInvocation,
+        mut invocation: ToolInvocation,
     ) -> Result<AnyToolResult, FunctionCallError> {
         let tool_name = invocation.tool_name.clone();
         let display_name = tool_name.display();
         let call_id_owned = invocation.call_id.clone();
         let otel = invocation.turn.session_telemetry.clone();
-        let payload_for_response = invocation.payload.clone();
-        let log_payload = payload_for_response.log_payload();
```
abhinav-oai, line 296: shifted this down so we log the rewritten tool payload

abhinav-oai:
@codex review

abhinav-oai:
@codex review

eternal-openai (dismissed):
hmm, one substantial concern and two minor ones:

1. so an MCP tool like mcp__foo__exec_command has name = "exec_command" -> since we only switch on the name, we could mistake this MCP tool as being the real builtin one
3. hooks spec says that when multiple PreToolUse hooks rewrite input, the last one to finish wins, with nondeterministic ordering because the hooks run in parallel. This implementation instead makes the last configured hook always win, which is a spec mismatch
5. since the PR description says we only support updatedInput paired with "allow", bare "allow" should still be rejected until the actual allow semantics are implemented

codex-rs/core/src/hook_runtime/tool_compat.rs
abhinav-oai: wait should this branch fail like we do in apply_updated_input since we're exiting early for mcp?

pakrym-oai (approved):

PR #20540, opened 2026-05-01
Move apply-patch file changes into turn items

pakrym-oai:
## Why

Apply-patch file changes are now part of the core turn item stream, so v2 clients can consume the same first-class item lifecycle path used by other turn items instead of relying on app-server-specific remapping from legacy patch events.

## What changed

- Added a core `TurnItem::FileChange` carrying apply-patch changes and completion metadata.
- Updated the apply-patch tool emitter to send `ItemStarted` / `ItemCompleted` with the new `FileChange` item while preserving legacy `PatchApplyBegin` / `PatchApplyEnd` fan-out.
- Updated app-server v2 conversion to render the new core item directly and stopped `event_mapping` from remapping old patch begin/end events into item notifications.
- Kept thread history reconstruction based on the existing old apply-patch events for rollout compatibility.

## Verification

- `cargo test -p codex-protocol -p codex-app-server-protocol`
- `cargo test -p codex-core --test all apply_patch_tool_executes_and_emits_patch_events`
- `cargo test -p codex-app-server bespoke_event_handling`

### DIFF
--- a/codex-rs/app-server-protocol/src/protocol/event_mapping.rs
+++ b/codex-rs/app-server-protocol/src/protocol/event_mapping.rs
@@ -1,7 +1,6 @@
 use crate::protocol::common::ServerNotification;
 use crate::protocol::item_builders::build_command_execution_begin_item;
 use crate::protocol::item_builders::build_command_execution_end_item;
-use crate::protocol::item_builders::build_file_change_begin_item;
 use crate::protocol::item_builders::convert_patch_changes;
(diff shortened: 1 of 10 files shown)


owenlin0 (approved):
nice, this makes the app-server layer simpler too! 

at some point (maybe once we migrate everything to CoreTurnItems?) we should persist these in the rollout instead of the legacy EventMsg ones

Issue #17492, opened 2026-04-11
tool_call_mcp_elicitation=false does not suppress elicitation.form in MCP initialize (Graylog handshake fails with -32603)


mbonenberger:
### What version of Codex CLI is running?

codex-cli 0.120.0

### What subscription do you have?

Pro

### Which model were you using?

gpt-5.3-codex

### What platform is your computer?

Linux 6.8.0-106-generic x86_64 x86_64

### What terminal emulator and version are you using (if applicable)?

running in a generic Ubuntu Linux shell environment

### What issue are you seeing?

When starting Codex with a configured Graylog MCP server, MCP startup fails during handshake with:

```
MCP client for `graylog` failed to start: MCP startup failed: handshaking with MCP server failed: JSON-RPC error: -32603: Unrecognized field "form" (class io.modelcontextprotocol.spec.McpSchema$ClientCapabilities$Elicitation), not marked as ignorable (0 known properties: ]) at [Source: UNKNOWN; byte offset: #UNKNOWN] (through reference chain:

io.modelcontextprotocol.spec.McpSchema$InitializeRequest["capabilities"] >io.modelcontextprotocol.spec.McpSchema$ClientCapabilities["elicitation"]->io.modelcontextprotocol.spec.McpSchema$ClientCapabilities$Elicitation["form"])
```

Then Codex shows:

 ⚠ MCP startup incomplete (failed: graylog)

### Reproduction details

Direct request to Graylog MCP endpoint with elicitation.form reproduces the server-side failure:

```
curl -i -X POST \
  -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{"elicitation": {"form":true}},"clientInfo":{"name":"probe","version":"1.0"}}}' \  
"http://<REDACTED_TOKEN>:token@127.0.0.1:9000/api/mcp"
````
  Response:

```
HTTP/1.1 500 Internal Server Error
  
{"jsonrpc":"2.0","id":1,"error":{"code":-32603,"message":"Unrecognized field \"form\" ..."}}
```
  
Control check (no elicitation.form) succeeds:

```
curl -i -X POST \
  -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo": {"name":"probe","version":"1.0"}}}' \
[...]

PR #20562, opened 2026-05-01
Use the 2025-06-18 elicitation capability shape

abhinav-oai:
# Why

Codex currently negotiates MCP `2025-06-18`, where the client elicitation capability is represented as an empty object. We were still serializing `capabilities.elicitation.form`, which belongs to the later capability shape and can cause strict `2025-06-18` servers to reject `initialize` with an unrecognized-field error.

This keeps the handshake aligned with the protocol version Codex actually negotiates and fixes the compatibility regression tracked in #17492.

# What

- Serialize the client elicitation capability as `elicitation: {}` for `2025-06-18`.
- Keep elicitation advertised for both Codex Apps and custom MCP servers.
- Tighten regression coverage so the unit test asserts both the Rust value and the serialized wire shape.
- Add an app-server integration test that round-trips a form elicitation from a custom MCP server; the existing connector round-trip continues to cover the connector path.

# Verification

- `cargo test -p codex-mcp`
- `cargo test -p codex-app-server mcp_server_elicitation_round_trip`
- `cargo test -p codex-app-server mcp_server_tool_call_round_trips_elicitation`

# Next steps

- Decide whether `tool_call_mcp_elicitation=false` should also suppress capability advertisement during `initialize`.
- Revisit `form` / `url` capability advertisement when Codex is ready to negotiate MCP `2025-11-25`, which defines that newer shape.

### DIFF
--- a/codex-rs/app-server/tests/suite/v2/mcp_tool.rs
+++ b/codex-rs/app-server/tests/suite/v2/mcp_tool.rs
@@ -13,10 +13,16 @@ use axum::Router;
 use codex_app_server_protocol::ItemCompletedNotification;
 use codex_app_server_protocol::JSONRPCError;
 use codex_app_server_protocol::JSONRPCResponse;
+use codex_app_server_protocol::McpElicitationSchema;
+use codex_app_server_protocol::McpServerElicitationAction;
+use codex_app_server_protocol::McpServerElicitationRequest;
+use codex_app_server_protocol::McpServerElicitationRequestParams;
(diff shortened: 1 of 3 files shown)


mzeng-openai:
Can we make sure to test elicitations from connectors and custom MCPs and make sure they don't regress? For context they are supported it's just we don't advertise the mcp version as 2025-11-25

mzeng-openai (approved):

Issue #20489, opened 2026-04-30
Add a screen-reader-friendly Codex TUI mode


taylorarndt:
I know external contributions may not be open right now, but I have a local fix ready if the team wants it or wants to use it as a reference.

VoiceOver users currently hear decorative terminal UI as content while Codex is working. For example, active command/status rows can be announced as repeated spaces, bullets, "running", and box-drawing characters. This interrupts reading code and makes it hard to review terminal output while Codex is still producing updates.

The main pain points:

- Live command/status rows are re-rendered with spinner/bullet/shimmer-style decoration.
- Box drawing such as `│` and `└` is announced as "box drawing".
- When new content arrives, VoiceOver may re-read older completed transcript content, causing the user to lose their place.

Recommended behavior:

- Add `codex --screen-reader` or `codex --accessible`.
- Disable spinner/shimmer/decorative bullets in that mode.
- Render plain indentation instead of box drawing.
- Minimize redraws that cause completed history to be treated as new spoken content.
- Keep the current visual TUI unchanged for users who want it.

I have a branch with a first patch that starts by making the no-animation path quieter for active exec/status rows and removing box drawing from exec output prefixes.

taylorarndt:
I pushed the local reference branch here: https://github.com/taylorarndt/codex/tree/fix/voiceover-screen-reader-tui\n\nVerification run locally:\n\n```text\ncargo test -p codex-tui\n```\n\nThe patch keeps the default visual TUI path unchanged and only makes the existing no-animation path quieter: active exec/status rows omit spinner/bullet decoration, and exec output prefixes use plain indentation instead of box drawing.

etraut-openai:
Thanks for the suggestion. As you are aware, we already have the beginnings of this with the `animations = false` config flag, which you can add to the `[tui]` section of your config file. This disables all spinners, shimmers, and other animations that can interfere with a screen reader. I like the approach you've explored where we build upon this existing feature for incremental improvements.

taylorarndt:
[...]

PR #20564, opened 2026-05-01
Enforce `animations = false` for screen readers

etraut-openai:
## Why

Issue #20489 calls out that animated TUI affordances can be noisy for screen-reader users. Codex already has `tui.animations = false` as a reduced-motion setting, but some live activity rows render spinner-style prefixes in that mode. These were relatively recent regressions.

We have also regressed this pattern more than once by adding new spinner/shimmer callsites that do not think through the reduced-motion path, so this PR adds a small guardrail while fixing the current surfaces.

## What changed

- Omit the live status-row spinner when animations are disabled, so the row starts with stable text like `Working (...)`.
- Render running hook headers without the spinner prefix when animations are disabled, while preserving shimmer/spinner behavior when animations are enabled.
- Centralize TUI activity indicators in `tui/src/motion.rs`, with explicit reduced-motion choices for hidden prefixes, static bullets, and plain shimmer-text fallbacks.
- Route existing spinner/shimmer callsites through the central motion helper, including exec rows, MCP/web-search/loading rows, hook rows, plugin loading, and onboarding loading text.
- Add a source-scan regression test that rejects direct `spinner(...)` or `shimmer_spans(...)` usage outside the central module and primitive definition.
- Add focused coverage that reduced-motion active exec rows are stable, status rows start without a spinner, running hooks omit the spinner, and MCP inventory loading stays stable.
- Update the one affected status-indicator snapshot; the existing detail tree prefix remains unchanged.

## Verification

- `cargo test -p codex-tui`

### DIFF
--- a/codex-rs/tui/src/chatwidget/plugins.rs
+++ b/codex-rs/tui/src/chatwidget/plugins.rs
@@ -16,10 +16,11 @@ use crate::bottom_pane::custom_prompt_view::CustomPromptView;
 use crate::history_cell;
 use crate::key_hint;
 use crate::legacy_core::config::Config;
+use crate::motion::MotionMode;
+use crate::motion::shimmer_text;
 use crate::onboarding::mark_url_hyperlink;
 use crate::render::renderable::ColumnRenderable;
 use crate::render::renderable::Renderable;
-use crate::shimmer::shimmer_spans;
 use crate::tui::FrameRequester;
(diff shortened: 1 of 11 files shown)


fcoury-oai (approved):
Smoke tested a couple of different affected areas, namely the "Working" animations and the activity spinners on the title and they both worked as before with no flags and were omitted/affected correctly when the flag was true.

Code looks good, approved.

PR #20566, opened 2026-05-01
[tool_suggest] More prompt polishes.

mzeng-openai:
Tool suggest still misfires when model needs tool_search, updating the prompts to further disambiguate it:

- [x] rename it from `tool_suggest` to `request_plugin_install`
- [x] rephrase "suggestion" to "install" in the tool descriptions.
- [x] disambiguate "the tool" vs "the plugin/connector". 

Tested with the Codex App and verified it still works.


sayan-oai (approved):
please rename files accordingly

codex-rs/core/src/tools/handlers/tool_suggest.rs
sayan-oai: why's this still called ToolSuggestHandler? whys the file still called tool_suggest.rs?

mzeng-openai: Did a larger sweep for renaming things.

sayan-oai (approved):

PR #20619, opened 2026-05-01
[codex] request desktop attestation from app

jiamingz42:
## Summary

TL;DR: teaches `codex-rs` / app-server to request a desktop-provided attestation token and attach it as `x-oai-attestation` on the scoped ChatGPT Codex request paths.

DeviceCheck attestation interface

## Details

This PR teaches the Codex app-server runtime how to request and attach an attestation token. It does not generate DeviceCheck tokens directly; instead, it relies on the connected desktop app to advertise that it can generate attestation and then asks that app for a fresh header value when needed.

The flow is:

1. The Codex desktop app connects to app-server.
2. During `initialize`, the app can advertise that it supports `requestAttestation`.
3. Before app-server calls selected ChatGPT Codex endpoints, it sends the internal server request `attestation/generate` to the app.
4. app-server receives a pre-encoded header value back.
5. app-server forwards that value as `x-oai-attestation` on the scoped outbound requests.

The code in this repo is mostly protocol and runtime plumbing: it adds the app-server request/response shape, introduces an attestation provider in core, wires that provider into Responses / compaction / realtime setup paths, and covers the intended scoping with tests. The signed macOS DeviceCheck generation remains owned by the desktop app PR.

## Related PR

- Codex desktop app implementation: https://github.com/openai/openai/pull/878649

## Validation

<details>
<summary>Tests run</summary>

```sh
cargo test -p codex-app-server-protocol
cargo test -p codex-core attestation --lib
cargo test -p codex-app-server --lib attestation
```

Also ran:

```sh
just fix -p codex-core
just fix -p codex-app-server
just fix -p codex-app-server-protocol
just fmt
just write-app-server-schema
```

</details>

<details>
<summary>E2E DeviceCheck validation</summary>

First validated the signed desktop app boundary directly: launched a packaged signed `Codex.app`, sent `attestation/generate`, decoded the returned `v1.` attestation header, and validated the extracted DeviceCheck token with `personal/jm/verify_devicecheck_token.py` using bundle ID `com.openai.codex`. Apple returned `status_code: 200` and `is_ok: true`.

Then ran the fuller app + app-server flow. The packaged `Codex.app` launched a current-branch app-server via `CODEX_CLI_PATH`, and a local MITM proxy intercepted outbound `chatgpt.com` traffic. The app-server requested `attestation/generate` from the real Electron app process, and the intercepted `/backend-api/codex/responses` traffic included `x-oai-attestation` on both routes:

```text
GET  /backend-api/codex/responses  Upgrade: websocket  x-oai-attestation: present
POST /backend-api/codex/responses  Upgrade: none       x-oai-attestation: present
```

The captured header decoded to a DeviceCheck token that also validated with Apple for `com.openai.codex` (`status_code: 200`, `is_ok: true`, team `2DC432GLL2`).

</details>


jiamingz42:
rebase modify some unrelated files - fixing it now

jiamingz42:
Addressed in 3cbcba1877: websocket handshake headers now go through the same scoped `AttestationPurpose::Response` path as HTTP `/responses`, and I added a regression test covering `x-oai-attestation` on the websocket handshake. (Reply by Codex)

jiamingz42:
Just did a rebase and resolved some conflicts. Need to double check it does not break anything. (Reply by Codex)

codex-rs/app-server/tests/suite/v2/initialize.rs
```diff
@@ -158,6 +158,7 @@ async fn initialize_opt_out_notification_methods_filters_notifications() -> Resu
             },
             Some(InitializeCapabilities {
                 experimental_api: true,
```
pakrym-oai, line 160: feature needs an integration test.

codex-rs/app-server/src/message_processor.rs
```diff
@@ -153,6 +158,86 @@ impl ExternalAuth for ExternalAuthRefreshBridge {
...
+        let outgoing = outgoing.clone();
+        let attestation_connection_ids = attestation_connection_ids.clone();
+        Box::pin(request_attestation_header_value(
+            outgoing,
+            attestation_connection_ids,
+        ))
+    })
+}
+
+async fn request_attestation_header_value(
```
pakrym-oai: what is the expected perf overhead?

jiamingz42: I ran 100 measured samples after a 10-request warmup. The E2E attestation/generate latency was p90 12.24ms and p95 12.42ms.

codex-rs/app-server/tests/suite/v2/initialize.rs
jiamingz42: 🚧 on it!

codex-rs/core/src/thread_manager.rs
```diff
@@ -316,6 +318,31 @@ impl ThreadManager {
...
+            analytics_events_client,
+            state_db,
+            thread_store,
+            agent_graph_store,
+            /*attestation_provider*/ None,
+        )
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    pub fn new_with_attestation_provider(
```
pakrym-oai: nit: let's avoid yet another overload and add the parameter to existing one

codex-rs/core/src/client.rs
```diff
@@ -305,6 +309,33 @@ impl ModelClient {
...
+            model_verbosity,
+            enable_request_compression,
+            include_timing_metrics,
+            beta_features_header,
+            /*attestation_provider*/ None,
+        )
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) fn new_with_attestation_provider(
```
pakrym-oai: no new ctor

codex-rs/core/src/client.rs
```diff
@@ -914,6 +983,13 @@ impl ModelClientSession {
                     turn_metadata_header.as_ref(),
                 );
                 headers.extend(self.client.build_responses_identity_headers());
+                self.client
+                    .extend_attestation_header_for(
+                        &mut headers,
+                        provider,
+                        AttestationPurpose::Response,
```
pakrym-oai: do we really need the purpose? this will make every internal feature change the protocol

codex-rs/core/src/client.rs
```diff
@@ -1601,6 +1689,43 @@ fn build_responses_headers(
...
+}
+
+fn should_send_attestation(provider: &codex_api::Provider, purpose: AttestationPurpose) -> bool {
+    let provider_is_chatgpt_codex = provider
+        .base_url
+        .trim_end_matches('/')
+        .eq_ignore_ascii_case(CHATGPT_CODEX_BASE_URL);
+    provider_is_chatgpt_codex
+        && matches!(
+            purpose,
```
pakrym-oai: these are the only 3 existing purposes.

codex-rs/core/src/client.rs
```diff
@@ -1601,6 +1689,43 @@ fn build_responses_headers(
...
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum AttestationPurpose {
+    Response,
+    Compaction,
+    RealtimeWebrtcCallSetup,
+}
+
+fn should_send_attestation(provider: &codex_api::Provider, purpose: AttestationPurpose) -> bool {
+    let provider_is_chatgpt_codex = provider
```
pakrym-oai: we use uses_chatgpt_auth in other places.

codex-rs/core/src/attestation.rs
```diff
@@ -0,0 +1,39 @@
...
+type GenerateAttestationFuture = Pin<Box<dyn Future<Output = Option<String>> + Send>>;
+type GenerateAttestationCallback = dyn Fn() -> GenerateAttestationFuture + Send + Sync + 'static;
+
+/// Session-scoped source for just-in-time attestation header values.
+///
+/// Host integrations provide the opaque string expected by the upstream
+/// `x-oai-attestation` header. Core validates only that it is legal as an HTTP
+/// header value before forwarding it.
+#[derive(Clone)]
+pub struct AttestationProvider {
```
pakrym-oai: Can attestation provider encapsulate all logic around generating the value (including filtering based on auth type)

codex-rs/core/src/client_tests.rs
```diff
@@ -464,3 +485,218 @@ fn auth_request_telemetry_context_tracks_attached_auth_and_retry_phase() {
...
+        headers
+            .get(crate::attestation::X_OAI_ATTESTATION_HEADER)
+            .and_then(|value| value.to_str().ok()),
+        Some("v1.header-1"),
+    );
+    assert_eq!(attestation_calls.load(Ordering::Relaxed), 1);
+}
+
+#[tokio::test]
+async fn compact_generate_fresh_attestation_headers_for_chatgpt_codex() {
```
pakrym-oai: these probably should just be integration tests

codex-rs/core/src/attestation.rs
pakrym-oai: Better yet, leave AttestationProvider just a trait in code and move the entire implemenation to app-server

codex-rs/app-server/src/message_processor.rs
```diff
@@ -83,7 +86,9 @@ use tokio::sync::watch;
 use tokio::time::Duration;
 use tokio::time::timeout;
 use tracing::Instrument;
+use tracing::warn;
 
+const ATTESTATION_GENERATE_TIMEOUT: Duration = Duration::from_secs(5);
```
pakrym-oai: this is too high, 100ms maybe?

codex-rs/app-server/src/message_processor.rs
```diff
@@ -153,6 +158,86 @@ impl ExternalAuth for ExternalAuthRefreshBridge {
...
+            outgoing,
+            attestation_connection_ids,
+        ))
+    })
+}
+
+async fn request_attestation_header_value(
+    outgoing: Arc<OutgoingMessageSender>,
+    attestation_connection_ids: Arc<Mutex<HashSet<ConnectionId>>>,
+) -> Option<String> {
```
pakrym-oai: nit skip this extra method

codex-rs/app-server/src/message_processor.rs
```diff
@@ -733,6 +836,12 @@ impl MessageProcessor {
                     .connection_initialized(connection_id)
                     .await;
             }
+            if session.request_attestation() {
```
pakrym-oai: can we somehow put attestation onto live_connections in thread_processor? So we don't have multiple places to track connections.

codex-rs/app-server/src/message_processor.rs
```diff
@@ -154,6 +159,75 @@ impl ExternalAuth for ExternalAuthRefreshBridge {
...
+
+    let result = match timeout(timeout_duration, rx).await {
+        Ok(Ok(Ok(result))) => result,
+        Ok(Ok(Err(err))) => {
+            warn!(
+                code = err.code,
+                message = %err.message,
+                "attestation generation request failed"
+            );
+            return Some(String::new());
```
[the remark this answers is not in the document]
jiamingz42: Update the transport protocol. This comment become obsolete.

codex-rs/app-server/src/thread_state.rs
```diff
@@ -214,12 +219,28 @@ impl ThreadStateManager {
...
         self.state
             .lock()
             .await
             .live_connections
-            .insert(connection_id);
+            .iter()
+            .filter_map(|(connection_id, capabilities)| {
+                capabilities.request_attestation.then_some(*connection_id)
+            })
+            .min_by_key(|connection_id| connection_id.0)
```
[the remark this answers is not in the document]
jiamingz42: image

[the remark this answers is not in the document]
jiamingz42: Makes sense. Codex suggests asking whichever client initiates the request. It probably doesn't matter for now since it's mostly a one-to-one relationship, but let's implement it in case that assumption changes.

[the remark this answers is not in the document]
jiamingz42: Try two approaches. Implementing the strong guarantee seems a bit over-engineering. 

I simplified this to only ask attestation-capable clients attached to the same thread. That still avoids using a token from an unrelated thread; the remaining limitation is that if multiple clients are attached to the same thread, we may not pick the exact initiator. The benefit is that the implementation is much simpler and stays local to app-server.

codex-rs/app-server-protocol/src/protocol/v2/attestation.rs
```diff
@@ -0,0 +1,17 @@
...
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct AttestationGenerateParams {}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct AttestationGenerateResponse {
+    /// Opaque client attestation payload to embed in the upstream header envelope.
+    pub header_value: String,
```
pakrym-oai: maybe call this just token? we can decide how to send it? don't feel strong about this.

codex-rs/core/src/client.rs
```diff
@@ -314,6 +318,7 @@ impl ModelClient {
         enable_request_compression: bool,
         include_timing_metrics: bool,
         beta_features_header: Option<String>,
+        attestation_provider: Option<Arc<dyn AttestationProvider>>,
     ) -> Self {
         let model_provider = create_model_provider(provider_info, auth_manager);
```
pakrym-oai, line 324: I can see the akwardness in trying to add uses_chatgpt_auth onto ApiProvider.

What if we put the logic onto ModelProvider trait as `supports_attestation`. You have an instance of provider here and you can store the `include_attestation` onto ModelClientState.

codex-rs/core/src/client.rs
```diff
@@ -1626,6 +1671,18 @@ fn build_responses_headers(
     headers
 }
 
+impl ModelClient {
+    async fn extend_attestation_header_for(
```
pakrym-oai: nit, do we need this tiny method in additoin to generate_attestation_header_for?

pakrym-oai (commented):
Let's try and drop uses_chatgpt_auth from ApiProvider. ApiProvider is an abstraction that should know about details of the provider implementation.

codex-rs/core/src/client.rs
jiamingz42: Addressed in 0712d3b921: attestation eligibility now lives on `ModelProvider::supports_attestation()`, and `ModelClientState` stores the resulting `include_attestation` flag. That lets `ApiProvider` stay focused on request transport details instead of carrying provider-auth semantics. (Reply by Codex)

euroelessar (approved):

codex-rs/core/src/client.rs
```diff
@@ -920,8 +955,9 @@ impl ModelClientSession {
     ///
     /// Keeping option construction in one place ensures request-scoped headers are consistent
     /// regardless of transport choice.
-    fn build_responses_options(
+    async fn build_responses_options(
         &self,
+        _provider: &codex_api::Provider,
```
pakrym-oai: nit: rem

pakrym-oai (approved):

jiamingz42:
🙏  @euroelessar @pakrym-oai it's green now. taking a last pass and then merging it.

PR #20638, opened 2026-05-01
[codex] Record model-input image metadata on turn traces

charley-openai:
## Why
Make slow-turn trace triage show whether images were included in model inference input, including screenshots returned by tools such as Browser Use.

[Example Trace](https://openai.datadoghq.com/apm/trace/4342321495833984941?graphType=waterfall&shouldShowLegend=true&spanID=6182567473583795561&traceQuery=)
Screenshot 2026-05-05 at 6 16 03 PM

## Summary
- record trace-safe model-input image metadata on `session_task.turn`
- keep `codex.user_prompt` limited to user-submitted prompt counts
- accumulate turn-level image evidence across sampling passes so later retries do not clear earlier image signal
- include aggregate image/mime types plus sanitized per-image source, type, byte length, and extension when available
- avoid recording local paths, remote URLs, or base64 payloads

## Validation
- `just fmt`
- `cargo test -p codex-otel`
- `cargo test -p codex-core --test all otel`
- `just fix -p codex-otel`
- `just fix -p codex-core`

### DIFF
--- a/codex-rs/core/src/session/turn.rs
+++ b/codex-rs/core/src/session/turn.rs
@@ -440,6 +440,11 @@ pub(crate) async fn run_turn(
                 .await
                 .for_prompt(&turn_context.model_info.input_modalities)
         };
+        turn_context.session_telemetry.record_model_input_images(
+            &turn_context.turn_span(),
+            &turn_context.model_input_images,
+            &sampling_request_input,
+        );
 
         let sampling_request_input_messages = sampling_request_input
             .iter()
--- a/codex-rs/otel/src/events/image_input_telemetry.rs
+++ b/codex-rs/otel/src/events/image_input_telemetry.rs
@@ -1,0 +1,283 @@
+use codex_protocol::user_input::UserInput;
+use serde::Serialize;
+use std::collections::BTreeSet;
+use std::io::Read;
+use std::path::Path;
+
+const IMAGE_HEADER_READ_LIMIT: u64 = 64 * 1024;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct ImageInputTelemetry {
+    pub(crate) details_json: String,
+    pub(crate) image_types: String,
+    pub(crate) mime_types: String,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+struct ImageInputDetail {
+    source: &'static str,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    image_type: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    mime_type: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    width: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    height: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    byte_length: Option<u64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    extension: Option<String>,
+}
+
+pub(crate) fn image_input_telemetry(items: &[UserInput]) -> Option<ImageInputTelemetry> {
+    let details: Vec<ImageInputDetail> = items
+        .iter()
+        .filter_map(|item| match item {
+            UserInput::Image { image_url } => Some(remote_or_data_url_detail(image_url)),
+            UserInput::LocalImage { path } => Some(local_image_detail(path)),
+            UserInput::Text { .. } | UserInput::Skill { .. } | UserInput::Mention { .. } => None,
+            _ => None,
+        })
+        .collect();
+
+    if details.is_empty() {
+        return None;
+    }
+
+    let image_types = comma_join(
+        details
+            .iter()
+            .filter_map(|detail| detail.image_type.as_deref()),
+    );
+    let mime_types = comma_join(
+        details
+            .iter()
+            .filter_map(|detail| detail.mime_type.as_deref()),
+    );
+    let details_json = serde_json::to_string(&details).ok()?;
+
+    Some(ImageInputTelemetry {
+        details_json,
+        image_types,
+        mime_types,
+    })
+}
+
+fn remote_or_data_url_detail(image_url: &str) -> ImageInputDetail {
+    if let Some(detail) = data_url_detail(image_url) {
+        return detail;
+    }
+
+    let extension = safe_image_extension_from_url(image_url);
+    let mime_type = extension.as_deref().and_then(mime_type_from_extension);
+    ImageInputDetail {
+        source: "remote_url",
+        image_type: mime_type.as_deref().and_then(image_type_from_mime),
+        mime_type,
+        width: None,
+        height: None,
+        byte_length: None,
+        extension,
+    }
+}
+
+fn data_url_detail(image_url: &str) -> Option<ImageInputDetail> {
+    let image_url = strip_data_scheme(image_url)?;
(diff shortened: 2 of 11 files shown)


codex-rs/otel/tests/suite/otel_export_routing_policy.rs
```diff
@@ -187,17 +264,87 @@ fn otel_export_routing_policy_routes_user_prompt_log_and_trace_events() {
...
-        Some("1")
+        Some("2")
     );
     assert_eq!(
         prompt_trace_attrs
             .get("local_image_input_count")
             .map(String::as_str),
         Some("1")
     );
+    assert!(!prompt_trace_attrs.contains_key("image_input_types"));
```
jif-oai, line 277: I think this is the wrong surface for the stated goal. The PR says we are enriching codex.user_prompt, but this test now locks in that the new metadata is not on that event
Am I missing something?

codex-rs/core/src/session/turn.rs
```diff
@@ -433,6 +433,9 @@ pub(crate) async fn run_turn(
                 .await
                 .for_prompt(&turn_context.model_info.input_modalities)
         };
+        turn_context
+            .session_telemetry
+            .record_model_input_images(&turn_context.turn_span(), &sampling_request_input);
```
jif-oai: This overwrites the turn-span fields on every sampling pass. If the slow first request had an image and a later retry/follow-up has none, the final span ends up saying 0, so the trace hides exactly the case we are trying to diagnose. Can we emit this per request or accumulate turn-level evidence instead?

codex-rs/otel/src/events/image_input_telemetry.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    let extension = safe_image_extension_from_url(image_url);
+    let mime_type = extension.as_deref().and_then(mime_type_from_extension);
+    ImageInputDetail {
+        source,
+        image_type: mime_type.as_deref().and_then(image_type_from_mime),
+        mime_type,
+        width: None,
+        height: None,
```
jif-oai: Both code path set it to `None`?  So what's the point?

codex-rs/otel/src/events/image_input_telemetry.rs
```diff
@@ -0,0 +1,212 @@
...
+    let mut details: Vec<ImageInputDetail> = Vec::new();
+    let mut user_message_image_count = 0;
+    let mut tool_output_image_count = 0;
+
+    for item in items {
+        match item {
+            ResponseItem::Message { content, .. } => {
+                for content_item in content {
+                    if let ContentItem::InputImage { image_url, .. } = content_item {
+                        user_message_image_count += 1;
```
jif-oai: How can you be sure it's a user message without checking the role? What about image generation for example? Or maybe citation one day or whatever?

codex-rs/otel/src/events/image_input_telemetry.rs
```diff
@@ -0,0 +1,212 @@
...
+            .iter()
+            .filter_map(|detail| detail.image_type.as_deref()),
+    );
+    let mime_types = collect_unique_strings(
+        details
+            .iter()
+            .filter_map(|detail| detail.mime_type.as_deref()),
+    );
+    let details_json = if details.is_empty() {
+        String::new()
```
jif-oai: This is not a valid JSON. Can we have an empty JSON instead?

codex-rs/core/src/session/turn.rs
charley-openai: nice catch, you're right!

codex-rs/otel/src/events/image_input_telemetry.rs
charley-openai: agreed, just ripping it out.

codex-rs/otel/tests/suite/otel_export_routing_policy.rs
charley-openai: Ugh, the title & desc were outdated. The intention was actually to include the logged data in the session_task.turn spans so that we'd be able to discern relative performance with/without image data.

codex-rs/otel/src/events/image_input_telemetry.rs
charley-openai: yep!

codex-rs/otel/src/events/image_input_telemetry.rs
charley-openai: you're right, changing it to just be inputted images, role doesn't actually matter as much for what we're trying to find.

codex-rs/core/src/session/turn.rs
```diff
@@ -440,6 +440,11 @@ pub(crate) async fn run_turn(
                 .await
                 .for_prompt(&turn_context.model_info.input_modalities)
         };
+        turn_context.session_telemetry.record_model_input_images(
```
jif-oai, line 443: I think this is too low in the stack for the stated goal. Inline compaction also sends model requests under the same turn, and those prompts can still contain images, but this only records the normal sampling loop. So a turn can spend its slow inference on an image-heavy compaction request and still end up tagged as having no images here.

codex-rs/otel/src/events/image_input_telemetry.rs
```diff
@@ -0,0 +1,250 @@
...
+    details: Vec<ImageInputDetail>,
+    image_types: BTreeSet<String>,
+    mime_types: BTreeSet<String>,
+    message_image_count: usize,
+    tool_output_image_count: usize,
+}
+
+impl ImageInputTelemetryState {
+    fn accumulate(&mut self, request: Self) {
+        self.message_image_count = self.message_image_count.max(request.message_image_count);
```
jif-oai, line 47: These maxima can come from different requests, then we add them back together as if they happened at once. After compaction or invalid-image sanitization, one pass can have 2 message images and another can have 1 tool image, and the turn span reports 3 even though no model request ever sent 3...

codex-rs/otel/src/events/image_input_telemetry.rs
```diff
@@ -0,0 +1,250 @@
...
+        self.tool_output_image_count = self
+            .tool_output_image_count
+            .max(request.tool_output_image_count);
+        self.image_types.extend(request.image_types);
+        self.mime_types.extend(request.mime_types);
```
jif-oai, line 52: This can make the fields disagree with each other. If one request has `[jpeg message]` and a later request has `[png tool_output]`, the aggregate types say both existed, but details keeps only the first list because the lengths tie... (not sure if plausible but I think)

codex-rs/otel/src/events/image_input_telemetry.rs
```diff
@@ -0,0 +1,250 @@
...
+        if request.details.len() > self.details.len() {
+            self.details = request.details;
+        }
+    }
+
+    fn snapshot(&self) -> ImageInputTelemetrySnapshot {
+        let details_json =
+            serde_json::to_string(&self.details).unwrap_or_else(|_| "[]".to_string());
```
jif-oai, line 60: Still unbounded

jif-oai (commented):
I'm not sure yet this is the good layer. The The thing we actually want to explain is a model request, but this PR stores mutable turn-level state and reconstructs a synthetic summary after the fact

I think I’d rather put the first-class telemetry on the actual model-request boundary, then optionally derive a coarse turn summary from that, than make TurnContext carry a lossy accumulator whose meaning is hard to reason about

charley-openai:
closing for now while i'm out on pat leave.

PR #20644, opened 2026-05-01
Fix elevated Windows PTY teardown after shell exit

iceweasel-oai:
## Why
Windows background terminals can outlive the shell process that started them because the elevated command runner keeps the ConPTY input pipe open while it waits for the parent IPC loop. When that happens, `conhost.exe` can keep the pseudoconsole alive even after the shell exits, the output reader never reaches EOF, and the runner never sends its final `Message::Exit`. From unified exec's point of view the PTY session is still running, so the background terminals UI grows and never shrinks.

## What Changed
- share the PTY stdin handle between the input thread and the main teardown path in `windows-sandbox-rs/src/elevated/command_runner_win.rs`
- close PTY stdin as soon as the child shell exits, before closing the pseudoconsole and waiting for stdout drain
- document the handle ownership and teardown ordering inline where the lifetime is managed

## Verification
- `cargo build -p codex-windows-sandbox --bin codex-command-runner`
- reproduced in the desktop harness by starting short-lived PTY sessions and confirming they now report normal exit instead of staying alive forever
- started longer-lived PTY sessions in the desktop harness and verified they appear while running and disappear after they finish

### DIFF
--- a/codex-rs/windows-sandbox-rs/src/elevated/command_runner_win.rs
+++ b/codex-rs/windows-sandbox-rs/src/elevated/command_runner_win.rs
@@ -360,13 +360,16 @@ fn spawn_output_reader(
 /// Read stdin/terminate frames and forward to the child process.
 fn spawn_input_loop(
     mut reader: File,
-    stdin_handle: Option<HANDLE>,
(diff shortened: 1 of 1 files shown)


iceweasel-oai:
@codex review

PR #20649, opened 2026-05-01
[connectors] Track host-owned Codex Apps MCP provenance

mzeng-openai:
## Why

MCP server names are routing labels, not a trustworthy ownership boundary. Connector-gateway behavior should only apply to the built-in Codex Apps MCP instance created by the host, not to an MCP server that merely has the same name or declares the same capabilities.

## How provenance is determined

`McpServerProvenance` is host-assigned and runtime-only. MCP servers loaded from user/plugin config deserialize as `UserConfigured`; the only path that marks a server `HostOwnedCodexApps` is the host path that injects Codex Apps MCP at `codex_apps` using the ChatGPT backend `/wham/apps` endpoint. Because provenance is not serialized, cached Codex Apps tools are restamped as host-owned when loaded.

## What changed

- Add runtime-only `McpServerProvenance` and mark the host-injected Apps MCP as `HostOwnedCodexApps`.
- Carry that provenance through managed MCP clients and emitted `ToolInfo`.
- Switch connector-gateway behavior from raw server-name checks to provenance checks across auth/cache setup, connector discovery, tool exposure, and MCP tool-call approval handling.
- Add regression coverage proving that a spoofed `codex_apps` server name is not treated as host-owned.

## Testing

- `cargo test -p codex-mcp`
- `cargo test -p codex-core mcp_tool_call`


viyatb-oai:
Flagging one security concern from Viyat's security review automation.

The cached Codex Apps tool path appears to drop host-owned provenance. `ToolInfo.server_provenance` is skipped during serde, so when cached tools are loaded they come back with the default `UserConfigured` provenance. `load_cached_codex_apps_tools` then returns those tools without re-stamping them as `HostOwnedCodexApps`.

That matters because the exposure filtering treats host-owned Codex Apps tools differently: connector accessibility/enabled checks only apply when `is_host_owned_codex_apps()` is true. A cached app tool can therefore be model/tool-search visible as a normal MCP tool even when its connector should be disabled or inaccessible. The actual tool call path may recover once the live server is connected, but the model-visible surface is already wider than intended.

The fix I'd suggest: after deserializing the cache, set every cached Codex Apps `ToolInfo.server_provenance = McpServerProvenance::HostOwnedCodexApps` before running exposure filtering. Longer term, consider a cache DTO that does not serialize `ToolInfo` directly and converts back into `ToolInfo` with explicit provenance. A small regression should load a cached app tool for a disabled/inaccessible connector and assert `build_mcp_tool_exposure` does not expose it.

mzeng-openai:
Good catch. Cached Codex Apps tools now get restamped as `HostOwnedCodexApps` at the cache boundary, and I added a regression for the startup-cache load. The current manager path was already stamping before exposure, but this makes the invariant local so future direct cache reads cannot miss it.

codex-rs/codex-mcp/src/mcp/mod.rs
bolinfest: This appears to read `config.chatgpt_base_url` and while it also goes through checks in `codex_apps_mcp_url_for_base_url`, what is not clear to me is whether someone could inject a malicious value of `chatgpt_base_url` into `config.toml` and get stamped with `McpServerProvenance::HostOwnedCodexApps` here.

I think this code should be written to make this easier to audit.

For example, one thing that might help is to return a type other than `String` where it is forced to go through a constructor that verifies that the `url` is "safe" for our purposes.

We might want to consider having such a function return this "url wrapper type" and the `McpServerProvenance` as fields of a single struct as further proof they were derived in a way that ensures they are in sync with one another.

codex-rs/codex-mcp/src/mcp/mod_tests.rs
```diff
@@ -228,6 +228,16 @@ fn codex_apps_server_config_uses_legacy_codex_apps_path() {
     assert_eq!(url, "https://chatgpt.com/backend-api/wham/apps");
 }
 
+#[test]
+fn codex_apps_server_config_is_marked_host_owned() {
+    let config = test_mcp_config(PathBuf::from("/tmp"));
```
bolinfest: Please do not use `/tmp` because that is not portable to Windows: prefer the `TempDir` abstraction.

codex-rs/codex-mcp/src/mcp/mod.rs
mzeng-openai: Done. The host-owned path now constructs a private `TrustedCodexAppsMcpUrl` and pairs it with `HostOwnedCodexApps` in one endpoint constructor. That constructor validates `chatgpt_base_url` before provenance is granted, with tests for spoofed hosts, non-HTTPS first-party URLs, credentials/query/fragment, and non-default ports.

codex-rs/codex-mcp/src/mcp/mod_tests.rs
mzeng-openai: Done, switched those tests to `tempfile::tempdir()`.

codex-rs/codex-mcp/src/mcp/mod.rs
```diff
@@ -199,10 +203,15 @@ pub fn with_codex_apps_mcp(
     config: &McpConfig,
 ) -> HashMap<String, McpServerConfig> {
     if config.apps_enabled && auth.is_some_and(CodexAuth::uses_codex_backend) {
-        servers.insert(
-            CODEX_APPS_MCP_SERVER_NAME.to_string(),
-            codex_apps_mcp_server_config(config),
-        );
+        match codex_apps_mcp_server_config(config) {
+            Ok(server_config) => {
+                servers.insert(CODEX_APPS_MCP_SERVER_NAME.to_string(), server_config);
```
bolinfest: Can/do we want to impose any restrictions on MCP server names read from `config.toml` so that "internal" MCP servers cannot run the risk of name collisions with user-defined ones?

You could also use https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.entry and ensure your `insert()` is not and overwrite and if so error?

codex-rs/codex-mcp/src/mcp/mod.rs
```diff
@@ -347,11 +356,9 @@ pub async fn collect_mcp_snapshot_from_manager(
     .await
 }
 
-pub(crate) fn codex_apps_mcp_url(config: &McpConfig) -> String {
-    codex_apps_mcp_url_for_base_url(
-        &config.chatgpt_base_url,
-        config.apps_mcp_path_override.as_deref(),
-    )
+#[cfg(test)]
```
bolinfest: Should this get moved to `mod_tests.rs` instead?

codex-rs/codex-mcp/src/mcp/mod.rs
```diff
@@ -383,19 +390,47 @@ fn codex_apps_mcp_bearer_token_env_var() -> Option<String> {
     }
 }
 
-fn normalize_codex_apps_base_url(base_url: &str) -> String {
-    let mut base_url = base_url.trim_end_matches('/').to_string();
-    if (base_url.starts_with("https://chatgpt.com")
-        || base_url.starts_with("https://chat.openai.com"))
-        && !base_url.contains("/backend-api")
+#[derive(Debug, Clone, PartialEq, Eq)]
```
bolinfest: This is adding a lot of code to a file that is already fairly large: I would put this in a separate file (and move any tests to a parallel file, as appropriate).

codex-rs/codex-mcp/src/mcp/mod.rs
```diff
@@ -383,19 +390,47 @@ fn codex_apps_mcp_bearer_token_env_var() -> Option<String> {
...
+struct HostOwnedCodexAppsMcpEndpoint {
+    url: TrustedCodexAppsMcpUrl,
+    provenance: McpServerProvenance,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct TrustedCodexAppsMcpUrl(String);
```
bolinfest: You really want something like this to ensure the integrity of the `String` value:

```rust
mod trusted_codex_apps_mcp_url {
    #[derive(Debug, Clone, PartialEq, Eq)]
    pub struct TrustedCodexAppsMcpUrl(String);

    impl TrustedCodexAppsMcpUrl {
        pub fn new(url: String) -> Option<Self> {
            if is_trusted_codex_apps_mcp_url(&url) {
                Some(Self(url))
            } else {
                None
            }
        }

        pub fn as_str(&self) -> &str {
            &self.0
        }
    }

    fn is_trusted_codex_apps_mcp_url(url: &str) -> bool {
        // your validation logic here
        url.starts_with("https://")
    }
}

pub use trusted_codex_apps_mcp_url::TrustedCodexAppsMcpUrl;
```

All the more reason to move this to a separate file to restrict the public API!

codex-rs/codex-mcp/src/mcp/mod.rs
mzeng-openai: Good point. I now treat `codex_apps` as reserved at the config boundary, so any user-configured server with that name is dropped before we add the host-owned endpoint rather than being overwritten later.

PR #20653, opened 2026-05-01
[codex] report running command durations

nornagon-openai:
## Summary

- add an internal `started_at_ms` to `ExecCommandBeginEvent` and populate it from real execution paths
- report `durationMs: 0` when command execution starts
- let `ThreadHistoryBuilder` derive elapsed command durations when it materializes live snapshots/history

## Why

Running commands only exposed a final duration today. A second client joining a live thread therefore had no server-authored elapsed runtime to render, which pushed clients toward local guesses that reset when UI state remounted.

The core event now carries the execution start time as the source of truth, while app-server still exposes elapsed milliseconds on the wire. That keeps remote clients clock-skew-safe and lets resumed clients render an accurate-enough runtime immediately, without a separate app-server refresh pass over already-projected turns.

## Validation

- `just fmt`
- `cargo test -p codex-protocol`
- `cargo test -p codex-app-server --lib`
- `cargo test -p codex-app-server` *(fails in unrelated `skills_changed_notification_is_emitted_after_skill_change`; the test environment currently times out after a websocket `401 Unauthorized`)*
- `cargo test -p codex-tui --lib` *(fails in unrelated sandbox-sensitive status snapshot expectations in this checkout)*
- `just fix -p codex-protocol -p codex-core -p codex-app-server`
- `cargo test` *(workspace run reached and passed app-server/app-server-protocol coverage, then failed in unrelated `codex-core` tests; many `js_repl` failures are due to local Node `v22.15.1` while the suite requires `>= v22.22.0`, plus existing guardian/config expectation failures)*


codex-rs/app-server/README.md
```diff
@@ -1140,7 +1140,7 @@ Today both notifications carry an empty `items` array even when item events were
 - `agentMessage` — `{id, text}` containing the accumulated agent reply.
 - `plan` — `{id, text}` emitted for plan-mode turns; plan text can stream via `item/plan/delta` (experimental).
 - `reasoning` — `{id, summary, content}` where `summary` holds streamed reasoning summaries (applicable for most OpenAI models) and `content` holds raw reasoning blocks (applicable for e.g. open source models).
-- `commandExecution` — `{id, command, cwd, status, commandActions, aggregatedOutput?, exitCode?, durationMs?}` for sandboxed commands; `status` is `inProgress`, `completed`, `failed`, or `declined`.
+- `commandExecution` — `{id, command, cwd, status, commandActions, aggregatedOutput?, exitCode?, durationMs?}` for sandboxed commands; `status` is `inProgress`, `completed`, `failed`, or `declined`. `durationMs` is elapsed runtime while the command is running and the final duration once it completes.
```
owenlin0, line 1143: hmmm, should we just have a `execution_started_at` timestamp instead, and then populate `duration_ms` once it's complete? feels better to me than a duration that increments and is "materialized" when you receive the commandExecution item

PR #20658, opened 2026-05-01
[codex-exec] Enforce add-dir for macOS sandbox

evawong-oai:
## Why

This PR is a user migration fix for the legacy exec sandbox path during the `PermissionProfile` migration.

The failing path was:

```text
codex exec
  --sandbox workspace-write
  --add-dir <extra>
  -C <workspace>
```

Then the model shell tool ran from the added root and tried:

```text
printf ok > normal.txt
```

macOS Seatbelt returned operation not permitted.

Legacy workspace write added roots are dropped only when `codex exec` starts or resumes an app server thread without an active permission profile.

That no active profile state is intentional for this legacy path. The local `Config` has the resolved effective permissions, including the added root, but there is no named profile that can be reselected without losing legacy root details. Before this PR, `codex exec` sent only `sandbox = workspace-write` into `thread/start` and `thread/resume`. App server rebuilt workspace write permissions from that lossy shape, so the added root was dropped.

## Change Details

1. Add `projectRoots` request fields to `thread/start` and `thread/resume`.
2. Store `project_roots` on `Config` as the final cwd plus the original `--add-dir` values resolved against that cwd.
3. Have `codex exec` forward `config.project_roots` when starting or resuming the app server thread.
4. Have app server translate those concrete roots into the existing `additional_writable_roots` override before config builds the thread permissions.
5. Keep the legacy no active profile path on the legacy sandbox selection. This PR does not synthesize a built in permission profile for that path.

## Validation

Current pushed head is `351d4995ef671c4ef1169bac7c2dc21b097c94b8`.

Current head validation:

1. `just fmt`
2. `cargo check -p codex-thread-manager-sample`
3. `cargo test -p codex-core add_dir_override_extends_workspace_writable_roots -- --nocapture`
4. `cargo test -p codex-exec thread_lifecycle_params_include_legacy_sandbox_when_no_active_profile -- --nocapture`
5. `git diff --check`

Earlier validation on this PR before the Config ownership cleanup:

1. `cargo test -p codex-app-server-protocol -- --nocapture`
2. `cargo test -p codex-app-server --no-run`

AWS macOS VM `i-01c31233406f6b2c4`, macOS `26.4.1` build `25E253`, validated current head `351d4995ef671c4ef1169bac7c2dc21b097c94b8`:

1. Direct Seatbelt metadata tier: `SUMMARY preflight_pass=6 pass=55 fail=0 skip=3 unsupported=0 invalid=0 root=/Users/ec2-user/codex-bugb15632-validation-20260503/macos-seatbelt-cases.jv6Xeg head=351d4995ef671c4ef1169bac7c2dc21b097c94b8 backend=macos-seatbelt`
2. `just c exec` Seatbelt E2E tier: `SUMMARY preflight_pass=5 pass=8 fail=0 skip=0 invalid=0 root=/Users/ec2-user/codex-bugb15632-validation-20260503/just-c-exec-e2e-cases.MeIPIc head=351d4995ef671c4ef1169bac7c2dc21b097c94b8 backend=just-c-exec-macos-seatbelt`


codex-rs/exec/src/lib.rs
```diff
@@ -1016,10 +1020,86 @@ fn sandbox_mode_from_permission_profile(
 }
 
 fn config_request_overrides_from_config(config: &Config) -> Option<HashMap<String, Value>> {
-    config
-        .active_profile
-        .as_ref()
-        .map(|profile| HashMap::from([("profile".to_string(), Value::String(profile.clone()))]))
+    let mut overrides = HashMap::new();
+    if let Some(profile) = config.active_profile.as_ref() {
+        overrides.insert("profile".to_string(), Value::String(profile.clone()));
```
bolinfest: FYI, we are going to be changing how profiles work in https://github.com/openai/codex/pull/17141.

codex-rs/exec/src/lib.rs
```diff
@@ -1016,10 +1020,86 @@ fn sandbox_mode_from_permission_profile(
...
+
+    overrides.insert(
+        "sandbox_workspace_write".to_string(),
+        json!({
+            "writable_roots": writable_roots,
+            "network_access": network_access,
+            "exclude_tmpdir_env_var": exclude_tmpdir_env_var,
+            "exclude_slash_tmp": exclude_slash_tmp,
+        }),
+    );
```
bolinfest: I feel like it would be ideal if, instead of propagating the use of the legacy `sandbox_workspace_write`, we injected the appropriate equivalent to something using the `[permissions]` shape here.

Or, perhaps more appropriately, addressed this at the `ConfigBuilder` level.

Note `config_request_overrides_from_config()` was historically created to support a `--profile` flag, which as I noted, we are trying to redo in a very different way. As such, I would prefer to avoid expanding the role of this function.

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -3589,6 +3589,11 @@ pub struct ThreadStartParams {
     pub approvals_reviewer: Option<ApprovalsReviewer>,
     #[ts(optional = nullable)]
     pub sandbox: Option<SandboxMode>,
+    /// Fully resolved runtime permissions for this thread. Cannot be combined
+    /// with `sandbox` or named `permissions`.
+    #[experimental("thread/start.permissionProfile")]
+    #[ts(optional = nullable)]
+    pub permission_profile: Option<PermissionProfile>,
```
bolinfest: As I noted in a comment on https://github.com/openai/codex/pull/15977, I do not think we should be exposing this via the app server API: we should only expose profile names. Clients should not be able to specify arbitrary profile values.

bolinfest:
I think https://github.com/openai/codex/pull/22610 made it possible to solve this in a cleaner way by supporting `workspace_roots` for profiles and `runtime_workspace_roots` for threads.

evawong-oai:
Closing this because latest main now covers the macOS validation slice this PR was intended to fix.

Validated on the AWS macOS VM against main at 249d50aafc552d2fdcd3836a50b03ce366d9744f. The Seatbelt harness passed with preflight_pass=5 pass=8 fail=0 skip=0 invalid=0. The key cases passed: normal write in added root, protected .git in added root, and protected metadata read but not write.

PR #20664, opened 2026-05-01
Add stdio exec-server client transport

starr-openai:
## Why

Configured environments need to connect to exec-server instances that are not necessarily already listening on a websocket URL. A command-backed stdio transport lets Codex start an exec-server process, speak JSON-RPC over its stdio streams, and clean up that child process with the client lifetime.

**Stack position:** this is PR 2 of 5. It builds on the server-side stdio listener from PR 1 and provides the client transport used by later environment/config PRs.

## What Changed

- Add `ExecServerTransport` variants for websocket URLs and stdio shell commands.
- Add stdio command connection support for `ExecServerClient`.
- Move websocket/stdio transport setup into `client_transport.rs` so `client.rs` stays focused on shared JSON-RPC client, session, HTTP, and notification behavior.
- Tie stdio child process cleanup to the JSON-RPC connection lifetime with a RAII lifetime guard.
- Keep existing websocket environment behavior by adapting URL-backed remotes to `ExecServerTransport::WebSocketUrl`.

## Stack

- **2. This PR:** https://github.com/openai/codex/pull/20664 - Add stdio exec-server client transport

## Validation

Not run locally; this was split out of the original draft stack and then refactored to separate transport setup from the base client.


codex-rs/exec-server/src/client_api.rs
```diff
@@ -25,6 +25,22 @@ pub struct RemoteExecServerConnectArgs {
     pub resume_session_id: Option<String>,
 }
 
+/// Stdio connection arguments for a command-backed exec-server.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct StdioExecServerConnectArgs {
```
jif-oai: A single shell string means callers have to hand-roll quoting, and we lose any way to control argv/env/cwd... is it on purpose?

codex-rs/exec-server/src/client_transport.rs
```diff
@@ -0,0 +1,226 @@
...
+    {
+        let mut command = Command::new("cmd");
+        command.arg("/C").arg(shell_command);
+        command
+    }
+
+    #[cfg(not(windows))]
+    {
+        let mut command = Command::new("sh");
+        command.arg("-lc").arg(shell_command);
```
jif-oai: Are we sure we want a login shell here? For example, what if the `.profile` has any kind of `echo "hello world"` (as lots of people do)
Then the stdout become 
```
hello world
{"id":1,"result":{"sessionId":"..."}}
```
and so we break the JSON-RPC

And I think this problem might become true in multiple other cases...

codex-rs/exec-server/src/client_transport.rs
```diff
@@ -0,0 +1,226 @@
...
+                    }
+                }
+            });
+        }
+
+        Self::connect(
+            JsonRpcConnection::from_stdio(
+                stdout,
+                stdin,
+                format!("exec-server stdio command `{shell_command}`"),
```
jif-oai: Can we avoid putting the raw command in the connection label? Any token/env/... assignment in the command now gets echoed back through transport errors/logs

codex-rs/exec-server/src/client_transport.rs
```diff
@@ -0,0 +1,226 @@
...
+#[cfg(unix)]
+fn terminate_stdio_child_process(process_group_id: Option<u32>, child: &mut Child) {
+    let Some(process_group_id) = process_group_id else {
+        kill_stdio_child(child);
+        return;
+    };
+
+    let should_escalate = match terminate_process_group(process_group_id) {
+        Ok(exists) => exists,
+        Err(err) => {
```
jif-oai: (found by Codex)
If `killpg(SIGTERM)` fails with anything other than “not found”, we just log it and then hand the child to `wait()`. i.e. dead lock

codex-rs/exec-server/src/rpc.rs
```diff
@@ -229,12 +231,24 @@ pub(crate) struct RpcClient {
     disconnected_rx: watch::Receiver<bool>,
     next_request_id: AtomicI64,
     transport_tasks: Vec<JoinHandle<()>>,
+    _transport_lifetime: Option<TransportLifetime>,
```
jif-oai: This keeps the child alive for the cached client lifetime, not the live connection lifetime. Is it on purpose?

codex-rs/exec-server/src/client.rs
```diff
@@ -930,6 +917,91 @@ mod tests {
             .expect("json-rpc line should write");
     }
 
+    #[cfg(not(windows))]
```
jif-oai, line 938: Could we add a windows test as well? Not sure this is a big deal for now but it might become one

codex-rs/exec-server/src/rpc.rs
```diff
@@ -372,6 +383,16 @@ impl Drop for RpcClient {
     }
 }
 
+fn drop_lifetime_on_disconnect(
+    lifetime: JsonRpcTransportLifetime,
+    mut disconnected_rx: watch::Receiver<bool>,
+) -> JoinHandle<()> {
+    tokio::spawn(async move {
+        let _lifetime = lifetime;
+        let _ = disconnected_rx.changed().await;
```
jif-oai: This only drops the child on the transport-close path. If the RPC reader stops because of a malformed message or an unexpected server request, the client is already dead but this watch never flips, so the stdio child keeps running

codex-rs/exec-server/src/client.rs
```diff
@@ -180,29 +191,23 @@ pub struct ExecServerClient {
...
-            .get_or_try_init(|| async {
-                ExecServerClient::connect_websocket(RemoteExecServerConnectArgs {
-                    websocket_url: self.websocket_url.clone(),
-                    client_name: "codex-environment".to_string(),
-                    connect_timeout: Duration::from_secs(5),
-                    initialize_timeout: Duration::from_secs(5),
-                    resume_session_id: None,
-                })
-                .await
+            .get_or_try_init(|| {
```
jif-oai, line 210: For StdioCommand, once the connection dies we now kill the child, but this OnceCell still keeps returning the same dead client forever

codex-rs/exec-server/src/client_transport.rs
jif-oai: Still valid no?

codex-rs/exec-server/src/client_transport.rs
```diff
@@ -0,0 +1,218 @@
...
+
+impl Drop for StdioChildGuard {
+    fn drop(&mut self) {
+        let Some(mut child) = self.child.take() else {
+            return;
+        };
+
+        terminate_stdio_child_process(self.process_id, &mut child);
+
+        if let Ok(handle) = Handle::try_current() {
```
jif-oai: If the last client clone is dropped off a tokio runtime, we skip `wait()` entirely. That kills the child but can leave it live on Unix, so we leak a zombie until Codex exits

codex-rs/exec-server/src/client_transport.rs
```diff
@@ -0,0 +1,218 @@
...
+        Ok(exists) => exists,
+        Err(err) => {
+            debug!("failed to terminate exec-server stdio process group {process_group_id}: {err}");
+            kill_stdio_child(child);
+            return;
+        }
+    };
+    if should_escalate {
+        spawn(move || {
+            sleep(STDIO_CHILD_TERM_GRACE_PERIOD);
```
jif-oai: This delayed `killpg` is detached from the child handle
What if the leader get reaped before the timeout? The PGID can be re-used and we race

jif-oai (commented):
Here stdio is not just another transport... it owns a process tree and generally, handling processes in UNIX is extremely racy. There should be some code around unified exec or pty we can re-use or take inspiration

pakrym-oai (approved):

codex-rs/exec-server/src/client_transport.rs
```diff
@@ -0,0 +1,125 @@
...
+            ),
+            args.into(),
+        )
+        .await
+    }
+
+    pub(crate) async fn connect_stdio_command(
+        args: StdioExecServerConnectArgs,
+    ) -> Result<Self, ExecServerError> {
+        let mut child = stdio_command_process(&args.command)
```
jif-oai, line 79: Can we avoid launching this as a direct child only? If the command is a wrapper that forks the real server, dropping the client only kills the wrapper and leaves the rest of the tree behind... very classic process leak

codex-rs/exec-server/src/rpc.rs
```diff
@@ -229,12 +227,19 @@ pub(crate) struct RpcClient {
     disconnected_rx: watch::Receiver<bool>,
     next_request_id: AtomicI64,
     transport_tasks: Vec<JoinHandle<()>>,
+    _transport: JsonRpcTransport,
```
jif-oai: This still keeps the stdio child for the cached client lifetime, not the live connection lifetime

codex-rs/exec-server/src/connection.rs
```diff
@@ -22,11 +24,55 @@ pub(crate) enum JsonRpcConnectionEvent {
...
+impl Drop for StdioTransport {
+    fn drop(&mut self) {
+        let Some(mut child_process) = self.child_process.take() else {
+            return;
+        };
+
+        if let Err(err) = child_process.start_kill() {
+            debug!("failed to terminate exec-server stdio child: {err}");
+        }
+        match tokio::runtime::Handle::try_current() {
```
jif-oai: If the last drop happens off-runtime, we kill the child but never reap it. That leaves an orphan (on UNIX at least) around until Codex exits

codex-rs/exec-server/src/client_api.rs
```diff
@@ -25,6 +27,32 @@ pub struct RemoteExecServerConnectArgs {
...
+    pub args: Vec<String>,
+    pub env: HashMap<String, String>,
+    pub cwd: Option<PathBuf>,
+}
+
+/// Parameters used to connect to a remote exec-server environment.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) enum ExecServerTransportParams {
+    WebSocketUrl(String),
+    #[allow(dead_code)]
```
jif-oai, line 52: Can we drop if this is dead code?

PR #20665, opened 2026-05-01
Make environment providers own default selection

starr-openai:
## Why

The next PR in this stack introduces configured environments, where the provider knows both which environments exist and which one should be selected by default. The existing manager derived the default internally by checking for the legacy `remote` and `local` ids, and it treated "remote" as equivalent to "has a websocket URL." That does not work cleanly for stdio-command remotes because they are remote environments without an `exec_server_url`.

**Stack position:** this is PR 3 of 5. It is the environment-model bridge between PR 2's transport enum and PR 4's TOML provider.

## What Changed

- Add `DefaultEnvironmentSelection` to the `EnvironmentProvider` contract:
  - `Derived` preserves the old `remote`-then-`local` fallback behavior.
  - `Environment(id)` lets a provider explicitly select a configured default.
  - `Disabled` lets a provider intentionally expose no default environment.
- Move the legacy `CODEX_EXEC_SERVER_URL=none` default-disabling behavior into `DefaultEnvironmentProvider`.
- Make `EnvironmentManager` validate explicit provider defaults and return an error if the selected id is missing.
- Track `remote_transport` separately from `exec_server_url` so stdio-command environments are still recognized as remote.
- Add `Environment::remote_stdio_shell_command(...)` for the TOML provider added in the next PR.

## Stack

- **3. This PR:** https://github.com/openai/codex/pull/20665 - Make environment providers own default selection

## Validation

Not run locally; this was split out of the original draft stack.

### DIFF
--- a/codex-rs/exec-server/src/environment.rs
+++ b/codex-rs/exec-server/src/environment.rs
@@ -73,12 +73,14 @@ impl EnvironmentManager {
 
     /// Builds a test-only manager with environment access disabled.
     pub fn disabled_for_tests(local_runtime_paths: ExecServerRuntimePaths) -> Self {
-        Self::from_environments(
+        match Self::from_environments(
             HashMap::new(),
             local_runtime_paths,
             DefaultEnvironmentSelection::Disabled,
-        )
-        .expect("disabled test environment manager")
+        ) {
+            Ok(manager) => manager,
+            Err(err) => panic!("disabled test environment manager: {err}"),
+        }
     }
 
     /// Builds a test-only manager from a raw exec-server URL value.
@@ -105,12 +107,14 @@ impl EnvironmentManager {
     ) -> Self {
         let provider = DefaultEnvironmentProvider::new(exec_server_url);
         let provider_environments = provider.environments(&local_runtime_paths);
-        Self::from_environments(
+        match Self::from_environments(
             provider_environments,
             local_runtime_paths,
             provider.default_environment_selection(),
-        )
-        .expect("default provider should create valid environments")
+        ) {
+            Ok(manager) => manager,
+            Err(err) => panic!("default provider should create valid environments: {err}"),
+        }
     }
 
     /// Builds a manager from a provider-supplied startup snapshot.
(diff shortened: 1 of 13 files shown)


codex-rs/exec-server/src/environment.rs
```diff
@@ -275,15 +259,28 @@ impl Environment {
...
+        )
+    }
+
+    fn remote_with_transport(
+        transport: ExecServerTransport,
+        local_runtime_paths: Option<ExecServerRuntimePaths>,
+    ) -> Self {
+        let exec_server_url = match &transport {
+            ExecServerTransport::WebSocketUrl(url) => Some(url.clone()),
+            ExecServerTransport::StdioShellCommand(_) => None,
```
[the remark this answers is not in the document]
jif-oai: I guess will be fixed in a follow-up?

codex-rs/exec-server/src/environment_provider.rs
```diff
@@ -11,16 +11,20 @@ use crate::environment::REMOTE_ENVIRONMENT_ID;
...
 pub trait EnvironmentProvider: Send + Sync {
     /// Returns the environments available for a new manager.
     async fn get_environments(
         &self,
         local_runtime_paths: &ExecServerRuntimePaths,
     ) -> Result<HashMap<String, Environment>, ExecServerError>;
+
+    /// Returns the provider-selected default environment id, or `None` to
+    /// disable model-facing environment access.
+    fn default_environment_id(&self) -> Option<String>;
```
jif-oai: This loses the old “derive the default for me” state entirely... `None` now means “disable”. I guess it's safe but we need to be 100% sure this is handled properly (IDK if we have other providers yet)

codex-rs/exec-server/src/environment.rs
```diff
@@ -99,36 +102,23 @@ impl EnvironmentManager {
...
-            local_runtime_paths,
-        )
-    }
-
-    fn from_provider_environments(
-        environments: HashMap<String, Environment>,
-        local_runtime_paths: ExecServerRuntimePaths,
-    ) -> Result<Self, ExecServerError> {
+        let environments = provider.get_environments(&local_runtime_paths).await?;
+        let default_environment_id = provider.default_environment_id();
```
jif-oai: nit but racy

jif-oai (approved):

codex-rs/exec-server/src/environment_provider.rs
starr-openai: sorta - providers convert omitted/default config into Some(local) explicitly. so we should only end up with None here if the exec server URL explicitly is set to "none"

PR #20666, opened 2026-05-01
Add CODEX_HOME environments TOML provider

starr-openai:
## Why

After stdio transports and provider-owned defaults exist, Codex needs a config-backed provider that can describe more than the single legacy `CODEX_EXEC_SERVER_URL` remote. This PR adds that provider without activating it in product entrypoints yet, keeping parser/validation review separate from runtime wiring.

**Stack position:** this is PR 4 of 5. It builds on PR 3's provider/default model and adds the `environments.toml` provider used by PR 5.

## What Changed

- Add `environment_toml.rs` as the TOML-specific home for parsing, validation, and provider construction.
- Keep the TOML schema/provider structs private; the public constructor added here is `EnvironmentManager::from_codex_home(...)`.
- Add `TomlEnvironmentProvider`, including validation for:
  - reserved ids such as `local` and `none`
  - duplicate ids
  - unknown explicit defaults
  - empty programs or URLs
  - exactly one of `url` or `program` per configured environment
- Support websocket environments with `url = "ws://..."` / `wss://...`.
- Support stdio-command environments with `program = "..."`.
- Add helpers to load `environments.toml` from `CODEX_HOME`, but do not wire entrypoints to call them yet.
- Add the `toml` dependency for parsing.

## Stack

- **4. This PR:** https://github.com/openai/codex/pull/20666 - Add CODEX_HOME environments TOML provider

## Validation

Not run locally; this was split out of the original draft stack.

## Documentation

This introduces the config shape for `environments.toml`; user-facing documentation should be added before this stack is treated as a documented public workflow.

### DIFF
--- a/codex-rs/exec-server/src/environment_toml.rs
+++ b/codex-rs/exec-server/src/environment_toml.rs
@@ -45,6 +45,7 @@ struct TomlEnvironmentProvider {
 }
 
 impl TomlEnvironmentProvider {
+    #[cfg(test)]
     fn new(config: EnvironmentsToml) -> Result<Self, ExecServerError> {
         Self::new_with_config_dir(config, /*config_dir*/ None)
     }
--- a/codex-rs/exec-server/src/environment_toml.rs
+++ b/codex-rs/exec-server/src/environment_toml.rs
@@ -1,0 +1,673 @@
+use std::collections::HashMap;
+use std::collections::HashSet;
+use std::path::Path;
+use std::path::PathBuf;
+
+use serde::Deserialize;
+use tokio_tungstenite::tungstenite::client::IntoClientRequest;
+
+use crate::DefaultEnvironmentProvider;
+use crate::Environment;
+use crate::EnvironmentProvider;
+use crate::ExecServerError;
+use crate::ExecServerRuntimePaths;
+use crate::client_api::ExecServerTransportParams;
+use crate::client_api::StdioExecServerCommand;
+use crate::environment::LOCAL_ENVIRONMENT_ID;
+
+const ENVIRONMENTS_TOML_FILE: &str = "environments.toml";
+const MAX_ENVIRONMENT_ID_LEN: usize = 64;
+
+#[derive(Deserialize, Debug, Default)]
+#[serde(deny_unknown_fields)]
+struct EnvironmentsToml {
+    default: Option<String>,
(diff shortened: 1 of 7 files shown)


codex-rs/exec-server/src/environment_toml.rs
```diff
@@ -0,0 +1,436 @@
...
+use crate::EnvironmentProvider;
+use crate::ExecServerError;
+use crate::ExecServerRuntimePaths;
+use crate::ExecServerTransport;
+use crate::environment::LOCAL_ENVIRONMENT_ID;
+
+const ENVIRONMENTS_TOML_FILE: &str = "environments.toml";
+
+#[derive(Deserialize, Debug, Default)]
+struct EnvironmentsToml {
```
jif-oai, line 26: I think we should take the habit of rejecting unknown fields. Mainly now that we control most of the backend if remote

codex-rs/exec-server/src/environment_toml.rs
```diff
@@ -0,0 +1,436 @@
...
+
+    #[serde(default)]
+    environments: Vec<EnvironmentToml>,
+}
+
+#[derive(Deserialize, Debug)]
+struct EnvironmentToml {
+    id: String,
+    url: Option<String>,
+    command: Option<String>,
```
jif-oai: Once again, this means we loose program, args, env, cwd, ...
Are we sure this is ok?

codex-rs/exec-server/src/environment_toml.rs
```diff
@@ -0,0 +1,436 @@
...
+    }
+    if id == LOCAL_ENVIRONMENT_ID || id.eq_ignore_ascii_case("none") {
+        return Err(ExecServerError::Protocol(format!(
+            "environment id `{id}` is reserved"
+        )));
+    }
+    Ok(())
+}
+
+fn validate_websocket_url(url: String) -> Result<String, ExecServerError> {
```
jif-oai, line 254: We should also validate it's a valid URL

jif-oai (approved):
Ok after my comments

codex-rs/exec-server/src/environment_toml.rs
starr-openai: good point, will update

PR #20687, opened 2026-05-01
[codex] Split tool handlers by tool name

pakrym-oai:
## Why

Tool registration used to bind a tool name to a handler externally, which left ownership split between the registry plan and the handler implementation. Some built-in handlers also multiplexed multiple in-core tools by switching on the invoked tool name internally.

This moves the registry identity onto the handler itself and makes built-in multi-tool areas use separate concrete handlers, so each registered handler instance owns exactly one tool name and one dispatch path.

## What Changed

- Added `ToolHandler::tool_name()` and changed `ToolRegistryBuilder::register_handler` to derive the registry key from the handler.
- Split built-in multiplexed handlers into concrete per-tool handlers for unified exec, shell/local shell/container exec, MCP resources, goal tools, and agent job tools.
- Kept name-carrying handler instances only where the runtime target is inherently external or dynamic, such as MCP tools, dynamic tools, and unavailable placeholders.
- Updated `ToolHandlerKind` and registry-plan construction so plan entries map directly to concrete handler registrations.

## Verification

- `cargo test -p codex-tools tool_registry_plan`
- `cargo test -p codex-core --lib tools::registry_tests`
- `just fix -p codex-tools`
- `just fix -p codex-core`


codex-rs/core/src/tools/handlers/mcp_resource.rs
```diff
@@ -29,8 +29,11 @@ use codex_protocol::protocol::EventMsg;
 use codex_protocol::protocol::McpInvocation;
```
jif-oai, line 32: The split is good but as a result, the begin/execute/end emission envelop seems to be copied 3 times in his file. I wonder if it's possible clean this? That does not look super future proof

codex-rs/core/src/tools/handlers/shell_tests.rs
```diff
@@ -205,20 +205,11 @@ fn shell_command_handler_rejects_login_when_disallowed() {
...
-                "printf hi".to_string(),
-            ],
-            workdir: None,
-            timeout_ms: None,
-            sandbox_permissions: None,
-            prefix_rule: None,
-            additional_permissions: None,
-            justification: None,
-        },
+    let payload = ToolPayload::Function {
```
jif-oai: Looks like this test got shifted off `local_shell` during the split, so we’re no longer covering the new LocalShellHandler hook path here. I’d keep this one pointed at `local_shell` or add a dedicated test, otherwise this refactor’s riskiest behavior change is basically untested

codex-rs/core/src/tools/spec.rs
```diff
@@ -195,116 +184,140 @@ pub(crate) fn build_specs_with_discoverable_tools(
     }
 
     for handler in plan.handlers {
+        let name = handler.name;
```
jif-oai, line 183: Feels like we still have two sources of truth here: the plan says one tool name, the handler hardcodes another, and nothing asserts they stay in sync. That’s pretty easy to break later with a rename and only notice once dispatch starts missing at runtime

codex-rs/core/src/tools/handlers/mcp_resource.rs
pakrym-oai: Was it a bad diff? I only see tool struct definitions being net new in this file.

codex-rs/core/src/tools/spec.rs
pakrym-oai: I plan to delete the ToolHandlerKind and this registration logic as the next step.

codex-rs/core/src/tools/handlers/mcp_resource.rs
pakrym-oai: I guess, emit_tool_call_begin is copied three times but it's not new.

jif-oai (approved):

PR #20689, opened 2026-05-01
Inject state DB, agent graph store

rasmusrygaard:
## Why

We want the agent graph store to be passed down the stack as a real dependency, the same way we already treat the thread store.

This will let us inject the agent graph store as a real dependency and support implementations other than the local SQLite-backed one. Right now most code instantiates a state DB and an agent graph store just-in-time. Ideally, we would not depend on the state DB directly but only read through the higher-level interfaces.

This change makes the dependency boundaries explicit and moves state DB initialization to process bootstrap instead of hiding it inside local store implementations.

## What changed

- `ThreadManager` now requires a `StateDbHandle` and an `AgentGraphStore` at construction time instead of treating them as optional internals.
- The local store constructors no longer lazily initialize SQLite. Callers now initialize the state DB once per process and use that shared handle to build:
  - `LocalThreadStore`
  - `LocalAgentGraphStore`
- App bootstraps (`app-server`, `mcp-server`, `prompt_debug`, and the thread-manager sample) now initialize the state DB up front and inject the resulting handle down the stack.
- `app-server` now consistently uses its process-scoped state DB handle instead of reopening SQLite or trying to recover it from loaded threads.
- Device-key storage now reuses the shared state DB handle instead of maintaining its own lazy opener.
- The thread archive / descendant traversal paths now use the injected `AgentGraphStore` instead of reaching through local thread-store-specific state.

## Verification

- `cargo check -p codex-core -p codex-thread-store -p codex-app-server -p codex-mcp-server -p codex-thread-manager-sample --tests`
- `cargo test -p codex-thread-store`
- `cargo test -p codex-core thread_manager_accepts_separate_agent_graph_store_and_thread_store -- --nocapture`
- `cargo test -p codex-app-server thread_archive_archives_spawned_descendants -- --nocapture`


codex-rs/app-server/src/device_key_api.rs
```diff
@@ -29,22 +29,17 @@ use codex_device_key::RemoteControlClientEnrollmentSignPayload;
...
-use tokio::sync::OnceCell;
 
 #[derive(Clone)]
 pub(crate) struct DeviceKeyApi {
     store: DeviceKeyStore,
 }
 
 impl DeviceKeyApi {
-    pub(crate) fn new(sqlite_home: PathBuf, default_provider: String) -> Self {
+    pub(crate) fn new(state_db: Arc<StateRuntime>) -> Self {
```
pakrym-oai: nit: use StateDbHandle consistently

codex-rs/app-server/src/in_process.rs
```diff
@@ -409,22 +411,60 @@ fn start_uninitialized(args: InProcessStartArgs) -> InProcessClientHandle {
...
-                feedback: args.feedback,
-                log_db: args.log_db,
-                config_warnings: args.config_warnings,
-                session_source: args.session_source,
-                auth_manager,
-                rpc_transport: AppServerRpcTransport::InProcess,
-                remote_control_handle: None,
-                plugin_startup_tasks: crate::PluginStartupTasks::Start,
-            }));
+            let state_db = match &args.config.experimental_thread_store {
```
pakrym-oai: it's a bit fishy to set state_db based on experimental_thread_store. should we leave in_process defaulting to local?

codex-rs/app-server/src/message_processor.rs
```diff
@@ -266,7 +270,7 @@ pub(crate) struct MessageProcessorArgs {
 impl MessageProcessor {
     /// Create a new `MessageProcessor`, retaining a handle to the outgoing
     /// `Sender` so handlers can enqueue messages to be written to stdout.
-    pub(crate) fn new(args: MessageProcessorArgs) -> Self {
+    pub(crate) async fn new(args: MessageProcessorArgs) -> Self {
         let MessageProcessorArgs {
```
pakrym-oai, line 271: does this need to become async?

codex-rs/app-server/src/message_processor.rs
```diff
@@ -289,14 +294,19 @@ impl MessageProcessor {
...
+        let thread_store = thread_store_from_config(config.as_ref(), state_db.clone());
+        let agent_graph_store = agent_graph_store_from_config(config.as_ref(), state_db.clone());
         let thread_manager = Arc::new(ThreadManager::new(
             config.as_ref(),
             auth_manager.clone(),
             session_source,
             environment_manager,
             Some(analytics_events_client.clone()),
-            Arc::clone(&thread_store),
+            ThreadManagerPersistence {
```
pakrym-oai: nit: maybe drop the extra struct?

codex-rs/app-server/src/lib.rs
```diff
@@ -647,24 +644,17 @@ pub async fn run_main_with_transport_options(
         AuthManager::shared_from_config(&config, /*enable_codex_api_key_env*/ false).await;
 
     let remote_control_config_enabled = config.features.enabled(Feature::RemoteControl);
-    let remote_control_enabled = remote_control_config_enabled && state_db.is_some();
-    if remote_control_config_enabled && state_db.is_none() {
-        error!("remote control disabled because sqlite state db is unavailable");
-    }
+    let remote_control_enabled = remote_control_config_enabled;
```
pakrym-oai: nit: rem

vmy7zg24hc-maker:
dev/rasmus/agent_store_interface_plumbing

codex-rs/thread-store/src/local/mod.rs
```diff
@@ -76,30 +75,19 @@ impl std::fmt::Debug for LocalThreadStore {
 }
 
 impl LocalThreadStore {
-    /// Create a local store from process-scoped local storage configuration.
-    pub fn new(config: LocalThreadStoreConfig) -> Self {
+    /// Create a local store from process-scoped local storage configuration and
+    /// the caller-provided shared state DB handle.
+    pub fn new(config: LocalThreadStoreConfig, state_db: StateDbHandle) -> Self {
```
pakrym-oai, line 78: LocalThreadStoreConfig has sqlite home on it

codex-rs/mcp-server/src/message_processor.rs
```diff
@@ -60,13 +63,22 @@ impl MessageProcessor {
             /*enable_codex_api_key_env*/ false,
         )
         .await;
+        let Some(state_db) = init_state_db_from_config(config.as_ref()).await else {
+            panic!("mcp server requires state db");
```
pakrym-oai: we usually have no panics in prod code

codex-rs/core/src/thread_manager.rs
```diff
@@ -263,25 +270,90 @@ pub fn build_models_manager(
...
+
+/// Process-scoped persistence dependencies required by [`ThreadManager`].
+#[derive(Clone)]
+pub struct ThreadManagerPersistence {
+    pub state_db: StateDbHandle,
+    pub thread_store: Arc<dyn ThreadStore>,
+    pub agent_graph_store: Arc<dyn AgentGraphStore>,
+}
+
+fn state_db_from_roots_for_tests(
```
pakrym-oai: this method probably needs to be async

codex-rs/core/src/thread_manager.rs
```diff
@@ -263,25 +270,90 @@ pub fn build_models_manager(
...
         ThreadStoreConfig::Remote { endpoint } => Arc::new(RemoteThreadStore::new(endpoint)),
         ThreadStoreConfig::InMemory { id } => InMemoryThreadStore::for_id(id),
     }
 }
 
+pub fn agent_graph_store_from_state_db(state_db: StateDbHandle) -> Arc<dyn AgentGraphStore> {
+    Arc::new(LocalAgentGraphStore::new(state_db))
+}
+
+pub fn agent_graph_store_from_config(
```
pakrym-oai: same q about deciding store based on experimental_thread_store

codex-rs/core/src/thread_manager.rs
```diff
@@ -263,25 +270,90 @@ pub fn build_models_manager(
...
             LocalThreadStoreConfig::from_config(config),
+            state_db,
         )),
         ThreadStoreConfig::Remote { endpoint } => Arc::new(RemoteThreadStore::new(endpoint)),
         ThreadStoreConfig::InMemory { id } => InMemoryThreadStore::for_id(id),
     }
 }
 
+pub fn agent_graph_store_from_state_db(state_db: StateDbHandle) -> Arc<dyn AgentGraphStore> {
+    Arc::new(LocalAgentGraphStore::new(state_db))
```
pakrym-oai, line 287: nit: agent_graph_store_from_state_db is 3 characters shorter than Arc::new(LocalAgentGraphStore::new, may be worth inlining

codex-rs/core/src/thread_manager.rs
```diff
@@ -836,6 +915,16 @@ impl ThreadManager {
     }
 }
 
+impl ThreadManagerState {
```
pakrym-oai, line 903: nit: there is an impl block just below

codex-rs/core/src/stream_events_utils.rs
```diff
@@ -174,7 +174,8 @@ pub(crate) async fn mark_thread_memory_mode_polluted_if_external_context(
 }
 
 async fn record_stage1_output_usage_and_detect_memory_citation(
-    turn_context: &TurnContext,
+    sess: &Session,
+    _turn_context: &TurnContext,
```
pakrym-oai: delete unused

codex-rs/core/src/prompt_debug.rs
```diff
@@ -35,13 +38,22 @@ pub async fn build_prompt_input(
         config.codex_linux_sandbox_exe.clone(),
     )?;
 
+    let Some(state_db) = init_state_db_from_config(&config).await else {
```
pakrym-oai: ?.map_err?

codex-rs/core/src/session/tests.rs
```diff
@@ -2331,8 +2331,8 @@ async fn thread_rollback_fails_when_num_turns_is_zero() {
 
 #[tokio::test]
 async fn set_rate_limits_retains_previous_credits() {
-    let codex_home = tempfile::tempdir().expect("create temp dir");
-    let config = build_test_config(codex_home.path()).await;
+    let codex_home = tempfile::tempdir().expect("create temp dir").keep();
```
pakrym-oai: fishy?

codex-rs/core/src/session/session.rs
```diff
@@ -442,22 +443,12 @@ impl Session {
...
-            } else {
-                None
-            }
-        }
-        .instrument(info_span!(
-            "session_init.state_db",
-            otel.name = "session_init.state_db",
-            session_init.ephemeral = config.ephemeral,
-        ));
+        let state_db_fut =
```
pakrym-oai: this doesn't have to be a future, does it?

pakrym-oai (commented):
Minor comments.

codex-rs/core/src/thread_manager.rs
rasmusrygaard: Yeah, we should have a separate flag here. Can we take in a follow-up to make the scope of this PR more manageable?

codex-rs/app-server/src/request_processors/thread_goal_processor.rs
```diff
@@ -74,18 +74,10 @@ impl ThreadGoalRequestProcessor {
 
     pub(crate) async fn pending_resume_goal_state(
         &self,
-        thread: &CodexThread,
+        _thread: &CodexThread,
```
pakrym-oai: delete?

codex-rs/app-server/src/in_process.rs
```diff
@@ -414,6 +417,12 @@ fn start_uninitialized(args: InProcessStartArgs) -> InProcessClientHandle {
         );
         let (processor_tx, mut processor_rx) = mpsc::channel::<ProcessorCommand>(channel_capacity);
         let mut processor_handle = tokio::spawn(async move {
+            let Some(state_db) = state_db else {
+                warn!(
+                    "in-process app-server state db initialization failed; shutting down processor task"
```
pakrym-oai, line 424: will we observe this correctly?

codex-rs/thread-store/src/local/live_writer.rs
```diff
@@ -66,12 +66,12 @@ pub(super) async fn resume_thread(
         })?;
     let config = RolloutConfig {
         codex_home: store.config.codex_home.clone(),
-        sqlite_home: store.config.sqlite_home.clone(),
+        sqlite_home: store.sqlite_home(),
```
pakrym-oai, line 69: why do need sqlite home is we already pass the state_db_ctx in?

codex-rs/core/src/stream_events_utils.rs
```diff
@@ -177,7 +174,7 @@ pub(crate) async fn mark_thread_memory_mode_polluted_if_external_context(
 }
 
 async fn record_stage1_output_usage_and_detect_memory_citation(
-    state_db_ctx: Option<&state_db::StateDbHandle>,
+    sess: &Session,
```
pakrym-oai: nit: pass in StateDbHandle?

codex-rs/core/src/thread_manager_tests.rs
```diff
@@ -47,6 +48,99 @@ fn assistant_msg(text: &str) -> ResponseItem {
...
+    let state_db = init_state_db_from_config(config)
+        .await
+        .expect("thread manager test requires state db");
+    let thread_store = thread_store_from_config(config, state_db.clone());
+    let agent_graph_store = agent_graph_store_from_config(config, state_db.clone());
+    (state_db, thread_store, agent_graph_store)
+}
+
+#[tokio::test]
+async fn thread_manager_accepts_separate_agent_graph_store_and_thread_store() {
```
pakrym-oai: is this a useful test?

codex-rs/core/src/thread_manager_tests.rs
```diff
@@ -47,6 +48,99 @@ fn assistant_msg(text: &str) -> ResponseItem {
...
+    std::fs::create_dir_all(&config.codex_home).expect("create codex home");
+
+    let state_db = init_state_db_from_config(&config)
+        .await
+        .expect("thread manager test requires state db");
+    let _ = agent_graph_store_from_state_db(state_db);
+}
+
+#[tokio::test]
+async fn agent_graph_store_from_config_uses_remote_endpoint() {
```
pakrym-oai: is this a useful test?

pakrym-oai (approved):

codex-rs/thread-store/src/local/live_writer.rs
rasmusrygaard: I'm not sure, looks like this was already the way this worked? Do you think this PR is the time to clean that up?

PR #20692, opened 2026-05-01
Support PreToolUse additionalContext

abhinav-oai:
# Why

`PreToolUse` already exposes `hookSpecificOutput.additionalContext` in the generated hook schema, but the runtime still rejected it as unsupported. That leaves `PreToolUse` out of step with the other context-injecting hooks and prevents hook authors from attaching model-visible guidance to a pending tool call before it runs.

# What

- Parse `PreToolUse.additionalContext` and carry it through the hook event pipeline.
- Record `PreToolUse` context at the hook boundary so successful context is preserved for both allowed and blocked calls without widening the tool registry surface.
- Preserve existing deny behavior when context is combined with either `permissionDecision: "deny"` or the legacy `decision: "block"` shape.


abhinav-oai:
@codex review

codex-rs/core/src/tools/registry.rs
```diff
@@ -354,20 +354,47 @@ impl ToolRegistry {
...
+            if let Some(pre_tool_use_payload) = handler.pre_tool_use_payload(&invocation) {
+                let outcome = run_pre_tool_use_hooks(
+                    &invocation.session,
+                    &invocation.turn,
+                    invocation.call_id.clone(),
+                    &pre_tool_use_payload.tool_name,
+                    &pre_tool_use_payload.tool_input,
+                )
+                .await;
+                if outcome.should_block
```
abhinav-oai: some of this logic moved here from `codex-rs/core/src/hook_runtime.rs`

didn't love it but that's the consequence of returning `PreToolUseOutcome` to carry the `additionalContext`

alternative was to create something like `PreToolUseRuntimeOutcome` (similar to `HookRuntimeOutcome`) but didn't wanna introduce a new concept here

abhinav-oai:
@codex review

codex-rs/core/src/hook_runtime.rs
```diff
@@ -161,8 +161,10 @@ pub(crate) async fn run_pre_tool_use_hooks(
         hook_events,
         should_block,
         block_reason,
+        additional_contexts,
     } = hooks.run_pre_tool_use(request).await;
     emit_hook_completed_events(sess, turn_context, hook_events).await;
+    record_additional_contexts(sess, turn_context, additional_contexts).await;
```
sayan-oai, line 167: is this uncapped? maybe an existing/broader concern (though `PreToolUse` is probably a high volume hook), but allowing injection of an uncapped amount of context directly into a dev message seems like an easy way for users to unknowingly blow up their own context window

sayan-oai (approved):
this pr logic seems fine; discussed in slack implementing a hook-level context injection cap as a follow up.

abhinav-oai:
merging as we got https://github.com/openai/codex/pull/21069 in

PR #20702, opened 2026-05-01
Support PreToolUse permissionDecision ask

abhinav-oai:
## Why

`PreToolUse` hooks can already block a tool call with `permissionDecision: deny`, but they cannot currently tighten an otherwise-permitted call into an explicit human approval.

This adds that control without letting a pre-hook weaken core policy before Codex reaches the concrete approval boundary for the tool call.

## What

- support `permissionDecision: "ask"` for `PreToolUse` output
  - surface the hook-authored reason to the user without adding it to model-visible context
- record hook-authored `ask` decisions once in turn-local state keyed by `tool_use_id`, then apply them at the concrete approval boundaries for that call
- accept `permissionDecision: "allow"` for compatibility, but ignore it so a pre-hook cannot waive prompts or later concrete checks
- reserve `permissionDecision: "defer"` in the schema for future Codex exec support while keeping it unimplemented today

## Precedence

| Case | Result |
| --- | --- |
| `PreToolUse: ask` | Route to the human user for that tool call, bypassing cached approvals, `PermissionRequest` hooks, and auto review |
| `PreToolUse: allow` | Accepted for compatibility and ignored |
| Core policy / strict auto review | Remain authoritative; `PreToolUse` cannot weaken them |


codex-rs/core/src/session/review.rs
```diff
@@ -146,6 +146,7 @@ pub(super) async fn spawn_review_thread(
         codex_linux_sandbox_exe: parent_turn_context.codex_linux_sandbox_exe.clone(),
         tool_call_gate: Arc::new(ReadinessFlag::new()),
         dynamic_tools: parent_turn_context.dynamic_tools.clone(),
+        pre_tool_use_permission_decisions: Default::default(),
```
abhinav-oai, line 149: [note] tracking the state here felt nicer than plumbing it all around

codex-rs/core/src/tools/orchestrator.rs
```diff
@@ -167,48 +192,50 @@ impl ToolOrchestrator {
...
-                    tool_ctx.call_id.as_str(),
-                    approval_ctx,
-                    tool_ctx,
-                    /*evaluate_permission_request_hooks*/ !strict_auto_review,
-                    &otel,
-                )
-                .await?;
-
-                Self::reject_if_not_approved(tool_ctx, guardian_review_id.as_deref(), decision)
+                if pre_tool_use_allows {
```
viyatb-oai: This makes `PreToolUse::Allow` outrank the `NeedsApproval` result we just computed. Once policy has already said this specific call needs approval, skipping the prompt here means the hook can waive an existing ask requirement. Could we preserve existing ask requirements and only use `allow` when there is not one?

codex-rs/core/src/tools/network_approval.rs
```diff
@@ -496,8 +513,17 @@ impl NetworkApprovalService {
                 }
             }
         }
-        let use_guardian = routes_approval_to_guardian(&turn_context);
-        let guardian_review_id = use_guardian.then(new_guardian_review_id);
+        if pre_tool_use_allows {
```
viyatb-oai: `PreToolUse::Allow` was decided from the tool input, but here we reuse it to auto-approve a later host-specific network denial. At this point we know the actual host/protocol/port, which the pre-tool hook did not have when it returned `allow`, so this can grant network access the hook never evaluated. Could we keep host approvals in the network/permission-request path instead of carrying a tool-level allow across this boundary?

jignaciofernandez:
I would like to see this reopened or landed. The current behavior parses `permissionDecision: "ask"` but still treats it as unsupported (`PreToolUse hook returned unsupported permissionDecision:ask`), so a `PreToolUse` hook can deny a tool call but cannot escalate an otherwise-allowed call into a native human approval prompt.

Use case: policy hooks that need to force explicit user approval for specific commands/tools without globally tightening the whole Codex approval policy, and without weakening core policy. This is especially important in Codex App/CLI setups where hooks are used as a durable safety boundary.

The design in this PR looks like the right direction: `PreToolUse: ask` should route that specific tool call to the human user, while `allow` must not bypass core policy or later concrete approval checks. Please consider reopening this PR or tracking an equivalent focused implementation for native `PreToolUse permissionDecision: "ask"` support in Codex App/CLI.

PR #20703, opened 2026-05-01
Support updatedToolOutput for PostToolUse

abhinav-oai:
# Why

`PostToolUse` hooks can already inspect the stable `tool_response` that Codex exposes after a tool runs, but they cannot currently replace the result that the model sees on the next turn. Supporting the existing `updatedToolOutput` / `updatedMCPToolOutput` hook fields lets hook authors redact or normalize direct model-visible tool output without mutating the underlying typed tool result.

That distinction matters for code mode: nested tools can be chained together inside a running program, so rewriting an intermediate value would change program behavior and break the caller's expectation about what the previous tool actually returned.

# What

- parse `updatedToolOutput` and/or `updatedMCPToolOutput` from `PostToolUse` hook output
  - `updatedToolOutput` takes precedence over `updatedMCPToolOutput`
- validate replacements by JSON kind against the original hook-facing `tool_response`; invalid replacements are ignored with a hook warning
- preserve the original typed `ToolOutput` while applying accepted replacements only to the direct model-visible response item
- keep code-mode nested tool results unchanged, so code mode continues to receive original structured outputs while direct model calls can see the rewritten projection
- cover the split with direct shell/MCP rewrite tests, invalid-shape coverage, and a code-mode regression that proves nested `exec_command` output is not rewritten

# Documentation

The public hook spec should document that `updatedToolOutput` / `updatedMCPToolOutput` affect the model-visible projection only and do not rewrite code-mode intermediate results.


abhinav-oai:
@codex review

abhinav-oai:
One thing to figure out: what should `updatedToolOutput` / `updatedMCPToolOutput` mean for code-mode callers?

Today `PostToolUse` runs below the direct-vs-code-mode split, but this implementation only rewrites the model-visible projection; `code_mode_result()` still returns the original typed result. That is coherent if the contract is intentionally "rewrite what the model sees, not the underlying tool result," and it preserves structured MCP results for code-mode consumers. But the other plausible reading is "rewrite the tool output for whoever called the tool," which would imply code mode should observe the rewritten value too.

I think we should make that choice explicit in the spec/docs and add a targeted code-mode test for whichever behavior we want, especially because redaction/sanitization hooks make the distinction observable.

abhinav-oai:
@codex review

PR #20708, opened 2026-05-02
Add Windows sandbox readiness RPC

iceweasel-oai:
## Why

The desktop app on Windows needs a read-only way to tell, before the next tool call, whether the local Windows sandbox setup is in a state that should block the user and ask for setup again.

The main case we want to cover is the elevated sandbox setup version bump. Today, if the app is configured for elevated Windows sandboxing and the installed setup is stale, the next sandboxed shell/exec path can end up triggering the elevated setup flow directly. That means the user can see an unexpected UAC prompt with no UI explanation.

This change adds a small app-server preflight so the desktop app can ask “is Windows sandbox ready, not configured, or update-required?” during startup and show the appropriate blocking UI before the user hits a tool call.

## What changed

- Added a new read-only app-server RPC: `windowsSandbox/readiness`
- Added a new protocol enum and response type:
  - `WindowsSandboxReadiness`
  - `WindowsSandboxReadinessResponse`
- Added core readiness logic in `core/src/windows_sandbox.rs`:
  - `ready`
  - `notConfigured`
  - `updateRequired`
- Wired the new request through `codex_message_processor`
- Regenerated the vendored app-server schema fixtures

## Readiness semantics

This is intentionally a coarse startup/version-bump readiness check, not a full predictor of every runtime repair case.

For now, readiness is determined from:
- the configured Windows sandbox level
- `sandbox_setup_is_complete()` for elevated mode

That means:
- `disabled` maps to `notConfigured`
- `restricted token` maps to `ready`
- `elevated` maps to `ready` or `updateRequired` depending on `sandbox_setup_is_complete()`

This is deliberate for the first UI integration because the common case we want to catch is “the app updated, the elevated setup version bumped, and the user should see an update-required blocker instead of a surprise UAC prompt”.

It does not attempt to model every case where the deeper runtime path might decide to repair or re-run setup.

## Testing

- Ran `cargo fmt --all -- app-server-protocol/src/protocol/common.rs app-server-protocol/src/protocol/v2.rs app-server/src/codex_message_processor.rs core/src/windows_sandbox.rs core/src/windows_sandbox_tests.rs`
- Added unit tests for the pure readiness mapping in `core/src/windows_sandbox_tests.rs`
- Regenerated vendored schema fixtures with `cargo run -p codex-app-server-protocol --bin write_schema_fixtures -- --schema-root app-server-protocol/schema`
- Did not run the full cargo test suite


iceweasel-oai:
@codex review

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -7139,6 +7139,15 @@ pub enum WindowsSandboxSetupMode {
     Unelevated,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub enum WindowsSandboxReadiness {
+    Ready,
+    NotConfigured,
+    UpdateRequired,
```
dylan-hurd-oai, line 7198: Should this include details on the version?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -7139,6 +7139,15 @@ pub enum WindowsSandboxSetupMode {
     Unelevated,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub enum WindowsSandboxReadiness {
+    Ready,
+    NotConfigured,
```
dylan-hurd-oai, line 7197: Should we distinguish between "Not yet configured", "attempted to configure but failed", and "user opted not to configure"?

codex-rs/app-server-protocol/src/protocol/v2.rs
```diff
@@ -7155,6 +7164,13 @@ pub struct WindowsSandboxSetupStartResponse {
     pub started: bool,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, JsonSchema, TS)]
+#[serde(rename_all = "camelCase")]
+#[ts(export_to = "v2/")]
+pub struct WindowsSandboxReadinessResponse {
```
dylan-hurd-oai, line 7220: Does this need its own endpoint? should we have something more generic, or is there an advantage to isolating this call?

codex-rs/core/src/windows_sandbox.rs
```diff
@@ -56,6 +56,41 @@ pub fn windows_sandbox_level_from_features(features: &Features) -> WindowsSandbo
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum WindowsSandboxReadiness {
+    Ready,
+    NotConfigured,
+    UpdateRequired,
+}
+
+pub fn determine_windows_sandbox_readiness(config: &Config) -> WindowsSandboxReadiness {
+    if !cfg!(target_os = "windows") {
+        return WindowsSandboxReadiness::Ready;
```
dylan-hurd-oai: not configured?

codex-rs/core/src/windows_sandbox.rs
```diff
@@ -56,6 +56,41 @@ pub fn windows_sandbox_level_from_features(features: &Features) -> WindowsSandbo
...
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum WindowsSandboxReadiness {
+    Ready,
+    NotConfigured,
+    UpdateRequired,
+}
+
+pub fn determine_windows_sandbox_readiness(config: &Config) -> WindowsSandboxReadiness {
```
dylan-hurd-oai: does this logic need to be in core? it seems like it only depends on config

codex-rs/app-server-protocol/src/protocol/v2.rs
iceweasel-oai: like the version from the marker file? I don't think the app cares about that, could add it later if there's a need

codex-rs/app-server-protocol/src/protocol/v2.rs
iceweasel-oai: users no longer have the choice to not configure. Similarly, open to adding more fields in the future but right now this RPC is really just for "should I block the UI for sandbox setup/upgrade"

codex-rs/core/src/windows_sandbox.rs
iceweasel-oai: yeah probably. The Mac app really shouldn't call this... but not configured is a bit more precise

codex-rs/app-server-protocol/src/protocol/v2.rs
iceweasel-oai: mostly just trying to keep this self contained and not pollute other endpoints. I considered adding it to app-server init, but felt weird to have some random windows fields in there. I guess I could imagine a more generic "some setup is required by the user" RPC but it felt like over-engineering given that currently there isn't anything like that.

codex-rs/core/src/windows_sandbox.rs
iceweasel-oai: yeah good point, this is a logical file to put it in, but I agree that adding stuff to core isn't always best. I'll take a look at a better location

iceweasel-oai: moved to `codex-rs/app-server/src/request_processors/windows_sandbox_processor.rs`

codex-rs/core/src/windows_sandbox.rs
iceweasel-oai: fixed

dylan-hurd-oai (approved):

PR #20709, opened 2026-05-02
Use responses request helpers for compact requests

aibrahim-oai:
## Why

`/responses/compact` was still building its own request payload and shared headers separately from `/responses`. That made it easy for new request fields or header behavior to land on `/responses` without reaching compaction, even though we want the compact request shape to stay identical to the normal responses path.

## What changed

- route remote compaction through the shared `ResponsesApiRequest` builder in `core`
- reuse shared body and header serialization helpers in `codex-api` for both `/responses` and `/responses/compact`
- remove the separate compact payload type so compact no longer has its own request body shape
- replace the manual client parity test with a `core` integration test that drives the real helper path, compares the last normal `/responses` request with the following `/responses/compact` request after five varied turns, and snapshots the resulting `input` diff

## Verification

- `cargo check -p codex-api -p codex-core --lib`


aibrahim-oai:
Closing this one in favor of #20719, which is the direct PR from openai/codex:aibrahim/compact-request-parity to main.

